PR #21, opened 2016-03-02
Initial setup for micromasters portal

### LINKED ISSUE #9: Set up initial app
start with cookie cutter app

noisecapella:
Fixes #9

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -1,0 +1,151 @@
+"""
+Django settings for ui pp. This is just a harness type
+project for testing and interacting with the app.
+
+
+For more information on this file, see
+https://docs.djangoproject.com/en/1.8/topics/settings/
+
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/1.8/ref/settings/
+
+"""
+import ast
+import os
+
+import dj_database_url
+import yaml
+
+CONFIG_PATHS = [
+    os.environ.get('MICROMASTERS_CONFIG', ''),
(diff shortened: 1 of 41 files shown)


micromasters/settings.py
```diff
@@ -0,0 +1,151 @@
...
+
+INSTALLED_APPS = (
+    'django.contrib.admin',
+    'django.contrib.auth',
+    'django.contrib.contenttypes',
+    'django.contrib.sessions',
+    'django.contrib.messages',
+    'django.contrib.staticfiles',
+    # Our INSTALLED_APPS
+    'ui'
```
noisecapella: We would have that problem with any name, wouldn't it? I'm not saying that all ui has to go within this app, it's just a place to start writing code

ui/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+"""ui
```
noisecapella: It's from the cookiecutter, I'll remove it

ui/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+Web Portal for MicroMasters
+"""
+
+__project__ = 'ui'
+__version__ = '0.0.0'
```
noisecapella: It might be best to just get rid of this and make it blank

bdero:
#24 is live at https://micromasters.herokuapp.com/, which is a child PR of this PR. :+1:

PR #28, opened 2016-03-03
Course models and api

### LINKED ISSUE #20: create model and rest api for micromaster_courses
as a learner, I need to be able to see a list of micromasters' courses. 

model needs to include:

- [ ] edx course key, if it exists
- [ ] title
- [ ] enrollment_start
- [ ] start_date
- [ ] enrollment url
- [ ] order (maybe just display order)
- [ ] prerequisites

justinabrahms:

Fixes #20 

#### What's this PR do?

Adds the baseline models for the courses and programs, as well as presents a simple REST API.

models.py -> api.py

Query the API.

http://localhost:8079/api/courses/
http://localhost:8079/api/programs/

There were some additional changes to support coverage.


noisecapella:
I'll start reviewing this but feel free to ignore my comments if something is a work in progress

micromasters/urls.py
```diff
@@ -1,22 +1,16 @@
...
+from courses.api import ProgramViewSet, CourseViewSet
+
+router = routers.DefaultRouter()
+router.register(r'programs', ProgramViewSet)
+router.register(r'courses', CourseViewSet)
 
 urlpatterns = [
     url(r'', include('ui.urls')),
     url(r'^admin/', include(admin.site.urls)),
+    url(r'^api/', include(router.urls)),
```
noisecapella: Can this be `api/v1` instead?

courses/models_test.py
```diff
@@ -0,0 +1,24 @@
...
+from .factories import ProgramFactory, CourseFactory
+
+
+class ProgramTests(TestCase):
+    """Tests for Program model"""
+
+    def test_to_string(self):
+        """Test for __str__ method"""
+        prog = ProgramFactory.build(title="Title")
+        assert "%s" % prog == "Title"
```
noisecapella: Can we use `format` here and below?

courses/models.py
```diff
@@ -0,0 +1,31 @@
...
+    A degree someone can pursue, e.g. "Supply Chain Management"
+    """
+    title = models.CharField(max_length=255)
+    live = models.BooleanField(default=False)
+
+    def __str__(self):
+        return self.title
+
+
+class Course(models.Model):
```
noisecapella: I was almost about to ask you to use `@python_2_unicode_compatible` here :)

micromasters/urls.py
justinabrahms: Do you think that comes with backwards compatibility guarantees? That's why I didn't put it there to begin with, given that this will likely evolve rather quickly.

noisecapella: In that case how about `v0`? I'd like to have some version so we don't have something ugly like `/api/` meaning the old API endpoint and `/api/v1/` meaning the new endpoint

courses/management/commands/gen_fake_data.py
```diff
@@ -0,0 +1,41 @@
...
+            dest='programs',
+            default=5,
+            help='Number of programs to generate',
+        )
+
+        parser.add_argument(
+            '--courses',
+            dest='courses',
+            default=3,
+            help='Number of courses to generate',
```
noisecapella: Can you add 'per program' here?

courses/management/commands/gen_fake_data.py
```diff
@@ -0,0 +1,41 @@
...
+        record_count = 0
+        for _ in range(int(options['programs'])):
+            program = ProgramFactory.create()
+            record_count += 1
+            module_range = range(randint(1, int(options['courses'])))
+            for _ in module_range:
+                record_count += 1
+                CourseFactory.create(program=program)
+
+        self.stdout.write("Wrote {} records.".format(record_count))
```
noisecapella: Any reason not to use `print()` here?

justinabrahms: https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/#management-commands-output

noisecapella:
Can you create an issue for Admin interfaces for these models? Currently the only way to set `live` for a program is in the shell

justinabrahms:
Done in #29

micromasters/settings.py
```diff
@@ -149,3 +151,10 @@ def get_var(name, default):
 STATICFILES_DIRS = (
     os.path.join(BASE_DIR, 'static'),
 )
+
+
+REST_FRAMEWORK = {
+    'DEFAULT_PERMISSION_CLASSES': [
+        'rest_framework.permissions.IsAuthenticatedOrReadOnly',
```
noisecapella: Is there a use case for write access to the API?

noisecapella:
Can you change the API views to be read only, at least for this PR?

justinabrahms:
Updated to be read-only. You can't do that as a base level permission, unfortunately.

noisecapella:
:+1: Remember to squash!

PR #37, opened 2016-03-03
Configured sass and css in webpack

### LINKED ISSUE #36: Configure webpack integration for SASS and CSS

noisecapella:
Fixes #36 

#### What's this PR do?
Sets up CSS and Sass integration with webpack, and points the index view to the right place. It also breaks the webpack javascript into three chunks: style, inner, and outer (feel free to suggest better names)

webpack.config.dev.js

Go to the index page, view source. The javascript links should be there and should be valid links. Also, if you add style information to static/scss/layout.scss, it should immediately show up in your browser due to the hot reloading


bdero:
This doesn't seem to be working on the deployed app:

ui/templates/base.html
```diff
@@ -9,6 +10,7 @@
   <body>
     {% block content %}
     {% endblock %}
+  <script type="text/javascript" src="{{ outer_src }}"></script>
```
jamiefolsom: `outer`, as in non-dashboard?

noisecapella: Yes, I'm open to better names if you have one

ui/views.py
```diff
@@ -1,6 +1,31 @@
...
+    """
+    if settings.DEBUG and settings.USE_WEBPACK_DEV_SERVER:
+        host = request.get_host().split(":")[0]
+
+        return "{host_url}/{bundle}".format(
+            host_url=settings.WEBPACK_SERVER_URL.format(host=host),
+            bundle=bundle_name
+        )
+    else:
+        return static("bundles/{bundle}".format(bundle=bundle_name))
```
bdero: Wasn't there a test associated with this code in teachersportal?

npm_install_if_prod.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euf -o pipefail
+if [[ "$NODE_ENV" != "development" ]]
+then
+    node node_modules/webpack/bin/webpack.js --config webpack.config.prod.js
+fi
```
bdero: :+1:

ui/views.py
noisecapella: Yes, I'll copy it over

ui/templates/base.html
jamiefolsom: No that's fine, just want to understand. I have been thinking in terms of subdomains: www and app. Not sure others agree with that.

bdero:
Verified it works on the deploy app now. Firing it up to try it out locally.

bdero:
:+1:, I confirmed this is working locally for me. Although this will need a rebase since https://github.com/mitodl/micromasters/pull/28 created fresh conflicts.

ui/templates/base.html
pdpinch: public? or anonymous_access?

(If I'm understanding the context correctly)

PR #39, opened 2016-03-03
Add pre and post compile hooks (migrations and git hash)

bdero:
#### What are the relevant tickets?
Closes #38 

#### What's this PR do?
This PR adds in the pre and post compile hook straight from TeachersPortal. It allows for running migrations and

#### Where should the reviewer start?
See if the PR app (listed in this PR with a squirrel icon) has successfully deployed (meaning the build passed without error). If it deploys, the migration command returned a success signal.

#### How should this be manually tested?
Check if the git hash is being served from the static files on the review app: https://[pr-app].herokuapp.com/static/hash.txt
If the hash is shown, that means the pre_compile hook successfully ran when Heroku was generating the build artifacts, and that static files are configured properly.

PR #43, opened 2016-03-03
Added models for user profiles

### LINKED ISSUE #18: create model & API for mm user profile
may need to talk to Erdin to understand scope & types of fields

foreign key to the user

for example:

GET /profiles/user_id
POST /profiles/
PATCH /profiles/user_id

noisecapella:
Connected to #18 

#### What's this PR do?
Adds models for the user profile. This PR does not include any REST API work

profiles/models.py

https://i.imgur.com/CaG0KPC.gifv


profiles/api.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+from rest_framework import serializers, viewsets
+from profiles.models import Profile
+
+
+class ProfileSerializer(serializers.ModelSerializer):
+    """Serializer for Program objects"""
+    class Meta:  # pylint: disable=missing-docstring
+        model = Profile
+        fields = ('id',)
```
giocalitri: I think this needs to go in a `serializers.py` file

profiles/api.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class ProfileViewSet(viewsets.ReadOnlyModelViewSet):
+    """API for the Program collection"""
+    queryset = Profile.objects.all()
+    serializer_class = ProfileSerializer
```
giocalitri: is this the REST view? if so it should be in a `views.py` file

I understand what you meant to do, but in the past the `api.py` module contained the api functions used across the app and not the REST APIs, so I would avoid this approach.

Another comment is that we might want to version the rest APIs using directories path like `/api/v0/...`

profiles/models.py
```diff
@@ -0,0 +1,28 @@
...
+from django.contrib.auth.models import User
+from django.db import models
+
+
+class Profile(models.Model):
+    """
+    A user profile
+    """
+    user = models.ForeignKey(User)
+    edx_username = models.TextField()
```
giocalitri: why is this here? can it be even possible to have a different username from edx?

profiles/api.py
noisecapella: Ok, we should also change `courses/api.py` to `courses/serializers.py` and `courses/views.py` to be consistent

bdero:
This should probably be rebased on top of https://github.com/mitodl/micromasters/pull/39 so that the endpoints are verifiable on the PR app.

profiles/models.py
noisecapella: It seems like we can't rely on the two usernames always being identical, but you probably know better than me. When the user logs in via OAuth is the User object created automatically?

justinabrahms: Yes. If that user is already logged in, they get an additional account associated to them.

This username will be queryable via python social auth.

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    PUBLIC_TO_MM = 'public_to_mm'
+    PUBLIC = 'public'
+    ACCOUNT_PRIVACY_CHOICES = (
+        (PRIVATE, 'Private'),
+        (PUBLIC_TO_MM, 'Public to logged in users'),
+        (PUBLIC, 'Public to everyone'),
+    )
+
+    user = models.ForeignKey(User)
+    account_privacy = models.TextField(
```
giocalitri: `TextField` sounds a bit excessive: why not `CharField`?

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+        default=PRIVATE,
+        choices=ACCOUNT_PRIVACY_CHOICES,
+    )
+    email_optin = models.BooleanField(default=False)
+    employer = models.TextField(blank=True, null=True)
+    job_title = models.TextField(blank=True, null=True)
```
giocalitri: also for `employer` and `job_title` maybe `TextField` is too much

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    state_or_territory = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField`

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+    name = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField`

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    bio = models.TextField(blank=True, null=True)
+    country = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField`

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    has_profile_image = models.BooleanField()
+    profile_url_full = models.TextField(blank=True, null=True)
+    profile_url_large = models.TextField(blank=True, null=True)
+    profile_url_medium = models.TextField(blank=True, null=True)
+    profile_url_small = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField` for all urls

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    requires_parental_consent = models.NullBooleanField()
+    year_of_birth = models.IntegerField(blank=True, null=True)
+    level_of_education = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField`

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    goals = models.TextField(blank=True, null=True)
+    language_proficiencies = JSONField(blank=True, null=True)
+    gender = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField`

profiles/models.py
```diff
@@ -0,0 +1,48 @@
...
+    mailing_address = models.TextField(blank=True, null=True)
```
giocalitri: some here about `TextField`

giocalitri:
I think some fields of the model should not be declared as `TextField`, but in general the PR is in a pretty good shape.

profiles/models.py
noisecapella: TextField and CharField are identical under the hood for postgres databases

pwilkins: I didn't read the StackOverflow answer to mean that they are identical in Postgres, just that there was no performance difference.  For Django there are fewer functional differences than if we weren't using an ORM.

profiles/models.py
noisecapella: I'm a bit reluctant to use UrlField or anything else here since these fields are copied from edX (at least initially) so they're not guaranteed to validate

profiles/models.py
giocalitri: they are the same if you do not put a max_length and it does not make much sense to not doing it when we know the max size of the field, like in this case and many of the following.

anyway feel free to leave it in this way if you have a strong opinion about it.

profiles/models.py
giocalitri: I agree about `UrlField`

profiles/models.py
noisecapella: It doesn't look like there's any max length in the edX model where this information comes from, so I'm reluctant to have a max length here

noisecapella:
I changed some of the fields to CharField with max_length=255, and gender and level_of_education have choices and max_length which matches the UserProfile model in edX

giocalitri:
fine with me :+1:

PR #47, opened 2016-03-08
Implemented proof of concept for Python Social Auth backend

### LINKED ISSUE #2: login with edX
As a learner, I need to be able to login and create an account using my edX account. 

This story is just for OAuth login. At a later date we will deal with other user requirements below. 

Learners can only login if:
1. they have an edX account
2. they have enrolled in a micromasters course
3. they have paid for an IDV verified track in the course 

If the learner doesn't meet all this criteria, she should be given a message explaining the situation and a path for correcting it. 

requirements:
- [X] did Justin's edX PRs work?
- [ ] #3 edX enrollment API that can be used to determine if the user has enrolled in one of the necessary courses
- [ ] #3 discovery: determine if edX enrollment API reveal's enrollments IDV state

This story is just for the happy path: user has account, registers and pays for micromasters

giocalitri:
fixes #2 

#### What's this PR do?
This PR is a proof of concept to demonstrate that an user can login using EDX as backend.

`backends/edxorg.py` and `micromasters/settings.py`

1. You need to have a working local instance of EDX. 
2. in EDX create an OAUTH client application for micromasters. NOTE, the Redirect URL does matter: if your micromasters app is at the URL `http://192.168.99.102:8079`, then the Redirect URL MUST BE `http://192.168.99.102:8079/complete/edxorg/`
3. update your `.env` file with the edx instance base url, client id and client secret
4. restart micromaster
5. in micrometers go to the url `/login/edxorg/`
6. verify that a new user is created in the micromasters django admin

In this PR there are no tests given that it is hard to test the OAuth login at this stage unless we can mock a full OAuth provider.


backends/edxorg.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        Loads user data from service.
+
+        This is the function that has to pull the data from edx
+        """
+        return super(EdxOrgOAuth2, self).user_data(access_token, *args, **kwargs)
+
+    def get_user_details(self, response):
+        """
+        Returns user details in a know internal struct.
```
justinabrahms: know/known

backends/edxorg.py
```diff
@@ -0,0 +1,55 @@
...
+    ACCESS_TOKEN_METHOD = 'POST'
+    REDIRECT_STATE = False
+    DEFAULT_SCOPE = ['email']
+
+    def user_data(self, access_token, *args, **kwargs):
+        """
+        Loads user data from service.
+
+        This is the function that has to pull the data from edx
+        """
```
justinabrahms: Can you document return values here and below? I think this returns just an http response?

giocalitri: as far as I understood it should return a dictionary, not an http response

justinabrahms: I was basing this on:

"this has to pull data from edx"
and the function below "given the data from edx, formats it into a dictionary"

Also the function below takes a "response", so I assumed http response.

justinabrahms:
:+1:

PR #56, opened 2016-03-10
homepage first draft

### LINKED ISSUE #11: simple django template for home page

annagav:
Fix #11

### DIFF
--- a/static/scss/layout.scss
+++ b/static/scss/layout.scss
@@ -1,1 +1,80 @@
-// Empty
+
+
+body {
+    background-color: lightgray;
+}
+#container {
+    position: relative;
+    max-width: 900px;
+    border: 1px solid grey;
+    min-height: 900px;
+    margin: 0 auto;
+    background-color: white;
+}
+.purpose-statement-box {
+    width: 600px;
+    border: 2px solid gray;
+    margin: 50px auto;
+    height: 200px;
+    display: table;
+
+}
+.purpose-statement-text {
+    display: table-cell;
+    vertical-align: middle;
+    text-align: center;
+    padding: 0 40px;
+}
+button {
+    border: 1px solid gray;
+    background-color: white;
+    box-shadow: 2px 2px 1px #888888;
+}
+button:hover {
+    background-color: lightgrey;
+}
+.learn-more-btn {
+    margin-top: 20px;
+}
+
+.login-button {
+    width: 150px;
+    height: 60px;
+}
+.login-button-wrapper {
+    position:relative;
+    left:50%;
+    margin-left: -75px;
+}
+.programs-container {
+    margin: 100px 0;
+    position: relative;
+    padding: 0 40px;
+    text-align: center;
+}
+.program-box {
+    display: inline-block;
+    width: 225px;
+    border: 2px solid gray;
+    margin: 10px;
+    overflow: hidden;
+}
+.program-name {
+    vertical-align: middle;
+    display: inline-block;
+}
+.footer-row {
+    font-size: 14px;
+    margin: 0 30px;
+    border-top: 3px solid grey;
+    position: absolute;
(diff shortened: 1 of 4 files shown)


static/css/layout.css
```diff
@@ -0,0 +1,80 @@
+
```
noisecapella: Can you copy this file's contents into `static/scss/layout.scss`? This wasn't working when you started with this functionality but it should work now

ui/templates/base.html
```diff
@@ -3,6 +3,8 @@
   <head>
     <meta charset="UTF-8">
     <script type="text/javascript" src="{{ style_src }}"></script>
+    {% load staticfiles %}
+    <link href="{% static 'css/layout.css'%}" rel="Stylesheet"/>
```
noisecapella: After the contents of `layout.css` are copied into `layout.scss` you should remove this line. The contents of `layout.scss` are included in the above line with `style_src` instead.

ui/templates/index.html
```diff
@@ -4,6 +4,38 @@
...
+            <div class="program_box">
+                <span style="line-height: 115px"></span>
+                <div class="program_name">{{ program }}</div>
+            </div>
+        {% endfor %}
+    </div>
+
+    <div class="footer_row">
+        <div class="footer_links">
+            <a href="#">MIT Giving</a>
```
noisecapella: Can you add URLs for these three links?

ui/urls.py
```diff
@@ -7,4 +7,5 @@
 
 urlpatterns = [
     url(r'^$', index, name='ui-index'),
+
```
noisecapella: Remove this line please, since otherwise the file is unchanged

ui/templates/index.html
annagav: MIT Giving : http://giving.mit.edu/ ?

static/scss/layout.scss
```diff
@@ -1 +1,80 @@
-// Empty
+
```
noisecapella: Can you change the underscores to dashes? It seems to be more popular to do it that way so we might as well follow along

static/scss/layout.scss
```diff
@@ -1 +1,80 @@
...
+
+body{
```
noisecapella: Can you add a space before the `{` for each item in this file?

ui/views.py
```diff
@@ -4,6 +4,7 @@
 from django.conf import settings
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.shortcuts import render
+from courses import models
```
noisecapella: Can you change this to `from courses.models import Program`? It's nitpicky but it's a more specific import which is good for code cleanliness

noisecapella:
It looks like you will need to edit the test to test for the existence of the program text instead of "Hi"

noisecapella:
Also, feel free to provide a reason not to change something if you don't agree with a comment, instead of changing it, for any of our code reviews :)

ui/views_test.py
```diff
@@ -17,10 +18,11 @@ def setUp(self):
 
     def test_index_view(self):
```
noisecapella: Just two more things:
 - Can you rename this test to `test_program_liveness`
 - In this test create another Program where live=False, and assert that the program title is not in the response

ui/views_test.py
```diff
@@ -17,10 +18,11 @@ def setUp(self):
...
         """Verify the index view is as expected"""
```
noisecapella: Also can you update this description?

noisecapella:
Looks great :+1: 

One more thing: Can you rename both the commit message and the PR title to something like "Implemented home page"? It should be in the past tense because our release script will use these titles as bullet points in our release notes

PR #60, opened 2016-03-10
Fetch information for user profile from edX

### LINKED ISSUE #50: get actual user profile data while logging in with edx
PR #47 was a proof of concept with hardcoded data returned.

This PR should implement a way to get the actual data.

During the "OAUTH dance" MM should first query edx at the URL `/api/mobile/v0.5/my_user_info` to get the user's username and then get the proper user profile using the url `/api/user/v1/accounts/{username}/`

This is a continuation of #2

justinabrahms:
Fixes #50 

#### What's this PR do?
Fetches information from edx for user's profile
edxorg.py
Same as https://github.com/mitodl/micromasters/pull/47 , but the user should be created with the correct username. If you're already logged in (as superuser), that account will be associated to you.

Status codes are [handled](https://github.com/omab/python-social-auth/blob/master/social/backends/base.py#L228) by python-social-auth.

admin-fail

### DIFF
--- a/backends/edxorg.py
+++ b/backends/edxorg.py
@@ -35,7 +35,12 @@ class EdxOrgOAuth2(BaseOAuth2):
             dict: a dictionary containing user information
                 coming from the remote service.
         """
-        return super(EdxOrgOAuth2, self).user_data(access_token, *args, **kwargs)
+        return self.get_json(
+            urljoin(EDXORG_BASE_URL, "/api/mobile/v0.5/my_user_info"),
+            headers={
+                "Authorization": "Bearer {}".format(access_token),
+            }
+        )
 
     def get_user_details(self, response):
         """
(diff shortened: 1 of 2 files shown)


backends/edxorg_test.py
```diff
@@ -0,0 +1,51 @@
...
+        })
+
+        assert {
+            'edx_id': 5,
+            'username': 'staff',
+            'fullname': 'staff',
+            'email': 'staff@example.com',
+            'first_name': 'staff',
+            'last_name': ''
+        } == result
```
giocalitri: What about a name like `Giovanni Di Milia`? :-)

giocalitri:
Code looks good: maybe a test missing (but I do not have strong opinions about it)

backends/edxorg_test.py
justinabrahms: https://github.com/omab/python-social-auth/blob/master/social/backends/base.py#L167

Handled. You think I would forget about you?

backends/edxorg.py
```diff
@@ -54,14 +59,17 @@ def get_user_details(self, response):
...
-            'edx_id': 'unique_edx_id_1',
-            'username': 'my_username',
-            'email': 'firstname_lastname@example.com',
-            'fullname': 'firstname lastname',
-            'first_name': 'firstname',
-            'last_name': 'lastname'
+        full, first, last = self.get_user_names(response['name'])
+
+        retval = {
+            'edx_id': response['id'],
```
giocalitri: do we want to use this as remote unique identifier? Isn't it better to use the email that will be unique even if there is a change in the DB?

justinabrahms: I don't care one way or another. Will defer to your preference.

giocalitri: let's use the username

justinabrahms:
Done.

giocalitri:
Perfect :+1: as soon as tests pass

giocalitri:
actually, for clarity, can you please update the tests?

giocalitri:
perfect, thanks

PR #85, opened 2016-03-16
Added REST API for profiles

### LINKED ISSUE #18: create model & API for mm user profile

noisecapella:
Fixes #18

#### What's this PR do?
 - Adds a REST API for viewing and editing user profiles: `/api/v0/profiles/<username>/`
 - Updates ProfileFactory to have randomized values for all fields

profiles/serializers.py

- In the Django shell, run

        from profiles.factories import ProfileFactory
        user = User.objects.create_user(username="user", password="pass")
        profile = ProfileFactory.create(user=user, account_privacy="public")

This will create a user named `user` with `account_privacy` set to `public`. Keep the shell open so we can edit profile later.

- Create a superuser and login as that superuser via Django Admin.
- View `/api/v0/profiles/user/`. You should see 12 key value pairs, including `bio` and `country` for example, but not `mailing_address`.
- In the Django shell run

        profile.account_privacy = "public_to_mm"
        profile.save()

- View `/api/v0/profiles/user/` again. You should see exactly the same data as before
- In the Django shell run

        profile.account_privacy = "private"
        profile.save()

- View `/api/v0/profiles/user/` again. You should see only fields related to profile image, and `account_privacy`.

This doesn't add any additional roles or permissions. See #89

### DIFF
--- a/profiles/factories.py
+++ b/profiles/factories.py
@@ -1,8 +1,22 @@
 """Factories for making test data"""
+from datetime import datetime, timezone
+
 from django.contrib.auth.models import User
-import factory
+from factory import (
+    Sequence,
+    SubFactory,
+)
 from factory.django import DjangoModelFactory
+from factory.fuzzy import (
+    FuzzyAttribute,
+    FuzzyChoice,
+    FuzzyDateTime,
+    FuzzyInteger,
+    FuzzyText,
+)
 import faker
+
+from profiles.models import Profile
 
 
 FAKE = faker.Factory.create()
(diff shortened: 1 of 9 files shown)


profiles/factories.py
```diff
@@ -20,9 +32,35 @@ class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,
...
+    year_of_birth = FuzzyInteger(1850, 2015)
+    level_of_education = FuzzyChoice(
+        [choice[0] for choice in Profile.LEVEL_OF_EDUCATION_CHOICES]
+    )
+    goals = FuzzyText()
+    gender = FuzzyChoice(
+        [choice[0] for choice in Profile.GENDER_CHOICES]
+    )
+    mailing_address = FuzzyText()
+    # 10000 days is 27 years
```
justinabrahms: This doesn't make sense to me in context.

profiles/permissions.py
```diff
@@ -0,0 +1,22 @@
+"""
+Permission classes for profiles
+"""
+from rest_framework.permissions import (
+    BasePermission,
+    SAFE_METHODS,
+)
+
+
+class CanEditIfOwner(BasePermission):
```
justinabrahms: Tests?

profiles/views_test.py
```diff
@@ -0,0 +1,136 @@
...
+        self.url1 = reverse('profile-detail', kwargs={'user': self.user1.username})
+
+        self.user2_credentials = {
+            "username": "user2",
+            "password": "pass"
+        }
+        self.user2 = User.objects.create_user(**self.user2_credentials)
+        self.url2 = reverse('profile-detail', kwargs={'user': self.user2.username})
+
+    def test_no_anonymous_access(self):
```
justinabrahms: Ideally, this test would be a simple assertion that the permission classes are set correctly (is authenticated and the edit one you made above). If that's true and they're both tested to work.. then DRF will handle this for us and we don't have to make (slow) integration tests for it. DRF already tests that functionality for us.

profiles/views_test.py
```diff
@@ -0,0 +1,136 @@
...
+    Tests for GET on profile view
+    """
+
+    def setUp(self):
+        """
+        Create a user and profile
+        """
+        super(ProfileTests, self).setUp()
+
+        self.user1_credentials = {
```
justinabrahms: We don't actually need to store these anymore. Users can be forcibly logged in with Django 1.9's `force_login` function. Additionally, `force_login` skips some of the authorization checks around login, which makes it faster. https://docs.djangoproject.com/en/1.9/topics/testing/tools/#django.test.Client.force_login

profiles/views_test.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Get another user's private profile
+        """
+        profile = ProfileFactory.create(user=self.user2, account_privacy=Profile.PRIVATE)
+        self.client.login(**self.user1_credentials)
+
+        resp = self.client.get(self.url2)
+        assert resp.json() == ProfilePrivateSerializer().to_representation(profile)
+
+    def test_patch_other_profile(self):
```
justinabrahms: This should be handled in the `CanEditIfOwner` tests, I believe.

profiles/views_test.py
noisecapella: We also need to test that we called `self.check_object_permissions` somewhere, otherwise the `has_object_permission` function won't be called. One reason I like integration tests is that they detect these edge cases automatically. But I think in this case we should be ok to break it up and assert that `has_object_permission` was called

profiles/views.py
```diff
@@ -0,0 +1,45 @@
...
+            return self.serializer_class_private
+        else:
+            return self.serializer_class_limited
+
+    def get_object(self):
+        """
+        Override get_object to use username
+        """
+        obj = get_object_or_404(Profile, user__username=self.kwargs['user'])
+        self.check_object_permissions(self.request, obj)
```
justinabrahms: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L76

This is the default implementation of get_object. It looks like if we set `lookup_field` to be `user__username` and `lookup_url_kwarg` to `user`, we don't have to write this method at all (and can skip the test mentioned below).

profiles/signals.py
```diff
@@ -12,9 +12,10 @@
 
 
 log = logging.getLogger(__name__)
+CREATE_PROFILE = "create_profile"
```
justinabrahms: Why the constant?

noisecapella: Leftover from a previous revision

justinabrahms:
:+1:

PR #86, opened 2016-03-18
Created Dashboard Container 

### LINKED ISSUE #75: Student Dashboard: Overall Page Template
As a student I want a page on which to see my dashboard information, which will contain other data, including my program and course listing, avatar etc.

jamiefolsom:

Fixes #75 

#### What's this PR do?

 - Adds a page template for the dashboard.
 - Removes checkbox redux stuff, which is only there as an example
 - General CSS/JS cleanup for consistency
 - Add react-bootstrap

static/js/containers/DashboardPage.js

Visit  `/dashboard` after logging in

screenshot from 2016-03-24 13-41-49

#### Background context

This isn't very responsive design, but it uses bourbon neat so we shouldn't be blocking ourselves from doing this work later.


ui/views.py
```diff
@@ -48,9 +57,23 @@ def dashboard(request):
     """
     The app dashboard view
     """
+    host = request.get_host().split(":")[0]
+
+    name = ""
+    if not request.user.is_anonymous():
+        name = request.user.username
```
justinabrahms: Should there ever be a case where a user has a username, but not a profile?

justinabrahms:
Minor question, but this looks good to me. :+1:

ui/views.py
noisecapella: Not since @giocalitri merged his PR. I'll remove the check below

PR #91, opened 2016-03-21
Profiles creation

### LINKED ISSUE #73: On first (OAuth) login, create a user profile
As a user, when I login to micromasters for the first time I want my profile to be created.

giocalitri:
fixes #73 

#### What's this PR do?
Creates a Profile for the user and populates it with the data existing in the connected EDX instance

`pipeline_api.py`

Add some personal info for your user on edx and then login again on micromasters.
The user should be new on micromasters otherwise the profile will not be updated with the info from edx

you need a working micromasters properly configured with a local instance of EDX


backends/pipeline_api.py
```diff
@@ -0,0 +1,79 @@
...
+from urllib.parse import urljoin
+
+from profiles.models import Profile
+
+log = logging.getLogger(__name__)
+
+
+def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):   # pylint: disable=unused-argument
+    """
+    Gets profile informations from EDX and saves them in the user profile
```
justinabrahms: information not informations.

backends/edxorg.py
```diff
@@ -16,6 +14,7 @@ class EdxOrgOAuth2(BaseOAuth2):
     name = 'edxorg'
     ID_KEY = 'edx_id'
     REQUEST_TOKEN_URL = None
+    EDXORG_BASE_URL = getattr(settings, 'EDXORG_BASE_URL', 'https://courses.edx.org/')
```
noisecapella: It seems like `settings.EDXORG_BASE_URL` should always exist here. Therefore the third argument here would never be used. Also, it seems like if `https://courses.edx.org/` was meant to be a default value, this should go in `settings.py` as `get_var('EDXORG_BASE_URL', 'https://courses.edx.org/')`

backends/pipeline_api.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    # this function is completely skipped if the backend is not edx or
+    # the user has not created now
+    if backend.name != 'edxorg' or not is_new:
+        return
+
+    access_token = response.get('access_token')
+    if not access_token:
+        # this should never happen for the edx oauth provider, but just in case...
```
justinabrahms: Can you add a logging statement here?

justinabrahms:
:+1: from me after you address my comments. George is also reviewing too, it seems.

backends/pipeline_api.py
```diff
@@ -0,0 +1,79 @@
...
+        return
```
noisecapella: Can we log this case?

backends/pipeline_api.py
```diff
@@ -0,0 +1,79 @@
...
+
+    try:
+        user_profile = Profile.objects.get(user=user)
+    except Profile.DoesNotExist:
+        # this should never happen, since the profile is created with a signal
+        # right after the user is created
+        return
```
noisecapella: Can we log this case?

backends/pipeline_api_test.py
```diff
@@ -0,0 +1,137 @@
...
+        }
+        pipeline_api.update_profile_from_edx(
+            edxorg.EdxOrgOAuth2, self.user, {'access_token': 'foo_token'}, True)
+        mocked_get_json.assert_called_once_with(
+            urljoin(
+                edxorg.EdxOrgOAuth2.EDXORG_BASE_URL,
+                '/api/user/v1/accounts/{0}'.format(self.user.username)
+            ),
+            headers={'Authorization': 'Bearer foo_token'}
+        )
```
noisecapella: Can you assert the updated values in `self.user_profile`?

backends/pipeline_api_test.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_update_profile_no_existing_profile(self):
+        """
+        The profile did not exist for the user
+        """
+        self.user_profile.delete()
+
+        # just checking that nothing raises an exception
+        pipeline_api.update_profile_from_edx(
+            edxorg.EdxOrgOAuth2, self.user, {'access_token': 'foo_token'}, True)
```
noisecapella: Can you add a check that no profiles have been created?

profiles/signals.py
```diff
@@ -0,0 +1,24 @@
...
+log = logging.getLogger(__name__)
+
+
+@receiver(post_save, sender=User, dispatch_uid="create_profile")
+def create_user_profile(sender, instance, created, **kwargs):  # pylint: disable=unused-argument
+    """
+    Signal handler to create a user profile every time a new user is created
+    """
+    if created:
+        log.debug('creting profile for user %s', instance.username)
```
noisecapella: creating

profiles/models.py
```diff
@@ -98,4 +99,7 @@ class Profile(models.Model):
         null=True,
     )
     mailing_address = models.TextField(blank=True, null=True)
-    date_joined = models.DateTimeField(blank=True, null=True)
+    date_joined_micromasters = models.DateTimeField(blank=True, null=True, default=datetime.now)
```
noisecapella: Can you change this to use `auto_now_add=True` instead?

noisecapella: I think you can get rid of `null=True` also

noisecapella:
Functionality works great

profiles/models.py
giocalitri: I chose `default=datetime.now` instead of `auto_now_add=True` because the latter makes the field read only. Not sure it matters...

giocalitri:

backends/edxorg.py
```diff
@@ -14,7 +14,7 @@ class EdxOrgOAuth2(BaseOAuth2):
     name = 'edxorg'
     ID_KEY = 'edx_id'
     REQUEST_TOKEN_URL = None
-    EDXORG_BASE_URL = getattr(settings, 'EDXORG_BASE_URL', 'https://courses.edx.org/')
+    EDXORG_BASE_URL = getattr(settings, 'EDXORG_BASE_URL')
```
noisecapella: Why not `settings.EDXORG_BASE_URL` here? It will always exist

noisecapella:
Just one nit, feel free to squash and merge :+1:

backends/edxorg.py
giocalitri: you are right: just missed

PR #102, opened 2016-03-24
Added dockerignore

### LINKED ISSUE #101: Add .dockerignore
This will make the COPY step faster for rebuilds, and reduce the chance of running out of disk space on a docker-machine VM

noisecapella:
Fixes #101 

#### What's this PR do?
Adds a .dockerignore file which lets us skip copying certain directories and files from the host to the docker container on build. The main offender here is `node_modules` which should not be dealt with at all during `docker-compose build`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.tox
+htmlcov
+.coverage
+node_modules
+static/bundles
+staticfiles/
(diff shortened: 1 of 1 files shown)


justinabrahms:
:+1: if PR passes.

PR #107, opened 2016-03-25
Use sitepackages=True so tox reuses the same packages in the docker environment

### LINKED ISSUE #108: When running tests on Travis use tox --sitepackages to avoid installing dependencies twice
Since the docker container and tox both use exact same dependencies for micromasters, we should be able to speed up tests a bit by using the --sitepackages option

noisecapella:
Fixes #108 

#### What's this PR do?
Updates tox.ini to use site packages. Since we only have one environment which is exactly the same as the docker container, this should not cause any problems, and it speeds up builds a little bit.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,7 @@ skip_missing_interpreters = True
 skipsdist = True
 
 [testenv]
+sitepackages = True
 deps =
     -r{toxinidir}/requirements.txt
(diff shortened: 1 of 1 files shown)


justinabrahms:
:+1: Awesome. That looks like a 1+ minute speedup!

PR #120, opened 2016-03-28
Fixed LICENSE organization

noisecapella:
From the 3 clause BSD license template [here](https://en.wikipedia.org/wiki/BSD_licenses) it seems like this should be our organization, not the Micromasters name specifically.


pdpinch:
:+1:

PR #127, opened 2016-03-30
Added sign-in-with LinkedIn backend support.

### LINKED ISSUE #126: add LinkedIn social auth plugin
Create an endpoint that prompts the user to login to LInkedIn via OAuth and returns all the json available. 

We will use this later once we've established the mapping of LinkedIn fields to user profile fields in micromasters.

justinabrahms:
Fixes #126 

#### What's this PR do?
Adds support for getting LinkedIn basic profile information and loads it into the db.

pipeline_api.py

Get some linkedin api credentials from https://www.linkedin.com/developer/apps (you'll need a valid linkedin account) and put them in .env

Then go to localhost:8079/login/linkedin-oauth2/
We can't get the full profile w/ past jobs until we make it into the linkedin partner program. Ferdi applied for it.
No, but look at the fixture I checked in with my own data.
http://s.mlkshk-cdn.com/r/181FE.gif


.env.example
```diff
@@ -2,3 +2,5 @@ STATUS_TOKEN=
 EDXORG_BASE_URL=
 EDXORG_CLIENT_ID=
 EDXORG_CLIENT_SECRET=
+LINKEDIN_CLIENT_ID=
+LINKEDIN_CLIENT_SECRET=
```
giocalitri: nit: newline at the end of the file

backends/pipeline_api.py
```diff
@@ -79,3 +79,35 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
...
+        user (User): user object
+        response (dict): dictionary of the user information coming
+            from previous functions in the pipeline
+        is_new (bool): whether the authenticated user created a new local instance
+
+    Returns:
+        None
+    """
+
+    # this function is completely skipped if the backend is not edx or
```
giocalitri: you meant `if the backend is not linkedin`

backends/pipeline_api_test.py
```diff
@@ -166,3 +167,33 @@ def test_update_profile(self, mocked_get_json):
...
+        backend.name = 'linkedin-oauth2'
+        user = UserFactory.create()
+        response = {'test': 'works'}
+        is_new = True
+        update_from_linkedin(backend, user, response, is_new)
+        is_new = False
+        update_from_linkedin(backend, user, {'doesnt': 'work'}, is_new)
+
+        profile = Profile.objects.get(user=user)
+        assert 'doesnt' not in profile.linkedin
```
giocalitri: Maybe it is too much, but what if the profile does not exist (after all there is a check in the code)

docker-compose.yml
```diff
@@ -13,6 +13,7 @@ web:
   volumes:
     - .:/src
   environment:
+    DJANGO_LOG_LEVEL: 'INFO'
```
giocalitri: I would put this in the .env file,  already set to `'INFO'`. If we put it there, then we cannot override from the .env file

giocalitri:
Just few minor comments

noisecapella:
Outside the scope of this PR but is the intent that this will only get triggered when the user clicks some 'Import from LinkedIn' button? In what cases do we want to have that overwrite the user's information that they may have previously updated themselves?

backends/pipeline_api_test.py
justinabrahms: We have a test that this shouldn't happen.

https://github.com/mitodl/micromasters/blob/master/profiles/signals_test.py#L17

docker-compose.yml
justinabrahms: Removed. I'll include it in my own personal `.env`

backends/pipeline_api.py
justinabrahms: Done.

.env.example
justinabrahms: Done.

justinabrahms:
Updated.

This will happen when a user hits that button, yes.

This will not update user's profile information. It only stores the response from linkedin in a specific json field. If we want to prepopulate their real profile information with information from linked in, we can pull it from that json blob.

giocalitri:
Actually @noisecapella is right: I think this should not be part of the python auth pipeline (at least this is what we said yesterday)

giocalitri:
@justinabrahms `is_new` is `True` for a new local django user or a new local social_auth user?

if it is the first, then the code does not work

giocalitri:
@justinabrahms where is the logging level? Did you remove it completely?

justinabrahms:
I do not suspect this button will exist in the anonymous areas of the website. It should only be in the profile editing section, as we didn't have "login with linkedin" as a baseline auth option, only for the "help users filling out their profile" option. I think the user should exist by this point, but perhaps @Ferdi or @pdpinch will have an answer.

justinabrahms:
I removed it completely. I have it in my own .env file now, which isn't checked into version control.

giocalitri:
actually we could put it back in the .env.example, but that's not important

giocalitri:
@justinabrahms sorry, there has been a misunderstanding from me
I was just talking about these lines:
https://github.com/mitodl/micromasters/blob/linkedin/backends/pipeline_api.py#L101-L102

does `is_new` means that the user logged in for the first time with linkedin or that she logged in for the very first time in general? I guess the first case, but I want to be sure.

justinabrahms:
Good catch. Removed that case entirely after a conversation w/ Peter. Clicking "get info from linkedin" a second time will update the linkedin information locally.

giocalitri:
great: feel free to merge :+1:

PR #129, opened 2016-03-31
Course runs

### LINKED ISSUE #124: Add a model to abstract courses from course runs
Because edX will offer multiple runs of the same course, but users need only take one of the runs to meet the micromasters requirement. 

The courses REST API shouldn't need any major changes, but the enrollment API will.  

See picture of new structure:

img_4231

justinabrahms:
Fixes #124 

#### What's this PR do?
Moves the old `Course` model to `CourseRun` and introduces a logical parent for it, called `Course`. `Course` denotes a course which isn't bound by a specific start/end date.

models.py

Should be functionally identical, save a new admin.

We need a way to track multiple courses runs. These will be manually entered, because edx course ids aren't deterministic.

course-admin


justinabrahms:
cc @giocalitri to make sure this is what you were thinking w/ your diagram above.

requirements.txt
```diff
@@ -14,3 +14,4 @@ requests==2.9.1
 static3==0.5.1
 tox==1.9.2
 uwsgi==2.0.12
+Pillow==3.1.1
```
giocalitri: what's this for?

giocalitri:
code looks good: only a minor comment :+1:

apt.txt
```diff
@@ -3,6 +3,7 @@ git
 curl
 libpq-dev
 python3-dev
+libjpeg-dev
```
noisecapella: Are we doing any image processing in this PR?

justinabrahms: We have an ImageField. This requires pillow which requires libjpeg.

giocalitri:
Can you squash?

courses/models.py
```diff
@@ -17,15 +17,40 @@ def __str__(self):
...
+
+
+class CourseRun(models.Model):
+    """
+    An individual run of a course within a Program, e.g. "Supply Chain 101
+    - Summer 2017". This is different than the logical notion of a course, but
+      rather a specific instance of that course being taught.
     """
     title = models.CharField(max_length=255)
     edx_course_key = models.CharField(max_length=255, blank=True, null=True)
```
pdpinch: I'm debating whether some help text may be useful here. 

`Can be null, a deprecated course key or a split mongo course key. Intended for course keys on edx.org`

PR #131, opened 2016-04-04
Removed redux-actions

### LINKED ISSUE #128: Remove redux-actions
I'd like to remove redux-actions, which is a helper library for simplifying how we define our reducers. The project is accumulating a backlog of issues and pull requests, and it doesn't really gain us much. It's also an extra library that people need to become familiar with when learning redux. Thoughts @justinabrahms?

noisecapella:
Fixes #128
#### What's this PR do?
Removes redux-actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "react-ga": "1.2.0",
     "react-redux": "4.4.0",
     "react-router": "2.0.0",
-    "redux-actions": "0.9.1",
     "redux-devtools": "2.1.5",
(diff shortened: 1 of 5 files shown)


justinabrahms:
I think this code less clear and is more complex than the previous code and "people will need to learn a library" is an argument to never use any library. I think introducing this change into Teacher's Portal will more clearly show the above, wherein we have more actions and the benefits of the library are clearer.

That said, I don't feel a strong sense of ownership on this code base, so :+1: if you want to merge it.

PR #133, opened 2016-04-05
Display student avatar and name

### LINKED ISSUE #67: Student dashboard: Display student avatar and student name
As a student, I want my avatar to appear on my dashboard page, and throughout the private (and in community areas, as appropriate based on my privacy preferences).

This requires: offering a way for students to authenticate with another service and select a hosted avatar (eg gravatar, twitter, facebook), or to upload one, size, center and crop.

MVP: display edX avatar

annagav:
Fix #67 

Go to student dashboard.

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -5,8 +5,14 @@ export const REQUEST_COURSE_LIST = 'REQUEST_COURSE_LIST';
 export const RECEIVE_COURSE_LIST_SUCCESS = 'RECEIVE_COURSE_LIST_SUCCESS';
 export const RECEIVE_COURSE_LIST_FAILURE = 'RECEIVE_COURSE_LIST_FAILURE';
 
-export const CLEAR_COURSE_LIST = 'CLEAR_COURSE_LIST';
+export const REQUEST_USER_PROFILE = 'REQUEST_USER_PROFILE';
+export const RECEIVE_USER_PROFILE_SUCCESS = 'RECEIVE_USER_PROFILE_SUCCESS';
+export const RECEIVE_USER_PROFILE_FAILURE = 'RECEIVE_USER_PROFILE_FAILURE';
 
+export const CLEAR_COURSE_LIST = 'CLEAR_COURSE_LIST';
+export const CLEAR_PROFILE = 'CLEAR_PROFILE';
+
+// constants for fetch status (these are not action types)
 export const FETCH_FAILURE = 'FETCH_FAILURE';
 export const FETCH_SUCCESS = 'FETCH_SUCCESS';
 export const FETCH_PROCESSING = 'FETCH_PROCESSING';
@@ -31,4 +37,27 @@ export function fetchCourseList() {
       then(json => dispatch(receiveCourseListSuccess(json))).
       catch(() => dispatch(receiveCourseListFailure()));
   };
+
+
 }
+
+const requestUserProfile = () => ({ type: REQUEST_USER_PROFILE });
+
+export const receiveUserProfileSuccess = profile =>({
+  type: RECEIVE_USER_PROFILE_SUCCESS,
+  payload: { profile }
+});
+
+const receiveUserProfileFailure = () => ({ type: RECEIVE_USER_PROFILE_FAILURE });
+
+export const clearProfile = () => ({ type: CLEAR_PROFILE });
+
+export function fetchUserProfile(username) {
+  return dispatch => {
+    dispatch(requestUserProfile());
+    return api.getUserProfile(username).
+      then(json => dispatch(receiveUserProfileSuccess(json))).
+      catch(()=> dispatch(receiveUserProfileFailure()));
+  };
+}
+
--- a/static/js/components/UserImage.js
+++ b/static/js/components/UserImage.js
@@ -1,0 +1,17 @@
+/* global SETTINGS: false */
+import React from 'react';
+
+class UserImage extends React.Component {
+  render() {
+    const { imageUrl } = this.props;
+
+    return <img
+      className="card-image" src={ imageUrl }
+      alt={"Profile image for " + SETTINGS.name}/>;
+  }
+}
+
+UserImage.propTypes = {
+  imageUrl: React.PropTypes.string.isRequired
+};
+export default UserImage;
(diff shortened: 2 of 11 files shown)


noisecapella:
Can you change this PR to store the entire profile we receive from the API rather than just `userImage`? That way we won't have to change anything when we need to retrieve more information from the profile

static/js/containers/DashboardPage.js
```diff
@@ -4,16 +4,19 @@ import Header from '../components/Header';
...
 
 class DashboardPage extends React.Component {
   componentDidMount() {
     this.fetchCourseList();
+    this.fetchUserImage();
   }
 
   componentDidUpdate() {
     this.fetchCourseList();
+
```
noisecapella: Can you add `fetchUserImage()` here too?

static/js/reducers/index.js
```diff
@@ -44,6 +47,34 @@ export const courseList = (state = INITIAL_COURSE_LIST_STATE, action) => {
   }
 };
 
+const INITIAl_USER_IMAGE_STATE = {
```
noisecapella: Can you change this to use all capital letters?

static/js/reducers/index.js
```diff
@@ -44,6 +47,34 @@ export const courseList = (state = INITIAL_COURSE_LIST_STATE, action) => {
...
+  imageUrl: SETTINGS.edx_base_url + 'static/images/profiles/default_120.png'
```
noisecapella: Can you investigate if there's a way to join these two pieces so that it doesn't matter if `SETTINGS.edx_base_url` has the trailing slash or not? You may want to start here: http://stackoverflow.com/questions/29855098/is-there-a-built-in-javascript-function-similar-to-os-path-join

static/js/components/UserImage.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Created by anna on 4/1/16.
```
noisecapella: Can you remove this and the other comments that PyCharm adds by default?

static/js/util/api.js
```diff
@@ -79,3 +79,7 @@ function fetchJSONWithCSRF(input, init) {
 export function getCourseList() {
   return fetchJSONWithCSRF('/api/v0/courses/');
 }
+
+export function getUserImage() {
+  return fetchJSONWithCSRF('/api/v0/profile/'+SETTINGS.username);
```
noisecapella: It would be cleaner to have `getUserImage` take an argument `username` and use that instead of `SETTINGS.username`, in case we wanted to use the function for another user in the future. Instead, have the code which dispatches the `fetchUserImage` action take `SETTINGS.username` as an argument

static/js/components/UserImage.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+import React from 'react';
+
+class UserImage extends React.Component{
+  render() {
+    const { imageUrl } = this.props;
+
+    return <img className="card-image" src={imageUrl}/>;
```
pdpinch: we should always have an `alt` attribute on img, even if it's empty. 

in this case can we use "profile image for <username>" (similar to edX)

annagav:
Addressed comments

static/js/actions/index.js
```diff
@@ -5,8 +5,13 @@ export const REQUEST_COURSE_LIST = 'REQUEST_COURSE_LIST';
 export const RECEIVE_COURSE_LIST_SUCCESS = 'RECEIVE_COURSE_LIST_SUCCESS';
 export const RECEIVE_COURSE_LIST_FAILURE = 'RECEIVE_COURSE_LIST_FAILURE';
 
+export const REQUEST_USER_PROFILE = 'REQUEST_USER_PROFILE';
+export const RECEIVE_USER_PROFILE_SUCCESS = 'RECEIVE_USER_PROFILE_SUCCESS';
+export const RECEIVE_USER_PROFILE_FAILURE = 'RECEIVE_USER_PROFILE_FAILURE';
```
noisecapella: Can you add a CLEAR_PROFILE and related actions and reducers and the dispatch in the `componentWillUnmount`, similar to CLEAR_COURSE_LIST below?

static/js/reducers/index.js
```diff
@@ -44,6 +47,37 @@ export const courseList = (state = INITIAL_COURSE_LIST_STATE, action) => {
   }
 };
 
+const INITIAL_USER_PROFILE_STATE = {
+  profile: {
+    profile_url_large: // eslint-disable-line camelcase
+      (SETTINGS.edx_base_url + '/static/images/profiles/default_120.png').
+        replace(/([^:]\/)\/+/g, "$1")
```
noisecapella: Can you add a comment describing what the regex is doing?

noisecapella:
Can you add tests:

 - in `api_test.js` for the new `getUserProfile` function
 - in `index_test.js` for the new user profile reducer function

They should be pretty similar to the existing course tests

static/js/components/UserImage.js
```diff
@@ -0,0 +1,17 @@
...
+    const { imageUrl } = this.props;
+
+    return <img
+      className="card-image" src={ imageUrl }
+      alt={"Profile image for " + SETTINGS.name}/>;
+  }
+}
+
+UserImage.propTypes = {
+  imageUrl: React.PropTypes.object.isRequired
```
noisecapella: This should be a string

static/scss/layout.scss
```diff
@@ -165,19 +165,39 @@ body {
...
+        .card-image-box {
+          @include span-columns(2);
+          min-width: 150px;
+          height: 220px;
+          padding-top: 40px;
+          .card-image{
+            border-radius: 50%;
+            height: 140px;
+            width: 140px;
+            float: right;
```
noisecapella: Can you use `margin-left` here instead to position the image? I'd like to avoid `float` as much as possible

static/scss/layout.scss
```diff
@@ -165,19 +165,40 @@ body {
...
+      .card-user {
+        background-color: darkslategrey;
         @include span-columns(12);
-        padding: 25px 0px;
+        .card-image-box {
+          @include span-columns(2);
+          min-width: 190px;
+          height: 220px;
+          padding-top: 40px;
+          border:1px solid red;
```
noisecapella: Can you remove this?

noisecapella:
Functionality looks good! :+1:

PR #141, opened 2016-04-06
Added redirect, "/dashboard" -> "/" if user not logged in

### LINKED ISSUE #98: Enforce login required for dashboard
Figure out what happens when a user is not logged in but they visit the dashboard

aliceriot:
this closes #98

#### What's this PR do?

this adds a simple redirect check when hitting `/dashboard`. if
`request.user` is not authenticated, we redirect to `/`.

in `/ui/views.py`

1. log in to micromasters
2. check that you can visit `/dashboard`
3. click the (brand new!) logout button
4. visit `/dashboard`
5. confirm

first python commit!

abe-simpson-gif

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -7,6 +7,7 @@ import logging
 from django.conf import settings
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.shortcuts import render
+from django.contrib.auth.decorators import login_required
 
 from courses.models import Program
 
@@ -52,6 +53,7 @@ def index(request):
     })
 
 
+@login_required()
 def dashboard(request):
     """
     The app dashboard view
--- a/ui/views_test.py
+++ b/ui/views_test.py
@@ -4,7 +4,10 @@ Test end to end django views.
 from django.test import TestCase
 from django.test.client import Client
 from django.core.urlresolvers import reverse
+from django.db.models.signals import post_save
+from factory.django import mute_signals
 from courses.factories import ProgramFactory
+from profiles.factories import ProfileFactory, UserFactory
 
 
 class TestViews(TestCase):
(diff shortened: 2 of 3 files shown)


noisecapella:
Can you use the [@login_required](https://docs.djangoproject.com/en/1.9/topics/auth/default/#django.contrib.auth.decorators.login_required) decorator instead?

noisecapella:
You will also need to write a test to verify this. There are a few UI tests currently in `views_test.py` in the same directory. You should be able to just get the page with `self.client` similar to the other tests and then assert whatever non-200 status code you get

ui/views_test.py
```diff
@@ -32,6 +32,14 @@ def test_program_liveness(self):
             status_code=200
         )
 
+    def test_unauthenticated_user_redirect(self):
+        """Verify that an unauthenticated user can't visit '/dashboard'"""
+        response = self.client.get('/dashboard/')
```
noisecapella: Can you use the `reverse` function here? It looks up the url given the url name, which are specified in `ui/urls.py`

ui/views_test.py
```diff
@@ -32,6 +35,27 @@ def test_program_liveness(self):
             status_code=200
         )
 
+    def test_unauthenticated_user_redirect(self):
+        """Verify that an unauthenticated user can't visit '/dashboard'"""
+        response = self.client.get(reverse('ui-dashboard'))
+        self.assertRedirects(
+            response,
+            "/?next=/dashboard/"
```
noisecapella: It looks like the path `/dashboard/` is used here, can you use `reverse` here too?

aliceriot: ah, yeah I suppose I could!

ui/views.py
```diff
@@ -52,6 +53,7 @@ def index(request):
     })
 
 
+@login_required(login_url="/")
```
noisecapella: `login_url` is unnecessary here since we define [LOGIN_REDIRECT_URL](https://docs.djangoproject.com/en/1.9/ref/settings/#login-redirect-url) in `settings.py` to be the same

ui/views_test.py
```diff
@@ -32,6 +35,27 @@ def test_program_liveness(self):
             status_code=200
         )
 
+    def test_unauthenticated_user_redirect(self):
+        """Verify that an unauthenticated user can't visit '/dashboard'"""
+        response = self.client.get(reverse('ui-dashboard'))
+        self.assertRedirects(
+            response,
+            "/?next=" + reverse('ui-dashboard')
```
noisecapella: It would be better to use a format string here (ie: `/?next={}".format(url)`). Format strings are easier to work with since they will convert any argument to a string, and they're better for localization since the string being localized is in one piece, for example: http://pylonsbook.com/en/1.1/internationalization-and-localization.html#marking-strings-for-internationalization

ui/views.py
aliceriot: for some reason I had trouble if I didn't set `login_url`, I checked and confirmed that in `settings.py` it's set to `/`, but it was redirecting me to `/login/` anyway.

from this: https://docs.djangoproject.com/en/1.9/topics/auth/default/#django.contrib.auth.decorators.login_required I think it might be because we don't associated `settings.LOGIN_URL` with a view? not sure!

noisecapella: Yes I think you're right. Can you add `LOGIN_URL` to `settings.py`?

noisecapella:
Looks good! :+1:

PR #148, opened 2016-04-07
Added google analytics to base ui template

### LINKED ISSUE #135: Add Google Analytics to django template
So we can get page views for the home page

aliceriot:

closes #135

#### What's this PR do?

this adds google analytics to the base template for the ui app.

first, this adds a template tag for the ui app that lets us pull values
out of `settings.py` (we store the google analytics id there).

then we can just pop the google analytics JS snippet into the `<head>`
of our base template, and pull in the analytics id using the template
tag. Nice!

`/ui/templatetags/ui_extras.py` <- this is the thing I'm least sure
about.

we can generate a throwaway google analytics account, set the
environment variable `GA_TRACKING_ID` to the tracking id for that
account (in the micromasters `.env`), and then confirm that visiting any
page results in a page view being registered.

I was a little unsure as to how to get the environment variable defining
our tracking ID into the template. this SE thread was handy:

https://stackoverflow.com/questions/433162/can-i-access-constants-in-settings-py-from-templates-in-django

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -11,6 +11,17 @@
     <title>{% block title %}{% endblock %}</title>
     {% block extrahead %}
     {% endblock %}
+    <!-- Google Analytics -->
+    <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+    ga('create', '{{ GA_TRACKING_ID }}', 'auto');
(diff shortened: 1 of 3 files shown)


ui/templates/base.html
```diff
@@ -11,6 +11,18 @@
...
     {% endblock %}
+    <!-- Google Analytics -->
+    <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+    {% load ui_extras %}
+    ga('create', '{% settings_value "GA_TRACKING_ID" %}', 'auto');
```
noisecapella: Can you pass in `{{ GA_TRACKING_ID }}` instead and set the value in the context instead? That would avoid the need for the `templatetags` directory

noisecapella:
:+1:

PR #164, opened 2016-04-11
Dashboard REST API

### LINKED ISSUE #125: Student Dashboard API
replaces #74 

For a given user id, return a list of courses in the current micromasters with a status for each course 

For example:

title: Intro to SCM
status: 
- course has not been offered (start date doesn't exist or is in the future)
- user not enrolled, but course is offered with enrollment date in the future
- user not enrolled, course is open for enrollment now
- user enrolled but not verified
- user enrolled but not verified and verification date has passed
- verified but not completed
- verified, completed and not passed
- verified, completed and passed
  
some additional fields, depending on what the status is, e.g. user grade

giocalitri:
fixes #125 

#### What's this PR do?
Introduces the dashboard api

course models, then dashboard api, then dashboard view

Add some course runs for your courses with dates in a way to have past present and future run.
You may want to make those run correspondent to real course on edX where the user you are using has enrolled/not enrolled


courses/admin.py
```diff
@@ -17,9 +17,10 @@ class CourseRunInline(admin.StackedInline):
 
 class CourseAdmin(admin.ModelAdmin):
     """ModelAdmin for Courses"""
-    list_display = ('title',)
+    list_display = ('position_in_program', 'title',)
     list_filter = ('program__live',)
     inlines = [CourseRunInline]
+    ordering = ('position_in_program',)
```
noisecapella: Can we set this ordering on the model instead?

courses/factories.py
```diff
@@ -18,10 +18,18 @@ class CourseFactory(DjangoModelFactory):
...
     program = factory.SubFactory(ProgramFactory)
+    position_in_program = factory.Sequence(lambda n: n)
 
     class Meta:  # pylint: disable=missing-docstring
         model = Course
 
+    @classmethod
+    def _setup_next_sequence(cls):
+        try:
+            return Course.objects.latest('position_in_program').position_in_program + 1
```
noisecapella: It seems like `latest` is defined to be for date values: https://docs.djangoproject.com/en/1.9/ref/models/querysets/#latest. Can you use `order_by()` and `last()` instead?

courses/factories.py
```diff
@@ -18,10 +18,18 @@ class CourseFactory(DjangoModelFactory):
...
+        except Course.DoesNotExist:
+            return 0
+
 
 class CourseRunFactory(DjangoModelFactory):
```
noisecapella: Can you update `CourseRunFactory` with the various start and end date fields? These will all default to `None` if we don't add a field here

requirements.txt
```diff
@@ -19,3 +19,4 @@ tox==2.3.1
 uwsgi==2.0.12
 Pillow==3.1.1
 wagtail==1.4.3
+-e git+git://github.com/mitodl/edx-api-client.git#egg=edx-api-client
```
noisecapella: Are we able to use a version published on pypi?

micromasters/urls.py
```diff
@@ -21,6 +22,7 @@
     url('', include('social.apps.django_app.urls', namespace='social')),
     url(r'^admin/', include(admin.site.urls)),
     url(r'^api/v0/', include(router.urls)),
+    url(r'^api/v0/dashboard/?$', UserDashboard.as_view(), name='dashboard_api'),
```
noisecapella: Can you remove the `?` from the regex? Django should redirect it automatically if the trailing slash is not present

courses/models_test.py
```diff
@@ -15,10 +24,248 @@ def test_to_string(self):
...
         """Test for __str__ method"""
-        course = CourseFactory.build(title="Title")
-        assert "{}".format(course) == "Title"
+        assert "{}".format(self.course) == "Title"
+
+    def test_get_next_run_no_run(self):
+        """Test for the get_next_run method"""
+        assert self.course.get_next_run() is None
+
+    def test_get_next_run_only_past(self):
```
noisecapella: Is it possible to use a parameterized test for this? Each parameter could be tested with before, after, and `None` values. It seems like we assume that if `end` is set then `start` is always set, for example, there are many unhandled test cases like this. Feel free to ignore if it's a pain to do

courses/admin.py
giocalitri: I do not think we need it in the model: I needed here just because it's easier to see the courses

noisecapella: Sounds good

courses/models_test.py
giocalitri: I actually thought about doing this, but we will basically move the messiness from inside the test to the parameters..

requirements.txt
giocalitri: we are still too unstable with the library: as soon as I add the certificates part, I will make a release and change this

courses/models.py
```diff
@@ -33,6 +37,30 @@ class Course(models.Model):
...
+        Note that it could be current as long as the enrollment window
+        has not closed.
+        """
+        now = datetime.now(pytz.utc)
+        next_run_set = self.courserun_set.filter(
+            # there is only the start date
+            (models.Q(enrollment_start=None) & models.Q(enrollment_end=None) &
+             models.Q(end_date=None) & models.Q(start_date__lte=now)) |
+            # there is no enrollment date but the course has not ended yet
+            (models.Q(enrollment_end=None) & models.Q(end_date__gte=now)) |
```
noisecapella: What if the enrollment end is in the past but the course end date is in the future? Just trying to get a handle on this expression

micromasters/urls.py
giocalitri: sure

courses/factories.py
giocalitri: sure

courses/models.py
giocalitri: this case is covered by the following expression (unless I misunderstood what you mean)

noisecapella: The next expression tests if enrollment end is in the future. If it's in the past then it wouldn't be covered by any of this I think

courses/models_test.py
```diff
@@ -15,10 +24,248 @@ def test_to_string(self):
...
+        """Test for the get_next_run method"""
+        # run in the past
```
noisecapella: Can you move these comments into the docstrings for each test?

courses/models_test.py
noisecapella: I looked over these test cases and I think these are uncovered. Can you add tests for the ones I didn't mark done?

    enrollment past, course past - done
    enrollment past, course present - done
    enrollment past, course future
    enrollment past, course none
    enrollment present, course past
    enrollment present, course present - done
    enrollment present, course future
    enrollment present, course none
    enrollment future, course past
    enrollment future, course present
    enrollment future, course future - done
    enrollment future, course none
    enrollment none, course past
    enrollment none, course present - done
    enrollment none, course future
    enrollment none, course none

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+
+    @classmethod
+    def get_assoc_field(cls, course_status):
+        """
+        Method to get from the ASSOCIATED_FIELDS dict
+        """
+        return cls.ASSOCIATED_FIELDS.get(course_status, [])
+
+
+class CourseRunUserStatus(object):
```
noisecapella: FYI we don't need this `(object)` syntax anymore with Python 3, but feel free to leave it

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def get_info_for_course(user, course, user_enrollments, user_certificates):
+    """
+    Checks the status of a course given the status of all its runs
+
+    Args:
+        user (User): an user object
+        course (Course): a course object
+        user_enrollments (Enrollments): the user enrollments object
```
noisecapella: I think this and `Certificates` below should both be `dict`

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Representation of a course run status for a specific user
+    """
+    def __init__(self, status, course_run=None, enrollment_for_course=None):
+        self.status = status
+        self.course_run = course_run
+        self.enrollment_for_course = enrollment_for_course
+
+
+def get_info_for_course(user, course, user_enrollments, user_certificates):
```
noisecapella: My IDE says that `user_certificates` is unused here. Will this be used in the future?

courses/models.py
giocalitri: but if the `enrollment_end` is in the past, it means that the enrollment are closed, so the `course_run` is not `next`

dashboard/api.py
giocalitri: the `user_certificates` are not used because there is not way to query for them yet.

dashboard/api.py
giocalitri: No, it will be a specific object like it is `Enrollments`

dashboard/api.py
giocalitri: BTW: `# pylint: disable=unused-argument` is below for this reason

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+
+def format_course_for_dashboard(course_run, status_for_user, course, certificate_for_course=None):
+    """
+    Helper function that formats a course run adding informations to the fields coming from the DB
+
+    Args:
+        course_run (CourseRun): a course run
+        status_for_user (str): a string representing the status of a course for the user
+        course (Course): a course
+        certificate_for_course (Certificate): an object representing the
```
noisecapella: Can you change this to `dict` too?

giocalitri: see comment before

courses/models.py
noisecapella: I see, so this function only gets next `CourseRun`s whose enrollments are open or where the enrollment is unspecified so we look to the course start/end dates for that

giocalitri: yes: just talked with @pdpinch and we agreed that I should not consider the `enrollment_start` at all, so I will slightly modify this function

courses/models_test.py
giocalitri: after talking with @pdpinch, we agreed that I should work with the assumption that the dates cannot be "wrong" (with "wrong" defined by the constraints that edX puts on the dates), meaning the cases should be:

    enrollment past, course past - done
    enrollment past, course present - done
    enrollment past, course future
    enrollment past, course none - this is not possible
    enrollment present, course past - this is not possible
    enrollment present, course present - done
    enrollment present, course future
    enrollment present, course none - this is not possible
    enrollment future, course past - this is not possible
    enrollment future, course present - this is not possible
    enrollment future, course future - done
    enrollment future, course none - this is not possible
    enrollment none, course past
    enrollment none, course present - done
    enrollment none, course future
    enrollment none, course none - this is not possible

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+        user_certificates (Certificates): the user certificates objects
+
+    Returns:
+        dict: dictionary representing the course status for the user
+    """
+    # pylint: disable=unused-argument, too-many-return-statements
+    if not course.courserun_set.count():
+        return format_course_for_dashboard(None, CourseStatus.NOT_OFFERED, course)
+    # get all the run statuses
+    run_statuses = []
```
noisecapella: Nit: Can you use a list comprehension?

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+    for course_run in course.courserun_set.all():
+        run_statuses.append(get_status_for_courserun(course_run, user_enrollments))
+    # sort them by end date
+    run_statuses.sort(key=lambda x: x.course_run.end_date or datetime.datetime(datetime.MAXYEAR, 1, 1), reverse=True)
```
noisecapella: According to the postgres docs it already sorts null values as being greater than any other value (from http://www.postgresql.org/docs/8.3/static/queries-order.html), so you may be able to just sort this with an `order_by` instead

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+    # pick the first `not enrolled` or the first
+    for run_status in run_statuses:
+        if run_status.status != CourseRunStatus.NOT_ENROLLED:
+            break
+    else:
+        run_status = run_statuses[0]
```
noisecapella: Nit: There's a tiny chance that `course.courserun_set.count()` will be non-zero but the later `course.courserun_set.all()` will have nothing in it, which would cause an out of bounds error here

giocalitri: sorry, missing the case when this could happen...

dashboard/api.py
noisecapella: Nevermind, it looks like you have to do this in Python

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+            status = CourseRunStatus.READ_CERT
+        elif course_run.is_future:
+            status = CourseRunStatus.WILL_ATTEND
+    else:
+        if course_run.is_current or course_run.is_future:
+            status = CourseRunStatus.UPGRADE
+        elif course_run.is_past:
+            status = CourseRunStatus.NOT_PASSED
+    return CourseRunUserStatus(
+        status=status,
```
noisecapella: Can you define `status = None` above? My IDE is complaining that `status` might not be defined in all cases

courses/models_test.py
noisecapella: This is fine with me but we should make sure in the future that only edX is able to alter these dates, or whatever UI we use to alter them does the necessary validations

dashboard/api.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    PASSED = 'passed'
+    NOT_PASSED = 'not-passed'
+    CURRENT_GRADE = 'verified-not-completed'
+    UPGRADE = 'enrolled-not-verified'
+    NOT_OFFERED = 'not-offered'
+    OFFERED = 'offered-not-enrolled'
+
+    @classmethod
+    def all_statuses(cls):
```
noisecapella: It looks like we aren't using this function anywhere but the tests. Do we need it in the future?

dashboard/api_test.py
```diff
@@ -0,0 +1,402 @@
...
+        all_constants = [value for name, value in vars(api.CourseStatus).items()
+                         if not name.startswith('_') and isinstance(value, str)]
+        assert sorted(all_constants) == sorted(api.CourseStatus.all_statuses())
+
+    def test_course_run_status(self):
+        """test for CourseRunStatus"""
+        for attr in ('NOT_ENROLLED', 'GRADE', 'READ_CERT', 'WILL_ATTEND', 'UPGRADE', 'NOT_PASSED'):
+            assert hasattr(api.CourseRunStatus, attr)
+
+    def test_corse_run_user_status(self):
```
noisecapella: Typo: `course`

dashboard/api_test.py
```diff
@@ -0,0 +1,402 @@
...
+        if edx_key is not None:
+            run.edx_course_key = edx_key
+            run.save()
+        return run
+
+
+class FormatRunTest(CourseMixin):
+    """Tests for the format_course_for_dashboard function"""
+
+    def test_format_run_no_run(self):
```
noisecapella: Can you add a test that uses one of the statuses in `CourseFormatConditionalFields` and verifies that it added the fields correctly?

dashboard/api_test.py
```diff
@@ -0,0 +1,402 @@
...
+            {
+                'title': self.course.title,
+                'status': 'status_foo',
+                'position_in_program': self.course.position_in_program
+            }
+        )
+
+    def test_format_run(self):
+        """Test for format_course_for_dashboard"""
+        now = datetime.now()
```
noisecapella: It looks like we use `datetime.now(pytz.utc)` everywhere else, should we use it here too?

courses/models.py
```diff
@@ -33,6 +37,30 @@ class Course(models.Model):
...
+    class Meta:
+        unique_together = ('program', 'position_in_program',)
+
+    def get_next_run(self):
+        """
+        Gets the next run associated with this course.
+        Note that it could be current as long as the enrollment window
+        has not closed.
+        """
+        now = datetime.now(pytz.utc)
```
noisecapella: Should we test that we're using `pytz.utc` here? I'm not sure if it's worth it

dashboard/api_test.py
```diff
@@ -0,0 +1,402 @@
...
+        with patch(
+            'dashboard.api.get_status_for_courserun',
+            autospec=True,
+            return_value=api.CourseRunUserStatus(
+                status=api.CourseRunStatus.WILL_ATTEND,
+                course_run=self.course_run
+            )
+        ):
+            api.get_info_for_course(self.user, self.course, None, None)
+        mock_format.assert_called_once_with(self.course_run, api.CourseStatus.CURRENT_GRADE, self.course)
```
noisecapella: What would the grade be if the status is WILL_ATTEND?

courses/models_test.py
giocalitri: In theory we should pull the dates from edX: we will need to implement this

noisecapella:
I'm impressed at the test coverage here given the scope of the problem. Testing functionality now

dashboard/api.py
giocalitri: I might change the `CourseFormatConditionalFields. get_assoc_field` to use this to enforce the existence of the status

courses/models.py
giocalitri: me neither...

dashboard/api.py
giocalitri: okay

dashboard/api_test.py

dashboard/api_test.py
giocalitri: yes, we should

dashboard/api_test.py
giocalitri: no clue: I guess it will depend on the implementation of the API that will be used to pull this data. I suspect it will be `None` or `0`

dashboard/api.py
noisecapella: I found a bug here in functional testing. If end_date is None and another end_date is a valid date, it will fail to compare to `datetime(datetime.MAXYEAR...` because the time zone is not set. I wrote a test to reproduce this:

    def test_course_run_end_date_mixed(self):
        """
        Test with a mix of end_date being None and also a valid date
        """
        def mocked_get_status_for_courserun(run, enrollments):  # pylint: disable=unused-argument
            """Mock get_status_for_courserun with different values for each run"""
            return api.CourseRunUserStatus(
                status=api.CourseRunStatus.NOT_ENROLLED,
                course_run=run
            )

        run1 = CourseRunFactory.create(
            start_date=datetime.now(pytz.utc),
            end_date=None
        )
        CourseRunFactory.create(
            start_date=datetime.now(pytz.utc),
            end_date=datetime.now(pytz.utc),
            course=run1.course
        )
        with patch(
            'dashboard.api.get_status_for_courserun',
            autospec=True,
            side_effect=mocked_get_status_for_courserun
        ):
            api.get_info_for_course(self.user, run1.course, None, None)

noisecapella:
Functionality looks good aside from the bug I commented on

courses/models_test.py
giocalitri: sure

dashboard/api.py
giocalitri: I actually forgot about this: will fix

dashboard/api.py
giocalitri: sometimes I wonder where my brain is :-)

courses/factories.py
noisecapella: It looks like this comment is unaddressed, but I'll just file a ticket for a later PR

dashboard/api.py
```diff
@@ -0,0 +1,236 @@
...
+def get_status_for_courserun(course_run, user_enrollments):
+    """
+    Checks the status of a course run for a user given her enrollments
+
+    Args:
+        course_run (CourseRun): a course run
+        user_enrollments (Enrollments): the user enrollments
+
+    Returns:
+        CourseRunStatus: an object representing the run status for the user
```
noisecapella: This should be `CourseRunUserStatus`

courses/factories.py
giocalitri: I'm sorry, I totally missed this

noisecapella:
Just the one comment :+1:

dashboard/api.py
giocalitri: will change before squashing

PR #172, opened 2016-04-12
Moved our pipeline functions down the python-social-auth pipeline

### LINKED ISSUE #170: User cannot login if an exception occurs early in the social auth pipeline
I ran into a problem when the `update_profile_from_edx` function threw an Exception due to unrelated work I did. In this case the `User` is created but the user can't login using that user. When they next attempt to login a new user is generated with a new username which won't match the edX username. This will cause us to fail to get their profile information, which prevents them from logging in the second time.

To reproduce:
 - Add a `raise Exception()` in `update_profile_from_edx`
 - Delete your `User` or start from scratch with a new docker container
 - Login and authorize with edX. You should get the exception you put in `update_profile_from_edx`
 - Remove the exception you added in the first step
 - Try to login again. This time you will get a different exception complaining about being forbidden from accessing the user profile REST API. If you look at the URL you will see that some characters are appended to the username. A new `User` was created for this login with a username that doesn't match edX, so future logins will always fail.

To solve, we can move our functions to the end of the social auth pipeline. At this point even if our code raises an exception, it won't prevent python-social-auth from linking edX to the account and they'll be able to login again with the same `User`

noisecapella:
Fixes #170 

#### What's this PR do?
Moves our pipeline functions to the end of the pipeline

See #170 for instructions on how to reproduce the problem. With this PR you should not see the problem anymore.

I tried writing a test for this but there's not much documentation on writing an integration test involving the pipeline so that we can reproduce the problem.

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -157,11 +157,11 @@ SOCIAL_AUTH_PIPELINE = (
     'social.pipeline.social_auth.social_user',
     'social.pipeline.user.get_username',
     'social.pipeline.user.create_user',
-    'backends.pipeline_api.update_profile_from_edx',
-    'backends.pipeline_api.update_from_linkedin',
     'social.pipeline.social_auth.associate_user',
     'social.pipeline.social_auth.load_extra_data',
     'social.pipeline.user.user_details',
+    'backends.pipeline_api.update_profile_from_edx',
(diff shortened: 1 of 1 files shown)


annagav:
I was able to login after an exception was thrown on first login to mm. 
👍

Issue #69, opened 2016-03-14
Student Dashboard: Create and display Student ID


jamiefolsom:
As a student, I want my "student identification number" stored in my profile, for use in verifying my identity, retrieving my enrollments, grades, etc.

The ID should look like : 

MMM ### ####
For example 

MMM 000 0001
MMM 000 0002
etc.

Numbers will be serial. We will use a separate field, to avoid confusing the user primary key with this "student id"

- [x] add to model
- [x] update REST API
- [x] add to dashboard react component

pdpinch:
MIT student IDs are 9 digit numbers. For now, let's plan to do the same thing. 

Perhaps at some point we will engage the bureaucracy to see if we can get bona fide student ids for micromasters candidates. I'm not sure where to start that conversation though.

pdpinch:
I updated the description to capture the discussion in planning.

aliceriot:
just opened #173 which addresses some of this (adds the attribute, adds it to the model serializer), but I think we're waiting on some more React work to include this somewhere in the frontend (unless I'm misunderstanding how it's to be included in the UI).

noisecapella:
I think you should be able to start now, the dashboard has the user profile image and name already. There are some mockups here: https://projects.invisionapp.com/share/KR6FYCZTX#/screens/142338434

aliceriot:
ah cool, not seeing that I thought the id was only supposed to appear at `/dashboard/profile`

PR #173, opened 2016-04-12
Added auto-incrementing `student_id` field to `Profile`

aliceriot:
#### relevant tickets:

this accomplishes part of #69

#### what's this pr do?

this adds a new attribute to the Profile model (`student_id`) that we
basically keep incremented.

to do so we override the `save` method on the Profile model, so that we
can insert our own logic to set the correct `student_id` on `self`
before we save (only if the model is new or the `student_id` field is
currently empty).

we also add this attribute to the `ProfileSerializer`, which is what
spits the profile JSON to the frontend.

#### where should the reviewer start?

the migration and the model change:

	new file:   profiles/migrations/0005_profile_student_id.py
	modified:   profiles/models.py

the serializer change is smaller, I think

	modified:   profiles/serializers.py

#### how should this be manually tested?

I think the key thing is testing that when you save a `Profile` instance
it has a unique `student_id`. It should save a `student_id` when saving
both new and previously created objects.

#### any background context?

I modeled this change over how we accomplished a somewhat (I think?)
similar thing over in Lore.

The relevant pieces are here:

https://github.com/mitodl/lore/blob/master/learningresources/models.py#L163

https://github.com/mitodl/lore/blob/master/rest/util.py#L66


profiles/models.py
```diff
@@ -99,6 +100,26 @@ class Profile(models.Model):
...
+        if self.id is None or self.student_id is None:
+            max_id = Profile.objects.aggregate(Max('student_id'))
+            if max_id['student_id__max'] is None:
+                max_id = 1
+            else:
+                max_id = max_id['student_id__max']
+            while Profile.objects.filter(student_id=max_id).exists():
+                max_id += 1
+            self.student_id = max_id
+            print("here")
```
noisecapella: Can you remove this line?

aliceriot: yep, not yet ready for review :)

noisecapella: Sorry, I missed that

aliceriot: no worries :smiley:

aliceriot:
Screenshot of appearance in the dashboard: 

react_view

static/js/components/Dashboard.js
```diff
@@ -2,9 +2,14 @@
 import React from 'react';
 import CourseList from './CourseList';
 import UserImage from './UserImage';
+import _ from 'lodash';
 
 class Dashboard extends React.Component {
 
+  formatId(id) {
+    return id ? _.padStart(id.toString(), 6, '0') : undefined;
```
pdpinch: Please format the ID `MMM ### ####` -- just for readability

aliceriot: sure!

profiles/models.py
```diff
@@ -99,6 +100,25 @@ class Profile(models.Model):
...
+
+    @transaction.atomic
+    def save(self, *args, **kwargs):
+        """need to handle setting the student_id number"""
+        if self.id is None or self.student_id is None:
+            max_id = Profile.objects.aggregate(Max('student_id'))
+            if max_id['student_id__max'] is None:
+                max_id = 1
+            else:
+                max_id = max_id['student_id__max']
```
giocalitri: you can write
```
if max_id['student_id__max'] is None:
    max_id = 1
else:
    max_id = max_id['student_id__max']
```
in a more pythonic way:
`max_id = max_id['student_id__max'] or 1`

profiles/models.py
```diff
@@ -99,6 +100,25 @@ class Profile(models.Model):
...
+            while Profile.objects.filter(student_id=max_id).exists():
+                max_id += 1
```
giocalitri: I suspect the reason for this while is the possibility of multiple save happening concurrently.
The idea is good, but I dislike the possibility (even if it is unlikely) of an infinite loop, so I would use a finite big number of retries.

profiles/models.py
```diff
@@ -99,6 +100,25 @@ class Profile(models.Model):
...
+            self.student_id = max_id
+        super(Profile, self).save(*args, **kwargs)
 
     def __str__(self):
         return 'Profile for "{0}"'.format(self.user.username)
+
+    def pretty_printed_student_id(self):
+        """pretty prints the student id for easy display"""
+        return "MMM{0:06}".format(self.student_id)
```
giocalitri: I would make this a property method:

```
@property
def pretty_printed_student_id(self):
```

so you can access it as a property and not a method:
`user_profile. pretty_printed_student_id`

profiles/serializers.py
```diff
@@ -35,6 +35,7 @@ class Meta:  # pylint: disable=missing-docstring
             'gender',
             'mailing_address',
             'date_joined_micromasters',
+            'student_id',
```
giocalitri: I would use the `pretty_printed_student_id` instead

giocalitri: maybe clarify with whoever will add this in the UI what's best

static/js/components/Dashboard.js
```diff
@@ -2,9 +2,14 @@
 import React from 'react';
 import CourseList from './CourseList';
 import UserImage from './UserImage';
+import _ from 'lodash';
 
 class Dashboard extends React.Component {
 
+  formatId(id) {
+    return id ? `MMM${_.padStart(id.toString(), 6, '0')}` : undefined;
```
giocalitri: ok, never mind the previous comment: if you "pretty print" the number in the backend, there is not reason to repeat the job in the front-end.
I think this should be done in the backend and here you just put the string

giocalitri:
There are few things to change, but nothing serious.

profiles/models.py
aliceriot: nice and compact! I like it.

profiles/models.py
noisecapella: Where is the infinite loop here?

profiles/models.py
aliceriot: ah, so much django learning to do!

in rails there is no distinction when calling a method or a property.

profiles/models.py
aliceriot: I think if we somehow ran into a perfect race condition it could be the case that `Profile.objects.filter(student_id=max_id).exists()` is true for a long tim.

profiles/models_test.py
```diff
@@ -0,0 +1,18 @@
...
+from profiles.factories import ProfileFactory, UserFactory
+
+
+class ProfileTests(TestCase):
+    """tests for the profile model"""
+
+    def student_id_on_save(self):
+        """test that a student id is set on save"""
+        user = UserFactory.create()
+        profile = ProfileFactory.create(user=user)
```
noisecapella: `ProfileFactory.create` should have saved the `Profile` to the database so it's surprising that `student_id` is `None` here. Maybe `ProfileFactory.create` is not calling save? Can you rewrite the test to use `Profile.objects.create` instead?

profiles/models_test.py
```diff
@@ -0,0 +1,18 @@
...
+        assert profile.student_id is None
+        profile.save()
```
noisecapella: Can you add a test that verifies that the student ids are incremented by one? And a test where `Profile` already has a student id and `save()` is called?

static/js/components/Dashboard.js
aliceriot: yeah, that makes sense.

I thought there might be some reason why we want the actually number in the frontend, but we can always change it later if we do.

profiles/models.py
noisecapella: But we're doing the transaction atomically so nothing else should be able to race us in the database.

profiles/models.py
giocalitri: that is actually pure python :-)

profiles/models_test.py
giocalitri: actually, since the profile is created with a signal after the user is created, this call should just fail.

static/js/components/Dashboard.js
noisecapella: I can go both ways here. It seems like we should only be providing raw data from the server (at least through REST APIs) and do all formatting on the frontend. On the other hand you could argue that the `MMM...` string is our identifier and therefore a piece of raw data.

Whichever you choose, we should have a test for this

profiles/models.py
giocalitri: @noisecapella if there is a `while` with no upper bound, I think we need to assume it will be an infinite loop.

giocalitri: @noisecapella you are doing a transaction atomically, but you are not pulling the number from a sequence, this means that, without this code, 2 different processes on the database could potentially pick the same `Max()` number

noisecapella: At that point wouldn't both processes increment the max number, try again, and then one transaction has to roll back, and an Exception is raised. At that point the user could try logging in again and it would probably succeed, since the chance of this error happening in the first place is pretty low

aliceriot: I made a change so that each time `max_id` is incremented we also decrement a counter that starts at 1000. if this reaches zero we let the transaction happen and raise an exception.

does that do what we want?

noisecapella:
@aliceriot The incrementing code looks fine to me, I think it would work either way

aliceriot:
cool, working on adding some more tests right now

profiles/models.py
giocalitri: @noisecapella If you are questioning the likelihood of this happening I agree with you. If we are talking about why a `while` without upper bound should exist, my opinion is absolutely no

giocalitri: @noisecapella I also think the right solution is not to let an exception be raised, but, if we really want to rely only on the DB transaction, we should have the field be a sequence (if we do not care about potential missing numbers).

profiles/models_test.py
noisecapella: @aliceriot You may find this code snippet useful: https://github.com/mitodl/micromasters/blob/master/profiles/permissions_test.py#L27

Since we use Django signals to create a profile (see https://github.com/mitodl/micromasters/blob/master/profiles/signals.py#L17) you will probably need to use this `mute_signals` context manager

aliceriot:
updated appearance:

updated_appearance

Ferdi:
Not sure about "MITx id" here.  Probably simply

ID: MMM-000001

aliceriot:
cool, was going based on this mockup: https://projects.invisionapp.com/share/KR6FYCZTX#/screens/142338434

I can change it to just `id`.

profiles/models.py
```diff
@@ -121,6 +122,25 @@ class Profile(models.Model):
...
+            self.student_id = max_id
+        super(Profile, self).save(*args, **kwargs)
 
     def __str__(self):
         return 'Profile for "{0}"'.format(self.user.username)
+
+    @property
+    def pretty_printed_student_id(self):
+        """pretty prints the student id for easy display"""
+        return "MMM{0:06}".format(self.student_id)
```
noisecapella: It looks like this complains when `self.student_id` is `None`. Can you return an empty string for that case?

noisecapella: Also add a test for this case please

static/js/components/Dashboard.js
```diff
@@ -14,6 +14,9 @@ class Dashboard extends React.Component {
         </div>
         <div className="card-name">
           { SETTINGS.name }
+          <div className="card-student-id">
+            { profile.student_id }
```
noisecapella: I think this needs to be changed to `profile.pretty_printed_student_id`

noisecapella: Can you also update `USER_PROFILE_RESPONSE` in `constants.js` to add this new field?

profiles/models.py
aliceriot: yes, good catch

static/js/components/Dashboard.js
aliceriot: indeed! before I was passing the integer to the frontend and formatting it there.

noisecapella:
Looks good to me :+1:

aliceriot:
woooo

PR #175, opened 2016-04-12
Added missing configuration for heroku to app.json

bdero:
We'll know this works if the respective environment variables are copied from micromasters-ci to the review app of this PR.
@noisecapella


noisecapella:
:+1:

Issue #149, opened 2016-04-11
User enrollment: personal info card (part of profile creation)


pdpinch:
When I first enroll, I need to view and update my personal profile information, as it was imported from edX. 

Fields are described in https://docs.google.com/spreadsheets/d/1ehN3FNymCOhpJT7h9XZDOITP_lOFDnEcStRa0le4MX8/edit?pref=2&pli=1#gid=0

This will be the first tab of three, presented the first time the user visits the site. The user will not be able to proceed without completing the required fields. 

edX will provide a full name which we should attempt to split into `family_name` and `given_name.` `preferred_name` is optional.  

Countries should be normalized, ideally using the same list as edX. (It's possible that we may need to restrict users from some countries)

![enroll_personal](https://cloud.githubusercontent.com/assets/430126/14503033/399853c0-017c-11e6-8707-a9abc381b998.png)

pdpinch:
"To enroll in MIT Micromasters you must fill in the information below"

pdpinch:
We will POST/PATCH when the user closes the modal -- to avoid losing data in progress

pdpinch:
@noisecapella I updated the fields in the spreadsheet

pdpinch:
I added a wireframe.

noisecapella:
Is preferred language a dropdown? Should we have state in addition to city and country?

noisecapella:
It looks like edX uses ISO 639-1 language codes but they only support English in the devstack by default. edx.org has four languages to choose from

noisecapella:
We should have a message that says that the user will opt in to receive email by submitting their profile

pdpinch:
Actually, I was referring to `language_proficiencies` in the edX profile, not `language`

I believe the user can choose from any of the ISO 639-1 languages for their proficiencies, but language is limited to the 4 supported languages. 

Our micromasters site is only going to support English (as far as I know) but I think MIT is interested in the range of it's users' proficiencies. 
> It looks like edX uses ISO 639-1 language codes but they only support English in the devstack by default. edx.org has four languages to choose from
>

pdpinch:
> We should have a message that says that the user will opt in to receive email by submitting their profile

Thanks for the reminder! I updated the wireframe. 

I assume we still have a field in the database for tracking opt-in status, even though we're not letting people opt-out here.

noisecapella:
Can you upload the wireframe again?

noisecapella:
[...]

PR #180, opened 2016-04-13
Added fields, updated serializer for personal data on profile

noisecapella:
Refs #149 

#### What's this PR do?
 - Adds some fields for personal data in profile, removes year_of_birth to be replaced by date_of_birth
 - Renames fields we don't plan on using to have `edx_` prefix. This includes certain job or school information which we'll deal with in a later PR
 - Populates some of these new fields with information from edX. first_name and last_name are created by putting the first word in first_name and all other words in last_name
 - Adds tests for dashboard `SETTINGS`
 - Some fields removed from serializer which we don't plan on using

profiles/models.py

Run database migrations, go to `/api/v0/profiles/<username>/` and see the updated fields. You should also be able to PATCH most fields.

This is part of #149, the UI portion will be in a separate PR


profiles/util_test.py
```diff
@@ -0,0 +1,54 @@
...
+        assert first_name == "two"
+        assert last_name == "names"
+
+    def test_split_name_more_than_two(self):
+        """
+        Split name should be limited to two names
+        """
+        first_name, last_name = split_name("three names here")
+        assert first_name == "three"
+        assert last_name == "names here"
```
giocalitri: I do not think this is an easy function to implement, but there are people with multiple first name (and no middle).
Not sure we care at all

giocalitri:
this looks pretty good: a single question.
I did not tests the functionality yet

profiles/util_test.py
noisecapella: `split_name` is just used to create inital first and last names, they will have the opportunity to adjust this if we get it wrong during the profile edit.

giocalitri:
I am not sure when this problem was introduced, but `account_privacy` in edx can have a value of `all_users` that does not seem to be allowed any more.

giocalitri:
functionality, other than the bug I found with the field coming from edX, is fine :+1:

noisecapella:
There are different settings for account privacy for edX and MM but we forgot to translate the account privacy correctly. I'll fix that and add a test for it

PR #181, opened 2016-04-13
Added stata center image to Dashboard user card background

aliceriot:
#### What are the relevant tickets?

this closes #161

#### What's this PR do?

This adds a background image for the user info card we should at
`/dashboard`.

#### Where should the reviewer start?

only non-asset change is in `static/scss/layout.scss`.

#### How should this be manually tested?

pull it down, make sure the image shows up.

#### Screenshots (if appropriate)

it looks like this:

background

PR #183, opened 2016-04-14
Fixed a small bug with the dashboard background image

aliceriot:

none, but I introduced the change here: #181

#### What's this PR do?

makes one css change to eliminate this strange gap.

changes:

	modified:   static/scss/layout.scss

make sure it looks right!

before:

bad_css_before

after:

bad_css_after


noisecapella:
:+1: 
It looks like the strange gap is actually part of the image, but I'm fine with this solution

PR #185, opened 2016-04-15
Fixed a JS console warning, react-router and react-bootstrap

aliceriot:

none, this fixes a bug I noticed.

#### What's this PR do?

this adds a library called react-router-bootstrap that is designed to
make react-router and react-bootstrap play nicely together.

basically, we were getting a JS warning that looks like this:

```
warning.js:45 Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>. See LoginButton > MenuItem > SafeAnchor > a > ... > Link > a
```

this is occurring because we had the following code:

```javascript
<MenuItem>
  <Link to="/profile">
    Profile
  </Link>
</MenuItem>
```

`MenuItem` comes from `react-bootstrap`, and `Link` is a `react-router`
link. basically, the problem is that `MenuItem` implicitly includes an
`<a ..></a>` element, and places `this.props.children` inside that
element.

hence, we get a console warning about improperly nesting `<a>` tags.

of course, the outer one has `href=undefined`, so only the inner
`<Link>` tag does anything, and the ui works as expected, but still, the
warning.

the `react-router-bootstrap` library saves us here. we get a component
called `<LinkContainer>` which can wrap any `react-bootstrap` component
and make it behave like a `react-router` `<Link>`. we just set a `to`
prop on the `<LinkContainer>` with the path we want and any url params,
and off we go!

I think just take a look in here:

	modified:   static/js/containers/LoginButton.js

pull down the branch, restart docker, visit `/dashboard` and confirm the
js console error is gone.

docs for the lib here:
https://github.com/react-bootstrap/react-router-bootstrap

### DIFF
--- a/static/js/containers/LoginButton.js
+++ b/static/js/containers/LoginButton.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import DropdownButton from 'react-bootstrap/lib/DropdownButton';
+import LinkContainer from 'react-router-bootstrap/lib/LinkContainer';
 import MenuItem from 'react-bootstrap/lib/MenuItem';
-import Link from 'react-router/lib/Link';
 
 class LoginButton extends React.Component {
   render() {
@@ -12,11 +12,11 @@ class LoginButton extends React.Component {
       <DropdownButton
         title={authentication.name}
         id="logout-button">
-        <MenuItem>
-          <Link to="/profile">
+        <LinkContainer to={{ pathname: '/profile' }} active={false}>
+          <MenuItem>
             Profile
-          </Link>
-        </MenuItem>
+          </MenuItem>
+        </LinkContainer>
         <MenuItem
           href="/settings">
           Settings
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "react-mdl": "1.5.3",
     "react-redux": "4.4.0",
     "react-router": "2.0.0",
+    "react-router-bootstrap": "0.22.1",
     "react-select": "1.0.0-beta12",
     "redux-devtools": "2.1.5",
     "redux-logger": "2.6.1",


static/js/containers/LoginButton.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import DropdownButton from 'react-bootstrap/lib/DropdownButton';
+import { LinkContainer } from 'react-router-bootstrap';
```
noisecapella: Can you use this syntax instead?

`import LinkContainer from 'react-router-bootstrap/lib/LinkContainer';`

Importing modules directly is best in case there's other stuff in `react-router-bootstrap` which webpack can leave out

aliceriot: sure

noisecapella:
Great find, I was wondering where this came from!

aliceriot:
@noisecapella I ran into this on a previous project, so I knew where to look!

static/js/containers/LoginButton.js
```diff
@@ -1,6 +1,7 @@
...
 import MenuItem from 'react-bootstrap/lib/MenuItem';
 import Link from 'react-router/lib/Link';
```
noisecapella: Can you remove this line since we don't use `Link` anymore?

aliceriot: what do you think about adding the `no-unused-vars` rule to `eslint.rc`? I think that should make noise about unused imports.

noisecapella: Sure, I kind of assumed this was a default rule but apparently not. Can we do this in a separate PR?

aliceriot: oof nvm it's waaaay loud, it complains about this:

```
`import React from 'react';`

class Foo extends React.Component {}
```

says react isn't used...

noisecapella: `React` is a strange one because every `<div />` gets translated to a `React.createElement("div")` which requires `React`, but you don't see it in the original code. It looks like we can filter out certain errors based on variable name: http://eslint.org/docs/rules/no-unused-vars#caughterrorsignorepattern

aliceriot: ooh nifty!

I'll open another pr and fiddle with this!

noisecapella:
Functionality looks good but I see this blue background now when I'm on the user profile:

screenshot from 2016-04-15 13-48-10

We can deal with it in another PR if you prefer :+1:

aliceriot:
@noisecapella should be fixed! just a quick prop change.

noisecapella:
Looks good!

PR #186, opened 2016-04-15
Added saveProfile action dispatcher, tests for saveProfile and updateProfile

noisecapella:
Part of #149 

#### What's this PR do?
Adds an action to PATCH the profile data

Go to the profile, edit a field. Click 'Save and continue' and it will stay on the page. Refresh the browser and see that the changed value is populated in the field

We should probably add a loader at some point.

Also, there is no validation done yet.

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -6,17 +6,24 @@ export const RECEIVE_COURSE_LIST_SUCCESS = 'RECEIVE_COURSE_LIST_SUCCESS';
 export const RECEIVE_COURSE_LIST_FAILURE = 'RECEIVE_COURSE_LIST_FAILURE';
 export const CLEAR_COURSE_LIST = 'CLEAR_COURSE_LIST';
 
-export const REQUEST_USER_PROFILE = 'REQUEST_USER_PROFILE';
-export const RECEIVE_USER_PROFILE_SUCCESS = 'RECEIVE_USER_PROFILE_SUCCESS';
-export const RECEIVE_USER_PROFILE_FAILURE = 'RECEIVE_USER_PROFILE_FAILURE';
+// user profile actions
+export const REQUEST_GET_USER_PROFILE = 'REQUEST_GET_USER_PROFILE';
+export const RECEIVE_GET_USER_PROFILE_SUCCESS = 'RECEIVE_GET_USER_PROFILE_SUCCESS';
+export const RECEIVE_GET_USER_PROFILE_FAILURE = 'RECEIVE_GET_USER_PROFILE_FAILURE';
 export const CLEAR_PROFILE = 'CLEAR_PROFILE';
 export const UPDATE_PROFILE = 'UPDATE_PROFILE';
+export const START_PROFILE_EDIT = 'START_PROFILE_EDIT';
+export const CLEAR_PROFILE_EDIT = 'CLEAR_PROFILE_EDIT';
+export const REQUEST_PATCH_USER_PROFILE = 'REQUEST_PATCH_USER_PROFILE';
+export const RECEIVE_PATCH_USER_PROFILE_SUCCESS = 'RECEIVE_PATCH_USER_PROFILE_SUCCESS';
+export const RECEIVE_PATCH_USER_PROFILE_FAILURE = 'RECEIVE_PATCH_USER_PROFILE_FAILURE';
 
 // constants for fetch status (these are not action types)
 export const FETCH_FAILURE = 'FETCH_FAILURE';
 export const FETCH_SUCCESS = 'FETCH_SUCCESS';
 export const FETCH_PROCESSING = 'FETCH_PROCESSING';
 
+// actions for course
 const requestCourseList = () => ({ type: REQUEST_COURSE_LIST });
 export const receiveCourseListSuccess = (courseList, programList) => ({
   type: RECEIVE_COURSE_LIST_SUCCESS,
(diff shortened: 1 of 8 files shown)


static/js/actions/index.js
```diff
@@ -6,17 +6,21 @@ export const RECEIVE_COURSE_LIST_SUCCESS = 'RECEIVE_COURSE_LIST_SUCCESS';
 export const RECEIVE_COURSE_LIST_FAILURE = 'RECEIVE_COURSE_LIST_FAILURE';
 export const CLEAR_COURSE_LIST = 'CLEAR_COURSE_LIST';
```
aliceriot: to be consistent, maybe a comment like `// user profile actions` here?

static/js/components/PersonalTab.js
```diff
@@ -25,19 +26,27 @@ class PersonalTab extends React.Component {
   updateGender(gender) {
     const { dispatch, profile: { profileCopy } } = this.props;
```
aliceriot: Am I right in understanding that `profileCopy` is an object we update locally (on the client), and then we only worry about persisting it to the server when we actually save? (`dispatch(saveProfile(...))`)

static/js/components/PersonalTab.js
```diff
@@ -25,19 +26,27 @@ class PersonalTab extends React.Component {
...
+      'preferred_language': language ? language.value : null
     });
     dispatch(updateProfile(clone));
   }
 
+  saveAndContinue() {
+    const { dispatch, profile: { profileCopy } } = this.props;
+
+    dispatch(saveProfile(SETTINGS.username, profileCopy)).then(() => {
+      // TODO: update to next tab
```
aliceriot: are we planning to use `react-router` for the different profile tabs?

so you'd have `/profile/personal`, `/profile/education`, etc

that strikes me as a nice way to organize that, and then transitioning to the next tab is as simple as

```
import { browserHistory } from 'react-router';

browserHistory.push(`/profile/${nextTabUrl}`)
```

static/js/components/PersonalTab.js
noisecapella: Yes that's correct. It could probably use a better name, maybe `profileInEdit` or something like that

aliceriot: yeah, I'd support a name change.

`profileCopy` is a little unclear - took me a bit to figure out why we were passing a copy of the whole thing back into the store, instead of updating the profile itself in the store.

maybe `unsavedProfile`, `profileEditInProgress`, something like that?

static/js/components/PersonalTab.js
```diff
@@ -69,6 +78,9 @@ class PersonalTab extends React.Component {
         placeholder="Preferred language"
       /><br />
 
+      <Button raised onClick={this.saveAndContinue.bind(this)}>
+        Save and continue
+      </Button>
```
aliceriot: could we maybe gray this out until the user has made any edits to their profile?

maybe diff `profile` and `profileCopy`? not sure if there's an easy way to do that.

alternatively, there could just be a `profileHasBeenSaved` boolean that we set to `false` anytime we dispatch `updateProfile`, and we set to `true` anytime `saveProfile` comes back from the server successfully.

that might be too much extra work, so feel free to ignore that suggestion :P

static/js/components/PersonalTab.js
aliceriot: was just thinking about this a little more.

we currently create `profileCopy` on app initialization - not when we actually start editing.

if we rename it to something relating to edits in progress, I think it should only be created when the user actually starts editing stuff (maybe when the first `onChange` callback fires?).

aliceriot: the copy of the profile we're currently editing we might not want to be a straight up copy of the profile anyway, i think it will eventually make sense to store the validation data in there too.

static/js/util/api.js
```diff
@@ -87,7 +87,14 @@ export function getProgramList() {
 }
 
 export function getUserProfile(username) {
-  return fetchJSONWithCSRF('/api/v0/profiles/' + username);
+  return fetchJSONWithCSRF(`/api/v0/profiles/${username}/`);
```
aliceriot: template strings are probably my favorite es6 features :)

aliceriot:
couple of questions, back to you

static/js/components/PersonalTab.js
noisecapella: Sounds good, I'll change it to `unsavedProfile` which will contain:

    {
      errors: [],
      profile: {}
    }

and will be created after dispatching an action

aliceriot: :100:

noisecapella: What if I just call it `edit` instead of `unsavedProfile`? It feels weird to call it `unsavedProfile` if it contains `{ errors, profile }`, and it can be accessed by `edit.profile` or `edit.errors` which seems nice to read

aliceriot: `edit` sounds good to me!

static/js/components/PersonalTab.js
noisecapella: That makes sense. Any thoughts @annagav? I'm going to avoid the work for this PR since there's only one tab

aliceriot: I'm going to be implementing the employment tab, I could take that on at the same time.

static/js/components/PersonalTab.js
noisecapella: I'm not sure `profileHasBeenSaved` feels right here since if they're successful, they're moved onto another tab.

Diffing `profile` and the edited copy is not difficult, it can be done with `_.isEqual`, but for simplicity's sake I'd like to only validate on button click for now

aliceriot: yeah, I think if the editing changes are now in `edit`, which is only created when we start editing, we could gray out the button until the `edit` object has been dispatched to the store - this is a minor UX thing really, that the `save` button visually indicates it won't do anything until you start editing.

noisecapella:
I think that addresses all your comments. Since I'm gone Tuesday would you mind taking over and merging?

aliceriot:
@noisecapella sure!

PR #187, opened 2016-04-15
Updated eslint configuration

aliceriot:

none.

#### What's this PR do?

updates the eslint config to properly lint ES6 / JSX code, sets a few options, and deals with the errors the linter points out.

`.eslintrc`, and then review the changes made to pass linting.

Make sure none of the style changes broke anything!

Originally I just wanted to enable the `no-unused-vars` option, but quickly discovered that we didn't have `eslint-plugin-react` configured quite right, so when enabling `no-unused-vars` eslint would complain that the following code wasn't using `React`, for instance:

```
import React from 'react';

class Foo extends React.Component {
  ...
}
```

using the right configuration base (setting `extends: [ "plugin:react/recommended" ]`) gave us that easily.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,10 @@
 {
   "parser": "babel-eslint",
-  "extends": "defaults/configurations/google",
+  "extends": [
+      "defaults/configurations/google",
+      "eslint:recommended",
+      "plugin:react/recommended"
+  ],
   "rules": {
     "quotes": [0], // no opinion on ' vs "
     "object-curly-spacing": [0],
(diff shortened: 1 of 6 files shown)


.eslintrc
```diff
@@ -10,7 +14,12 @@
     "newline-after-var": [0],
     "react/jsx-indent-props": [2, 2], // no tabs, indent is two spaces
     "react/jsx-key": [2], // validate that key prop exists
-    "react/jsx-no-undef": [2] // disallow undeclared variables in JSX
+    "react/jsx-no-undef": [2], // disallow undeclared variables in JSX
+    "no-unused-vars": [ 1, {
+      "vars": "local",
+      "argsIgnorePattern": "action"
+    }],
+    "comma-dangle": [0]
```
aliceriot: this I'm not sure about - we had several violations in the codebase so I thought it might be something we want turned off.

noisecapella: It's a common pattern in Python, encouraged because lists like this

    [
      'a',
      'b',
      'c',
    ]

Can be updated with one line diffs more easily. I'd vote to silence the warning

aliceriot: yeah, makes sense to me! clean diffs are :100:

noisecapella:
:+1:

PR #189, opened 2016-04-15
Added validation to personal page on profile

noisecapella:
#### What's this PR do?
Adds support for validation of profile fields in the front end (specifically personal tab for now).

Edit the profile and make all fields blank. When you click 'save and continue' you should see red text next to each field which is required (all of them for now).

preferred name is made required here due to a comment in #149

### DIFF
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -10,7 +10,7 @@ import { browserHistory } from 'react-router';
 
 // bind to this.boundTextField in the constructor of a form component
 // to update text fields when editing the profile
-function boundTextField(key, label) {
+function boundTextField(key, label, error) {
   const { updateProfile, profile } = this.props;
   let onChange = e => {
     let clone = Object.assign({}, profile);
@@ -22,6 +22,7 @@ function boundTextField(key, label) {
       floatingLabel
       label={label}
       value={profile[key]}
+      error={error}
       onChange={onChange} />
   );
 }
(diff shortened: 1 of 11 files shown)


noisecapella:
cc @aliceriot 
This is my current thinking on validation, it's a bit out of date with the other PR however

noisecapella:
The UI is a bit weird in certain ways:
 - if there's a validation error on a different tab you don't have any notification of it
 - should the button do something different if there's an error?
 - the 'save and continue' field will save the entire profile, not just that particular tab. May not be terribly important if we're not enforcing required fields on the server side

Any thoughts?

aliceriot:
do you think it would be relatively straightforward to refactor this to set the `error` prop on the inputs? on the `TextField` components it ends up being a much nicer looking error:

mdl_error

(I just set `error="oh no"` in the `boundTextField` method in `static/js/util/profile_edit.js` to take the screenshot)

noisecapella:
Haha, guess I should have read the documentation more closely for that element. I'll make the change

aliceriot:
unfortunately it doesn't look like there's an equally easy way forward with the `react-select` dropdown.

have we considered using the [Material UI](http://www.material-ui.com/) components? they have some really nice select components.

static/js/util/util.js
```diff
@@ -95,3 +95,31 @@ export function makeCourseStatusDisplay(course, now = moment()) {
...
+ */
+export function validateProfile(profile) {
+  let errors = {};
+
+  let required = {
+    'first_name': "First name",
+    'last_name': "Last name",
+    'preferred_name': "Preferred name",
+    'gender': "Gender",
+    'preferred_language': "Preferred language"
```
aliceriot: these don't match the things we display in the UI (i.e. 'First Name' is displayed in the UI as 'Given Name').

might be good to make them match, it's a little incongruous:

name_mismatch

noisecapella:
We've had bad experiences with material-ui in the past... but maybe if we just use the select component it won't be so bad. I'll try it out

static/js/util/util.js
noisecapella: Will do

aliceriot:
Ah, gotcha. I've never used it for anything, I just saw this: http://www.material-ui.com/#/components/select-field and thought it looked fancier than what we're using.

noisecapella:
I'm going to try material-ui after the other work is done so that it doesn't block the demo

aliceriot:
sounds good, I think what we have is good for now

noisecapella:
Comments should be addressed

static/js/util/profile_edit.js
```diff
@@ -22,27 +22,29 @@ function boundTextField(key, label) {
       floatingLabel
       label={label}
       value={profile[key]}
+      error={error}
```
aliceriot: this is nice and clean :)

aliceriot:
I think this LGTM, I like integrating the errors right into the return value of `boundTextField` and `boundSelectField`.

:+1:

PR #191, opened 2016-04-20
Build employment tab

### LINKED ISSUE #150: User enrollment: employment card (part of profile creation)
As a user, I need to enter information about my professional experience. This will help me network with others from my company and industry, and it will help me apply for a masters degree. 

Users can enter as many positions as they choose, but they cannot close the modal with zero positions unless they have also answered "No" to the question "Are you currently employed?"

For each employer, the user must enter:

- Position
- Industry
- Company Name (someday we will normalize these, but not for MVP)
- Company City 
- Company Country (normalized)
- Start Date
- End Date (optional)

Positions without an end date are considered "current"

Self-employed users are encouraged to enter 'self-employed' for position. 

enroll_ professional

aliceriot:

closes #150 

#### What's this PR do?

this adds an 'employment' tab to the profile editing page.

There are a lot of changes. I would start by reviewing the changes to the `bound${inputType}Field` methods in `util/profile_edit` and the changes to the validation logic (in `util/util`).

Pull down the branch, and make sure you can edit all the fields on the employment and personal tabs, and that they are all correctly validated.

dogtongue

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "devDependencies": {
     "babel-eslint": "5.0.0",
     "babel-plugin-transform-react-jsx": "6.6.4",
+    "babel-preset-stage-1": "6.5.0",
     "babel-register": "6.7.2",
     "eslint": "2.2.0",
     "eslint-config-defaults": "9.0.0",
@@ -35,6 +36,7 @@
     "moment": "2.12.0",
     "node-neat": "1.7.2",
     "node-sass": "3.4.2",
+    "object.entries": "1.0.3",
     "react": "0.14.7",
     "react-bootstrap": "0.28.3",
     "react-datepicker": "0.25.0",
--- a/static/js/components/EmploymentTab.js
+++ b/static/js/components/EmploymentTab.js
@@ -1,21 +1,135 @@
 import React from 'react';
+import Button from 'react-mdl/lib/Button';
+import Grid from 'react-mdl/lib/Grid';
+import { Cell } from 'react-mdl/lib/Grid';
 
-class EmploymentTab extends React.Component {
+import ProfileTab from "../util/ProfileTab";
+import { saveAndContinue } from '../util/profile_edit';
+
+class EmploymentTab extends ProfileTab {
+  constructor(props) {
+    super(props);
+    this.saveAndContinue = saveAndContinue.bind(this, '/dashboard', EmploymentTab.validation);
+    this.blankWorkHistoryEntry = {
+      id: null,
+      position: "",
+      industry: "",
+      company_name: "",
+      start_date: "",
+      end_date: "",
+      city: "",
+      country: "",
+      state_or_territory: '',
+    }
+  }
+
+  static propTypes = {
+    profile:        React.PropTypes.object,
+    saveProfile:    React.PropTypes.func,
+    updateProfile:  React.PropTypes.func,
+  };
+
+  static defaultProps = {
+    requiredFields: [
+      ['currently_employed']
+    ],
+    validationMessages: {
+      'currently_employed': 'Current employment status',
+      'position': 'Position',
+      'industry': 'Industry',
+      'company_name': 'Company Name',
+      'start_date': 'Start Date',
+      'end_date': 'End Date',
+      'city': 'City',
+      'country': 'Country',
+      'state_or_territory': 'State or Territory',
+    },
+  };
+
+  static validation (profile, requiredFields) {
+    const keys = [
+      'position',
+      'industry',
+      'company_name',
+      'start_date',
+      'end_date',
+      'city',
+      'country',
+      'state_or_territory',
+    ];
+
+    let nestedFields = (index) => {
+      let keySet = (key) => ['work_history', index, key];
+      return keys.map(key => keySet(key));
+    }
+
+    return requiredFields.concat(
+      ...profile.work_history.map( (v,i) => nestedFields(i))
+    );
+  }
+
+  employmentEntriesForm (index) {
+    let keySet = (key) => ['work_history', index, key];
+    return (
+      <Grid key={index} className="profile-tab-grid">
+        <Cell col={12}>
+          {this.boundTextField(keySet('company_name'), 'Company Name')}
+        </Cell>
+        <Cell col={12}>
+          {this.boundTextField(keySet('position'), 'Position')}
+        </Cell>
+        <Cell col={12}>
+          {this.boundTextField(keySet('industry'), 'Industry')}
+        </Cell>
+        <Cell col={12}>
+          {this.boundTextField(keySet('city'), 'City')}
+        </Cell>
+        <Cell col={12}>
+          {this.boundTextField(keySet('state_or_territory'), 'State or Territory')}
+        </Cell>
+        <Cell col={12}>
+          {this.boundSelectField(keySet('country'), 'Country', this.countryOptions)}
+        </Cell>
+        <Cell col={12}>
+          {this.boundDateField(keySet('start_date'), 'Start Date')}
+        </Cell>
+        <Cell col={12}>
+          {this.boundDateField(keySet('end_date'), 'End Date')}
+        </Cell>
+      </Grid>
+    );
+  }
+
   render () {
     return (
       <div>
-        Employment
-        <br/>
-        id: {this.props.profile.pretty_printed_student_id}
+        <Grid className="profile-tab-grid">
+          <Cell col={6}>
+            Are you currently employed?
+          </Cell>
+          <Cell col={6}>
+            {this.boundSelectField(['currently_employed'], 'Yes/No (required)', this.boolOptions)}
+          </Cell>
+        </Grid>
+        {this.editProfileObjectArray(
+          'work_history',
+          this.blankWorkHistoryEntry,
+          this.employmentEntriesForm.bind(this)
+        )}
+        <Grid className="profile-tab-grid">
+          <Cell className="profile-terms" col={6}>
+            By clicking Save, you agree to the MIT Micromasters Terms of Service and you agree to
+            receiving email from MIT.
+          </Cell>
+          <Cell col={6}>
+            <Button raised onClick={this.saveAndContinue}>
+              Save and continue
+            </Button>
+          </Cell>
+        </Grid>
       </div>
     );
   }
-}
-
-EmploymentTab.propTypes = {
-  profile:    React.PropTypes.object,
-  saveProfile:    React.PropTypes.func,
-  updateProfile:  React.PropTypes.func,
 }
 
 export default EmploymentTab;
--- a/static/js/reducers/index_test.js
+++ b/static/js/reducers/index_test.js
@@ -50,6 +50,7 @@ import configureTestStore from 'redux-asserts';
 import rootReducer, { INITIAL_USER_PROFILE_STATE } from '../reducers';
 import assert from 'assert';
 import sinon from 'sinon';
+import PersonalTab from '../components/PersonalTab';
 
 describe('reducers', () => {
   let sandbox, store, dispatchThen;
@@ -251,11 +252,19 @@ describe('reducers', () => {
       store.dispatch(receiveGetUserProfileSuccess(USER_PROFILE_RESPONSE));
       store.dispatch(startProfileEdit());
 
-      dispatchThen(validateProfile(USER_PROFILE_RESPONSE), [UPDATE_PROFILE_VALIDATION]).then(profileState => {
+      dispatchThen(validateProfile(
+        USER_PROFILE_RESPONSE,
+        PersonalTab.defaultProps.requiredFields,
+        PersonalTab.defaultProps.validationMessages
+      ), [UPDATE_PROFILE_VALIDATION]).then(profileState => {
         assert.deepEqual(profileState.edit.errors, {});
 
         // also assert that it returns a resolved promise
-        store.dispatch(validateProfile(USER_PROFILE_RESPONSE)).then(() => {
+        store.dispatch(validateProfile(
+          USER_PROFILE_RESPONSE,
+          PersonalTab.defaultProps.requiredFields,
+          PersonalTab.defaultProps.validationMessages
+        )).then(() => {
           done();
         });
       });
@@ -265,21 +274,73 @@ describe('reducers', () => {
       let profileWithError = Object.assign({}, USER_PROFILE_RESPONSE, {
         first_name: ''
       });
+      let keysToCheck = [["first_name"]];
+      let messages = {"first_name": "Given name"};
 
       // populate a profile
       store.dispatch(receiveGetUserProfileSuccess(USER_PROFILE_RESPONSE));
       store.dispatch(startProfileEdit());
-      dispatchThen(validateProfile(profileWithError), [UPDATE_PROFILE_VALIDATION]).then(profileState => {
+      dispatchThen(validateProfile(
+        profileWithError,
+        keysToCheck,
+        messages
+      ), [UPDATE_PROFILE_VALIDATION]).then(profileState => {
         assert.deepEqual(profileState.edit.errors, {
           first_name: 'Given name is required'
         });
 
         // also assert that it returns a rejected promise
-        store.dispatch(validateProfile(profileWithError)).catch(() => {
+        store.dispatch(validateProfile(
+          profileWithError,
+          keysToCheck,
+          messages
+        )).catch(() => {
           done();
         });
       });
     });
+
+    it('should validate a profile with nested objects and errors', done => {
+      let profileWithError = Object.assign({}, USER_PROFILE_RESPONSE, {
+        work_history: [{
+          company_name: "foocorp",
+          position: undefined
+        }]
+      });
+      let keysToCheck = [
+        ["work_history", 0, "company_name"],
+        ["work_history", 0, "position"],
+      ];
+      let messages = {
+        "company_name": "Company name",
+        "position": "Position",
+      };
+
+      // populate a profile
+      store.dispatch(receiveGetUserProfileSuccess(USER_PROFILE_RESPONSE));
+      store.dispatch(startProfileEdit());
+      dispatchThen(validateProfile(
+        profileWithError,
+        keysToCheck,
+        messages
+      ), [UPDATE_PROFILE_VALIDATION]).then(profileState => {
+        assert.deepEqual(profileState.edit.errors, {
+          work_history: [
+            {position: 'Position is required'}
+          ]
+        });
+
+        // also assert that it returns a rejected promise
+        store.dispatch(validateProfile(
+          profileWithError,
+          keysToCheck,
+          messages
+        )).catch(() => {
+          done();
+        });
+      });
+    });
+
 
     it("can't edit a profile if we never get it successfully", done => {
       dispatchThen(startProfileEdit(), [START_PROFILE_EDIT]).then(profileState => {
--- a/static/js/util/ProfileTab.js
+++ b/static/js/util/ProfileTab.js
@@ -4,20 +4,44 @@ import {
   boundDateField,
   boundTextField,
   boundSelectField,
+  editProfileObjectArray,
 } from './profile_edit';
 import COUNTRIES from '../countries';
+import LANGUAGE_CODES from '../language_codes';
 
 class ProfileTab extends React.Component {
   constructor(props) {
     super(props);
+
+    // bind our field methods to this
     this.boundTextField = boundTextField.bind(this);
     this.boundSelectField = boundSelectField.bind(this);
     this.boundDateField = boundDateField.bind(this);
+    this.editProfileObjectArray = editProfileObjectArray.bind(this);
+
+    // options we set (for select components)
     this.countryOptions = COUNTRIES.map(country => ({
       value: country.Code,
       label: country.Name
     }));
+    this.boolOptions = [
+      { value: 'yes', label: 'Yes' },
+      { value: 'no', label: 'No' }
+    ];
+    this.genderOptions = [
+      { value: 'm', label: 'Male' },
+      { value: 'f', label: 'Female' },
+      { value: 'o', label: 'Other/Prefer not to say' }
+    ];
+    this.languageOptions = LANGUAGE_CODES.map(language => ({
+      value: language.alpha2,
+      label: language.English
+    }));
   }
+
+  static contextTypes = {
+    router: React.PropTypes.object.isRequired
+  };
 }
 
 export default ProfileTab;
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -2,8 +2,10 @@ import React from 'react';
 import TextField from 'react-mdl/lib/Textfield';
 import Select from 'react-select';
 import DatePicker from 'react-datepicker';
-import { browserHistory } from 'react-router';
+import Button from 'react-mdl/lib/Button';
+import Grid, { Cell } from 'react-mdl/lib/Grid';
 import moment from 'moment';
+import _ from 'lodash';
 
 // utility functions for pushing changes to profile forms back to the
 // redux store.
@@ -12,19 +14,26 @@ import moment from 'moment';
 
 // bind to this.boundTextField in the constructor of a form component
 // to update text fields when editing the profile
-export function boundTextField(key, label, error) {
-  const { updateProfile, profile } = this.props;
+// we pass in a keyset looking like this:
+//
+// ["top-level-key", index, "nested_object_key"] or just ["top_level_key"]
+export function boundTextField(keySet, label) {
+  const {
+    profile,
+    errors,
+    updateProfile
+  } = this.props;
   let onChange = e => {
-    let clone = Object.assign({}, profile);
-    clone[key] = e.target.value;
+    let clone = _.cloneDeep(profile);
+    _.set(clone, keySet, e.target.value);
     updateProfile(clone);
   };
   return (
     <TextField
       floatingLabel
       label={label}
-      value={profile[key]}
-      error={error}
+      value={_.get(profile, keySet)}
+      error={_.get(errors, keySet)}
       onChange={onChange} />
   );
 }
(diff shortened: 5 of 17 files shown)


profiles/models.py
```diff
@@ -123,6 +130,13 @@ class Profile(models.Model):
     date_joined_micromasters = models.DateTimeField(blank=True, null=True, auto_now_add=True)
     linkedin = JSONField(blank=True, null=True)
     student_id = models.IntegerField(blank=True, null=True, unique=True)
+    currently_employed = models.CharField(
```
noisecapella: Why not use a boolean here? Are we expecting other options in the future?

noisecapella:
Instead of creating the extra bound field functions, can we change the existing bound field functions to take a selector function for the object in addition to the string key? For the existing functions this function would be `profile => profile` but for the other functions it may be `profile => profile['employment'][index]`

pdpinch:
We've got some design changes to this form. Please talk to me or @Ferdi before reviewing or updating this PR.

profiles/models.py
aliceriot: that...does make more sense, haha.

aliceriot:
@noisecapella I think that's the right way to go about it. I'm going to try cooking up something like this:

```
function boundTextField(keySet, label) { ... }
```

where keySet is an array of keys and indices into `profile`. in the case of a top-level property (e.g. `first_name`) our call would look like:

```
<div>
  {this.boundTextField(["first_name"], "First Name")}
</div>
```

We can re-used the `keySet` both to get the current value for `first_name` and to get any validation errors. A more complicated example (the `company_name` property on the first entry in `profile.work_history`) would look like this:

```
<div>
  {this.boundTextField(["work_history"][0]["company_name"], "Company Name")}
</div>
```

(in this case we'd have to insert the index programmatically, by mapping across the objects in `profile.work_history` and calling some function that draws the right UI for each of them).

does that generally make sense? I wrote a general purpose function ( in `util/util.js` ) called `dig` which should be helpful here - it's basically a JS implementation of `Hash.dig` in Ruby.

noisecapella:
I think something like this is cleaner personally:

    {this.boundTextField(profile => profile["work_history"][0], "company_name", "Company Name")}

But feel free to keep it as is if you prefer, this is just a matter of taste

If we do write the `dig` and `insert` functions they should be unit tested well. Is there some third party library we can pull these functions from?

noisecapella:
By the way I'm going to leave some comments on this PR but I'm aware it's not yet marked Needs Review. Feel free to ignore if you intend to get to it in the future!

aliceriot:
haha will do!

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+}
 
-  for (let key of Object.keys(required)) {
-    if (!profile[key]) {
-      errors[key] = `${required[key]} is required`;
+export function dig (object, keys) {
+  let obj = _.cloneDeep(object);
+  let keySet = _.cloneDeep(keys);
+  while (keySet.length > 0) {
+    let key = keySet.shift();
```
noisecapella: If you iterate over the keys using the `for (let key of keySet) ...` syntax you won't need to clone either `obj` or `keySet` here

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+    if (obj[key]) {
```
noisecapella: If `obj[key]` is `0` this will return `undefined`.

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+  return obj;
+}
 
-  return errors;
+export function insert (object, keys, payload) {
+  let keySet = _.cloneDeep(keys);
+  let temp = object;
+  while (keySet.length > 1) {
+    let key = keySet.shift()
+    if ( temp[key] ) {
```
noisecapella: Same comment here as above, we should use `hasOwnProperty` here to see if the object has the key

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+      return undefined;
     }
   }
+  return obj;
+}
 
-  return errors;
+export function insert (object, keys, payload) {
+  let keySet = _.cloneDeep(keys);
+  let temp = object;
```
noisecapella: Since most of the objects we deal with here will be immutable, can we change this to deep clone the object and return the modified clone instead? It looks like existing callers of the function don't do anything with the return value.

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+    } else {
+      return undefined;
     }
   }
+  return obj;
+}
 
-  return errors;
+export function insert (object, keys, payload) {
+  let keySet = _.cloneDeep(keys);
```
noisecapella: Can you change the while loop below to a for loop? You can do a for loop over the index if you need to skip the first item. That way we won't need to clone the keys here.

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
-    'gender': "Gender",
-    'preferred_language': "Preferred language",
-    'city': "City",
-    'country': "Country",
-    'birth_city': 'City',
-    'birth_country': "Country",
-    'date_of_birth': "Date of birth"
-  };
+  for (let keySet of requiredFields) {
+    if (!dig(profile, keySet)) {
```
noisecapella: `dig` may return a valid `0` so this should handle that case

static/js/util/util.js
noisecapella: Also, `shift()` can cause subtle performance problems since you are copying the remainder of the array each time you use it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+  let nullObj = (obj) => {
+    for (let key in obj) { obj[key] = null };
+  }
+  fields.map( (pair) => {
+    for (let obj of pair[1]) {
+      nullObj(obj);
+    }
+  });
   let errors = {};
+  fields.map( pair => errors[pair[0]] = pair[1]);
```
noisecapella: Can you use `forEach` instead of `map`? There's not much difference but `map` usually indicates you wanted to have no side effects and return the changed value, while `forEach` is explicitly for causing side effects of some kind (like populating the `errors` object above)

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
  * @param {Object} profile The user profile
  * @returns {Object} Validation errors or an empty object if no errors
  */
-export function validateProfile(profile) {
+export function validateProfile(profile, requiredFields, messages) {
+  let clone = _.cloneDeep(profile);
```
noisecapella: It would be nice to avoid cloning the profile here since we are only generating a list of errors, not modifying the profile. Is there a way around it?

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
+  let fields = Object.entries(clone).filter(pair => _.isObject(pair[1]));
+  let nullObj = (obj) => {
+    for (let key in obj) { obj[key] = null };
+  }
+  fields.map( (pair) => {
```
noisecapella: You can use ES6 destructuring here: `([key, value])` for example

static/js/util/util.js
```diff
@@ -168,30 +169,58 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
  * @param {Object} profile The user profile
  * @returns {Object} Validation errors or an empty object if no errors
  */
-export function validateProfile(profile) {
+export function validateProfile(profile, requiredFields, messages) {
+  let clone = _.cloneDeep(profile);
+  let fields = Object.entries(clone).filter(pair => _.isObject(pair[1]));
+  let nullObj = (obj) => {
+    for (let key in obj) { obj[key] = null };
```
noisecapella: Can you change this to `for (let key of Object.keys(obj))` instead? It's slightly more explicit and it guards against possible prototype properties (which are admittedly not a problem for us)

static/js/util/profile_edit.js
```diff
@@ -81,3 +104,44 @@ export function saveAndContinue(next) {
     browserHistory.push(next)
   });
 }
+
+export function editProfileObjectArray (arrayName, blankEntry, formCallback) {
+  const { updateProfile, profile } = this.props;
+  if ( profile[arrayName] ) {
+    let editForms = profile[arrayName].map( (entry, index) => (
+      formCallback.bind(this)(entry, index)
```
noisecapella: Instead of `bind(this)(entry, index)` you can use `call(this, entry, index)`. However, why not `bind(this)` in the calling function instead? That would allow any function that returns a `ReactElement` to be used here

static/js/util/profile_edit.js
```diff
@@ -81,3 +104,44 @@ export function saveAndContinue(next) {
...
+    ));
+    let addAnotherBlankEntry = () => {
+      let clone = Object.assign({}, profile);
+      clone[arrayName].push(Object.assign({}, blankEntry));
```
noisecapella: Since `Object.assign` only does a shallow copy, can you do this instead:

    clone[arrayName] = clone[arrayName].concat(blankEntry);

Since `blankEntry` isn't changed here you shouldn't need to clone it

static/js/util/profile_edit.js
```diff
@@ -81,3 +104,44 @@ export function saveAndContinue(next) {
     browserHistory.push(next)
   });
 }
+
+export function editProfileObjectArray (arrayName, blankEntry, formCallback) {
+  const { updateProfile, profile } = this.props;
+  if ( profile[arrayName] ) {
```
noisecapella: Can you update this to use `hasOwnProperty` to handle `0` and `false` correctly?

static/js/util/profile_edit.js
noisecapella: If you make that change to `formCallback` you should be able to simplify this to

    let editForms = profile[arrayName].map(formCallback);

static/js/util/profile_edit.js
```diff
@@ -81,3 +104,44 @@ export function saveAndContinue(next) {
...
+          </Button>
+        </Cell>
+      </Grid>
+    );
+    return editForms;
+  } else {
+    let startEditing = () => {
+      let clone = Object.assign({}, profile);
+      if ( clone[arrayName] === null ) {
+        clone[arrayName] = [ Object.assign({}, blankEntry) ];
```
noisecapella: You shouldn't need to clone `blankEntry` here. Since all redux code treats these objects as immutable, we can just pass around the reference to `blankEntry` instead of making a defensive copy of it

static/js/util/profile_edit.js
```diff
@@ -81,3 +104,44 @@ export function saveAndContinue(next) {
...
+            Add another entry
+          </Button>
+        </Cell>
+      </Grid>
+    );
+    return editForms;
+  } else {
+    let startEditing = () => {
+      let clone = Object.assign({}, profile);
+      if ( clone[arrayName] === null ) {
```
noisecapella: This will only check for `null` specifically. Besides, if the intent is to check if it has anything in here already, that check was already done in the outside if statement so you can skip the check here.

static/js/util/profile_edit.js
aliceriot: ah nice

profiles/models.py
aliceriot: these are temporary anyway, and just in this pr so that I have something to fiddle with in the frontend

static/js/util/profile_edit.js
aliceriot: I actually ran into issues with edits being mirrored across objects if I took out `Object.assign` here, so I put it back in. Not exactly sure what that's happening, because looking at the reducer it looks like it shouldn't, so ¯\_(ツ)_/¯

static/js/util/profile_edit.js
aliceriot: I changed this like to:

```js
profile.hasOwnProperty(arrayName) && !_.isEmpty(profile[arrayName])
```

static/js/util/profile_edit.js
noisecapella: I think the issue is that `insert` currently mutates the object passed in, which might include `blankEntry`

static/js/util/profile_edit.js
aliceriot: I updated this to

```js
profile.hasOwnProperty(arrayName) && !_.isEmpty(profile[arrayName])
```

static/js/util/profile_edit.js
aliceriot: I'm now binding the form callback in the `render` function on the tab component, so we just do

```js
let editForms = profile[arrayName].map( (v, i) => formCallback(i) );
```

static/js/util/util.js
aliceriot: lodash includes `_.get`, which I'm using instead now. It's a little bit better behaved.

static/js/util/profile_edit.js
aliceriot: I'm no longer using insert, instead I'm using `_.set`, which does indeed mutate the object in place.

noisecapella:
If I save the employment page without adding any employment information I get this console error: TypeError: profile.work_history is null

static/js/components/EmploymentTab.js
```diff
@@ -1,21 +1,131 @@
...
+        <Cell col={12}>
+          {this.boundDateField(keySet('end_date'), 'End Date')}
+        </Cell>
+      </Grid>
+    );
+  }
 
-class EmploymentTab extends React.Component {
   render () {
+    this.employmentEntriesForm = this.employmentEntriesForm.bind(this);
```
noisecapella: Can you create a local variable here instead of reassigning since this variable is only used inside the render method?

static/js/util/profile_edit.js
```diff
@@ -2,8 +2,12 @@ import React from 'react';
 import TextField from 'react-mdl/lib/Textfield';
 import Select from 'react-select';
 import DatePicker from 'react-datepicker';
+import Button from 'react-mdl/lib/Button';
+import { Cell } from 'react-mdl/lib/Grid';
+import Grid from 'react-mdl/lib/Grid';
```
noisecapella: Can you combine the `Grid` import with the previous line?

static/js/util/profile_edit.js
```diff
@@ -12,72 +16,141 @@ import moment from 'moment';
...
+//
+// ["top-level-key", index, "nested_object_key"] or just ["top_level_key"]
+export function boundTextField(keySet, label) {
+  const {
+    profile,
+    errors,
+    updateProfile
+  } = this.props;
   let onChange = e => {
     let clone = Object.assign({}, profile);
```
noisecapella: This will need to be a deep clone since `_.set` can alter data at any level within the object

static/js/util/profile_edit.js
```diff
@@ -12,72 +16,141 @@ import moment from 'moment';
...
-  const { updateProfile, profile } = this.props;
+export function boundSelectField(keySet, label, options) {
+  const {
+    profile,
+    errors,
+    validationMessages,
+    updateProfile,
+  } = this.props;
   let onChange = value => {
     let clone = Object.assign({}, profile);
```
noisecapella: This will need to be a deep clone

static/js/util/profile_edit.js
```diff
@@ -12,72 +16,141 @@ import moment from 'moment';
...
+export function boundDateField(keySet, label) {
+  const {
+    profile,
+    errors,
+    validationMessages,
+    updateProfile,
+  } = this.props;
+
   let onChange = date => {
     let clone = Object.assign({}, profile);
```
noisecapella: This will need to be a deep clone

static/js/util/profile_edit.js
aliceriot: ah, that makes sense

static/js/components/EmploymentTab.js
aliceriot: I'm not sure what you mean, are you suggesting changing the above to something like this?

```js
render () {
  const employmentForm = this.employmentEntriesForm.bind(this)
```

noisecapella: Yes, it's bad practice to have `render()` mutate its own object in any way

static/js/util/profile_edit.js
```diff
@@ -12,72 +15,141 @@ import moment from 'moment';
...
+    _.set(clone, keySet, e.target.value);
     updateProfile(clone);
   };
   return (
     <TextField
       floatingLabel
       label={label}
-      value={profile[key]}
-      error={error}
+      value={_.get(profile, keySet, undefined)}
```
noisecapella: Can you remove `undefined` here and below since it's the default argument value?

static/js/util/profile_edit.js
```diff
@@ -12,72 +15,141 @@ import moment from 'moment';
...
+  } else {
+    fields = requiredFields;
+  }
+
+  saveProfile( profile, fields, validationMessages).then(() => {
     browserHistory.push(next)
   });
 }
+
+export function editProfileObjectArray (arrayName, blankEntry, formCallback) {
```
noisecapella: Can you add a description for this function?

static/js/util/profile_edit.js
```diff
@@ -12,72 +15,141 @@ import moment from 'moment';
...
+  const { updateProfile, profile } = this.props;
+  if ( profile.hasOwnProperty(arrayName) && !_.isEmpty(profile[arrayName]) ) {
+    let editForms = profile[arrayName].map((v, i) => formCallback(i));
+    let addAnotherBlankEntry = () => {
+      let clone = Object.assign({}, profile);
+      clone[arrayName] = clone[arrayName].concat(Object.assign({}, blankEntry));
```
noisecapella: After you add the deep clones above can you check if the `Object.assign({}, blankEntry)` is still necessary?

static/js/util/util.js
```diff
@@ -168,30 +169,16 @@ export function makeCourseProgressDisplay(course, isFirst, isLast) {
...
-    'birth_city': 'City',
-    'birth_country': "Country",
-    'date_of_birth': "Date of birth"
-  };
-
-  for (let key of Object.keys(required)) {
-    if (!profile[key]) {
-      errors[key] = `${required[key]} is required`;
+  for (let keySet of requiredFields) {
+    if ( _.isEmpty(_.get(profile, keySet, undefined)) ) {
```
noisecapella: `_.isEmpty` is true for all numbers and certain other non-collection types, for example `_.isEmpty(1)` is true.

static/js/util/profile_edit.js
aliceriot: haha, somehow I decided that it returned something else by default :stuck_out_tongue_closed_eyes:

static/js/util/util_test.js
```diff
@@ -211,35 +211,61 @@ describe('utility functions', () => {
...
+      };
+      const keysToCheck = [
+        ["nested_array", 0, "foo"],
+        ["nested_array", 0, "baz"],
+      ];
+      const nestMessages = {"baz": "Baz"};
+      let errors = validateProfile(profile, keysToCheck, nestMessages);
+      assert.deepEqual({nested_array: [ { baz: "Baz is required" } ] }, errors);
+    });
   });
```
noisecapella: Can you add a test to validate that a field can have a value of `0` and pass the required validation?

package.json
```diff
@@ -35,6 +35,7 @@
     "moment": "2.12.0",
     "node-neat": "1.7.2",
     "node-sass": "3.4.2",
+    "object.entries": "^1.0.3",
```
noisecapella: Can you change this to an exact version?

static/js/util/util.js
aliceriot: we aren't currently dealing with any numerical inputs - should I make everything numerical-safe anyway?

noisecapella: Yes, it would save us the effort if we have numerical input in the future. We should also support boolean values which also require similar care

static/js/util/profile_edit.js
aliceriot: just tried it out, seems like it's working just fine with those `Object.assign(...)` calls taken out.

noisecapella:
Can you remove the python changes from this PR since we're breaking that into #243?

static/js/components/EmploymentTab.js
```diff
@@ -1,21 +1,131 @@
...
+    this.saveAndContinue = saveAndContinue.bind(this, '/dashboard', this.validation);
+    this.blankWorkHistoryEntry = {
+      id: null,
+      position: "",
+      industry: "",
+      company_name: "",
+      start_date: "",
+      end_date: "",
+      city: "",
+      country: "",
```
noisecapella: Can you add `state_or_territory` here and below?

static/js/util/ProfileTab.js
```diff
@@ -4,19 +4,39 @@ import {
...
     this.boundDateField = boundDateField.bind(this);
+    this.editProfileObjectArray = editProfileObjectArray.bind(this);
+
+    // options we set (for select components)
     this.countryOptions = COUNTRIES.map(country => ({
       value: country.Code,
       label: country.Name
     }));
+    this.boolOptions = [
+      { value: 'yes', label: 'Yes' },
```
noisecapella: For the `value` piece can you use `true` and `false`?

aliceriot:
@noisecapella I'm going to remove the python changes when it's ready to merge, I need them until then to keep testing things

static/js/components/EmploymentTab.js
aliceriot: sure!

aliceriot: so we don't have `state_or_territory` on the mockup ([here](https://github.com/mitodl/micromasters/issues/150), unless there's a more recent version I haven't seen) is this something we're adding?

noisecapella: I feel like it's an oversight that it's not in the mockups. I'll see if I can grab @pdpinch to be sure

static/js/util/profile_edit.js
```diff
@@ -12,72 +15,147 @@ import moment from 'moment';
 
 // bind to this.boundTextField in the constructor of a form component
 // to update text fields when editing the profile
-export function boundTextField(key, label, error) {
-  const { updateProfile, profile } = this.props;
+// we pass in a keyset looking like this:
+//
+// ["top-level-key", index, "nested_object_key"] or just ["top_level_key"]
+export function boundTextField(keySet, label) {
```
noisecapella: It would be useful to have unit tests for these functions. Would it be helpful if I worked on that?

static/js/util/profile_edit.js
```diff
@@ -12,72 +15,147 @@ import moment from 'moment';
...
-  saveProfile(profile).then(() => {
+  let fields;
+  if ( _.isFunction(nestedValidationCallback) ) {
+    fields = nestedValidationCallback(profile, requiredFields);
+  } else {
+    fields = requiredFields;
+  }
+
+  saveProfile( profile, fields, validationMessages).then(() => {
     browserHistory.push(next)
```
noisecapella: Can you use `context.router.push` here?

aliceriot: ah yes good idea :)

static/js/util/profile_edit.js
aliceriot: I could do that!

static/js/components/EmploymentTab.js
aliceriot: cool, I just addded it, can always revert that commit if we don't want it

static/js/util/profile_edit.js
noisecapella: Sounds good, let me know if you need any help

aliceriot: just added some!

static/js/util/profile_edit_test.js
```diff
@@ -0,0 +1,112 @@
...
+  const change = (clone) => that.props.profile = clone;
+  beforeEach(() => {
+    that = {
+      props: {
+        profile: {
+          "first_name": "",
+          "date_of_birth": "",
+          "gender": undefined,
+        },
+        errors: {},
```
noisecapella: Can you add an error here (or below) so that we can assert it in the return value?

aliceriot: sure!

static/js/util/ProfileTab.js
```diff
@@ -4,20 +4,44 @@ import {
...
     this.boundDateField = boundDateField.bind(this);
+    this.editProfileObjectArray = editProfileObjectArray.bind(this);
+
+    // options we set (for select components)
     this.countryOptions = COUNTRIES.map(country => ({
       value: country.Code,
       label: country.Name
     }));
+    this.boolOptions = [
+      { value: true, label: 'Yes' },
```
noisecapella: It looks like `true` and `false` are not actually working :open_mouth:  https://github.com/JedWatson/react-select/issues/546
Can you change it back?

aliceriot:
last commit may be controversial - I added the `stage-1` babel present, which lets us use class properties inside of the class declaration, like this:

```js
class Foo extends React.Component {
  static myProp = "hey!";
}
```

it's a nice change, I think, because then the class properties are inside the class declaration, as opposed to this situation:

```js
class Foo extends React.Component {
  ...
}

Foo.myProp = "hey!";
```

which is a lot harder to read IMO, and it's easy to see the closing brace for the class and assume you've read all that's relevant.

static/js/util/ProfileTab.js
aliceriot: oh noes! ya I'll change it back

static/js/util/profile_edit_test.js
aliceriot: I'll add an assert in each function's test case too

noisecapella:
Is there the ability to delete educations in the UI mockups?

aliceriot:
looking at this: https://github.com/mitodl/micromasters/issues/150 I don't see an 'X' button or anything - should be that hard to add one though (famous last words)

noisecapella:
I'm fine just leaving it as is, we can add that feature in a future PR

noisecapella:
It looks like the red we use and the red used in material-design-lite is slightly different. Can you update our CSS to #DA0000?

Also found 'undefined is required' :)

screenshot from 2016-05-02 15-32-43

noisecapella:
It doesn't look like there's any way to clear the datefield, and the existing UI doesn't have a nice way of saying that a job doesn't have an end date (ie, your current job). But I think we should get something merged sooner and deal with those issues separately

noisecapella:
Can you squash, remove the Python code and otherwise get this into a state ready for merge? I'm happy with the code for the most part, and it would be nice to have a working UI in master to iterate on so we can deal with the quirks

aliceriot:
cool, sounds good to me!

aliceriot:
aha forgot to add a validation message when I added `state_or_territory`

aliceriot:
ok I just squashed it down

static/js/reducers/index_test.js
```diff
@@ -265,17 +274,27 @@ describe('reducers', () => {
...
         // also assert that it returns a rejected promise
-        store.dispatch(validateProfile(profileWithError)).catch(() => {
+        store.dispatch(validateProfile(
+          profileWithError,
+          keysToCheck,
+          messages
+        )).catch(() => {
           done();
         });
       });
```
noisecapella: Can you add a test to cover validation for an employment tab field? Specifically I want to make sure keys with more than one item are covered somewhere

aliceriot: ah yeah, that's a good idea, I meant to put one it and it slipped my mind.

static/js/reducers/index_test.js
```diff
@@ -300,6 +300,49 @@ describe('reducers', () => {
...
+
+      // populate a profile
+      store.dispatch(receiveGetUserProfileSuccess(USER_PROFILE_RESPONSE));
+      store.dispatch(startProfileEdit());
+      dispatchThen(validateProfile(
+        profileWithError,
+        keysToCheck,
+        messages
+      ), [UPDATE_PROFILE_VALIDATION]).then(profileState => {
+        console.log(profileState.edit.errors);
```
noisecapella: Stray console.log

aliceriot: oop good eye :)

noisecapella:
I got to go but I'll do a quick once over tomorrow morning to file issues for the little things that have come up. But it looks like it's in great shape!

aliceriot:
@noisecapella does that mean this is good to merge?

aliceriot:
@noisecapella nvm I gotta head out too, we can take care of it tomorrow morning.

noisecapella:
Ok let's merge :+1:

aliceriot:
weeee

Issue #97, opened 2016-03-24
Home Page Theming


pdpinch:
Make the home page look like ~~https://projects.invisionapp.com/share/KR6FYCZTX#/screens/142338433~~ https://projects.invisionapp.com/share/AF6XG4X62#/screens/152781225_Public_Home_V3-1

- [x] Depends on #95 logo graphic
- [x] Depends on #103 thumbnails and descriptions in Programs model
- [ ] other images available from the launch site repo, https://github.com/mitodl/mga/tree/gh-pages/images (private)

pdpinch:
we'll ask Nicholae to do this

pdpinch:
![bg-banner-1](https://cloud.githubusercontent.com/assets/430126/14188695/ec37ca58-f757-11e5-95a8-6af1f6ab40a8.png)
![bg-banner-1 2x](https://cloud.githubusercontent.com/assets/430126/14188700/ec44a3ea-f757-11e5-98d2-21f147827439.png)
![bg-main-1](https://cloud.githubusercontent.com/assets/430126/14188698/ec428cf4-f757-11e5-81f7-942d307ea2bb.png)
![bg-main-1 2x](https://cloud.githubusercontent.com/assets/430126/14188699/ec42a8f6-f757-11e5-81c0-c70556a7245b.png)
![logo-mit-bluewhite](https://cloud.githubusercontent.com/assets/430126/14188696/ec3ff76e-f757-11e5-9647-4eea5f43591d.png)
![logo-mit-bluewhite 2x](https://cloud.githubusercontent.com/assets/430126/14188697/ec4033dc-f757-11e5-9bb2-73e5e3edeb04.png)

pdpinch:
@aliceriot there's a new home page design from @nicolaeherrera at https://github.com/mitodl/micromasters-templates/commit/e293184d53111161f247dbda078491bfd1c55cc4

pdpinch:
There's also an updated invision page too: https://projects.invisionapp.com/share/AF6XG4X62#/screens/152781225_Public_Home_V3-1

PR #198, opened 2016-04-21
incorporate homepage template

aliceriot:

this accomplishes part of #97 

#### What's this PR do?

implements a temporary version of the homepage

the stuff going on in `cms/templates/cms/home_page.html` and `/static/scss`.

Pull down the branch, check it out, click around and make sure I didn't clobber the CSS for any other part of the side.

mm_homepage


giocalitri:
it is missing the actual programs in the template

cms/templates/cms/home_page.html
```diff
@@ -9,54 +9,156 @@
...
+          MIT credentials are now attainable by anyone in the world with the discipline, dedication, and intelligence to master course material.
+        </h1>
+        <a href="add" role="button" class="btn btn-primary btn-lg btn-learn-more">
+          <span>LEARN MORE</span>
+        </a>
+      </div>
+      <div class="banner-bottom">
+        <div class="banner-wrapper-content">
+          <h2>Get Started Today!
+            <a href="add" role="button" class="btn btn-success sign-in-button margin-left-10">
```
giocalitri: the `href` for this button is wrong: but I am not sure what we should do in case the user is already signed in. 
For the moment let's just show it also when the user is logged in.

giocalitri:
there are a couple of small problems, but I do not think it's worth fixing them

cms/templates/cms/home_page.html
aliceriot: cool, it's not that complicated to only show the button when the user isn't signed in, or we could change it to 'visit the dashboard' or something similar when they are.

giocalitri: Not it is better to show it always and wait for the new UI to do it right

giocalitri: (IMHO)

aliceriot: cool, sounds good

aliceriot:
updated: 

mm_homepage_updated

I made a few css tweaks as well.

Is it correct that programs do not yet have descriptions and images that are stored in the database? I looked at the `Program` model in `courses/models.py` and didn't see anything to substitute for the stock image and the lorum ipsum.

aliceriot:
updated again: 

mm_homepage_title_descriptions

giocalitri:
Can you please change the default "homepage title" to the text "MIT credentials are now attainable by anyone in the world with the discipline, dedication, and intelligence to master course material."
In this way people should not change it manually later.

giocalitri:
Another problem I see is when there is no description, there is a "None" text. There should be something more meaningful like "No description available" o simply no text at all (whatever you prefer).

cms/templates/cms/home_page.html
```diff
@@ -9,54 +9,115 @@
...
-            {% if program.programpage %}
-            <a href="{{ program.programpage.url }}">
-              <div class="program-name">{{ program }}</div>
+
+    <div class="banner-wrapper">
+      <div class="banner-wrapper-content">
+        <h1>
+          {{ title }}
+        </h1>
+        <a href="add" role="button" class="btn btn-primary btn-lg btn-learn-more">
```
giocalitri: Not sure where this `href` should point, but definitively not to this url.

cms/templates/cms/home_page.html
```diff
@@ -9,54 +9,115 @@
...
+                    <div class="title">{{ program }}</div>
+                  </a>
+                  {% else %}
+                  <div class="title">{{ program }}</div>
+                  {% endif %}
+                  <div class="description">
+                    {{ program.description }}
+                  </div>
+                  <div class="media-footer">
+                    <div class="widget-metas"><span class="meta-courses">{{ program.course_set.count }} courses</span></div>
```
giocalitri: edge case here:
you need to take in account that there might me a single course and in this case `courses` should be `course`

aliceriot:
@giocalitri for changing the default, do you mean by setting a default value for the column in the database?

cms/templates/cms/home_page.html
aliceriot: not sure either, this was set to `add` in the template I was incorporating and I forgot to reset it.

we could make it `#` for now?

giocalitri:
@aliceriot yes, I guess we can make a migration for it. Let me know if you have issues finding how to do it

aliceriot:
updated again!

mm_updated_again

cms/migrations/0006_auto_20160421_1712.py
```diff
@@ -0,0 +1,23 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def set_default_homepage_text(apps, schema_editor):
+    HomePage = apps.get_model("cms", "HomePage")
+    homepage = HomePage.objects.first()
+    homepage.title = """MIT credentials are now attainable by \
+anyone in the world with the discipline, dedication, and \
+intelligence to master course material."""
```
giocalitri: You should not use the triple quotes for this: they are used for docstrings.
You can use string literal concatenation for this, so something like:
```
("MIT credentials are now attainable by"
 "anyone in the world with the discipline, dedication, and"
 "intelligence to master course material."
```
Note that there are no commas

cms/templates/cms/home_page.html
```diff
@@ -80,7 +80,11 @@
                   <div class="title">{{ program }}</div>
                   {% endif %}
                   <div class="description">
+                    {% if program.description %}
                     {{ program.description }}
+                    {% else %}
+                    No description available for this course.
```
giocalitri: I guess you need to say `program` and not `course` here

cms/templates/cms/home_page.html
```diff
@@ -9,54 +9,123 @@
...
+                    <div class="title">{{ program }}</div>
+                  </a>
+                  {% else %}
+                  <div class="title">{{ program }}</div>
+                  {% endif %}
+                  <div class="description">
+                    {% if program.description %}
+                    {{ program.description }}
+                    {% else %}
+                    No description available for this course.
```
giocalitri: this should be `program` and not `course`

cms/migrations/0006_auto_20160421_1712.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.4 on 2016-04-21 17:12
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def set_default_homepage_text(apps, schema_editor):
```
noisecapella: Can you rename this migration to something human readable, ie `0006_change_homepage_title.py`?

giocalitri:
I think now it is fine: please squash 👍

aliceriot:
awesome, thanks for reviewing!

PR #199, opened 2016-04-21
Deleted the `employment` route placeholder

aliceriot:

none

#### What's this PR do?

This removes the employment route placeholder from our router. We
probably aren't going to have the employment tab ready in time for the
demo, and we'd rather not show off the placeholder.

    modified:   static/js/dashboard.js

This just removes the route for the `EmploymentTab` component.

    modified:   static/js/components/PersonalTab.js

This sets the `PersonalTab` component to redirect to `/dashboard` on
save, instead of to `/profile/professional`.

Make sure nothing is wacky with react-router.

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -18,7 +18,7 @@ class PersonalTab extends React.Component {
       value: language.alpha2,
       label: language.English
     }));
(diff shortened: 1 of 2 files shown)


noisecapella:
:+1:

PR #200, opened 2016-04-21
Added country and city fields

### LINKED ISSUE #149: User enrollment: personal info card (part of profile creation)

noisecapella:
Fixes #149 
#### What's this PR do?
Adds current location, birth location and date of birth fields

PersonalTab.js

Go into the profile, save to confirm validation errors. Fill out the new fields and confirm that validation works.

### DIFF
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -1,7 +1,9 @@
 import React from 'react';
 import TextField from 'react-mdl/lib/Textfield';
 import Select from 'react-select';
+import DatePicker from 'react-datepicker';
 import { browserHistory } from 'react-router';
+import moment from 'moment';
 
 // utility functions for pushing changes to profile forms back to the
 // redux store.
@@ -10,7 +12,7 @@ import { browserHistory } from 'react-router';
 
 // bind to this.boundTextField in the constructor of a form component
 // to update text fields when editing the profile
-function boundTextField(key, label, error) {
+export function boundTextField(key, label, error) {
   const { updateProfile, profile } = this.props;
   let onChange = e => {
     let clone = Object.assign({}, profile);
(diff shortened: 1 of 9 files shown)


static/js/util/profile_edit.js
```diff
@@ -47,14 +49,35 @@ function boundSelectField(key, label, options, error) {
...
+      onChange={onChange}
+    />
+    <span className="validation-error-text">{error}</span>
+  </div>;
+}
+
+
 // bind to this.saveAndContinue.bind(this, '/next/url')
-function  saveAndContinue(next) {
+export function  saveAndContinue(next) {
```
aliceriot: extra space here (between `function` and `saveAndContinue`)

static/js/util/profile_edit.js
```diff
@@ -47,14 +49,35 @@ function boundSelectField(key, label, options, error) {
...
+  return <div>
+    <DatePicker
+      selected={newDate}
+      placeholderText={label}
+      showYearDropdown
+      onChange={onChange}
+    />
+    <span className="validation-error-text">{error}</span>
+  </div>;
+}
```
aliceriot: this is nice and clean! I like the way we've set up these functions, it's a nice pattern that encourages lots of re-use.

aliceriot:
one little formatting nit, other than that I think this is :100: / :+1:

a lot of what you've done here should be helpful when working on the employment history tab too.

PR #205, opened 2016-04-22
Certificates taken in account in dashboard

giocalitri:
#### What's this PR do?
Takes in account certificates in the dashboard

dashboard/api.py

In Micromasters, the course needs to be "past", meaning the end date should be past.
In EDX, you need to create a verified enrollment and certificate for the micromaster course for a user.
To manually create a certificate for a user `bob` for a course `course-v1:odl+GIO101+CR-FALL15`:
```
from certificates.models import CertificateStatuses
from certificates.tests.factories import GeneratedCertificateFactory
from opaque_keys.edx.keys import CourseKey
bob = User.objects.get(username='bob')
course_key = CourseKey.from_string('course-v1:odl+GIO101+CR-FALL15')
GeneratedCertificateFactory.create(user= bob, course_id=course_key, status=CertificateStatuses.downloadable, mode='verified', download_url='www.google.com', grade="0.88",)
```
Make sure the user `bob` is enrolled in the course, then change the enrollment mode to `verified` in the following way:
```
from course_modes.models import CourseMode
from enrollment.api import update_enrollment
CourseMode.objects.create(course_id=course_key, mode_slug='verified', mode_display_name='Gio verified', min_price=50, )
update_enrollment('bob', 'course-v1:odl+GIO101+CR-FALL15', mode='verified')

```
You need to rebuild `web` because the `edx-api-client` library has changed.
You also need the latest edx master to use the certificate rest apis

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -105,7 +105,7 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
     Returns:
         dict: dictionary representing the course status for the user
     """
-    # pylint: disable=unused-argument, too-many-return-statements
+    # pylint: disable=too-many-return-statements
     with transaction.atomic():
         if not course.courserun_set.count():
             return format_course_for_dashboard(None, CourseStatus.NOT_OFFERED, course)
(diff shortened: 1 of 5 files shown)


noisecapella:
Ok, functionality looks good. Looking at code now

dashboard/api.py
```diff
@@ -132,13 +132,21 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
...
+        # if there is no certificate for the user, the user never passed
+        # the course, so she needs to enroll in the next one
+        if not user_certificates.has_verified_cert(run_status.course_run.edx_course_key):
+            return format_course_for_dashboard(
+                course.get_next_run(),
+                CourseStatus.OFFERED if course.get_next_run() is not None else CourseStatus.NOT_OFFERED,
+                course
+            )
+        # pull the verified certificate for course
+        cert = user_certificates.get_verified_cert(run_status.course_run.edx_course_key)
```
noisecapella: Just to confirm, a user does not get a certificate at all if the grade is too low to pass, right?

giocalitri: from my yesterday's conversation with Matt Dryer, no she does not get it. There is a special case when the instructor can make the student pass the course anyway, but in that case the certificate should be similar to a student that passed normally

dashboard/api_test.py
```diff
@@ -309,6 +363,15 @@ def setUpTestData(cls):
             enr_end=now+timedelta(weeks=1),
             edx_key="course-v1:MITx+8.MechCX+2014_T1"
         )
+        # and a run that is past and verified
```
noisecapella: Would we ever have a certificate that's for a course run that has not reached its end date yet? Maybe the course gave our grades early? If you think that's a possibility can you add a test case for it?

noisecapella:
Just the one question, looks great. If there's no work left to be done feel free to merge :+1:

dashboard/api_test.py
giocalitri: I think the case you are describing is not possible in the current logic: if the course is current, we will return the current grade

PR #206, opened 2016-04-22
Removed node-sass rebuild step

noisecapella:
#### What's this PR do?
This removes the `node rebuild node-sass` step which was introduced here: https://github.com/mitodl/teachersportal/pull/43. I don't think anyone uses the OS X libsass library so I believe this isn't useful.

Make sure `docker-compose up` still works fine


giocalitri:
tested: it seems to work fine 👍

PR #207, opened 2016-04-22
After login the user is redirected to dashboard

giocalitri:
#### What's this PR do?
The redirect after the login is to the user dashboard page

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -164,7 +164,7 @@ SOCIAL_AUTH_PIPELINE = (
     'backends.pipeline_api.update_profile_from_edx',
     'backends.pipeline_api.update_from_linkedin',
 )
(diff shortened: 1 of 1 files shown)


Ferdi:
I thought they'd go to dashboard ?

giocalitri:
I thought the idea was that they had to go to profile and then, if the profile was complete, go to the dashboard.
Honestly now it is not a problem either way

noisecapella:
Maybe we should change this to redirect to /dashboard instead? Then in our future PR we would redirect to /profile in the frontend if they have not filled out the profile yet

giocalitri:
fine with me

giocalitri:
done

noisecapella:
:+1:

PR #211, opened 2016-04-25
Incorporated update homepage design

### LINKED ISSUE #97: Home Page Theming

aliceriot:

this closes #97 

#### What's this PR do?

Implements the latest design for the home page. You can see the unincorporated template over [here](https://github.com/mitodl/micromasters-templates)

Oh god, everywhere.

major changes:

- changed webpack config (sharing a file between dev and prod)
- added new webpack entry point (to bundle stylesheets for the homepage)
- updated html for homepage (`/cms/templates/cms/home_page.html`)
- pulled in vendor css and vendor JS (under `/static/scss/public` and `/static/js/public`, respectively)
- added new image assets
- added new navigation theme to the dashboard side of things as well

I tried to cut down on what we needed to check in to version control as much as possible. Unfortunately a number of things are used here that I couldn't find npm packages for (like the `raty` plugin that shows the rating stars, and a bunch of css).

Pull it down, make sure everything looks good. If you can think of any ways to cut the diff down further that would be :100: 

Second version of the homepage. Tried to keep the diff as small as possible.

micromasters_home_page


cms/templates/cms/home_page.html
```diff
@@ -10,122 +10,164 @@
...
+          <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+
+          </div>
+        -->
       </div>
     </div>
   </div>
-</footer>
 
+  <div id="wtf">
```
noisecapella: Can you remove this? Is it necessary for something?

aliceriot: oh gosh, good eye :) was a debugging things

cms/templates/cms/home_page.html
```diff
@@ -10,122 +10,161 @@
...
+            {{ title }}
+          </h1>
         </div>
       </div>
-    </div>
 
+      <div class="row sub-banner">
+        <div class="col-md-12">
+          <div class="sub-banner-content">
+            MIT has created a new credential for online learners: the Micromasters, which will be granted by <a href="https://www.edx.org/school/mitx" target="_blank">MITx</a> to students who do exceptionally well in a given set of graduate-level online courses and do well in a subsequent exam. <a href="javascript:void()" data-target="#modal-about" data-toggle="modal">Learn more</a>.
```
noisecapella: Is 'Micromasters' meant to be plural here? It is in the original template. cc @pdpinch

noisecapella:
When you hover over the program box it shows a link cursor even though there's no link. Not sure if we care enough to change it though, I assume all programs will have links eventually anyway

static/js/components/Header.js
```diff
@@ -1,24 +1,28 @@
...
-        <LoginButton />
-      </div>
-    </div>;
+  render () {
+    return (
+      <Navbar bsStyle="default" fluid={true}>
+        <Navbar.Header>
+          <Navbar.Brand>
+            <a href="/">
+              <img src="/static/images/logo-micromasters@2x.png" width="215" alt="MIT Micromasters" />
```
noisecapella: Can you add `height` too?

static/js/components/Header.js
```diff
@@ -1,24 +1,28 @@
...
+            </a>
+          </Navbar.Brand>
+          <ul className="nav navbar-toolbar navbar-left">
+            <NavItem href="/dashboard">
```
noisecapella: Is there a way to integrate this better with `react-router`? Right now it does a page refresh when you click this

static/js/public/core.js
```diff
@@ -0,0 +1,291 @@
+/* eslint-disable no-undef, indent, no-console, no-undef */
```
noisecapella: Is this all copied from somewhere or did you author it?

package.json
```diff
@@ -26,10 +26,12 @@
     "babel-plugin-jsx": "1.2.0",
     "babel-preset-es2015": "6.6.0",
     "babel-preset-react": "6.5.0",
+    "bootstrap": "3.3.6",
```
noisecapella: FYI we already have a link to the bootstrap CSS in `base.html` somewhere. Feel free to ignore for this PR but eventually we should switch to using whatever comes with this library

package.json
```diff
@@ -50,6 +52,7 @@
     "sass-loader": "3.1.2",
     "striptags": "2.1.1",
     "style-loader": "0.13.0",
+    "url-loader": "0.5.7",
```
noisecapella: What do we use this for?

ui/templates/base.html
```diff
@@ -7,6 +7,10 @@
     <script type="text/javascript">
     var SETTINGS = {{ js_settings_json|safe }};
     </script>
+    {% if style_public_src %}
+    <script type="text/javascript" src="{{ style_public_src }}"></script>
+    <script src="https://code.jquery.com/jquery-2.2.0.min.js"></script>
```
noisecapella: Aren't we already pulling in jquery in `public.js`?

noisecapella:
Looks good overall, just some questions.

static/js/components/Header.js
aliceriot: yes, I think so. I think a `Link` with the right `className` set will be styled correctly

static/js/public/core.js
aliceriot: do you mean the file? nicholae wrote it (or it was included with the newer version of the template anyway).

package.json
aliceriot: this is for the JS

ui/templates/base.html
aliceriot: yeah, I had this in for debugging and missed it :D

aliceriot: hrm, actually tried removing this and ran into some issues.

do you think it's fine to rely on the cdn version? it works fine if I just leave the cdn `script` tag up here and remove jquery from the `public.js` bundle.

package.json
aliceriot: check `loaders` in the webpack config, it's for including static assets like images and stuff.

here:

loaders

aliceriot: I think the `url-loader` will serve the asset via base64 encoding in a url if it's small enough (which is why it's a little niftier than `file-loader`)

ui/templates/base.html
noisecapella: I'm a bit worried about race conditions if it's not in `public.js`. The CDN might take a little longer than our JS blob to download.

static/js/components/Header.js
aliceriot: what do you mean? in `header.scss` we set `min-height` on `.navbar` and padding on `.navbar-brand`

ui/templates/base.html
aliceriot: true, I'll try to debug using it in `public.js` then

static/js/components/Header.js
noisecapella: There's `width="215"` in this line but no `height`. It's not strictly necessary but it would be good to be complete

ui/templates/base.html
aliceriot: well, aren't `<script>` tags executed in the order of their declaration, regardless of how long the request takes? we aren't setting the `async` or `defer` attributes.

static/js/components/Header.js
aliceriot: ah gotcha, I had just copy-pasted that from nicholae's code.

ui/templates/base.html
aliceriot: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

noisecapella: Huh, I didn't realize that. Feel free to leave it

aliceriot: sounds good :)

pdpinch:
@noisecapella @aliceriot can we merge this?

cms/templates/cms/home_page.html
```diff
@@ -10,122 +10,161 @@
...
+          <h3>Join the MIT Micromasters community</h3>
+          <p>
+          Micromasters aren't just about courseware. Part of the the MIT experience is interacting with other exceptional students like yourself. Fill out your profile on this site and build professional relationships with like-minded learners.
+          </p>
+          <h3>Micromasters courses are delivered through edX.org</h3>
+          <p>
+          Use your edX.org account to enroll in a Micromasters program and share your grades and progress with MIT. When you complete all the requirements and earn a Micromasters credential, you can use it to apply for a Masters degree at MIT.
+          </p>
+        </div>
+        <!--
```
noisecapella: Can you remove these lines here?

noisecapella:
:+1:

cms/templates/cms/home_page.html
aliceriot: removed!

PR #212, opened 2016-04-25
Added ProfileTab utility class

aliceriot:

none

#### What's this PR do?

this creates a utility class called `ProfileTab` which lets us cut down on boilerplate a little (when setting up the profile tabs).

we just write components that inherit from `ProfileTab`, and the `boundTextField` and so on functions are already bound to `this` (just remember to call `super(props)`in your constructor!).

read the new class (`static/js/util/ProfileTab.js) and review the changes to the personal tab (`static/js/components/PersonalTab.js`).

Make sure this doesn't clobber any existing behavior.

my name is alice and I hate boilerplate.

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -1,21 +1,13 @@
 import React from 'react';
 import Button from 'react-mdl/lib/Button';
 
-import {
-  boundDateField,
-  boundTextField,
-  boundSelectField,
-  saveAndContinue,
-} from '../util/profile_edit';
+import ProfileTab from "../util/ProfileTab";
+import { saveAndContinue } from "../util/profile_edit";
 import LANGUAGE_CODES from '../language_codes';
-import COUNTRIES from '../countries';
 
-class PersonalTab extends React.Component {
+class PersonalTab extends ProfileTab {
   constructor(props) {
     super(props);
-    this.boundTextField = boundTextField.bind(this);
-    this.boundSelectField = boundSelectField.bind(this);
-    this.boundDateField = boundDateField.bind(this);
     this.genderOptions = [
       { value: 'm', label: 'Male' },
       { value: 'f', label: 'Female' },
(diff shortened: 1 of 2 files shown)


static/js/components/PersonalTab.js
```diff
@@ -25,26 +17,22 @@ class PersonalTab extends React.Component {
...
-        )}<br />
+      {this.boundTextField("last_name", "Family name", errors.last_name)}<br />
+      {this.boundTextField("preferred_name", "Preferred name", errors.preferred_name)}<br />
+      {this.boundSelectField('gender', 'Gender', this.genderOptions, errors.gender)}<br />
+      {this.boundSelectField(
+        'preferred_language',
+        'Preferred language',
+        this.languageOptions,
+        errors.preferred_language
+      )}<br />
```
aliceriot: this is unrelated, I just noticed that the indentation wasn't correct.

static/js/util/ProfileTab.js
```diff
@@ -0,0 +1,23 @@
...
+
+class ProfileTab extends React.Component {
+  constructor(props) {
+    super(props);
+    this.boundTextField = boundTextField.bind(this);
+    this.boundSelectField = boundSelectField.bind(this);
+    this.boundDateField = boundDateField.bind(this);
+    this.countryOptions = COUNTRIES.map(country => ({
+      value: country.Code,
+      label: country.Name
```
aliceriot: I pulled `countryOptions` in here since I think we'll also want these options on both the employment and education tabs.

static/js/components/PersonalTab.js
```diff
@@ -25,26 +17,22 @@ class PersonalTab extends React.Component {
       value: language.alpha2,
       label: language.English
     }));
-    this.countryOptions = COUNTRIES.map(country => ({
-      value: country.Code,
-      label: country.Name
-    }));
     this.saveAndContinue = saveAndContinue.bind(this, '/dashboard');
```
aliceriot: this we bind with a specific argument, so we still need to bind it in each tab's constructor.

we could instead make `saveAndContinue` push a `nextTabUrl` prop onto the history, and then it could be bound in `ProfileTab` instead of here, but eh, same amount of code either way (still would have to declare what the `nextTabUrl` prop looked like on every tab component).

noisecapella:
Looks nice and simple :+1:

PR #213, opened 2016-04-25
Added script to run development server on OS X without docker

### LINKED ISSUE #142: Create a script to run micromasters natively for OS X users
I'm thinking the script will:
 - create a virtualenv in the directory if necessary and pip install requirements
 - npm install
 - read environment variables from `.env`
 - fork three processes for Django, postgres and the webpack watcher

It would require some manual configuration for postgres but hopefully nothing too complicated

noisecapella:
Fixes #142

#### What's this PR do?
Provides a script to let OS X users (and Linux users) run the webserver and webpack dev servers from the host environment instead of within Docker.

This requires a whole lot of manual installation steps compared to Docker setup, but it should make the development process easier:
 - Postgres must be installed and its binaries added to the `$PATH`
 - The postgres user must have the password 'postgres'
 - Python 3 should be installed
 - Node JS 4.x or later should be installed
 - All dependencies of the various Python libraries need to be installed first. (There are many here.)

After the manual installation is done, the user should just need to run `./osx_start.sh`. The python and npm dependencies will be installed, then the servers will start. The script will trap SIGTERM and SIGINT to kill these servers when Ctrl+C is pressed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,5 @@ codekit-config.json
 
 # Django static
 staticfiles/
+
+.venv
--- a/osx_start.sh
+++ b/osx_start.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+set -euf -o pipefail
+set -a
+
+export NODE_ENV=development
+export DEBUG=True
+export PORT=8079
+export COVERAGE_DIR=htmlcov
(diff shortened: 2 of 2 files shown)


annagav:
Thank you for the script! 👍

PR #235, opened 2016-04-26
Moved onUpdate from Route to Router

noisecapella:
#### What's this PR do?
Fixes a bug where we do the google analytics update in the wrong place

This is somewhat difficult to test locally. It's probably best to just confirm the fix with the docs for react-router, then merge, then look at the google analytics dashboard for micromasters-ci.


annagav:
👍

PR #238, opened 2016-04-27
Added page for terms of service

### LINKED ISSUE #178: Terms of Service
We need a page for terms of service, and a link to it from the footer.

noisecapella:
Fixes #178 

#### What's this PR do?
Adds a required terms of service page. When the user clicks 'Accept' the user will be directed to the profile. If they click 'Cancel' the user will be logged out and redirected to the homepage. There is also a new field in the REST API and model to keep track of whether a user agreed to this.

Since there are UI tests here there is a bit of new code to handle this testing. The routes are also moved into `dashboard_routes.js`. 

static/js/containers/TermsOfServicePage.js

You should be forced to go to the terms of service page whenever you are logged in and are in the dashboard area. When you agree to the terms of service (currently blank) you should be allowed to go anywhere.


giocalitri:
The code looks good and functions well, but there is a small problem of UI:
every time I load a page before I accept the terms of services, I can actually see the page.
terms_of_service

noisecapella:
In order to solve this UI quirk we need to block the UI on loading the profile from the REST API, since the profile has the information on whether the user already agreed to it

noisecapella:
What if I add code to redirect to `/terms_of_service` or `/profile` after login on the Django side? I think as long as we are redirected to the right place the first time, the UI quirk you mentioned is only achieved by a user experimenting around with the URL

giocalitri:
this does not really work after the very first login

noisecapella:
I added a decorator to redirect to terms_of_service on the Django side. Can you try the login flow again?

ui/decorators.py
```diff
@@ -0,0 +1,27 @@
+"""
+Decorators for views
+"""
+from functools import wraps
+
+from django.shortcuts import redirect
+
+
+def require_terms_of_service(f):
```
giocalitri: please do not use 1 letter variable

ui/decorators.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    @wraps(f)
+    def wrapper(request, *args, **kwargs):
+        """
+        Wrapper for require_terms_of_service
+
+        Args:
+            request (django.http.request.HttpRequest): A request
+        """
+        if not request.user.is_anonymous() and request.path != '/terms_of_service/':
```
giocalitri: `'/terms_of_service/'` should probably go in a variable: at least a global one inside this module

giocalitri:
two very small changes and then good to go 👍

giocalitri:
you need to rebase

noisecapella:
The UI looks a bit weird after the rebase but we need to update the UI anyway in the future to actually add terms of service, so I'm going to punt on this

PR #241, opened 2016-04-28
Education tab

### LINKED ISSUE #151: User Enrollment: education card (profile creation)
As a user, I need to enter information about my educational history. This will help me network with others, and it will help me apply for a masters degree.

~~The first field, "Education Completed" is populated from edX, but the user can change it.~~

We will present the user with a series of switches, one for each type of degree. If switch is set to "on" the user must add a degree and complete the form before continuing. However, the user can set any and all degrees to "off" if they in fact don't have the degree (none are required). High school and Bachelors will default to being turned on, since these are the degrees that we expect micromasters applicants to have. 

The user can enter as many degrees/certifications as she chooses.

For each degree/certification, the user must enter:

Degree/Certification
School Name or Granting Authority
School City
School State/Territory (if Country == US)
School Country 
Subject
Online (switch component)
Graduation Month and Year (or expected graduation)

For now, we won't worry about normalizing the data. Later, when we have more users we will clean up the data. 

Screen designs:

https://projects.invisionapp.com/share/KR6FYCZTX#/screens/153562861
https://projects.invisionapp.com/share/KR6FYCZTX#/screens/155766490

Wireframes: 

education_default
education_modal
education_list

stretch goal: When we get a value from edX, we can use the "Education Completed" Field to make some assumptions about which switches to display. 

Doctorate => Ph.D, Bachelor's Degree, High School
Master's or Professional => Master's Degree, Bachelor's Degree, High School
Associate Degree => Associate Degree, High School
High School => High School

annagav:
Fix #151
#### What's this PR do?
Add education tab to profile page.
EducationTab.js
- go to profile page, 
- click on the 'Education' tab, 
- click the + button,
- fill out the form to create new education instance


noisecapella:
Thanks! I'm going to make some comments even though this is not ready for review yet, feel free to ignore if you're going to get to them eventually

profiles/models.py
```diff
@@ -144,3 +145,21 @@ def __str__(self):
...
+        (MASTERS, "Master's or professional degree"),
+        (BACHELORS, "Bachelor's degree"),
+        (ASSOCIATE, "Associate degree"),
+        (HIGH_SCHOOL, "High school"),
+        (OTHER_EDUCATION, "Other education"),
+    )
+    profile = models.ForeignKey(Profile, related_name='education')
+    degree_name = models.CharField(max_length=30, choices=DEGREE_CHOICES)
+    graduation_date = models.DateField()
+    subject = models.CharField(max_length=30)
```
noisecapella: Can you change the length limit for these four lines to 255? 30 or 50 may be too low

annagav: I am going to define it as `models.TextField()`. Does it make sense to  set  `blank=True, null=True`, given that we require the student to fill in everything?

profiles/models.py
```diff
@@ -144,3 +145,21 @@ def __str__(self):
...
+    school_name = models.CharField(max_length=50)
+    school_city = models.CharField(max_length=30)
```
noisecapella: Can you add school_state_or_territory?

profiles/models.py
noisecapella: I don't think it's necessary. For the profile we needed it most places because the most of profile starts off blank, but we're only creating `Education` objects when the user is editing their own profile.

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+    this.boundEducationMonthField = boundEducationMonthField.bind(this);
+    this.boundEducationYearField = boundEducationYearField.bind(this);
+    this.educationLevelOptions = [
+      { value: 'hs', label: "High school", enabled: true },
+      { value: 'a', label: 'Associate degree', enabled: true },
+      { value: 'b', label: "Bachelor's degree", enabled: true },
+      { value: 'm', label: "Master's or professional degree", enabled: false },
+      { value: 'p', label: "Doctorate", enabled: false }
+    ];
+    this.educationLevelLabels = {
```
noisecapella: Can you create this object using the data from `educationLevelOptions` to reduce duplication?

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+} from '../actions';
+import { Grid, Cell, Switch, Dialog, DialogTitle, DialogContent, DialogActions, FABButton, Icon } from 'react-mdl';
+
+class EducationTab extends ProfileTab {
+  constructor(props) {
+    super(props);
+    this.boundEducationMonthField = boundEducationMonthField.bind(this);
+    this.boundEducationYearField = boundEducationYearField.bind(this);
+    this.educationLevelOptions = [
+      { value: 'hs', label: "High school", enabled: true },
```
noisecapella: Can we define each of the values here as constants in `constants.js` since we use these in the reducers too?

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+      { value: 'a', label: 'Associate degree', enabled: true },
```
noisecapella: It looks like `enabled` isn't used here, can you remove it?

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+  handleOpenDialog(level, index){
+    const { dispatch, profile, updateProfile } = this.props;
+    let new_index = index;
+    if (index === null){
+      new_index = profile['education'].length;
+    }
+    dispatch(openEducationForm(level, new_index));
+
+    let clone = Object.assign({}, profile);
+    clone['education'].push(this.addNewEducation(level));
```
noisecapella: The clone is a shallow copy but this modifies an item within the object. Can you change this to

    clone['education'] = clone['education'].concat(this.addNewEducation(level));

Or do it within the `Object.assign`:

      education: profile.education.concat(this.addNewEducation(level))
    });

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+    this.handleOpenDialog = this.handleOpenDialog.bind(this);
+    this.handleCloseDialog = this.handleCloseDialog.bind(this);
+  }
+  printAddDegree(level){
+    const { educationLevels } = this.props;
+    if (educationLevels[level.value]){
+      return <Grid key={"add-"+level.value}>
+        <Cell col={11}></Cell>
+        <Cell col={1}>
+          <FABButton mini onClick={this.handleOpenDialog.bind(this, level.value, null )} raised ripple>
```
noisecapella: It looks like the `index` parameter is never set to anything but null. Is there another place where `handleOpenDialog` is meant to be used?

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+            <Icon name="add" />
+          </FABButton>
+        </Cell>
+
+      </Grid>;
+    }
+  }
```
noisecapella: Can you add a `return null;` here, just to be explicit?

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+    updateProfile(clone);
+
+  }
+  handleCloseDialog(){
+    const { dispatch } = this.props;
+    dispatch(closeEducationForm())
+  }
+  addNewEducation(level){
+    return {
+      'degree_name': level,
```
noisecapella: Can we move this function into `util.js` so we can unit test it? There's a similar object in the employment tab which we should move into a function also

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+          </FABButton>
+        </Cell>
+
+      </Grid>;
+    }
+  }
+  printExistingEducations(level){
+    const { profile } = this.props;
+    return profile['education'].map(education => {
+      if(education.degree_name === level.value && 'id' in education) {
```
noisecapella: I like the `'id' in education` here. We should remove the `id: null` default from the employment tab default data so the two are consistent

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+
+  render() {
+    let {educationDialog, educationLevels} = this.props;
+
+    var levels_grid = this.educationLevelOptions.map(level =>{
+      return <Cell col={12} key={level.value} >
+        <Grid key={level.value} className="education-level-header">
+          <Cell col={11} key="name"><h5 className="education-level-name">{level.label}</h5></Cell>
```
noisecapella: We shouldn't need the `key` attribute anywhere except at the root `Cell` object

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,198 @@
...
+          <Cell col={1} key="switch">
+            <Switch ripple id="switch1" onChange={()=>{this.handleSwitchClick(level.value)}}
```
noisecapella: Can you remove the `id="switch1"` if it's not relevant?

static/js/reducers/index.js
```diff
@@ -56,6 +61,45 @@ export const courseList = (state = INITIAL_COURSE_LIST_STATE, action) => {
...
+        educationIndex: action.payload.index
+      });
+    case HIDE_EDUCATION_FORM_DIALOG:
+      return INITIAL_DIALOG_STATE;
+    default:
+      return state;
+  }
+};
+export const INITIAL_EDUCATION_LEVEL_STATE = {
+  hs: true,
```
noisecapella: Can you use constants for the keys here?

static/js/reducers/index.js
```diff
@@ -56,6 +61,45 @@ export const courseList = (state = INITIAL_COURSE_LIST_STATE, action) => {
...
+  a: true,
+  b: true,
+  m: false,
+  d: false
+};
+export const educationLevels = (state = INITIAL_EDUCATION_LEVEL_STATE, action) => {
+  switch (action.type) {
+    case TOGGLE_EDUCATION_LEVEL:
```
noisecapella: Can you change this action to provide the new state as an argument? It will make it slightly easier to test the action.

static/js/util/profile_edit.js
```diff
@@ -71,7 +73,68 @@ export function boundDateField(key, label, error) {
     <span className="validation-error-text">{error}</span>
   </div>;
 }
+export function boundEducationMonthField(keySet, label) {
+  const {updateProfile, profile} = this.props;
+  let onChange = month => {
+    let clone = Object.assign({}, profile);
```
noisecapella: Can you change this to use `_.cloneDeep`? We will need to do a deep clone here since `_.set` can modify any object within the clone, not just the root object.

static/js/util/profile_edit.js
```diff
@@ -71,7 +73,68 @@ export function boundDateField(key, label, error) {
...
 
+    let current_value = _.get(clone, keySet);
+        // format as ISO-8601
+    if(month === null){
+      _.set(clone, keySet, null);
+    }else if (current_value){
+      _.set(clone, keySet, moment(current_value).set('month', month.value).format('YYYY-MM-DD'));
```
noisecapella: Can you create a constant for `YYYY-MM-DD` and replace the string in all places within the project where it is used?

static/js/util/profile_edit.js
```diff
@@ -71,7 +73,68 @@ export function boundDateField(key, label, error) {
...
   </div>;
 }
+export function boundEducationMonthField(keySet, label) {
+  const {updateProfile, profile} = this.props;
+  let onChange = month => {
+    let clone = Object.assign({}, profile);
 
+    let current_value = _.get(clone, keySet);
+        // format as ISO-8601
+    if(month === null){
```
noisecapella: Can you change this if statement to set a local variable instead, and use `_.set` once at the end of the if statement with the local variable?

static/js/util/profile_edit.js
```diff
@@ -71,7 +73,68 @@ export function boundDateField(key, label, error) {
...
+    { value: 6, label: 'July'},
+    { value: 7, label: 'August'},
+    { value: 8, label: 'September'},
+    { value: 9, label: 'October'},
+    { value: 10, label: 'November'},
+    { value: 11, label: 'December'}
+  ];
+
+
+  return <Select
```
noisecapella: Should the user be able to clear the month select field? If so we should handle it in the above logic.

static/scss/layout.scss
```diff
@@ -56,3 +56,28 @@ body {
     width: 100%;
     height: 100%;
 }
+
```
noisecapella: Can you move this to `dashboard.scss`?

static/js/util/profile_edit.js
noisecapella: Actually this code may handle the case already, need to double check

pdpinch:
If it's not too much trouble, can you add a screenshot of what this looks like currently?

annagav:
screen shot 2016-05-03 at 3 44 52 pm 1

annagav:
screen shot 2016-05-03 at 3 47 01 pm

noisecapella:
It looks like you will need to add a dialog polyfill to use dialogs in non-Chrome browsers. I'm investigating how to fix this

noisecapella:
I'm going to make a few more comments, feel free to ignore if you already fixed them locally

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+      options={month_options}
+      value={month}
+      placeholder={label}
+      onChange={onChange}
+    />;
+}
+
+export function boundEducationYearField(keySet, label){
+  const {updateProfile, profile} = this.props;
+  let onChange = year => {
```
noisecapella: This takes an event as an argument since it's a textfield, so you'll need to do something like `e.target.value` to retrieve the value here.

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+    let clone = Object.assign({}, profile);
+    let current_value = _.get(clone, keySet);
+    // format as ISO-8601
+    if(month === null) {
```
noisecapella: This should be `year`

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+      value={month}
+      placeholder={label}
+      onChange={onChange}
+    />;
+}
+
+export function boundEducationYearField(keySet, label){
+  const {updateProfile, profile} = this.props;
+  let onChange = year => {
+    let clone = Object.assign({}, profile);
```
noisecapella: This should be a `_.cloneDeep` since we are using `_.set` which can modify anywhere in the profile object

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+    let current_value = _.get(clone, keySet);
+    // format as ISO-8601
+    if(month === null) {
+      _.set(clone, keySet, null);
+    }else if (current_value){
+      _.set(clone, keySet, moment(current_value).set('year', year.value).format('YYYY-MM-DD'));
```
noisecapella: `year.value` should be `year`

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+
+  return <Select
+      options={month_options}
+      value={month}
+      placeholder={label}
+      onChange={onChange}
+    />;
+}
+
+export function boundEducationYearField(keySet, label){
```
noisecapella: Can you rename to `boundYearField` since it doesn't need to be specific to education?

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
     <span className="validation-error-text">{_.get(errors, keySet)}</span>
   </div>;
 }
+export function boundEducationMonthField(keySet, label) {
```
noisecapella: Can you rename to `boundMonthField`?

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+      onChange={onChange}
+    />;
+}
+
+export function boundEducationYearField(keySet, label){
+  const {updateProfile, profile} = this.props;
+  let onChange = year => {
+    let clone = Object.assign({}, profile);
+    let current_value = _.get(clone, keySet);
+    // format as ISO-8601
```
noisecapella: Since this is a textfield `year` might include some non-numeric characters. moment.js has [isValid](http://momentjs.com/docs/#/parsing/is-valid/) which can be used to check if a date is valid, and this will handle things like negative numbers too. Can you change this function to use `isValid` to check that `moment(current_value).set('year', year)` results in a valid date object, and `_.set` with null if not?

static/js/util/profile_edit.js
```diff
@@ -89,7 +89,68 @@ export function boundDateField(keySet, label) {
...
+    if(month === null) {
+      _.set(clone, keySet, null);
+    }else if (current_value){
+      _.set(clone, keySet, moment(current_value).set('year', year.value).format('YYYY-MM-DD'));
+    } else {
+      _.set(clone, keySet, moment().set('year', year.value).format('YYYY-MM-DD'));
+    }
+    updateProfile(profile_with_education);
+  };
+  let year = _.get(profile, keySet)? moment(_.get(profile, keySet), "YYYY-MM-DD").year() : null;
```
noisecapella: Can you replace `null` with `""`? React has started complaining about assigning `null` and `undefined` as the `value` of a textfield

Ferdi:
@annagav  Looking at the screenshots , it looks like we are making great progress !

Can you move those screenshots as part of the issue #151 so that we can discuss the functionality there and only code review here ?

annagav:
@Ferdi I added the screenshots to issue #151.

static/js/components/EducationTab.js
annagav: Should I do it in this PR?

noisecapella: No need, it was just done in a different PR

static/js/util/profile_edit.js
annagav: yes it takes care of that. The only problem is that the month and the year are getting cleared.

static/js/util/profile_edit.js
annagav: or just leave it the value it had before, if the date is not valid.

noisecapella: Sounds good to me

static/js/actions/index.js
```diff
@@ -79,6 +79,7 @@ export const updateProfileValidation = errors => ({
 
 export const validateProfile = (profile, requiredFields, messages) => {
   return dispatch => {
+    console.log(requiredFields);
```
noisecapella: Can you remove this?

static/js/components/EducationTab.js
```diff
@@ -2,50 +2,95 @@ import React from 'react';
...
 } from '../actions';
+import { HIGH_SCHOOL, ASSOCIATE, BACHELORS, MASTERS, DOCTORATE } from '../constants';
 import { Grid, Cell, Switch, Dialog, DialogTitle, DialogContent, DialogActions, FABButton, Icon } from 'react-mdl';
 
 class EducationTab extends ProfileTab {
   constructor(props) {
     super(props);
-    this.boundEducationMonthField = boundEducationMonthField.bind(this);
-    this.boundEducationYearField = boundEducationYearField.bind(this);
+    this.boundEducationMonthField = boundMonthField.bind(this);
```
noisecapella: Can you also rename these two variables to remove the `Education` part?

static/js/components/EducationTab.js
```diff
@@ -2,50 +2,95 @@ import React from 'react';
...
+  openNewEducationForm(level, index) {
+    const {dispatch, profile, updateProfile} = this.props;
     let new_index = index;
     if (index === null){
       new_index = profile['education'].length;
     }
     /* add empty education */
     let clone = Object.assign({}, profile);
-    clone['education'].push(this.addNewEducation(level));
+    clone['education'].push(generateNewEducation(level));
```
noisecapella: You can't use `push` here since it alters the list and you've only made a shallow copy. Can you rewrite using `concat` instead?

static/js/components/EducationTab.js
```diff
@@ -82,54 +129,51 @@ class EducationTab extends ProfileTab {
...
-    };
-  }
+
 
 
   render() {
-    let {educationDialog, educationLevels} = this.props;
+    let { profile, educationDialog, educationLevels} = this.props;
+
+    if (!profile['education']){
```
noisecapella: Can you change to `if (profile['education'] === undefined)`?

static/js/components/EducationTab.js
```diff
@@ -138,7 +182,13 @@ class EducationTab extends ProfileTab {
         
       </Cell>
     });
-    let keySet = (key) =>['education', educationDialog.educationIndex, key];
+    let index = educationDialog.educationIndex;
+    if (!index) {
```
noisecapella: This won't work if `index` is `0`. Can you check against `null` explicitly?

static/js/reducers/index.js
```diff
@@ -28,38 +28,42 @@ import {
...
+
+export const INITIAL_EDUCATION_LEVEL_STATE = education_level_state;
+
 export const educationLevels = (state = INITIAL_EDUCATION_LEVEL_STATE, action) => {
   switch (action.type) {
     case TOGGLE_EDUCATION_LEVEL:
-      let updated_state = {};
-      updated_state[action.payload.level]= !state[action.payload.level];
-      return Object.assign({}, state, updated_state);
+      return Object.assign({}, state, action.payload.new_state);
```
noisecapella: Can you use camelcase here? Unless we are dealing with data from the REST API we should use camelcase everywhere

static/js/actions/index.js
```diff
@@ -37,10 +41,20 @@ export const updateProfile = profile => ({
   type: UPDATE_PROFILE,
   payload: { profile }
 });
+export const openEducationForm = (level, index, education_id) => ({
```
noisecapella: It seems like this should only take the `index`, `education_id` duplicates this purpose. Can you remove `education_id` from the reducer too?

giocalitri:
@annagav please fix the unit tests

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+  boundYearField,
+  saveAndContinue
+} from '../util/profile_edit';
+import {
+  toggleEducationLevel,
+  openEducationForm,
+  closeEducationForm
+} from '../actions';
+import { HIGH_SCHOOL, ASSOCIATE, BACHELORS, MASTERS, DOCTORATE } from '../constants';
+import { Grid, Cell, Switch, Dialog, DialogTitle, DialogContent, DialogActions, FABButton, Icon } from 'react-mdl';
```
aliceriot: it seems like we usually put dependency imports first, followed by a line break and then imports from our own stuff, eg:

```js
import React from 'react';
import _ from 'lodash';

import ProfileTab from '../components/ProfileTab';
```

might make sense to stick with that here, especially when there's so much stuff being imported

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+      'online_degree': 'Online Degree',
+      'school_name': 'School name',
+      'school_city': 'City',
+      'school_state_or_territory': 'State',
+      'school_country': 'Country'
+    }
+  };
+
+  static validation(profile, requiredFields) {
+    const keys = [
```
aliceriot: can you refactor this so that the keys for validating nested objects are a class property of the `EducationTab` class?

should be pretty simple, something like

```js
static nestedValidationFields = [ ... ];

validation ( ... ) {
  let nestedFields = (index) => {
    let keySet = (key) => ['education', index, key];
    return EducationTab.nestedValidationKeys.map(key => keySet(key));
  };
}
```

this will make it easier to implement the complete profile check later for the education tab (specifically, it's handy for testing).

aliceriot: ( you can go look here: https://github.com/mitodl/micromasters/pull/271 for an example of how this works, look at my changes to `EmploymentTab` and the tests associated with it)

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+    const { educationLevels } = this.props;
+    if (educationLevels[level.value]){
+      return <Grid key={"add-"+level.value}>
+        <Cell col={11}></Cell>
+        <Cell col={1}>
+          <FABButton mini onClick={this.openNewEducationForm.bind(this, level.value, null )} raised ripple>
+            <Icon name="add" />
+          </FABButton>
+        </Cell>
+
```
aliceriot: whitespace, ahh!

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+  openNewEducationForm(level, index) {
+    const {dispatch, profile, updateProfile} = this.props;
+    let new_index = index;
+    if (index === null){
+      new_index = profile['education'].length;
+    }
+    /* add empty education */
+    let clone = Object.assign({}, profile);
+    clone['education'].concat(generateNewEducation(level));
```
aliceriot: `.concat` doesn't mutate an array in place, so I think this your call to `generateNewEducation` is being thrown on the floor here (unless I'm misreading this)

aliceriot: you can do

```js
clone['education'] = clone['education'].concat(...)
```

instead

aliceriot:
I think it would be good to write some tests for the new functions you wrote in `profile_edit.js`, just similar to what's already there (making sure callbacks get call, they receive the right props, etc).

static/js/components/EducationTab.js
aliceriot: (or use `clone['education'].push`)

annagav: I think you are right. Not sure why George asked me to replace push with concat.

aliceriot: well, I think `concat` is generally preferable, mutating arrays can get messy, esp. when dealing with the store, cloning, `Object.assign` and so on.

he might have meant to do something like `clone['education'] = clone['education'].concat(...)`

annagav: my mistake :P

static/js/components/EducationTab.js
annagav: good to know! Thanks!

profiles/serializers.py
```diff
@@ -69,16 +88,34 @@ def get_pretty_printed_student_id(self, obj):
...
                     continue
                 else:
                     setattr(instance, attr, value)
-
             instance.save()
             if 'work_history' in validated_data:
                 update_work_history(validated_data['work_history'], instance.id)
 
-        return instance
+            profile_id = instance.id
```
noisecapella: It would be nice to move this block of code into an `update_education` function similar to `update_work_history` above, but we can do this in another PR

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+    saveProfile: React.PropTypes.func,
+    updateProfile: React.PropTypes.func,
+    errors: React.PropTypes.object,
+    dispatch: React.PropTypes.func.isRequired
+  };
+
+  static defaultProps = {
+    requiredFields: [],
+    validationMessages: {
+      'degree_name': "'Degree level",
```
noisecapella: Can you remove the apostrophe before `Degree level`?

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+    super(props);
+    this.boundMonthField = boundMonthField.bind(this);
+    this.boundYearField = boundYearField.bind(this);
+    this.educationLevelOptions = [
+      {value: HIGH_SCHOOL, label: "High school"},
+      {value: ASSOCIATE, label: 'Associate degree'},
+      {value: BACHELORS, label: "Bachelor's degree"},
+      {value: MASTERS, label: "Master's or professional degree"},
+      {value: DOCTORATE, label: "Doctorate"}
+    ];
```
aliceriot: what do you think about moving this stuff into the constructor for `ProfileTab`? I think we may want the month and year fields for the work history tab, I've also been putting stuff like `educationLevelOptions` in the `ProfileTab` constructor so that these classes inheriting from `ProfileTab` end up looking more declarative (just saying what kinds of inputs should be drawn and so on)

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,249 @@
...
+
+  openEditEducationForm(level, education_id){
+    const { dispatch, profile } = this.props;
+
+    let index = profile['education'].findIndex((education) => {
+      return ('id' in education) && education_id === education.id;
+    });
+    dispatch(openEducationForm(level, index));
+
+  }
```
aliceriot: style nit: I like leaving a blank line between methods:

```js
class Foo {
  fooBar () {
    ...
  }

  fooBarBaz () {
    ...
  }
}
```

static/js/reducers/index.js
```diff
@@ -23,6 +27,48 @@ import {
...
+education_level_state[BACHELORS] = true;
+education_level_state[MASTERS] = false;
+education_level_state[DOCTORATE] = false;
+
+export const INITIAL_EDUCATION_LEVEL_STATE = education_level_state;
+
+export const educationLevels = (state = INITIAL_EDUCATION_LEVEL_STATE, action) => {
+  switch (action.type) {
+  case TOGGLE_EDUCATION_LEVEL:
+    return Object.assign({}, state, action.payload.newState);
```
noisecapella: Since `newState` contains the entire set of education levels, we could just `return action.payload.newState` here instead

static/js/util/profile_edit.js
```diff
@@ -125,7 +126,79 @@ export function boundDateField(keySet, label) {
...
+
+export function boundYearField(keySet, label) {
+  const {updateProfile, errors, profile} = this.props;
+  let onChange = year => {
+    let clone = _.cloneDeep(profile);
+    let current_value = _.get(clone, keySet);
+    // format as ISO-8601
+    let date_value;
+    if (!year) {
+      date_value = "";
```
noisecapella: This won't work with `date_value.isValid()` below. Can you use some value which is invalid, for example `moment("")`?

static/js/util/profile_edit.js
```diff
@@ -125,7 +126,79 @@ export function boundDateField(keySet, label) {
...
+        // format as ISO-8601
+    let date_value;
+    if(!month){
+      date_value = "";
+    }else if (current_value){
+      date_value = moment(current_value).set('month', month.value).format(DATE_FORMAT);
+    } else {
+      date_value = moment().set('day', 1).set('month', month.value).format(DATE_FORMAT);
+    }
+    _.set(clone, keySet, date_value);
```
noisecapella: It looks like you need to use `date_value.isValid` here like you do below. `date_value = ""` should also be updated as below

static/js/util/profile_edit.js
annagav: I moved `fromat` into the if statement below.

noisecapella:
If the user clicks the add button, it will add a new education which is being edited. However when the user cancels the dialog this education is left behind in the profile edit. Can you add code to remove the newly added dialog if necessary?

annagav:
@noisecapella I added code to remove new education that has not been saved, when the dialog closes.

static/scss/layout.scss
```diff
@@ -48,3 +48,8 @@ body {
...
+.profile-tab-grid {
+      max-width: 900px;
+   border: 1px grey solid;
```
noisecapella: Is this CSS only for debugging?

annagav: not really. It is part of the mockup styles :)

noisecapella:
The education item is deleted correctly when I cancel the dialog for a new education, but this also happens for an existing education.

Since each item on the education page is edited via dialog box, maybe you could just reset the `profile.edit` if the dialog box is canceled? This would also undo changes in an edited dialog which is canceled

profiles/serializers_test.py
```diff
@@ -51,10 +61,14 @@ def test_full(self):  # pylint: disable=no-self-use
             'birth_city': profile.birth_city,
             'preferred_language': profile.preferred_language,
             'pretty_printed_student_id': profile.pretty_printed_student_id,
+            'edx_level_of_education': profile.edx_level_of_education,
+            'education': [
+                EducationSerializer().to_representation(education) for education in profile.education.all()
+                ],
```
noisecapella: Can you change the indentation here and in `work_history` below to remove four spaces, so that the `]` lines up with `'education'`?

static/js/actions/index.js
```diff
@@ -57,10 +61,20 @@ export const updateProfile = profile => ({
...
+export const openEducationForm = (level, index) => ({
+  type: SHOW_EDUCATION_FORM_DIALOG,
+  payload: { level, index }
+});
+export const closeEducationForm = () => ({type: HIDE_EDUCATION_FORM_DIALOG});
 
 export const startProfileEdit = () => ({ type: START_PROFILE_EDIT });
 export const clearProfileEdit = () => ({ type: CLEAR_PROFILE_EDIT });
 
+export const toggleEducationLevel = newState => ({
```
noisecapella: Can you rename `newState` to `educationLevels` here? You will also need to change the reducer accordingly

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,240 @@
...
+        </Grid>;
+      }
+    });
+  }
+
+  openEditEducationForm(level, education_id){
+    const { dispatch, profile } = this.props;
+
+    let index = profile['education'].findIndex((education) => {
+      return ('id' in education) && education_id === education.id;
```
noisecapella: We shouldn't need the `('id' in education)` part here. `education_id` will always be a number, so the `education_id === education.id` comparison should only be true if `education.id` is also a number

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,240 @@
...
+    let newState = Object.assign({}, educationLevels);
+    newState[level] = !educationLevels[level];
+    dispatch(toggleEducationLevel(newState));
+  }
+
+  handleCloseDialog(){
+    const { dispatch, updateProfile, profile, educationDialog} = this.props;
+    dispatch(closeEducationForm());
+    let clone = Object.assign({}, profile);
+    clone['education'].splice(educationDialog.educationIndex, 1);
```
noisecapella: We can't use mutable operations on objects that come from `props` since React requires that we treat them as immutable. `Object.assign` only does a shallow clone above which means `clone['education']` is still the same object as `profile['education']'. You will need to make a copy of the list first or use `_.cloneDeep` above. Although see my previous comment about reverting `profile.edit` entirely.

static/js/reducers/index.js
```diff
@@ -63,6 +68,46 @@ export const ui = (state = INITIAL_UI_STATE, action) => {
...
+  }
+};
+let education_level_state = {};
+education_level_state[HIGH_SCHOOL] = true;
+education_level_state[ASSOCIATE] = true;
+education_level_state[BACHELORS] = true;
+education_level_state[MASTERS] = false;
+education_level_state[DOCTORATE] = false;
+
+export const INITIAL_EDUCATION_LEVEL_STATE = education_level_state;
```
noisecapella: FYI, `const` only applies to the reference to the variable, not the contents, so you don't need the `education_level_state` here (feel free to leave it though)

annagav:
@noisecapella All comments addressed.

static/js/components/EducationTab.js
```diff
@@ -0,0 +1,239 @@
...
+      </Grid>;
+    }
+    return null;
+  }
+
+  printExistingEducations(level) {
+    const {profile} = this.props;
+    return profile['education'].map(education => {
+      if (education.degree_name === level.value && 'id' in education) {
+        return <Grid className="existing-education-grid">
```
noisecapella: React is complaining about a lack of a `key` attribute here in the console output. Can you add one at this point using `education.id`?

annagav:
@noisecapella I added a key to Grid.

static/js/components/EducationTab.js
```diff
@@ -105,7 +105,7 @@ class EducationTab extends ProfileTab {
     const {profile} = this.props;
     return profile['education'].map(education => {
       if (education.degree_name === level.value && 'id' in education) {
-        return <Grid className="existing-education-grid">
+        return <Grid key={"education-row-" + education.id} className="existing-education-grid">
```
noisecapella: For future reference you only need `education.id` here since keys only need to be unique compared to other keys in the same collection. It's not like the `id` attribute which needs to be unique across all elements in the browser

noisecapella:
Ok I believe that should be it! Squash and merge :+1:

PR #246, opened 2016-04-29
Fixed JS console error with react-router history

aliceriot:

none

#### What's this PR do?

React router throws up a console warning if you grab `history` out of
`this.props` instead of importing the history provider (e.g.
`browserHistory` or `hashHistory`) and calling `.push` on that.

This just makes that change, so the warning doesn't show up!

`static/js/containers/App.js` and `static/js/containers/TermsOfServicePage.js`

- update your profile so that `agreed_to_terms_of_service == False`
- visit `/dashboard`
- click 'agree'
- confirm no JS console warning

original warning (can repro by following the 'manual testing' steps on master):

react-router-error

### DIFF
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -58,13 +58,13 @@ class App extends React.Component {
   }
 
   requireTermsOfService() {
-    const { userProfile, location: { pathname }, history } = this.props;
+    const { userProfile, location: { pathname } } = this.props;
     if (
       userProfile.getStatus === FETCH_SUCCESS &&
       !userProfile.profile.agreed_to_terms_of_service &&
       !(TERMS_OF_SERVICE_REGEX.test(pathname))
     ) {
-      history.push('/terms_of_service');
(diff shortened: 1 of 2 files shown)


noisecapella:
Unfortunately the tests rely on being able to provide a history object that isn't a browser history. Did you investigate the `context.router` suggestion in the error?

aliceriot:
I did not, I'll take a look

aliceriot:
@noisecapella yeah, I have no idea how to get around that, haha

we can just delete this if we're fine with living with the warning

noisecapella:
It would be nice to fix this eventually but it's not blocking anything right now

aliceriot:
@noisecapella I just took another stab at this, was following along with the docs here: https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md#navigating-inside-deeply-nested-components

seems to be working ok! tests are passing too.

noisecapella:
That's great! :+1:

PR #247, opened 2016-04-29
Made `npm test` run the tests

aliceriot:

none

#### What's this PR do?

^^ what it says!

Pull it down and make sure it runs the tests!

I moved the JS testing into `package.json` a while ago, and I thought this was working then, but I just tried it and it is not.

This makes it work, so you can do:

```
docker-compose run web npm test
```

to run the JS tests.


noisecapella:
:+1:

Issue #217, opened 2016-04-26
Profile: update personal info form styling to match current mock-ups


pdpinch:
Make https://micromasters-ci.herokuapp.com/profile/personal look like https://projects.invisionapp.com/share/AF6XG4X62#/screens/153562859_Step_One_-_Personal_Info

<img width="686" alt="screen shot 2016-05-18 at 2 14 02 pm" src="https://cloud.githubusercontent.com/assets/430126/15370050/2a4c9d34-1d03-11e6-83cc-c851f3b1c123.png">

PR #250, opened 2016-04-29
Added field for state or territory

### LINKED ISSUE #229: discovery: find a python library for countries & states/provinces
bonus if we get cities too. 

django-countries maybe?

this is implemented in the front end though

noisecapella:
Fixes #229 

#### What's this PR do?
Adds a field to generate a list of states or territories based on the selected country

`static/js/util/profile_edit.js`

Go to the personal tab and select a country. The currently selected state should be cleared and the user should be able to select from a list of territories for that country.


noisecapella:
This should be done but will require some revision after the employment tab PR merges so I'll wait for that to merge first

static/js/countries.js
```diff
@@ -1,999 +0,0 @@
-// from http://data.okfn.org/data/core/country-list
-export default [
```
aliceriot: code deletion is a beautiful thing

static/js/util/ProfileTab.js
```diff
@@ -16,14 +19,16 @@ class ProfileTab extends React.Component {
...
     this.boundDateField = boundDateField.bind(this);
     this.editProfileObjectArray = editProfileObjectArray.bind(this);
 
     // options we set (for select components)
-    this.countryOptions = COUNTRIES.map(country => ({
-      value: country.Code,
-      label: country.Name
+    let countryOptions = Object.keys(iso3166.data).map(code => ({
+      value: code,
+      label: iso3166.data[code]['name']
```
aliceriot: I think this could be simplified to:

```js
Object.keys(iso3166.data).map(code => (
  { value: code, label: iso3166.data[code]['name'] }
));
```

style nit really, so feel free to ignore this, I have a perchant for one-line methods (holdover from being a rubyist :D)

static/js/components/PersonalTab.js
```diff
@@ -57,13 +57,14 @@ class PersonalTab extends ProfileTab {
       )}<br />
       <h4>Where do you live?</h4>
       {this.boundTextField(['city'], 'City')}<br />
-      {this.boundTextField(['state_or_territory'],'State or Territory')}<br />
+      {this.boundStateSelectField(['state_or_territory'], ['country'], 'State or Territory')}<br />
```
aliceriot: I think it might make sense to move the state select after the country select, otherwise if you don't have a country selected you see this:

states_and_territories

static/js/components/EmploymentTab.js
```diff
@@ -85,7 +85,7 @@ class EmploymentTab extends ProfileTab {
           {this.boundTextField(keySet('city'), 'City')}
         </Cell>
         <Cell col={12}>
-          {this.boundTextField(keySet('state_or_territory'), 'State or Territory')}
+          {this.boundStateSelectField(keySet('state_or_territory'), keySet('country'), 'State or Territory')}
```
aliceriot: same comment as below, re: moving state after country

static/js/components/PersonalTab.js
```diff
@@ -57,13 +57,14 @@ class PersonalTab extends ProfileTab {
...
       {this.boundSelectField(['country'], 'Country', this.countryOptions)}<br />
       <h4>Where were you born?</h4>
       {this.boundTextField(['birth_city'], 'City')}<br />
       {this.boundTextField(['birth_state_or_territory'], 'State or Territory')}<br />
```
aliceriot: do we also want to change this to use the new state and country fields?

aliceriot:
couple comments, back to you

static/js/components/PersonalTab.js
noisecapella: I agree, but can we postpone this until https://github.com/mitodl/micromasters/issues/217?

static/js/components/PersonalTab.js
noisecapella: Yes, I missed this one

pdpinch:
ISO_3166-2 is just country and region (state / province / territory) right?

Is there anything comparable that includes cities?

noisecapella:
There's UNLOCODE, however we would be bundling a large amount of information with the javascript at that point, roughly 10mb just for the united states: http://www.unece.org/fileadmin/DAM/cefact/locode/us.htm

static/js/util/ProfileTab.js
noisecapella: I generally like to put object literals with more than one item on multiple lines, but no reason other than style preference :)

noisecapella:
Accidentally force pushed, but the only thing I changed was using the state select field with `birth_state_or_territory`

static/js/components/PersonalTab.js
aliceriot: yah that makes sense, we'll be moving things around then anyway

aliceriot:
:+1:

static/js/components/PersonalTab.js
pdpinch: In my fantasy, the "State of Territory" component wasn't displayed at all until a country was chosen. But I was assuming that we couldn't get a complete set of states for all countries.

pdpinch:
10MB is definitely more than I want to be sending down the wire. Since #229 was a discovery story, it would have been good to note that info there before we got to implementation. But I know you guys are moving fast.

static/js/components/PersonalTab.js
aliceriot: I think that wouldn't be unreasonable behavior to add in later

PR #251, opened 2016-04-29
Removed unused authentication state, use profile state instead for name

noisecapella:
#### What's this PR do?
Removes the unused `authentication` reducer, and uses the name from the profile instead. Now the name updates in the header immediately after the profile is saved

### DIFF
--- a/static/js/containers/LoginButton.js
+++ b/static/js/containers/LoginButton.js
@@ -1,3 +1,4 @@
+/* global SETTINGS: false */
 import React from 'react';
 import { connect } from 'react-redux';
 import DropdownButton from 'react-bootstrap/lib/DropdownButton';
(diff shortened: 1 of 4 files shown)


aliceriot:
Would it also make sense to change the display in `static/js/components/Dashboard.js`? 

In the "card-name" div we currently show `{ SETTINGS.name }`.

static/js/containers/LoginButton.js
```diff
@@ -34,12 +35,12 @@ class LoginButton extends React.Component {
...
 LoginButton.propTypes = {
   dispatch: React.PropTypes.func.isRequired,
-  authentication: React.PropTypes.object.isRequired
+  userProfile: React.PropTypes.object.isRequired
 };
 
 const mapStateToProps = (state) => {
   return {
-    authentication: state.authentication
+    userProfile: state.userProfile
```
aliceriot: could we change this to `profile: state.userProfile.profile`? if we only need things on the profile object (`profile.preferred_name`) this might be cleaner.

noisecapella:
Sounds good, changes are made

aliceriot:
looks good! :+1:

PR #253, opened 2016-05-02
Removed calls to programs and courses APIs in frontend

### LINKED ISSUE #237: Remove /api/v0/courses and /api/v0/programs
And related react work to fetch from these APIs. We get everything from /api/v0/dashboard/ now.

If we're concerned we may need this in the future we should at least get rid of the code on the frontend side which fetches from these APIs

noisecapella:
Fixes #237 

#### What's this PR do?
Removes unused calls to `/api/v0/programs/` and `/api/v0/courses/` from the frontend

Everything should work the same as before, but there should not be these requests in the Network tab

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -3,12 +3,6 @@ import _ from 'lodash';
 import * as api from '../util/api';
 import * as util from '../util/util';
 
-// course list actions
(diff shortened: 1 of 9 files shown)


giocalitri:
I think this is good to go 👍

Issue #223, opened 2016-04-26
Add course not-passed to Dashboard UI


pdpinch:
related to the API story, #219 

Need UI design before we estimate.

giocalitri:
the new UI should take in account the following ideas among all:
- in the program each line will represent always a course (and not a course run)
- there will be a collapsable list under each course with the runs and their statuses
- the actions at the level of the course will be  related only to the future or current enrolled run

PR #254, opened 2016-05-02
Old not passed courses in dashboard API

### LINKED ISSUE #219: Add "not passed" courses to student dashboard API
As a micromasters student I want to see courses that I enrolled in but didn't pass on my student dashboard. 

After discussing the issue at the Product Council, it is clear that we do need to display failed courses, so students know what has happened.

giocalitri:
fixes #219 

#### What's this PR do?
the old not-passed runs now are added to the list of runs for a course together with the next one to take or with the current one

dashboard/api.py

the rest api interface should show multiple runs. the additional ones can be only `passed` or `not-passed`

the UI will most likely not work properly until #223 is done

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -93,6 +93,13 @@ class CourseRunUserStatus():
         self.course_run = course_run
         self.enrollment_for_course = enrollment_for_course
 
+    def __repr__(self):
+        return "<CourseRunUserStatus for course {course} status {status} at {address}>".format(
+            status=self.status,
+            course=self.course_run.title if self.course_run is not None else '"None"',
+            address=hex(id(self))
+        )
+
 
 def get_info_for_program(program, user, enrollments, certificates):
     """
@@ -135,6 +142,7 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
     Returns:
         dict: dictionary representing the course status for the user
     """
+    # pylint: disable=too-many-statements
     # data about the course to be returned anyway
     course_data = {
         "id": course.pk,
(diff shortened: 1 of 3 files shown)


dashboard/api.py
```diff
@@ -93,6 +93,13 @@ def __init__(self, status, course_run=None, enrollment_for_course=None):
         self.course_run = course_run
         self.enrollment_for_course = enrollment_for_course
 
+    def __repr__(self):
+        return "<CourseRunUserStatus for course {course} status {status} at {address}>".format(
+            status=self.status,
+            course=self.course_run.title if self.course_run is not None else '"None"',
```
noisecapella: nit: Should `"None"` have quotes around it?

dashboard/api.py
```diff
@@ -212,7 +248,43 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
...
+                # in this case the user might have passed the course also in the past
+                cert = user_certificates.get_verified_cert(run_status.course_run.edx_course_key)
+                course_data['runs'].append(
+                    format_courserun_for_dashboard(
+                        run_status.course_run,
+                        CourseStatus.PASSED,
+                        cert,
+                        position=index
+                    )
+                )
```
noisecapella: Is it intentional that we are not doing `index += 1` here also?

dashboard/api.py
```diff
@@ -212,7 +248,43 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
...
+        if next_run is None or run_status.course_run.pk != next_run.pk:
+            course_data['runs'].append(
+                format_courserun_for_dashboard(
+                    run_status.course_run, CourseStatus.NOT_PASSED, position=index)
+            )
+            index += 1
+
+    # add all the other runs with status != NOT_ENROLLED
+    # the first one (or two in some cases) has been added with the logic before
+    for run_status in run_statuses:
```
noisecapella: Are we doing this logic here so that they are always after the other courseruns in the list?

noisecapella:
Just some questions, functionality looks good

dashboard/api.py
giocalitri: It was to have the string with the quotes in the `repr` string, this is a weird case and I wanted to make it more visible.

giocalitri: I can remove them, no strong opinion

dashboard/api.py
giocalitri: yes: another option could be to have the not-passed in a different list... I am actually struggling with this...
what do you think?

dashboard/api.py
giocalitri: there is one for both cases at the end of the `if/else`

dashboard/api.py
noisecapella: No, it's fine, feel free to leave it as is

dashboard/api.py
noisecapella: Putting the not-passed in a different list might be better so that we group together the different cases. At the end we can just concat the two lists together

dashboard/api.py
noisecapella: If we are just updating `index` to keep up with the length of `course_data['runs']` why not have `position=len(course_data['runs']) + 1` for each case here?

giocalitri: sounds like a good idea: will do

dashboard/api.py
giocalitri: ok, so for a course we will have `runs` that will contain the current or the future runs and `runs-notpassed` (replace with anything you like more) win the `not-passed` ones. Is this ok?

noisecapella: Sounds good

giocalitri: will do then

dashboard/api.py
```diff
@@ -172,27 +172,31 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
...
     run_statuses.remove(run_status)
 
-    # an index to sort the runs for a course
-    index = 1
-    if run_status.status in (CourseRunStatus.NOT_ENROLLED, CourseRunStatus.NOT_PASSED):
+    if run_status.status == CourseRunStatus.NOT_ENROLLED:
         next_run = course.get_next_run()
         status = CourseStatus.OFFERED if next_run is not None else CourseStatus.NOT_OFFERED
         course_data['status'] = status
         if next_run is not None:
```
noisecapella: It seems like the `next_run is None` case is removed here, is it handled elsewhere? Code coverage was good before so I assume it is but I wanted to double check

noisecapella:
Nevermind I think I was looking at a weird diff :+1:

PR #260, opened 2016-05-03
Added semicolon enforcement rule

noisecapella:
#### What's this PR do?
Enforces that semicolons should always be used, and adds an unreachable code rule to guard against the case mentioned here: http://eslint.org/docs/rules/semi

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,7 +24,9 @@
       "vars": "local",
       "argsIgnorePattern": "action"
     }],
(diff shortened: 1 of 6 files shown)


aliceriot:
:+1:

aliceriot:
I'm on the 'using semicolons is not a big deal' side of the semicolon debate, personally. I've never really understood why people object to it...

PR #263, opened 2016-05-03
Moved stage1 plugin into production

noisecapella:
#### What's this PR do?
 - Moves the stage1 babel requirement from devDependencies to dependencies so production builds can use it
 - Adds `--bail` to webpack command so it returns a non-zero status code on error (surprisingly not the default behavior)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "devDependencies": {
     "babel-eslint": "5.0.0",
     "babel-plugin-transform-react-jsx": "6.6.4",
-    "babel-preset-stage-1": "6.5.0",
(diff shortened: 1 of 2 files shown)


giocalitri:
details for the change in `webpack_if_prod.sh`?

noisecapella:
The `--bail` command causes webpack to exit with a non-zero code if there's an error. Since it didn't do this before we were unaware of the requirements error. Some context: https://github.com/webpack/webpack/issues/708#issuecomment-73857915

giocalitri:
then it is 👍

PR #264, opened 2016-05-03
Upgraded npm packages

noisecapella:
#### What's this PR do?
 - Upgrades all NPM packages to latest available (for example React 15)
 - Removes redux-devtools and replaces it with configuration to allow use with the redux devtools chrome extension

Make sure everything functions as it did before, and the look and feel is the same.


aliceriot:
two new errors on `/profile`, possibly related to what I have saved, but these weren't happening before:

errors_on_profile

stack traces aren't super helpful about exactly where this is coming from exactly.

aliceriot:
things I've confirmed are working:

- [ ] eslint
- [ ] react-router transitions (with no warnings)
- [ ] redirection to the terms of service page
- [ ] form validation
- [ ] JS test suite
- [ ] logout button

noisecapella:
It looks like a legit warning due to React 15. We are passing `undefined` to the `value` prop of the `TextField` and React is saying that by doing that we're indicating that the component is uncontrolled, as if we weren't setting `value` at all. Should be fixed now!

static/js/containers/LoginButton.js
```diff
@@ -9,9 +9,16 @@ class LoginButton extends React.Component {
   render() {
     const { profile } = this.props;
 
+    // span tags are a workaround for weird indentation with react-bootstrap
+    // and React 15. React 15 removed span tags but react-bootstrap still expects
+    // them.
```
aliceriot: oh react-bootstrap, you're so fun

aliceriot:
cool, I think this is :ship: shape! :+1:

Issue #259, opened 2016-05-03
Change yes/no values being sent from employment UI


noisecapella:
The backend should handle true/false values here, so at some point we should translate the yes/no in the state to true/false, either in the API function or maybe `boundSelectField`

For #150

pdpinch:
We don't need to do this. If the user has no employment history, we won't store anything.

PR #271, opened 2016-05-04
Added check for profile completeness before visiting dashboard

### LINKED ISSUE #227: require user profile be complete before allowing user to view dashboard
After some further thought, this should depend on checking a value from the REST API, filled_out

### LINKED ISSUE #286: Profile: return user to point they left off if their enrollment was interrupted
As a user I may not complete my profile in on session. When I return, I need to be directed to the sections that I still need to complete.

aliceriot:

closes #286 

#### What's this PR do?

This makes it so that users are directed to fill out unfilled portions of their profile before they can interact with the rest of the dashboard.

There's a utility function (`profileComplete`) defined in `util.js`. This returns `true` if a profile is complete, and else it returns an object with messages about what it is incomplete (in general).

We also have a function to check completion (in `containers/App.js`) which runs on `componentDidMount` and `componentDidUpdate`, and it follows this logic:

```
if (
  good profile &&
  already did TOS &&
  not already in /profile && 
  profile isn't complete
) {
  // do some stuff
}
```

so it plays nicely with the TOS page, if a user has both an incomplete profile and has not yet agreed to the terms of service they will be redirected first to the TOS page, and then after agreeing to that they will be redirected to the profile page.

The `profileComplete` utility function returns something that looks like this:

```
{
  url: "/profile/personal",
  title: "Personal Info",
  text: "Please complete personal info"
}
```

where the `url` is where we want to redirect, and the `title` and `text` fields are used to show the user a dialog box. I added a few new actions and a new reducer to handle this (which should be useful in a more general sense for other dialog boxes we might want to show).

Returning the url here means that we can first check the `personal` tab, and if that isn't valid redirect there, then go on to check the professional tab, etc. 

I guess with the new reducers / redux stuff changes - this is what I feel least confident about (is `ui` the right name for a top-level key? what would be better?).

Try interacting with it in a couple of different ways. If a field is incomplete in the profile it should be redirecting there in all circumstances (i.e. you shouldn't be able to click away from the profile). It shouldn't redirect to the profile if the user hasn't yet accepted the ToS, however.


static/js/containers/App.js
```diff
@@ -9,27 +17,49 @@ import {
...
   clearDashboard,
+  clearUI,
+  updateDialogText,
+  updateDialogTitle,
+  setDialogVisibility,
 } from '../actions/index';
+import { profileComplete } from '../util/util';
 
 const TERMS_OF_SERVICE_REGEX = /\/terms_of_service\/?/;
+const PROFLE_REGEX = /^\/profile\/?[a-z]?/;
```
noisecapella: `PROFILE`

aliceriot: is the regex pattern insufficient? I'm kind of bad at regex, haha

noisecapella: I mean you misspelled profile

static/js/containers/App.js
```diff
@@ -57,6 +87,48 @@ class App extends React.Component {
...
+      dispatch(setDialogVisibility(true));
+      this.context.router.push(url);
+    }
+  }
+
+  showDialog () {
+    const { ui, dispatch } = this.props;
+    let visible = _.get(ui, ['dialog', 'visible']);
+    let text = _.get(ui, ['dialog', 'text']);
+    if ( visible == true && text != "" ) {
```
noisecapella: Can we use triple equals here? I thought we checked this in the linter but I'm not sure

static/js/containers/App.js
aliceriot: ah ok that makes more sense now

noisecapella:
Can you add tests to reducers/index_test.js for the new actions?

aliceriot:
@noisecapella sure, I also just added a commit turning on the eslint rule to require `===` (which for some reason is called `eqeqeq`): http://eslint.org/docs/rules/eqeqeq

aliceriot:
also I realized I didn't let amir know about the `currently_employed` field, so I'm adding that here (we need to have it so we can decide whether to validate for `work_history`).

aliceriot:
ah just got the joke, `eqeqeq` is triple-eq. good one, eslint people!

profiles/models.py
```diff
@@ -142,6 +148,12 @@ class Profile(models.Model):
     date_joined_micromasters = models.DateTimeField(blank=True, null=True, auto_now_add=True)
     linkedin = JSONField(blank=True, null=True)
     student_id = models.IntegerField(blank=True, null=True, unique=True)
+    currently_employed = models.CharField(
```
noisecapella: We should store this as a boolean and update our frontend to send this value as a boolean also (#259). Can you revert this and we can do it in a separate PR?

aliceriot: sure

static/js/containers/App.js
```diff
@@ -57,6 +87,48 @@ class App extends React.Component {
...
+    ) {
+      const { url, title, text } = profileCompleteCheck;
+      dispatch(updateDialogText(text));
+      dispatch(updateDialogTitle(title));
+      dispatch(setDialogVisibility(true));
+      this.context.router.push(url);
+    }
+  }
+
+  showDialog () {
```
noisecapella: We should always show the dialog and use the `open` prop of `Dialog` to say where it is open or not. Also, we will need to use `dialogPolyfill` here to support Firefox and IE

aliceriot: yeah, I saw the other PR, was waiting for it to be merged.

cool, I'll remove this line:

```
if ( visible === true && text !== "" ) {
````

File: static/js/util/util.js [hunk 5]
```diff
@@ -191,6 +193,42 @@ export function validateProfile(profile, requiredFields, messages) {
...
 
 /* eslint-enable camelcase */
+/*
+check that the profile is complete. we make the assumption that a
+complete profile consists of:
+  - a valid personal tab
+  - an entry for 'currently employed', and a work history entry if
+    `currently employed == 'yes'`
+*/
+export function profileComplete(profile) {
```
noisecapella: Can you rename this `validateProfileComplete`? Can you also return `false` instead of `true`, since the object returned is truthy when there are validation errors?

aliceriot: can do the rename, I'm not sure about returning `false` in the case where there are no errors.

what if I changed the return values to look like `[bool, {...}]`, so if we get all the way to the end the return value is `[true, {}]`, and if we have an error it looks like `[false, { url: "profile/", ... } ]`?

aliceriot: alternatively I could just return `{}` in the 'no errors' case...hrm

noisecapella: I like the tuple idea, it's nice and explicit

aliceriot: cool :) I've been writing a little Go lately for some personal projects, that sort of thing seems like a common idiom in Go-land.

giocalitri:
@aliceriot what's the status of this story?

aliceriot:
@giocalitri working on review comments!

giocalitri:
I was testing functionality and I see a problem:
if the profile is not complete and I type `/dashboard` the dashboard page starts loading and then I am redirected to the profile.
This looks bad: maybe we should add something like a loader to not show the page until it is fully loaded.
Not for this PR but we should probably open a PR.
@pdpinch ?

giocalitri:
Again probably this is not a bug of this PR itself, but let's say I go to the profile: how am I supposed to know what is mandatory to be filled? If I skip state/territory, I can go through, for example.
Another issue probably.

giocalitri:
bug:
if I go to the professional tab and I select "No" if I am currently employed and I try so save, I get redirected to the current tab and my choice is lost.

giocalitri:
the redirect functionality by itself seems to work fine, but I cannot have a final opinion because of the last bug I reported

aliceriot:
Yeah, the currently employed tab doesn't save. If you look through the
commit history for this PR there's some discussion about it.

giocalitri:
okay... which story should take care of this?

aliceriot:
I don't think there is one for it yet - I can file a new issue.

aliceriot:
What about putting the unfilled fields into the popup message? I considered adding this functionality but thought it might make the popup too busy. It would be trivial to add that feature though - I'm using the same code we wrote for validating fields on submit to check that the profile is complete.

giocalitri:
for the unfilled fields, I think it is not up to us engineer to solve: it is an usability problem and I have no idea what's the best approach: just open an issue for it

aliceriot:
ok, will do

giocalitri:
for the profile page, I am bit puzzled: which story was covering this part? this should be part of that

aliceriot:
@giocalitri the changes we discussed this morning have been added, so I think it's ready to be re-reviewed.

giocalitri:
will review immediately

giocalitri:
I think this is ok to go 👍 
There will be another story to cover what is missing

giocalitri:
please squash

PR #273, opened 2016-05-04
Updated profile PATCH code to use return value as new profile

### LINKED ISSUE #272: After PATCH the employment id should be saved to the profile
After a new employment is PATCHed, the PATCH returns an updated profile. We should take this updated profile and replace it in the redux state

noisecapella:
Fixes #272

#### What's this PR do?
Instead of assigning the profile in edit to be the profile on successful PATCH, use the return value from the PATCH. This is necessary so we can keep track of the id number

Add a new work history item to your profile and save it. Then, without refreshing, look at your redux state and verify that the work history item in your profile has an id number.

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -54,7 +54,7 @@ export const saveProfile = (username, profile) => {
   return dispatch => {
     dispatch(requestPatchUserProfile());
     return api.patchUserProfile(username, profile).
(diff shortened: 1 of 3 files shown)


aliceriot:
seems to be working correctly so far!

work_history_ids

aliceriot:
:+1:

PR #279, opened 2016-05-05
Added Privacy tab to profile

### LINKED ISSUE #215: Add a public/private option to the user profile
As a micromasters student, I want to be able to control who can see my profile information. In particular, I may want to share my professional experience with MIT, but I may not want my employer to know that I'm seeking further education. 

Add a field to the profile form. We might as well follow the same format as edX, i.e. 

"Micromasters learners can see my: " ("public to the world", "public to other micromasters students" "private")

We will share private profiles with MIT faculty. We will share "public to other micromasters students" only to logged in users. "public to the world" will be shared with all. 

Default to private. 

Until we have views that show other learners, this won't have any impact, but we need to collect the data at registration. 

Because this is relevant to whole profile, it needs to be a final step, i.e. "Profile Privacy"

annagav:
Fix #215 
#### What's this PR do?
Add Privacy tab to profile. Set account privacy default value to be 'private'.
PrivacyTab.js
Go to Privacy tab. For a newly created account the value should be set to 'private'. For already existing accounts the value was set from the edx profile on creation of the mm account, and is public by default.

### DIFF
--- a/static/js/components/PrivacyTab.js
+++ b/static/js/components/PrivacyTab.js
@@ -22,7 +22,7 @@ class PrivacyTab extends ProfileTab {
       ['account_privacy']
     ],
     validationMessages: {
-      'account_privacy': 'Privacy'
+      'account_privacy': 'Privacy level'
     }
   };
 
--- a/static/js/components/PrivacyTab.js
+++ b/static/js/components/PrivacyTab.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import Button from 'react-mdl/lib/Button';
+
+import ProfileTab from "../util/ProfileTab";
+import { saveAndContinue } from "../util/profile_edit";
+
+class PrivacyTab extends ProfileTab {
+  constructor(props) {
+    super(props);
+    this.saveAndContinue = saveAndContinue.bind(this, '/dashboard');
+  }
+
+  static propTypes = {
+    profile:        React.PropTypes.object,
+    errors:         React.PropTypes.object,
+    saveProfile:    React.PropTypes.func,
+    updateProfile:  React.PropTypes.func
+  };
+
+  static defaultProps = {
+    requiredFields: [
+      ['account_privacy']
+    ],
+    validationMessages: {
+      'account_privacy': 'Privacy'
+    }
+  };
+
+  render() {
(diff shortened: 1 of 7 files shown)


static/js/util/profile_edit.js
```diff
@@ -50,7 +50,7 @@ export function boundSelectField(keySet, label, options) {
   } = this.props;
   let onChange = value => {
     let clone = _.cloneDeep(profile);
-    _.set(clone, keySet, value? value.value : null);
+    _.set(clone, keySet, value? value.value : '');
```
aliceriot: was this causing a console warning? I know we had an issue with the `TextField` complaining when `this.props.value == undefined`.

static/js/components/PrivacyTab.js
```diff
@@ -0,0 +1,41 @@
...
+    saveProfile:    React.PropTypes.func,
+    updateProfile:  React.PropTypes.func
+  };
+
+  static defaultProps = {
+    requiredFields: [
+      ['account_privacy']
+    ],
+    validationMessages: {
+      'account_privacy': 'Privacy'
```
aliceriot: 'Privacy' seems a little vague as a validation message to me, maybe something 'Privacy Level'  instead?

aliceriot:
I think this looks good! I'm psyched about how little code we need to write that's specific to each one of these tabs.

I had a couple of questions / comments, back to you.

static/js/util/profile_edit.js
annagav: When I was trying to click save and continue with an empty value, it would show the error but then immediately reload.
Also, in my other PR, George mentioned something about react complaining about null values.

static/js/components/PrivacyTab.js
annagav: Sounds good :)

static/js/util/profile_edit.js
aliceriot: cool, I was just wondering what was going on. react, so tricky sometimes...

aliceriot:
:+1:, assuming tests are passing

annagav:
@aliceriot Thank you!

Issue #226, opened 2016-04-26
Add button and note for applying to masters program to student dashboard


pdpinch:
For MVP, this should just be an inactive button -- so users know that it's coming later. It should look like https://projects.invisionapp.com/share/AF6XG4X62#/screens/142338434

For the future:

There should be a message, "You need to complete X more courses before you can apply for a masters" 

X = (total number of courses in the program minus the user's course with status PASSED)

button is inactive until user has passed all courses.

pdpinch:
I don't think this is necessary for MVP, @Ferdi ?

Ferdi:
A static , inactive button that does nothing would be nice. To give the impression to users that we are working on it

pdpinch:
I update the description to make the expectation for MVP more clear.

amir-qayyum-khan:
Done with changes in #282 

<img width="1067" alt="screen shot 2016-05-06 at 8 42 16 pm" src="https://cloud.githubusercontent.com/assets/10431250/15078203/79c1685e-13cb-11e6-82e5-4ac25bff0d7b.png">

Issue #162, opened 2016-04-11
Make enrollment and upgrade links active


pdpinch:
Provide the same endpoints as edX does for enrollment and upgrade to verified. 

All we need is a course key and a logged in user.

pdpinch:
For "enroll" we can use the enrollments API: http://edx.readthedocs.org/projects/edx-platform-api/en/latest/enrollment/enrollment.html#view-and-add-to-a-users-course-enrollments This supports a POST, so it looks like its easiest and best to enroll the user without leaving micromasters. 

For "upgrade to verified" we have no choice but to drop the user into the edX shopping cart flow. The URLs look like this: https://courses.stage.edx.org/course_modes/choose/course-v1:ASUx+CreditDemo+Demo_Course/ (See ### for discussion).

pdpinch:
In the interests of getting this done, we can implement the enroll button as a link instead of a POST for now also.

amir-qayyum-khan:
@pdpinch can i get ui image where to add these links?

pdpinch:
See https://invis.io/UY77QX9N4#/142338434_Student_-_Dashboard

PR #282, opened 2016-05-06
Added dashboard links to enroll and upgrade courses, and a disabled button to apply for masters on campus

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/226
fixes https://github.com/mitodl/micromasters/issues/162
#### What's this PR do?

- Added button to apply for Master's in the bottom of program.
- Added navigation on Enroll button and Upgrade button on courses inside program.

@pdpinch @giocalitri @aliceriot 

You can start review from dashboard listing. Then you can see utility function that creates course status.

Open dashboard page, click enroll button and upgrade button.
See disabled button at the bottom of each program which states apply for Master's in given program.

screen shot 2016-05-06 at 8 42 16 pm

### DIFF
--- a/static/js/components/CourseList.js
+++ b/static/js/components/CourseList.js
@@ -1,10 +1,12 @@
 import React from 'react';
 import _ from 'lodash';
+import Button from 'react-bootstrap/lib/Button';
 
 import {
   makeCourseStatusDisplay,
   makeCourseProgressDisplay,
 } from '../util/util';
+import { STATUS_PASSED } from '../constants';
 
 class CourseList extends React.Component {
   render() {
@@ -12,10 +14,16 @@ class CourseList extends React.Component {
 
     let programs = dashboard.programs.map(program => {
       let sortedCourses = _.sortBy(program.courses, 'position_in_program');
+      let totalCourses = program.courses.length;
+      let coursesPassed = 0;
 
       let table = sortedCourses.map((course, i) => {
         let isTop = i === 0;
         let isBottom = i === sortedCourses.length - 1;
+
+        if (course.status === STATUS_PASSED) {
+          coursesPassed++;
+        }
 
         return <ul
           key={course.id}
@@ -33,18 +41,28 @@ class CourseList extends React.Component {
         </ul>;
       });
 
-      return <div key={program.id} className="course-list">
-        <ul className="course-list-header">
-          <li className="course-title">
-            {program.title}
-          </li>
-          <li className="course-status "/>
-          <li className="course-progress">
-            Progress
-          </li>
-        </ul>
-        {table}
-      </div>;
+      return (
+        <div key={program.id}>
+          <div className="course-list">
+            <ul className="course-list-header">
+              <li className="course-title">
+                {program.title}
+              </li>
+              <li className="course-status "/>
+              <li className="course-progress">
+                Progress
+              </li>
+            </ul>
+            {table}
+          </div>
+          <div className="clear-float apply-for-ms">
+            <br/>
+            <p>You need to pass {totalCourses - coursesPassed} more courses
+              before you can apply for <strong>{program.title}</strong> Master's Degree.</p>
+            <Button bsStyle="primary" disabled>Apply for <strong>{program.title}</strong> Master Degree</Button>
+          </div>
+        </div>
+      );
     });
 
     return <div>
--- a/static/js/components/CourseList_test.js
+++ b/static/js/components/CourseList_test.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import ReactDOM from 'react-dom';
 import TestUtils from 'react-addons-test-utils';
 import sinon from 'sinon';
 import assert from 'assert';
@@ -6,7 +7,7 @@ import _ from 'lodash';
 
 import CourseList from './CourseList';
 import * as util from '../util/util';
-import { DASHBOARD_RESPONSE } from '../constants';
+import { STATUS_PASSED, DASHBOARD_RESPONSE } from '../constants';
 
 describe("CourseList", () => {
   let sandbox;
@@ -16,6 +17,44 @@ describe("CourseList", () => {
 
   afterEach(() => {
     sandbox.restore();
+  });
+
+  it('renders apply for master button and text', () => {
+    let renderedComponent = TestUtils.renderIntoDocument(
+      <CourseList dashboard={{programs: DASHBOARD_RESPONSE}} />
+    );
+    let programBottomDivComponent = TestUtils.scryRenderedDOMComponentsWithClass(
+      renderedComponent,
+      'apply-for-ms'
+    );
+    assert(programBottomDivComponent.length > 0);
+
+    let programCount = 0;
+    for (let program of DASHBOARD_RESPONSE) {
+      let totalCourses = program.courses.length;
+      let coursesPassed = 0;
+      for (let course of program.courses) {
+        if (course.status === STATUS_PASSED) {
+          coursesPassed++;
+        }
+      }
+
+      // Tests for program's `apply for master` bottom UI.
+      let coursesNeedToPass = totalCourses - coursesPassed;
+      let programBottomText = "You need to pass " + coursesNeedToPass +
+          " more courses before you can apply for " + program.title + " Master's Degree.";
+      let programBottomButtonText = "Apply for " + program.title + " Master Degree";
+
+      let div = ReactDOM.findDOMNode(programBottomDivComponent[programCount]);
+      let renderedProgramBottomMessage = div.querySelectorAll("p");
+      let renderedProgramBottomButton = div.querySelectorAll("button");
+
+      assert(renderedProgramBottomButton.length > 0);
+      assert(renderedProgramBottomMessage[0].textContent === programBottomText);
+      assert(renderedProgramBottomButton[0].textContent === programBottomButtonText);
+
+      programCount++;
+    }
   });
 
   it('renders courses using util functions', () => {
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -1,3 +1,4 @@
+/* global SETTINGS:false */
 import React from 'react';
 import ReactDOM from 'react-dom';
 import ga from 'react-ga';
@@ -50,6 +51,9 @@ export function makeCourseStatusDisplay(course, now = moment()) {
     courseOrRun = course.runs[0];
   }
 
+  let courseUpgradeUrl = `${SETTINGS.edx_base_url}course_modes/choose/${courseOrRun.course_id}/`;
+  let courseInfoUrl = `${SETTINGS.edx_base_url}courses/${courseOrRun.course_id}/info`;
+
   switch (courseOrRun.status) {
   case STATUS_PASSED:
   case STATUS_NOT_PASSED:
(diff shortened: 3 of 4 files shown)


static/js/util/util.js
```diff
@@ -50,6 +51,38 @@ export function makeCourseStatusDisplay(course, now = moment()) {
...
+    if (courseOrRun.course_id) {
+      url.push(
+        SETTINGS.edx_base_url,
+        'courses/',
+        courseOrRun.course_id,
+        '/info'
+      );
+      window.open(url.join(""), '_blank');
+    } else {
+      alert("Please enter valid course id");
```
amir-qayyum-khan: - is there always course id?
- if not the what kind of validation message should we show?

pdpinch: A course run will always have a course id (course key). If there is no course run, there will be no course id and we shouldn't attempt to display an enroll or upgrade button.

static/js/components/CourseList.js
```diff
@@ -1,22 +1,32 @@
...
 } from '../util/util';
+import {
+  STATUS_PASSED,
+} from '../constants';
 
 class CourseList extends React.Component {
   render() {
     const { dashboard } = this.props;
+    var coursesPassed = 0;
+    var totalCourses = 0;
```
aliceriot: can you use `let` instead of `var` here? ES2015 introduced the `let` and `const` keywords for variable binding, they have better behavior variable scoping and so on: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

static/js/util/util.js
giocalitri: To be precise, if there is a status that requires to show a button, there will always be a course id available.

static/js/components/CourseList.js
```diff
@@ -1,22 +1,32 @@
...
 import _ from 'lodash';
-
+import Button from 'react-bootstrap/lib/Button';
 import {
   makeCourseStatusDisplay,
   makeCourseProgressDisplay,
 } from '../util/util';
+import {
+  STATUS_PASSED,
+} from '../constants';
```
aliceriot: I'd leave this import on one line, kind of a style nit tho

static/js/components/CourseList.js
```diff
@@ -44,6 +54,13 @@ class CourseList extends React.Component {
...
         </ul>
         {table}
+      </div>
+      <div className="clear-float">
+        <br/>
+        <p>You need to pass {totalCourses - coursesPassed} more courses
+          before you can apply for <strong>{program.title}</strong> Master's Degree.</p>
+        <Button bsStyle="primary" disabled>Apply for <strong>{program.title}</strong> Master Degree</Button>
+      </div>
       </div>;
```
aliceriot: is this an extra closing `</div>` tag?

amir-qayyum-khan: yes

aliceriot: oh I see, I missed the two opening tags on one line.

could you reformat this so it's a little more readable? mainly 1 opening `<div>` tag per line and correct indentation will help a lot.

aliceriot: additionally I prefer this style for JSX returns:

```js
return (
  <div id="foo">
    { content }
  </div>
);
```

but we seem to also use this style:

```js
return <div>
  { content }
</div>
```

so if you prefer that I think it's a matter of preference

static/js/util/util.js
aliceriot: also I think we don't want to lean on `alert` here. in https://github.com/mitodl/micromasters/commit/0e8cc743c2354a1c79d631456176e6bcd25b70ea I added support for using the redux store to set text to show in a dialog, if we really want to alert the user I think that's a better option.

amir-qayyum-khan:
Done @aliceriot @giocalitri

static/js/components/CourseList.js
```diff
@@ -1,22 +1,30 @@
 import React from 'react';
 import _ from 'lodash';
-
+import Button from 'react-bootstrap/lib/Button';
 import {
   makeCourseStatusDisplay,
   makeCourseProgressDisplay,
 } from '../util/util';
+import STATUS_PASSED from '../constants';
```
aliceriot: I think you will need to do a qualified import here, like this:

```js
import { STATUS_PASSED } from '../constants;'
```

since that module doesn't have `STATUS_PASSED` as a default export

static/js/components/CourseList.js
```diff
@@ -45,6 +53,18 @@ class CourseList extends React.Component {
         </ul>
         {table}
       </div>;
+
+      let applyForMastersUI = <div className="clear-float">
```
aliceriot: Can you leave the JSX here as one declaration, inlined in the value we `return`? there's no control flow happening here so there's not really a need to separate it out like this (ditto for `programListingUI` above)

aliceriot:
also, when responding to review comments, can you add intermediate commits instead of rebasing? it makes it easier to follow the changes. you can squash them down into one commit at merge time instead.

static/js/components/CourseList.js
amir-qayyum-khan: you want me to combine this html instead of separating it into variables?

aliceriot: yeah, I there's no need to separate it out since we're always including it anyway. if we were conditionally including e.g. the `programListingUI` JSX above it would make sense to me to separate it out, but as-is I think it decreases readability.

amir-qayyum-khan: Done

static/js/components/CourseList.js
```diff
@@ -1,22 +1,30 @@
...
 
 class CourseList extends React.Component {
   render() {
     const { dashboard } = this.props;
+    let coursesPassed = 0;
+    let totalCourses = 0;
 
     let programs = dashboard.programs.map(program => {
       let sortedCourses = _.sortBy(program.courses, 'position_in_program');
+      totalCourses = (program.courses.length > 0) ? program.courses.length: 0;
```
aliceriot: style nit: imo a JS ternary should be formatted like this:

```js
foo = bar ? baz : undefined
```

with spaces around the operators (`?` and `:`).

static/js/components/CourseList.js
```diff
@@ -1,22 +1,30 @@
 import React from 'react';
 import _ from 'lodash';
-
+import Button from 'react-bootstrap/lib/Button';
```
aliceriot: we usually keep a blank line between dependency imports (react and so on) and our own code (the `util.js` import below this)

amir-qayyum-khan: Done

aliceriot:
this LGTM :+1: :shipit: :ship:

giocalitri:
@aliceriot did you test functionality?

aliceriot:
yes, it seems to be working correctly to me.

giocalitri:
okay

amir-qayyum-khan:
@aliceriot merged commits

static/js/components/CourseList.js
```diff
@@ -1,22 +1,31 @@
...
 
 class CourseList extends React.Component {
   render() {
     const { dashboard } = this.props;
+    let coursesPassed = 0;
+    let totalCourses = 0;
 
     let programs = dashboard.programs.map(program => {
       let sortedCourses = _.sortBy(program.courses, 'position_in_program');
+      totalCourses = program.courses.length > 0 ? program.courses.length : 0;
```
noisecapella: When would `program.courses.length` not work here? It's always a number greater than 0

static/js/components/CourseList.js
```diff
@@ -33,17 +42,25 @@ class CourseList extends React.Component {
...
+            <li className="course-status "/>
+            <li className="course-progress">
+              Progress
+            </li>
+          </ul>
+          {table}
+        </div>
+        <div className="clear-float">
+          <br/>
+          <p>You need to pass {totalCourses - coursesPassed} more courses
```
noisecapella: This text is per program but `coursesPassed` will keep incrementing for every program and course. It seems like you should define these two inside the `map` function and not the `render` function

static/js/util/util.js
```diff
@@ -85,7 +110,7 @@ export function makeCourseStatusDisplay(course, now = moment()) {
 
     let verificationDate = moment(courseOrRun.verification_date);
     if (verificationDate.isAfter(now, 'day')) {
-      return <Button bsStyle="success">UPGRADE TO VERIFIED</Button>;
+      return <Button bsStyle="success" onClick={onUpgradeBtnClick}>UPGRADE TO VERIFIED</Button>;
```
noisecapella: I don't think you need `onClick` here at all, you should be able to just use `href` and `target` here

static/js/util/util.js
```diff
@@ -100,7 +125,7 @@ export function makeCourseStatusDisplay(course, now = moment()) {
     if (enrollmentDate.isAfter(now, 'day')) {
       return "Enrollment starting: " + enrollmentDate.format("M/D/Y");
     } else {
-      return <Button bsStyle="success">ENROLL</Button>;
+      return <Button bsStyle="success" onClick={onEnrollBtnClick}>ENROLL</Button>;
```
noisecapella: Remove `onClick` here please

aliceriot:
okay so nvm not ready to do :D

static/js/components/CourseList.js
amir-qayyum-khan: Sorry i misunderstood edge case, where a program has no course. Now just saw api it was returning empty course list. Now has been fixed.

amir-qayyum-khan:
@noisecapella done with fixing issues

static/js/util/util.js
```diff
@@ -46,10 +47,27 @@ function asPercent(number) {
...
 export function makeCourseStatusDisplay(course, now = moment()) {
   let courseOrRun = course;
+  let courseUpgradeUrl = [];
+  let courseInfoUrl = [];
+
   if (course.runs.length > 0) {
     courseOrRun = course.runs[0];
   }
 
+  courseInfoUrl.push(
```
noisecapella: What advantage is there to using `push` here? Since `join` is just simple string concatenation, can you use a ES6 template string here:

    let courseInfoUrl = `${SETTINGS.edx_base_url}courses/${courseOrRun.course_id}/info`;

noisecapella:
Can you add a test to `CourseList_test.js` which uses `DASHBOARD_RESPONSE` from `constants.js` and asserts the text at the bottom?

amir-qayyum-khan:
Done @noisecapella

static/js/components/CourseList_test.js
```diff
@@ -22,19 +23,31 @@ describe("CourseList", () => {
     let makeCourseProgressDisplaySpy = sandbox.spy(util, 'makeCourseProgressDisplay');
     let makeCourseStatusDisplaySpy = sandbox.spy(util, 'makeCourseStatusDisplay');
 
-    TestUtils.renderIntoDocument(
+    let renderedComponent = TestUtils.renderIntoDocument(
```
noisecapella: Please put all this in a separate test

amir-qayyum-khan: Done @noisecapella

static/js/components/CourseList.js
```diff
@@ -33,17 +40,25 @@ class CourseList extends React.Component {
...
-          <li className="course-title">
-            {program.title}
-          </li>
-          <li className="course-status "/>
-          <li className="course-progress">
-            Progress
-          </li>
-        </ul>
-        {table}
+      return <div key={program.id + "_1"}>
```
noisecapella: We only need the key attribute at the root element. Can you remove the `key={program.id}` below and change this to `key={program.id}`?

amir-qayyum-khan: Done @noisecapella

amir-qayyum-khan:
@noisecapella is PR this ok to merge?

noisecapella:
Looks good to me :+1: 

I think the wording needs some work but that's more a question for the designer

PR #298, opened 2016-05-09
Fix osx_run.sh to let tox run correctly

### LINKED ISSUE #274: Change osx_run.sh to not prefix python3 to shell commands

noisecapella:
WIP

Fixes #274

### DIFF
--- a/_osx_setup.sh
+++ b/_osx_setup.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+set -euf -o pipefail
+set -a
+
+export NODE_ENV=development
+export DEBUG=True
+export PORT=8079
+export COVERAGE_DIR=htmlcov
+export DATABASE_URL=postgres://postgres:postgres@localhost:5432/postgres
+export MICROMASTERS_USE_WEBPACK_DEV_SERVER=True
+export MICROMASTERS_SECURE_SSL_REDIRECT=False
+export MICROMASTERS_DB_DISABLE_SSL=True
+
+# Workaround for a python package complaining about uwsgi not having a compiler
+# http://stackoverflow.com/questions/11669594/error-while-installing-uwsgi-on-mac
+export CC=gcc
+
+source .env
+
+# Before this script is run make sure that:
+# - Python3 is installed from the official Python 3 installer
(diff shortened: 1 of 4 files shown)


osx_run.sh
```diff
@@ -1,53 +1,7 @@
...
 
-export NODE_ENV=development
-export DEBUG=True
-export PORT=8079
-export COVERAGE_DIR=htmlcov
-export DATABASE_URL=postgres://postgres:postgres@localhost:5432/postgres
-export MICROMASTERS_USE_WEBPACK_DEV_SERVER=True
-export MICROMASTERS_SECURE_SSL_REDIRECT=False
-export MICROMASTERS_DB_DISABLE_SSL=True
+. ./_osx_setup.sh
```
aliceriot: does this mean that `osx_run.sh` will do `npm install`, `pip install`, etc every time? it seems like that would make the command slow unnecessarily.

am I missing something, or is there a reason why `osx_run.sh` couldn't just export the necessary environment variables, source `.env`, and then run the command?

_osx_setup.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
gsidebo: i feel like the underscore should precede all of these osx_* files or none of them

osx_run.sh
gsidebo: can you explain more about the purpose of _run.sh and _start.sh? naming could probably be better, since semantically they would seem to be doing the same thing (eg: "i want to start my machine" and "i want to run my machine" mean roughly the same thing). this file just seems to be running setup then executing the user's arguments as a command. is that going to be useful to us? also, what are the use cases where a user would want to execute _start?

_osx_setup.sh
gsidebo: ok i think i see what you're trying to do here. are you trying to indicate that this file is 'private' in a sense, and shouldn't be run directly by users? personally i think it's completely valid to expect a user to run two different shell scripts. one for setup, one to start/run the machine, especially since the setup script would be need far less often than the start/run scripts

noisecapella:
Closing, we will remove these scripts in #307

PR #299, opened 2016-05-10
Implemented refresh token

### LINKED ISSUE #278: Implement OAuth refresh token
edX has updated its OAuth library and now supports refresh tokens. The new library also implements a 10 hour expiration of OAuth tokens, making refreshes more important.

giocalitri:
closes #278 
fixes mitocw/edx-platform#217

#### What's this PR do?
switches the backend to use the new edX oauth toolkit library and implements the OAUTH2 refresh token.

first `backend/pipeline_api.py`
then `backend/edxorg.py`
then `dashboard/views.py`

you need to create in your edx instance a new application in the section `django oauth toolkit`, then update `client_id` and `client_secret` in your `.env` file.

Until the current edx PR#12435 gets merged, the  endpoint for Django OAUTH Toolkit cannot be used (the certificates rest api will not work).
In the meantime it woks with the old library (that turned out to support refresh tokens as well)

### DIFF
--- a/backends/edxorg.py
+++ b/backends/edxorg.py
@@ -1,6 +1,7 @@
 """
 EdX.org backend for Python Social Auth
 """
+from datetime import datetime
 from urllib.parse import urljoin
 
 from django.conf import settings
@@ -15,11 +16,25 @@ class EdxOrgOAuth2(BaseOAuth2):
     ID_KEY = 'edx_id'
     REQUEST_TOKEN_URL = None
     EDXORG_BASE_URL = settings.EDXORG_BASE_URL
+
+    # Settings for Django OAUTH toolkit
+    # AUTHORIZATION_URL = urljoin(EDXORG_BASE_URL, '/_o/authorize/')
+    # ACCESS_TOKEN_URL = urljoin(EDXORG_BASE_URL, '/_o/token/')
+    # DEFAULT_SCOPE = ['read', 'write']
+
+    # Settings for old edx OAUTH library
     AUTHORIZATION_URL = urljoin(EDXORG_BASE_URL, '/oauth2/authorize/')
     ACCESS_TOKEN_URL = urljoin(EDXORG_BASE_URL, '/oauth2/access_token/')
+    DEFAULT_SCOPE = ['email', ]
+
     ACCESS_TOKEN_METHOD = 'POST'
     REDIRECT_STATE = False
-    DEFAULT_SCOPE = ['email']
+    EXTRA_DATA = [
+        ('refresh_token', 'refresh_token', True),
+        ('expires_in', 'expires_in'),
+        ('token_type', 'token_type', True),
+        ('scope', 'scope'),
+    ]
 
     def user_data(self, access_token, *args, **kwargs):
         """
@@ -82,3 +97,17 @@ class EdxOrgOAuth2(BaseOAuth2):
             string: the unique identifier for the user in the remote service.
         """
         return details.get(self.ID_KEY)
+
+    def refresh_token(self, token, *args, **kwargs):
+        """
+        Overridden method to add extra info during refresh token.
+
+        Args:
+            token (str): valid refresh token
+
+        Returns:
+            dict of information about the user
+        """
+        response = super(EdxOrgOAuth2, self).refresh_token(token, *args, **kwargs)
+        response['updated_at'] = datetime.utcnow().timestamp()
+        return response
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -2,6 +2,7 @@
 APIs for extending the python social auth pipeline
 """
 import logging
+from datetime import datetime
 from urllib.parse import urljoin
 
 from profiles.models import Profile
@@ -115,3 +116,18 @@ def update_from_linkedin(backend, user, response, *args, **kwargs):   # pylint: 
 
     user_profile.linkedin = response
     user_profile.save()
+
+
+def set_last_update(details, *args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Pipeline function to add extra information about when the social auth
+    profile has been updated.
+
+    Args:
+        details (dict): dictionary of informations about the user
+
+    Returns:
+        dict: updated details dictionary
+    """
+    details['updated_at'] = datetime.utcnow().timestamp()
+    return details
--- a/backends/utils.py
+++ b/backends/utils.py
@@ -1,0 +1,50 @@
+"""
+Utility functions for the backends
+"""
+from datetime import datetime, timedelta
+
+from requests.exceptions import HTTPError
+from social.apps.django_app.utils import load_strategy
+
+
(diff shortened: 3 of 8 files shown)


backends/edxorg.py
```diff
@@ -15,11 +16,25 @@ class EdxOrgOAuth2(BaseOAuth2):
     ID_KEY = 'edx_id'
     REQUEST_TOKEN_URL = None
     EDXORG_BASE_URL = settings.EDXORG_BASE_URL
+
+    # Settings for Django OAUTH toolkit
+    # AUTHORIZATION_URL = urljoin(EDXORG_BASE_URL, '/_o/authorize/')
+    # ACCESS_TOKEN_URL = urljoin(EDXORG_BASE_URL, '/_o/token/')
+    # DEFAULT_SCOPE = ['read', 'write']
+
+    # Settings for old edx OAUTH library
```
gsidebo: will we be deleting these 3 lines when the edX PR gets merged?

backends/edxorg.py
```diff
@@ -15,11 +16,25 @@ class EdxOrgOAuth2(BaseOAuth2):
...
     AUTHORIZATION_URL = urljoin(EDXORG_BASE_URL, '/oauth2/authorize/')
     ACCESS_TOKEN_URL = urljoin(EDXORG_BASE_URL, '/oauth2/access_token/')
+    DEFAULT_SCOPE = ['email', ]
+
     ACCESS_TOKEN_METHOD = 'POST'
     REDIRECT_STATE = False
-    DEFAULT_SCOPE = ['email']
+    EXTRA_DATA = [
```
gsidebo: i just checked out the branch and i'm not seeing where this is used. in general it's not very clear what this is supposed to be doing

backends/pipeline_api.py
```diff
@@ -115,3 +116,18 @@ def update_from_linkedin(backend, user, response, *args, **kwargs):   # pylint:
 
     user_profile.linkedin = response
     user_profile.save()
+
+
+def set_last_update(details, *args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Pipeline function to add extra information about when the social auth
+    profile has beem updated.
```
gsidebo: beem -> been :)

backends/utils.py
```diff
@@ -0,0 +1,50 @@
...
+class InvalidCredentialStored(Exception):
+    """Custom exception to throw in some specific situations"""
+    def __init__(self, message, http_status_code):
+        super(InvalidCredentialStored, self).__init__(message)
+        self.http_status_code = http_status_code
+
+
+def _refresh_timestamp(user_social):
+    """
+    Private function that refresh an user access token
```
gsidebo: this is a bit confusing. would `_send_refresh_request` be a better name for this method? the docstring says this is here to "refresh an user access token" but there's already a method called `refresh_user_token`. i feel like more specific names for these methods would help

dashboard/views_test.py
```diff
@@ -153,3 +161,99 @@ def test_with_runs(self):
...
+            "expires_in": 100  # seconds
+        }
+        self.update_social_extra_data(extra_data)
+
+        with patch(
+            'edx_api.enrollments.CourseEnrollments.get_student_enrollments',
+            autospec=True,
+            return_value=self.enrollments
+        ):
+            res = self.client.get(self.url)
```
gsidebo: looks like everything from line 206 to this line can be moved into a helper method that returns `res`. the code in all 3 test cases is identical

backends/edxorg.py
giocalitri: if you mena the following 3 lines, yes

backends/edxorg.py
giocalitri: sorry, what are you referring to?

backends/utils.py
giocalitri: will do

backends/pipeline_api.py
giocalitri: will do

dashboard/views_test.py
giocalitri: if I do not find problems with the helper function, will do

backends/edxorg.py
gsidebo: `EXTRA_DATA` variable. i'm not seeing where it's used and for what

giocalitri: It is used by the `'social.pipeline.social_auth.load_extra_data'`function in the `python social auth` pipeline: this is just configuration, the code is from the library itself.

static/js/util/api.js
```diff
@@ -62,6 +66,10 @@ function fetchJSONWithCSRF(input, init) {
     // Also note that text is a promise here, not a string
     let text = response.text();
 
+    // For 400 and 401 errors, force login
```
noisecapella: Why a 400 error? That's often a validation error from the API

giocalitri: the 400 is also an error we get from edX in a special case: when we send a refresh for a token without a refresh token string. This is a very edge case, I admit it.

noisecapella: Is the 400 also coming from our server?

giocalitri: well, I propagate the 400 and 401, so technically yes, but this is up to us.

noisecapella: Can you add a comment about the refresh token string for the 400? Otherwise I'm fine with this since we're only using it on getDashboard

giocalitri:
@gsidebo @noisecapella Sorry for not putting the changes in a different commit

giocalitri:
@gsidebo In the original PR there was a missing file `utils_test.py` that I forgot to add to the tree

backends/utils_test.py
```diff
@@ -0,0 +1,121 @@
...
+        social_user.save()
+
+    @patch('backends.edxorg.EdxOrgOAuth2.refresh_token', autospec=True)
+    def test_refresh(self, mock_refresh):
+        """The refresh needs to be called"""
+        extra_data = {
+            "updated_at": (self.now - timedelta(weeks=1)).timestamp(),
+            "expires_in": 100  # seconds
+        }
+        self.update_social_extra_data(extra_data)
```
gsidebo: `update_social_extra_data` seems like it's doing too much. it's instantiating some user object, then updating a property and saving. in this method, you're calling `update_social_extra_data`, which itself creates a user object, then in the next line you're creating another user object identically. is this intentional? i feel like this should be changed in one of two ways: (1) it gets/creates the user, updates the property, saves, and returns the user object, or (2) it takes an existing `social_user` as a parameter in addition to `data`, then just updates the property and saves

giocalitri: I think there is a misunderstanding here. `update_social_extra_data` does exactly that it says: it updates the `extra_data` field of the `user_social` model instance. I do not create any new instances here; the only place where something is created is in the `setUpTestData` method.

Anyway I think I might simplify the tests by returning the `social_user` object after updating the field.

gsidebo:
👍  looks good

PR #301, opened 2016-05-10
Updated redux-asserts

noisecapella:
https://github.com/mitodl/redux-asserts/pull/8

#### What's this PR do?
Updates version number to use latest redux-asserts (there is no functional change here)

This is a library for testing so as long as the tests pass, everything should be good


bdero:
:+1:

Issue #276, opened 2016-05-05
Profile: make month / year component more material


pdpinch:
from https://github.com/mitodl/micromasters/issues/151#issuecomment-216932574

> The month-year doesn't look very material design.  Here is one suggestion

> ![month-year](https://cloud.githubusercontent.com/assets/125561/15022029/b69ace6c-11f7-11e6-8c90-b331ee26de89.png)

This needs to be applied to the education and employment profiles

PR #302, opened 2016-05-10
Added material-ui AutoComplete 

### LINKED ISSUE #277: Profile: make select component more material
from https://github.com/mitodl/micromasters/issues/151#issuecomment-216932574

> The select element (e.g. country) doesn't look material design. Can you look around for a select MD select component ?

noisecapella:
Fixes #277 

#### What's this PR do?
Adds material-ui so we can use the AutoComplete component to replace our select fields. Also updates `boundStateSelectField` to call `boundSelectField` instead of implementing duplicate behavior

Go to the profile and edit the gender field, or the state/country fields. You should notice these behaviors:
 - the autocomplete now only matches prefixes (case insensitive)
 - when user is not actively typing in the autocomplete, it should always show the currently selected item or a blank string.
 - when the user types some text and presses enter and they have something in the autocomplete, the first item in the autocomplete menu will become the selected item
 - when the user types some text which doesn't match anything and presses enter, it should revert back to whatever was previously selected
 - when user clicks on a selected item, it should become the new selected item
 - user should not be able to clear the field

Weird behaviors:
 - when you press down on the keyboard the autocomplete loses focus and the text field reverts to the currently selected item. Ideally we should let the user choose options by pressing the down key and pressing enter, but I don't know how to distinguish a loss of focus to go to the menu vs a loss of focus to go to another part of the form

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import Button from 'react-mdl/lib/Button';
+import Grid, { Cell } from 'react-mdl/lib/Grid';
 
 import ProfileTab from "../util/ProfileTab";
 import { saveAndContinue } from "../util/profile_edit";
@@ -45,30 +46,34 @@ class PersonalTab extends ProfileTab {
   }
 
   render() {
-    return <div>
-      {this.boundTextField(["first_name"], "Given name")}<br />
-      {this.boundTextField(["last_name"], "Family name")}<br />
-      {this.boundTextField(["preferred_name"], "Preferred name")}<br />
-      {this.boundSelectField(['gender'], 'Gender', this.genderOptions)}<br />
-      {this.boundSelectField(
+    return <Grid className="profile-tab-grid">
+      <Cell col={4}>{this.boundTextField(["first_name"], "Given name")}</Cell><Cell col={8} />
+      <Cell col={4}>{this.boundTextField(["last_name"], "Family name")}</Cell><Cell col={8} />
+      <Cell col={4}>{this.boundTextField(["preferred_name"], "Preferred name")}</Cell><Cell col={8} />
+      <Cell col={4}>{this.boundSelectField(['gender'], 'Gender', this.genderOptions)}</Cell><Cell col={8} />
+      <Cell col={4}>{this.boundSelectField(
         ['preferred_language'],
         'Preferred language',
         this.languageOptions
-      )}<br />
-      <h4>Where do you live?</h4>
-      {this.boundTextField(['city'], 'City')}<br />
-      {this.boundStateSelectField(['state_or_territory'], ['country'], 'State or Territory')}<br />
-      {this.boundSelectField(['country'], 'Country', this.countryOptions)}<br />
-      <h4>Where were you born?</h4>
-      {this.boundTextField(['birth_city'], 'City')}<br />
-      {this.boundStateSelectField(['birth_state_or_territory'], ['birth_country'], 'State or Territory')}<br />
-      {this.boundSelectField(['birth_country'], 'Country', this.countryOptions)}<br />
-      {this.boundDateField(['date_of_birth'], 'Date of birth')}<br />
+      )}</Cell><Cell col={8} />
+      <Cell col={4}><h4>Where do you live?</h4></Cell><Cell col={8} />
+      <Cell col={4}>{this.boundTextField(['city'], 'City')}</Cell><Cell col={8} />
+      <Cell col={4}>
+        {this.boundStateSelectField(['state_or_territory'], ['country'], 'State or Territory')}
+      </Cell><Cell col={8} />
+      <Cell col={4}>{this.boundSelectField(['country'], 'Country', this.countryOptions)}</Cell><Cell col={8} />
+      <Cell col={4}><h4>Where were you born?</h4></Cell><Cell col={8} />
+      <Cell col={4}>{this.boundTextField(['birth_city'], 'City')}</Cell><Cell col={8} />
+      <Cell col={4}>
+        {this.boundStateSelectField(['birth_state_or_territory'], ['birth_country'], 'State or Territory')}
+      </Cell><Cell col={8} />
+      <Cell col={4}>{this.boundSelectField(['birth_country'], 'Country', this.countryOptions)}</Cell><Cell col={8} />
+      <Cell col={4}>{this.boundDateField(['date_of_birth'], 'Date of birth')}</Cell><Cell col={8} />
 
       <Button raised onClick={this.saveAndContinue}>
         Save and continue
       </Button>
-    </div>;
+    </Grid>;
   }
 }
 
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -11,9 +11,6 @@ import { makeDashboardRoutes } from './dashboard_routes';
 import 'style!css!react-mdl/extra/material.css';
 import 'react-mdl/extra/material.js';
 
-// requirements for react-select
-import 'style!css!react-select/dist/react-select.css';
-
 // requirement for react-datepicker
 import 'style!css!react-datepicker/dist/react-datepicker.css';
 
--- a/static/js/dashboard_routes.js
+++ b/static/js/dashboard_routes.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Route, Router, IndexRedirect } from 'react-router';
 import { Provider } from 'react-redux';
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
+import getMuiTheme from 'material-ui/styles/getMuiTheme';
 
 import App from './containers/App';
 import DashboardPage from './containers/DashboardPage';
@@ -10,8 +12,6 @@ import EmploymentTab from './components/EmploymentTab';
 import PrivacyTab from './components/PrivacyTab';
 import EducationTab from './components/EducationTab';
 import TermsOfServicePage from './containers/TermsOfServicePage';
-import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
-import getMuiTheme from 'material-ui/styles/getMuiTheme';
 
 /**
  * Create the dashboard routes (the root React elements to be rendered into our container)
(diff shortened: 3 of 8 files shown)


noisecapella:
Needs a CSS update to make the text color consistent with react-mdl

static/js/dashboard.js
```diff
@@ -24,6 +21,10 @@ if (!Object.entries) {
 // requirement for dialog-polyfill
 import 'style!css!dialog-polyfill/dialog-polyfill.css';
 
+// material-ui requirement
+let injectTapEventPlugin = require("react-tap-event-plugin");
+injectTapEventPlugin();
```
aliceriot: is this required just for the material-ui component to work? I'm just wondering if we're worrying about the usability of MM on a touch device at all.

noisecapella: I'm not sure if it's required for AutoComplete but I wasn't able to click a SelectField without this code for some reason

aliceriot: haha, sounds like a good idea to include it then :P

aliceriot:
Functionally this seems pretty good, and I think it's a huge step up in appearance from the `react-select`.

Not sure exactly what to do about the point you bring up (w/r/t typing in text which doesn't correspond to one of the options). There's another confusing behavior there, which is if you already have a value entered for a field. Then it's possible to type in some text which doesn't match any option, as which point the field looks exactly the same as if you had picked a valid choice, and you aren't prevented from proceeding to the next point. This is confusing because it appears to the user that whatever they've just typed in ('asdfasdfasdf' or something) has actually been saved for, say, 'Country', but actually 'Country' is still just set to whatever value it was before.

Anyway, it looks like we can add a bunch of callbacks (`onInput` and so on) so we may be able to hack the behavior we want in? not sure.

aliceriot:
Anyway, code looks good to me. If you're happy with the behavior for now I say :+1:

giocalitri:
it needs a rebase

noisecapella:
Ok, I've changed a lot of stuff so you may want to go through it all again. By the way I removed `boundMonthField` and `boundYearField` in favor of `boundDateField`. We will add back a `boundMonthYearField` in #276

static/js/components/PersonalTab.js
```diff
@@ -45,30 +46,34 @@ class PersonalTab extends ProfileTab {
   }
 
   render() {
-    return <div>
-      {this.boundTextField(["first_name"], "Given name")}<br />
-      {this.boundTextField(["last_name"], "Family name")}<br />
-      {this.boundTextField(["preferred_name"], "Preferred name")}<br />
-      {this.boundSelectField(['gender'], 'Gender', this.genderOptions)}<br />
-      {this.boundSelectField(
+    return <Grid>
```
aliceriot: you may want to add `className="profile-tab-grid"` here, so the spacing and width is in line with the other tabs.

static/js/dashboard_routes.js
```diff
@@ -27,15 +27,15 @@ export function makeDashboardRoutes(browserHistory, store, onRouteUpdate) {
...
+              <IndexRedirect to="personal" />
+              <Route path="personal" component={PersonalTab} />
               <Route path="education" component={EducationTab}/>
-              <Route path="professional" component={EmploymentTab}/>
-              <Route path="privacy" component={PrivacyTab}/>
+              <Route path="professional" component={EmploymentTab} />
+              <Route path="privacy" component={PrivacyTab} />
             </Route>
-            <Route path="/terms_of_service" component={TermsOfServicePage}/>
+            <Route path="/terms_of_service" component={TermsOfServicePage} />
```
aliceriot: I agree that `<Foo prop="asdf" />` is more readable than `<Foo prop="asdf"/>`, is there an eslint rule we can turn on to enforce this?

aliceriot:
I'm noticing that the material-ui autocomplete and textfields are noticeably slower than what we're replacing. is this something you've noticed too @noisecapella ?

noisecapella:
Yes I noticed this too. I'm hoping it will be faster in the minimized version but I have no idea

noisecapella:
Do you have your console open while messing around with these fields?

For AutoComplete there is a `menuCloseDelay` field which makes things slower for some reason. We can also turn off animation where we can

aliceriot:
yeah I had the console open, I'll try with it closed. my computer is probably a little bit slower that yours.

aliceriot:
ah ok yeah closing the console makes it a lot faster, not a problem.

aliceriot:
yeah, we may want to turn off some of the animation. the animation when you select something in the autocomplete is a bit strange, it takes longer than it should to get out of the way.

static/js/dashboard_routes.js
aliceriot: I took a look, it doesn't seem like there is one :(

static/js/util/profile_edit.js
```diff
@@ -30,12 +32,16 @@ export function boundTextField(keySet, label) {
     _.set(clone, keySet, e.target.value);
     updateProfile(clone);
   };
+
+  // fullWidth means set width to 100% instead of 256px. This lets us use the
+  // Grid and Cell to manage its size
```
aliceriot: :100:

noisecapella: One of many quirks of material-ui

static/js/util/profile_edit.js
```diff
@@ -49,36 +55,66 @@ export function boundSelectField(keySet, label, options) {
     errors,
     updateProfile,
   } = this.props;
-  let onChange = value => {
+  let onChange = option => {
     let clone = _.cloneDeep(profile);
-    _.set(clone, keySet, value? value.value : '');
+    let toStore;
+    if (option.value !== null && option.value !== undefined) {
+      toStore = option.value.props.value;
```
aliceriot: so here we're pulling the option value out of the `MenuItem` component we declare below (in `convertOption`). is there any reason we couldn't also stick a `value` prop on the return value of that function? it might make the code here a little simpler.

static/js/util/profile_edit.js
```diff
@@ -49,36 +55,66 @@ export function boundSelectField(keySet, label, options) {
...
-      options={options}
-      value={_.get(profile, keySet)}
-      placeholder={label}
-      onChange={onChange} />
-    <span className="validation-error-text">{_.get(errors, keySet)}</span>
-  </div>;
+  let selectedValue = _.get(profile, keySet);
+  let selectedOption;
+  for (let option of options) {
+    if (option.value === selectedValue) {
```
aliceriot: you might be able to use `Array.find` here, something like

```js
options.find(option => option.value === _.get(profile, keySet))
```

static/js/util/profile_edit.js
```diff
@@ -49,36 +55,66 @@ export function boundSelectField(keySet, label, options) {
...
+  let convertOption = option => {
+    if (option) {
+      return {
+        text: option.label,
+        value: <MenuItem primaryText={option.label} value={option.value}/>
+      };
+    } else {
+      return {};
+    }
+  };
```
aliceriot: this seems like we have to do some juggling to get the way we currently declare select options (`[{value: 3, label: "Foo" }]`) to work with this field. Would it make more sense to change how they're declared to match the requirements of this new component? (e.g. in `ProfileTab.js`

static/js/util/profile_edit.js
```diff
@@ -89,14 +125,7 @@ export function boundStateSelectField(stateKeySet, countryKeySet, label) {
...
 
-  return <div>
-    <Select
-      options={options}
-      value={_.get(profile, stateKeySet)}
-      placeholder={label}
-      onChange={onChange} />
-    <span className="validation-error-text">{_.get(errors, stateKeySet)}</span>
-  </div>;
+  return mockedBoundSelectField.call(this, stateKeySet, label, options);
```
aliceriot: this is slick!

aliceriot:
couple comments!

static/js/util/profile_edit.js
noisecapella: I feel like the current way is cleaner since the `<MenuItem ...` stuff is hidden away in this function, and we don't need to duplicate `option.label` like we do here.

static/js/util/profile_edit.js
noisecapella: I'd rather not since we can't call it `option.value` and I can't think of a better name for it. It also means I'd have to update the current unit tests so that `onChange` can make use of it

aliceriot: ok, in that case do you think it would be wise to change this line to:

```js
toStore = _.get(option, ['value', 'props', 'value'])
```

for defensive-ness' sake?

noisecapella:
I updated the behavior to set the field to whatever the currently selected value is when the focus leaves the field. Still need to clean it up and write tests for it, it's getting pretty messy

aliceriot:
cool, I'll take another look after standup!

noisecapella:
Should we allow the user to clear the autocomplete in any way? I feel like it's simpler if we don't allow them to, and I can't think of how the UI would work, but I'm not sure if there's a use case we want to allow.

noisecapella:
So the current behavior should look like this:
 - the autocomplete now only matches prefixes (case insensitive)
 - when user is not actively typing in the autocomplete, it should always show the currently selected item or a blank string.
 - when the user types some text and presses enter and they have something in the autocomplete, the first item in the autocomplete menu will become the selected item
 - when the user types some text which doesn't match anything and presses enter, it should revert back to whatever was previously selected
 - when user clicks on a selected item, it should become the new selected item
 - user should not be able to clear the field

Weird behaviors:
 - when you press down on the keyboard the autocomplete loses focus and the text field reverts to the currently selected item. Ideally we should let the user choose options by pressing the down key and pressing enter, but I don't know how to distinguish a loss of focus to go to the menu vs a loss of focus to go to another part of the form

noisecapella:
Tests are now in place!

aliceriot:
Just going through your bullet points above:

- [x] the autocomplete now only matches prefixes (case insensitive)
- [ ] when user is not actively typing in the autocomplete, it should always show the currently selected item or a blank string.
- [ ] when the user types some text and presses enter and they have something in the autocomplete, the first item in the autocomplete menu will become the selected item
- [ ] when the user types some text which doesn't match anything and presses enter, it should revert back to whatever was previously selected
- [ ] when user clicks on a selected item, it should become the new selected item
user should not be able to clear the field

noisecapella:
We can change it to always show all options (do no filtering) but this breaks the behavior where Enter picks the first item in the list. I suppose we could have Enter do a comparison that the text in the text field matches exactly, then choose that item, else revert to what was already selected.

For the hint text do you have an idea where to put that in the UI?

aliceriot:
hmm not sure, it's kind of hard to make things like that unobtrusive while still being clear. maybe we could conditionally modify the `floatingTextLabel` prop?

noisecapella:
We don't have a way to tell whether the user opened the menu via mouse click or by entering text, at least not without digging deeper into AutoComplete, which is a little unpalatable

noisecapella:
By the way review comments should be addressed now

aliceriot:
cool, I just looked over the code and it LGTM. how do we feel about merging what we have, and then coming back around sometime in the future to deal with some of the small UX edge cases?

noisecapella:
Sounds good

aliceriot:
:+1:

PR #308, opened 2016-05-12
Added material-ui, replaced react-mdl Dialog with material-ui Dialog

noisecapella:
Required for #277

#### What's this PR do?
There's an issue with the material-ui AutoComplete because it uses z-index to hover above the textfield area. The HTML5 dialog element which react-mdl uses is defined as top-layer when activated using `dialog.showModal()` so we can't use it with the AutoComplete. Fortunately the material-ui Dialog is a drop in replacement

Make sure the dialogs still work. There's one in App.js and one in the education tab when you click the edit or plus button. Cancel or ESC or clicking outside the dialog should cancel, and the other buttons should act like before.

The react-select component will be truncated within the textbox, but we're getting rid of it in #277 anyway


aliceriot:
there are a bunch of css changes that appear to be just code cleanup, is that correct?

noisecapella:
Not exactly, it removes the surrounding `#dashboard { }`. material-ui puts the dialog at the body level so the CSS needs to be modified to handle those fields correctly.

aliceriot:
ah gotcha, that makes sense.

static/js/containers/App.js
```diff
@@ -23,7 +18,6 @@ import {
 } from '../actions/index';
 import {
   validateProfileComplete,
-  doDialogPolyfill
 } from '../util/util';
```
aliceriot: nit: this could probably go on one line now

aliceriot:
aside from that one comment this looks good to me! :+1:

PR #311, opened 2016-05-12
Added field for month/year

### LINKED ISSUE #276: Profile: make month / year component more material

noisecapella:
Fixes #276
#### What's this PR do?
Uses the textfields to bring up month and year fields. Both fields should validate the numbers provided

`profile_edit.js`

There shouldn't be any odd behavior (at least in the UI):
 - There should be two text fields and a label which look like this: Label MM / YYYY
 - you should be able to edit either textbox independently. If one is blank, there should be a validation error when you click save. After the validation error you should be able to save if you fill in proper values for each textbox.
 - Only numbers 1-12 and 1-9999 should be accepted by the field. You should be able to type in text but it will cause a validation error

### DIFF
--- a/static/js/util/ProfileTab.js
+++ b/static/js/util/ProfileTab.js
@@ -5,6 +5,7 @@ import {
   boundDateField,
   boundTextField,
   boundSelectField,
+  boundMonthYearField,
   editProfileObjectArray,
   boundStateSelectField,
 } from './profile_edit';
@@ -22,6 +23,7 @@ class ProfileTab extends React.Component {
     this.boundSelectField = boundSelectField.bind(this);
     this.boundStateSelectField = boundStateSelectField.bind(this);
     this.boundDateField = boundDateField.bind(this);
+    this.boundMonthYearField = boundMonthYearField.bind(this);
     this.editProfileObjectArray = editProfileObjectArray.bind(this);
 
     // options we set (for select components)
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -10,17 +10,27 @@ import MenuItem from 'material-ui/MenuItem';
 import TextField from 'material-ui/TextField';
 
 import { DATE_FORMAT } from '../constants';
+import {
+  validateMonth,
+  validateYear,
+} from '../util/util';
 
 // utility functions for pushing changes to profile forms back to the
 // redux store.
 // this expects that the `updateProfile` and `profile` props are passed
 // in to whatever component it is used in.
 
-// bind to this.boundTextField in the constructor of a form component
-// to update text fields when editing the profile
-// we pass in a keyset looking like this:
-//
-// ["top-level-key", index, "nested_object_key"] or just ["top_level_key"]
+/**
+ * bind to this.boundTextField in the constructor of a form component
+ * to update text fields when editing the profile
+ * we pass in a keyset looking like this:
+ *
+ * ["top-level-key", index, "nested_object_key"] or just ["top_level_key"]
+ *
+ * @param keySet {String[]} Path to the field
+ * @param label {String} Label for the field
+ * @returns {ReactElement}
+ */
 export function boundTextField(keySet, label) {
   const {
     profile,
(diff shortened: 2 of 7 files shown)


static/js/components/PersonalTab.js
```diff
@@ -45,30 +46,34 @@ class PersonalTab extends ProfileTab {
   }
 
   render() {
-    return <div>
-      {this.boundTextField(["first_name"], "Given name")}<br />
-      {this.boundTextField(["last_name"], "Family name")}<br />
-      {this.boundTextField(["preferred_name"], "Preferred name")}<br />
-      {this.boundSelectField(['gender'], 'Gender', this.genderOptions)}<br />
-      {this.boundSelectField(
+    return <Grid>
```
aliceriot: probably want to throw `className="profile-tab-grid"` in here

noisecapella:
Ready for review! Note that this is rebased on #302

noisecapella:
One quirk of this implementation is that, if you have some invalid text in one of the fields you will get a validation error saying that the field is required, not that the validation of the text failed. I'd like to push that off for another PR, if we ever deem it important enough to work on

aliceriot:
yeah, our current validation is doing it's job, but eventually we may want to do more complicated stuff (verifying date A is later than date B, verifying that field B contains a valid email address, etc).

static/js/util/util.js
```diff
@@ -280,3 +292,36 @@ export function makeStrippedHtml(textOrElement) {
...
+  return month;
+}
+
+/**
+ * Validate a year string is an integer and fits into YYYY
+ * @param {String} string The input string
+ * @returns {Number|undefined} The valid year or undefined if not valid
+ */
+export function validateYear(string) {
+  let year = filterPositiveInt(string);
```
aliceriot: can we simplify this to just `Number(string)`, and then as `_.isNaN(year)` below?

noisecapella: I wanted to catch text like "2e0" which will convert to an integer but isn't a meaningful year or month

static/js/util/profile_edit.js
```diff
@@ -312,22 +312,22 @@ export function boundMonthYearField(keySet, label) {
...
+      onChange={e => setNewDate(e.target.value, undefined)}
+    /> <span className="slash">/</span> <TextField
+      floatingLabelText="YYYY"
+      style={{
+        maxWidth: "4em"
+      }}
+      fullWidth={true}
+      value={monthYear.year !== undefined ? monthYear.year : ""}
+      onChange={e => setNewDate(undefined, e.target.value)}
+    />
```
aliceriot: I think this may just be a differing style preference, but I find JSX a lot easier to read when the opening and closing tags are generally one per line (obviously excepting very short tags like `<span>{ alertText() }</span>`).

noisecapella: I did the `/> <span` so that React knows there's a space between the two elements. If they're on separate lines React assumes I meant to concatenate them together without any spaces. But it looks like here I can move the spaces inside the `span`

static/js/util/util.js
aliceriot: cool, that makes sense to check for.

what about this:

```js
let numberFilter = string => (
  Number(string.replace(/\D/,""))
);
```

a little more compact, I guess? haha. your call.

noisecapella: For the `"2e0"` case that would become `20` since it just strips out non-numeric characters

aliceriot: mm yeah I guess so, probably just keep what's here then

static/js/util/ProfileTab.js
```diff
@@ -22,7 +23,9 @@ class ProfileTab extends React.Component {
     this.boundSelectField = boundSelectField.bind(this);
     this.boundStateSelectField = boundStateSelectField.bind(this);
     this.boundDateField = boundDateField.bind(this);
+    this.boundMonthYearField = boundMonthYearField.bind(this);
     this.editProfileObjectArray = editProfileObjectArray.bind(this);
+    this.boundMonthYearField = boundMonthYearField.bind(this);
```
aliceriot: binding it to `this` twice?

aliceriot:
found one little typo, but other than that it's :+1:

Issue #233, opened 2016-04-26
When profile is completed for first time, set email_optin and filled_out to True


noisecapella:
Not sure which issue this connects to so I'm filling out this story just in case

noisecapella:
This is blocked by #150 #151

PR #312, opened 2016-05-12
Implemented new employment page design

### LINKED ISSUE #255: Handle deletion of employment items in UI
For #150

### LINKED ISSUE #289: style employment step to look like current mock-ups
See https://invis.io/4E791TKU3#/153562860_Step_Three_-_Professional_Info

- [ ] Implement switch instead of "are you employed" question

### LINKED ISSUE #290: Profile: allow user to submit employment info without an end date
See https://invis.io/VC7850ZBS#/156281164_Step_Three_-_Professional_Info_-_Add_Emp_Modal

- [x] add label for "end date"
- [x] add hint text "Leave blank if this is a current position"
- [x] allow End Date to be null

aliceriot:

closes #289, closes #255, closes #290 

#### What's this PR do?

This implements the new design for the employment step, as seen here:

https://projects.invisionapp.com/share/4E791TKU3#/screens/153562860

There are a lot of changes! So probably just looking at the UI, making sure things look good and work as they should.

Then the additions to `EmploymentTab.js`.

Make sure you can edit work history! it should let you:

- add a new entry
- edit an existing entry
- delete an existing entry
- turn editing the work history on or off
- leave the end date blank (for a current position)

it shouldn't:

- let you 'save' an invalid entry
- let you set an empty start date (and the rest of the required fields)

it should:

- alert you with a little exclamation thing when there's a validation error on a particular entry
- show 'Current' as the end date if a job has no end date entered
- keep the + button in the same place on the bottom of the card\
- show some filler text when the editor is switched off

current UI state:

employment-ui

employment-edit

### DIFF
--- a/static/js/components/EmploymentTab.js
+++ b/static/js/components/EmploymentTab.js
@@ -1,7 +1,14 @@
 import React from 'react';
+import Grid, { Cell } from 'react-mdl/lib/Grid';
+import Card from 'react-mdl/lib/Card/Card';
 import Button from 'react-mdl/lib/Button';
-import Grid from 'react-mdl/lib/Grid';
-import { Cell } from 'react-mdl/lib/Grid';
+import Switch from 'react-mdl/lib/Switch';
+import FABButton from 'react-mdl/lib/FABButton';
+import Icon from 'react-mdl/lib/Icon';
+import IconButton from 'react-mdl/lib/IconButton';
+import _ from 'lodash';
+import Dialog from 'material-ui/Dialog';
+import moment from 'moment';
 
 import ProfileTab from "../util/ProfileTab";
 import { saveAndContinue } from '../util/profile_edit';
@@ -9,34 +16,35 @@ import { saveAndContinue } from '../util/profile_edit';
 class EmploymentTab extends ProfileTab {
   constructor(props) {
     super(props);
-    this.saveAndContinue = saveAndContinue.bind(this, '/profile/privacy', EmploymentTab.validation);
     this.blankWorkHistoryEntry = {
-      position: "",
-      industry: "",
-      company_name: "",
-      start_date: "",
-      end_date: "",
-      city: "",
-      country: "",
-      state_or_territory: '',
+      position: null,
+      industry: null,
+      company_name: null,
+      start_date: null,
+      end_date: null,
+      city: null,
+      country: null,
+      state_or_territory: null,
     };
   }
+  nextUrl = "/profile/privacy";
 
   static propTypes = {
-    profile:        React.PropTypes.object,
-    saveProfile:    React.PropTypes.func,
-    updateProfile:  React.PropTypes.func,
+    profile:            React.PropTypes.object,
+    ui:                 React.PropTypes.object,
+    saveProfile:        React.PropTypes.func,
+    updateProfile:      React.PropTypes.func,
+    setWorkHistoryEdit: React.PropTypes.func,
+    setWorkDialogIndex: React.PropTypes.func,
   };
 
   static defaultProps = {
     requiredFields: [],
     validationMessages: {
-      'currently_employed': 'Current employment status',
       'position': 'Position',
       'industry': 'Industry',
       'company_name': 'Company Name',
       'start_date': 'Start Date',
-      'end_date': 'End Date',
       'city': 'City',
       'country': 'Country',
       'state_or_territory': 'State or Territory',
@@ -48,7 +56,6 @@ class EmploymentTab extends ProfileTab {
     'industry',
     'company_name',
     'start_date',
-    'end_date',
     'city',
     'country',
     'state_or_territory',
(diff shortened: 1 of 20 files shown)


aliceriot:
note to reviewers: there are a couple of issues I need to sort out after rebasing following the merge of #302 and #311

static/js/util/profile_edit.js
```diff
@@ -294,6 +293,9 @@ export function boundMonthYearField(keySet, label) {
       // store the edit value and make the formatted date null so it doesn't pass validation
       _.set(clone, editKeySet, newMonthYear);
       _.set(clone, keySet, undefined);
+    } else if ( validatedMonth === "" || validatedYear === "") {
+      _.set(clone, editKeySet, undefined);
+      _.set(clone, keySet,
```
noisecapella: Can you fix this?

aliceriot: currently working on it!

static/js/util/util.js
```diff
@@ -299,6 +299,9 @@ export function makeStrippedHtml(textOrElement) {
  * @returns {Number|undefined} The valid month or undefined if not valid
  */
 export function validateMonth(string) {
+  if ( string === "" ) {
```
noisecapella: Why are we allowing empty strings here? Is there a need for it in the new UI?

static/js/util/profile_edit.js
noisecapella: That's fine, ping me when you want me to take a look at it again

static/js/util/util.js
aliceriot: yeah, I need to be able to leave the end month blank

aliceriot: (for current positions)

static/js/util/profile_edit.js
aliceriot: should be all set now!

static/js/util/util.js
noisecapella: That's fine, can you add this information to the JSDoc here and below?

static/js/reducers/ui_test.js
```diff
@@ -0,0 +1,100 @@
...
+      done();
+    });
+  });
+
+  it('should set the work history dialog visibility', done => {
+    dispatchThen(setWorkDialogVisibility(true), [SET_WORK_DIALOG_VISIBILITY]).then(state => {
+      assert.equal(state.workDialogVisibility, true);
+    });
+
+    dispatchThen(setWorkDialogVisibility(false), [SET_WORK_DIALOG_VISIBILITY]).then(state => {
```
noisecapella: I think you'll need to move this within the `then` above, or else you may reach `done` without executing the `then` above

static/js/reducers/ui_test.js
```diff
@@ -0,0 +1,100 @@
...
+      done();
+    });
+  });
+
+  it('should set work history edit', done => {
+    dispatchThen(setWorkHistoryEdit(true), [SET_WORK_HISTORY_EDIT]).then(state => {
+      assert.equal(state.workHistoryEdit, true);
+    });
+
+    dispatchThen(setWorkHistoryEdit(false), [SET_WORK_HISTORY_EDIT]).then(state => {
```
noisecapella: Same comment here

noisecapella:
Can you change the formatted dates in the UI to exclude the date portion?

static/js/util/util.js
aliceriot: yah

noisecapella:
Does the switch do anything meaningful at the moment? Any difference in validation?

static/js/reducers/ui_test.js
aliceriot: ah good catch

aliceriot:
@noisecapella it doesn't do anything currently besides turn the ui on / off

aliceriot:
w/r/t to the formatted dates, do you mean what I'm showing in the ui here?

dates

noisecapella:
At some point we should update these dialog boxes to scroll vertically (#320)

noisecapella:
@aliceriot yes, we should be showing 2016-05 or maybe 05/2016 instead

aliceriot:
ah ok so we should hide the day value, gotcha.

noisecapella:
It looks like the mockups use MM/YYYY so we should format it like that in the UI

noisecapella:
I'm getting this warning:
Warning: Switch is changing a uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

noisecapella:
In `blankWorkHistoryEntry` can you set those fields to default to null instead of an empty string? Right now the empty string being sent to the REST API is confusing it

noisecapella:
I'm getting a new warning:

warning.js:44 Warning: `value` prop on `input` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components.

Can you change `boundTextField` to check if the value is null or undefined and use an empty string in the `value` prop in that case?

static/js/util/profile_edit.js
```diff
@@ -334,10 +340,10 @@ export function boundMonthYearField(keySet, label) {
  * pass an option callback if you need nested validation
  * (see EmploymentTab for an example)
  *
- * @param next {String} URL to redirect to after successful validation
+ * @param afterSave {String} URL to redirect to or function to call after successful validation 
  * @param nestedValidationCallback {func} If present, a function to retrieve validation fields
  */
-export function saveAndContinue(next, nestedValidationCallback) {
+export function saveAndContinue(afterSave, nestedValidationCallback) {
```
noisecapella: I think we can rewrite this to return a `Promise` instead, the one returned by `saveProfile.then()` below. Then the calling function could do `this.context.router.push('')` or execute other custom code. What do you think?

noisecapella: We should also unit test this at some point, but I can do that as part of #233

aliceriot: I think that makes sense - I'll give it a try.

static/js/reducers/ui.js
```diff
@@ -0,0 +1,56 @@
...
+    return Object.assign({}, state, {
+      dialog: Object.assign(
+        {},
+        state.dialog,
+        { visible: action.payload }
+      )
+    });
+  case SET_WORK_HISTORY_EDIT:
+    return Object.assign({}, state, {
+      workHistoryEdit: action.payload
```
noisecapella: Since this is a boolean, can you have this default to `false` in `INITIAL_UI_STATE`? This will also avoid the need for checking for `undefined` since it will always be true or false

aliceriot: yeah, that's a good idea, I had to do some fiddling to check for it.

static/js/components/EmploymentTab.js
```diff
@@ -65,56 +79,232 @@ class EmploymentTab extends ProfileTab {
...
+  addWorkHistoryEntry = () => {
+    const {
+      updateProfile,
+      profile,
+      setWorkDialogIndex,
+      setWorkDialogVisibility,
+    } = this.props;
+    let clone = Object.assign({}, profile);
+    clone['work_history'] = clone['work_history'].concat(
+      _.cloneDeep(this.blankWorkHistoryEntry)
```
noisecapella: Unless we're editing `blankWorkHistoryEntry` we shouldn't need to clone it. Can you rewrite as:
        'work_history': profile.work_history.concat(this.blankWorkHistoryEntry)
    });

static/js/reducers/ui.js
noisecapella: Can you also do the same for work dialog visibility?

static/js/components/EmploymentTab.js
```diff
@@ -65,56 +79,232 @@ class EmploymentTab extends ProfileTab {
...
+      </Grid>
+    );
+  }
+
+  renderWorkHistory () {
+    const { ui, profile: { work_history } } = this.props;
+    if ( _.isUndefined(ui.workHistoryEdit) || ui.workHistoryEdit === true ) {
+      let workHistory = [];
+      if ( !_.isEmpty(work_history) ) {
+        workHistory.push(...work_history.map((entry, i) => this.jobRow(entry, i)));
```
noisecapella: Why not `workHistory = work_history.map(...);`?

static/js/components/EmploymentTab.js
```diff
@@ -65,56 +79,232 @@ class EmploymentTab extends ProfileTab {
...
+          </span>
+        </Cell>
+      </Grid>
+    );
+  }
+
+  renderWorkHistory () {
+    const { ui, profile: { work_history } } = this.props;
+    if ( _.isUndefined(ui.workHistoryEdit) || ui.workHistoryEdit === true ) {
+      let workHistory = [];
```
noisecapella: Is it possible to rename this? It seems odd to have `workHistory` and `work_history` in the same scope

static/js/components/EmploymentTab.js
```diff
@@ -65,56 +79,232 @@ class EmploymentTab extends ProfileTab {
...
+      setWorkDialogIndex,
+      errors,
+    } = this.props;
+    let editCallback = () => {
+      setWorkDialogIndex(index);
+      setWorkDialogVisibility(true);
+    };
+    let validationAlert = () => {
+      if (_.get(errors, ['work_history', index])) {
+        return <IconButton name="error" onClick={editCallback} />;
```
noisecapella: This is nice!

static/js/components/EmploymentTab.js
```diff
@@ -65,56 +79,232 @@ class EmploymentTab extends ProfileTab {
...
+          <IconButton name="edit" onClick={editCallback} />
+          <IconButton name="clear" onClick={deleteEntry} />
         </Cell>
       </Grid>
     );
   }
 
   render () {
+    const {
+      ui: { workHistoryEdit },
```
noisecapella: Can we merge this with the next line?

noisecapella:
Functionality looks great!

static/js/components/EmploymentTab.js
aliceriot: :D

static/js/components/EmploymentTab.js
aliceriot: I think the current state is an artifact from an refactor - I was declaring `workHistory`, adding some stuff to it, and then pushing the `jobRows` on. I didn't change things around when I moved stuff around I guess! haha

aliceriot:
cool! I just made the change to `saveAndContinue` to have it return a promise. It's a little bit of a refactor.

Now, instead of binding `saveAndContinue` to `this` in the constructor, we declare an instance method (`saveAndContinue`) on `ProfileTab`. 

This just looks like:

```js
  saveAndContinue = () => {
    saveAndContinue.call(this, this.constructor.prototype.validation).then(() => {
      this.context.router.push(this.nextUrl);
    });
  }
```

since `saveAndContinue` (from `profile_edit.js`) now returns a promise we can chain off of it. Instead of binding an argument to `saveAndContinue` we set an instance property (`nextUrl`) on each tab class (`PrivacyTab`, etc) which we can use to get the correct url to push.

`saveAndContinue` now just takes one argument - if `this.constructor.prototype` doesn't have `validation` defined on it (`EmploymentTab` and `EducationTab` do) this value will be `undefined`. 

this also made it easier to implement the `saveWorkHistoryEntry` method on `EmploymentTab` (this is run when we click 'save' on the work history entry dialog). this just looks like:

```js
  saveWorkHistoryEntry = () => {
    saveAndContinue.call(this, EmploymentTab.validation).then(() => {
      this.closeWorkDialog();
    });
  }
```

static/js/components/EmploymentTab.js
```diff
@@ -235,10 +235,7 @@ class EmploymentTab extends ProfileTab {
   }
 
   render () {
-    const {
-      ui: { workHistoryEdit },
-      ui: { workDialogVisibility },
-    } = this.props;
+    const { ui: { workHistoryEdit }, ui: { workDialogVisibility } } = this.props;
```
noisecapella: You should be able to merge this further: `{ ui: { workHistoryEdit, workDialogVisibility } }`

noisecapella:
I feel like the employment history should be sorted somehow, but maybe it's better to file an issue and skip it for now

aliceriot:
right now it's just database order I guess? to me it would make sense to have the entries sorted by date or something...not sure.

static/js/components/EmploymentTab.js
aliceriot: I changed it to this:

```js
    clone['work_history'] = clone['work_history'].concat(this.blankWorkHistoryEntry);
```

without cloning profile it's possible to achieve a situation where two of the work_history array entries point at the same object, so that editing one of them appears to edit both entries in the UI.

prototypes

You're right that we don't need to clone `this.blankWorkHistoryEntry` though. The above was with this code:

```js
    'work_history': profile.work_history.concat(this.blankWorkHistoryEntry)
});
```

steps to repro it:

1. remove all entries (click the X)
2. add two entries without saving either (hit the cancel button when the dialog opens)
3. edit one, see edits replayed twice (two entries in the array are pointing at the same object).

noisecapella: I feel like having multiple references to the same object should be legal in the redux store, since those objects are not supposed to be modified. I investigated and it looks like `_.cloneDeep` will clone objects with two or more references but those references all get the same clone. Python does this too in `deepcopy`, it looks like it keep track of already cloned objects to prevent recursion errors.

I'm thinking maybe this would be best handled in the `bound` methods, but cloning `blankWorkHistoryEntry` is probably a simpler solution for now

noisecapella: On a side note, we should revert the state of the edit when the user clicks 'Cancel'. Since the user shouldn't have any state which is being edited which is not in the dialog box, it should be enough to clear the `edit` using the action similar to how `EducationTab` does it.

aliceriot: cool, I just added a method to the `ProfilePage` class that we pass down to the tabs via `props`, and added a call to that method in the function to handle closing the dialog.

static/js/components/EmploymentTab.js
aliceriot: oh indeed /facepalm

static/js/components/EmploymentTab.js
aliceriot: I changed the array name to `workHistoryRows`

noisecapella:
:+1: 

At some point we should get everything consistent again, for example the new save and continue button here doesn't match anywhere else

aliceriot:
agreed - I think there are tickets to implement the mockups for education, personal, privacy tabs open now. deff looks weird when you're clicking through them :)

Issue #307, opened 2016-05-12
Update documentation and add scripts for OS X so that webpack-dev-server runs locally but everything else runs via Docker


noisecapella:
This will remove the need for docker-osx-dev. We will need to update our handling of the host string since right now it assumes the dev server host is the same as the Django host. We will need to figure out how to reference the host IP address from within docker-machine, and we should add a script to let OS X users start the dev server more easily (this would be used by docker-compose.yml too)

giocalitri:
#356 fixed this: closing

PR #313, opened 2016-05-13
Update documentation

gsidebo:

No ticket for this one

#### What's this PR do?

Brings MM setup documentation to a somewhat current/complete state. Further updates will be made to the readme after https://github.com/mitodl/micromasters/issues/307 (moving npm/webpack setup to the host machine rather than the Docker container). Some other cleanup was also done (moving Makefile commands into a shell script, adding .idea/ to .rsync_ignore)

Look at the readme on its own (not in a diff) and see if you spot any glaring omissions/inaccuracies.

Getting MM running locally for the first time involved a lot of steps that weren't in the docs. This was an attempt to get the readme to a point where a user could get the app running with minimal outside help. Getting it done now was a good idea since the process was still fresh in my mind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,58 +1,41 @@
 # Micromasters
 Portal for learners and course teams to access MITx Micromasters programs
 
-## Getting Started
+## Major Dependencies
+- Docker
+  - OSX recommended install method: [Download from Docker website](https://docs.docker.com/mac/)
+- docker-compose
+  - Recommended install: pip `pip install docker-compose`
+- Virtualbox (https://www.virtualbox.org/wiki/Downloads)
 
-Although you can run Micromasters locally with a default sqlite database after
-installing the ``requirements.txt`` file, the recommended way is to 
-use [Docker](https://www.docker.io). Install Docker, then install 
-``docker-compose`` and run the app: 
+## (OSX only) Getting your machine Docker-ready
 
-    pip install docker-compose 
-    docker-compose up
+#### Create your docker container:
 
-### (OSX only) Create your docker machine:
-
-    docker-machine create default
-    docker-machine start default
-    docker-machine env default
-    eval "$(docker-machine env default)"
-
-These commands create a Docker container named ``default``, start the
+The following commands create a Docker machine named ``mm``, start the
 container, and configure environment variables to facilitate communication
-with the ``edX`` instance.
+with the edX instance.
 
+    docker-machine create --driver virtualbox mm
+    docker-machine start mm
+    # 'docker-machine env (machine)' prints export commands for important environment variables
+    eval "$(docker-machine env mm)"
 
-This will set up a near production-ready containerized development 
-environment that runs migrations, with the Django development server 
-running on port ``8079``.
+#### Enable file syncing between the host machine and Docker:
 
-To run one-off commands, like shell, you can run:
+Due to file notification issues using Docker in OSX, development is more
+efficient if you install [docker-osx-dev](https://github.com/brikis98/docker-osx-dev).
+This synchronizes host file-system changes with the Docker container.
 
-    docker-compose run web python manage.py shell
+First, set up ``docker-osx-dev`` via shell script.
 
-or to create root user:
+    . ./osx_docker_dev_setup.sh
 
-    docker-compose run web python manage.py createsuperuser
-
-
-## For OS X Development
-
-Due to issues using Docker in OSX, development is more efficient if you
-install [docker-osx-dev](https://github.com/brikis98/docker-osx-dev).
-docker-osx-dev synchronizes host file-system changes with the Docker
-container. If you have [``homebrew``](http://brew.sh/) installed, you can 
-install ``docker-osx-dev`` by typing ``make`` in the root of the Micromasters
-project directory. The make file will install or update Docker using Homebrew
-and go on to install ``docker-osx-dev``.
-
-Subsequently, before you run the application with ``docker-compose up``, 
-you would run:
+Subsequently, before you run the application with ``docker-compose up``,
+you would run this command in a separate terminal tab (assumes machine
+name ``mm``):
   
-    docker-osx-dev -m default -s ./ --ignore-file '.rsync-ignore'
-
-(Assuming your Docker VM is called ``default``, and your current working
-directory is the root of the ``micromasters`` source directory).
+    docker-osx-dev -m mm -s ./ --ignore-file '.rsync-ignore'
 
 This starts a process that monitors the file system for changes. On startup
 you may receive this error:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,58 +1,54 @@
...
-These commands create a Docker container named ``default``, start the
+These commands create a Docker container named ``mm``, start the
 container, and configure environment variables to facilitate communication
-with the ``edX`` instance.
+with the edX instance.
 
+    docker-machine create mm
+    docker-machine start mm
+    # 'docker-machine env (container)' prints export commands for important environment variables
+    eval "$(docker-machine env mm)"
```
pdpinch: Curious about the recommendation to name it mm instead of default. Makes sense I guess.

README.md
```diff
@@ -1,58 +1,54 @@
...
-environment that runs migrations, with the Django development server 
-running on port ``8079``.
-
-To run one-off commands, like shell, you can run:
-
-    docker-compose run web python manage.py shell
+_**NOTE:** Setting an environment variable called
+**MM_CONTAINER_NAME** will allow shell/python scripts and one-off commands
+to easily work with your Docker container. Add this to your shell config
+file (.bashrc, et al): ``export MM_CONTAINER_NAME='<your_desired_container_name>'``_
```
pdpinch: given that you called in mm above, shouldn't we just name it mm here?

README.md
```diff
@@ -68,96 +64,128 @@ you may receive this error:
...
 
-Create a user who has permission to view the course:
+Micromasters can work with a live instance of edX, but it's recommended that
+you get it running locally. It's obviously more configurable that way, and you'll
+likely need to run it locally for other projects in the future.
 
-    docker-compose run web python manage.py createsuperuser
+#### 1) Install edX
+Run through the
+[instructions provided by edX](https://openedx.atlassian.net/wiki/display/OpenOPS/Running+Devstack)
```
pdpinch: I should have mentioned this to you earlier, but I think https://edx-installing-configuring-and-running.readthedocs.io/en/latest/installation/devstack/install_devstack.html may be more authoritative than confluence.

README.md
```diff
@@ -68,96 +64,128 @@ you may receive this error:
...
+        user.is_superuser=True
+        user.save()
+
+#### 4) Add an OAuth client
+
+Run Django admin (see "In Django admin" section from step 2), navigate to the
+OAuth2 clients section (eg: http://192.168.33.10:8000/admin/oauth2/client/), and add a
+new client. Fill in the values as follows:
+
+- **User**: Use the lookup (magnifying glass) to find your superuser
```
pdpinch: I'm not sure this is actually necessary. @giocalitri ?

README.md
```diff
@@ -68,96 +65,128 @@ you may receive this error:
...
+- **Name**: Anything you want. Something like 'micromasters-local'
+- **Url**: The URL where Micromasters will be running. _OSX Users: this should
+           be http://192.168.99.100:8079_
```
pdpinch: The IP is going to be dependent on docker. `docker-machine env` is the definitive way to get the IP address in use.

README.md
```diff
@@ -68,96 +65,128 @@ you may receive this error:
...
+a template to create your ``.env`` file. For Micromasters to work, it needs 3 values:
+
+- ``EDXORG_BASE_URL``
 
-    edxapp@precise64:~/edx-platform$ pip install ipython
+    Run ``ifconfig eth1`` while ssh'ed into the machine. The
+    appropriate IP is the ``inet addr:`` value, and the LMS server runs
+    on port ``8000`` by default. The value for ``EDXORG_BASE_URL`` should be
+    something like ``http://192.168.33.10:8000``
+- ``EDXORG_CLIENT_ID`` and ``EDXORG_CLIENT_SECRET``
```
pdpinch: This IP is actually [specified in the devstack VagrantFile](https://github.com/edx/configuration/blob/master/vagrant/release/devstack/Vagrantfile#L100). Unless edX changes something, it will always be ``http://192.168.33.10:8000``

README.md
```diff
@@ -68,96 +65,128 @@ you may receive this error:
...
-    Url - the URL of your local Micromasters instance, a URL on port ``8087``.
-    Redirect url: the same URL, but with ``/complete/edxorg/`` at the end. 
-    Client type: Chose "Confidential (Web applications)"
+Note: Your IP address may vary depending on what address Docker 
+assigns to your container.  If the IP address above doesn't work, enter 
+this command:
 
-Make a note of the Client id and the Client secret; you need them to
-configure Micromasters.  Don't forget to save your new client.
+    docker-machine ls
```
pdpinch: this is another solution to my [earlier comment](https://github.com/mitodl/micromasters/pull/313/files#r63185965)

pdpinch:
Some quick comments. I need to take another look at the whole thing together.

README.md
gsidebo: thanks! that was bothering me

gsidebo:
@pdpinch i think i took care of all of the existing comments. have a look

README.md
```diff
@@ -68,96 +65,119 @@ you may receive this error:
...
+The Micromasters codebase contains a ``.env.sample`` file which will be used as
+a template to create your ``.env`` file. For Micromasters to work, it needs 3 values:
 
-      In [1]: from django.contrib.auth.models import User
+- ``EDXORG_BASE_URL``
 
-      In [2]: User.objects.get(username='staff')
-      Out[2]: <User: staff>
+    This value _should_ be ``http://192.168.33.10:8000``. The Vagrant VM IP is hard-coded
+    in the Vagrantfile, and it's unlikely that edX will change that. Tthe LMS server runs
```
pdpinch: typo: `Tthe`

README.md
```diff
@@ -68,96 +65,119 @@ you may receive this error:
...
+#### 1) Install edX
+Run through the
+[instructions provided by edX](https://edx-installing-configuring-and-running.readthedocs.io/en/latest/installation/devstack/install_devstack.html)
+up to and including the **LMS Workflow** section. Note that this section mentions
+some helpful dummy accounts that edX devstack comes preloaded with (eg: staff@example.com).
 
-Start the machine:
+#### 2) Run the machine and SSH into it
+Next, start the VM and create a superuser (or set an existing user to have
+superuser permissions).
```
pdpinch: you actually create the superuser in step 3) so this can be just:

```
Next, start the VM and create a superuser.
```

pdpinch:
I feel like we need a brief final step, to confirm you're done and everything is working. Something like,

```
You should now be able to visit Micromasters in your browser on port `8079` and the IP chosen by docker. You should be able to and login via the edx-platform server you configured.
```

(doesn't have to be exactly that)

README.md
```diff
@@ -1,58 +1,55 @@
...
+- docker-compose (_if you're using Docker, this is required_)
+  - Recommended install: pip `pip install docker-compose`
 
-    pip install docker-compose 
-    docker-compose up
+## Getting Started
 
-### (OSX only) Create your docker machine:
+1. Create and activate a virtualenv in the root folder of the Micromasters project directory.
+1. Install the python dependencies ``pip install -r requirements.txt``
```
aliceriot: could we divide this into Linux and Mac OS sections? The Linux section will only really say 'install docker, docker-compose and do `docker-compose up`' but it would good to be totally clear that these instructions are specific to OSX.

gsidebo: well these two particular steps are for all OSes. after this, a linux user would skip the next two sections (both labeled "OSX only") to "Running edX Locally", then "Docker Container Configuration and Start-up". i _think_ everything that is OSX-specific in those sections is appropriately labeled, but i wouldn't mind more eyes on that

aliceriot: gotcha, for most linux distros though I don't think you would need a virtualenv. on arch, for instance, `docker` and `docker-compose` are the in repos, so all I would have to do if I were starting MM on a new machine would be to install those two packages and then do `docker-compose up`.

gsidebo: ahh i see. i'll move these two steps into the OSX-only section then

aliceriot: I'm excited the documentation is improving, when I started (~month ago) I was like ????

osx_docker_dev_setup.sh
```diff
@@ -0,0 +1,4 @@
+curl -o /usr/local/bin/docker-osx-dev https://raw.githubusercontent.com/brikis98/docker-osx-dev/master/src/docker-osx-dev
```
noisecapella: Can you add `#!/usr/bin/env bash` as the first line?

README.md
```diff
@@ -1,58 +1,55 @@
...
-docker-osx-dev synchronizes host file-system changes with the Docker
-container. If you have [``homebrew``](http://brew.sh/) installed, you can 
-install ``docker-osx-dev`` by typing ``make`` in the root of the Micromasters
-project directory. The make file will install or update Docker using Homebrew
-and go on to install ``docker-osx-dev``.
+This synchronizes host file-system changes with the Docker container.
+
+First, set up ``docker-osx-dev`` via shell script. You can specify a container
+name as an argument or in the **MM_CONTAINER_NAME** env. variable. If the script
+gets neither, the container name is assumed to be 'default'.
```
noisecapella: Can you change this from `default` to `mm`, for consistency with the rest of the file?

gsidebo: i actually got rid of that handling of container name entirely. forgot to update this accordingly. good catch :)

README.md
```diff
@@ -1,58 +1,56 @@
 # Micromasters
 Portal for learners and course teams to access MITx Micromasters programs
 
-## Getting Started
+## Major Dependencies
+- pip, virtualenv, homebrew (OSX only)
+- Postgres >= 9.4
+  - OSX recommended install method: [Postgres.app](http://postgresapp.com/)
+- Python 3
+  - OSX recommended install method: Homebrew `brew install python3`
```
noisecapella: cc @blarghmatey Is this preferred over the app you get from the python website?

README.md
```diff
@@ -1,58 +1,56 @@
...
 
-    docker-machine create default
-    docker-machine start default
-    docker-machine env default
-    eval "$(docker-machine env default)"
+1. Create and activate a virtualenv in the root folder of the Micromasters project directory.
+1. Install the python dependencies ``pip install -r requirements.txt``
 
-These commands create a Docker container named ``default``, start the
+The following commands create a Docker container named ``mm``, start the
```
noisecapella: `mm` is the name of the docker machine, not the container

README.md
```diff
@@ -1,58 +1,56 @@
...
 Portal for learners and course teams to access MITx Micromasters programs
 
-## Getting Started
+## Major Dependencies
+- pip, virtualenv, homebrew (OSX only)
+- Postgres >= 9.4
+  - OSX recommended install method: [Postgres.app](http://postgresapp.com/)
+- Python 3
+  - OSX recommended install method: Homebrew `brew install python3`
+- Docker (_optional, but strongly recommended_)
```
noisecapella: It seems odd that we aren't saying Docker is required here but the rest of our instructions are how to set this up with Docker.

README.md
```diff
@@ -1,58 +1,56 @@
...
 
-Although you can run Micromasters locally with a default sqlite database after
-installing the ``requirements.txt`` file, the recommended way is to 
-use [Docker](https://www.docker.io). Install Docker, then install 
-``docker-compose`` and run the app: 
+## (OSX only) Getting your machine Docker-ready
 
-    pip install docker-compose 
-    docker-compose up
+#### Install Python dependencies and create your docker container:
```
noisecapella: The Docker container should include all setup of the Python dependencies. Since we're using it for Linux and OS X anyway, why not leave this information out?

README.md
gsidebo: oof... ok, i just updated the rest of the README to use what i _think_ are the correct terms for machine and container. wouldn't mind if you took a look and made sure i'm not confusing one for the other

README.md
gsidebo: yeah i thought about that. old doc said that docker wasn't explicitly required, but you're right. we should definitely center around one method of running this app

README.md
gsidebo: so we don't need to run `pip install -r requirements.txt` on the host machine at all?

README.md
```diff
@@ -1,58 +1,56 @@
 # Micromasters
 Portal for learners and course teams to access MITx Micromasters programs
 
-## Getting Started
+## Major Dependencies
+- pip, virtualenv, homebrew (OSX only)
+- Postgres >= 9.4
+  - OSX recommended install method: [Postgres.app](http://postgresapp.com/)
```
noisecapella: For this PR we should assume all work is being in Docker whether on Linux or OS X (this would change for OS X in #307). Can we remove these requirements here, except for docker and docker-compose?

README.md
```diff
@@ -1,58 +1,56 @@
...
+    eval "$(docker-machine env mm)"
 
-This will set up a near production-ready containerized development 
-environment that runs migrations, with the Django development server 
-running on port ``8079``.
+_**NOTE:** Setting an environment variable called
+**MM_MACHINE_NAME** will allow shell/python scripts and one-off commands
+to easily work with your Docker machine. Add this to your shell config
+file (.bashrc, et al) with your desired machine name:
+``export MM_MACHINE_NAME='mm'``_
```
noisecapella: How would this environment variable be used by the user?

README.md
```diff
@@ -1,58 +1,56 @@
...
 
-To run one-off commands, like shell, you can run:
+#### Enable file syncing between the host machine and Docker:
 
-    docker-compose run web python manage.py shell
-
-or to create root user:
-
-    docker-compose run web python manage.py createsuperuser
+Due to issues using Docker in OSX, development is more efficient if you
```
noisecapella: Can you change `Due to issues` to `Due to file notification issues`?

README.md
pdpinch: Let's change it to match with the plan to us docker for everything but web pack -- that is to say, docker is required.

README.md
gsidebo: removed. we had the docker-osx-dev commands in the Makefile and the machine name was hardcoded as 'default'. those lines have since been scrapped, so this env variable isnt really relevant anymore

README.md
gsidebo: got rid of these in the dependencies list and the pip/virtualenv instructions

osx_docker_dev_setup.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
noisecapella: Can you make this file executable?

README.md
blarghmatey: I don't have a particular preference, but I've never actualy installed it with the OSX package from python.org. I've always done it through homebrew and itworked out fine.

osx_docker_dev_setup.sh
gsidebo: good catch

README.md
```diff
@@ -1,58 +1,42 @@
...
-environment that runs migrations, with the Django development server 
-running on port ``8079``.
-
-To run one-off commands, like shell, you can run:
-
-    docker-compose run web python manage.py shell
+with the edX instance.
 
-or to create root user:
+    docker-machine create mm
```
pdpinch: I'm finding I need to specify a driver, i.e. `docker-machine create --driver virtualbox mm`

This is with docker-machine version 0.7.0, build a650a40

gsidebo: i _knew_ i was forgetting something. i had a note about this and it got lost in the pile. i'll add this (and add virtualbox to the major dependencies)

README.md
```diff
@@ -68,96 +52,121 @@ you may receive this error:
...
-    name - your choice of name, mine is ``micromasters-local``.
-    Url - the URL of your local Micromasters instance, a URL on port ``8087``.
-    Redirect url: the same URL, but with ``/complete/edxorg/`` at the end. 
-    Client type: Chose "Confidential (Web applications)"
+In another terminal tab, navigate the the Micromasters directory
+and add a superuser in the now-running Docker container:
 
-Make a note of the Client id and the Client secret; you need them to
-configure Micromasters.  Don't forget to save your new client.
+    docker-compose run web python manage.py createsuperuser
```
pdpinch: Strangely, I'm finding that I need to specify python3 here. 

```
pdpinch$ docker-compose run web python --version
ERROR: Container command 'python' not found or does not exist.
pdpinch$ docker-compose run web python3 --version
Python 3.4.3
```

How is that I have a container without python?

pdpinch:
I'm having a couple of odd problems, but I think this is good to merge nonetheless. 

👍

README.md
gsidebo: i just poked around to see where the docker container is being provisioned with python. looks like it's specified in ./apt.txt (`python3-dev`) and it's being installed via apt-get. i wonder if installing with apt-get fails to link `python` to the `python3` command, which homebrew apparently does for you. @blarghmatey any thoughts?

README.md
```diff
@@ -1,58 +1,42 @@
 # Micromasters
 Portal for learners and course teams to access MITx Micromasters programs
 
-## Getting Started
+## Major Dependencies
+- Python 3
```
noisecapella: This is only installed inside Docker, we shouldn't list it as a requirement here

README.md
noisecapella: @pdpinch this is expected behavior, all commands need `python3` to work properly

gsidebo: Maybe not in scope for this pr, but I would think we would want to allow a user to just use 'python'. Using 'python' all the time on your host machine and needing to use 'python3' on your docker container seems unnecessary, but I may be missing something

Issue #285, opened 2016-05-06
alert user to which fields are unfilled using red text


aliceriot:
When checking for profile completeness we need to alert the user about which fields are currently unfilled.

A couple possibilities:

- move `errors` into the redux store, then the `validateProfileComplete` method could dispatch the errors into the store (so that validation messages would show up on individual fields in the same way they currently do on save)
- put information about missing fields into the dialog, something like:

![message](https://cloud.githubusercontent.com/assets/6207644/15076500/f586dfa4-1377-11e6-9f09-023552e2f442.png)

- probably others?

pdpinch:
To be clear, this is an alert in addition to moving the user to the first step with invalid fields?

I'd like to highlight the fields the same way we do when the user is still within a step. Is that the first option you listed?
> move errors into the redux store, then the validateProfileComplete method could dispatch the errors into the store (so that validation messages would show up on individual fields in the same way they currently do on save)

aliceriot:
yes, that would be the first step.

to accomplish that I think we'd need to refactor how we do validation slightly, currently the validation errors are stored in the redux store but under `userProfile.edit.errors`. The `edit` key on `userProfile` doesn't exist unless we've started editing the profile, so we'd need to change that around a little bit.

aliceriot:
although if we're redirecting people there we'd expect them to be editing things soon anyway, so maybe it would work as-is, with a few changes to the `validateProfileComplete` method.

Ferdi:
I think the modal is a bit too much here. There is only one place they'd end up - where they left off . Why dont we simply redirect there displaying the inline validation errors ?   In addition we can show a simple message at the top of the form  "Please complete your profile to continue".

amir-qayyum-khan:
@aliceriot can you guide me where to start?

aliceriot:
@amir-qayyum-khan sure - I think all you should need to do is that when we check for profile completeness (see the `validateProfileComplete` method in `util/util.js` and the `requireCompleteProfile` method on `App.js`) if the profile isn't complete we dispatch the `START_PROFILE_EDIT` and `UPDATE_PROFILE_VALIDATION` actions. The `errors` object that holds the validation results is passed down into each profile tab from its parent component, so the errors should then just show up.

aliceriot:
does that make sense? lmk if you have more questions

amir-qayyum-khan:
yeah, thnaks

PR #314, opened 2016-05-13
Added highlight approach to missing fields instead of popup

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/285

#### What's this PR do?
- Removed popup message for missing contents in fields
- Added highlight validation.

see Apps.js and utils.js

- Left any filed content missing on profile tab
- Move to last tab and press save and continue
- You will be navigate to profile where you have missing contents.
- The filed with missing contents will be highlight

@aliceriot @noisecapella @pdpinch 

- screen shot 2016-05-16 at 4 04 11 pm
- screen shot 2016-05-16 at 4 04 21 pm
- screen shot 2016-05-16 at 4 04 11 pm


static/js/containers/App.js
```diff
@@ -88,17 +89,14 @@ class App extends React.Component {
...
       userProfile.getStatus === FETCH_SUCCESS &&
       profile.agreed_to_terms_of_service &&
       !PROFILE_REGEX.test(pathname) &&
       !complete
     ) {
-      const { url, title, text } = info;
-      dispatch(updateDialogText(text));
-      dispatch(updateDialogTitle(title));
-      dispatch(setDialogVisibility(true));
+      dispatch(updateProfileValidation(errors));
```
aliceriot: the change you need to make is right here.

before we can dispatch the `updateProfileValidation` action we need to have already dispatched the `startProfileEdit` action - otherwise there is no in-progress 'edit' object (which we keep separate in the store from the object pulled from the API) to store the validation errors in.

So with what you have here they aren't being cleared, they're simply being dropped on the floor.

aliceriot: all you need to do is add

```js
import { startProfileEdit } from '../actions/index';
```
up top, and do

```js
  dispatch(startProfileEdit);
```
right here (before dispatching the validation updates)

static/js/util/util.js
```diff
@@ -228,40 +228,33 @@ complete profile consists of:
 */
 export function validateProfileComplete(profile) {
   let reqFields = [];
+  let validationMessages = {};
```
aliceriot: do you need to build up this object of all the validation messages? in the function body below we return any time we find validation errors relevant to a particular profile tab, so I think for each you should be able to just do something like:

```js
let errors = validateProfile(profile, reqFields, PersonalTab.defaultProps.validationMessages);
```

then we check to see if that's empty, and if not, we return right there. So in any case where we move on to the next tab (professional) the validation messages for the personal tab won't be necessary to check.

static/js/util/util.js
```diff
@@ -228,40 +228,33 @@ complete profile consists of:
...
+
+  let errors = validateProfile(profile, reqFields, validationMessages);
   if ( !_.isEqual(errors, {}) ) {
-    return ([false, {
-      url: "/profile/personal",
-      title: "Personal Info",
-      text: "Please complete your personal information.",
-    }]);
+    return ([false, "/profile/personal", errors]);
   }
```
aliceriot: can you add a check for completeness of the education tab? you should be able to follow the same pattern we've got here for the others.

static/js/util/util.js
aliceriot: We probably don't need to build up `reqFields` either, hmmmm

static/js/util/util.js
```diff
@@ -227,44 +228,72 @@ complete profile consists of:
     `currently employed == 'yes'`
 */
 export function validateProfileComplete(profile) {
```
aliceriot: Can you undo the changes to this function? this is a lot of extra code that doesn't really give us anything. we always want to redirect the user to the first tab (personal -> education -> professional -> privacy) which has an invalid field on it, so we don't need to check the professional tab if we find something missing on the personal tab (which is why the function previously returned any time it found any validation error).

I think the only thing we really need to change to get the functionality we want is passing in the validation messages.

static/js/containers/ProfilePage.js
```diff
@@ -40,9 +39,7 @@ class ProfilePage extends React.Component {
       dispatch(startProfileEdit());
     }
     return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
-      dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
-        dispatch(clearProfileEdit());
-      });
+      dispatch(saveProfile(SETTINGS.username, profile));
```
aliceriot: We don't want to remove the `clearProfileEdit` dispatch here. We want to clear the edit object after we successfully save the edits.

the `clearProfileEdit` action here is only going to be dispatched after we've finished editing and we've successfully saved the object (well, after the `Promise` has been resolved), so it shouldn't affect the changes you're making.

static/js/containers/App.js
```diff
@@ -88,17 +88,16 @@ class App extends React.Component {
...
       userProfile.getStatus === FETCH_SUCCESS &&
       profile.agreed_to_terms_of_service &&
       !PROFILE_REGEX.test(pathname) &&
       !complete
     ) {
-      const { url, title, text } = info;
-      dispatch(updateDialogText(text));
-      dispatch(updateDialogTitle(title));
-      dispatch(setDialogVisibility(true));
+
```
aliceriot: whitespace

static/js/containers/App.js
```diff
@@ -12,9 +12,9 @@ import {
   fetchDashboard,
   clearDashboard,
   clearUI,
-  updateDialogText,
-  updateDialogTitle,
   setDialogVisibility,
```
aliceriot: we don't need this action anymore either (`setDialogVisibility`)

static/js/util/util.js
aliceriot: (previously we were just checking that the returned `errors` objects was non-empty, so we didn't care about the specific messages, but now we want them to dispatch to the store and so on)

static/js/containers/ProfilePage.js
amir-qayyum-khan: @aliceriot it is navigating user to ``personal tab`` before it clears edit profile. 
Thats why when user lands on ``personal profile tab`` all the error messages get clear.

amir-qayyum-khan: @aliceriot it seems action/saveProfile is async tasks. The task complete after we land on first tab (personal tab). Because of this error message clear from background. there is no point to have ``-        dispatch(clearProfileEdit());`` here.

aliceriot: Sorry, I guess I'm not quite understanding what the problem is. I pulled down your code and the problem seems to me to be that when we run the `requireCompleteProfile` method (in `App.js`, called on `componentDidMount` and `componentDidUpdate`) we're still not dispatching the `startProfileEdit` action - I don't understand exactly why that's not happening.

We shouldn't need to make any modification of the code in this method here (`saveProfile`) since we won't be running it when we load the tab.

Does that make sense? This code here (where we dispatch the `clearProfileEdit` action) is only run when we save a tab (click on the 'save and continue' button or w/e).

.eslintrc
```diff
@@ -32,6 +32,7 @@
     "camelcase": [2, {
       "properties": "never"
     }],
+   "no-console": [0] //allow use of javascript console in tests.
```
aliceriot: I think we'd like to keep this set to `[2]`, @noisecapella thoughts?

aliceriot: also if you'd like to violate a rule in one place eslint lets you temporarily disable a rule with a comment (similar to pylint). see the docs: http://eslint.org/docs/user-guide/configuring#configuring-rules

static/js/containers/App.js
```diff
@@ -88,49 +85,19 @@ class App extends React.Component {
...
       userProfile.getStatus === FETCH_SUCCESS &&
       profile.agreed_to_terms_of_service &&
       !PROFILE_REGEX.test(pathname) &&
       !complete
     ) {
-      const { url, title, text } = info;
-      dispatch(updateDialogText(text));
-      dispatch(updateDialogTitle(title));
-      dispatch(setDialogVisibility(true));
+      dispatch(startProfileEdit);
```
aliceriot: Found the problem - syntax error. `startProfileEdit` is a function, so we need to change this line to:

```js
dispatch(startProfileEdit())
```

then it should work fine.

static/js/containers/ProfilePage.js
amir-qayyum-khan: @aliceriot  Actually the scenario is when user press ``save and continue`` on privacy tab it should take user to ``personal tab``.
- And it does this well.
- The issue is when it take user to ``personal tab`` after few milisec the highlighted errors gets remove.
- And my investigation shows that it is because of `` dispatch(clearProfileEdit());``.
- It seems, app is saving profile on my machine after it navigate to ``personal tab``. (I mean to say save request complete after few milli sec and at that time it is already on ``personal tab`` and edit mode gets clear.

aliceriot: See the comment above - the problem was not with the `clearProfileEdit` action here, but a syntax error. Since we weren't actually dispatching the `startProfileEdit` action (`startProfileEdit` is a function that returns the action - see the definition in `actions/index.js`) there wasn't an `errors` object to save our validation errors into, and they were being dropped on the floor.

static/js/containers/App.js
amir-qayyum-khan: let me try

static/js/containers/ProfilePage.js
aliceriot: I just made the change above locally, and if I have a validation error on the `personal` tab and click 'save and continue' on the privacy tab I see the issue you're talking about.

static/js/containers/App.js
amir-qayyum-khan: @aliceriot same issue. may be you cant produce it because my machine is slow (13" osx i5 and 8gb ram)?

static/js/containers/ProfilePage.js
aliceriot: you're right, the problem is that when the AJAX for the PATCH to save the profile comes back we dispatch the `clearProfileEdit` action - I'm not sure exactly what the best way forward is here, since we do want to clear the profile edit object after we save. I'll talk to george a little bit and get back to you!

static/js/containers/App.js
aliceriot: not quite the same issue, with this change (`startProfileEdit` -> `startProfileEdit()`) I do actually see validation errors (before they weren't showing up at all).

this is a different issue than the profile PATCH promise async issue.

amir-qayyum-khan: @aliceriot error object looks good. I checked it. 
there is on error on console that appears on pressing ``save and continue`` on personal tab

**undefined:1 Uncaught (in promise) undefined**

aliceriot: yeah, I think that predates this change

.eslintrc
noisecapella: I agree, we should not allow `console` in our code. If we're using it in tests for debugging we should remove it before the code is merged

aliceriot:
OK I just spoke with george about this, I think this is what we need to do.

First off, the case where a user manages to get to the privacy tab with an empty field on another tab is definitely an edge case - in the final version of the profile UI the tabs arent going to be clickable, so it's not going to be possible to advance without filling out all fields.

Still, we can fix the error in the following way.

In `static/js/containers/ProfilePage.js` we need to refactor the `saveProfile` method slightly - we need to be able to chain off off a promise (the whole function needs to return a promise) so that we can ensure that our route transition happens synchronously after the `saveProfile` action has been dispatch, and that promise has been resolved.

So we just need `saveProfile` (on `ProfilePage.js`) to return a promise, like this:

```js
  saveProfile(isEdit, profile, requiredFields, messages) {
    if (!isEdit) {
      // Validation errors will only show up if we start the edit
      dispatch(startProfileEdit());
    }
    return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
      return dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
        dispatch(clearProfileEdit());
        return Promise.resolve();
      });
    });
  }
```

then in `static/util/profile_edit.js`, in our `saveAndContinue` function, we will actually guarantee that the route transition will occur after the PATCH promise is resolved when we do:

```js
  saveProfile( profile, fields, validationMessages).then(() => {
    this.context.router.push(next);
  });
```

that should be all we need to do.

amir-qayyum-khan:
@aliceriot please review.

static/js/containers/ProfilePage.js
```diff
@@ -42,6 +42,7 @@ class ProfilePage extends React.Component {
     return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
       dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
         dispatch(clearProfileEdit());
+        return Promise.resolve();
```
aliceriot: the race condition we discovered is still occurring with this change. I posted a comment on the pr showing the correct way to change this function:

```js
  saveProfile(isEdit, profile, requiredFields, messages) {
    if (!isEdit) {
      // Validation errors will only show up if we start the edit
      dispatch(startProfileEdit());
    }
    return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
      return dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
        dispatch(clearProfileEdit());
        return Promise.resolve();
      });
    });
  }
```

basically, each of the chained arrow functions needs to return a promise, this is what gives us guarantees of the statements being executed synchronously.

you're change returns `Promise.resolve()` at the end, but because the first arrow function (opened on line 42) doesn't return a promise we aren't actually going to be chaining them together. if you change this function to what I have above we should get the behavior that we want, which is the actions happening strictly in this order:

```
validateProfile -> saveProfile -> clearProfileEdit -> return Promise.resolve()
```

then, with the guarantees around strict order of execution that this gives us, we can ensure that we won't trigger a route transition (to `/dashboard`) before the `PATCH` request comes back. The problem before was this line (in the `saveAndContinue` function, defined in `util/profile_edit.js`:

```js
  saveprofile( profile, fields, validationmessages).then(() => {
    this.context.router.push(next);
  });
```

before we weren't chaining promises, so we didn't actually have any kind of guarantee that our callback (`() => this.context.router.push(next);`) would execute after all of the actions in `saveProfile` had finished dispatching. With the change above we do have such a guarantee - so we can do the route transition ( `this.context.router....`) safely, knowing that we're all done with saving the profile, and then we can safely dispatch the `startProfileEdit` and `updateValidation` actions when we call the `requireCompleteProfile` function.

Does this all make sense?

noisecapella: Now that I think about it, the `return Promise.resolve();` should not be necessary at all. `then` should always return a resolved promise by default, and we know that `dispatch(clearProfileEdit` will never need to return a rejected promise, so it should just work. Need more :coffee: when I talk about these things :)

aliceriot: haha, no worries :)

so to confirm, this is what we need to do? 

```js
  saveProfile(isEdit, profile, requiredFields, messages) {
    if (!isEdit) {
      // Validation errors will only show up if we start the edit
      dispatch(startProfileEdit());
    }
    return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
      return dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
        dispatch(clearProfileEdit());
      });
    });
  }
```

aliceriot: yeah, I just confirmed - the above does what we want. Nice!

noisecapella: Yes, that looks correct

static/js/util/util_test.js
```diff
@@ -325,10 +325,17 @@ describe('utility functions', () => {
...
+        first_name: 'Given name is required',
+        last_name: 'Family name is required',
+        preferred_name: "Preferred name is required",
+        gender: "Gender is required",
+        preferred_language: "Preferred language is required",
+        city: "City is required",
+        country: "Country is required",
+        birth_city: "City is required",
+        birth_country: "Country is required",
+        date_of_birth: "Date of birth is required",
```
aliceriot: this object literal here is exactly the validation messages on PersonalTab (`PersonalTab.defaultProps.validationMessages`) - we might want to just pull that in here instead of explicitly declaring it.

static/js/containers/App.js
```diff
@@ -88,49 +85,19 @@ class App extends React.Component {
       location: { pathname },
       dispatch,
     } = this.props;
-    let [ complete, info ] = validateProfileComplete(profile);
+    let [ complete, url, errors] = validateProfileComplete(profile);
```
aliceriot: we should probably prefer `const` over `let` here, could you change that real quick?

static/js/util/util.js
```diff
@@ -227,44 +228,60 @@ complete profile consists of:
...
+    PersonalTab.defaultProps.validationMessages
+  );
+
+  if (!_.isEqual(errors, {})) {
+    return [false, '/profile/personal', errors];
   }
 
   // check professional tab
-  if ( _.isArray(profile.work_history) && !_.isEmpty(profile.work_history) ) {
+  if ( _.isArray(profile.work_history) && !_.isEmpty(profile.work_history)) {
```
aliceriot: total style nit: for `if .. else` conditions I think it's best to do one of the following:

```js
if ( condition && otherCondition ) {
  ...
}
```
or 
```js
if (condition && otherCondition) {
  ...
}
```

just consistency in the spacing of the wrapping parentheses.

aliceriot:
functionality is looking good, just a few small comments. back to you!

giocalitri:
needs to be rebased

static/js/util/util_test.js
amir-qayyum-khan: actually it is different. please see the ``PersonalTab.defaultProps.validationMessages``.The **required** string is missing in ``validationMessages``.

```validationMessages: {
      'first_name': "Given name",
      'last_name': "Family name",
      'preferred_name': "Preferred name",
      'gender': "Gender",
      'preferred_language': "Preferred language",
      'city': "City",
      'state_or_territory': 'State or Territory',
      'country': "Country",
      'birth_city': 'City',
      'birth_state_or_territory': 'State or Territory',
      'birth_country': "Country",
      'date_of_birth': "Date of birth"
    }```

Comment (amir-qayyum-khan) [inline static/js/util/util_test.js]: i made a change for that
> let errors = {};
      _.forIn(new Foo, function(value, key) {
        _.set(errors, keySet, `${key} is required`);
      });
      const expectation = [false, "/profile/personal", errors];

File: static/js/util/util_test.js [hunk 14]
```diff
@@ -324,12 +325,12 @@ describe('utility functions', () => {
...
-      let expectation = [false, {
-        url: "/profile/personal",
-        title: "Personal Info",
-        text: "Please complete your personal information.",
-      }];
+    it('should return fields for an empty profile', () => {
+      let errors = {};
+      _.forIn(PersonalTab.defaultProps.validationMessages, (value, key) => {
+        _.set(errors, key, `${value} is required`);
+      });
```
aliceriot: this works, but I think this might be cleaner:

```js
let errors = Object.assign({}, ...Object.entries(PersonalTab.defaultProps.validationMessages).map(
  ([k,v]) => ({[k]: `${v} is required`})
));
```

we'd need to add the `Object.entries` polyfill though, by adding the following to `static/js/global_init.js`:

```js
import entries from 'object.entries';
if (!Object.entries) {
  entries.shim();
}
```

your call.

static/js/util/util_test.js
```diff
@@ -379,10 +378,12 @@ describe('utility functions', () => {
...
-      let expectation = [ false, {
-        url: "/profile/professional",
-        title: "Professional Info",
-        text: "Please complete your work history information.",
+      let expectation = [false, "/profile/professional", {
+        work_history: [
+          {
+            country: "Country is required"
+          }
+        ]
```
aliceriot: style nit (ignore at your discretion): I would format this in this way instead:

```js
let expectation = [false, "/profile/professional", {
  work_history: [{country: "Country is required"}]
}];
```

aliceriot:
I added a couple of comments, this is getting close!

aliceriot:
@noisecapella @amir-qayyum-khan I'm wondering if we should delete the UI actions relating to setting the dialog text and so on in this PR, we aren't using them anywhere else, and while I had conceived of the dialog as being more widely useful we're already deleting the helper function from `App.js` here.

amir-qayyum-khan:
@aliceriot lets create a separate PR for that? what do you say?

aliceriot:
sure, that sounds good. no rush, code deletion can happen whenever :)

aliceriot:
something isn't quite right - if I set my profile's `first_name` to `None` and then reload the application in the browser (by typing in `http://localhost:8079/dashboard`) I see the following:

no_error

aliceriot:
I suspect this has to do with the changes introduced in https://github.com/mitodl/micromasters/pull/357

amir-qayyum-khan:
@aliceriot please review.

aliceriot:
functionality looks good, errors are showing up correctly. looking at the code now.

aliceriot:
:+1:

Issue #234, opened 2016-04-26
handle http errors in dashboard UI


giocalitri:
Now if there is a 500 error in the backend there is no information in the front-end, for example when the user has multiple social auth objects. 

The error message should include a link to e-mail mitx-support@mit.edu, with the text of the error message. 

If the error is coming from edX, we should still display the dashboard, but with a message that some data is missing.

giocalitri:
The scope of this story will be to handle the errors ONLY in the front end and leave the data as is.
We will need an additional story to make the dashboard REST API more robust and send some data back in case there is only a problem with edX.

PR #317, opened 2016-05-13
Fixed profile submission to wait for a 200

### LINKED ISSUE #270: Save and continue buttons should wait for 200 from PATCH before continuing
Needs UI for any error message

noisecapella:
Fixes #270

#### What's this PR do?
When we add `catch` to a `Promise` without returning another rejected `Promise`, future promises are resolved with that return value. This meant that even when the API function returned the rejected promise, the action dispatcher function returned a resolved promise.

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -69,7 +69,11 @@ export const saveProfile = (username, profile) => {
     dispatch(requestPatchUserProfile());
     return api.patchUserProfile(username, profile).
       then(newProfile => dispatch(receivePatchUserProfileSuccess(newProfile))).
-      catch(() => dispatch(receivePatchUserProfileFailure()));
+      catch(e => {
+        dispatch(receivePatchUserProfileFailure());
+        // propagate exception
+        return Promise.reject(e);
+      });
   };
 };
 export const updateProfileValidation = errors => ({
@@ -94,7 +98,11 @@ export function fetchUserProfile(username) {
     dispatch(requestGetUserProfile());
     return api.getUserProfile(username).
       then(json => dispatch(receiveGetUserProfileSuccess(json))).
-      catch(()=> dispatch(receiveGetUserProfileFailure()));
+      catch(e => {
+        dispatch(receiveGetUserProfileFailure());
+        // propagate exception
+        return Promise.reject(e);
+      });
   };
 }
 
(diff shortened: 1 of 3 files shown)


static/js/containers/ProfilePage.js
```diff
@@ -40,9 +40,9 @@ class ProfilePage extends React.Component {
       dispatch(startProfileEdit());
     }
     return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
-      dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
-        dispatch(clearProfileEdit());
-      });
+      return dispatch(saveProfile(SETTINGS.username, profile));
+    }).then(() => {
+      return dispatch(clearProfileEdit());
```
aliceriot: what's the reason for changing these two lines? is there a reason we can't do this?

```js
return dispatch(validateProfile(profile, requiredFields, messages)).then(() => {
  return dispatch(saveProfile(SETTINGS.username, profile)).then(() => {
    dispatch(clearProfileEdit());
  });
});
```

still trying to wrap my head around promises a little more, and understand how this change relates to the change over in #314.

static/js/actions/index.js
```diff
@@ -137,6 +145,10 @@ export function fetchDashboard() {
     dispatch(requestDashboard());
     return api.getDashboard().
       then(dashboard => dispatch(receiveDashboardSuccess(dashboard))).
-      catch(() => dispatch(receiveDashboardFailure()));
+      catch(e => {
+        dispatch(receiveDashboardFailure());
+        // propogate exception
+        return Promise.reject(e);
+      });
```
aliceriot: does `e` have error code information and so on?

aliceriot:
really one question, back to you!

static/js/containers/ProfilePage.js
noisecapella: Just a stylistic difference, I'll change it to what we're doing in #314 

Technically the `return dispatch(clearProfileEdit())` is different but we don't use the returned value anyway

static/js/actions/index.js
noisecapella: It depends. If a `throw` happened in either `getDashboard` or in the `then()` immediately afterward, it would have the exception. But if there's a rejected promise then `e` is just whatever the promise value happened to be. In our `fetchJSONWithCSRF` function, if we get a non-2xx status code the value would be the content of the response

aliceriot: ok cool, I was just wondering if it would make sense to surface this to the user in some way - like an 'error: connection troubles' or 'error: not authorized', etc. maybe something to worry about later.

aliceriot:
LGTM :+1: :ship: :shipit:

static/js/actions/index.js
noisecapella: Yeah, I think we would want to do it when we do #234

PR #325, opened 2016-05-16
added edx_course_key to courserun list display in admin

### LINKED ISSUE #324: admin enhancements
As an admin user, it would be handy to be able to see the edx course key in the courserun list view. 

Since so many of the course runs have the same name, the only way to differentiate them is by course key.

pdpinch:

fixes #324 

#### What's this PR do?

Adds edx_course_key to the courserun list display in admin.
Changes the order of fields in course list display, to make it consistent with other admin lists.

By the time the reviewer starts, they will be done. 

Take a look at /admin/courses/courserun/ and /admin/courses/course/

There's a lot of courseruns with the same name, so we need some way to differentiate them in list view. 

learning to play guitar

### DIFF
--- a/courses/admin.py
+++ b/courses/admin.py
@@ -1,36 +1,58 @@
-"""Admin views for Courses & Programs"""
+"""
+Admin views for Courses & Programs
+"""
+
 from django.contrib import admin
-from .models import Course, CourseRun, Program
+
+from courses.models import Course, CourseRun, Program
 
 
-class ProgramAdmin(admin.ModelAdmin):
-    """ModelAdmin for Programs"""
-    list_display = ('title', 'live',)
-    list_filter = ('live',)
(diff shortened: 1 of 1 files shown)


courses/admin.py
```diff
@@ -17,15 +17,15 @@ class CourseRunInline(admin.StackedInline):
 
 class CourseAdmin(admin.ModelAdmin):
     """ModelAdmin for Courses"""
-    list_display = ('position_in_program', 'title',)
+    list_display = ('title', 'position_in_program',)
```
giocalitri: given that we are touching this, I would add the program in this list and the modify the ordering below to be first by program and then by position in program.

giocalitri:
Just one comment

pdpinch:
Thanks Giovanni. This helps a lot. 👍

PR #336, opened 2016-05-16
Marked profile as incomplete when it's being filled out and complete when past the last tab

### LINKED ISSUE #233: When profile is completed for first time, set email_optin and filled_out to True

noisecapella:
Fixes #233 

#### What's this PR do?
If the user clicks the save button on a profile page, it will send a `filled_out`: false along with the rest of the data, which will force them to complete the rest of the profile. When they get to the last tab and click Save, it will send `filled_out`: true and `email_optin`: true.

### DIFF
--- a/static/js/util/ProfileTab.js
+++ b/static/js/util/ProfileTab.js
@@ -56,7 +56,8 @@ class ProfileTab extends React.Component {
   }
 
   saveAndContinue = () => {
-    saveAndContinue.call(this, this.constructor.prototype.validation).then(() => {
+    let lastTab = this.nextUrl === "/dashboard";
+    saveAndContinue.call(this, this.constructor.prototype.validation, lastTab).then(() => {
       this.context.router.push(this.nextUrl);
     });
   }
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -155,9 +155,6 @@ class ProfileSerializer(ModelSerializer):
             'edx_level_of_education',
             'education'
         )
-        read_only_fields = (
-            'filled_out',
-        )
 
 
 class ProfileLimitedSerializer(ModelSerializer):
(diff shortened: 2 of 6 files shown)


static/js/components/PrivacyTab.js
```diff
@@ -25,6 +26,12 @@ class PrivacyTab extends ProfileTab {
     }
   };
 
+  saveAndContinue = () => {
+    saveAndContinue.call(this, null, true).then(() => {
+      this.context.router.push(this.nextUrl);
+    });
```
aliceriot: I added an instance method ('saveAndContinue`) to `ProfileTab` that we might want to use instead here - we could refactor that method to take an optional boolean to signify it's the last tab (something like `lastTab`? haha probs not a good name).

noisecapella:
Needs tests

noisecapella:
Ok, tests are in place

static/js/util/ProfileTab.js
```diff
@@ -56,7 +56,9 @@ class ProfileTab extends React.Component {
   }
 
   saveAndContinue = () => {
-    saveAndContinue.call(this, this.constructor.prototype.validation).then(() => {
+    // this tab is the last tab if it redirects to the dashboard
+    let lastTab = this.nextUrl === "/dashboard";
+    saveAndContinue.call(this, this.constructor.prototype.validation, lastTab).then(() => {
```
aliceriot: this is slick! what do you think about deleting the comment? the line (`let lastTab = ...`) is self-explanatory to me.

noisecapella: Will do

aliceriot:
LGTM (assuming the tests pass)

:+1: :100: :shipit:

PR #339, opened 2016-05-17
Added requirement for state_or_territory on PersonalTab

### LINKED ISSUE #288: state/territory is required in all profile forms
Since our country data includes at least one state/territory for every country, we can make state/territory a required field. 

The UX will be better once we change the order of city, state, country as described in #281

aliceriot:

closes #288 

#### What's this PR do?

This adds a requirement that the `state_or_territory` and `birth_state_or_territory` fields be filled out on the personal tab. We're currently already requiring the equivalent fields on the employment and education tabs.

Changes in `PersonalTab`.

Change the profile record in the database to have `state_or_territory = None`, save it, and then reload the dashboard. When you try to advance past the personal tab you should see a warning that the `state_or_territory` field is required. Same thing should happen for `birth_state_or_territory`.

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -26,8 +26,10 @@ class PersonalTab extends ProfileTab {
       ['gender'],
       ['preferred_language'],
       ['city'],
+      ['state_or_territory'],
       ['country'],
       ['birth_city'],
+      ['birth_state_or_territory'],
       ['birth_country'],
       ['date_of_birth'],
     ],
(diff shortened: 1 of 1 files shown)


pdpinch:
Works as advertised, although we still have the odd behavior that if the users clears the value after saving, the component reverts to the saved value. That's out of scope for this story though. 

:+1:

pdpinch:
Oh, hold that thumb until tests pass...

aliceriot:
yeah, I think that's a quirk of the material-ui autoselect. which, incidentally, is why I suggested changing the database value and reloading the dashboard, because I couldn't find another way to make the field blank! haha.

pdpinch:
I happened to have a profile with the fields already empty and could test with that.

PR #340, opened 2016-05-17
Changed all class properties to use es7 class property syntax

### LINKED ISSUE #330: Convert all remaining class.propTypes and etc to use class property syntax

aliceriot:

closes #330 

#### What's this PR do?

This moves all class properties to the newer class property syntax, which uses the `static` keyword and looks like this:

```js
class Foo {
  static someClassProperty = 12;
}
```

as opposed to the older syntax:

```js
class Foo {
  ...
}

Foo.someClassProperty = 12;
```

the newer syntax moves the class properties into the body of the class declaration, meaning that the property declarations are closer to the code they relate to. On classes that extend `React.Component` we can declare our `propTypes` and `defaultProps` as class properties:

```js
class Foobar extends React.Component {
  static propTypes = {
    foobar:  React.PropTypes.func.isRequired,
  }

  static defaultProps = {
    foobar: () => console.log("FOOBAR");
  }
}
```

It's a bit more readable and just generally nicer to work with. This syntax is a [stage-1 proposal](https://github.com/jeffmo/es-class-fields-and-static-properties), which we're able to use by adding the `stage-1` preset to Babel (see `.babelrc`).

review the changes, make sure I didn't make any typos :P

click around and make sure everything works!

### DIFF
--- a/static/js/components/CourseList.js
+++ b/static/js/components/CourseList.js
@@ -9,6 +9,10 @@ import {
 import { STATUS_PASSED } from '../constants';
 
 class CourseList extends React.Component {
+  static propTypes = {
+    dashboard: React.PropTypes.object.isRequired,
+  };
+
   render() {
     const { dashboard } = this.props;
 
@@ -70,9 +74,5 @@ class CourseList extends React.Component {
     </div>;
   }
 }
-
-CourseList.propTypes = {
-  dashboard: React.PropTypes.object.isRequired,
-};
 
 export default CourseList;
--- a/static/js/components/Dashboard.js
+++ b/static/js/components/Dashboard.js
@@ -4,6 +4,10 @@ import CourseList from './CourseList';
 import UserImage from './UserImage';
 
 class Dashboard extends React.Component {
+  static propTypes = {
+    profile:    React.PropTypes.object.isRequired,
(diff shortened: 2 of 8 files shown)


gsidebo:
general question: does webpack (or some other tool) take this es6/7 syntax and 'translate' it to something that is compatible with all the browsers we support? keep in mind i dont even know what browsers we support

aliceriot:
Yeah, we use babel for this purpose.

aliceriot:
you can read more here: https://babeljs.io/

aliceriot:
Babel is a transpiler, so basically it implements the ES2015 (and newer) syntax in standards-compliant ES5 JS. Webpack has some knowledge of ES2015 syntax, but really right now (until the 2.0 release hits) it pretty much consumes ES5 JS, so we basically chain them together - babel transpiles the code we write to clean ES5, and Webpack does the minification and bundling (for production) or the hot-reload dev stuff.

aliceriot:
We also polyfill prototype methods that don't yet have wide browser support (like `Object.entries`).

aliceriot:
also I like talking about JS a lot so if you have question hmu haha

gsidebo:
👍  looks good

Issue #281, opened 2016-05-06
Profile: country > state > city


pdpinch:
Because choosing the country populates the state / territory drop-down, change the order of these fields.

Issue #318, opened 2016-05-13
Use state select in education profile view


noisecapella:
can be addressed the same time as #281 and #288

PR #341, opened 2016-05-17
Changed location order to Country -> State -> City

aliceriot:

#281 

#### What's this PR do?

Changes the order of the location fields on the personal and education tabs (they were changed in #312 for employment).

The changes are in `PersonalTab.js` and `EducationTab.js`.

Make sure that you can edit the location fields on both of these tabs, and that nothing got mushed up.

### DIFF
--- a/static/js/components/EducationTab.js
+++ b/static/js/components/EducationTab.js
@@ -210,13 +210,13 @@ class EducationTab extends ProfileTab {
         </Grid>
         <Grid>
           <Cell col={4}>
-            {this.boundTextField(keySet('school_city'), 'City')}
+            {this.boundSelectField(keySet('school_country'), 'Country', this.countryOptions)}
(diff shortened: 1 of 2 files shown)


giocalitri:
it looks good
I noticed that the `state or region` in the education tab does not provide suggestions as in the personal tab. 
I suppose this problem is not related, but I would like a confirmation

noisecapella:
@giocalitri it's not related, I filed an issue earlier for this: #318

noisecapella:
Actually, we should probably just knock out that problem too, as the fix is a one liner

giocalitri:
I am fine keeping it separated, since you already filed an issue for it.
👍

giocalitri:
please next time put `closes` or `fixes` <story_number>, so the story can be closed when the PR is merged

PR #345, opened 2016-05-17
Implemented UI for showing course runs in dashboard

### LINKED ISSUE #223: Add course not-passed to Dashboard UI

noisecapella:
Fixes #223 

#### What's this PR do?
Shows course runs in dashboard and handles the not passed behavior more accurately. 

Add a course which is not passed to your local server, then view the dashboard. You should see a red circle next to the course, and a expander link which would allow you to see individual runs.

screenshot from 2016-05-18 13-43-09

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,6 +32,7 @@
     "camelcase": [2, {
       "properties": "never"
     }],
+    "guard-for-in": [2]
   },
   "env": {
     "es6": true,
--- a/static/js/components/CourseList.js
+++ b/static/js/components/CourseList.js
@@ -1,20 +1,29 @@
 import React from 'react';
 import _ from 'lodash';
 import Button from 'react-bootstrap/lib/Button';
+import Grid, { Cell } from 'react-mdl/lib/Grid';
 
+import { toggleDashboardExpander } from '../actions/ui';
 import {
   makeCourseStatusDisplay,
   makeCourseProgressDisplay,
+  makeRunStatusDisplay,
 } from '../util/util';
-import { STATUS_PASSED } from '../constants';
+import {
+  STATUS_PASSED,
+  DASHBOARD_COURSE_HEIGHT,
+  DASHBOARD_RUN_HEIGHT,
+} from '../constants';
 
 class CourseList extends React.Component {
   static propTypes = {
     dashboard: React.PropTypes.object.isRequired,
+    expander: React.PropTypes.object.isRequired,
+    dispatch: React.PropTypes.func.isRequired,
   };
 
   render() {
-    const { dashboard } = this.props;
+    const { dashboard, dispatch, expander } = this.props;
 
     let programs = dashboard.programs.map(program => {
       let sortedCourses = _.sortBy(program.courses, 'position_in_program');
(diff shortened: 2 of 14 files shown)


noisecapella:
I need to remove the carot link when there are no runs for a course

aliceriot:
functionality looks good, going to start reading the code now.

.eslintrc
```diff
@@ -32,6 +32,7 @@
     "camelcase": [2, {
       "properties": "never"
     }],
+    "guard-for-in": [2]
```
aliceriot: :100:

static/js/actions/index.js
```diff
@@ -132,3 +132,12 @@ export function fetchDashboard() {
...
 }
+
+
+// dashboard expander actions
+export const TOGGLE_DASHBOARD_EXPANDER = 'TOGGLE_DASHBOARD_EXPANDER';
+
+export const toggleDashboardExpander = (courseId, newValue) => ({
+  type: TOGGLE_DASHBOARD_EXPANDER,
+  payload: { courseId, newValue }
+});
```
aliceriot: do you think it might make sense to stick these actions in with the `ui` actions? I had imagined the `ui` top level key on the store being for storing miscellaneous pieces of UI state (dialog open / closed, etc).

static/js/components/CourseList.js
```diff
@@ -29,37 +43,75 @@ class CourseList extends React.Component {
...
+                className="course-run-title"
+              >
+                {run.title}
+              </Cell>
+              <Cell col={4} className="course-run-status">
+                {makeRunStatusDisplay(run)}
+              </Cell>
+            </Grid>;
+          });
+        }
```
aliceriot: I think you should be able to do this for the arrow function here:

```js
run => (
  <Grid ...>
    ...
  </Grid>
);
```

since I believe JSX is transformed into a single expression we're ok doing an implicit-return style arrow function (we don't need curly braces around the function body or the `return` keyword).

static/js/components/CourseList.js
```diff
@@ -29,37 +43,75 @@ class CourseList extends React.Component {
...
+
+        let expanderClass = "glyphicon glyphicon-chevron-down";
+        if (expander[course.id]) {
+          expanderClass = "glyphicon glyphicon-chevron-up";
+        }
```
aliceriot: suggestion: change this to

```js
let setExpanderIcon = () => (
  expander[course.id] ? "glyphicon-chevron-up" : "glyphicon-chevron-down"
);
```

then when we need this class later just do

```js
className={`glyphicon ${setExpanderIcon}`}
```

or alternatively I guess we could just do

```js
let expanderClass = `glyphicon glyphicon-chevron-${expander[course.id] ? "up" : "down"}`;
```

that could also just be stuck directly in `className`. hmmm....

static/js/components/CourseList.js
```diff
@@ -29,37 +43,75 @@ class CourseList extends React.Component {
...
+
+        let height = DASHBOARD_COURSE_HEIGHT + courseRuns.length * DASHBOARD_RUN_HEIGHT;
+        return <div
           key={course.id}
-          className={"course-list-body course-list-status-" + course.status}
+          className={"course-list-body course-list-row course-list-status-" + course.status}
```
aliceriot: template string might be preferred here? there's an eslint rule for this (http://eslint.org/docs/rules/prefer-template) I've thought we should maybe turn on.

static/js/components/CourseList.js
```diff
@@ -1,20 +1,34 @@
...
   static propTypes = {
     dashboard: React.PropTypes.object.isRequired,
+    expander: React.PropTypes.object.isRequired,
+    dispatch: React.PropTypes.func.isRequired,
   };
 
+  toggleExpander(courseId) {
+    const { dispatch, expander } = this.props;
+    dispatch(toggleDashboardExpander(courseId, !expander[courseId]));
+  }
```
aliceriot: if you declare `toggleExpander` as an arrow function we won't need to bind it to `this` later:

```js
toggleExpander = courseId => {
  ...
}
```

static/js/components/Dashboard.js
```diff
@@ -1,39 +1,42 @@
...
-      <div className="card-user">
-        <div className="card-image-box">
+      <Grid className="card-user">
+        <Cell col={2} className="card-image-box">
             <UserImage imageUrl={imageUrl}/>
-        </div>
-        <div className="card-name">
+        </Cell>
+        <Cell col={5} className="card-name">
           { profile.preferred_name || SETTINGS.name }
```
aliceriot: do we still want to fall back on `SETTINGS.name`?

aliceriot:
couple comments and some questions, back to you!

noisecapella:
@pdpinch I added a screenshot

static/js/components/CourseList.js
noisecapella: Looks like a good idea, I'll file an issue to deal with this separately since there's a lot of code to change

static/js/components/Dashboard.js
noisecapella: I think this is just for the moment after the page loads but before we get data from the profile API

static/js/components/CourseList.js
aliceriot: sounds good!

static/js/components/Dashboard.js
aliceriot: ah gotcha, probs fine to leave it then.

noisecapella:
Comments are addressed

aliceriot:
:+1:, assuming all tests are passing

Issue #335, opened 2016-05-16
Style profile privacy page like mock-ups 


pdpinch:
/profile/privacy should look like https://invis.io/4E791TKU3#/156281163_Step_Four_-_Profile_Privacy

Notably, by changing to radio buttons, users will be able to see the options more easily.

noisecapella:
We also need to move the email optin text from the work history tab to the privacy tab

amir-qayyum-khan:
Made changes in UI.

<img width="1209" alt="screen shot 2016-05-23 at 9 21 24 pm" src="https://cloud.githubusercontent.com/assets/10431250/15477113/6e5a56ba-212c-11e6-8270-5aac1fcae6ba.png">

PR #347, opened 2016-05-18
Updated privacy tab in profile.

amir-qayyum-khan:
Fixes https://github.com/mitodl/micromasters/issues/335

#### What's this PR do?
- Update UI of privacy tab.

- See profile_edit.js

- Open profile and go to privacy.

screen shot 2016-05-20 at 4 49 18 pm

### DIFF
--- a/static/js/util/ProfileTab.js
+++ b/static/js/util/ProfileTab.js
@@ -7,6 +7,7 @@ import {
   boundSelectField,
   boundMonthYearField,
   boundStateSelectField,
+  boundRadioGroupField,
   saveAndContinue,
 } from './profile_edit';
 import { HIGH_SCHOOL, ASSOCIATE, BACHELORS, MASTERS, DOCTORATE } from '../constants';
@@ -24,6 +25,7 @@ class ProfileTab extends React.Component {
     this.boundStateSelectField = boundStateSelectField.bind(this);
     this.boundDateField = boundDateField.bind(this);
     this.boundMonthYearField = boundMonthYearField.bind(this);
+    this.boundRadioGroupField = boundRadioGroupField.bind(this);
 
     // options we set (for select components)
     let countryOptions = Object.keys(iso3166.data).map(code => ({
@@ -42,9 +44,12 @@ class ProfileTab extends React.Component {
     }));
     this.languageOptions = _.sortBy(languageOptions, 'label');
     this.privacyOptions = [
-      { value: 'public', label: 'Public to the world'},
-      { value: 'public_to_mm', label: 'Public to other micromasters students'},
-      { value: 'private', label: 'Private'}
+      { value: 'public', label: 'Public to the world', helper: `We will publish your Micromasters 
+        profile on our website.` },
+      { value: 'public_to_mm', label: 'Public to other micromasters students', helper: `Your Micromasters profile 
+        will only be viewable by other learners in your program, and by MIT faculity and staff.` },
+      { value: 'private', label: 'Private', helper: `Your Micromasters profile will be viewable only by 
+        MIT faculty and staff.` }
     ];
     this.educationLevelOptions = [
       {value: HIGH_SCHOOL, label: "High school"},
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -6,6 +6,7 @@ import _ from 'lodash';
 import AutoComplete from '../components/AutoComplete';
 import MenuItem from 'material-ui/MenuItem';
 import TextField from 'material-ui/TextField';
+import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 
 import { DATE_FORMAT } from '../constants';
 import {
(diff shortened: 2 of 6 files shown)


amir-qayyum-khan:
@noisecapella please review this

aliceriot:
it looks on the mockup that we want some whitespace around the options:

w

I think if you just add a `<Grid className="profile-tab-grid">` component around the options themselves that should take care of it.

aliceriot:
the button should also be wider, and the spacing between the options, and between the radio button and each option should be smaller.

aliceriot:
the `font-weight` on `.employment-splash` is too much, we should leave it at 400

aliceriot:
the font-weights on the option text labels look a little heavy to me too - I think the grey text (`.radio-label-hint`) should be a couple `px` smaller about `font-weight: 400;`

static/js/util/profile_edit.js
```diff
@@ -19,6 +20,49 @@ import {
...
+  } = this.props;
+
+  let onChange = e => {
+    let clone = _.cloneDeep(profile);
+    _.set(clone, keySet, e.target.value);
+    updateProfile(clone);
+  };
+
+  const selectedValue = _.get(profile, keySet);
+  const radioButtons = _.map(options, (obj) => {
```
aliceriot: can you use `Array.prototype.map` here instead of `_.map`?

something like 

```js
const radioButtons = options.map( obj => {
  ... (function body)
});
```

static/js/util/profile_edit.js
```diff
@@ -19,6 +20,49 @@ import {
...
+      value={obj.value}
+      label={label}/>);
+  });
+   
+  return (<RadioButtonGroup
+    name={label}
+    onChange={onChange}
+    defaultSelected={selectedValue}>
+      {radioButtons}
+    </RadioButtonGroup>
```
aliceriot: when returning JSX I think we should use one of these two styles:

```js
return <div>
  { foo }
</div>;

// or

return (
  <div>
    { foo }
  </div>
);
```

I prefer the latter b/c then the opening and closing tags line up, but it's a style preference which one you want to use.

static/js/util/profile_edit.js
aliceriot: also when declaring an arrow function you don't need `( )` around the function arguments if there's a single argument, so this:

```js
let foo = arg => arg;
```
is equivalent to:

```js
let foo = (foo) => foo;
```
I think it looks better to leave off the parentheses if their not necessary

static/js/util/ProfileTab.js
```diff
@@ -42,9 +44,12 @@ class ProfileTab extends React.Component {
...
     this.privacyOptions = [
-      { value: 'public', label: 'Public to the world'},
-      { value: 'public_to_mm', label: 'Public to other micromasters students'},
-      { value: 'private', label: 'Private'}
+      { value: 'public', label: 'Public to the world', helper: 'We will publish your Micromasters ' +
+        'profile on our website.' },
+      { value: 'public_to_mm', label: 'Public to other micromasters students', helper: 'Your Micromasters profile ' +
+        'will only be viewable by other learners in your program, and by MIT faculity and staff.' },
+      { value: 'private', label: 'Private', helper: 'Your Micromasters profile will be viewable only by ' +
+        'MIT faculty and staff.' }
```
aliceriot: can you use template literals here instead of string concatenation? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

aliceriot:
couple questions and comments, back to you

static/js/util/profile_edit.js
```diff
@@ -6,6 +6,7 @@ import _ from 'lodash';
 import AutoComplete from '../components/AutoComplete';
 import MenuItem from 'material-ui/MenuItem';
 import TextField from 'material-ui/TextField';
+import {RadioButton, RadioButtonGroup} from 'material-ui/RadioButton';
```
aliceriot: style nit:

```js
// this import style
import { foo, bar } from 'foobar';
// is preferable to
import {foo,bar} from 'foobar';
```

static/js/util/profile_edit.js
```diff
@@ -19,6 +20,62 @@ import {
...
+ * @param keySet {String[]} Path to the field
+ * @param label {String} Label for the field
+ * @param options {Object[]} A list of options for the select field
+ * @returns {ReactElement}
+ */
+export function boundRadioGroupField(keySet, label, options) {
+  const {
+    profile,
+    updateProfile
+  } = this.props;
```
aliceriot: can you condense this to one line?

```js
const { profile, updateProfile } = this.props;
```

static/js/util/profile_edit.js
```diff
@@ -19,6 +20,62 @@ import {
...
+      }
+    };
+
+    let onChange = e => {
+      let clone = _.cloneDeep(profile);
+      _.set(clone, keySet, e.target.value);
+      updateProfile(clone);
+    };
+
+    const radioButtons = options.map((obj) => {
```
aliceriot: style nit:

```js
// parenthesis aren't necessary with single-argument arrow functions
// so this:
let bar = (foo) => foo * 2;
// can become
let bar = foo => foo * 2;
```

IMHO it's cleaner to leave them out

aliceriot:
Just two small comments, fix those and it looks good to me! (assuming tests are passing, travisCI is having issues right now :( )

amir-qayyum-khan:
@aliceriot on local machine test are working fine what should i do?

aliceriot:
this is related to a change that was just merged - if you rebase on master things should start passing.

amir-qayyum-khan:
@aliceriot i am going to merge it

aliceriot:
:+1:

noisecapella:
In the future please delete the branch after merging

PR #351, opened 2016-05-18
Used state select field for education tab

### LINKED ISSUE #318: Use state select in education profile view

noisecapella:
Fixes #318


aliceriot:
:+1:, conditional on tests passing

PR #352, opened 2016-05-18
Redirect to profile if the profile is not complete

### LINKED ISSUE #227: require user profile be complete before allowing user to view dashboard

giocalitri:
closes #227 

#### What's this PR do?
Redirects the user to the profile page if the profile is not complete

`ui/decorators.py`

uncheck `Filled out` and `Agreed to terms of service` from the user profile and test that the user is redirected first to the terms of service and then to the profile.

### DIFF
--- a/ui/url_utils.py
+++ b/ui/url_utils.py
@@ -1,0 +1,7 @@
+"""
+Utils for URLs (to avoid circular imports)
+"""
+
+DASHBOARD_URL = '/dashboard/'
+PROFILE_URL = '/profile/'
+TERMS_OF_SERVICE_URL = '/terms_of_service/'
--- a/static/js/constants.js
+++ b/static/js/constants.js
@@ -5,7 +5,7 @@ export const MASTERS = 'm';
 export const DOCTORATE = 'p';
 
 export const USER_PROFILE_RESPONSE = {
-  "filled_out": false,
+  "filled_out": true,
   "agreed_to_terms_of_service": true,
   "account_privacy": "all_users",
   "email_optin": false,
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -41,6 +41,7 @@ class App extends React.Component {
     this.fetchUserProfile(SETTINGS.username);
     this.fetchDashboard();
     this.requireTermsOfService();
+    this.requireProfileFilledOut();
     this.requireCompleteProfile();
   }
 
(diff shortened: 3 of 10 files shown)


noisecapella:
This handles the Django side of the problem but we still need to do this check in the frontend as well. I think we just need to edit `requireCompleteProfile` in `App.js` to check for `filled_out`. I'm not sure how this overlaps with #314, I'd be happy to merge this PR separately and do the frontend work later if you prefer.

noisecapella:
Looks good :+1: 
But we shouldn't close #227 until the frontend piece is fixed as well

ui/url_utils.py
```diff
@@ -0,0 +1,8 @@
+"""
+Utils for URLs (to avoid circular imports)
+"""
+
+DASHBOARD_URL = '/dashboard/'
+PROFILE_URL = '/profile/'
+PROFILE_PERSONAL_URL = '/profile/personal/'
```
noisecapella: Would you mind getting rid of this one? Since our regex doesn't have a terminating character `/profile/` should match just as well

giocalitri:
@noisecapella I do not think this overlaps with the front-end: the front-end should take care of setting the property on the profile and this should take care of not letting the user go anywhere else if such property is not set.

ui/url_utils.py
giocalitri: I will

noisecapella:
But the user can still go to the dashboard in the front end even if the profile is not filled out by clicking the link on the upper right. Shouldn't we handle that case?

giocalitri:
The view for the dashboard is decorated, meaning from wherever you try to access is, it will still redirect to the profile.
Maybe I do not understand what you mean...

noisecapella:
Since the dashboard page is a single page application handling the dashboard, profile, terms of service, etc, the user can navigate between those pages without touching the server

giocalitri:
I understand, but I am honestly incapable to find the case you are describing in the current UI (I just checked).
I agree with you and we can open an issue for the front-end part.

giocalitri:
I also I have 2 comments about this:
1- if this story was about the backend and front-end, why is it only 2 points?
2- I am starting to think that we should not put all the front end in a single page app

noisecapella:
It should have been pointed higher probably, but the work is not very complicated and already done for terms of service on the front end. What are your thoughts regarding (2)?

giocalitri:
done with comments and added front-end part

noisecapella:
Looks good :+1:

PR #354, opened 2016-05-18
Added prefer-template eslint rule

### LINKED ISSUE #350: Use prefer-template eslint rule

aliceriot:

closes #350 

#### What's this PR do?

Adds the `prefer-template` rule to eslint and fixes all lint errors.

Review all the changes, and make sure the newer template literal will produce the same string that the concatenation would have.

Click around and make sure there are no regressions (clobbered spaces, etc).

### DIFF
--- a/static/js/components/Dashboard.js
+++ b/static/js/components/Dashboard.js
@@ -14,7 +14,7 @@ class Dashboard extends React.Component {
 
   render() {
     const { profile, dashboard, expander, dispatch } = this.props;
-    let imageUrl = (SETTINGS.edx_base_url + '/static/images/profiles/default_120.png').
+    let imageUrl = `${SETTINGS.edx_base_url}/static/images/profiles/default_120.png`.
     //replacing multiple "/" with a single forward slash, excluding the ones following the colon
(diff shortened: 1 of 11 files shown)


static/js/components/Dashboard.js
```diff
@@ -14,7 +14,7 @@ class Dashboard extends React.Component {
 
   render() {
     const { profile, dashboard, expander, dispatch } = this.props;
-    let imageUrl = (SETTINGS.edx_base_url + '/static/images/profiles/default_120.png').
+    let imageUrl = `${SETTINGS.edx_base_url}/static/images/profiles/default_120.png`.
```
noisecapella: It would be nice to use a path join library similar to `os.path.join` in Python, but out of scope for this issue

aliceriot: does `path` do anything for us? It's part of node, so it wouldn't run in the browser, but if there were some way to signify that an expression should be evaluated and then substituted when compiling / building the JS we could maybe use that? idk that's far-fetched haha

noisecapella:
:+1:

PR #355, opened 2016-05-18
Fixed AutoComplete quirks

### LINKED ISSUE #327: Fix behavior where AutoComplete leaves text in field which doesn't correspond with selected item
There are two ideas:
 - we store the text being edited as well as the selected text. We clear the text being edited when a new item is selected, so if we check that there is no text being edited when we validate, we should be able to force the user to select this
 - copy AutoComplete into our repository and modify it to our needs. This will give us better control over what callbacks we have access to

noisecapella:
Fixes #327 

#### What's this PR do?
Forks the AutoComplete component from material-ui and makes these changes:
 - When the user tabs onto the field with the keyboard, the menu does not open anymore. Instead the user has to type text first.
 - When the user clicks the field the entire set of choices will be shown rather than the choices which match the text in the textbox. After the user edits text in the textbox, the choices will be filtered to match that text.
 - The onBlur handler is now not invoked when the user presses enter, down, or escape. This means the user can navigate through the menu with their arrow keys to select an option.

Other changes to our code:
 - Added back the onBlur handler to show the selected value when the user removes focus from the textbox, but not when the user presses down, enter, or escape.
 - When text is edited, any other value will be deselected. This will prevent people from leaving half completed text in the autocomplete and having it pass validation

Use the select field in the conditions described above. The main change of behavior from what is on master is, when the user navigates away from the select field without selecting anything first, the field is cleared. This will hopefully prevent confusion on what is being submitted in the profile.

https://twitter.com/mononcqc/status/723876305951924226/photo/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "history": "2.1.1",
     "iso-3166-2": "0.4.0",
     "isomorphic-fetch": "2.2.1",
+    "keycode": "2.1.1",
     "lodash": "4.11.2",
     "material-ui": "0.15.0",
     "moment": "2.13.0",
@@ -57,6 +58,7 @@
     "striptags": "2.1.1",
     "style-loader": "0.13.1",
     "url-loader": "0.5.7",
+    "warning": "2.1.0",
     "webpack": "1.13.0"
   },
   "scripts": {
--- a/static/js/components/AutoComplete.js
+++ b/static/js/components/AutoComplete.js
@@ -1,0 +1,568 @@
+// copied from https://raw.githubusercontent.com/callemall/material-ui/master/src/AutoComplete/AutoComplete.js
+/* eslint-disable indent */
+import React, {Component, PropTypes} from 'react';
+import ReactDOM from 'react-dom';
+import keycode from 'keycode';
+import TextField from 'material-ui/TextField';
+import Menu from 'material-ui/Menu';
+import MenuItem from 'material-ui/MenuItem';
+import Divider from 'material-ui/Divider';
+import Popover from 'material-ui/Popover/Popover';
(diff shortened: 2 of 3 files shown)


noisecapella:
This still has a quirk where the user presses the down key, clicks elsewhere to lose focus, and the text field is not cleared. I'm also not sure whether limiting the select to 5 items is a weird experience for the user, though it certainly improves responsiveness

giocalitri:
I think I understood the change here, even if I suppose most of the library has been copied from the original one.
I think this is fine as it is, but you might want to open a PR to the original library and propose the additional feature.
👍 from me in the meantime

giocalitri:
just realized there was no proper label: is the work not finished?

noisecapella:
It's still a WIP, I'd like to get the other quirks ironed out first.

I'm not sure about making a PR for this because this work is kind of a hack, and they say it will be refactored soon: https://github.com/callemall/material-ui/issues/2735

Ferdi:
Is all this work worth it ? I was the one to suggest this feature, but it's getting more complicated than I thought.  It looks like we are forking the whole library and maintain ourselves ?
Couldn't we use one of many tested (in terms of code and usability) libraries out there and style it a bit to look like material design.  For example : https://twitter.github.io/typeahead.js/examples/  ?
Just a thought .

noisecapella:
@Ferdi We are only forking the AutoComplete component, not the entirety of material-ui. Does that address your concerns? I agree that it adds a maintenance burden but I think it's acceptable in this case since the work to style other selectboxes to look exactly like the other material-ui components is not trivial.

@giocalitri I fixed the quirk (mostly) and put the PR up for review. Any other thoughts?

giocalitri:
I think this is fine 👍

PR #356, opened 2016-05-18
Changed OSX Docker workflow to expect the webpack server to be run on the host machine

gsidebo:

#307 

#### What's this PR do?

Changes our workflow so OSX users will run the webpack dev server on their host machines rather than inside the Docker container. This eliminates the need for docker-osx-dev.

Start at `docker-compose.yml` for the config changes, `micromasters/utils.py` for the code changes

OSX users should run `./webpack_dev_server.sh --install` in one terminal tab, and `docker-compose up`. Linux users can just run the latter command. We'll definitely want a Linux user to test this since I hacked on this in OSX. 

**Major implications:**

1. (as mentioned) OSX users will need to run the webpack server locally in a separate terminal tab, and can do so by running the helper script `./webpack_dev_server.sh`
1. (as mentioned) docker-osx-dev is no longer needed
1. OSX users will now need to add the following to the `.env` file: `WEBPACK_DEV_SERVER_HOST='localhost'`. Before now, webpack was assumed to be running in the container with the web server.
1. The 'watch' service in `docker-compose.yml` now invokes the helper script to start the server. In the case of OSX users, 'watch' will do nothing. I couldn't figure out a good way to entirely exclude the 'watch' service for OSX users that wouldn't require some annoying flags to be specified every single time the `docker-compose` command was used. Any feedback on that is quite welcome.


gsidebo:
forgot README updates. working on that now...

gsidebo:
some things i'm not sure about:

- the new file `micromasters/utils.py` containing the functions that provide the webpack host/url. couldn't think of a better place for them
- those functions requiring `request` to be passed in. kind of a code smell. is there a cleaner way to determine the machine's IP in django than grabbing it from the request?
- like i said in the first comment, the 'watch' service is still available when running from OSX and it just does nothing
- the webpack_dev_server script is a bit awkward in general

giocalitri:
There are some linting errors.
Please fix

micromasters/settings.py
```diff
@@ -257,7 +257,8 @@ def get_var(name, default):
 
 # Request files from the webpack dev server
 USE_WEBPACK_DEV_SERVER = get_var('MICROMASTERS_USE_WEBPACK_DEV_SERVER', False)
-WEBPACK_SERVER_URL = get_var('MICROMASTERS_WEBPACK_SERVER_URL', 'http://{host}:8078')
+WEBPACK_DEV_SERVER_HOST = get_var('WEBPACK_DEV_SERVER_HOST', '')
```
giocalitri: why an empty default for this?

webpack_dev_server.sh
```diff
@@ -0,0 +1,37 @@
...
+#       the webpack server on the host machine rather than the container.
+
+if [[ -z "$WEBPACK_DEV_SERVER_HOST" || "$IS_OSX_HOST_MACHINE" == 'true' ]] ; then
+    if [[ "$1" == "--install" ]] ; then
+      npm install --no-bin-links && echo "Finished npm install"
+    fi
+    # Start the webpack dev server on the appropriate host and port
+    node ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.config.dev.js -d \
+        --content-base ./static --host "$WEBPACK_HOST" --port "$WEBPACK_PORT" --progress --inline --hot
+fi
```
giocalitri: empty line at the end of the file

giocalitri:
I think I need to talk to you in person tomorrow morning to fully understand this PR

giocalitri:
In the meantime, please rebase and fix conflicts

micromasters/settings.py
gsidebo: when this isn't specified, we want the webpack server host to be the same as the django host. before this PR, we were pulling the django host value out of the request in our views (`request.get_host()`). i kept that approach. when `WEBPACK_DEV_SERVER_HOST` is blank, we get the host via `request.get_host()`. i mentioned above that if there's an easy way to get the django host that doesn't involve pulling it out of the request, we should go with that

ui/views_test.py
```diff
@@ -91,13 +91,14 @@ def test_authenticated_user_doesnt_redirect(self):
     def test_webpack_url(self):
```
giocalitri: I know you did not write this test, but (as note for the future), it is better to use `ddt` or `@pytest.mark.parametrize` for multiple inputs for a test.

webpack_dev_server.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -e
+
+if [[ $(uname -s) == "Darwin" ]] ; then
+  IS_OSX_HOST_MACHINE="true"
+else
+  IS_OSX_HOST_MACHINE="false"
```
giocalitri: does the double quote vs single quote make a difference in bash?

giocalitri:
I have an additional question: on Mac, where the npm packages are installed? in the current directory or in a global/system one?

giocalitri:
First pass done: overall it looks good: I have only a couple of questions and few minor comments

webpack_dev_server.sh
noisecapella: Only if there are variables inside the string

webpack_dev_server.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -e
```
noisecapella: Can you use `set -euf -o pipefail` here? See https://sipb.mit.edu/doc/safe-shell/

gsidebo:
@giocalitri 

> I have an additional question: on Mac, where the npm packages are installed? in the current directory or in a global/system one?

as @aliceriot kindly reminded me, npm packages are installed locally (ie: in the same directory as the package.json) by default. you need to set the `-g` flag explicitly to install them on the entire system

> In the meantime, please rebase ...

Just want to make sure: when you say 'rebase', you mean that I should squash all the branch commits to one commit, right?

giocalitri:
No, when I say `rebase`, there are conflicts between this branch and master, so you should rebase to fix them.

gsidebo:
@giocalitri just rebased onto master

giocalitri:
I get this error
```
gioMacMit:micromasters gdimilia$ ./webpack_dev_server.sh
./webpack_dev_server.sh: line 18: WEBPACK_DEV_SERVER_PORT: unbound variable
```

gsidebo:
@giocalitri fixed

aliceriot:
not sure if it's ready for linux testing yet, but I just checked out this branch and did `docker-compose up` with no problems.

gsidebo:
@aliceriot it's ready, and thank you! you're way ahead of me. i'll merge this when i get the +1

giocalitri:
from the OS X point of view this is good to go. 👍 
we might need a second opinion from a linux user ( @aliceriot ?)

giocalitri:
please squash the commits

gsidebo:
yeah @aliceriot some very minor changes were made since you tested. should be good to go, but it certainly wouldn't hurt if you have it one more try.

aliceriot:
looks good to me! everything's working just fine.

Issue #216, opened 2016-04-26
Allow user to view & edit profile


pdpinch:
As a micromasters student, I'd like to be able to view my profile in one page, with the ability to edit it. 

The user can view at /users/{username}, and editing will return to the profile form at /profile/{username} with the proper tab open. 

Modeled after aboutme.google.com. 

Current draft of the design at https://invis.io/4E791TKU3#/157446163_Public_Profile

pdpinch:
double check the url structure

noisecapella:
Is this necessary for the MVP?

pdpinch:
@noisecapella yes. 

The expectation is that this is the same API as the first profile. And the hope is that we can reuse the react components for education and employment.

noisecapella:
I'm going to submit a PR to put this in place (currently non-functional), then we can figure out what we want in more detail

![screenshot from 2016-05-19 14-31-39](https://cloud.githubusercontent.com/assets/863262/15405110/90713a36-1dce-11e6-9d10-44b2a52db9c9.png)

Ferdi:
@noisecapella that looks pretty good !

pdpinch:
I'm changing this to its own epic.

PR #357, opened 2016-05-19
Refactored profiles reducer to handle multiple profiles

noisecapella:
Required for #216 

#### What's this PR do?
Renames the `userProfile` reducer to `profiles` and changes the structure and related actions to use the profile username so we can store multiple profiles.

Everything should work as before


annagav:
Looks good 👍

PR #358, opened 2016-05-19
Added User page

noisecapella:
Part of #216 

#### What's this PR do?
Adds a page at /users/:username. Currently just a template

See screenshot in comment on #216


giocalitri:
needs to be rebased

noisecapella:
Rebased

static/js/components/UserImage.js
```diff
@@ -1,18 +0,0 @@
...
-  render() {
-    const { imageUrl } = this.props;
-
-    return <img
-      className="card-image" src={ imageUrl }
-      alt={`Profile image for ${SETTINGS.name}`}/>;
-  }
-}
-
-export default UserImage;
```
aliceriot: code deletion :100:

static/js/containers/UserPage.js
```diff
@@ -0,0 +1,64 @@
...
+      <User profile={profile} />
+    </Loader>;
+  }
+}
+
+const mapStateToProps = state => ({
+  profiles: state.profiles
+});
+
+export default connect(mapStateToProps)(UserPage);
```
aliceriot: is there a reason we can't just pass the `profile` and other stuff down as props from `App`? I guess we have `ProfilePage` connected to the store as well, so maybe this is ok, just wondering.

aliceriot:
functionality looks good, going to read the code more closely now.

static/js/util/util.js
```diff
@@ -359,3 +359,15 @@ export function validateYear(string) {
...
+export function makeProfileImageUrl(profile) {
+  let imageUrl = `${SETTINGS.edx_base_url}/static/images/profiles/default_120.png`.
+  //replacing multiple "/" with a single forward slash, excluding the ones following the colon
+    replace(/([^:]\/)\/+/g, "$1");
+  if (profile.profile_url_large) {
+    imageUrl = profile.profile_url_large;
+  }
+
+  return imageUrl;
+}
```
aliceriot: oh so the `[^:]` means that we don't clobber the double slash after `https://`.

I'm still a regex noob, is there a reason we can just do:

```js
imageUrl.replace(/[^:]\/\//, "/");
```

aliceriot:
this generally looks great, just a couple of questions.

static/js/util/util.js
noisecapella: I think that will remove the character before the double slash as well. I'm not a regex expert either, I just moved the above code from the existing Dashboard component

aliceriot: ah yes indeed it does! what does the `/+/g` do? If I do this:

```js
var url = "https://foo.bar//asdf";
url.replace(/([^:]\/)\//, "$1");
```

it seems to have the same result ("https://foo.bar/asdf") as the pattern above. ahhh regex, so inscrutable sometimes...

static/js/containers/UserPage.js
noisecapella: The profile is loaded in the lifecycle of this component which requires dispatching actions. We could pass down `dispatch` and `profiles` from `App` but this seems to fit better with convention

static/js/util/util.js
aliceriot: ah figured it out, /g does the whole string

static/js/containers/UserPage.js
aliceriot: sounds good to me, I was just wondering about when to pass down props w/ helper methods vs when to use `connect` - in the education reducer refactor I disconnected that component (which made sense to me) and was just wondering what our general protocol for this was.

static/js/util/util.js
noisecapella: I think it grabs multiple slashes in a row, if you have the pathological path which has three or more slashes in a row for some reason

aliceriot: haha, yeah I think `/c+/` matches one or more occurrences of `c`

aliceriot: anyway this is fine as is, haha

var regex = /what/;

static/js/containers/UserPage.js
noisecapella: Yeah, I'm not sure of a general rule of thumb on that either. In teachersportal we had a problem where we were passing a lot of props and functions down from the top of the component tree since we didn't use `connect` on the lower level components. But on the other hand passing props and functions makes it a lot easier to test. ¯\_(ツ)_/¯

aliceriot: yeah - I think as long we're holding to the pattern of trying to make our presentational components as functional as possible we're doing well. like we definitely don't want to have components which are way down the component tree and just render some small thing to be connected to the store, they should just received props and spit them out.

aliceriot:
this looks good to me (assuming tests pass)

:+1:

PR #359, opened 2016-05-19
Added favicon

### LINKED ISSUE #334: Add favicon
Unless we plan to customize, we can re-use the MIT favicon in use on the micromasters marketing site, https://github.com/mitodl/mga/blob/gh-pages/images/favicon.ico

noisecapella:
Fixes #334 

#### What's this PR do?
Adds a favicon

Make sure that the favicon shows up when you bookmark the micromasters PR deploy


ui/templates/base.html
```diff
@@ -4,7 +4,9 @@
     <meta charset="UTF-8">
     <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700" />
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7" crossorigin="anonymous">
-    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
+    {% load static %}
+    <link rel="shortcut icon" href="{% static 'images/favicon.ico' %}" />
```
gsidebo: what does `{% load static %}` do exactly? this feels a little clunky to me. could you create a new template tag (et al) that could just load the correct href value in one line?

noisecapella: `{% load static %}` imports the static template tag which lets us reference static assets. Would creating a new template tag that be simpler than just using the built in `static` tag?

gsidebo: yeah a new template tag might not be the answer here, but loading the `static` template tag within the actual template strikes me as anti-pattern. i feel like it would be cleaner to auto-load the `static` template tag for all templates (or at least this one). i say this while acknowledging the following:

1. other templating libraries (jsp?) expect certain functionality to be loaded in the actual template file
1. after poking around a little, it seems that django docs think this approach is OK
1. this brings up an explicit vs. implicit argument

i just don't love the idea of template code that doesn't directly influence the markup that gets generated. if you agree with the sentiment, feel free to autoload the static tag, but this is ship-able as-is

👍

PR #361, opened 2016-05-19
Updated style of education tab

### LINKED ISSUE #331: Style education form to match mock-ups
See https://invis.io/4E791TKU3#/153562861_Step_Two_-_Educational_Info and https://invis.io/4E791TKU3#/155766490_Step_Two_-_Educational_Info_-_Add_Degree

- [ ] increase margins
- [ ] match colors
- [ ] make add button (`+`) green and move down to [similar location to employment](https://github.com/mitodl/micromasters/issues/289#issuecomment-219046258)
- [ ] border around each component
- [ ] make active component white / shade other components

step two - educational info
step two - educational info - add degree

noisecapella:
Fixes #331

#### What's this PR do?
Refactors the education tab to update its look and feel to match the mockups

EducationTab.js

All existing functionality should work as before. The look and feel should match the comments in the issue #331

screenshot from 2016-05-19 17-26-52

### DIFF
--- a/static/js/components/EducationTab.js
+++ b/static/js/components/EducationTab.js
@@ -1,15 +1,17 @@
 import React from 'react';
 import Button from 'react-mdl/lib/Button';
 import IconButton from 'react-mdl/lib/IconButton';
-import _ from 'lodash';
-
-import ProfileTab from "../util/ProfileTab";
-import { generateNewEducation } from "../util/util";
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 import Switch from 'react-mdl/lib/Switch';
 import FABButton from 'react-mdl/lib/FABButton';
 import Icon from 'react-mdl/lib/Icon';
 import Dialog from 'material-ui/Dialog';
+import { Card } from 'react-mdl/lib/Card';
+import _ from 'lodash';
+import moment from 'moment';
+
+import ProfileTab from "../util/ProfileTab";
+import { generateNewEducation } from "../util/util";
 import { saveAndContinue } from '../util/profile_edit';
 
 class EducationTab extends ProfileTab {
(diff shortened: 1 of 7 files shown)


aliceriot:
@noisecapella do we want the 'plus' buttons to sit on the bottom of the degree cards (as the one on the work history tab does)?

right now they're only aligned when you have two or more entries in the degree category:

button_lineup

haven't dug into the code to see how you're positioning them yet, but on the employment tab I had a little helper function that changed the `className` prop on the `FABButton` in response to how many entries there were, we might want something similar here.

aliceriot:
also I think it's worth mentioning that in google's material design guidelines they recommend only using one FAB per screen, although I can't really think of a nice way around that here.

noisecapella:
I'm using the same code as in the employment tab because the linked issue recommended positioning them the same way, though we may want to rethink that slightly

aliceriot:
one other appearance-level thing, it looks on the mockup like the switch elements should be greyed-out when they're turned off.

static/scss/dashboard.scss
```diff
@@ -188,16 +188,18 @@ c.validation-wrapper {
...
+    .profile-add-button.no-entries {
       top: 68px;
     }
-    .add-employment-button.one-entry {
+    .profile-add-button.one-entry {
       top: 38px;
     }
-    .add-employment-button.more-entries {
+    .profile-add-button.more-entries {
       top: 28px;
```
aliceriot: re: the comment above about spacing, I think we may need to either have different spacing for the education and professional tabs, or do some tweaking to what the education tab currently looks like to ensure that the same `top: XXpx;` style will put the button in the right place (currently it looks like the fonts are a little bigger on the education tab, and that might be causing the issue).

static/js/components/EducationTab.js
```diff
@@ -97,48 +99,70 @@ class EducationTab extends ProfileTab {
...
+        <FABButton
+          colored
+          onClick={() => this.openNewEducationForm(level.value, null)}
+          key="I'm unique!"
+          className={`profile-add-button education-add-button ${fabSpacing()}`}>
+          <Icon name="add" />
+        </FABButton>
+      );
+      return rows;
     }
```
aliceriot: this is so similar to the function on `EmploymentTab`...probably not worth abstracting it right now but maybe something to do later.

aliceriot:
Code generally looks great! I think the way we draw the ui for the education tab is a lot more readable and sensible with these changes.

Couple questions about small styling issues, back to you!

noisecapella:
@pdpinch do you want to have a quick meeting on the style issues @aliceriot brought up?

pdpinch:
I think @Ferdi okayed this on Friday, but I can meet to confirm.

noisecapella:
Just pushed some style updates

static/js/components/EducationTab.js
```diff
@@ -104,14 +104,16 @@ class EducationTab extends ProfileTab {
...
+        rows = Object.entries(education).
+          filter(([index, education]) => education.degree_name === level.value).
+          map(([index, education]) => this.educationRow(education, index));
       }
       rows.push(
         <FABButton
           colored
           onClick={() => this.openNewEducationForm(level.value, null)}
           className="profile-add-button education-add-button"
+          key="I'm unique!"
```
aliceriot: :D

aliceriot:
I have basically one question at this point, about the look and feel on this tab and the employment tab.

first, employment:

employment_ffff

education:

education_asdf

I think we should standardize a few things here, basically:

- the minimum amount of whitespace between the last row (education or workhistory entry) and the bottom of the card
- colors (for the switch and the 'plus' button)
- delete icon

right now I think it's a little jarring to go back and forth between the two tabs, since they're just a little bit different.

static/js/components/EmploymentTab.js
```diff
@@ -168,7 +168,7 @@ class EmploymentTab extends ProfileTab {
           colored
           onClick={this.addWorkHistoryEntry}
           key="I'm unique!"
-          className={`add-employment-button ${fabSpacing()}`}>
+          className={`profile-add-button ${fabSpacing()}`}>
```
aliceriot: do we still need to use `fabSpacing()` here? It looks like you're not using something similar above for the education `<FABButton>`

noisecapella: I'm absolute positioning the FABButton there, I'll update this for the employment tab too

pdpinch:
Thanks @aliceriot for putting a finer point on what needs to be addressed. I totally agree they should be the same. Some answers, after consulting with Ferdi:
    
- the whitespace in the employment example is preferred. 
- colors should be like the mockups (green)
- we like the trash can delete icon 

right now I think it's a little jarring to go back and forth between the two tabs, since they're just a little bit different.

static/js/components/EmploymentTab.js
aliceriot: ahhh nice, that's a much cleaner approach than the hack I wrote >_<

noisecapella:
The two tabs should be consistent now. I also removed whatever education or employment specific CSS I could find

aliceriot:
the UX is great! feels really nice and unified.

there is one small issue, which I'm not sure is strictly related to this PR, but I noticed it just now.

everything except the navbar is inside of `<div class='app-media layout-boxed'>`. For some reason different `right` and `left` margins are being computed for this class on the different tabs. Specifically, education and personal have one value and work history and privacy another. This is despite the fact that the card maintains the same width across all of these, and that `margin-left` and `margin-right` are set to `auto` on `.layout-boxed`. I'm kind of puzzled about where the difference is coming from...

maybe better file another issue for that.

static/js/components/EducationTab.js
```diff
@@ -88,79 +90,88 @@ class EducationTab extends ProfileTab {
...
-          </FABButton>
-        </Cell>
-      </Grid>;
+  renderEducationLevel(level){
+    const { ui: { educationDegreeInclusions }, profile: { education } } = this.props;
+    if (educationDegreeInclusions[level.value]) {
+      let rows = [];
+      if (education !== undefined) {
+        rows = Object.entries(education).
+          filter(([index, education]) => education.degree_name === level.value).  // eslint-disable-line no-unused-vars
```
aliceriot: perhaps I'm misunderstanding something, but why is `Object.entries` necessary here? Isn't `profile.education` an array? Why could we do something like:

```js
rows = education.filter( education => education.degree_name === level.value ).
map( ... ) 
```

aliceriot: ah I see, we're using the index below in `this.educationRow`, but we want the index in the original array.

otherwise we could do

```js
education.filter( ...).map( ( education, index ) => this.educationRow(education, index) )
```

but the index we'd be passing in would be the index of the elements in the return value of `education.filter( ... )`, and not the index in `profile.education`.

is that correct?

aliceriot: In any case, we can get rid of the eslint rule disable by doing:

```js
filter(([ , education]) => education.degree_name === level.value).
```

the comma lets us ignore the first entry in the array we're destructuring

static/js/containers/ProfilePage_test.js
```diff
@@ -66,10 +66,7 @@ describe("ProfilePage", function() {
       );
 
       renderComponent(`/profile/${page}`).then(([component, div]) => {  // eslint-disable-line no-unused-vars
```
aliceriot: same comment here as above, mind ignoring `component` so we can remove the rule disable?

static/js/components/EducationTab.js
noisecapella: Yes, getting the index after the filter caused a subtle bug. I'll remove the eslint rule

aliceriot:
:+1: :shipit: :100: :fireworks: :confetti_ball: :cake:

PR #371, opened 2016-05-20
Added middleware to redirect canceled authorizations

### LINKED ISSUE #169: If user clicks cancel when authorizing edX an AuthCanceled exception is thrown
I think we just need to capture this exception and redirect them to the index

noisecapella:
Fixes #169 

#### What's this PR do?
When the user logs into a server and clicks 'Cancel', they currently get a 500 error on micromasters. This will cause it to redirect to the home page instead.

Run your local instance with `DEBUG = False`. Log into the sandbox or your devstack and cancel the authorization when it asks for read and write scope. You should be redirected to `/` instead of getting a 500 error

This will silence oauth authorization exceptions in general


giocalitri:
code is fine. testing functionality

giocalitri:
👍

PR #372, opened 2016-05-20
Removed node-neat

noisecapella:
#### What's this PR do?
Removes node-neat and related imports. `bourbon-neat` was removed in #345 but this was overlooked


aliceriot:
this might be something unrelated, but the homepage looks like this on my machine:

whats_up

something's not quite right.

noisecapella:
Nope, I missed some references. I force pushed to update

aliceriot:
looks good now! :+1:

PR #395, opened 2016-05-23
Cleared state field when country field changes

### LINKED ISSUE #383: Bug: selecting a new country doesn't clear the state/territory state
To reproduce: start with the state and country fields selected. Change the country field and you see the state field clear in the UI. However the redux state for the state field doesn't change until you interact with the state field, so it will pass validation despite being a state for a different country.

I think the solution is to create a `boundCountrySelectField` and specify the state field that's related to it. When a new country is selected, wipe the state for the state/territory field

noisecapella:
Fixes #383 

#### What's this PR do?
Clear state field state when the country changes. This prevents a subtle bug where the state field appears to clear but its state keeps the previous state/territory

Change the country in the employment and education and the birth place and current place of the personal tab. In each case don't touch the state field, then try to save it. You should see a validation error for the state indicating that it wasn't selected. (We require the state field for each of these cases, right?)

### DIFF
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -120,9 +120,11 @@ export function boundTextField(keySet, label) {
  * @param keySet {String[]} Path to the field
  * @param label {String} Label for the field
  * @param options {Object[]} A list of options for the select field
+ * @param onChange {func} Handler which is called when the state changes or is cleared.
+ * Takes the updated profile as argument
  * @returns {ReactElement}
  */
-export function boundSelectField(keySet, label, options) {
+export function boundSelectField(keySet, label, options, onChange) {
   const {
     profile,
     errors,
@@ -162,6 +164,9 @@ export function boundSelectField(keySet, label, options) {
     _.set(clone, editKeySet, undefined);
 
     updateProfile(clone);
+    if (_.isFunction(onChange)) {
+      onChange(clone);
+    }
   };
 
   let selectedValue = _.get(profile, keySet);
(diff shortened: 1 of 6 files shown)


aliceriot:
We do require the state fields, per #288

aliceriot:
Something isn't quite right when editing an existing entry in work history.

first, selecting a new country doesn't cause the state field to clear:

country

here I've changed the country from Andora to Afghanistan, but the state or territory list still reflects the Andora selection. I can press save, and the dialog closes with no problems, but when I open it again the country and state selections are still Andora / Ordino.

then if I change the state field and try to save I get this:

wh_edit

noisecapella:
@aliceriot It looks like the AutoComplete blur fix wasn't included in this PR. I've just rebased, can you try it again?

aliceriot:
will do

aliceriot:
yep, that's fixed!

static/js/util/profile_edit.js
```diff
@@ -162,6 +164,9 @@ export function boundSelectField(keySet, label, options) {
     _.set(clone, editKeySet, undefined);
 
     updateProfile(clone);
+    if (onChange !== undefined) {
```
aliceriot: maybe `_.isFunction` instead? if we get some unexpected value passed in we could get a runtime error (`Uncaught TypeError: foobar is not a function`).

noisecapella: I prefer runtime errors for those cases since there's a clear logic error on our part if we're passing a non-function as the onChange argument and a runtime error might make that easier to diagnose. But in this case it probably doesn't matter, an error on our part would mean the country field would not be synced up with the state field anymore, which would be preferable to a runtime error halting the render method

aliceriot:
cool, LGTM :+1: :100: :fireworks: :confetti_ball:

PR #397, opened 2016-05-23
Used later version of React to fix test failures

noisecapella:
#### What's this PR do?
Updates the minor version of React so that packages we depend on are all consistent with each other. This should fix the test failures


aliceriot:
:+1:

```js
npm = Garbage.fire();
```

PR #400, opened 2016-05-23
Fixed webpack_if_prod script to make it OSX-friendly

gsidebo:
Noticed this bug when running `npm install` locally. The -u flag requires env variables to be set. There's no reason for us to require that (at least not that I know of). Now that this script is going to be run in OSX host machines as well as Docker containers, this change was necessary

### DIFF
--- a/webpack_if_prod.sh
+++ b/webpack_if_prod.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
-set -euf -o pipefail
-if [[ "$NODE_ENV" != "development" ]]
+set -ef -o pipefail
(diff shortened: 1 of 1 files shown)


noisecapella:
The redux logger uses the `NODE_ENV` variable to say whether or not to output logging statements to the console. Perhaps OS X users could define `NODE_ENV=development` in their `.bash_profile` instead? (I forget exactly which file this is)

noisecapella:
Actually it looks like I misunderstood which script you were altering. I think this is a reasonable change :+1:

PR #402, opened 2016-05-24
Changed JS setup to use npm-shrinkwrap

### LINKED ISSUE #399: proposal: use npm-shrinkwrap instead of pinning exact versions
TL;DR: We've seen issues even with pinning exact versions of npm packages. I think using [npm-shrinkwrap](https://docs.npmjs.com/cli/shrinkwrap) could help us avoid this in the future.

## Details

An `npm-shrinkwrap.json` file works much like a `Gemfile.lock` file does in Ruby/bundler. Every time a dependency is added `npm` will automatically update the shrinkwrap file to reflect the addition of the dependency, and when you do `npm install` it will decide what to install (and at which versions) based on the shrinkwrap file, instead of looking at the allowed versions in the `package.json` and deciding what to do from there (something `npm` is notoriously, shall we say, creative with).

I've used it on some side projects - here's an example: https://github.com/aliceriot/literate-crypto/blob/master/npm-shrinkwrap.json - and found it really helpful.

The shrinkwrap file pins exact versions for our explicit dependencies (which we currently pin in `package.json`) and also for any dependencies that those dependencies pull in. The problem we saw today (5/23/16) was due to a new version of React being pushed, which our dependencies allowed (because they specify a range). Using npm shrinkwrap would circumvent this problem, because we pin exact dependency-of-dependency versions, so newer version of React (which meet the dependency version requirements of our dependencies) won't cause the versions of packages being pulled in to change.

This should give us more confidence in `npm install` doing the same (or at least, a really similar) thing every time.

## Implementation of proposal

If we decide to do this I think we'll need to bump our node / npm version, shrinkwrap support didn't get that great until npm 3.

Then there's only one other catch: `npm shrinkwrap` doesn't include `devDependencies` by default. I'd suggest two options here:

1. we can pass the `--dev` flag to `npm shrinkwrap` to include dev dependencies in the shrinkwrap file
2. we can discard `devDependencies` altogether and put all dependencies in `dependencies`.

Personally I favor 2, it's more simple overall, and (as far as I can tell) we aren't installing that much extra stuff in `devDependencies` anyhow.

Then switching to `npm-shrinkwrap` would look basically like:

1. `rm -rf node_modules`
2. reorganize `devDependencies`
3. bump node version so we can use npm 3
4. `npm install`
5. `npm shrinkwrap`
6. commit `npm-shrinkwrap.json`

Anyway, feedback?

aliceriot:

closes #399 

#### What's this PR do?

This does 4 things:

- bump our nodejs and npm version to 6.2
- use a nodejs image for the watch container (in docker-compose)
- move all `devDependencies` into `dependencies`
- create an `npm-shrinkwrap.json` file (after doing a clean install with the new `package.json`)

Look over the changes!

Make sure nothing JS wise is broken. 

There's discussion about this on #399.


aliceriot:
anecdotally, it looks like moving the JS stuff to it's own container shaves around two minutes
off of the CI test suite time!

ci

noisecapella:
I'm getting this when I run `docker-compose up` locally:

> watch_1 | ERROR in Missing binding /src/node_modules/node-sass/vendor/linux-x64-48/binding.node
watch_1 | Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 6.x
watch_1 | 
watch_1 | Found bindings for the following environments:
watch_1 |   - Linux 64-bit with Node.js 4.x
watch_1 | 
watch_1 | This usually happens because your environment has changed since running `npm install`.
watch_1 | Run `npm rebuild node-sass` to build the binding for your current environment.

I'll investigate more tomorrow

aliceriot:
hmm yeah I'm not sure what's up with that error, I've had that come up several times on the current setup as well.

aliceriot:
did you `rm -rf node_modules` and do `docker-compose build --no-cache` before doing `docker-compose up`?

aliceriot:
OK so I think I've found the problem. If I do the following:

```
rm -rf node_modules
docker rm micromasters_watch_1
docker-compose build
docker-compose up
```

I get the following error:

sass_why

if I remove `--no-bin-links` from line 35 in `webpack_dev_server.sh` and redo the above steps I get a working install with no problems.

is this behaving the same way on your machine @noisecapella ?

noisecapella:
We shouldn't need `--no-bin-links` anymore since that was a workaround for OS X shared folder behavior. I'll try it out without it

test_suite.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
```
noisecapella: This won't exit with a non-zero code on error. Can you add

    set -euf -o pipefail

From https://sipb.mit.edu/doc/safe-shell/

.travis.yml
```diff
@@ -3,5 +3,8 @@ language: ruby
 cache: bundler
 script:
   - docker-compose -f travis-docker-compose.yml run web tox
+  - docker-compose -f travis-docker-compose.yml run watch npm install
```
noisecapella: FYI We aren't setting `NODE_ENV=development` anymore here or below. I suppose it doesn't matter for tests though, it may even be preferable not to set it

noisecapella:
Functionality works for me. Can you update this PR to remove `--no-bin-links`?

We should also have an OS X user test this setup

.travis.yml
aliceriot: I have `node_env='production'` in `travis-docker-compose.yml`, which we had before. Should we not be setting that?

aliceriot: Oh I see, we had `tox` setting that.

noisecapella: We had set NODE_ENV=development in `tox.ini` which overrides anything set in the docker-compose file. But it doesn't look like there was a good reason for doing that so it's best to leave it out

aliceriot: ok, sounds good

aliceriot:
@annagav @gsidebo @giocalitri would one of you be able to make sure that everything works ok on OSX after these changes?

You'll probably need to rebuild the docker containers (we're not installing nodejs in web anymore) and remove the `node_modules` directory.

giocalitri:
I will take a look

giocalitri:
👍 it works fine (and it looks it somehow solves the problem I had with node-sass)

noisecapella:
:+1: from me too

aliceriot:
thanks @giocalitri !

aliceriot:
gonna merge now

PR #401, opened 2016-05-24
Fixed UI for month/year field

noisecapella:
#### What are the relevant tickets?
Fixes #379

#### What's this PR do?
Uses the month field to show the label. MM and YYYY are now hint text

#### Screenshots (if appropriate)
monthyear_error
monthyear_month_focused
monthyear_year_focused

PR #404, opened 2016-05-24
Replaced react-datepicker with textfields

### LINKED ISSUE #378: Personal Profile: make date picker for date of birth more material
The current date picker doesn't look like the others. 

We'll change to 3 text fields so users don't feel like they need to page backwards through the years.

noisecapella:
Rebased on #401 
Fixes #378

#### What's this PR do?
Removes react-datepicker, changes boundMonthYearField, and adds a date textbox

This is only present in the date of birth field in the personal tab. You should see this behavior:
 - a validation error should be shown after save if the date is missing or invalid
 - If the user clears one text field, all fields should clear at the same time
 - The user should be able to save a valid date
 - When the user refreshes the browser the date should show up correctly
 - The label should show up above the text fields
 - There should be MM DD YYYY hint text in the text fields

screenshot from 2016-05-24 15-19-37

### DIFF
--- a/static/js/components/EducationTab.js
+++ b/static/js/components/EducationTab.js
@@ -209,7 +209,7 @@ class EducationTab extends ProfileTab {
         {this.boundTextField(keySet('field_of_study'), 'Field of Study')}
       </Cell>
       <Cell col={6}>
-        {this.boundMonthYearField(keySet('graduation_date'), 'Graduation Date')}
+        {this.boundDateField(keySet('graduation_date'), 'Graduation Date', true)}
       </Cell>
       <Cell col={6}>
         {this.boundTextField(keySet('school_name'), 'School Name')}
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import DatePicker from 'react-datepicker';
 import moment from 'moment';
 import iso3166 from 'iso-3166-2';
 import _ from 'lodash';
@@ -8,10 +7,11 @@ import MenuItem from 'material-ui/MenuItem';
 import TextField from 'material-ui/TextField';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 
-import { DATE_FORMAT } from '../constants';
+import { ISO_8601_FORMAT } from '../constants';
 import {
   validateMonth,
   validateYear,
+  validateDay,
 } from '../util/util';
 
 // utility functions for pushing changes to profile forms back to the
@@ -281,44 +281,10 @@ export function boundCountrySelectField(stateKeySet, countryKeySet, label) {
  *
  * @param keySet {String[]} Path to look up and set a field
  * @param label {String} Label for the field
+ * @param omitDay {bool} If false, there is a date textbox in addition to month and year
  * @returns {ReactElement}
  */
-export function boundDateField(keySet, label) {
-  const {
-    profile,
-    errors,
-    updateProfile,
-  } = this.props;
-
-  let onChange = date => {
-    let clone = _.cloneDeep(profile);
-    // format as ISO-8601
-    let newDate = _.isNull(date) ? date : date.format(DATE_FORMAT);
-    _.set(clone, keySet, newDate);
-    updateProfile(clone);
-  };
-  let newDate = _.get(profile, keySet) ? moment(_.get(profile, keySet), DATE_FORMAT) : null;
-  return <div>
-    <DatePicker
-      selected={newDate}
-      placeholderText={label}
-      showYearDropdown
-      onChange={onChange}
-    />
-    <span className="validation-error-text">{_.get(errors, keySet)}</span>
-  </div>;
-}
-
-/**
- * bind this to this.boundMonthYearField in the constructor of a form component
- * to update date fields
- * pass in the name (used as placeholder), key for profile.
- *
- * @param keySet {String[]} Path to look up and set a field
- * @param label {String} Label for the field
- * @returns {ReactElement}
- */
-export function boundMonthYearField(keySet, label) {
+export function boundDateField(keySet, label, omitDay) {
   const {
     profile,
     errors,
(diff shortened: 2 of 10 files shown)


aliceriot:
checking the functionality:

- [ ] a validation error should be shown after save if the date is missing or invalid
- [ ] If the user clears one text field, all fields should clear at the same time
- [ ] The user should be able to save a valid date
- [ ] When the user refreshes the browser the date should show up correctly
- [ ] The label should show up above the text fields
- [ ] There should be MM DD YYYY hint text in the text fields

aliceriot:
the only thing I feel unsure about is the `If the user clears one text field, all fields should clear at the same time` behavior. Here's a scenario in which I think it's sub-optimal. Say I've entered a date:

date

but that's not my birthday, I meant to type `9` for day. If I click on the day field and delete the existing number all of my input goes away - so I have to type the date in correctly again. When selecting the field the current input is initially highlighting, so just clicking in the field and typing `9` would allow me to edit the value without having the fields all be cleared, but I don't think we should trust a user to know to do that.

noisecapella:
I think we can fix that behavior so that if any fields are blank, the field is treated as blank when the user submits the form.

aliceriot:
that sounds good to me!

noisecapella:
Ok, ready for review again

aliceriot:
functionality looks good!

static/js/util/profile_edit.js
```diff
@@ -281,44 +281,10 @@ export function boundCountrySelectField(stateKeySet, countryKeySet, label) {
  *
  * @param keySet {String[]} Path to look up and set a field
  * @param label {String} Label for the field
+ * @param omitDate {bool} If false, there is a date textbox in addition to month and year
```
gsidebo: shouldn't this be `omitDay`? when this is true, we just intend to omit the day input, right?

static/js/util/profile_edit.js
```diff
@@ -334,72 +300,94 @@ export function boundMonthYearField(keySet, label) {
...
+    // keep text up to date
+    _.set(clone, editKeySet, newTuple);
 
-    if (validatedMonth === undefined || validatedYear === undefined) {
-      // store the edit value and make the formatted date null so it doesn't pass validation
-      _.set(clone, editKeySet, newMonthYear);
-      _.set(clone, keySet, undefined);
-    } else if ( validatedMonth === "" || validatedYear === "") {
-      _.set(clone, editKeySet, undefined);
+    if (validatedDate === undefined || validatedMonth === undefined || validatedYear === undefined) {
```
aliceriot: maybe this might be cleaner:

```js
if ( [validatedDate, validatedMonth, validatedYear].includes(undefined) ) {
```

static/js/util/profile_edit.js
```diff
@@ -334,72 +300,94 @@ export function boundMonthYearField(keySet, label) {
...
     }
     return null;
   };
 
-  // Get a tuple { month, year } which contains the values being edited in the textbox
+  // Get a tuple { date, month, year } which contains the values being edited in the textbox
   // values may be strings or numbers. Otherwise return empty object.
-  let getMonthYear = () => {
-    let monthYear = _.get(profile, editKeySet, {});
+  let getTuple = () => {
```
gsidebo: if this is just supposed to return an object with day/month/year as keys, i think this should be called `getDateObject`/`getDateJSON`/`getSerializedDate` and update the below references to `tuple` accordingly. `getTuple` isn't very descriptive

static/js/util/profile_edit.js
```diff
@@ -334,72 +300,94 @@ export function boundMonthYearField(keySet, label) {
...
         return {
           month: date.month() + 1,
-          year: date.year()
+          year: date.year(),
+          date: date.date()
         };
       }
     }
-    return monthYear;
+    return tuple;
```
aliceriot: it looks like we're dealing with an object ( `{ month: ..., year: ..., date: ...}` ) here, which makes me feel like `tuple` is a bit of a confusing name (JS doesn't have tuples, after all).

static/js/util/util.js
```diff
@@ -355,15 +355,28 @@ export function makeStrippedHtml(textOrElement) {
 }
 
 /**
+ * Validate a month date
+ * @param {String} string The input string
+ * @returns {Number|undefined} The valid date if a valid date value or undefined if not valid
+ */
+export function validateDate(string) {
```
gsidebo: same comment above re: date vs. day

static/js/util/profile_edit.js
```diff
@@ -334,72 +300,94 @@ export function boundMonthYearField(keySet, label) {
...
+      floatingLabelText=" "
+      floatingLabelFixed={true}
+      style={{
+        maxWidth: "2em"
+      }}
+      fullWidth={true}
+      value={tuple.date !== undefined ? tuple.date : ""}
+      onChange={e => setNewDate(e.target.value, undefined, undefined)}
+    />;
+  }
```
aliceriot: I would refactor this bit to look more like:

```js
let dateSlash = () => omitDate ? undefined : <span className="slash"> / </span>;
let dateField = () => {
  if ( !omitDate ) {
    return <TextField>
      ...
    </TextField>
}
```

then in the JSX below we can just call `{dateField()}`.

idk why but this way feels cleaner to me, feel free to stick with what you have though!

static/js/components/EducationTab.js
```diff
@@ -209,7 +209,7 @@ class EducationTab extends ProfileTab {
         {this.boundTextField(keySet('field_of_study'), 'Field of Study')}
       </Cell>
       <Cell col={6}>
-        {this.boundMonthYearField(keySet('graduation_date'), 'Graduation Date')}
+        {this.boundDateField(keySet('graduation_date'), 'Graduation Date', true)}
```
aliceriot: this is nice! I like when we can build more general things like this.

static/js/util/profile_edit_test.js
```diff
@@ -428,91 +414,240 @@ describe('Profile Editing utility functions', () => {
...
+        date: "28",
+        month: "",
+        year: "2066"
+      });
+    });
+
+    it("treats an empty year string as deleting the format text but not edit data", () => {
+      that.props.profile.date_of_birth = "2066-02-28";
+      rerender(false);
+      yearTextField.props.onChange({target: {value: ""}});
```
gsidebo: do we ever use helper methods in our test files? i kinda feel like we should. lots of repetition with this case and the two above. i think at the least it would be nice to have a helper method for the assertions which takes an object of expected values

static/js/util/profile_edit.js
noisecapella: moment.js uses `date` to refer to the day of month and `day` to refer to day of month. But most other libraries refer to day like you said so I'll change the references

static/js/util/profile_edit_test.js
```diff
@@ -391,35 +360,52 @@ describe('Profile Editing utility functions', () => {
...
+      }
+    };
+
     it("has proper props for an invalid or missing value", () => {
       for (let dateOfBirth of ['', {}, null, undefined]) {
         that.props.profile.date_of_birth = dateOfBirth;
-        [monthTextField, , yearTextField] = renderMonthYearField();
+        rerender(false);
 
         assert.equal(monthTextField.props.floatingLabelText, "Date of birth");
```
gsidebo: general question: will we ever want to reuse this component to capture user input for a date that _isn't_ a date of birth? i would think the answer is yes but i'll defer to you. if we do want this to be a date-of-birth-exclusive input, i think it should have a more accurate name than `boundDateField`

aliceriot: If you look above you can see it being used on the education and employment tabs of the profile (in `EmploymentTab.js` and `EducationTab.js`)

static/js/util/profile_edit.js
noisecapella: I renamed it to `getEditObject`. It doesn't have date in the name but the function is defined within the `boundDateField` function so I think the context should be clear enough there

static/js/util/profile_edit_test.js
noisecapella: We do use helper functions. I'm not sure there's enough repetition in this file to justify a helper function. What function would you suggest?

noisecapella:
Ready for review again!

aliceriot:
:+1: LGTM

PR #405, opened 2016-05-24
Profile REST api honors privacy settings

### LINKED ISSUE #387: Honor user privacy setting when viewing profile
We have 3 types of user and 3 privacy settings, but only 7 use cases and 3 possible outcomes.

### Users  
1. The current user
2. A verified micromasters user (to be defined, but we will add a field to the user profile)
3. An anonymous use (not logged in). 

### Privacy Settings 
1. Private
2. Public to other verified micromasters users
3. Public to all

### Possible Outcomes
1. Access to the complete profile
2. Access to the public profile (all fields, except birthdate, `filled_out`, `agreed_to_terms_of_service`, `email_optin`, `pretty_printed_student_id`, `birth_city`, `birth_state_or_territory`
3. Access to a restricted profile (only the fields `username`, profile images, and `account_privacy`

### Use Cases
Case 1: User Foo looks at profile Foo (same user)
This is easy: the user should see everything: Outcome 1

Case 2: The user Foo (a MM student) looks at the profile Bar (with privacy set to "private")
Restricted Profile: Outcome 3

Case 3: the verified micromasters user Foo looks at the profile Bar (with privacy set to "private to MM students")
Public Profile: Outcome 2

Case 4: the verified micromasters user Foo looks at the profile Bar (with privacy set to "public")
Public Profile: Outcome 2 (same as case 3)

Case 5: The user Foo (not a verified micromasters user) looks at the profile Bar (with privacy set to "private")
Restricted Profile: Outcome 3

Case 6: The user Foo (not a verified micromasters user) looks at the profile Bar (with privacy set to "private to MM students")
Restricted Profile: Outcome 3 (same as case 5)

Case 7: The user Foo (not a verified micromasters user) looks at the profile Bar (with privacy set to "public")
Public Profile: Outcome 2

giocalitri:
closes #387 

#### What's this PR do?
Makes sure that the profile rest view honors the profile privacy setting

`profiles/views.py`, then `profiles/serializers.py`

You need 2 users (one with the `verified_micromaster_user` set to `True` and another with the same setting set to `False`) and a user with the profile filled out (let's call her `Foo`).
Then for each profile privacy setting of the Foo user, verify that the result of the profiles REST api works accordingly to what described in the story description.
Note you need to test also with the anonymous user.


noisecapella:
FYI I created an issue #415 to cover changes in the frontend

noisecapella:
I created a new edX user and tried to log in as them, and I got this error:

    null value in column "verified_micromaster_user" violates not-null constraint

Maybe we need to set this explicitly in pipeline_api.py?

noisecapella:
Nevermind, I checked out a different branch

noisecapella:
Functionality looks good, checking code

profiles/serializers.py
```diff
@@ -102,14 +106,13 @@ class Meta:
 
 class ProfileSerializer(ModelSerializer):
     """Serializer for Profile objects"""
-    pretty_printed_student_id = SerializerMethodField()
+    username = SerializerMethodField()
     work_history = EmploymentSerializer(many=True)
     education = EducationSerializer(many=True)
 
-    def get_pretty_printed_student_id(self, obj):
```
noisecapella: I'm guessing this was unnecessary, the serializer just needs the name in the fields list?

noisecapella:
Looks good :+1:

profiles/serializers.py
giocalitri: exactly

PR #406, opened 2016-05-24
Added ability to edit employment on profile page

### LINKED ISSUE #393: Profile viewer: allow user to edit employment info
Clicking an edit button on the row next to a job, open the employment form component.

aliceriot:

closes #393

#### What's this PR do?

This PR makes some changes to enable editing profile fields on the `/users/:username` page. Doing this entailed the following changes:

- A new class, `EmploymentForm`, which actually draws the form portion of the employment history editing UI
- A new class, `EmploymentValidation`, which simply defines the values used for validation (some `propTypes`, `defaultProps`, and the `validation` class method).
- changes to the `ProfileTab`. Now instead of a class that all the tabs inherit from, it's a mixin (called `ProfileFormFields`) which provides bindings (on `this`) for all of our input field types. I didn't really want to have `EmploymentTab extends EmploymentValidation` and `EmploymentValidation extends ProfileTab` (and `ProfileTab extends React.Component`) which is the reason for this change.
- A new class, `ProfileFormContainer`, which is a utility class that defines helpers for dispatching actions and for injecting props into `this.props.children`. Basically, the idea here is that `ProfilePage` and `UserPage` inherit from `ProfileFormContainer`, and then are connected to the store. `ProfileFormContainer` basically makes it easier to do the right thing in two places.

the diff looks a lot bigger than it really is, this is mostly moving around existing code to different places.

I would start by reviewing the changes to `EmploymentTab` and `EmploymentForm`.

Make sure there are no regressions on the profile. You should be able to edit fields and so on everywhere.

Also, visit `/users/:username` and make sure you can edit entries in the work history form. things should validate and so on as normal.

...whew

I didn't do any ui / pixel polish work in this PR, I think in one of the subsequent ones I'll try to make things look pretty, but here it is:

users_things


static/js/util/profile_edit.js
```diff
@@ -446,7 +446,7 @@ export function saveAndContinue(nestedValidationCallback, finalStep) {
   }
 
   let clone = Object.assign({}, profile, {
-    filled_out: !!finalStep
+    filled_out: profile.filled_out || finalStep
```
noisecapella: If `finalStep` is ever not a boolean (for example if we omit the argument) then this value may end up being `undefined`, which will in turn cause the `PATCH` not to send it in the JSON at all, since `JSON.stringify` ignores undefined values.

static/js/util/ProfileFormFields.js
```diff
@@ -14,8 +14,7 @@ import { HIGH_SCHOOL, ASSOCIATE, BACHELORS, MASTERS, DOCTORATE } from '../consta
 import LANGUAGE_CODES from '../language_codes';
 import iso3166 from 'iso-3166-2';
 
-
-class ProfileTab extends React.Component {
+export const ProfileFormFields = Base => class extends Base {
```
noisecapella: Why not `export default` here, since there's only one thing being exported?

static/js/components/EmploymentForm.js
```diff
@@ -0,0 +1,223 @@
...
+import IconButton from 'react-mdl/lib/IconButton';
+import _ from 'lodash';
+import moment from 'moment';
+
+import { saveAndContinue } from '../util/profile_edit';
+import { generateNewWorkHistory } from '../util/util';
+import EmploymentValidation from  '../util/EmploymentValidation';
+import { ProfileFormFields } from '../util/ProfileFormFields';
+
+class EmploymentForm extends ProfileFormFields(EmploymentValidation) {
```
noisecapella: It feels like the inheritance structure in these classes is getting complicated. Is there a way we can move the validation pieces into a standalone class which can be used by the other classes? We should generally favor composition over inheritance because the relationships are made more explicit.

noisecapella: In short I'm suggesting moving from this to

    class EmploymentForm extends ProfileFormFields {
      static validationClass = EmploymentValidation;
      ...

static/js/dashboard_routes.js
```diff
@@ -37,7 +38,9 @@ export function makeDashboardRoutes(browserHistory, store, onRouteUpdate) {
               <Route path="privacy" component={PrivacyTab} />
             </Route>
             <Route path="/terms_of_service" component={TermsOfServicePage} />
-            <Route path="/users/:username" component={UserPage} />
+            <Route path="/users" component={UserPage} >
```
noisecapella: Is there any case where we would want to use `UserPage` without `User`? When the user goes to `/users` they will see `UserPage` without any child components

aliceriot: I could add an `IndexRedirect` here that would redirect to `/users/${username}` where username is the users username. Would that make sense?

The reason I wanted to have the `:username` route nested under `/users/` is that it simplifies sharing the prop injection pattern between `ProfilePage` and `UserPage` (both have routes underneath them and inject props into `this.props.children` to provide helper functions for dispatching actions and data from the store).

noisecapella: Sounds good

giocalitri:
please rebase

aliceriot:
@noisecapella I just finished writing tests and getting all the tests I broke to pass (ha!). Ready for review!

static/js/components/User.js
```diff
@@ -5,18 +5,19 @@ import IconButton from 'react-mdl/lib/IconButton';
...
   };
 
   render() {
     const { profile } = this.props;
 
-    let getStateName = () => {
-      return iso3166.subdivision(profile.state_or_territory).name;
-    };
+    let getStateName = () => (
+      iso3166.subdivision(profile.state_or_territory).name
```
noisecapella: I ran into an error here when I went to a missing profile page. I think we need some code in this page to at least guard if `profile.state_or_territory` is undefined. I'm not sure how we want our 404 page to look like or function (cc @pdpinch)

noisecapella:
Can you rebase on master please?

aliceriot:
@noisecapella rebased

noisecapella:
Can you update the employment code to filter out work history items without ids?
screenshot from 2016-05-27 14-32-28

aliceriot:
sure, can you explain what situation that's occuring in?

noisecapella:
The entry there comes from the item currently being edited in the dialog box. If we filter out entries where id = undefined, it should show only work history items which have been persisted via the API

aliceriot:
Oh Ok, I see what you mean. just a sec!

noisecapella:
Functionality looks good, reviewing code now

static/js/components/EmploymentForm.js
```diff
@@ -0,0 +1,224 @@
...
+    );
+  }
+
+  renderWorkHistory () {
+    const { ui, profile: { work_history } } = this.props;
+    if ( ui.workHistoryEdit === true ) {
+      let workHistoryRows = [];
+      if ( !_.isUndefined(work_history) ) {
+        workHistoryRows = work_history.filter(entry =>
+          entry.id !== undefined
```
noisecapella: The `filter` will mess up the indexes which will cause problems with deleting and editing here (although in this case the user will probably be unable to click these buttons while the dialog is open). Can you use `Object.entries` here similar to in EducationTab to preserve the correct index value?

noisecapella:
Sorry, need to rebase again :)

static/js/containers/ProfileFormContainer.js
```diff
@@ -0,0 +1,151 @@
...
+  setWorkDialogIndex,
+  setEducationDialogVisibility,
+  setEducationDialogIndex,
+  setEducationDegreeLevel,
+  setEducationDegreeInclusions,
+} from '../actions/ui';
+
+class ProfileFormContainer extends React.Component {
+  static propTypes = {
+    profile:    React.PropTypes.object,
```
noisecapella: Where does profile come from? I don't see it used in this class, though maybe you use it in a subclass

aliceriot: It's injected via `connect` into the two classes that inherit from this class (`ProfilePage` and `UserPage`).

aliceriot: I set the `propTypes` here to DRY things up a bit (lots of shared stuff between `UserPage` and `ProfilePage`).

noisecapella: It looks like only `profiles` and `ui` are injected using connect. The React inspector doesn't show these props being sent to either `ProfilePage` or `UserPage`.

aliceriot: oh yes you're right, I forgot that I changed things around to pull the relevant profile out of profiles in each page's render method.

noisecapella:
Currently I am unable to view other users in the users page, though we can fix that as part of #415 if you prefer

aliceriot:
That seems like something it would make sense to do all at once (all of the logic relating to viewing other users) to me, so if that sounds like a good approach to you then let's do it

static/js/util/validation.js
```diff
@@ -0,0 +1,115 @@
+import _ from 'lodash';
+
+let handleNestedValidation = (profile, keys, nestedKey) => {
+  let nestedFields = index => (
+    keys.map(key => [nestedKey, index, key])
+  );
+  return profile[nestedKey].map((v, i) => nestedFields(i)).reduce((a,b) => a.concat(b));
```
noisecapella: The `concat` here in the `reduce` will cause a new list to be created for each iteration of `reduce`, though the number of lists here is likely to be small. Instead could you use `_.flatten`?

aliceriot: sure

static/js/util/validation_test.js
```diff
@@ -0,0 +1,71 @@
...
+import { USER_PROFILE_RESPONSE } from '../constants';
+
+describe('Profile validation functions', () => {
+  describe('Personal validation', () => {
+    it('should return an empty object when all fields are present', () => {
+      assert.deepEqual({}, personalValidation(USER_PROFILE_RESPONSE));
+    });
+
+    it('should return an appropriate error if a field is missing', () => {
+    });
```
noisecapella: It looks like this test case is missing

aliceriot: indeed it is! oopsie

noisecapella:
This PR is in good shape overall, just a few comments

aliceriot:
@noisecapella I think I've addressed all your comments!

noisecapella:
Looks good, squash and merge :+1:

Issue #373, opened 2016-05-22
On Dashboard, gray-out "Apply for Masters" button


pdpinch:
The button isn't clickable, but it's still too subtle. We need to actually change the color of the button and the text. 

See, for example, https://getmdl.io/components/index.html#buttons-section

amir-qayyum-khan:
@pdpinch what should be change in text?

pdpinch:
The text should be:

```
Apply for Supply Chain Management Master’s Degree
```

I forgot that it was already white, so the color is fine. 

The button should be greyed out, as it is in the design. I.e. `#B2C9D2`

PR #411, opened 2016-05-25
Fixed design of apply for master button on dashboard

amir-qayyum-khan:
Fixes https://github.com/mitodl/micromasters/issues/373

#### What's this PR do?
- Fixed design of apply for master button of the program in dashboard.

- CourseList.js

- Open dashboard and see at the end of a program.

screen shot 2016-05-25 at 6 50 13 pm

### DIFF
--- a/static/js/components/CourseList.js
+++ b/static/js/components/CourseList.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
-import Button from 'react-bootstrap/lib/Button';
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
+import RaisedButton from 'material-ui/RaisedButton';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
 import { toggleDashboardExpander } from '../actions/ui';
@@ -88,6 +89,11 @@ class CourseList extends React.Component {
         </div>;
       });
 
+      let applyForMSBtnLabel = `Apply for the ${program.title} Master’s Degree`;
+      const btnStyle = {
+        'text-transform': 'none'
+      };
+
       return (
         <div key={program.id} className="program">
           <Grid>
@@ -112,8 +118,10 @@ class CourseList extends React.Component {
           <div className="apply-for-ms">
             <br/>
             <p>You need to pass {totalCourses - coursesPassed} more courses
-              before you can apply for <strong>{program.title}</strong> Master's Degree.</p>
-            <Button bsStyle="primary" disabled>Apply for <strong>{program.title}</strong> Master Degree</Button>
+              before you can apply for the <strong>{program.title}</strong> Master’s Degree.</p>
+            <MuiThemeProvider>
+                <RaisedButton label={applyForMSBtnLabel} disabled={true} labelStyle={btnStyle} />
+            </MuiThemeProvider>
           </div>
         </div>
       );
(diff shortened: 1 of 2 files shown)


static/js/components/CourseList.js
```diff
@@ -112,8 +112,10 @@ class CourseList extends React.Component {
           <div className="apply-for-ms">
             <br/>
             <p>You need to pass {totalCourses - coursesPassed} more courses
-              before you can apply for <strong>{program.title}</strong> Master's Degree.</p>
-            <Button bsStyle="primary" disabled>Apply for <strong>{program.title}</strong> Master Degree</Button>
+              before you can apply for <strong>{program.title}</strong> Master’s Degree.</p>
```
pdpinch: Copy here needs an article. 

```
before you can apply for the <strong>{program.title}</strong> Master’s Degree.</p>
``

File: static/scss/dashboard.scss [hunk 2]
```diff
@@ -287,3 +287,15 @@ c.validation-wrapper {
 .header-privacy-tab {
   color: #A2AAB3;
 }
+
+button.btn-apply-for-ms {
+  font-family: "Open Sans";
+  background: #B2C9D2;
```
pdpinch: This color is only for the disabled state. I know we don't show the button enabled yet, but doesn't it make sense to put this in the disabled style?

Ferdi:
I'm afraid that wont pass accessibility requirements - low contrast.   Can you please look at Material Design guidelines / examples ?

pdpinch:
Good point Ferdi. I wasn't watching that. 

However, as far as I can tell, the mdl disabled button is low contrast too. See https://www.google.com/design/spec/components/buttons.html#buttons-raised-buttons

This led me to this [stackoverflow question](http://ux.stackexchange.com/questions/80881/do-disabled-buttons-still-need-to-be-contrast-compliant-for-accessibility), which points to the [relevant w3c standard](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html):

> Text or images of text that are part of an inactive user interface component... have no contrast requirement.

Should we go ahead and use the mdl-recommended grey button with grey text? Or leave it as is?

amir-qayyum-khan:
@pdpinch we can also look into material-ui buttons.
please see this as well.
http://www.material-ui.com/#/components/raised-button

pdpinch:
The material-ui treatment for disabled buttons is the same as the Google material design guidelines. Go with whichever is easier for you to implement. 

I'll check with our accessibility director, but we should proceed.

amir-qayyum-khan:
should i use same color i.e grayish text that is in the material-ui example?

pdpinch:
Yes. You can use the same gray colors from the material-ui example for now.

amir-qayyum-khan:
@pdpinch made the change and button looks like this.

screen shot 2016-05-28 at 12 30 54 am

static/js/components/CourseList.js
```diff
@@ -88,6 +89,12 @@ class CourseList extends React.Component {
         </div>;
       });
 
+      let applyForMSBtnLabel = `Apply for the ${program.title} Master’s Degree`;
+      const btnStyle = {
+        'text-transform': 'none',
+        'padding': '25px 16px'
+      };
```
pdpinch: Why is this necessary?

amir-qayyum-khan: if i not use text-transform the all text will be capitalize.

amir-qayyum-khan:
@gsidebo @pdpinch please review

pdpinch:
👍

PR #417, opened 2016-05-25
Fixed radio options to be shown if nothing is selected

### LINKED ISSUE #416: Bug: radio buttons don't show up if value is null
Instead, they should all show up but none should be selected

noisecapella:
Fixes #416 

#### What's this PR do?
Changes the behavior to show radio buttons even when nothing is selected, and adds better tests for it. There is also now a label and error text.

Set gender to null for a user, then go to the personal tab. You should see three options which are all unchecked. Click save, and you should see a validation error in red text below the gender options. You should be able to select an option, click Save, and have it proceed.

We should be more careful on code reviews in the future. There was no code for validation and the tests referenced props that don't exist in the RadioButtonGroup component.


aliceriot:
:+1: works as expected, code looks clean!

PR #420, opened 2016-05-26
Switched to social auth username

### LINKED ISSUE #239: The requests to edX that contain the username should use the proper one
Currently we assume that the local username on micromaster is the same we have on edX.

This might not be the case in some edge situations. The solution is to pull the edX username from the `User social Auth` model. The right field is `uid`.

noisecapella:
Fixes #239

#### What's this PR do?
Uses the username stored in the social auth model instead of the one attached to the `User` model. In logging we still use `user.username` but in all other places we should use the one attached to the edx social auth model instead. Tests are updated to provide a different username for the social auth model.

Also, the logic to show either the username or preferred name was moved out of the server into the JS in the function `getPreferredName`.

backends/pipeline_api.py

 - Find a user who has an account on edX but who has not yet logged into edX via MicroMaster's. If necessary, delete an existing user from MM using the Django shell first.
 - Create a new user using the Django admin, or the shell, or using createsuperuser. Do not create it by logging in. The user must have the same username as your edX user.
 - Log into MM using the web site. Everything should work as expected. You should see your username without any characters appended.
 - In the Django shell, verify that a user was created such that `user.social_auth.get(provider='edxorg').uid` is your username but `user.username` has some extra random characters appended to it.
 - Fill out the profile and make sure everything works like before.

### DIFF
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -5,6 +5,7 @@ import logging
 from datetime import datetime
 from urllib.parse import urljoin
 
+from backends.edxorg import EdxOrgOAuth2
 from profiles.models import Profile
 from profiles.util import split_name
 
@@ -28,7 +29,7 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):  
 
     # this function is completely skipped if the backend is not edx or
     # the user has not created now
-    if backend.name != 'edxorg' or not is_new:
+    if backend.name != EdxOrgOAuth2.name or not is_new:
         return
 
     access_token = response.get('access_token')
@@ -45,8 +46,9 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):  
         log.error('No profile found for the user %s', user.username)
         return
 
+    username = user.social_auth.get(provider=EdxOrgOAuth2.name).uid
     user_profile_edx = backend.get_json(
-        urljoin(backend.EDXORG_BASE_URL, '/api/user/v1/accounts/{0}'.format(user.username)),
+        urljoin(backend.EDXORG_BASE_URL, '/api/user/v1/accounts/{0}'.format(username)),
         headers={
             "Authorization": "Bearer {}".format(access_token),
         }
--- a/cms/models.py
+++ b/cms/models.py
@@ -38,7 +38,9 @@ class HomePage(Page):
 
         username = None
         if not request.user.is_anonymous():
-            username = request.user.social_auth.get(provider=EdxOrgOAuth2.name).uid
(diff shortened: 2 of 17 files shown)


noisecapella:
By the way staff users will get a 500 error when they visit the front page since they don't have an edX social auth defined. We can work around that but I'm not sure to what degree we want to support that case, since staff users will probably just want to go straight to /admin/

backends/pipeline_api.py
```diff
@@ -45,8 +46,9 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
         log.error('No profile found for the user %s', user.username)
         return
 
+    username = user.social_auth.get(provider=EdxOrgOAuth2.name).uid
     user_profile_edx = backend.get_json(
-        urljoin(backend.EDXORG_BASE_URL, '/api/user/v1/accounts/{0}'.format(user.username)),
+        urljoin(backend.EDXORG_BASE_URL, '/api/user/v1/accounts/{0}'.format(username)),
```
giocalitri: Note for the future: We should use the edx-api-client for this...

cms/models.py
```diff
@@ -35,14 +36,17 @@ def get_context(self, request):
             "host": webpack_dev_server_host(request)
         }
 
+        username = None
+        if not request.user.is_anonymous():
+            username = request.user.social_auth.get(provider=EdxOrgOAuth2.name).uid
```
giocalitri: does this mean that a staff user will get a 500 in the CMS?

noisecapella: Yes. Fixing it here is not difficult but I'm not sure where we want staff users to be able to go

giocalitri: well, I guess the staff users need at least to be able to work in the CMS and the django admin. We can open another issue to handle the dashboard.

giocalitri:
The code is fine overall, but the staff users (user who do not login via edX) should be able to use the CMS as well.

noisecapella:
Ok, staff users should now be able to see CMS pages. They will have no username, but I'm not sure that's important

cms/models.py
```diff
@@ -38,7 +38,9 @@ def get_context(self, request):
 
         username = None
         if not request.user.is_anonymous():
-            username = request.user.social_auth.get(provider=EdxOrgOAuth2.name).uid
+            social_auths = request.user.social_auth.filter(provider=EdxOrgOAuth2.name)
+            if social_auths.exists():
+                username = social_auths.first().uid
```
giocalitri: This work, but a DoesNotExist exception handling was working as well, I guess

giocalitri:
can you please rebase master?

noisecapella:
Rebased

giocalitri:
Functionality is good
👍 after squash and tests pass again

PR #422, opened 2016-05-26
Converted industry text field to select field

### LINKED ISSUE #315: Make "industry" a select / dropdown element
Can we make this a select element ?

industry

It looks like this is the list linkedin uses https://developer.linkedin.com/docs/reference/industry-codes

noisecapella:
Fixes #315 

#### What's this PR do?
Adds the Linkedin industry list and converts our text field to a select field using that list.

On the employment panel in the profile, click the plus to add and select an industry in the list


aliceriot:
:+1:

PR #423, opened 2016-05-26
Changed AutoComplete to focus on text field after selecting an item

### LINKED ISSUE #419: Bug: Autocomplete + Enter key breaks tabbing behavior in profile form fields
To reproduce:

- log in and navigate to the profile edit page
- highlight an autocompleted form field like "Preferred language"
- type a partial value, use the arrow keys to select an autocomplete option, and press Enter
- press Tab

Expected: the next input would be selected as if the Enter key was never pressed
Actual: the MIT Micromasters logo is highlighted and the page scrolls to the top

Looks like hitting Enter loses the active input, and Tab just selects the first tab-able element on the page

noisecapella:
Fixes #419

#### What's this PR do?
Focuses the text field after selecting an item. This should let the user tab after pressing enter

In an AutoComplete component, type some text. Press the down key and highlight an option you want to select. Press enter. The item should be selected and the text field should be focuses. When you press tab you should go to the next field. Also try pressing backspace after pressing enter to select an item, and verify that the behavior is expected.

### DIFF
--- a/static/js/components/AutoComplete.js
+++ b/static/js/components/AutoComplete.js
@@ -269,6 +269,11 @@ class AutoComplete extends Component {
         searchText: searchText,
       });
       this.close();
+      this.setState({
+        focusTextField: true
+      }, () => {
+        this.focus();
+      });
       this.timerTouchTapCloseId = null;
     }, this.props.menuCloseDelay);
   };


giocalitri:
this code by itself works as expected, but there is a side-effect: when I click in a field that has already a value, it does not limit the dropdown menu to that value:

tabbing_dropdown

noisecapella:
That's by design, we don't filter the list unless the user has started typing something. If they click the list it will show everything.

giocalitri:
Ok, then :1:

giocalitri:
Or whatever is the symbol for thumb up :-)

PR #425, opened 2016-05-26
Support for gravatar

### LINKED ISSUE #418: Gravatar support for images
Based on [this comment](https://github.com/mitodl/micromasters/issues/394#issuecomment-221675266) we are going to implement Gravatar support for profile images.

We are also going to postpone the implementation of #394 and #179

giocalitri:
closes #418 

#### What's this PR do?
Introduces support for gravatar images. It also removes the images URL from the profile

`profiles/serializers.py`

You need to create an account with an email that has a gravatar image. Also test that for an email that has not a gravatar image, the default image is shown.

To create a gravatar account go to https://en.gravatar.com


profiles/models.py
```diff
@@ -122,10 +122,7 @@ class Profile(models.Model):
     )
 
     has_profile_image = models.BooleanField(default=False)
```
noisecapella: Do we still need this field?

profiles/util.py
```diff
@@ -21,3 +23,39 @@ def split_name(name):
...
+        size (int): an integer representing the image size to be requested
+    Returns:
+        string: a string representi the gravatar url
+    """
+    base_gravatar_url = "https://www.gravatar.com/avatar/"
+    default_image_url = "https://s3.amazonaws.com/odl-micromasters-production/avatar_default.png"
+
+    # return the gravatar image
+    md5_hash = hashlib.md5()
+    md5_hash.update(user_email.lower().encode())
```
noisecapella: It looks like the official docs include a trim function: https://en.gravatar.com/site/implement/hash/
Can we add it here, even if it's not strictly necessary for our case?

noisecapella:
Functionality works fine

noisecapella:
Code looks good :+1:

profiles/util.py
giocalitri: you are right: will do

profiles/models.py
giocalitri: I thought it might be useful in the future... no strong opinion

noisecapella: No strong opinion here either

PR #427, opened 2016-05-26
Added missing action to integration tests

noisecapella:
#### What's this PR do?
Adds an action we should be listening for. There's no reason to listen for REQUEST_DASHBOARD without RECEIVE_DASHBOARD_SUCCESS

If tests pass, everything should be fine


giocalitri:
tests pass, so 👍

PR #429, opened 2016-05-26
Added minimal Dockerfile to run tests with

### LINKED ISSUE #428: node_modules is owned by root
When we switched to the `node:6.2` image we started running its commands as root. We can create a minimal Dockerfile to add the mitodl user to fix this.

FYI, the mitodl user is uid 1000 which happens to be the default first normal user, this is why files created inside the container are owned by the same user as the one outside the container. If you had two normal users on your system the docker container would be creating files with the uid of the first user which is a bit surprising.

noisecapella:
Fixes #428 

#### What's this PR do?
Adds a minimal Dockerfile for the watch image which runs commands as the mitodl user

### DIFF
--- a/Dockerfile-node
+++ b/Dockerfile-node
@@ -1,0 +1,5 @@
+FROM node:6.2
+MAINTAINER ODL DevOps <mitx-devops@mit.edu>
+
+RUN adduser --disabled-password --gecos "" mitodl
+USER mitodl
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,8 @@ web:
     - db
 
 watch:
-  image: node:6.2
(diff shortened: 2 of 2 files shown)


giocalitri:
I do not understand what you could not do before this PR

noisecapella:
When @aliceriot changed the `watch` container to use the `node` image from the docker hub instead of the same image as `web`, all commands in `node` were being run as root. This meant that the files created locally were also owned by root since they had the same uid

giocalitri:
oh, I see: that's why I do not see the problem 👍

giocalitri:
maybe @aliceriot wants to test it

aliceriot:
LGTM

PR #445, opened 2016-05-27
Added ability to edit education to users page

### LINKED ISSUE #392: Profile viewer: allow user to edit educational info
Clicking an edit button on the row next to a degree, open the degree form component.

aliceriot:

closes #392 

#### What's this PR do?

This does a similar thing to #406, we move the actual form components out of `EducationTab` and into `EducationForm`, which we can then drop on the `User` page.

Review the changes to `EducationTab` and `User`.

Make sure that you can edit all educational entries on both the `User` page and on the `EducationTab`.


static/js/components/EducationForm.js
```diff
@@ -0,0 +1,285 @@
...
+import { Card } from 'react-mdl/lib/Card';
+import _ from 'lodash';
+import moment from 'moment';
+
+import ProfileFormFields from '../util/ProfileFormFields';
+import { generateNewEducation } from "../util/util";
+import { saveAndContinue } from '../util/profile_edit';
+import { HIGH_SCHOOL } from '../constants';
+
+class EducationForm extends ProfileFormFields {
```
amir-qayyum-khan: why not  instance property pattern ?
```js
class EducationForm extends ProfileFormFields {
   educationLevelLabels = {};
```

static/js/components/EducationForm.js
```diff
@@ -0,0 +1,285 @@
...
+      setEducationDialogIndex,
+      setEducationDegreeLevel,
+      setEducationDialogVisibility,
+    } = this.props;
+    let newIndex = index;
+    if (index === null){
+      newIndex = profile['education'].length;
+    }
+    /* add empty education */
+    let clone = Object.assign({}, profile);
```
amir-qayyum-khan: can you rename it to meaningful name like profileClone?

static/js/components/EducationForm.js
```diff
@@ -0,0 +1,285 @@
...
+    clone['education'] = clone['education'].concat(generateNewEducation(level));
```
amir-qayyum-khan: what is level here? can you add doc strings to these modules?

static/js/components/EducationForm.js
aliceriot: We use `clone` for similar temporary objects elsewhere in the codebase, and I think given that we're only using it on three lines it's clear what's going on. I'd prefer to stick with the naming convention we're using.

static/js/components/EducationForm.js
aliceriot: This code is all copied from an existing class, i.e. this was the constructor for `EducationTab` before this change. I was trying to get the behavior I wanted with as little refactoring as possible, which I why I didn't change anything here.

I suppose it would make sense to do some refactoring at the same time, but I didn't want to make the PR more complicated or harder to review. This applies also to your comment on line 53 below, I could refactor that method as well but I held off on doing so because I'm really just renaming an existing class and changing some of the JSX in `render()`.

Does that make sense? I'm down to refactor these method if we think it's important, just wanted to share my thought process.

amir-qayyum-khan:
@aliceriot 
- Can you add screenshots, I dont know what is user page. 
- After adding high school i am not able to see what school i just added. Is this ok. 
- by the way i disabled all edu levels

screen shot 2016-06-01 at 11 59 38 pm

amir-qayyum-khan:
- Saving a school takes time i.e dialog in not responsive,
- Novice use can click many times because he dont get clue what is going on.

screen shot 2016-06-02 at 12 00 32 am

amir-qayyum-khan:
switch for ``Associate degree`` and ``Associate degree`` never gets disable. Each time i disable the and press save and continue and the visit education tab if found them enable. Is this right behaviour?



screen shot 2016-06-02 at 12 26 18 am

screen shot 2016-06-02 at 12 22 04 am

aliceriot:
Would you mind checking out master and seeing if the behavior is the same there? I think for this PR we should just be ensuring there are no specific regressions, and if you spot issues or usability problems we can file new issues to deal with those.

static/js/components/EducationForm.js
aliceriot: see above, but this is just moving existing code around.

amir-qayyum-khan:
@aliceriot 
- I can not save hight school because of error ``{"education":[{},{"field_of_study":["This field may not be null."]}]}`` and there is no such field like field_of_study on dialog
- slow to close is persist on master branch
- disable issue persist on master branch

screen shot 2016-06-02 at 12 00 32 am

aliceriot:
OK, I think that was fixed on master, I'll let you know when the branch is
rebased and ready for review again.

aliceriot:
@amir-qayyum-khan I just rebased and updated the remote branch - it looks like that issue with the high school validation is fixed.

static/js/components/EducationForm.js
aliceriot: so for the constructor here, I can move `educationLevelLabels` to be an instance property, which makes the constructor a little shorter, but I need to keep the `forEach` loop either here or in maybe `componentWillMount` so ensure it gets run before and rendering happens. 

what do you think? I think if we need to have the constructor for the `forEach` it doesn't really make sense to move the one thing to an instance property, but maybe it is a little cleaner that way.

noisecapella:
Code looks good to me, looking at functionality now

noisecapella:
:+1:

PR #446, opened 2016-05-27
Updated Programs page

### LINKED ISSUE #370: create program template for CMS
We can create programs and program pages through the Wagtail CMS, but we can't actually publish the pages because there's no template. 

@nicolaeherrera has create a template for us at https://github.com/mitodl/micromasters-templates/blob/master/program.html. It just needs to be adapted.

annagav:
Fix #370 
#### What's this PR do?
Use Nicolae's template for Program page. 
Add FAQs and Courses panels to Programs page. 
cms/models.py
As an instructor: 
1)login to wagtail as a superuser 
2)Create a ProgramPage, fill out all the required fields. FAQs and Courses should appear as multifield panels, add a few.
As a user: Go to homepage, click on the link for any program. Explore the page with all three tabs populated with your data: Description, FAQs, Courses.


pdpinch:
@annagav can you take a look at these test failures?

noisecapella:
Is `Courses` supposed to be a general text description or are we listing the courses there too?

annagav:
It should be a list of courses with name and description for each course.

noisecapella:
Should we use the title and description from the `Course` model for that purpose, or does it need to be editable by the user in the CMS?

annagav:
I think we decided not to link them, and have the user edit it.

cms/templates/cms/program_page.html
```diff
@@ -5,8 +5,248 @@
...
+                                                           aria-expanded="true">
+                                                            {{ course.title }}
+                                                        </a>
+                                                    </div>
+                                                    <div class="panel-collapse collapse in"
+                                                         id="course-{{ forloop.counter }}"
+                                                         aria-labelledby="exampleHeadingContinuousOne"
+                                                         role="tabpanel">
+                                                        <div class="panel-body">
+                                                            {{ course.description }}
```
noisecapella: It looks like this is outputting escaped HTML. Can you figure out how this should work so we see the formatted output here?

cms/models.py
```diff
@@ -59,4 +60,60 @@ class ProgramPage(Page):
...
+        context["style_public_src"] = get_bundle_url(request, "style_public.js")
+        context["authenticated"] = not request.user.is_anonymous()
+        context["username"] = request.user.username
+        context["js_settings_json"] = json.dumps(js_settings)
+        context["title"] = self.title
+
+        return context
+
+
+class Courses(Orderable):
```
noisecapella: We should use singular names for models. Can you rename this `Course` and `FrequentlyAskedQuestion`?

cms/migrations/0008_frequentlyaskedquestions.py
```diff
@@ -0,0 +1,31 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='FrequentlyAskedQuestions',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('sort_order', models.IntegerField(blank=True, editable=False, null=True)),
+                ('question', models.TextField()),
+                ('answer', models.TextField()),
+                ('page', modelcluster.fields.ParentalKey(on_delete=django.db.models.deletion.CASCADE, related_name='FAQs', to='cms.ProgramPage')),
```
noisecapella: Can you change `related_name` to `faqs`? Usually these names are lowercase

cms/migrations/0007_courses.py
```diff
@@ -0,0 +1,32 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Courses',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('sort_order', models.IntegerField(blank=True, editable=False, null=True)),
+                ('title', models.CharField(default='', max_length=255)),
+                ('description', wagtail.wagtailcore.fields.RichTextField(blank=True, null=True)),
+                ('program_page', modelcluster.fields.ParentalKey(on_delete=django.db.models.deletion.CASCADE, related_name='Courses', to='cms.ProgramPage')),
```
noisecapella: Can you change the related_name to `courses`?

cms/models.py
annagav: Course collides with `courses.models Course`

cms/models.py
```diff
@@ -59,4 +60,60 @@ class ProgramPage(Page):
...
+            ]
+        )
+    ]
+
+
+class FrequentlyAskedQuestions(Orderable):
+    """
+    FAQs for the program
+    """
+    page = ParentalKey(ProgramPage, related_name='FAQs')
```
noisecapella: Can you rename to `program_page` for consistency with its use in the above model?

cms/models.py
noisecapella: They're in different apps though. If we need to use both `Course` models somewhere we should import one or the other with a different name

noisecapella:
Functionality looks great

noisecapella:
Can you update `program_page.html` to set the web page title using the title block already defined in `base.html`?

courses/models.py
```diff
@@ -2,7 +2,6 @@
 Models for course structure
 """
 from datetime import datetime
-
```
noisecapella: Can you revert this change to make the commit cleaner?

cms/templates/cms/program_page.html
```diff
@@ -5,8 +5,248 @@
...
-<h1>{{ page.title }}</h1>
+    <div class="app-media layout-boxed">
+        <div class="main-content-wrapper">
+            <section class="main-content">
+                <div class="banner-wrapper banner-sub banner-program-1">
+                    <nav class="navbar navbar-default" role="navigation">
+                        <div class="container-fluid">
+                            <div class="navbar-header">
+                                <a class="navbar-brand" href="javascript:void(0)"><img
+                                        src="/static/images/logo-mit-micromasters-whitered@2x.png" width="215"
```
noisecapella: Actually we should copy `logo-mit-micromasters-whitered.png` and use `width="215"` and `height="40"` instead

noisecapella:
Can you refactor the footer into a partial template shared between home_page.html and program_page.html?

noisecapella:
Can you add tests to ui/views_test.py similar to the existing tests for the home page? Specifically these tests:
 - The ones starting with `test_index_context` should be copied and used for a program page
 - The two tests `test_login_button` and `test_sign_out_button`

annagav:
@noisecapella, I added the tests, and addressed all the comments.

noisecapella:
Functionality looks good, looking at code one more time

cms/templates/cms/program_page.html
```diff
@@ -1,12 +1,234 @@
...
-<h1>{{ page.title }}</h1>
+  <div class="app-media layout-boxed">
+    <div class="main-content-wrapper">
+      <section class="main-content">
+        <div class="banner-wrapper banner-sub banner-program-1">
+          <nav class="navbar navbar-default" role="navigation">
+            <div class="container-fluid">
+              <div class="navbar-header">
+                <a class="navbar-brand" href="javascript:void(0)"><img
+                        src="/static/images/logo-mit-micromasters-whitered@2x.png"
```
noisecapella: Can you change this to `logo-mit-micromasters-whitered.png` from the templates repository, then change width and height to be half their values?

cms/templates/cms/program_page.html
```diff
@@ -1,12 +1,234 @@
...
+          <div class="tab-content tab-content-programs">
+            <div role="tabpanel" class="tab-pane active fade in" id="home">
+
+
+              <div class="panel-body">
+                <div class="row">
+                  <div class="col-md-8">
+                    <h3>{{ page.title }}</h3>
+
+                    <p>{{ page.description|safe }}</p>
```
noisecapella: Can you change to `|richtext`? The [documentation](http://docs.wagtail.io/en/v1.5/topics/writing_templates.html) says that this will also allow the user to show images and handles internal links correctly which the `safe` filter won't do

annagav:
Addressed last two comments!

noisecapella:
Looks great :+1:

Issue #369, opened 2016-05-20
add wizard / progress UI to profile / enrollment forms


pdpinch:
Instead of tabs, the profile / enrollment forms should have an indicator so users know which step they are on. 

There are no links to jump between forms. The only way to advance is to click the "save and continue" button. (we need to add a "Previous Step" button but there's no design yet)

Here is the most recent design:
![step two - alt wizard 2](https://cloud.githubusercontent.com/assets/430126/15434412/afa501ce-1e84-11e6-8f40-2a23bf0a4a3e.png)

noisecapella:
@gsidebo FYI this will need tests too, similar to existing tests for `makeCourseProgressDisplay`

noisecapella:
It seems like we should remove the react-router handling for the tabs as well, so the user can't just go to /profile/education in the browser as a workaround. What do you think @aliceriot?

aliceriot:
hmm that might be what we want but I think that will entail a decent refactor of how the `ProfilePage` component injects props into the tab components - we might need a dummy component that just renders:

```js
<ProfilePage>
  <EmploymentTab />
</ProfilePage>
```

to keep the existing pattern. Either way I think this is blocked until https://github.com/mitodl/micromasters/issues/375 and https://github.com/mitodl/micromasters/issues/393 are merged, the former because we need the 'back' button and the latter because it entails a major refactor of related classes anyway (`ProfilePage`, mainly).

noisecapella:
Ok, I'll work on removing the links but otherwise not touch the react-router pieces. I'll file a new issue to cover the removal of the react-router support for individual profile pages

aliceriot:
I do agree that it doesn't necessarily make sense to have the user able to navigate around by URL, but at the same time with the `requireProfileFilledOut` and `requireCompleteProfile` methods on `App.js` we might have those edge cases covered anyway - meaning, it shouldn't be possible for a user to do any interaction with the site until they've filled out the profile, so someone editing the url to skip ahead to e.g. `/profile/professional` and skip the education form might not be a problem (since they'll be redirected to the education tab as soon as they click 'I'm done' on the privacy tab).

noisecapella:
At the moment we don't actually validate that the user has to enter any work history or education, even if the switches are enabled indicating they had something to add... but when we fix that it should work like you said

aliceriot:
ah
right yeah, for the work history validation if `profile.work_history` is an empty array we don't validate it.

PR #447, opened 2016-05-31
Added previous button to profile pages

### LINKED ISSUE #375: Profile: add previous button
As a user, I would like to be able to navigate back through the profile if need to make changes. 

Add a previous button to the user profile steps, to the leff of the "save and continue" buttons on the Education, Employment and Privacy steps.

gsidebo:

Closes #375 

#### What's this PR do?

adds the 'previous' button to appropriate profile pages. this accounts for a new profile navigation along with #369 

fill out a profile starting with /profile/personal and use the "Previous" and "Save And Continue" buttons

ss 2016-05-31 at 10 54 08 am
ss 2016-05-31 at 10 54 18 am
ss 2016-05-31 at 10 54 28 am

### DIFF
--- a/static/js/components/EducationTab.js
+++ b/static/js/components/EducationTab.js
@@ -12,7 +12,7 @@ import moment from 'moment';
 
 import ProfileFormFields from '../util/ProfileFormFields';
 import { generateNewEducation } from "../util/util";
-import { saveAndContinue } from '../util/profile_edit';
+import { saveProfileStep } from '../util/profile_edit';
 import { HIGH_SCHOOL } from '../constants';
 
 class EducationTab extends ProfileFormFields {
@@ -23,8 +23,10 @@ class EducationTab extends ProfileFormFields {
       this.educationLevelLabels[level.value] = level.label;
     });
   }
-  nextUrl = "/profile/professional";
 
+  prevUrl = "/profile/personal";
+  nextUrl = "/profile/professional";
+  
   static propTypes = {
     profile:                        React.PropTypes.object,
     ui:                             React.PropTypes.object,
@@ -134,7 +136,7 @@ class EducationTab extends ProfileFormFields {
   };
 
   saveEducationForm = () => {
-    saveAndContinue.call(this, EducationTab.validation).then(() => {
+    saveProfileStep.call(this).then(() => {
       this.clearEducationEdit();
     });
   };
@@ -288,19 +290,10 @@ class EducationTab extends ProfileFormFields {
         >
           {this.editEducationForm(educationDegreeLevel)}
         </Dialog>
-
         {levelsGrid}
-
         <Cell col={1} />
         <Cell col={10}>
-          <Button
-            raised
-            colored
-            className="profile-save-and-continue"
-            onClick={this.saveAndContinue}
-          >
-            <span>Save and Continue</span>
-          </Button>
+          {this.progressControls()}
         </Cell>
         <Cell col={1} />
       </Grid>
--- a/static/js/components/EmploymentTab.js
+++ b/static/js/components/EmploymentTab.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
-import Button from 'react-mdl/lib/Button';
 
 import EmploymentForm from './EmploymentForm';
 import ProfileFormFields from '../util/ProfileFormFields';
 
 class EmploymentTab extends ProfileFormFields {
+  prevUrl = "/profile/education";
   nextUrl = "/profile/privacy";
 
   render () {
@@ -24,13 +24,7 @@ class EmploymentTab extends ProfileFormFields {
           <Cell col={1}></Cell>
           <Cell col={1}></Cell>
           <Cell col={10}>
-            <Button
-              raised
-              colored
-              className="profile-save-and-continue"
-              onClick={this.saveAndContinue}>
-              <span>Save and Continue</span>
-            </Button>
+            {this.progressControls()}
           </Cell>
           <Cell col={1}></Cell>
         </Grid>
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -1,13 +1,9 @@
 import React from 'react';
-import Button from 'react-mdl/lib/Button';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
 import ProfileFormFields from '../util/ProfileFormFields';
 
 class PersonalTab extends ProfileFormFields {
-  constructor(props) {
-    super(props);
-  }
   nextUrl = "/profile/education";
 
   static propTypes = {
@@ -83,14 +79,7 @@ class PersonalTab extends ProfileFormFields {
         <Cell col={1} />
         <Cell col={1} />
         <Cell col={10}>
-          <Button
-            raised
-            colored
-            className="profile-save-and-continue"
-            onClick={this.saveAndContinue}
-          >
-            <span>Save and Continue</span>
-          </Button>
+          {this.progressControls()}
         </Cell>
         <Cell col={1} />
       </Grid>
--- a/static/js/containers/ProfilePage_test.js
+++ b/static/js/containers/ProfilePage_test.js
@@ -17,11 +17,20 @@ describe("ProfilePage", function() {
   this.timeout(5000);  // eslint-disable-line no-invalid-this
 
   let listenForActions, renderComponent, helper, patchUserProfileStub;
+  let pageUrlStubs = [
+    '/profile/personal',
+    '/profile/education',
+    '/profile/professional',
+    '/profile/privacy'
+  ];
+  let lastPage = pageUrlStubs[pageUrlStubs.length - 1];
+  let prevButtonSelector = '.progress-button.previous';
+  let nextButtonSelector = '.progress-button.next';
+
   beforeEach(() => {
     helper = new IntegrationTestHelper();
     listenForActions = helper.listenForActions.bind(helper);
     renderComponent = helper.renderComponent.bind(helper);
-
     patchUserProfileStub = helper.sandbox.stub(api, 'patchUserProfile');
   });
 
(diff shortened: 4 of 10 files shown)


gsidebo:
@aliceriot @noisecapella @amir-qayyum-khan please review!

gsidebo:
some very rough merge conflicts here. probably from #369. ill try to hammer through them

static/js/util/ProfileTab.js
```diff
@@ -58,14 +60,44 @@ class ProfileTab extends React.Component {
...
   saveAndContinue = () => {
-    let lastTab = this.nextUrl === "/dashboard";
-    saveAndContinue.call(this, this.constructor.prototype.validation, lastTab).then(() => {
+    saveProfileStep.call(this, this.constructor.prototype.validation, !!this.isLastTab).then(() => {
       this.context.router.push(this.nextUrl);
     });
-  }
+  };
+
+  progressControls = (cellWidth=10) => {
```
amir-qayyum-khan: shouldnt it be in profile_edit.js all ui components are there

amir-qayyum-khan:
Are you adding tests for previous button?

static/js/components/EmploymentTab.js
```diff
@@ -12,10 +12,14 @@ import moment from 'moment';
...
 import ProfileTab from "../util/ProfileTab";
-import { saveAndContinue } from '../util/profile_edit';
+import { saveProfileStep } from '../util/profile_edit';
 
 class EmploymentTab extends ProfileTab {
-  nextUrl = "/profile/privacy";
+  constructor(props) {
+    super(props);
+    this.prevUrl = "/profile/education";
+    this.nextUrl = "/profile/privacy";
```
aliceriot: I prefer the instance property pattern to the constructor pattern, I think it's just a little cleaner. Could you switch this to:

```js
class EmploymentTab extends ProfileTab {
  nextUrl = "...";
  prevUrl = "...";
  ...
}
```

gsidebo: i do too, but overriding those in child classes causes an error (or maybe just a warning?)

eg: PrivacyTab sets nextButtonText to "I'm Done!" instead of the default "Save And Continue". if i set that variable as you suggest and override it in PrivacyTab, i get the error/warning

aliceriot: why not have this pattern:

```js
nextButtonText = () => (
  this.isLastTab ? "I'm done!" : "Save and Continue"
);
```

we could define that instance method on `ProfileTab` and then set instance properties on all the child classes.

aliceriot:
I think you're going to have a number of conflicts with https://github.com/mitodl/micromasters/pull/406, let me know if it's confusing (#406 was a big pr).

static/js/util/ProfileTab.js
gsidebo: i think this component is defined in the right place. it is relevant to all components that subclass from `ProfileTab`. profile_edit.js may have most of our current low-level components, but i don't think that's a reason to put _all_ low-level components there. a lot of them make sense in there  because they can be used across many different components (`boundTextField`, `boundStateSelectField`, etc.). this `progressControls` component is only intended to be used by `ProfileTab` children

gsidebo:
@amir-qayyum-khan re: Previous button tests

we currently have "Save and Continue" tests, but they don't test that the URL changes correctly. the only thing that "Previous" changes is the URL. the functionality is very simple, but if we want to add tests, I'm happy to do it

amir-qayyum-khan:
@gsidebo on previous button click you can check if content fully renders gain. i mean no error messages, data last sate is at place etc?

gsidebo:
@aliceriot i'm working on the merge and i'm really confused

you changed 'saveAndContinue' in profile_edit.js to get rid of the validation argument and only accept a boolean `finalPage`, right? this is the function as edited by your PR in profile_edit.js: https://github.com/mitodl/micromasters/pull/406/files#diff-2160cc7afb8b1aa3f724d90e8ba5a01fR446

here's `EducationTab` calling that same function with something that looks like it should be a validation function: https://github.com/mitodl/micromasters/blob/3440571e3013469a74457b1df80af06b9e6c5daa/static/js/components/EducationTab.js#L137

aliceriot:
Yeah, the `saveAndContinue` call in `EducationTab` should have been changed in that PR probably. The validation is moved now into `util/validation.js`, so `EducationTab.validation` is `undefined`.

aliceriot:
I have a pr open that makes similar chnages to the education tab: https://github.com/mitodl/micromasters/pull/445

static/js/components/EmploymentTab.js
gsidebo: yeah i can probably incorporate something like that

gsidebo:
conflicts resolved. still need to look into unit testing the previous button and refactoring nextButtonText

static/js/components/EducationTab.js
```diff
@@ -134,7 +135,7 @@ class EducationTab extends ProfileFormFields {
   };
 
   saveEducationForm = () => {
-    saveAndContinue.call(this, EducationTab.validation).then(() => {
+    saveProfileStep.call(this).then(() => {
```
aliceriot: :100:

gsidebo:
apparently i broke a few js tests too. looking into that as well

gsidebo:
@noisecapella @aliceriot i plan on opening an issue related to this, but i'll mention it here too. debugging tests in ProfilePage_test.js has been very frustrating. the promises in the IntegrationTestHelper class and in redux-asserts seem to swallow all manner of errors and reports them as something like this:

```
  1) ProfilePage it respects the current value (false) when saving on /profile/personal:
     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
```

if, for example, i stick `assert(false)` into the body of the method that handles `renderComponent.then`, i don't get a message saying an assertion failed at line X. i get an error like the one above. i've been unsuccessful trying to add my own error handling to these various promises, and redux-asserts is pretty opaque to me (which may well be due to my lack of redux experience).

gsidebo:
for the time being i'm going to skip this test error. it's really bizarre and i just don't have the tools to figure out what the error actually is. i restored the bad test case to match its equivalent on master exactly (apart from the button selector string) and it's still failing. any input would be _greatly_ appreciated. again, this is on static/js/containers/ProfilePage_test.js and the bad test case is line 75 (`it(`it respects the current value (false) when saving on /profile/${page} ...`)

gsidebo:
@amir-qayyum-khan i added a Previous button unit test. please review

amir-qayyum-khan:
thanks, let me review

amir-qayyum-khan:
Go to education tab ``/profile/education`` press previous twice you will be land on education it self instead of personal tab.

gsidebo:
@amir-qayyum-khan when you press "Previous" twice, i think what is actually happening is your press "Previous" once, then the personal tab loads and your second click is counted as a click on "Save And Continue", which returns you to the education tab. does that make sense?

gsidebo:
@noisecapella @aliceriot opened an issue related to my JS testing difficulties: https://github.com/mitodl/micromasters/issues/448

gsidebo:
finally figured out the test failure

static/js/containers/ProfilePage_test.js
```diff
@@ -17,106 +17,76 @@ describe("ProfilePage", function() {
...
+      START_PROFILE_EDIT,
+      UPDATE_PROFILE_VALIDATION
+    ], () => {
+      TestUtils.Simulate.click(button);
+    });
+  };
+
+  it('navigates backward when Previous button is clicked', () => {
+    let pageIndex = 1;
+    renderComponent(pageUrlStubs[pageIndex]).then(([, div]) => {
```
aliceriot: I think indexing into `pageUrlStubs` here makes it a lot less clear what this test is actually testing. I think it makes sense to use that pattern when you want to write a test that checks that some behavior occurs across all the pages (something like `pageUrlStubs.map( index => { some.test.code(index) });` ) but if we're only testing what occurs on one page I need to look around unnecessarily to figure out which page we're actually on (`pageUrlStubs[pageIndex] == pageUrlStubs[1] == '/profile/education'`).

Can we change these tests on a single page to just use the string literals? (this test and the next one down).

static/js/util/ProfileFormFields.js
```diff
@@ -64,13 +66,41 @@ export default class ProfileFormFields extends React.Component {
...
+        colored
+        className="progress-button next"
+        onClick={this.saveAndContinue}>
+        <span>{this.isLastTab ? "I'm Done!" : "Save and Continue"}</span>
+      </Button>;
+    }
+    return <Cell col={cellWidth}>
+        {prevButton}
+        {nextButton}
+      </Cell>;
```
aliceriot: What do you think about this function just returning something like:

```js
<div>
  {prevButton}
  {nextButton}
</div>
```

and then leaving the responsibility of wrapping the buttons in a `<Cell>` component to the calling context? I think the default argument and so on is a little bit of indirection, when we could just have the information right there where the function is called at render time. If we have this JSX on, for instance, the 'professional' tab, it's a little clearer when reading the code what will actually render on the screen:

```js
<Cell col={10}>
  {this.progressControls()}
</Cell>
```
whereas if we have

```js
{this.progressControls(12)}
```

I have to go read the definiton of `progressControls` to figure out what that argument is going to do.

gsidebo: i thought about that, but didn't go for it initially b/c it's just adding an extraneous html element. looking at it with fresh eyes, i think that's preferable to what is here currently. i think an even cleaner solution might be using an actual react component rather than a simple function. something like...

```javascript
return <Cell {...this.props}>
  {prevButton}
  {nextButton}
</Cell>;

// then in the parent component
<Cell col={2} />
<ProgressControlCell col="10" />
<Cell col={2} />
```

not perfect, but put that forward b/c i've been wondering why we don't do this more in general. seems like a lot of our components are simple function calls. i can't say for sure whether or not we're missing out on important functionality, but i think we definitely lose some semantic cleanliness when some components in a `render` method are appear as `<SomeComponent />` and others appear as `{this.someComponent()}`. 

anyway, i'll change this to your suggestion

aliceriot: I think the main reason we prefer methods (e.g. on `ProfileFormFields`) over separate components for a lot of stuff is that many of those methods need to pull stuff out of `this.props`, and end up being tightly coupled to profile display functionality. In general, I think if we don't see a real need to use a component elsewhere it's a little easier (and less verbose! haha) to write instance helper methods that we call in the `render` method of our component. With a separate component (such as `<ProfileProgressControls>` or something) and the current setup for `nextUrl` and so on we'd need to splat in `this.props` and also assign props for all of the instance variables we'd need, so the call would end up looking like:

```js
<Cell col={2} />
<ProfileProgressControls {...this.props} col={10} nextUrl={this.nextUrl} prevUrl={this.prevUrl} />
<Cell col={2} />
```

IMO in this specific case the helper method (instance method on `ProfileFormFields`) has less boilerplate and ends up being cleaner, but in other cases this might not be true. I think the line between separate component / helper method is a little unclear sometimes, and I had a bit of trouble pulling the work history editing component out of the work history tab (so I could display it on the `/users/` page) partially for that reason. So we should definitely always be ensuring we're being somewhat sane with this stuff - we don't want tons and tons of tiny components, but we also don't want god classes that end up having far too much responsibility and becoming difficult to refactor.

static/js/containers/ProfilePage_test.js
```diff
@@ -17,106 +17,78 @@ describe("ProfilePage", function() {
...
+    renderComponent(secondPage).then(([, div]) => {
+      let button = div.querySelector(prevButtonSelector);
+      assert.equal(helper.currentLocation.pathname, secondPage);
+      TestUtils.Simulate.click(button);
+      assert.equal(helper.currentLocation.pathname, firstPage);
+    });
+  });
 
+  it('marks email_optin and filled_out when saving ${lastPage}', done => {
+    renderComponent(lastPage).then(([, div]) => {
```
gsidebo: @aliceriot happy medium (this line and lines 58-60). readable and keeps the url stub definitions in one place

aliceriot: yeah that's fine. is that a template literal though? it looks like a single quoted string, in which case the `'${lastPage'` substitution won't happen:

subs

static/js/util/ProfileFormFields.js
gsidebo: 💯  understood. thanks for the perspective. i imagine this is something that snags React newbies pretty frequently

static/js/containers/ProfilePage_test.js
gsidebo: oops

aliceriot: :D no worries, the syntax is annoyingly close on the page. backwards compatability...

static/js/components/PersonalTab.js
```diff
@@ -1,13 +1,9 @@
 import React from 'react';
-import Button from 'react-mdl/lib/Button';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
 import ProfileFormFields from '../util/ProfileFormFields';
 
 class PersonalTab extends ProfileFormFields {
-  constructor(props) {
-    super(props);
-  }
```
aliceriot: hooray code deletion!

static/scss/dashboard.scss
```diff
@@ -145,17 +145,23 @@ c.validation-wrapper {
...
   span {
     padding: 10px;
   }
 }
+  &.previous {
+    background: #F9F9F9!important;
+  }
+  &.next {
+    background: #4C8399!important;
+  }
```
aliceriot: the indentation is a little off here - we prefer 2 spaces for `scss`.

aliceriot:
one little comment, and :+1: once that's fixed.

gsidebo:
@aliceriot (continuing [this](https://github.com/mitodl/micromasters/pull/447#discussion-diff-65720190) discussion here since it's now an outdated diff...)

just had a thought. while i was doing assorted debugging for this story, i noticed that the `progressControls` method was being executed 3 times on every page load. i'm pretty sure this was caused by some actions being fired by various components which would cause the parent component (or some other component upstream) to be redrawn. i know that React selectively redraws components via a diffing engine. my question: do we run the risk of unnecessary redrawing if we use helper methods instead of formal React components? example: let's say B is a child component of A, and some action causes a change to A, but B is effectively unchanged. if B is a helper method, it seems to me that it would execute every time A changes. if B is a React component on the other hand, would it not be redrawn due to React's diffing engine?

@noisecapella you can chime in as well if you have any input

aliceriot:
I don't have much experience with trying to optimize for performance in react so I'm not sure. My understanding is that the main way people do that is by setting custom `componentShouldUpdate` methods that reduce the number of unnecessary updates.

currently, we're just leaning on react's virtual DOM without really any optimization - it might be worth looking into easy low hanging fruit, but react is so fast already I'm not sure it's worth putting tons of time into.

gsidebo:
@aliceriot 👍

noisecapella:
Regarding the point on performance, the render method is meant to be called frequently. The virtual DOM does a diff to check if any changes need to be made in the DOM, then the DOM changes are made. We should measure the performance with the browser profiler before doing any optimization here. By the way I profiled the AutoComplete field a little while ago and it looked like the DOM manipulation was by far the largest source of slowdown. For us I think it means we should make sure we don't render too much stuff, but otherwise I don't think we will have a performance problem here.

gsidebo:
@aliceriot @noisecapella sounds good. just wanted to see if this is something we're aware of. this PR is ready for review again

aliceriot:
above comment still applies, there's a little indentation error but besides that :+1:

gsidebo:
sorry, forgot to push

aliceriot:
:+1:

PR #463, opened 2016-06-03
Allowed nulls for education and employment fields

### LINKED ISSUE #462: Another high school validation error
cc @giocalitri 

When we sent `null` as the field of study to the backend it complains. We should update the backend with a database migration to allow nulls for all education and employment fields

noisecapella:
Fixes #462 
#### What's this PR do?
Changes fields to allow null values for education and employment. Since high schools don't require the field_of_study field we need to allow null there, and we should probably allow nulls in the other fields since we are doing this validation on the front end, not the back end.

Add a new high school to your education profile and save it


pdpinch:
I left a question on the issue https://github.com/mitodl/micromasters/issues/462#issuecomment-223645620

> It seems extreme to allow nulls on everything because of this one field. Are you anticipating some other changes?

noisecapella:
I changed the code to only allow nulls on `field_of_study`

aliceriot:
:+1:

PR #466, opened 2016-06-03
Docker with python35

### LINKED ISSUE #458: Upgrade to Python 3.5
It looks like if we switch the Dockerfile to point to `python:3.5` instead of `ubuntu:trusty` everything should just work

giocalitri:
fixes #458
#### What's this PR do?
Switches docker to python 3.5

everything should work as is but using python 3.5: one way to test is to run `docker-compose run web tox`


noisecapella:
:+1:

PR #479, opened 2016-06-06
Fixed two JS console warnings

aliceriot:

none

#### What's this PR do?

I noticed two JS console warnings that were introduced (near as I can tell) in https://github.com/mitodl/micromasters/commit/1dcaf65475ea61ba0c2e8ddc3e3ba932ec2d2af7

if, on master, you visit `/dashboard`, you get these errors in the JS console:

errors

check out the changes.

Make sure the JS warnings are gone, and that there are no regressions.

### DIFF
--- a/static/js/components/CourseList.js
+++ b/static/js/components/CourseList.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import _ from 'lodash';
-import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import RaisedButton from 'material-ui/RaisedButton';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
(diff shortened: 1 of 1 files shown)


static/js/components/CourseList.js
```diff
@@ -119,7 +120,7 @@ class CourseList extends React.Component {
             <br/>
             <p>You need to pass {totalCourses - coursesPassed} more courses
               before you can apply for the <strong>{program.title}</strong> Master’s Degree.</p>
-            <MuiThemeProvider>
```
noisecapella: I don't believe we need this line at all... this is done in `dashboard_routes.js`. Can you see if it still works if we remove it?

aliceriot: ah, indeed we don't! if I remove it things look the same as well (I believe this change was introduced to fix a styling issue).

without:
without

with:
with

noisecapella:
:+1:

PR #491, opened 2016-06-06
Fixed handling of multiple validation errors for education and work history

### LINKED ISSUE #490: Only one validation error is shown in education form at a time
To reproduce, open a new education and click Save

noisecapella:
Fixes #490 

#### What's this PR do?
Fixes bug in `checkFieldPresence` to combine errors correctly. Currently it uses `Object.assign` which won't combine nested values correctly

Open a new education, then click save without typing anything. All fields should be marked as needing validation

### DIFF
--- a/static/js/util/validation.js
+++ b/static/js/util/validation.js
@@ -10,12 +10,15 @@ let handleNestedValidation = (profile, keys, nestedKey) => {
 };
 
 let checkFieldPresence = (profile, requiredFields, messages) => {
-  return Object.assign({}, ...requiredFields.map( keySet => {
+  let errors = {};
+
+  for (let keySet of requiredFields) {
     let val = _.get(profile, keySet);
     if (_.isUndefined(val) || _.isNull(val) || val === "" ) {
-      return _.set({}, keySet,  `${messages[keySet.slice(-1)[0]]} is required`);
+      _.set(errors, keySet,  `${messages[keySet.slice(-1)[0]]} is required`);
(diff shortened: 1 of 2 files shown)


static/js/util/validation.js
```diff
@@ -10,12 +10,15 @@ let handleNestedValidation = (profile, keys, nestedKey) => {
...
+
+  for (let keySet of requiredFields) {
     let val = _.get(profile, keySet);
     if (_.isUndefined(val) || _.isNull(val) || val === "" ) {
-      return _.set({}, keySet,  `${messages[keySet.slice(-1)[0]]} is required`);
+      _.set(errors, keySet,  `${messages[keySet.slice(-1)[0]]} is required`);
     }
-  }));
+  }
+  return errors;
```
aliceriot: Ah I see, before I had written this:

```js
Object.assign({}, ...requiredFields.map (keySet => { ... });
```

where the arrow function returns something that should look like this:

```js
{'work_history': [
  {'fieldName': 'Field Name is required'}
]}
```

so we're splatting a bunch of these objects together, but whichever one ends up being last sets the `work_history` key on the object we end up returning to just the one error. bleh, this is what happens when you get too fancy...

aliceriot:
looks good, all the validation errors show up.

:+1:

static/js/util/validation.js
noisecapella: Yes, exactly

Issue #472, opened 2016-06-05
Mailing list form on home page should behave the same as form on current site


pdpinch:
- [x] default state: comments field and Institution name field are hidden
- [x] when the user selects "I represent an interested corporation" or "I represent an interested university or educational institution" the "institution name" field is displayed
- [x] when the user selects "Other" the "comments" field is displayed
- [x] when the user selects "I am an interested learner" the form reverts to its default state

pdpinch:
@amir-qayyum-khan I thought this was going to be part of #287. Is this a bug, or did you just leave it out?

amir-qayyum-khan:
ops, that was not assigned to me, i havent seen this at all. By the way is is already fixed in #287.

amir-qayyum-khan:
Fixed via #287.

pdpinch:
This isn't displaying properly for me on micromasters-ci or micromasters-rc:

<img width="1364" alt="screen shot 2016-06-06 at 11 24 46 am" src="https://cloud.githubusercontent.com/assets/430126/15827506/64702104-2bd9-11e6-8c5b-9dc1cf6b57d6.png">

In the default state, and when individual learner is selected, the institution and comment fields should not display.

amir-qayyum-khan:
@pdpinch hello,

The issue is in deployment of https://micromasters-rc.herokuapp.com,
- you cannot able to open my files from source of home page for example https://micromasters-rc.herokuapp.com/static/bundles/js/home_page/custom.js
- But you can open the public.js file which is existing deployed. https://micromasters-rc.herokuapp.com/static/bundles/public.js

You can check you self by right clicking the home page and the opn view as source link

<img width="1280" alt="screen shot 2016-06-07 at 4 43 38 pm" src="https://cloud.githubusercontent.com/assets/10431250/15856476/0d0bc426-2ccf-11e6-8a70-7d6ce323ca92.png">

##### Note: on my local docker machine it is working as expected. the issue is with location of static/assets of micromasters-rc.

pdpinch:
@bdero @blarghmatey what's going on here with the static asset build (on both micromasters-ci and micromasters-rc)

bdero:
I'm looking into this.

bdero:
Nevermind, Tobias is working on this.

pdpinch:
This was fixed by #493

PR #306, opened 2016-05-11
Updated home page to match current mock-ups

amir-qayyum-khan:
#### What are the relevant tickets?
Fixes https://github.com/mitodl/micromasters/issues/287

#### What's this PR do?
- Larger program cards as in mockups.
- Added placeholder cards for right column
- Removed star ratings from card.
- Refactor cards to use bootstrap

@pdpinch @noisecapella @aliceriot 

#### Where should the reviewer start?
- cms home page.

#### How should this be manually tested?
- home page.
- mockups https://projects.invisionapp.com/share/2R78JXBUG#/screens/142338433_Public_Home

#### Screenshots (if appropriate)
screen shot 2016-05-31 at 6 00 52 pm
screen shot 2016-05-31 at 6 01 09 pm

PR #493, opened 2016-06-07
Fixing path for JS assets on Mailchimp form

blarghmatey:

The javascript files used for the Mailchimp signup form are not located
in the bundles directory. This commit fixes the paths being generated
for the template context in which they are loaded.

Addresses #472


cms/models.py
```diff
@@ -46,9 +46,9 @@ def get_context(self, request):
         context["programs"] = Program.objects.filter(live=True)
         context["style_src"] = get_bundle_url(request, "style.js")
         context["public_src"] = get_bundle_url(request, "public.js")
-        context["custom_src"] = get_bundle_url(request, "js/home_page/custom.js")
-        context["ajaxchimp_src"] = get_bundle_url(request, "js/home_page/jquery.ajaxchimp.min.js")
-        context["wow_src"] = get_bundle_url(request, "js/home_page/wow.min.js")
+        context["custom_src"] = static("js/home_page/custom.js")
+        context["ajaxchimp_src"] = static("js/home_page/jquery.ajaxchimp.min.js")
+        context["wow_src"] = static("js/home_page/wow.min.js")
```
aliceriot: so we aren't bundling this JS? `custom.js` isn't minified like the mailchimp and `wow.min.js` code, are we fine with that?

noisecapella:
It seems like we get our javascript from three places now: A CDN for bootstrap, our webpack bundle for the dashboard, and the mailchimp assets we host locally. I'm fine with this since the scope of the bootstrap and mailchimp code is limited, just making sure everyone else knows about it.

Checking functionality...

noisecapella:
Oh nevermind I'll leave this to @aliceriot

cms/models.py
noisecapella: It looks like `custom.js` contains `__webpack_public_path__` as if it were going to be another webpack bundle. I'm not sure what the intent was in the original commit

blarghmatey: I didn't look too closely at custom.js but it is a pretty small file so I think we can get away with just leaving it as is. I defer to the judgment of @aliceriot and @noisecapella for final determination.

aliceriot:
I may be doing something incorrect, but I get this when I try to visit `/`:

static

blarghmatey:
Most likely my fault. Looking now.

cms/models.py
aliceriot: I could go either way I suppose, on the one hand we have already a webpack entry point for the homepage (`/static/js/public`) so it could be included there, but it's a very small file so maybe it doesn't matter.

noisecapella: custom.js was added in PR #306, copied from here: https://github.com/mitodl/mga/blob/gh-pages/js/custom.js

I think we should roll `custom.js` into the public bundle and leave the other two files as static assets. I'm not sure if the mailchimp code would be in npm to import, and in any case we don't manage jquery with npm, it's included in `base.html` from a CDN

noisecapella: Actually we could just require the specific javascript files like we do for the other assets in `public.js`

aliceriot: I agree, and I think if we can get the mailchimp code from npm and avoid checking in the dependency we should.

noisecapella:
@aliceriot I made a commit to this PR to move the JS into the bundle, can you take another look?

static/js/public.js
```diff
@@ -13,6 +13,10 @@ require("./public/site.js");
 // jquery components
 require("./public/components/raty.js");
 
+// mailchimp requirements
+require("ajaxchimp");
+require('imports?this=>window!wowjs');
```
aliceriot: does this automatically set a `window.wowjs` variable or something similar?

noisecapella: Yes, I couldn't get it to do something like `import WOW from '...'`

aliceriot: cool, I haven't used the `imports` loader before.

noisecapella: I haven't either, to be honest I just copied this from https://github.com/matthieua/WOW/issues/233#issuecomment-173640648

aliceriot:
Functionality seems good, so it's a :+1: from me.

noisecapella:
Ok I'll hand it back over to @blarghmatey

PR #495, opened 2016-06-07
Fixed bug: No enroll button if edx_course_key is not defined

### LINKED ISSUE #384: Bug: Don't attempt to display enroll button when edx_course_key is null
We allow courseruns to be created without an edx_course_key when there are plans to create one later. 

Unfortunately, this means we have to protect against situations when a course key is required, like providing an enroll button. 

In normal usage, this is unlikely to happen because we wouldn't pass the enrollment_start date without having a course code. But in our testing this came up.

annagav:
Fix #384

### DIFF
--- a/static/js/util/courseList.js
+++ b/static/js/util/courseList.js
@@ -76,8 +76,6 @@ export function makeCourseStatusDisplay(course, now = moment()) {
     }
   }
   case STATUS_OFFERED_NOT_ENROLLED: {
-    let courseInfoUrl = `${SETTINGS.edx_base_url}/courses/${firstRun.course_id}/about`;
-
     if (!firstRun.enrollment_start_date) {
       return firstRun.fuzzy_enrollment_start_date;
     }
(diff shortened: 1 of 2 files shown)


noisecapella:
Can you add a test for the change you made?

static/js/util/courseList.js
```diff
@@ -85,8 +85,11 @@ export function makeCourseStatusDisplay(course, now = moment()) {
     let enrollmentDate = moment(firstRun.enrollment_start_date);
     if (enrollmentDate.isAfter(now, 'day')) {
       return `Enrollment starting: ${enrollmentDate.format("M/D/Y")}`;
-    } else {
+    } else if(firstRun.course_id) {
+      let courseInfoUrl = `${SETTINGS.edx_base_url}/courses/${firstRun.course_id}/about`;
```
noisecapella: It looks like this is already defined above

annagav:
Fixed the tests!

noisecapella:
Code looks good, testing functionality

noisecapella:
Functionality looks good :+1: 

Can you rename the PR and commit message after you squash to be in the past tense?

annagav:
Is this commit message better?

noisecapella:
Can you change `btn` to `button`, and also change the PR title?

PR #496, opened 2016-06-07
Added confirmation dialog when deleting education and work entries

### LINKED ISSUE #439: Confirmation when user deletes employment / education
We should have a confirmation dialog when users try to delete an education/employment entry.

We can use the standard material-ui dialog, with the message "Are you sure you want to delete this?" with the buttons, "Cancel" and "Confirm"

aliceriot:

closes #439 

#### What's this PR do?

This adds a confirmation dialog that pops-up when clicking the 'delete'
icon for a work or education history entry. This entails some associated
reducers and actions, and a new class called `ConfirmDeletion` which
will, hopefully, be generally useful for this sort of thing.

Most of the key stuff ends up being implemented as instance methods on
`ProfileFormFields` and `ProfileFormContainer`.

Make sure that any deletion action causes a confirmation dialog to appear, and make sure that the buttons ('yes' , 'no') do what they should.

This should work on `/users` and on `/profile`.

delete_education

### DIFF
--- a/static/js/actions/ui.js
+++ b/static/js/actions/ui.js
@@ -65,3 +65,18 @@ export const SET_USER_PAGE_DIALOG_VISIBILITY = 'SET_USER_PAGE_DIALOG_VISIBILITY'
 export const setUserPageDialogVisibility = bool => (
   { type: SET_USER_PAGE_DIALOG_VISIBILITY, payload: bool }
 );
+
+export const SET_SHOW_EDUCATION_DELETE_DIALOG = 'SET_SHOW_EDUCATION_DELETE_DIALOG';
+export const setShowEducationDeleteDialog = bool => (
+  { type: SET_SHOW_EDUCATION_DELETE_DIALOG, payload: bool }
+);
+
+export const SET_SHOW_WORK_DELETE_DIALOG = 'SET_SHOW_WORK_DELETE_DIALOG';
+export const setShowWorkDeleteDialog = bool => (
+  { type: SET_SHOW_WORK_DELETE_DIALOG, payload: bool }
+);
+
+export const SET_DELETION_INDEX = 'SET_DELETION_INDEX';
+export const setDeletionIndex = index => (
+  { type: SET_DELETION_INDEX, payload: index }
+);
--- a/static/js/components/ConfirmDeletion.js
+++ b/static/js/components/ConfirmDeletion.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import Dialog from 'material-ui/Dialog';
+import Button from 'react-mdl/lib/Button';
+
+export default class ConfirmDeletion extends React.Component {
+  static propTypes = {
+    close:        React.PropTypes.func,
+    deleteEntry:  React.PropTypes.func,
+    open:         React.PropTypes.bool,
+  };
+
+  deleteAndClose = () => {
+    const { close, deleteEntry } = this.props;
+    deleteEntry();
+    close();
+  };
+
+  render () {
+    const { close, open } = this.props;
+    let actions = [
+      <Button
+        type='button'
+        key='close'
+        className="cancel-button"
+        onClick={close}>
+        No
+      </Button>,
+      <Button
+        key='delete'
+        type='button'
+        className="delete-button"
+        onClick={this.deleteAndClose}>
+        Yes
+      </Button>
+    ];
+    return (
+      <Dialog
+        className={"deletion-confirmation"}
+        open={open}
+        onRequestClose={close}
+        actions={actions}
+        autoScrollBodyContent={true}
+      >
+        Delete this entry?
+      </Dialog>
+    );
+  }
+}
--- a/static/js/containers/UserPage_test.js
+++ b/static/js/containers/UserPage_test.js
@@ -20,6 +20,9 @@ import {
   SET_EDUCATION_DIALOG_INDEX,
   SET_EDUCATION_DIALOG_VISIBILITY,
   SET_USER_PAGE_DIALOG_VISIBILITY,
+  SET_DELETION_INDEX,
+  SET_SHOW_WORK_DELETE_DIALOG,
+  SET_SHOW_EDUCATION_DELETE_DIALOG,
 } from '../actions/ui';
 import IntegrationTestHelper from '../util/integration_test_helper';
 import * as api from '../util/api';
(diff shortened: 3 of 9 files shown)


static/js/components/ConfirmDeletion.js
```diff
@@ -0,0 +1,42 @@
...
+        key='close'
+        className="cancel-button"
+        onClick={close}>
+        No
+      </Button>,
+      <Button
+        key='delete'
+        type='button'
+        className="save-button"
+        onClick={deleteEntry}>
```
noisecapella: Can you change `deleteEntry` to not close the dialog itself? This handler could then be changed to call `deleteEntry` then `close` afterwards

static/js/components/ConfirmDeletion.js
```diff
@@ -0,0 +1,42 @@
...
+        type='button'
+        key='close'
+        className="cancel-button"
+        onClick={close}>
+        No
+      </Button>,
+      <Button
+        key='delete'
+        type='button'
+        className="save-button"
```
noisecapella: Can you rename to `delete-button`?

static/js/containers/UserPage_test.js
```diff
@@ -41,11 +44,23 @@ describe("UserPage", () => {
...
+
   afterEach(() => {
     helper.cleanup();
   });
 
   describe("Education History", () => {
+    let deleteButton = div => {
+      return div.getElementsByClassName('profile-tab-card')[1].
+        getElementsByClassName('profile-row-icons')[0].
+        getElementsByClassName('mdl-button')[1];
```
noisecapella: Can you use the CSS class for the delete button here, like you do for the cancel button below?

static/js/components/ConfirmDeletion.js
aliceriot: Sorry, I'm not totally sure what you mean. Is this what you're suggesting?

- refactor the callback passed in to `ConfirmDeletion` as `deleteEntry` to just perform the deletion and not close the dialog
- refactor this class (`ConfirmDeletion`) to call `deleteEntry().then(() => close())`

static/js/containers/UserPage_test.js
```diff
@@ -41,11 +44,23 @@ describe("UserPage", () => {
 
   let userActions = [RECEIVE_GET_USER_PROFILE_SUCCESS, REQUEST_GET_USER_PROFILE];
 
+  let openDialog = () => {
+    return [...document.getElementsByClassName('deletion-confirmation')].find(dialog => (
+      dialog.style["left"] === "0px"
```
noisecapella: What does this do? Is there more than one dialog being shown at a time?

static/js/components/ConfirmDeletion.js
noisecapella: Yes, although `deleteEntry` doesn't have to return a promise

static/js/containers/UserPage_test.js
aliceriot: the material-ui dialog isn't an actual `<dialog>` element. basically, when it isn't being displayed (when the `open` prop is set to false) it sets `left = -10000px`, and when it is displayed it changes that to `0px`. this let's us find the one which is currently being 'shown' so we can get buttons and so on inside of it.

static/js/components/ConfirmDeletion.js
aliceriot: kk that makes sense.

static/js/actions/ui.js
```diff
@@ -65,3 +65,18 @@ export const SET_USER_PAGE_DIALOG_VISIBILITY = 'SET_USER_PAGE_DIALOG_VISIBILITY'
 export const setUserPageDialogVisibility = bool => (
   { type: SET_USER_PAGE_DIALOG_VISIBILITY, payload: bool }
 );
+
+export const SET_SHOW_EDUCATION_DELETE_DIALOG = 'SET_SHOW_EDUCATION_DELETE_DIALOG';
```
noisecapella: Do we need both this and `SET_SHOW_WORK_DELETE_DIALOG`? It looks like the only difference between the two is the delete function which is passed into the component.

static/js/containers/UserPage_test.js
noisecapella: Right, that makes sense

static/js/actions/ui.js
aliceriot: I think we need both for the `/users` page, we have two different popups there which each have different callbacks bound to the action buttons. I initially had just `SET_SHOW_DELETE_CONFIRM_DIALOG`, but on the `/users` page that causes a race condition between the popups for education and work history (we'll delete the entry at `DELETION_INDEX` for whichever one ends up being on top).

Does that make sense?

aliceriot:
@noisecapella updated!

static/js/actions/ui.js
noisecapella: Does the race condition happen when both popups are shown at the same time?

aliceriot: Yeah, and because they occupy the same screen space one will end up on top. Without the separate actions it was possible to have a situation where clicking the delete button on a work history entry and then confirming when the dialog appeared would cause one of the education entries to be deleted (on `/users`).

aliceriot: This is because the sequence looked something like this:

- user clicks 'delete' icon on a work history entry
- dispatch `SET_DELETION_INDEX`
- dispatch `SET_SHOW_CONFIRM_DELETE_DIALOG`
- education and work history popups both display (they both have the `open` prop set to the value of `state.showConfirmDeleteDialog`)
- education dialog ends up on top
- user clicks 'Yes' in the confirm dialog
- the `onClick` handler bound to the 'Yes' button (in the education dialog) is run, which does something like this:
    - get `state.deletionIndex`
    - `_.cloneDeep(profile)`
    - `profile.education.splice(deletionIndex, 1)`

then there's some cleanup after that. So basically without two actions we don't have enough information to tell whether a user is trying to delete a work history entry or an education entry, and if both pop-ups display we'll try to delete an entry from whichever happens to be on top.

so I think we need to either do what I've got here or something more complicated that places one `ConfirmDeletion` component on `/users`, dispatches an action to differentiate education and work history, and binds another set of callbacks when  a delete icon is clicked.

noisecapella: Yes, I see what you mean

aliceriot: I got super confused when I ran into it!

noisecapella:
Code looks good, checking functionality

noisecapella:
:+1:

PR #499, opened 2016-06-07
Removed python3-dev which points to a python 3.4 package

noisecapella:
Related to #458 

#### What's this PR do?
When we installed python3-dev it pulled in python 3.4 dependencies. In the `python:3.5` image these are already installed so this PR gets rid of this step. Note that python2.7 is also installed but it's required by `lsb-release` which we probably don't want to get rid of

The build should succeed

### DIFF
--- a/apt.txt
+++ b/apt.txt
@@ -2,7 +2,6 @@
 git
 curl
 libpq-dev
-python3-dev
 libjpeg-dev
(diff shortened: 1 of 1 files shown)


aliceriot:
build finished correctly, successfully ran all tests! :+1:

PR #500, opened 2016-06-07
Added Roboto font everywhere

### LINKED ISSUE #434: change font to Roboto
We should be using Roboto as the default font for the site. In some places it appears to be specified in the CCS but, in Firefox at least, Helvetica is actually being used. 

When we list degrees and employment in the profile form, the labels should be Roboto 16px bold and the dates should be Roboto 16px Light.

See https://projects.invisionapp.com/d/main#/console/6633865/153562860/comments/63129265 or https://invis.io/AF6XG4X62#/153562860_Step_Three_-_Professional_Info

aliceriot:

closes #434 

#### What's this PR do?

this changes the font to [Roboto](https://www.google.com/fonts/specimen/Roboto) everywhere.

Review the css changes. Unfortunately I think we need the `!important` tag to override some styling set by the material design light components :cry: :disappointed: 

I guess just make sure everything looks ok?

### DIFF
--- a/static/scss/dashboard.scss
+++ b/static/scss/dashboard.scss
@@ -1,7 +1,3 @@
-body {
-  font-family: "Open Sans" !important;
-}
-
 c.validation-wrapper {
   padding-bottom: 10px;
 }
@@ -139,7 +135,6 @@ c.validation-wrapper {
   border-bottom: 1px dashed #CCCCCC;
   .mdl-cell {
     margin: 20px auto;
-    font-family: "Open Sans";
     text-align: center;
     color: #414141;
     font-weight: 400;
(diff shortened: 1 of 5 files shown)


pdpinch:
Dumb question: is the base.html template used by both the django app and the react app?

aliceriot:
Yes, `base.html` is used everywhere. Basically, when we're in react-land we drop one `div` with `id='dashboard'` on the page, and we use `react-dom` to render our whole React app into that div.

noisecapella:
It looks like there are some uses of `Open Sans` in homepage.scss, can you remove those too?

aliceriot:
@noisecapella updated!

noisecapella:
Looks good :+1:

Issue #433, opened 2016-05-27
Profile: if education or work history is missing, and the relevant switch is turn on, we should cause a validation error


noisecapella:
The validation error message should be displayed where the collapsed degree would have gone.

PR #501, opened 2016-06-07
Refactored profile classes

noisecapella:
Part of #433 

#### What's this PR do?
The `Tab` classes don't need to inherit from `ProfileFormFields` since they have `Form` components which inherit from `ProfileFormFields` instead. Also, the progress controls were moved into `ProfileProgressControls` for better separation of concerns

Everything should work exactly as before. The reviewer should make sure edits still work in both profile pages

### DIFF
--- a/static/js/components/PrivacyTab.js
+++ b/static/js/components/PrivacyTab.js
@@ -1,13 +1,10 @@
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
+import ProfileProgressControls from './ProfileProgressControls';
 import ProfileFormFields from '../util/ProfileFormFields';
 
 class PrivacyTab extends ProfileFormFields {
-  prevUrl = "/profile/professional";
-  nextUrl = "/dashboard";
-  isLastTab = true;
-
   static propTypes = {
     profile:        React.PropTypes.object,
     saveProfile:    React.PropTypes.func,
@@ -15,6 +12,7 @@ class PrivacyTab extends ProfileFormFields {
   };
 
   render() {
+    const { saveProfile, profile } = this.props;
     return (
       <div>
         <Grid className="profile-splash">
@@ -29,7 +27,13 @@ class PrivacyTab extends ProfileFormFields {
             { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) } <br />
           </Cell>
           <Cell col={12}>
-            {this.progressControls()}
+            <ProfileProgressControls
+              prevUrl="/profile/professional"
+              nextUrl="/dashboard"
+              isLastTab={true}
+              saveProfile={saveProfile}
+              profile={profile}
+            />
           </Cell>
         </Grid>
       </div>
--- a/static/js/components/ProfileProgressControls.js
+++ b/static/js/components/ProfileProgressControls.js
@@ -1,0 +1,57 @@
+import React from 'react';
+import Button from 'react-mdl/lib/Button';
+
+import { saveProfileStep } from '../util/profile_edit';
+
+export default class ProfileProgressControls extends React.Component {
+  static propTypes = {
+    nextUrl: React.PropTypes.string,
+    prevUrl: React.PropTypes.string,
+    isLastTab: React.PropTypes.bool,
+    saveProfile: React.PropTypes.func.isRequired,
+    profile: React.PropTypes.object.isRequired
+  };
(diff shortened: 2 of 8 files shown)


static/js/components/ProfileProgressControls.js
```diff
@@ -0,0 +1,53 @@
...
+import Button from 'react-mdl/lib/Button';
+
+import { saveProfileStep } from '../util/profile_edit';
+
+export default class ProfileProgressControls extends React.Component {
+  static propTypes = {
+    nextUrl: React.PropTypes.string,
+    prevUrl: React.PropTypes.string,
+    isLastTab: React.PropTypes.bool,
+    saveProfile: React.PropTypes.func.isRequired,
```
gsidebo: is `saveProfile` actually being used anywhere? i can only see it being passed down to `ProfileProgressControls` as a prop, where it doesn't seem to be used

noisecapella: It's used by `saveProfileStep` which pulls `profile` and `saveProfile` from `this.props`

gsidebo: right, thanks for the reminder. there's just something very opaque about all this. finding out where `saveProfile` comes from and where it's used involves a bunch of different files. `saveProfileStep` itself relies on a side effect of sorts to work properly - it depends on `saveProfile` and `profile` to be available on `this.props`. maybe that is an understood pattern in our codebase and/or the React community itself. still, i would think an API like this would make so much more sense: `function saveProfileStep(profile, isLastStep=false)`. in the body of that function, you'd call `profile.save()`. this would involve creating a new class that encapsulates profile data and functionality. i'm not sure how heavy a lift that would be, but i think it's definitely preferable to the current approach.

anyway, that could be a future refactor. for now, this PR definitely improves what we have

static/js/components/ProfileProgressControls.js
```diff
@@ -0,0 +1,57 @@
...
+    return <div>
+      {prevButton}
+      {nextButton}
+    </div>;
+  }
+
+  static contextTypes = {
+    router: React.PropTypes.object.isRequired
+  };
+}
```
gsidebo: we do EOF newlines right?

static/js/components/PrivacyTab.js
```diff
@@ -1,20 +1,18 @@
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
+import ProfileProgressControls from './ProfileProgressControls';
 import ProfileFormFields from '../util/ProfileFormFields';
 
 class PrivacyTab extends ProfileFormFields {
```
gsidebo: maybe not in scope for this PR, but a future refactor might be to simplify `ProfileFormFields` or get rid of it entirely. it's doing 2 things: (1) defining constant values for _all_ tabs, which it shouldn't be doing IMO, and (2) binding input elements to `this`. if we want a one-stop shop to make all form element types available to these tabs, then it would make some sense to keep `ProfileFormFields` around and to set it as some property in the tab class, eg:

```javascript
class PrivacyTab extends React.Component {
  constructor() {
    this.formFields = ProfileFormFields()
    //...
  }
  
  render() {
    return <div>{this.formFields.boundTextField()}</div>
  }
```

either way, we should really move the tab-specific values to their own tab classes. as for this specific tab class, it doesn't make semantic sense that `PrivacyTab` inherits from `ProfileFormFields`, and even less so now that all of the other tabs aren't subclassing it

gsidebo:
👍  outstanding comments are just suggestions for future refactors. this is good to merge

static/js/components/PrivacyTab.js
noisecapella: I agree, we should favor composition over inheritance

static/js/components/ProfileProgressControls.js
noisecapella: Yes, I pushed a change for this

static/js/components/ProfileProgressControls.js
noisecapella: I agree that the use of `this.props` is kind of opaque and maybe we should send our arguments more explicitly. I don't think `profile.save()` would be clean or easy to implement however. To make changes to the state we need to dispatch actions, then we need to wait for React to send new props to have an update. It's not like a Django model where `save()` is synchronous and the model is updated right after the call.

PR #505, opened 2016-06-08
Removed singleTest command

noisecapella:
#### What's this PR do?
Removes `singleTest` script from package.json and moves the test command into a separate script `js_test.sh`. This allows the user to run `npm test` to run all JS tests or `npm test ./static/js/...` for a single file.

It looks like parameter substitutions like `$1` don't work in `package.json`. It ignores them and appends all parameters to the end of the line instead, so we can't do the fancy bash default parameters inside `package.json`, it has to go in a separate file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,8 +64,7 @@
   "scripts": {
     "postinstall": "./webpack_if_prod.sh",
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",
-    "test": "node ./node_modules/mocha/bin/_mocha --compilers js:babel-register --require static/js/global_init.js 'static/**/*/*_test.js'",
-    "singleTest": "node ./node_modules/mocha/bin/_mocha --compilers js:babel-register --require static/js/global_init.js $1",
+    "test": "./js_test.sh",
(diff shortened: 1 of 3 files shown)


gsidebo:
can you update the readme too? i think there's a note about this at the bottom

package.json
```diff
@@ -64,8 +64,7 @@
   "scripts": {
     "postinstall": "./webpack_if_prod.sh",
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",
-    "test": "node ./node_modules/mocha/bin/_mocha --compilers js:babel-register --require static/js/global_init.js 'static/**/*/*_test.js'",
-    "singleTest": "node ./node_modules/mocha/bin/_mocha --compilers js:babel-register --require static/js/global_init.js $1",
+    "test": "./js_test.sh",
```
gsidebo: can you not use this nifty syntax `${1:-'static/**/*/*_test.js'}` directly in the command definition in package.json? it's not so bad to have these sh files around but i feel like we should avoid it if possible

noisecapella: In the background context above I mentioned that `$1` doesn't work in `package.json`, it doesn't replace with anything. npm just appends all arguments to the end of the line

gsidebo:
👍

PR #506, opened 2016-06-08
Added check to only show edit buttons for a user's own profile

### LINKED ISSUE #415: Profile viewer: remove add, delete, or edit icons if profile is not user's
This is the frontend part of #387

This should be a simple comparison between the current user's id (could be null) and the username in the URL, but we do need tests.

aliceriot:

closes #415 

#### What's this PR do?

This adds a little utility function called `userPrivilegeCheck`, which
basically just checks that the username on the profile in question is
the same as `SETTINGS.username`. We pass a callback in to
`userPrivilegeCheck`, and `userPrivilegeCheck` returns the value of
executing that callback if the usernames match, else it returns an
optional second argument or, if that isn't provided, it returns
undefined.

Make another profile on your dev box. Check two things:

- you should see edit icons on your own profile, and be able to edit normally
- you shouldn't see any edit icons on another user's profile

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -11,7 +11,7 @@ import _ from 'lodash';
 import moment from 'moment';
 
 import ProfileFormFields from '../util/ProfileFormFields';
-import { generateNewEducation } from "../util/util";
+import { generateNewEducation, userPrivilegeCheck } from "../util/util";
 import { saveProfileStep } from '../util/profile_edit';
 import { HIGH_SCHOOL } from '../constants';
 import ConfirmDeletion from './ConfirmDeletion';
@@ -66,7 +66,11 @@ class EducationForm extends ProfileFormFields {
   };
 
   renderEducationLevel(level){
-    const { ui: { educationDegreeInclusions }, profile: { education } } = this.props;
+    const {
+      ui: { educationDegreeInclusions },
+      profile: { education },
+      profile
+    } = this.props;
     if (educationDegreeInclusions[level.value]) {
       let rows = [];
       if (education !== undefined) {
@@ -74,22 +78,24 @@ class EducationForm extends ProfileFormFields {
           filter(([, education]) => education.degree_name === level.value).
           map(([index, education]) => this.educationRow(education, index));
       }
-      rows.push(
-        <FABButton
-          colored
-          onClick={() => this.openNewEducationForm(level.value, null)}
-          className="profile-add-button"
-          key="I'm unique!"
-        >
-          <Icon name="add" />
-        </FABButton>
-      );
+      userPrivilegeCheck(profile, () => {
+        rows.push(
+          <FABButton
+            colored
+            onClick={() => this.openNewEducationForm(level.value, null)}
+            className="profile-add-button"
+            key="I'm unique!"
+          >
+            <Icon name="add" />
+          </FABButton>
+        );
+      });
       return rows;
     }
   }
 
   educationRow = (education, index) => {
-    const { errors } = this.props;
+    const { errors, profile } = this.props;
     if (!('id' in education)) {
       // don't show new educations, wait until we saved on the server before showing them
       return;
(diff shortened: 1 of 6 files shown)


noisecapella:
I feel like `userPrivilegeCheck` would be cleaner if it just took the `profile` as its argument and returned a boolean of the check against `SETTINGS.username`. Passing `profile` as an argument would be cleaner because it doesn't rely on `this.props.profile` existing or any particular `this` being bound. In general I feel like we should only bind `this` if there is necessarily a tight coupling between the component and the function being called.

If you wanted to keep the promises it would be cleaner to make a separate function, something like `ifThen(conditionValue, thenFunc, elseFunc)` which could reused and tested more easily.

aliceriot:
Yeah, I see what you mean. I wanted to have a single function that I could just wrap certain things with which would take care of all the details, instead of having to check the return value of `userPrivilegeCheck` (if it returned a boolean) every time - to me this would lead to extra boilerplate code around it. 

Maybe that's not so important though. I agree that we should just be taking `profile` as an argument, so I refactored to do that.

noisecapella:
Can you add unit test(s) for `userPrivilegeCheck`?

aliceriot:
Sure! That's a good idea.

static/js/components/EducationForm.js
```diff
@@ -103,19 +108,26 @@ class EducationForm extends ProfileFormFields {
       }
     };
     let dateFormat = date => moment(date).format("MM[/]YYYY");
-
+    let icons = () => {
+      return userPrivilegeCheck(profile,
```
noisecapella: Can you remove the explicit return here?

static/js/components/EducationForm.js
```diff
@@ -66,30 +66,35 @@ class EducationForm extends ProfileFormFields {
...
-        <FABButton
-          colored
-          onClick={() => this.openNewEducationForm(level.value, null)}
-          className="profile-add-button"
-          key="I'm unique!"
-        >
-          <Icon name="add" />
-        </FABButton>
-      );
+      userPrivilegeCheck(profile, () => (
```
noisecapella: Can you add braces here? Without the function braces it looks like we are using the return value of `rows.push` for something

static/js/components/EmploymentForm.js
```diff
@@ -102,23 +102,24 @@ class EmploymentForm extends ProfileFormFields {
...
-      workHistoryRows.push(
-        <FABButton
-          colored
-          onClick={this.addWorkHistoryEntry}
-          key="I'm unique!"
-          className="profile-add-button">
-          <Icon name="add" />
-        </FABButton>
-      );
+      userPrivilegeCheck(profile, () => (
```
noisecapella: Can you add the function braces here too?

static/js/containers/UserPage_test.js
```diff
@@ -50,247 +38,288 @@ describe("UserPage", () => {
...
-        done();
-      });
+  describe("Unauthenticated user page", () => {
+    beforeEach(() => {
+      helper = new IntegrationTestHelper();
+      listenForActions = helper.listenForActions.bind(helper);
+      renderComponent = helper.renderComponent.bind(helper);    
+      patchUserProfileStub = helper.sandbox.stub(api, 'patchUserProfile');
+      helper.profileGetStub.returns(
+        Promise.resolve(Object.assign({}, USER_PROFILE_RESPONSE))
```
noisecapella: Do we need to clone `USER_PROFILE_RESPONSE` here?

noisecapella: Oh, this is copied from before

static/js/util/util.js
```diff
@@ -16,6 +17,13 @@ export function sendGoogleAnalyticsEvent(category, action, label, value) {
   ga.event(event);
 }
 
+export function userPrivilegeCheck (profile, privileged, unPrivileged) {
+  if ( profile.username === SETTINGS.username ) {
+    return privileged();
```
noisecapella: Can handle this case like the `unPrivileged` case below?

static/js/containers/UserPage_test.js
aliceriot: yeah, the only difference is that we're not assigning a username into the profile.

static/js/containers/UserPage_test.js
```diff
@@ -50,247 +38,288 @@ describe("UserPage", () => {
...
-      renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
-        let title = div.getElementsByClassName('profile-card-title')[1];
-        assert.equal(title.textContent, 'High school');
-        done();
-      });
+      patchUserProfileStub = helper.sandbox.stub(api, 'patchUserProfile');
+
+      helper.profileGetStub.returns(
+        Promise.resolve(Object.assign({}, USER_PROFILE_RESPONSE, {
+          username: SETTINGS.username
```
noisecapella: This will update the username in the profile but we will still be using whatever username was in the url to store it in `profiles` object. Can you change `helper.profileGetStub` here and in the unauthenticated version to use `withArgs` to make sure we have the correct username in the URL too?

noisecapella:
Functionality looks great!

noisecapella:
:+1:

PR #508, opened 2016-06-08
Used level of attainment from edX to set default switch values

### LINKED ISSUE #507: Use level_of_education to specify which switches are on by default
It looks like we overlooked this when we first implemented the education tab.

We should also make sure that switches are on if there is employment or education items so they are not obscured

noisecapella:
Fixes #507 

#### What's this PR do?
Populates switches with the level of attainment from edX. If none is found all switches are on by default.

static/js/util/util.js

Log in, go to `/profile`, navigate to the education, then verify that the switches are:
 - always on if there is one or more education items in the card, regardless of anything else, or
 - all on if no edx_level_of_education is set in the profile, or
 - on only if the level of education is less or equal to the card's level of education

You should be able to change the level in the profile model and refresh to see the different states.

### DIFF
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -5,6 +5,8 @@ import ga from 'react-ga';
 import striptags from 'striptags';
 import _ from 'lodash';
 
+import { EDUCATION_LEVELS } from '../constants';
+
 export function sendGoogleAnalyticsEvent(category, action, label, value) {
   let event = {
     category: category,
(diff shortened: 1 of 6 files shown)


static/js/util/util.js
```diff
@@ -221,3 +223,23 @@ export function makeProfileImageUrl(profile) {
...
+export function calculateDegreeInclusions(profile) {
+  let found = false;
+  let inclusions = {};
+  for (const { value } of EDUCATION_LEVELS) {
+    inclusions[value] = !found;
+    if (value === profile.edx_level_of_education) {
+      // every level after this point is higher than the user has attained according to edx
+      found = true;
+    }
+  }
```
aliceriot: ok so here we're looping through `'HIGH_SCHOOL'`, `'ASSOCIATE'`, etc and setting `inclusions[value] = true`, until we encounter a `value` which is equal to the level of education from edx - then we set `found = true`, and we'll set the rest of the `values` from `EDUCATION_LEVELS` to false.

I get what this code does but it took me 5 minutes to get there - here's a suggestion for how to refactor it so it reads more clearly:

```js
let highestLevelFound = false;
let inclusions = {};
for (const { value } of EDUCATION_LEVELS) {
  inclusions[value] = highestLevelFound ? false : true;
  if (value === profile.edx_level_of_education) {
    // every level after this point is higher than the user has attained according to edx
    highestLevelFound = true;
  }
}
```

again, just a suggestion about how to make it clearer, use your judgement about what's best.

aliceriot:
functionality looks good, one question about the code. back to you!

noisecapella:
Addressed the comment

aliceriot:
LGTM :+1:

PR #512, opened 2016-06-09
Added background_image to ProgramsPage

### LINKED ISSUE #510: Allow hero image to be changed in programs template
As a course staff, I need to be able to customize my program page. In particular, I need to be able to change the large image at the top.

screenshot-micromasters-ci herokuapp com 2016-06-09 11-54-27-2

annagav:
Fix #510 
#### What's this PR do?
Add background image to ProgramsPage.

- login {url}/cms/
- create a program page, or use existing one
- Upload an image
- In the background_image field of the program page select that image
- save and publish the page


cms/templates/cms/program_page.html
```diff
@@ -11,7 +12,8 @@
   <div class="app-media layout-boxed">
     <div class="main-content-wrapper">
       <section class="main-content">
-        <div class="banner-wrapper banner-sub banner-program-1">
+      {% image page.background_image fill-1310x613 as tmp_photo %}
```
gsidebo: 1. just curious: what is this `image` function and `fill-1310x613` argument accomplishing that can't be done with CSS?
1. can you rename `tmp_photo` to something more descriptive? i think simply renaming it `background_image` would work

annagav: 1. The `image` tag  takes two arguments `{% image [image] [resize-rule] %}`, otherwise would not work.
2. I will rename the variable to `background_image`.

gsidebo: i see. i'm guessing we're doing that b/c a user would be uploading their own image, which we need to apply our own sizing rules to? makes sense, but i feel like it would make more sense to do this when the image is uploaded. maybe something for a future story

annagav: Yes, the user uploads the image, and we are using `wagtailimages_tags` to insert it into the template.

gsidebo:
👍

PR #517, opened 2016-06-09
Added error handling in the dashboard.

### LINKED ISSUE #234: handle http errors in dashboard UI

giocalitri:
closes #234

#### What's this PR do?
Introduces error handling in the dashboard UI

`static/js/util/api.js`

The easiest way is to go to the dashboard, stop edx platform and reload the dashboard.
If you want to test the error code and error message, in `dashboard/views.py` change the get function to:

```
    def get(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-self-use
        """
        Returns information needed to display the user dashboard for a program.
        """
        return Response(
            status=500,
            data={
                  'error_code': 'AB123',
                  'user_message': 'foo message'
            }
        )
```

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -73,10 +73,9 @@ export const saveProfile = (username, profile) => {
     dispatch(requestPatchUserProfile(username));
     return api.patchUserProfile(username, profile).
       then(newProfile => dispatch(receivePatchUserProfileSuccess(username, newProfile))).
-      catch(e => {
+      catch(() => {
         dispatch(receivePatchUserProfileFailure(username));
-        // propagate exception
-        return Promise.reject(e);
+        // the exception is assumed handled and will not be propagated
       });
   };
 };
@@ -90,10 +89,9 @@ export function fetchUserProfile(username) {
     dispatch(requestGetUserProfile(username));
     return api.getUserProfile(username).
       then(json => dispatch(receiveGetUserProfileSuccess(username, json))).
-      catch(e => {
+      catch(() => {
         dispatch(receiveGetUserProfileFailure(username));
-        // propagate exception
-        return Promise.reject(e);
+        // the exception is assumed handled and will not be propagated
       });
   };
 }
@@ -109,7 +107,10 @@ export const receiveDashboardSuccess = programs => ({
   type: RECEIVE_DASHBOARD_SUCCESS,
   payload: { programs }
 });
-const receiveDashboardFailure = () => ({ type: RECEIVE_DASHBOARD_FAILURE });
+export const receiveDashboardFailure = errorInfo => ({
+  type: RECEIVE_DASHBOARD_FAILURE,
+  payload: { errorInfo }
+});
 export const clearDashboard = () => ({ type: CLEAR_DASHBOARD });
 
 export function fetchDashboard() {
@@ -117,10 +118,9 @@ export function fetchDashboard() {
     dispatch(requestDashboard());
     return api.getDashboard().
       then(dashboard => dispatch(receiveDashboardSuccess(dashboard))).
-      catch(e => {
-        dispatch(receiveDashboardFailure());
-        // propagate exception
-        return Promise.reject(e);
+      catch(error => {
+        dispatch(receiveDashboardFailure(error));
+        // the exception is assumed handled and will not be propagated
       });
   };
 }
--- a/static/js/components/ErrorMessage.js
+++ b/static/js/components/ErrorMessage.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import Alert from 'react-bootstrap/lib/Alert';
+
+
+export default class ErrorMessage extends React.Component {
+  static propTypes = {
+    errorInfo:   React.PropTypes.object.isRequired,
+  };
+
+  render() {
+    const { errorInfo: {error_code: errorCode, user_message: userMessage} } = this.props;
+
+    let errorCodeStr = '';
+    let userMessageStr = '';
+    if (errorCode !== undefined) {
+      errorCodeStr = `${errorCode}: `;
+    }
+    if (userMessage !== undefined) {
+      userMessageStr = `Additional info: ${userMessage}`;
+    }
+
+    return (
+      <div className="alert-message">
+        <Alert bsStyle="danger">
+          <p>{ errorCodeStr }Sorry, we were unable to load the data necessary to
+          process your request. Please reload the page.</p>
+          <p>{ userMessageStr }</p>
+          <p>
+            If the error persists, please contact <a href="mailto:mitx-support@mit.edu">
+            mitx-support@mit.edu</a> specifying
+            this entire error message.
+          </p>
+        </Alert>
+      </div>
+    );
+  }
+}
--- a/static/js/containers/App_test.js
+++ b/static/js/containers/App_test.js
@@ -8,18 +8,25 @@ import _ from 'lodash';
 import {
   CLEAR_DASHBOARD,
   CLEAR_PROFILE,
+  RECEIVE_DASHBOARD_FAILURE,
+  RECEIVE_GET_USER_PROFILE_SUCCESS,
   START_PROFILE_EDIT,
   UPDATE_PROFILE_VALIDATION,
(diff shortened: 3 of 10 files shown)


static/js/util/api.js
```diff
@@ -73,17 +73,31 @@ export function fetchJSONWithCSRF(input, init, loginOnError) {
...
     } else {
       return "";
     }
+  }).catch(([text, statusCode]) => {
+    let respJson = {};
+    if (text.length !== 0) {
+      try {
+        respJson = JSON.parse(text);
+      }
+      catch(e) {}
```
noisecapella: We should handle this case if you're not planning on this already

static/js/components/ErrorMessage.js
```diff
@@ -0,0 +1,54 @@
...
+  static propTypes = {
+    errorInfo:   React.PropTypes.object.isRequired,
+  };
+
+  state = {
+    alertVisible: true
+  };
+
+  render() {
+    if (!this.state.alertVisible) {
```
noisecapella: In general we should refrain from using component state since it's more easily maintained in the redux store. `ErrorMessage` should take an additional `close` function in the `props` which should take care of making the error message go away.

Also, we should assume that `ErrorMessage` is not being shown if there is no error. For example here is a `render()` method:

    let errorMessage;
    if (hasError) {
      errorMessage = <ErrorMessage errorInfo={errorInfo} ... />;
    }

    <Page>
      {errorMessage}
    </Page>

`ErrorMessage` is only shown if there is something to show, so inside this class we don't need to check if it is visible or if the error message exists, for example.

static/js/components/ErrorMessage.js
```diff
@@ -0,0 +1,54 @@
...
+    const {errorInfo} = this.props;
+    if (errorInfo.errorStatusCode === undefined) {
+      return null;
+    }
+    let errorCode, userMessage = '';
+    if (errorInfo.error_code !== undefined) {
+      errorCode = errorInfo.error_code + ': ';
+    }
+    if (errorInfo.user_message !== undefined) {
+      userMessage = 'Additional info: ' + errorInfo.user_message;
```
noisecapella: We are using template literals instead of string concatenation, so this would be `` `Additional info: ${errorInfo.user_message}` ``

static/js/components/ErrorMessage.js
```diff
@@ -0,0 +1,54 @@
...
+  state = {
+    alertVisible: true
+  };
+
+  render() {
+    if (!this.state.alertVisible) {
+      return null;
+    }
+
+    const {errorInfo} = this.props;
```
noisecapella: You can use destructuring to pull out pieces of `errorInfo`:

    const { errorInfo: { errorStatusCode } } = this.props;

Will produce `errorStatusCode`, possibly with `undefined` if the key doesn't exist.

static/js/components/ErrorMessage.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import LoginButton from '../containers/LoginButton';
+import { Alert } from 'react-bootstrap';
```
noisecapella: It's generally better to import from the specific file so we don't necessarily include all of `react-bootstrap` for a single component:

    import Alert from 'react-bootstrap/lib/Alert';

static/js/containers/DashboardPage.js
```diff
@@ -42,9 +46,20 @@ const mapStateToProps = (state) => {
...
+  }
+
+  let errorInfoJson = {};
+  if (state.dashboard.errorInfo !== undefined) {
+    errorInfoJson = state.dashboard.errorInfo;
+  }
   return {
     profile: profile,
-    dashboard: state.dashboard,
+    dashboard: dashboardJson,
```
noisecapella: Is `dashboardJson` doing anything useful here?

static/js/actions/index.js
```diff
@@ -124,18 +124,21 @@ export const receiveDashboardSuccess = programs => ({
...
 export const clearDashboard = () => ({ type: CLEAR_DASHBOARD });
 
 export function fetchDashboard() {
   return dispatch => {
     dispatch(requestDashboard());
     return api.getDashboard().
       then(dashboard => dispatch(receiveDashboardSuccess(dashboard))).
-      catch(e => {
-        dispatch(receiveDashboardFailure());
+      catch(errorJson => {
```
noisecapella: nit: JSON is the string representation of the object, not the object itself. Can we rename this to just `error` instead?

static/js/components/ErrorMessage.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default class ErrorMessage extends React.Component {
+  static propTypes = {
+    errorInfo:   React.PropTypes.object.isRequired,
+  };
+
+  render() {
+    const { errorInfo: {error_code: errorCode, user_message: userMessage} } = this.props;
+
+    let errorCodeStr, userMessageStr = '';
```
noisecapella: FYI only `userMessageStr` is being set to a string here, the other is `undefined`. Both will be treated the same way by React AFAIK but we should choose one or the other here

noisecapella:
Looks good so far!

giocalitri:
@noisecapella I think I am done with your comments: waiting for the new ones :-)

noisecapella:
Can you rebase?

giocalitri:
done

static/js/containers/App_test.js
```diff
@@ -43,6 +54,65 @@ describe('App', () => {
     });
   });
 
+  describe('dashboard errors', () => {
+    let errorString = `Sorry, we were unable to load the data necessary
+      to process your request. Please reload the page.`;
+    errorString = errorString.replace(/\s\s+/g, ' ');
+
+    it('error from the backend triggers error message in dashboard', done => {
```
noisecapella: We just merged a PR which changes how these tests are written. I found that the test runner mocha natively handles tests with promises so instead of using `done` we can just return the promise in this function. For each test below can you remove `done` from the arguments, the `done()` call and use `return renderComponent(...` instead?

static/js/containers/App_test.js
```diff
@@ -43,6 +54,65 @@ describe('App', () => {
...
+      let response = _.cloneDeep(DASHBOARD_RESPONSE_ERROR);
```
noisecapella: We shouldn't need to clone this here or below since we shouldn't be modifying this value in our actions or API functions

static/js/util/integration_test_helper.js
```diff
@@ -45,26 +45,35 @@ class IntegrationTestHelper {
...
-      div = document.createElement("div");
-      component = ReactDOM.render(
-        makeDashboardRoutes(this.browserHistory, this.store, () => null),
-        div
-      );
-    }).then(() => {
-      return Promise.resolve([component, div]);
+      let component, div;
+
+      this.listenForActions(expectedTypes, () => {
```
noisecapella: This function was refactored in a recently merged PR, can you update this to preserve the refactoring except for the `isSuccessExpected` parts?

static/js/components/ErrorMessage.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    return (
+      <div className="alert-message">
+        <Alert bsStyle="danger">
+          <p>{ errorCodeStr }Sorry, we were unable to load the data necessary to
+          process your request. Please reload the page.</p>
+          <p>{ userMessageStr }</p>
+          <p>
+            If the error persists, please contact <a href="mitx-support@mit.edu">
```
noisecapella: We need to start the link with `mailto:` here

noisecapella:
Functionality looks great. The `Your Status` header still appears with the error message, can you change the code to output only the error message if there is one?

giocalitri:
done with the comments

noisecapella:
:+1: 

Can you rename the commit message and PR to be in the past tense?

Issue #485, opened 2016-06-06
user settings


pdpinch:
Users should be able to change their settings from the user drop-down.

The settings page will currently just have the privacy setting.

amir-qayyum-khan:
@pdpinch  is their UI mocks for this task. Should we reuse profile/privacy tab here?

pdpinch:
There's no mock-up for this task. We should use the same UI as /profile/privacy

pdpinch:
@amir-qayyum-khan thanks for the PR. You should have copied the screenshots here for discussion:

#### Screenshots 

<img width="364" alt="screen shot 2016-06-10 at 6 39 08 pm" src="https://cloud.githubusercontent.com/assets/10431250/15966097/015fccb2-2f3b-11e6-9c38-63f77cd17413.png">
<img width="785" alt="screen shot 2016-06-10 at 6 41 30 pm" src="https://cloud.githubusercontent.com/assets/10431250/15966096/015eb872-2f3b-11e6-9529-17d7dcef7984.png">

pdpinch:
repeating my https://github.com/mitodl/micromasters/pull/519#issuecomment-225273413 on the PR:

- [ ] remove text "We care about your privacy"
- [ ] remove dotted line and move the form up 
- [ ] The button label should be "Save" instead of "I'm Done!"

amir-qayyum-khan:
fixed via https://github.com/mitodl/micromasters/pull/519

pdpinch:
You can open another issue to fix it

PR #519, opened 2016-06-10
Added settings page

amir-qayyum-khan:
resolves https://github.com/mitodl/micromasters/issues/485

#### What's this PR do?
Added settings page to change privacy settings

- see SettingPage.js

- click menu icon on top right of user page. From their select settings.
 @pdpinch @noisecapella @aliceriot 

screen shot 2016-06-10 at 6 39 08 pm
screen shot 2016-06-10 at 6 41 30 pm

### DIFF
--- a/static/js/components/PrivacyForm.js
+++ b/static/js/components/PrivacyForm.js
@@ -1,0 +1,26 @@
+// @flow
+/* global SETTINGS: false */
+import React from 'react';
+
+import ProfileFormFields from '../util/ProfileFormFields';
+import type { Profile } from '../flow/profileTypes';
+import type { UIState } from '../reducers/ui';
+
+class PrivacyForm extends ProfileFormFields {
+  props: {
+    profile:        Profile,
+    ui:             UIState,
+    updateProfile:  Function,
+  };
+
+  render() {
+    return (
+      <div>
+        <h4 className="privacy-form-heading">Who can see your profile?</h4>
+        { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) }
+      </div>
+    );
+  }
+}
+
+export default PrivacyForm;
--- a/static/js/components/PrivacyTab.js
+++ b/static/js/components/PrivacyTab.js
@@ -2,6 +2,7 @@
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
+import PrivacyForm from './PrivacyForm';
 import ProfileProgressControls from './ProfileProgressControls';
 import ProfileFormFields from '../util/ProfileFormFields';
 import {
@@ -31,10 +32,9 @@ class PrivacyTab extends ProfileFormFields {
             We care about your privacy.
           </Cell>
         </Grid>
-        <Grid className="profile-tab-grid">
+        <Grid className="profile-tab-grid privacy-form">
           <Cell col={12}>
-            <h4>Who can see your profile?</h4>
-            { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) } <br />
+            <PrivacyForm {...this.props} />
           </Cell>
           <Cell col={12}>
             <ProfileProgressControls
--- a/static/js/containers/ProfileFormContainer.js
+++ b/static/js/containers/ProfileFormContainer.js
@@ -49,16 +49,16 @@ class ProfileFormContainer extends React.Component {
       profiles: state.profiles,
       ui: state.ui,
     };
-  }
+  };
 
-  fetchProfile: Function = (): void => {
-    const { dispatch, profiles, params: { username } } = this.props;
+  fetchProfile: Function = (username: string): void => {
+    const { dispatch, profiles } = this.props;
     if (profiles[username] === undefined || profiles[username].getStatus === undefined) {
       dispatch(fetchUserProfile(username));
     }
-  };
+  }
 
-  updateProfile(isEdit: boolean, profile: Profile) {
+  updateProfile: Function = (isEdit: boolean, profile: Profile): void => {
     const { dispatch } = this.props;
     const username = SETTINGS.username;
 
--- a/static/js/containers/SettingsPage.js
+++ b/static/js/containers/SettingsPage.js
@@ -1,0 +1,96 @@
+// @flow
+/* global SETTINGS: false */
+import React from 'react';
+import { connect } from 'react-redux';
+import Loader from 'react-loader';
+import Grid, { Cell } from 'react-mdl/lib/Grid';
+
+import Jumbotron from '../components/Jumbotron';
+import {
+  startProfileEdit,
+  FETCH_PROCESSING,
+} from '../actions/index';
+import ProfileFormContainer from './ProfileFormContainer';
+import PrivacyForm from '../components/PrivacyForm';
+import ProfileProgressControls from '../components/ProfileProgressControls';
+import {
+    combineValidators,
+    privacyValidation,
+} from '../util/validation';
+import { getPreferredName } from '../util/util';
+import type { Profile } from '../flow/profileTypes';
+import type { UIState } from '../reducers/ui';
+
+class SettingsPage extends ProfileFormContainer {
+  props: {
+    profiles: {[key: string]: Profile},
+    dispatch: Function,
+    ui:       UIState,
+  };
+
+  componentWillMount() {
+    this.startSettingsEdit();
+  }
+
+  startSettingsEdit() {
+    const { dispatch } = this.props;
+    dispatch(startProfileEdit(SETTINGS.username));
+  }
+
+  render() {
+    const { profiles, ui } = this.props;
+    let profile;
+    let errors;
+    let isEdit = true;
+    let loaded = false;
+    let username = SETTINGS.username;
+    let preferredName = SETTINGS.name;
+    
+    if (profiles[username] !== undefined) {
+      let profileFromStore = profiles[username];
+      loaded = profileFromStore.getStatus !== FETCH_PROCESSING;
+      if (profileFromStore.edit !== undefined) {
+        errors = profileFromStore.edit.errors;
+        profile = profileFromStore.edit.profile;
+      } else {
+        profile = profileFromStore.profile;
+        errors = {};
+        isEdit = false;
+      }
+      preferredName = getPreferredName(profile);
+    }
+
+    return (
+      <Loader loaded={loaded}>
+        <Jumbotron profile={profile} text={preferredName}>
+          <div className="card-copy">
+            <Grid className="profile-tab-grid privacy-form">
+              <Cell col={12}>
+                <PrivacyForm
+                  {...this.props}
+                  errors={errors}
+                  profile={profile}
+                  updateProfile={this.updateProfile.bind(this, isEdit)}
+                />
+              </Cell>
+              <Cell col={12}>
+                <ProfileProgressControls
+                  nextUrl="/dashboard"
+                  isLastTab={true}
+                  saveProfile={this.saveProfile.bind(this, isEdit)}
+                  profile={profile}
+                  ui={ui}
+                  validator={
+                    combineValidators(privacyValidation)
+                  }
+                />
+              </Cell>
+            </Grid>
+          </div>
+        </Jumbotron>
+      </Loader>
+    );
+  }
+}
+
+export default connect(ProfileFormContainer.mapStateToProps)(SettingsPage);
--- a/static/js/dashboard_routes.js
+++ b/static/js/dashboard_routes.js
@@ -8,6 +8,7 @@ import getMuiTheme from 'material-ui/styles/getMuiTheme';
 
 import App from './containers/App';
 import DashboardPage from './containers/DashboardPage';
+import SettingsPage from './containers/SettingsPage';
 import ProfilePage from './containers/ProfilePage';
 import PersonalTab from './components/PersonalTab';
 import EmploymentTab from './components/EmploymentTab';
@@ -40,6 +41,7 @@ export function makeDashboardRoutes(browserHistory: Object, store: Object, onRou
               <Route path="privacy" component={PrivacyTab} />
             </Route>
             <Route path="/terms_of_service" component={TermsOfServicePage} />
+            <Route path="/settings" component={SettingsPage}  />
             <Route path="/users" component={UserPage} >
               <IndexRedirect to={`${SETTINGS.username}`} />
               <Route path=":username" component={User} />
--- a/ui/url_utils.py
+++ b/ui/url_utils.py
@@ -6,3 +6,4 @@ DASHBOARD_URL = '/dashboard/'
 PROFILE_URL = '/profile/'
 TERMS_OF_SERVICE_URL = '/terms_of_service/'
 USERS_URL = '/users/'
+SETTINGS_URL = "/settings/"
--- a/static/js/components/ProfileProgressControls.js
+++ b/static/js/components/ProfileProgressControls.js
@@ -10,7 +10,7 @@ export default class ProfileProgressControls extends React.Component {
     prevUrl: React.PropTypes.string,
     isLastTab: React.PropTypes.bool,
     saveProfile: React.PropTypes.func.isRequired,
-    profile: React.PropTypes.object.isRequired,
+    profile: React.PropTypes.object,
     ui: React.PropTypes.object.isRequired,
     validator: React.PropTypes.func.isRequired
   };
--- a/static/js/containers/LoginButton.js
+++ b/static/js/containers/LoginButton.js
@@ -30,6 +30,11 @@ class LoginButton extends React.Component {
           title={title}
           bsStyle="danger"
           id="logout-button">
+          <LinkContainer to={{ pathname: '/settings' }} active={false}>
+            <MenuItem>
+              Settings
+            </MenuItem>
+          </LinkContainer>
           <MenuItem
             href="/logout"
             eventKey="logout">
--- a/static/js/containers/SettingsPage_test.js
+++ b/static/js/containers/SettingsPage_test.js
@@ -1,0 +1,96 @@
+/* global SETTINGS: false */
+import '../global_init';
+import TestUtils from 'react-addons-test-utils';
+import { assert } from 'chai';
+
+import {
+  START_PROFILE_EDIT,
+  UPDATE_PROFILE_VALIDATION,
+  REQUEST_PATCH_USER_PROFILE,
+  RECEIVE_PATCH_USER_PROFILE_SUCCESS,
+  CLEAR_PROFILE_EDIT,
+
+  receiveGetUserProfileSuccess
+} from '../actions';
+import IntegrationTestHelper from '../util/integration_test_helper';
+import * as api from '../util/api';
+import { USER_PROFILE_RESPONSE } from '../constants';
+
+describe("SettingsPage", function() {
+  this.timeout(5000);
+  let nextButtonSelector = '.progress-button.next';
+  let listenForActions, renderComponent, helper, patchUserProfileStub;
+  let userActions = [START_PROFILE_EDIT];
+
+  describe ("Settings page tests", () => {
+    beforeEach(() => {
+      helper = new IntegrationTestHelper();
+      listenForActions = helper.listenForActions.bind(helper);
+      renderComponent = helper.renderComponent.bind(helper);
+      patchUserProfileStub = helper.sandbox.stub(api, 'patchUserProfile');
+
+      helper.profileGetStub.
+        withArgs(SETTINGS.username).
+        returns(
+        Promise.resolve(Object.assign({}, USER_PROFILE_RESPONSE, {
+          username: SETTINGS.username
+        }))
+      );
+    });
+
+    afterEach(() => {
+      helper.cleanup();
+    });
+
+    let confirmSaveButtonBehavior = (updatedProfile, pageElements, validationFailure=false) => {
+      let { div, button } = pageElements;
+      button = button || div.querySelector(nextButtonSelector);
+      patchUserProfileStub.throws("Invalid arguments");
+      patchUserProfileStub.withArgs(SETTINGS.username, updatedProfile).returns(Promise.resolve(updatedProfile));
+
+      let actions = [];
+      if (!validationFailure) {
+        actions.push(
+          REQUEST_PATCH_USER_PROFILE,
+          RECEIVE_PATCH_USER_PROFILE_SUCCESS,
+          START_PROFILE_EDIT,
+          CLEAR_PROFILE_EDIT
+        );
+      }
+      actions.push(
+        UPDATE_PROFILE_VALIDATION
+      );
+      return listenForActions(actions, () => {
+        TestUtils.Simulate.click(button);
+      });
+    };
+
+    it('shows the privacy form', () => {
+      return renderComponent("/settings", userActions).then(([, div]) => {
+        let question = div.getElementsByClassName('privacy-form-heading')[0];
+        assert.equal(question.textContent, 'Who can see your profile?');
+      });
+    });
+
+    describe('save privacy form', () => {
+      it('save privacy changes', () => {
+        return renderComponent("/settings", userActions).then(([, div]) => {
+          let button = div.querySelector(nextButtonSelector);
+          let receivedProfile = Object.assign({}, USER_PROFILE_RESPONSE, {
+            account_privacy: 'public'
+          });
+
+          helper.store.dispatch(receiveGetUserProfileSuccess(SETTINGS.username, receivedProfile));
+
+          assert(button.innerHTML.includes("I'm Done!"));
+          let updatedProfile = Object.assign({}, receivedProfile, {
+            email_optin: true,
+            filled_out: true
+          });
+
+          return confirmSaveButtonBehavior(updatedProfile, {button: button});
+        });
+      });
+    });
+  });
+});
--- a/static/js/containers/UserPage.js
+++ b/static/js/containers/UserPage.js
@@ -9,11 +9,13 @@ import ProfileFormContainer from './ProfileFormContainer';
 
 class UserPage extends ProfileFormContainer {
   componentDidMount() {
-    this.fetchProfile();
+    const { params: { username } } = this.props;
+    this.fetchProfile(username);
   }
 
   componentDidUpdate() {
-    this.fetchProfile();
+    const { params: { username } } = this.props;
+    this.fetchProfile(username);
   }
 
   componentWillUnmount() {
--- a/ui/urls.py
+++ b/ui/urls.py
@@ -8,6 +8,7 @@ from ui.url_utils import (
     PROFILE_URL,
     TERMS_OF_SERVICE_URL,
     USERS_URL,
+    SETTINGS_URL,
 )
 from ui.views import (
     dashboard,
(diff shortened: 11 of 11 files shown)


pdpinch:
@amir-qayyum-khan you can remove the intro text, "We care about your privacy" and the dotted line from this. I think that will make the form more compact.

amir-qayyum-khan:
sure, let me

aliceriot:
I think we want that text (and the dotted line) on the privacy step of the profile fill-out process - maybe a way to do this would be to pull out the actual form into `PrivacyForm` and then `PrivacyTab` can render that text (w/ dotted line) and on the settings page we can just plop another `PrivacyForm` on the page. This is pretty much what I did to pull out the personal and work history forms from the tab components so we could re-use them on the `/users` page.

pdpinch:
Thanks Alice. That sounds like a good solution.

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  console.log(state);
+  return {
+    userProfile: profile,
+    ui: state.ui,
+    isSettingPage: true,
+    errors: errors,
+    updateProfile: state.updateProfile,
+    saveProfile: state.saveProfile
```
aliceriot: Are you sure this is a property of the state? I don't believe we can store functions in the redux store.

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,100 @@
...
+    userProfile:    React.PropTypes.object.isRequired,
+    ui:             React.PropTypes.object.isRequired,
+    errors:         React.PropTypes.object,
+    isSettingPage:  React.PropTypes.bool
+  };
+
+  constructor(props) {
+    super(props);
+    this.updateProfile = this.updateProfile.bind(this);
+    this.saveProfile = this.saveProfile.bind(this);
```
aliceriot: instead of declaring these methods below and binding them to `this` in the constructor you can just use the class property pattern instead:

```js
class SettingsPage extends React.Component {
  updateProfile = profile => {
    ...
  };
}
```

if you declare a class property as an arrow function `this` will already be bound to the instance at run-time (it will do what you expect, basically).

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  render() {
+    const { userProfile: { profile }, ui, errors, isSettingPage } = this.props;
+    let preferredName = profile.preferredName || SETTINGS.name;
```
aliceriot: we have a `getPreferredName` function in `util/util.js` which you could use here instead.

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,100 @@
...
+
+    return (
+      <Jumbotron profile={ profile } text={ preferredName }>
+        <div className="card-copy">
+          <PrivacyForm {...this.props} />
```
aliceriot: here you aren't passing `saveProfile` and `updateProfile` in to privacy form as props, you'll need to do something like this instead:

```js
<PrivacyForm
  {...this.props}
  saveProfile={this.saveProfile}
  updateProfile={this.updateProfile}
/>
```

aliceriot:
You may have an easier time of this if you make `PrivacyPage` subclass the `ProfileFormContainer` component - then you won't have to rewrite the functions to save and update the profile, and you won't need to fiddle with `connect` and all the store business (`ProfileFormContainer` should already have everything you need as instance methods, `this.updateProfile` and so on).

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,81 @@
...
+  static propTypes = {
+    profile:        React.PropTypes.object,
+    saveProfile:    React.PropTypes.func,
+    updateProfile:  React.PropTypes.func,
+    ui:             React.PropTypes.object
+  };
+
+  render() {
+    const { saveProfile, profile, isSettingPage, ui, updateProfile } = this.props;
+    let controlls = (
```
aliceriot: `s/controlls/controls/`

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,81 @@
...
+          validator={
+            combineValidators(
+              personalValidation,
+              educationValidation,
+              educationUiValidation,
+              employmentValidation,
+              employmentUiValidation,
+              privacyValidation
+            )
+          }
```
aliceriot: it might be a little more manageable to deal with the `<ProfileProgressControls>` if you pulled the things we need to be different out into little helper methods, which you could then just call inline in the props we're passing to the `<ProfileProgressControls>` component.
```js
const { isSettingsPage } = this.props;
let prevUrl = () => (
  isSettingsPage ? undefined : "/profile/professional"
);

let validator = () => (

File: static/js/components/PrivacyTab.js [hunk 7]
```diff
@@ -1,61 +1,27 @@
...
-  privacyValidation,
-} from '../util/validation';
+import PrivacyForm from './PrivacyForm';
 
-class PrivacyTab extends ProfileFormFields {
+
+class PrivacyTab extends React.Component {
   static propTypes = {
     profile:        React.PropTypes.object,
+    isSettingPage:  React.PropTypes.bool,
```
aliceriot: why does `PrivacyTab` need the `isSettingsPage` prop?

static/js/dashboard_routes.js
```diff
@@ -39,6 +40,7 @@ export function makeDashboardRoutes(browserHistory, store, onRouteUpdate) {
               <Route path="privacy" component={PrivacyTab} />
             </Route>
             <Route path="/terms_of_service" component={TermsOfServicePage} />
+            <Route path="/setting" component={SettingsPage} />
```
aliceriot: is the url supposed to be `/setting` or `/settings`? @pdpinch ?

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,77 @@
...
+
+  constructor(props) {
+    super(props);
+    const { dispatch } = this.props;
+    const username = SETTINGS.username;
+    dispatch(startProfileEdit(username));
+  }
+
+  render() {
+    const { userProfile: { profile }, ui, errors, isSettingPage } = this.props;
```
aliceriot: is `isSettingPage` passed to this component as a prop? what do we need it for here?

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,77 @@
...
+    userProfile:    React.PropTypes.object.isRequired,
+    ui:             React.PropTypes.object.isRequired,
+    errors:         React.PropTypes.object
+  };
+
+  constructor(props) {
+    super(props);
+    const { dispatch } = this.props;
+    const username = SETTINGS.username;
+    dispatch(startProfileEdit(username));
```
aliceriot: I think it might be better to do this in a `componentWillMount` callback instead of a constructor

amir-qayyum-khan:
@aliceriot how this line know the children attached to this view https://github.com/mitodl/micromasters/blob/master/static/js/containers/UserPage.js#L36

aliceriot:
I'm not sure what you mean. The `UserPage` component has children because it's declared to be the component to render at a react-router `<Route>` which has sub-routes - take a look in `dashboard_routes.js` for details.

You shouldn't need to fiddle with that though. The methods you need for the settings page (`saveProfile`, `updateProfile`, etc) should all be available at `this.saveProfile` on the `PrivacyPage` if you make `PrivacyPage` subclass `ProfileFormContainer`.

aliceriot:
You will need to connect `PrivacyPage` to the store, but you can use the `ProfileFormContainer.mapStateToProps` function to cut down on boilerplate.

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,61 @@
...
+  static propTypes = {
+    profile:        React.PropTypes.object,
+    saveProfile:    React.PropTypes.func,
+    updateProfile:  React.PropTypes.func,
+    ui:             React.PropTypes.object
+  };
+
+  render() {
+    const { saveProfile, profile, ui } = this.props;
+    let showPreBtn = (this.props.location.pathname === "/settings/privacy") ? false : true;
```
aliceriot: why not just

```js
let showPreBtn = this.props.location.pathname !== "/settings/privacy";
```

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,81 @@
...
+      preferredName = profile.preferredName || SETTINGS.name;
+    }
+
+    let childrenWithProps = this.childrenWithProps(profile);
+
+    return (
+      <Loader loaded={loaded}>
+        <Jumbotron profile={ profile } text={ preferredName }>
+          <div className="card-copy">
+             { childrenWithProps }
```
aliceriot: I don't think you need to do this here, you should be able to just render the `PrivacyForm` like this:

```js
<PrivacyForm
  saveProfile={this.saveProfile}
  {...this.props}
/>
```

if you do that you can simplify the stuff going on in `dashboard_routes.js` too, you won't need to have a child route rendering the `<PrivacyForm>` under the route rendering the `<SettingsPage>`.

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,81 @@
...
+  componentDidMount() {
+    this.fetchProfile();
+  }
+
+  componentDidUpdate() {
+    this.fetchProfile();
+  }
+
+  componentWillUnmount() {
+    const { dispatch, params: { username } } = this.props;
```
aliceriot: when we use params on the `UserPage` component it's a prop that's injected at runtime by react-router, we don't need to worry about it here.

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  return {
+    profiles: state.profiles,
+    ui: state.ui,
+    params: params
+  };
+};
+
+export default connect(mapStateToProps)(SettingsPage);
```
aliceriot: Since we don't need to inject the `params` prop here you should be able to use `ProfileFormContainer.mapStateToProps` instead of declaring a new state -> props function here.

static/js/components/PrivacyForm.js
aliceriot: actually I'd even just inline this down where it's used.

static/js/containers/SettingsPage.js
amir-qayyum-khan: it need location, router in props thats why we need to place it in dashboard_routs.

static/js/containers/SettingsPage.js
amir-qayyum-khan: @aliceriot i dont think so, I think this is injecting because of routs. https://github.com/mitodl/micromasters/blob/master/static/js/dashboard_routes.js#L43, 

I checked it, even if it add this line in settings ``export default connect(ProfileFormContainer.mapStateToProps)(UserPage);`` it do not supply user name param

static/js/containers/SettingsPage.js
amir-qayyum-khan: https://github.com/mitodl/micromasters/pull/519#discussion_r67499200

amir-qayyum-khan:
@aliceriot these methods are not working in settings page but working on userpage, can you give me idea how to debug it or is their any issue in code?
- ```componentDidMount()```
- ```componentDidUpdate()```

https://github.com/mitodl/micromasters/pull/519/files#diff-0edec7f878bf81603aeb87f756925f46R17

static/js/containers/SettingsPage.js
aliceriot: Yes, that's what I meant above, react-router injects the `params` prop because we declare the url parameter for the `/users/:username` page in the route declaration.

We don't need the username here at all, however, because on `/settings` we're never going to be editing another users profile.

aliceriot: Does that make sense? We need to worry about the `username` url parameter (and check that it is or isn't the same as `SETTINGS.username`) when we're on `/users` because you can visit another user's profile by visiting `/users/some_other_users_username`.

On the settings page we're only concerned with editing the current user's profile, so all of this is ancillary. Compare how we're injecting the profile into the profile form steps (`/profile/personal`, etc) to what we're doing on `/users`.

ui/url_utils.py
```diff
@@ -6,3 +6,4 @@
 PROFILE_URL = '/profile/'
 TERMS_OF_SERVICE_URL = '/terms_of_service/'
 USERS_URL = '/users/'
+SETTING = "/settings/"
```
noisecapella: This should be `SETTINGS_URL`

static/js/dashboard_routes.js
```diff
@@ -39,6 +41,10 @@ export function makeDashboardRoutes(browserHistory, store, onRouteUpdate) {
               <Route path="privacy" component={PrivacyTab} />
             </Route>
             <Route path="/terms_of_service" component={TermsOfServicePage} />
+            <Route path="/settings" component={SettingsPage} >
```
noisecapella: `/settings/<username>` doesn't make sense here since we only want to view the settings for the current user. Why not just `/settings`?

aliceriot: There's a bunch of comments earlier on the PR addressing this issue, fyi

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,66 @@
...
+import {
+  startProfileEdit,
+  FETCH_PROCESSING,
+  clearProfile,
+} from '../actions/index';
+import ProfileFormContainer from './ProfileFormContainer';
+
+class SettingsPage extends ProfileFormContainer {
+  componentDidMount() {
+    this.fetchProfile();
```
noisecapella: This should be done already in App.js

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    return (
+      <Grid className="profile-tab-grid">
+        <Cell col={12}>
+          <h4>Who can see your profile?</h4>
+          { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) } <br />
+        </Cell>
+        <Cell col={12}>
+          <ProfileProgressControls { ...profileProgressControlsOptions } />
```
noisecapella: This shouldn't be part of the `PrivacyForm` since it's specific to the privacy tab. Can you move it to `PrivacyTab` similar to `EmploymentTab` and `EducationTab`?

static/js/containers/SettingsPage.js
amir-qayyum-khan: @noisecapella  there is issue with that logic, when i dont use ``fetchProfile`` then in ``render()`` 
 ``const { profiles } = this.props;`` gets undefined. it seems app.js triggers after SettingsPage.js render

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,26 @@
...
+    ui:             React.PropTypes.object,
+    updateProfile:  React.PropTypes.func,
+  };
+
+  render() {
+    return (
+      <Grid className="profile-tab-grid privacy-form">
+        <Cell col={12}>
+          <h4 className="privacy-form-heading">Who can see your profile?</h4>
+          { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) } <br />
```
aliceriot: can we take care of the spacing and layout with css instead of `<br />` tags?

amir-qayyum-khan:
@aliceriot one of my test is failing, any idea how to debug? https://github.com/mitodl/micromasters/pull/519/files#diff-8a199ab411c143cdcd0f260400602486R104
```trace
1) SettingsPage Authenticated settings page save privacy form save privacy changes:
     TypeError: Cannot read property 'then' of undefined
      at index.js:74:12
      at node_modules/redux-thunk/lib/index.js:12:16
      at SettingsPage.saveProfile (ProfileFormContainer.js:138:14)
      at ProfileProgressControls.saveProfileStep (profile_edit.js:280:10)
      at ProfileProgressControls._this.saveAndContinue (ProfileProgressControls.js:24:21)
      at Object.invokeGuardedCallback [as invokeGuardedCallbackWithCatch] (node_modules/react/lib/ReactErrorUtils.js:26:12)
      at executeDispatch (node_modules/react/lib/EventPluginUtils.js:85:21)
      at Object.executeDispatchesInOrder (node_modules/react/lib/EventPluginUtils.js:110:5)
      at executeDispatchesAndRelease (node_modules/react/lib/EventPluginHub.js:42:22)
      at executeDispatchesAndReleaseSimulated (node_modules/react/lib/EventPluginHub.js:50:10)
      at forEachAccumulated (node_modules/react/lib/forEachAccumulated.js:26:8)
      at Object.EventPluginHub.processEventQueue (node_modules/react/lib/EventPluginHub.js:213:7)
      at node_modules/react/lib/ReactTestUtils.js:436:22
      at ReactDefaultBatchingStrategyTransaction.Mixin.perform (node_modules/react/lib/Transaction.js:136:20)
      at Object.ReactDefaultBatchingStrategy.batchedUpdates (node_modules/react/lib/ReactDefaultBatchingStrategy.js:63:19)
      at Object.batchedUpdates (node_modules/react/lib/ReactUpdates.js:98:20)
      at Object.click (node_modules/react/lib/ReactTestUtils.js:434:18)
      at SettingsPage_test.js:75:28
      at node_modules/redux-asserts/lib/index.js:117:9
      at IntegrationTestHelper.listenForActions (node_modules/redux-asserts/lib/index.js:91:12)
      at confirmSaveButtonBehavior (SettingsPage_test.js:74:14)
      at SettingsPage_test.js:104:18
```

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,131 @@
...
+
+    return React.Children.map([<PrivacyForm />], (child) => (
+      React.cloneElement(child, {
+        profile: profile,
+        errors: errors,
+        ui: ui,
+        updateProfile: this.updateProfile.bind(this, isEdit),
+        saveProfile: this.saveProfile.bind(this, isEdit),
+        clearProfileEdit: this.clearProfileEdit,
+        fetchProfile: this.fetchProfile
```
aliceriot: I don't think you need to do this wrangling with `React.Children.map` here - there's a concrete reason why we use this pattern on `ProfileFormContainer` to inject props into `this.children`, but we don't need to do so here.

When we're subclassing `ProfileFormContainer` we don't know what components we'll need to inject props into at runtime - which components are going to be found on `this.props.children` is determined by what route we're at, and which components `react-router` renders into `this.props.children`. Here we have complete knowledge of what components will be rendered at runtime, so we don't need this more general pattern.

You should be able to instead just do something like this:

```js
<PrivacyForm
  {...this.props}
  updateProfile={this.updateProfile.bind(this, isEdit)}
  saveProfile={this.saveProfile.bind(this, isEdit)}
  clearProfileEdit={this.clearProfileEdit}
  fetchProfile={this.fetchProfile}
/>
```

aliceriot: (this would apply also for the `privacyControlsWithProps` method below)

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,26 @@
...
+class PrivacyForm extends ProfileFormFields {
+  static propTypes = {
+    profile:        React.PropTypes.object,
+    ui:             React.PropTypes.object,
+    updateProfile:  React.PropTypes.func,
+  };
+
+  render() {
+    return (
+      <Grid className="profile-tab-grid privacy-form">
```
aliceriot: I think it would overall make more sense to have the parent component worry about layout, and have the `PrivacyForm` just return basically the form field and wire everything up correctly, so when I want to plop it on, say, `PrivacyTab` I'd just do somehting like

```js
<Cell col={12}>
  <PrivacyForm {...this.props} otherProps={...} />
</Cell>
```
That way the layout is the responsibility of the context where we actually want to render it, and the `PrivacyForm` component ends up being a bit more flexible and minimal.

static/js/containers/ProfileFormContainer.js
```diff
@@ -47,8 +47,10 @@ class ProfileFormContainer extends React.Component {
 
   fetchProfile = () => {
     const { dispatch, profiles, params: { username } } = this.props;
-    if (profiles[username] === undefined || profiles[username].getStatus === undefined) {
-      dispatch(fetchUserProfile(username));
+    let _username = username || SETTINGS.username;
```
noisecapella: Maybe we should change `fetchProfile` to take in the username explicitly? Then the calling function can determine which username to use

aliceriot: I think that would make more sense, as is were just increasing the degree to which it's coupled to particular use-cases.

static/js/containers/SettingsPage.js
aliceriot: you could also bind `this.updateProfile` and `this.saveProfile` in the constructor, kind of preference which one you think is cleaner.

amir-qayyum-khan:
Done @noisecapella  @aliceriot

aliceriot:
thanks, looking now

static/js/components/PrivacyTab.js
```diff
@@ -30,10 +30,9 @@ class PrivacyTab extends ProfileFormFields {
             We care about your privacy.
           </Cell>
         </Grid>
-        <Grid className="profile-tab-grid">
+        <Grid className="profile-tab-grid privacy-form">
           <Cell col={12}>
-            <h4>Who can see your profile?</h4>
-            { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) } <br />
+            <PrivacyForm {...this.props} />
```
aliceriot: Here you have `PrivacyForm` inside of a `<Cell col={12}>` component, this makes the `<Grid>` and `<Cell>` components declared in the render method of `PrivacyForm` unnecessary - if you were getting the warning about 'adjacent JSX elements' a simple solution is to wrap the `<h4>` element and the call to `boundRadioGroupField` in a `<div>`.

aliceriot: (same applies below in `SettingsPage`)

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,101 @@
...
+      if (profileFromStore.edit !== undefined) {
+        errors = profileFromStore.edit.errors;
+        profile = profileFromStore.edit.profile;
+        isEdit = true;
+      } else {
+        profile = profileFromStore.profile;
+        errors = {};
+        isEdit = false;
+      }
+      preferredName = profile.preferredName || SETTINGS.name;
```
aliceriot: there's a function in `util.js` called `getPreferredName` that you may want to use here (it's tested and so on).

aliceriot: It just takes a profile object as an argument and get's the most up-to-date preferred name (`Profile.preferred_name` || `SETTINGS.preferred_name`, etc).

aliceriot:
functionality looks good!

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,26 @@
+/* global SETTINGS: false */
```
aliceriot: could you add flow typechecking to this file? you can do so by adding a comment on the first line like this:

```js
// @flow
```

aliceriot: I tried it out and it passes with no changes

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,101 @@
+/* global SETTINGS: false */
```
aliceriot: it'd be good to add flow to this as well, looks like it also passes without modification.

aliceriot:
couple of small comments on the code, this is looking good though!

aliceriot:
I think this looks pretty good, but the typechecker isn't passing. Would you be able to fix the type errors? The first one I think is that you need to add a semicolon at the end of the `mapStateToProps` method on `ProfileFormContainer`, then after that there are like 3 type mismatch errors.

If you haven't played with it yet you can run the typechecker by doing `docker-compose run watch npm runs-script flow`.

aliceriot:
or wait, I think you're on OSX so maybe that's just `npm run-script flow`.

amir-qayyum-khan:
Done @aliceriot

static/js/components/PrivacyForm.js
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+
+import ProfileFormFields from '../util/ProfileFormFields';
+
+class PrivacyForm extends ProfileFormFields {
+  static propTypes = {
+    profile:        React.PropTypes.object,
+    ui:             React.PropTypes.object,
+    updateProfile:  React.PropTypes.func,
```
aliceriot: would you mind switching this to a flow-style props type declaration? That would look like:

```js
import type { Profile } from '../flow/profileTypes';

class PrivacyForm extends ProfileFormFields {
  props: {
    profile:        Profile,
    ui:             UIState,
    updateProfile:  Function,
  };

  ...
}
```

this lets flow check the types for things on `this.props` properly, so it can know the type of, for instance, `profile.preferred_name`.

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,101 @@
...
+import ProfileFormContainer from './ProfileFormContainer';
+import PrivacyForm from '../components/PrivacyForm';
+import ProfileProgressControls from '../components/ProfileProgressControls';
+import {
+    combineValidators,
+    privacyValidation,
+} from '../util/validation';
+import { getPreferredName } from '../util/util';
+
+class SettingsPage extends ProfileFormContainer {
```
aliceriot: same comment here, would you mind adding a `props` declaration? 

something like:

```js
import type { Profile } from '../flow/profileTypes';

class SettingsPage extends ProfileFormContainer {
  props: {
    profiles: {[key: string]: Profile}, 
    dispatch: Function,
    ui:       UIState,
  };

  ...
}
```

aliceriot:
one little tweak, almost there! back at you @amir-qayyum-khan

amir-qayyum-khan:
done @aliceriot

aliceriot:
cool, taking a look now

static/js/containers/SettingsPage.js
```diff
@@ -0,0 +1,108 @@
...
+    ui:       UIState,
+  };
+
+  componentDidMount() {
+    this.fetchProfile(SETTINGS.username);
+  }
+
+  componentDidUpdate() {
+    this.fetchProfile(SETTINGS.username);
+  }
```
aliceriot: Sorry, I just noticed this. I deleted these two methods (`componentDidUpdate` and `componentDidMount`) and it seems like everything is taken care of by `App`, which also calls these methods. 

Can you:

- delete those two methods
- add a default value to `profile` (e.g. `profile = {};`) in the render method, so that if the profile GET request hasn't come back yet we have something to pass to `ProfileProgressControls`.

sorry! I should have realized that earlier >_M

aliceriot:
one little thing, then I think it should be good!

amir-qayyum-khan:
done @aliceriot

aliceriot:
hrm, so as is this isn't passing the typechecker - here's what we can do to make it work:

- remove the object literal for `profile` in `render` on `Settings` (so just go back to `let profile;`
- remove `.isRequired` from the profile prop on `ProfileProgressControls`.

amir-qayyum-khan:
@aliceriot still tests are failing, after doing above 2 steps

amir-qayyum-khan:
@aliceriot fixed issue by updating action flags in test suit

aliceriot:
merge it! :shipit: :+1: :confetti_ball: :fireworks:

PR #520, opened 2016-06-10
Collapse education on `/users/`

### LINKED ISSUE #498: Collapse multiple education cards on User page into one card
Instead of:

screenshot-micromasters-ci herokuapp com 2016-06-07 13-43-15

we'd like something more like:

screen shot 2016-06-07 at 1 45 57 pm

I'm not sure if this means that we can't reuse the components from the education profile form. Also, if we collapse all the education into one form, we'll need a way for users to identify the level of a particular degree.

aliceriot:

closes #498 

#### What's this PR do?

This adds a different display component for the education history which we can use on the `/users` page.

Review the new components (`EducationDisplay` and `EducationDialog`) and the changes to `EducationForm`.

Visit `/users`. You should see your education history, and be able to delete and add new entries. On a different user's page, the edit, add, and delete buttons should be hidden. There should be no regressions on the education profile step.

### DIFF
--- a/static/js/containers/UserPage_test.js
+++ b/static/js/containers/UserPage_test.js
@@ -68,7 +68,7 @@ describe("UserPage", () => {
       it('shows the education component', done => {
         renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
           let title = div.getElementsByClassName('profile-card-title')[1];
-          assert.equal(title.textContent, 'High school');
+          assert.equal(title.textContent, 'Education');
           done();
         });
       });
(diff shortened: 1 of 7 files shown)


noisecapella:
Functionality looks good, checking code

aliceriot:
Cool, I just rebased over your changes.

static/js/containers/UserPage_test.js
```diff
@@ -149,7 +149,7 @@ describe("UserPage", () => {
 
       it('should let you add an education entry', done => {
```
noisecapella: This is outside the scope of the issue but would you mind adding a test to save the dialog box? It looks like we don't have any coverage for the code which saves or cancels this dialog or the work history dialog

aliceriot: Sure, that's a good idea to add!

noisecapella: Or if you'd rather merge this first that would be fine, we can do it as a separate issue

noisecapella:
:+1:

static/js/containers/UserPage_test.js
aliceriot: I'll open an issue, I'll probably just do it right away.

PR #504, opened 2016-06-07
Added validation for employment and education switches

noisecapella:
#### What are the relevant tickets?
Fixes #433 
Rebased on #508 

#### What's this PR do?
Adds validation for switches for education and employment, so that a validation error will be produced if:
 - the switch is on but there is nothing in the card
 - the switch is off but there's something in the card

#### Where should the reviewer start?
static/js/util/validation.js

#### How should this be manually tested?
In the profile pages, start at personal and go forward to the education tab. You should not get a validation error on the personal page. In the education page leave some cards blank and go forward. For all cards which are blank you should see a red validation error which has the label of the card within it. If you turn off a card or add an item to another card and continue, the validation error should go away.

Go back to the education page and add an item to one of the cards if necessary. Close the tab and click 'Save and Continue'. You should see a validation error.

For the employment tab, make it blank but have the switch turned on. When you proceed you should see a validation error. Add an item, turn the switch off, and click 'Save and Continue'. You should see a different validation error.

Also check that the user page functions exactly as it did before.

PR #523, opened 2016-06-10
Added extra validation for dates

### LINKED ISSUE #403: Better validation error when date is invalid
Currently we just use '... is required' for all validation messages. If the date is invalid because it's impossible we should adjust our error message accordingly

Should handle these cases:
 - A date which is out of range or has invalid characters (what should a valid range be?)
 - An end date which is before the begin date
 - A birthdate which is after today (do we want to restrict it further to at least 13 years old?)

noisecapella:
Fixes #403 
Rebased on #504 

#### What's this PR do?
Handles the three validation cases mentioned in the issue

 - Verify that an error occurs if you change your date of birth to a date in the future and click Save
 - In a work history entry try to add a year but not a month for the end date. Verify that the error message is: "Please enter a valid end date or leave it blank". If you remove both year and month, everything should work fine since that means you are currently working there.
 - In a work history entry verify that you get an error if your end date is before your start date. Verify that the message is: "End date cannot be before start date"
 - Also in work history, remove the start date, click 'Save' and verify that the error message is "Please enter a valid start date"
 - In an education entry verify that you see this error message when the graduation date is missing or invalid: "Please enter a valid graduation date"

### DIFF
--- a/static/js/util/validation.js
+++ b/static/js/util/validation.js
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import moment from 'moment';
 
 import { filterPositiveInt } from './util';
 import { HIGH_SCHOOL, EDUCATION_LEVELS } from '../constants';
@@ -10,13 +11,15 @@ let handleNestedValidation = (profile, keys, nestedKey) => {
   return _.flatten(profile[nestedKey].map((v, i) => nestedFields(i)));
 };
 
+let isNilOrEmptyString = val => _.isNil(val) || val === "";
+
 let checkFieldPresence = (profile, requiredFields, messages) => {
   let errors = {};
 
   for (let keySet of requiredFields) {
     let val = _.get(profile, keySet);
-    if (_.isUndefined(val) || _.isNull(val) || val === "" ) {
-      _.set(errors, keySet,  `${messages[keySet.slice(-1)[0]]} is required`);
+    if (isNilOrEmptyString(val)) {
+      _.set(errors, keySet,  messages[keySet.slice(-1)[0]]);
     }
   }
   return errors;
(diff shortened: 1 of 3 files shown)


aliceriot:
code looks good, testing functionality

aliceriot:
- [x] Verify that an error occurs if you change your date of birth to a date in the future and click Save
- [ ] In a work history entry try to add a year but not a month for the end date. Verify that the error message is: "Please enter a valid end date or leave it blank". If you remove both year and month, everything should work fine since that means you are currently working there.
- [ ] In a work history entry verify that you get an error if your end date is before your start date. Verify that the message is: "End date cannot be before start date"
- [ ] Also in work history, remove the start date, click 'Save' and verify that the error message is "Please enter a valid start date"
- [ ] In an education entry verify that you see this error message when the graduation date is missing or invalid: "Please enter a valid graduation date"

aliceriot:
for the second case I found an edge case.

1. create a work history entry that has a proper end date
2. edit that work history
3. delete the 'month' field value
4. click 'save'
5. entry is saved without an end date (despite there being a year value in the year field when 'save' is clicked)

aliceriot:
one edge case, other than that things seem to be working correctly.

noisecapella:
Good catch, it looks like `_.isEmpty(number)` returns true... :see_no_evil:

noisecapella:
Ok, that case should be handled now. Also, I discovered that moment JS parses two digit years 19xx despite being passed a format string with YYYY, so I changed the year validation to require all dates to be within 1800 >= year > 2100

aliceriot:
Yep, confirmed that case is fixed! looking over the code now.

static/js/util/validation.js
```diff
@@ -11,12 +11,17 @@ let handleNestedValidation = (profile, keys, nestedKey) => {
   return _.flatten(profile[nestedKey].map((v, i) => nestedFields(i)));
 };
 
+/**
+ * Is the value === undefined, null, or ""?
+ */
```
aliceriot: I don't think this comment is super necessary, I think this function documents itself. Personal style preference though.

aliceriot:
One little comment, other than that I think the code looks :100:

noisecapella:
Comment addressed

aliceriot:
it's a :+1: from me!

PR #535, opened 2016-06-13
Used chai to assert Promise behavior

### LINKED ISSUE #475: Use chai eventually assert to assert promise results
This wouldn't change functionality, it would just make our tests cleaner and more explicit

http://chaijs.com/plugins/chai-as-promised/

cc @gsidebo

noisecapella:
Fixes #475 

#### What's this PR do?
Replaces `assert` with `chai` and `chai-as-promised`. Tests are changed to always return a promise. Tests that have a rejected promise will use `assert.isRejected` instead.

Just check that the tests pass


aliceriot:
everything's passing, code looks good, :+1:

PR #538, opened 2016-06-14
Flipped CourseList arrow

### LINKED ISSUE #536: Switch arrows on dashboard
When the expander is open the arrow should be down, otherwise it should be up

noisecapella:
Fixes #536 

#### What's this PR do?
Flips the arrows so they're consistent with most people's use of expander elements on web sites.

View the dashboard with a course with at least one run. The chevron should point up by default. When you click it it should point down and you should see the run appear below the course.

### DIFF
--- a/static/js/components/CourseList.js
+++ b/static/js/components/CourseList.js
@@ -62,7 +62,7 @@ class CourseList extends React.Component {
           };
           expanderSpan = <span
             onClick={toggleExpander}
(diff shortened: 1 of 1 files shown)


noisecapella:
Actually I'm not sure about this... for example in edX studio the arrows are pointing down by default and turn to point up when the list has opened

noisecapella:
I'm going to close this and reopen if we decide it's something we should change

PR #540, opened 2016-06-14
Set work history switch to be on by default

### LINKED ISSUE #516: Change work history switch to be on by default
This will have the user be required to turn the switch off before proceeding if they don't want to enter their information

aliceriot:

closes #516 

#### What's this PR do?

Removes a `componentWillMount` lifecycle callback on `EmploymentForm` which was setting the switch to default to 'off' (in the case that a user has not yet entered any work history).

Now the switch should always initially have the default value from the store, which is `true`.


giocalitri:
just a question before checking the functionality: aren't there any tests for this part of the code?

aliceriot:
Not really, we don't have many full on integration tests.

giocalitri:
👍

PR #542, opened 2016-06-14
Updated field of study select to use JSON data

### LINKED ISSUE #443: Update Field of Study in education profile to use JSON data
Currently this field is a free text field. We would like to convert it to a type-ahead field. As the user types, we will look up the field of study, and populate a dynamic drop-down menu with the results. 

For the future (not this story), we would like to allow the user to enter their own text (i.e. not matching anything returned from the API). 

We will store the CIP code in the profile, similar to what we do with country codes.

gsidebo:

closes #443 

#### What's this PR do?

- changes the field of study select field to get its values from a static JSON file
- major refactor to all of our select fields. SelectField is now a 'formal' React component, and several changes were made to initialization, props, and code reuse
- adds an option to the select field to limit autocomplete results to a given number

AutoComplete.js and the *SelectField.js files

*NOTE: a lot of ./static/js/components/inputs/SelectField.js and ./static/js/components/inputs/inputs_test.js are copied from the old boundSelectField function and profile_page_test.js respectively*

Fill out a profile, paying special attention to the fields of study select fields in the education tab


static/js/components/inputs/SelectField.js
```diff
@@ -0,0 +1,164 @@
+import React from 'react';
+import _ from 'lodash';
+
+import MenuItem from 'material-ui/MenuItem';
+import AutoComplete from '../AutoComplete';
+
+
+export default class SelectField extends React.Component {
+  constructor(props) {
+    super(props);
```
gsidebo: this constructor probably needs to be refactored. working on that (and linting errors) now

static/js/components/inputs/CountrySelectField.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import _ from 'lodash';
+import iso3166 from 'iso-3166-2';
+import SelectField from './SelectField';
+
+let countryOptions = Object.keys(iso3166.data).map(code => ({
+  value: code,
+  label: iso3166.data[code]['name']
+}));
+countryOptions = _.sortBy(countryOptions, 'label');
```
aliceriot: any particular reason this is outside of the class declaration? I would have probably declared this in a `render` method as a little helper, probably fine either way though.

static/js/components/inputs/StateSelectField.js
```diff
@@ -0,0 +1,41 @@
...
+    const {
+      profile,
+      countryKeySet
+    } = this.props;
+    let options = [];
+    let country = _.get(profile, countryKeySet);
+    if (iso3166.data[country] !== undefined) {
+      options = Object.keys(iso3166.data[country].sub).map(code => ({
+        value: code,
+        label: iso3166.data[country].sub[code].name
```
aliceriot: If I'm understanding this correctly should should be able to make this a little more concise with `Object.entries`:

```js
Object.entries(iso3166.data[country].sub).map(([key, value]) => (
  { value: key, label: value.name }
);
```

static/js/components/inputs/CountrySelectField.js
gsidebo: we only need to create this set of country options once. any instance of CountrySelectField is going to use the exact same set of options, so to me it made sense to avoid calculating the options every time a CountrySelectField renders

static/js/components/inputs/StateSelectField.js
gsidebo: indeed. this was copy/paste from the old method. i'll look into this

aliceriot: Ah, yeah, we haven't always had the `Object.entries` polyfill, this might predate us adding it.

static/js/components/inputs/CountrySelectField.js
aliceriot: gotcha, that makes sense.

static/js/components/inputs/StateSelectField.js
```diff
@@ -0,0 +1,41 @@
...
+      }));
+      options = _.sortBy(options, 'label');
+    }
+    return options;
+  }
+
+  render() {
+    const {
+      stateKeySet
+    } = this.props;
```
aliceriot: I'd put this destructuring on one line (just looks a little nicer that way).

static/js/components/inputs/CountrySelectField.js
```diff
@@ -0,0 +1,41 @@
...
+    updateProfile:            React.PropTypes.func,
+    stateKeySet:              React.PropTypes.array,
+    countryKeySet:            React.PropTypes.array
+  };
+
+  onChange = newProfile => {
+    const {
+      stateKeySet,
+      updateProfile
+    } = this.props;
```
aliceriot: this can probably go on one line too

static/js/components/inputs/FieldsOfStudySelectField.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import _ from 'lodash';
+import SelectField from './SelectField';
+import FIELDS_OF_STUDY from '../../fields_of_study';
+
+let fieldOfStudyOptions = _.toPairs(FIELDS_OF_STUDY).map(fosEntry => ({
+  value: fosEntry[0],
+  label: fosEntry[1]
+}));
```
aliceriot: you could do a destructuring thing here in the arguments to the arrow function:

```js
_.toPairs(FIELDS_OF_STUDY).map(([code, label]) => (
  { value: code, label: label }
));
```

static/js/components/inputs/StateSelectField.js
gsidebo: will do

static/js/components/inputs/FieldsOfStudySelectField.js
gsidebo: will do

static/js/components/inputs/FieldsOfStudySelectField.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import _ from 'lodash';
+import SelectField from './SelectField';
+import FIELDS_OF_STUDY from '../../fields_of_study';
+
+let fieldOfStudyOptions = _.map(FIELDS_OF_STUDY, (name, code) => ({
+  value: code,
+  label: name
+}));
```
aliceriot: I prefer using standard library stuff when possible over lodash - I think this is personal preference though, so if you feel strongly that this looks / feels better than something like this:

```js
Object.entries(FIELDS_OF_STUDY).map( ([k,v]) => ( ... ) );
```

feel free to keep what you have.

static/js/components/inputs/CountrySelectField.js
```diff
@@ -0,0 +1,37 @@
...
+import _ from 'lodash';
+import iso3166 from 'iso-3166-2';
+import SelectField from './SelectField';
+
+let countryOptions = _(iso3166.data)
+  .map((country_info_obj, country_code) => ({
+    value: country_code,
+    label: country_info_obj.name
+  }))
+  .sortBy('label').value();
```
aliceriot: same standard library vs lodash thing here

static/js/components/inputs/SelectField.js
```diff
@@ -0,0 +1,185 @@
...
+
+    // Set some inter-related option display properties on this component
+    let optionDisplayProps = { resultLimit: resultLimit };
+    let optionDisplaySettingsFunc = resultLimit ? showLimitedOptionSettings : showAllOptionSettings;
+    _.assignIn(this, optionDisplaySettingsFunc(optionDisplayProps));
+
+    this.autocompleteStyleProps = this.props.autocompleteStyleProps || {
+      menuStyle: {maxHeight: 300},
+      fullWidth: true
+    };
```
aliceriot: why not just set a default prop value for this?

static/js/components/inputs/CountrySelectField.js
gsidebo: the previous one i agree with, but in this case, we use _.sortBy immediately after the map. to me it makes a lot more sense to use _ and chain them

aliceriot: fair enough

static/js/components/inputs/SelectField.js
```diff
@@ -0,0 +1,185 @@
...
+    updateProfile(clone);
+  };
+
+  onBlur = () => {
+    // clear the edit value when we lose focus. In its place we will display
+    // the selected option label if one is selected, or an empty string
+    const {
+      profile,
+      updateProfile
+    } = this.props;
```
aliceriot: could you put this destructuring on one line?

static/js/components/inputs/FieldsOfStudySelectField.js
gsidebo: here's a fun one. using `Object.entries` inside of the React component (or, most likely, anything that isn't executed immediately on startup) works fine, but using it outside of the class definition like this results in 'Object.entries is undefined'. the polyfill apparently hasn't had a chance to run yet when this line is executed

aliceriot: oh interesting! we're doing the polyfill in `dashboard.js`, but I bet the `import` statements cause this code to be executed first. you could try moving this:

```js
import entries from 'object.entries';
if (!Object.entries) {
  entries.shim();
}
```

higher in `dashboard.js`?

static/js/components/inputs/SelectField.js
gsidebo: forgot about that whole thing. will do

static/js/components/inputs/FieldsOfStudySelectField.js
gsidebo: ss 2016-06-15 at 4 39 28 pm

moved it to the top. just doesn't want to execute first

aliceriot: hrm, maybe it's something to do with how webpack bundles stuff? haha. maybe if we import and run the shim in this file too?

gsidebo: pretty hacky IMO. if we care about this enough i think we should open an issue to look into it

aliceriot: yeah, in this case let's just stick with the lodash solution

gsidebo: fyi https://github.com/mitodl/micromasters/issues/554

aliceriot:
Hadn't actually tested the functionality yet, but everything seems to be working correctly.

static/js/components/inputs/FieldsOfStudySelectField.js
```diff
@@ -0,0 +1,34 @@
...
+  static autocompleteStyleProps = {
+    menuStyle: {maxHeight: 300, width: 600},
+    listStyle: {width: '100%'},
+    fullWidth: false
+  };
+
+  render() {
+    return <SelectField
+      options={fieldOfStudyOptions}
+      resultLimit={this.props.resultLimit}
```
aliceriot: `resultLimit` should get passed automatically when we splat in `{...this.props}` below.

gsidebo:
another somewhat ugly merge conflict with master now. @aliceriot i may ask you to pair with me to figure this one out

aliceriot:
sure!

gsidebo:
finally, mercifully... ready for review again

static/js/util/ProfileFormFields.js
```diff
@@ -60,28 +48,36 @@ export default class ProfileFormFields extends React.Component {
...
   boundDateField: Function;
   boundRadioGroupField: Function;
-  countryOptions: Option[];
   genderOptions: Option[];
   languageOptions: Option[];
   privacyOptions: Array<{value: string, label: string, helper: string}>;
   educationLevelOptions: Option[];
   industryOptions: Option[];
 
+  defaultInputComponentProps = () => {
```
aliceriot: can you add a type annotation here? this is the only type error I'm seeing on the branch currently.

if should look something like

```js
defaultInputComponentProps: Function = (): void => {
  ...
};
```

aliceriot:
This is looking good, and the functionality all looks correct.

I have one small comment about a type annotation, and I think it would be good to add typechecking to all the new classes here. Just to see I added `// @flow` to all the new classes, and there's currently around 20 type errors. 

If you don't want to worry about that in this PR I have a branch open where I'm adding typechecking to all the remaining files we aren't currently checking, so I could just take care of adding it after this is merged.

gsidebo:
@aliceriot added the type checking

aliceriot:
awesome, taking a look

aliceriot:
LGTM :+1: :fireworks: :confetti_ball: :shipit:

PR #543, opened 2016-06-14
Hide switch on user page

### LINKED ISSUE #533: Remove work history switch for users page

aliceriot:
this closes #533 

This removes the switch from the user page, by adding a boolean prop to `EmploymentForm` which we can set to `false` whenever we want to hide the switch.

Make sure there no regressions in adding items and so on (on `/profile/professional`, the user page will not be usable until #540 is merged).

### DIFF
--- a/static/js/components/EmploymentForm.js
+++ b/static/js/components/EmploymentForm.js
@@ -183,6 +183,7 @@ class EmploymentForm extends ProfileFormFields {
         showWorkDeleteDialog,
       },
       errors,
+      showSwitch,
     } = this.props;
     const actions = [
       <Button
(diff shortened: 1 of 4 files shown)


static/js/components/EmploymentForm.js
```diff
@@ -16,6 +16,14 @@ import ProfileFormFields from '../util/ProfileFormFields';
 import ConfirmDeletion from './ConfirmDeletion';
 
 class EmploymentForm extends ProfileFormFields {
+  static propTypes = {
```
noisecapella: Does this combine these `propTypes` with the ones from its parent class?

noisecapella:
Can you rebase please?

static/js/components/EmploymentForm.js
aliceriot: oh, you know I'm not sure how that works. I can just move this onto `ProfileFormFields`

noisecapella: I did a `console.log` and it looks like this replaces the propTypes so you'll need to combine the two or move it into the parent class like you said

aliceriot: moved it to the parent class.

noisecapella:
:+1:

PR #545, opened 2016-06-14
Added flow typechecking for JS

### LINKED ISSUE #522: Investigate gradual typing options for JS
At PyCon there was a lot of excitement about gradual typing in Python, and the libraries that let you start checking type annotations in Python code, such as [mypy](http://mypy-lang.org/) and [pytype](https://github.com/google/pytype).

I think it's worth investigating options for gradual typing in JavaScript, we have several functions (especially in `util/util.js` and `util/validation.js`) that are used now in several components, as well as several components that are rendered in different contexts. Adding type annotations to the 'core functionality' of our app could give us a lot more confidence about the correctness of our code.

I've done a little bit of initial investigation into using the [Flow](https://flowtype.org/) typechecker for JS. I think it's well-suited to our needs, in that it's possible to gradually add typing annotations to an existing project and it's not necessary to type check the whole project. My understanding is that something like TypeScript would require a more thorough overhaul and tooling change, although that may be a misconception due to focusing on Flow.

I've added Flow to a personal project on this branch: https://github.com/aliceriot/literate-crypto/pull/12
There's a lot of other random noise in that PR (I'm messy on personal projects, sue me) but if you look at the changes to `Nav.js` and a couple of other files you can see an example of what the type annotations for Flow look like, as well as the tooling changes that were necessary. 

I found the tooling for using it to be fairly straightforward to use. There's a [babel plugin](https://www.npmjs.com/package/babel-plugin-transform-flow-strip-types) that will automatically strip out the type annotations when transpiling code, and there's also an [ESLint](https://www.npmjs.com/package/eslint-plugin-flowtype) plugin to set rules for linting the type annotations. The typechecker itself is installed via npm and run from the command-line, so I think it would be straightforward to integrate into our test suite.

I'm unclear on how much work it would be to actually get to a useful level of annotation on our project. With Flow you opt-in to typechecking on certain files by adding a `// @flow` comment at the top of the file, so we could probably make some minimal changes to get the whole project passing the typechecker and then add types and add files to the checker as we proceed and as we see fit.

I'd love to know what other folks writing lots of JS think! Do we think this would be helpful? Would it get in the way? Is there another, better option for annotating JS types that would work well with our current tooling?

aliceriot:

closes #522 (I guess?)

#### What's this PR do?

This adds the flow typechecker to our project. This comprises a few changes:

- adding the package for the typechecker itself (flow-bin)
- a babel plugin to automatically strip out flow type annotations
- two eslint plugins to handle imported types correctly
- adding a bunch of type annotations to our code

Make sure nothing is broken! I think I only added annotations and didn't change any functionality, but it's a little hard to be sure.

You'll need to rebuild your docker containers for the typechecker to work, once you have you should be able to do:

```
docker-compose run watch npm run-script flow
```

to type-check the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "babel-eslint": "6.0.4",
     "babel-loader": "6.2.4",
     "babel-plugin-jsx": "1.2.0",
+    "babel-plugin-transform-flow-strip-types": "^6.8.0",
     "babel-plugin-transform-react-jsx": "6.8.0",
     "babel-preset-es2015": "6.6.0",
     "babel-preset-react": "6.5.0",
@@ -20,9 +21,12 @@
     "eslint": "2.9.0",
     "eslint-config-defaults": "9.0.0",
     "eslint-plugin-babel": "3.2.0",
+    "eslint-plugin-flow-vars": "^0.4.0",
+    "eslint-plugin-flowtype": "^2.2.7",
     "eslint-plugin-react": "5.0.1",
     "fetch-mock": "4.4.0",
     "file-loader": "0.8.5",
+    "flow-bin": "^0.27.0",
     "history": "2.1.1",
     "imports-loader": "^0.6.5",
     "iso-3166-2": "0.4.0",
@@ -67,6 +71,7 @@
     "postinstall": "./webpack_if_prod.sh",
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",
     "test": "./js_test.sh",
-    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --compilers js:babel-register --require static/js/global_init.js 'static/**/*/*_test.js'"
+    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --compilers js:babel-register --require static/js/global_init.js 'static/**/*/*_test.js'",
+    "flow": "flow check"
   }
 }
--- a/static/js/actions/ui.js
+++ b/static/js/actions/ui.js
@@ -1,82 +1,85 @@
+// @flow
 // general UI actions
+import type { Action } from '../flow/generalTypes';
+
 export const CLEAR_UI = 'CLEAR_UI';
 export const clearUI = () => ({ type: CLEAR_UI });
 
 export const UPDATE_DIALOG_TEXT = 'UPDATE_DIALOG_TEXT';
-export const updateDialogText = text => (
+export const updateDialogText = (text: string): Action => (
   { type: UPDATE_DIALOG_TEXT, payload: text }
 );
 
 export const UPDATE_DIALOG_TITLE = 'UPDATE_DIALOG_TITLE';
-export const updateDialogTitle = title => (
+export const updateDialogTitle = (title: string): Action => (
   { type: UPDATE_DIALOG_TITLE, payload: title }
 );
 
 export const SET_DIALOG_VISIBILITY = 'SET_DIALOG_VISIBILITY';
-export const setDialogVisibility = bool => (
+export const setDialogVisibility = (bool: boolean): Action => (
   { type: SET_DIALOG_VISIBILITY, payload: bool }
 );
 
 // work history actions
 export const SET_WORK_HISTORY_EDIT = 'SET_WORK_HISTORY_EDIT';
-export const setWorkHistoryEdit = bool => (
+export const setWorkHistoryEdit = (bool: boolean): Action => (
   { type: SET_WORK_HISTORY_EDIT, payload: bool }
 );
 
 export const SET_WORK_DIALOG_VISIBILITY = 'SET_WORK_DIALOG_VISIBILITY';
-export const setWorkDialogVisibility = bool => (
+export const setWorkDialogVisibility = (bool: boolean): Action => (
   { type: SET_WORK_DIALOG_VISIBILITY, payload: bool }
 );
 
 export const SET_WORK_DIALOG_INDEX = 'SET_WORK_DIALOG_INDEX';
-export const setWorkDialogIndex = index => (
+export const setWorkDialogIndex = (index: number): Action => (
   { type: SET_WORK_DIALOG_INDEX, payload: index }
 );
 
 // dashboard actions
 export const TOGGLE_DASHBOARD_EXPANDER = 'TOGGLE_DASHBOARD_EXPANDER';
-export const toggleDashboardExpander = (courseId, newValue) => ({
+export const toggleDashboardExpander = (courseId: number, newValue: boolean): Action => ({
   type: TOGGLE_DASHBOARD_EXPANDER,
   payload: { courseId, newValue }
 });
 
 // education actions
 export const SET_EDUCATION_DIALOG_VISIBILITY = 'SET_EDUCATION_DIALOG_VISIBILITY';
-export const setEducationDialogVisibility = bool => (
+export const setEducationDialogVisibility = (bool: boolean): Action => (
   { type: SET_EDUCATION_DIALOG_VISIBILITY, payload: bool }
 );
 
 export const SET_EDUCATION_DIALOG_INDEX = 'SET_EDUCATION_DIALOG_INDEX';
-export const setEducationDialogIndex = index => (
+export const setEducationDialogIndex = (index: number): Action => (
   { type: SET_EDUCATION_DIALOG_INDEX, payload: index }
 );
 
 export const SET_EDUCATION_DEGREE_LEVEL = 'SET_EDUCATION_DEGREE_LEVEL';
-export const setEducationDegreeLevel = level => (
+export const setEducationDegreeLevel = (level: string): Action => (
   { type: SET_EDUCATION_DEGREE_LEVEL, payload: level }
 );
 
 export const SET_EDUCATION_DEGREE_INCLUSIONS = 'SET_EDUCATION_DEGREE_INCLUSIONS';
-export const setEducationDegreeInclusions = degreeInclusions => (
+export const setEducationDegreeInclusions = (degreeInclusions: Object): Action => (
   { type: SET_EDUCATION_DEGREE_INCLUSIONS, payload: degreeInclusions }
 );
 
 export const SET_USER_PAGE_DIALOG_VISIBILITY = 'SET_USER_PAGE_DIALOG_VISIBILITY';
-export const setUserPageDialogVisibility = bool => (
+export const setUserPageDialogVisibility = (bool: boolean): Action => (
   { type: SET_USER_PAGE_DIALOG_VISIBILITY, payload: bool }
 );
 
 export const SET_SHOW_EDUCATION_DELETE_DIALOG = 'SET_SHOW_EDUCATION_DELETE_DIALOG';
-export const setShowEducationDeleteDialog = bool => (
+export const setShowEducationDeleteDialog = (bool: boolean): Action => (
   { type: SET_SHOW_EDUCATION_DELETE_DIALOG, payload: bool }
 );
 
 export const SET_SHOW_WORK_DELETE_DIALOG = 'SET_SHOW_WORK_DELETE_DIALOG';
-export const setShowWorkDeleteDialog = bool => (
+export const setShowWorkDeleteDialog = (bool: boolean): Action => (
   { type: SET_SHOW_WORK_DELETE_DIALOG, payload: bool }
 );
 
 export const SET_DELETION_INDEX = 'SET_DELETION_INDEX';
-export const setDeletionIndex = index => (
+export const setDeletionIndex = (index: number): Action => (
   { type: SET_DELETION_INDEX, payload: index }
 );
--- a/static/js/components/EducationDialog.js
+++ b/static/js/components/EducationDialog.js
@@ -1,3 +1,4 @@
+// @flow
 import React from 'react';
 import Button from 'react-mdl/lib/Button';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
@@ -9,13 +10,14 @@ import ProfileFormFields from '../util/ProfileFormFields';
 import { educationValidation } from '../util/validation';
 
 export default class EducationDialog extends ProfileFormFields {
-  constructor(props) {
+  constructor(props: Object) {
     super(props);
     this.educationLevelLabels = {};
     this.educationLevelOptions.forEach(level => {
       this.educationLevelLabels[level.value] = level.label;
     });
   }
+  educationLevelLabels: Object;
 
   static propTypes = {
     open:           React.PropTypes.bool,
@@ -24,7 +26,7 @@ export default class EducationDialog extends ProfileFormFields {
     showLevelForm:  React.PropTypes.bool,
   };
 
-  clearEducationEdit = () => {
+  clearEducationEdit: Function = (): void => {
     const {
       setEducationDialogVisibility,
       setEducationDegreeLevel,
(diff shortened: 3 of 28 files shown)


static/js/util/validation.js
```diff
@@ -135,7 +142,8 @@ export function employmentValidation(profile) {
     let requiredFields = handleNestedValidation(profile, nestedKeys, 'work_history');
     let errors = checkFieldPresence(profile, requiredFields, messages);
 
-    for (let [index, workHistory] of Object.entries(profile.work_history)) {
+    for (let entry: WorkEntry of Object.entries(profile.work_history)) {
+      let index = entry[0], workHistory = entry[1];
```
aliceriot: this is the only place I had to do something unfortunate to deal with the destructuring problem - it's ugly but not too bad, I think.

static/js/util/validation.js
```diff
@@ -252,12 +257,10 @@ export function validateMonth(string) {
...
 /**
  * Validate a year string is an integer and fits into YYYY
- * @param {String} string The input string
- * @returns {Number|undefined} The valid year if a valid year value or undefined if not valid
  */
-export function validateYear(string) {
+export function validateYear(string: string): ?number {
   let year = filterPositiveInt(string);
-  if (year === undefined) {
+  if (year == null) { // eslint-disable-line eqeqeq
```
noisecapella: Why did you switch to using double equals here and above?

aliceriot: It was necessary because of the return type of `filterPositiveInt`, which was `?number`. the `?number` type is a shorthand for `number|null|void`, so the typechecker was complaining about how we weren't checking for the case when `year` is `null` in addition to the case in which is undefined.

I just checked, and if I change the return type of `filterPositiveInt` from `?number` to `number|null` I think we can keep this as it was.

aliceriot: I changed the return type and I was able to revert these changes (back to just `=== undefined`)

aliceriot: oh and for context, `null == undefined`

package.json
```diff
@@ -67,6 +71,7 @@
     "postinstall": "./webpack_if_prod.sh",
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",
     "test": "./js_test.sh",
-    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --compilers js:babel-register --require static/js/global_init.js 'static/**/*/*_test.js'"
+    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --compilers js:babel-register --require static/js/global_init.js 'static/**/*/*_test.js'",
+    "flow": "flow; test $? -eq 0 -o $? -eq 2"
```
noisecapella: What's the `test` for? I've noticed that there's a zero return value even when flow spots errors

static/js/actions/ui.js
```diff
@@ -1,82 +1,85 @@
...
 export const SET_WORK_DIALOG_INDEX = 'SET_WORK_DIALOG_INDEX';
-export const setWorkDialogIndex = index => (
+export const setWorkDialogIndex = (index: number): Action => (
   { type: SET_WORK_DIALOG_INDEX, payload: index }
 );
 
 // dashboard actions
 export const TOGGLE_DASHBOARD_EXPANDER = 'TOGGLE_DASHBOARD_EXPANDER';
-export const toggleDashboardExpander = (courseId, newValue) => ({
+export const toggleDashboardExpander = (courseId: number, newValue: any): Action => ({
```
noisecapella: I think `newValue` is always a bool here

static/js/components/EducationDialog.js
```diff
@@ -8,21 +9,22 @@ import ProfileFormFields from '../util/ProfileFormFields';
 import { educationValidation } from '../util/validation';
 
 export default class EducationDialog extends ProfileFormFields {
-  constructor(props) {
+  constructor(props: {}) {
```
noisecapella: I think I prefer `props: Object` from a readability point of view. Is there any difference to flow?

aliceriot: I don't think so - some things have subtle differences, like `Number` vs `number`, but I think we can prefer `Object` without it making a difference.

static/js/actions/ui.js
aliceriot: cool, ty

package.json
aliceriot: I'm not sure exactly, I copied this from I think the 'getting started' portion of the flow docs.

static/js/actions/ui.js
aliceriot: there were a couple places where I had trouble tracking down exactly what type something should be - usually I tried to narrow it down (e.g. `string|number` or something), so if you spot something where we should be more specific that's great. I think we want as few things with type `any` as possible.

static/js/components/EducationDialog.js
noisecapella: I was thinking that `{}` might refer to an empty object specifically but I'm not sure. What's the difference between `Number` and `number`?

static/js/components/EducationDisplay.js
```diff
@@ -21,19 +22,19 @@ import {
...
-  }
+  };
 
-  openNewEducationForm = (level, index) => {
+  openNewEducationForm: Function = (level, index) => {
     openNewEducationForm.call(this, level, index);
   };
 
-  deleteEducationEntry = () => {
+  deleteEducationEntry: Function = () => {
```
noisecapella: Can you replace the `Function` here and below with more specific annotations, ie `deleteEducationEntry: (() => void) = () => {`

static/js/components/EducationDialog.js
aliceriot: my understanding is that `number` refers to anything declare with a literal (e.g. `var foo: number = 2`) whereas `Number` refers to things explicitly wrapped with the `Number` constructor (e.g. `var foo: Number = new Number(2)`)

aliceriot: I got some confusing errors before I realized the difference mattered :P

static/js/components/EducationDisplay.js
aliceriot: yeah, I think that's generally a good idea

aliceriot: idk, which of these do you think is preferable?

first, explicitly annotated the function type signature on the class property:

```js
deleteEducationEntry: (() => void) = () => {
  ...
};
```

or putting the return value annotation on the function literal and just using the `Function` annotation on the class property:

```js
deleteEducationEntry: Function = (): void => {
  ...
};
```

both are valid flow syntax. I'm a little inclined towards the latter because it's maybe a little more consistent, not sure.

static/js/components/EducationDialog.js
aliceriot: https://flowtype.org/docs/builtins.html#number

static/js/components/EducationDisplay.js
noisecapella: I agree, the latter looks better to me

static/js/components/EmploymentForm.js
```diff
@@ -100,7 +102,7 @@ class EmploymentForm extends ProfileFormFields {
     if ( ui.workHistoryEdit === true ) {
       let workHistoryRows = [];
       if ( !_.isUndefined(work_history) ) {
-        workHistoryRows = Object.entries(work_history).filter(([,entry]) =>
+        workHistoryRows = Object.entries(work_history).filter(([,entry]: WorkHistoryEntry[]) =>
```
noisecapella: Can you change the type here to `[string, WorkHistoryEntry]` which is more accurate?

static/js/components/EducationDisplay.js
aliceriot: cool, I like having explicit return types so I'll go through and add those to our class properties.

static/js/flow/declarations.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable no-unused-vars */
+declare var SETTINGS: Object;
```
noisecapella: Can we make this use the keys defined in `ui/views.py`?

static/js/reducers/ui.js
```diff
@@ -25,8 +26,24 @@ import {
 } from '../actions/ui';
 import { HIGH_SCHOOL, ASSOCIATE, BACHELORS, MASTERS, DOCTORATE } from '../constants';
 import { calculateDegreeInclusions } from '../util/util';
+import type { Action } from '../flow/generalTypes';
 
-export const INITIAL_UI_STATE = {
+export type UIState = {
```
noisecapella: Would be nice to combine this and `INITIAL_UI_STATE` below somehow, not sure if that's something flow supports right now though

static/js/util/ProfileFormFields.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
noisecapella: Is there another way to specify which files we want to cover? I'm fine with this for now but if we end up running this on most JS files we have, it would be nice to use a exclude instead if that's supported by flow

static/js/reducers/ui.js
aliceriot: yeah, I was hoping for something that could just take an object literal and return a named type from it, but it doesn't seem like that's supported yet.

static/js/util/ProfileFormFields.js
aliceriot: I think so, from [this](https://flowtype.org/docs/new-project.html#_) it looks like we can pass a `--all` command line flag to have it look at all JS files, and we could combine that with some `exclude` rules to avoid checking things we'd rather not bother with typechecking (for instance the dependencies that are checked-in in `static/js/public` for the homepage).

static/js/components/EmploymentForm.js
aliceriot: oh yeah, that's probs better

static/js/util/validation.js
noisecapella: It looks like you can simplify this to `let [index, workHistory] = entry` at least

static/js/util/validation.js
```diff
@@ -155,7 +163,7 @@ export function employmentValidation(profile) {
   }
 }
 
-export function employmentUiValidation(profile, ui) {
+export function employmentUiValidation(profile: Profile, ui: Object): ValidationErrors {
```
noisecapella: It looks like you can use `UIState` here

static/js/util/validation.js
```diff
@@ -269,11 +272,9 @@ export function validateYear(string) {
 
 /**
  * Returns a function which merges the results of the given functions on a set of arguments
- * @param validators {Array} A list of validator functions taking (profile, ui)
- * @returns {function()} A function taking (profile, ui) which combines the outputs of the validators
  */
-export function combineValidators(...validators) {
-  return (...args) => _.merge({}, ...validators.map(
+export function combineValidators(...validators: Function[]): Function {
```
noisecapella: I've been playing around with this function trying to make the type checking more specific here. Here's what I have:

    type ValidatorFunc = (profile: Profile, ui: UIState) => ValidationErrors;
    /**
     * Returns a function which merges the results of the given functions on a set of arguments
     */
    export function combineValidators(...validators: ValidatorFunc[]) : ValidatorFunc {
      return (profile: Profile, ui: UIState) => _.merge({}, ...validators.map(
        validator => validator(profile, ui)
      ));
    }

But I can swap `profile` and `ui` inside `validator` and it doesn't complain. It will complain if I provide a number to the `ui` argument, but if I pass a number to the `profile` argument it doesn't error. :sob:

static/js/flow/declarations.js
aliceriot: sure

static/js/util/util.js
```diff
@@ -141,11 +143,8 @@ export function makeProfileProgressDisplay(active) {
 /* eslint-disable camelcase */
 /**
  * Generate new education object 
- *
- * @param {String} level The select degree level
- * @returns {Object} New empty education object
  */
-export function generateNewEducation(level) {
+export function generateNewEducation(level: string): Object {
```
noisecapella: Can you set the return type here and for the work history equivalent below?

static/js/util/profile_edit.js
```diff
@@ -440,13 +410,10 @@ export function boundDateField(keySet, label, omitDay) {
...
  * Validates the profile then PATCHes the profile if validation succeeded.
- *
- * @param validator {func} The function used to validate profile
- * @param isLastStep {bool} If true, this is the last tab in the profile
- * @returns {Promise} A promise which resolves with the new profile if validation and PATCH succeeded,
+ * Returns a promise which resolves with the new profile if validation and PATCH succeeded,
  * or rejects if either failed
  */
-export function saveProfileStep(validator, isLastStep=false) {
+export function saveProfileStep(validator: Function, isLastStep: boolean = false): Promise {
```
noisecapella: Can we make a type for validator functions and use that in place of `Function` here?

aliceriot: yeah, def.

noisecapella: Also can you use `Promise<Profile>` here? It looks like it handles the resolve case but for the reject case it will just return `any`, so we can't specify a type for that here

static/js/util/profile_edit.js
```diff
@@ -131,15 +123,8 @@ export function boundTextField(keySet, label) {
...
- *
- * @param keySet {String[]} Path to the field
- * @param label {String} Label for the field
- * @param options {Object[]} A list of options for the select field
- * @param onChange {func} Handler which is called when the state changes or is cleared.
- * Takes the updated profile as argument
- * @returns {ReactElement}
  */
-export function boundSelectField(keySet, label, options, onChange) {
+export function boundSelectField(keySet: string[], label: string, options: Object[], onChange: Function|void) {
```
noisecapella: Can you specify a type for `onChange`? It should take only a profile as its argument

static/js/util/profile_edit.js
```diff
@@ -23,13 +24,8 @@ import {
...
  * to update radio buttons.
  * pass in the name (used as placeholder), key for profile, and the options.
- *
- * @param keySet {String[]} Path to the field
- * @param label {String} Label for the field
- * @param options {Object[]} A list of options for the select field
- * @returns {ReactElement}
  */
-export function boundRadioGroupField(keySet, label, options) {
+export function boundRadioGroupField(keySet: string[], label: string, options: Object[]): React$Element {
```
noisecapella: Can we use `Option[]` in place of `Object[]`?

static/js/util/courseList.js
```diff
@@ -23,11 +24,8 @@ function asPercent(number) {
 
 /**
  * Determine React elements for the UI given a course status
- * @param {object} course A course coming from the dashboard
- * @param {moment} now The current time
- * @returns {ReactElement} Some React element or string to display for course status
  */
-export function makeCourseStatusDisplay(course, now = moment()) {
+export function makeCourseStatusDisplay(course: Object, now: moment = moment()): string|React$Element {
```
noisecapella: It would be nice to specify a type for course and run here though I'm happy doing that in another PR

package.json
noisecapella: From [their source](https://github.com/facebook/flow/blob/b352b4c41283c1cb109ee2e8f6ef604ad4ac381b/src/common/flowExitStatus.ml#L63-L86) it looks like `2` is a type error so it looks like it was trying to silence that error code. Since we want type errors to fail the build we should remove the `test` portion here.

aliceriot:
@noisecapella updated!

package.json
aliceriot: I think we might want to just do `flow check` here, that seems to be a bit faster (flow starts a server and stuff)

noisecapella:
LGTM, just need to rebase :+1:

static/js/util/validation.js
aliceriot: ah indeed!

Issue #483, opened 2016-06-06
Users API should return 404 for users with privacy setting set to "private"


pdpinch:

amir-qayyum-khan:
@pdpinch sorry i forget abt Users API. Can you give me api url?

pdpinch:
The web page view is at /users/ but the API is /api/v0/profiles/<username>

amir-qayyum-khan:
thanks

amir-qayyum-khan:
@pdpinch raises PermissionDenied. Are you ok with it?

<img width="975" alt="screen shot 2016-06-08 at 3 20 38 pm" src="https://cloud.githubusercontent.com/assets/10431250/15891171/d5de0766-2d8c-11e6-844a-02ef636585da.png">

pdpinch:
No, if we throw a 500 due to the exception, or return a 403, it leaks information that the account exists. 

We need to return a 404.

pdpinch:
@amir-qayyum-khan what's the status of this issue?

amir-qayyum-khan:
@pdpinch made this PR https://github.com/mitodl/micromasters/pull/546

amir-qayyum-khan:
fixed via https://github.com/mitodl/micromasters/pull/546

PR #546, opened 2016-06-15
Raised 404 from users API for users with privacy setting set to "private" 

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/483#issuecomment-225736543

#### What's this PR do?
- Raised 404 exception when user wants to access someones profile whose privacy mode is set tp private

- see ```profiles/views.py```

- Open api for user whose [**account_privacy = private**](http://ip:8079/api/v0/profiles/{user_name}/)

@pdpinch  @aliceriot  @giocalitri 

- In case **account_privacy = private**:
screen shot 2016-06-15 at 5 35 28 pm

- In case **account_privacy = public_to_mm**:
screen shot 2016-06-15 at 5 38 41 pm

- In case **account_privacy = public**:
screen shot 2016-06-15 at 5 40 33 pm

### DIFF
--- a/profiles/permissions.py
+++ b/profiles/permissions.py
@@ -1,10 +1,14 @@
 """
 Permission classes for profiles
 """
+from django.http import Http404
+from rest_framework.generics import get_object_or_404
 from rest_framework.permissions import (
     BasePermission,
     SAFE_METHODS,
 )
+
+from profiles.models import Profile
 
 
 class CanEditIfOwner(BasePermission):
@@ -20,3 +24,35 @@ class CanEditIfOwner(BasePermission):
             return True
 
         return profile.user == request.user
+
+
+class CanSeeIfNotPrivate(BasePermission):
+    """
+    Only owner of a profile can view and edit their profile.
+    Verified micromaster users can view other profiles to limited view only if
+    account_privacy of profile is set to public public_to_mm
+    """
+
+    def has_permission(self, request, view):
+        """
+        Implementation of the permission class.
+        """
+        profile = get_object_or_404(Profile, user__social_auth__uid=view.kwargs['user'])
+
+        if request.user == profile.user:
+            return True
+
+        # private profiles
+        if profile.account_privacy == Profile.PRIVATE:
+            raise Http404
+        elif profile.account_privacy == Profile.PUBLIC_TO_MM:
+            # anonymous user accessing profiles.
+            if request.user.is_anonymous():
+                raise Http404
+            # user who is not a micromaster verified user is accessing profiles.
+            elif not request.user.profile.verified_micromaster_user:
+                raise Http404
+        elif profile.account_privacy not in [Profile.PRIVATE, Profile.PUBLIC_TO_MM, Profile.PUBLIC]:
+            raise Http404
+
+        return True
--- a/profiles/permissions_test.py
+++ b/profiles/permissions_test.py
@@ -1,14 +1,22 @@
 """
 Tests for profile permissions
 """
-
 from mock import Mock
+from django.http import Http404
 from django.db.models.signals import post_save
 from django.test import TestCase
 from factory.django import mute_signals
 
-from profiles.factories import ProfileFactory, UserFactory
-from profiles.permissions import CanEditIfOwner
+from backends.edxorg import EdxOrgOAuth2
+from profiles.factories import (
+    ProfileFactory,
+    UserFactory,
+)
+from profiles.models import Profile
+from profiles.permissions import (
+    CanEditIfOwner,
+    CanSeeIfNotPrivate,
+)
 
 
 # pylint: disable=no-self-use
@@ -47,6 +55,135 @@ class CanEditIfOwnerTests(TestCase):
         other_user = UserFactory.create()
         for method in ('POST', 'PATCH', 'PUT'):
             with mute_signals(post_save):
-                profile = ProfileFactory.create()
+                profile = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
             request = Mock(method=method, user=other_user)
             assert not perm.has_object_permission(request, None, profile)
+
+
+# pylint: disable=no-self-use
+class CanSeeIfNotPrivateTests(TestCase):
+    """
+    Tests for CanSeeIfNotPrivate permissions
+    """
+
+    def setUp(self):
+        super(CanSeeIfNotPrivateTests, self).setUp()
+        with mute_signals(post_save):
+            self.other_user = other_user = UserFactory.create()
+            username = "{}_edx".format(other_user.username)
+            social_auth = other_user.social_auth.create(
+                provider=EdxOrgOAuth2.name,
+                uid=username
+            )
+            self.other_user_id = social_auth.uid
+            ProfileFactory.create(user=other_user, verified_micromaster_user=False)
+
+        with mute_signals(post_save):
+            self.profile_user = profile_user = UserFactory.create()
+            username = "{}_edx".format(profile_user.username)
+            social_auth = profile_user.social_auth.create(
+                provider=EdxOrgOAuth2.name,
+                uid=username
+            )
+            self.profile_user_id = social_auth.uid
+
+    def test_cant_view_if_privacy_is_private(self):
+        """
+        Users are not supposed to view private profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PRIVATE)
+
+        request = Mock(user=self.other_user)
+        view = Mock(kwargs={'user': self.profile_user_id})
+
+        with self.assertRaises(Http404):
+            perm.has_permission(request, view)
+
+    def test_cant_view_if_anonymous_user(self):
+        """
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PUBLIC_TO_MM)
+
+        request = Mock(user=Mock(is_anonymous=Mock(return_value=True)))
+        view = Mock(kwargs={'user': self.profile_user_id})
+
+        with self.assertRaises(Http404):
+            perm.has_permission(request, view)
+
+    def test_cant_view_if_non_verified_mm_user(self):
+        """
+        Non verified micromaster users are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PUBLIC_TO_MM)
+
+        request = Mock(user=self.other_user)
+        view = Mock(kwargs={'user': self.profile_user_id})
+
+        with self.assertRaises(Http404):
+            perm.has_permission(request, view)
+
+    def test_cant_view_if_privacy_weird(self):
+        """
+        Users can not open profiles with ambiguous account_privacy settings.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy='weird_setting')
+
+        request = Mock(user=self.other_user)
+        view = Mock(kwargs={'user': self.profile_user_id})
+
+        with self.assertRaises(Http404):
+            perm.has_permission(request, view)
+
+    def test_can_view_own_profile(self):
+        """
+        Users are allowed to view their own profile.
+        """
+        perm = CanSeeIfNotPrivate()
+        request = Mock(user=self.other_user)
+        view = Mock(kwargs={'user': self.other_user_id})
+
+        assert perm.has_permission(request, view) is True
+
+    def test_users_can_view_public_profile(self):
+        """
+        Users are allowed to view public profile.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PUBLIC)
+
+        request = Mock(user=self.other_user)
+        view = Mock(kwargs={'user': self.profile_user_id})
+        assert perm.has_permission(request, view) is True
+
+    def test_can_view_if_verified_mm_user(self):
+        """
+        Verified MM users are allowed to view public_to_mm profile.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PUBLIC_TO_MM)
+
+        with mute_signals(post_save):
+            verified_user = UserFactory.create()
+            username = "{}_edx".format(verified_user.username)
+            verified_user.social_auth.create(
+                provider=EdxOrgOAuth2.name,
+                uid=username
+            )
+            ProfileFactory.create(user=verified_user, verified_micromaster_user=True)
+
+        request = Mock(user=verified_user)
+        view = Mock(kwargs={'user': self.profile_user_id})
+        assert perm.has_permission(request, view) is True
--- a/profiles/views.py
+++ b/profiles/views.py
@@ -9,16 +9,18 @@ from profiles.models import Profile
 from profiles.serializers import (
     ProfileSerializer,
     ProfileLimitedSerializer,
-    ProfilePrivateSerializer,
 )
-from profiles.permissions import CanEditIfOwner
+from profiles.permissions import (
+    CanEditIfOwner,
+    CanSeeIfNotPrivate,
+)
 
 
 class ProfileViewSet(RetrieveModelMixin, UpdateModelMixin, GenericViewSet):
     """API for the Program collection"""
     # pylint: disable=too-many-return-statements
 
-    permission_classes = (CanEditIfOwner, )
+    permission_classes = (CanEditIfOwner, CanSeeIfNotPrivate, )
     lookup_field = 'user__social_auth__uid'
     lookup_url_kwarg = 'user'
     queryset = Profile.objects.all()
@@ -26,7 +28,6 @@ class ProfileViewSet(RetrieveModelMixin, UpdateModelMixin, GenericViewSet):
     # possible serializers
     serializer_class_owner = ProfileSerializer
     serializer_class_limited = ProfileLimitedSerializer
-    serializer_class_private = ProfilePrivateSerializer
 
     def get_serializer_class(self):
         """
@@ -34,24 +35,14 @@ class ProfileViewSet(RetrieveModelMixin, UpdateModelMixin, GenericViewSet):
         """
         profile = self.get_object()
 
-        # Case #1: Owner of the profile
+        # Owner of the profile
         if self.request.user == profile.user:
             return self.serializer_class_owner
-        # Case #2: Profile is private
-        elif profile.account_privacy == Profile.PRIVATE:
-            return self.serializer_class_private
-        # Case #3: Profile is public
+        # Profile is public
         elif profile.account_privacy == Profile.PUBLIC:
             return self.serializer_class_limited
-        # Case #4: Profile is public to mm verified users only
+        # Profile is public to mm verified users only
         elif profile.account_privacy == Profile.PUBLIC_TO_MM:
-            # Case #3: anonymous user
-            if self.request.user.is_anonymous():
-                # the profile at this point is not public so the anonymous users can see only basic info
-                return self.serializer_class_private
-            # Case #4: the user is not a micromaster verified user
-            elif not self.request.user.profile.verified_micromaster_user:
-                return self.serializer_class_private
             return self.serializer_class_limited
-        # Case #5: this should never happen, but just in case
-        return self.serializer_class_private
+        # this should never happen, but just in case
+        return self.serializer_class_limited
--- a/profiles/views_test.py
+++ b/profiles/views_test.py
@@ -11,6 +11,7 @@ from django.test import TestCase
 from factory.django import mute_signals
 from rest_framework.status import (
     HTTP_405_METHOD_NOT_ALLOWED,
+    HTTP_404_NOT_FOUND
 )
 
 from backends.edxorg import EdxOrgOAuth2
@@ -19,7 +20,6 @@ from profiles.models import Profile
 from profiles.permissions import CanEditIfOwner
 from profiles.serializers import (
     ProfileLimitedSerializer,
-    ProfilePrivateSerializer,
     ProfileSerializer,
 )
 from profiles.views import ProfileViewSet
@@ -73,7 +73,7 @@ class ProfileTests(TestCase):
         Make sure check_object_permissions is called at some point so the permissions work correctly
         """
         with mute_signals(post_save):
-            ProfileFactory.create(user=self.user1)
+            ProfileFactory.create(user=self.user1, account_privacy=Profile.PUBLIC)
         self.client.force_login(self.user1)
 
         with patch.object(ProfileViewSet, 'check_object_permissions', autospec=True) as check_object_permissions:
@@ -128,21 +128,21 @@ class ProfileTests(TestCase):
         An anonymous user gets user's public_to_mm profile.
         """
         with mute_signals(post_save):
-            profile = ProfileFactory.create(user=self.user2, account_privacy=Profile.PUBLIC_TO_MM)
+            ProfileFactory.create(user=self.user2, account_privacy=Profile.PUBLIC_TO_MM)
         self.client.logout()
         resp = self.client.get(self.url2)
-        assert resp.json() == ProfilePrivateSerializer().to_representation(profile)
+        assert resp.status_code == HTTP_404_NOT_FOUND
 
     def test_mm_user_get_public_to_mm_profile(self):
         """
         An unverified mm user gets user's public_to_mm profile.
         """
         with mute_signals(post_save):
-            profile = ProfileFactory.create(user=self.user2, account_privacy=Profile.PUBLIC_TO_MM)
+            ProfileFactory.create(user=self.user2, account_privacy=Profile.PUBLIC_TO_MM)
             ProfileFactory.create(user=self.user1, verified_micromaster_user=False)
         self.client.force_login(self.user1)
         resp = self.client.get(self.url2)
-        assert resp.json() == ProfilePrivateSerializer().to_representation(profile)
+        assert resp.status_code == HTTP_404_NOT_FOUND
 
     def test_vermm_user_get_public_to_mm_profile(self):
         """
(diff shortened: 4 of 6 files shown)


profiles/views.py
```diff
@@ -55,3 +56,36 @@ def get_serializer_class(self):
             return self.serializer_class_limited
         # Case #5: this should never happen, but just in case
         return self.serializer_class_private
+
+    def get_object(self):
```
giocalitri: why did you override `get_object`?

amir-qayyum-khan: because line no 83 to 86

https://github.com/mitodl/micromasters/pull/546/files#diff-30ff6c0b423aeb83e6ee53260122cac7R83

If we dont use this then instead of 404 it throws permission denied exception.

giocalitri: Ok, but I think this job needs to be done in the Permission classes and not here.

profiles/views_test.py
```diff
@@ -163,7 +163,7 @@ def test_anonym_user_get_private_profile(self):
             profile = ProfileFactory.create(user=self.user2, account_privacy=Profile.PRIVATE)
         self.client.logout()
         resp = self.client.get(self.url2)
-        assert resp.json() == ProfilePrivateSerializer().to_representation(profile)
+        assert resp.status_code == 404
```
giocalitri: please uses
```
from rest_framework import status
...
assert resp.status_code == status.HTTP_404_NOT_FOUND
```

profiles/views.py
amir-qayyum-khan: Done @giocalitri

amir-qayyum-khan:
Done @giocalitri

noisecapella:
We should get rid of `ProfilePrivateSerializer` since we don't need it anymore

profiles/permissions.py
```diff
@@ -19,4 +20,9 @@ def has_object_permission(self, request, view, profile):
         if request.method in SAFE_METHODS:
             return True
 
+        if profile.account_privacy == Profile.PRIVATE:
+            # disbaled permission denied for profile with account_privacy = private,
+            # so that we can raise 404 exception.
+            return True
+
```
giocalitri: I do not think this is the right place either: this check should go in a specific new permission.

amir-qayyum-khan:
@noisecapella there are some places where we are using ``ProfilePrivateSerializer `` https://github.com/mitodl/micromasters/blob/master/profiles/views.py#L49-L57

giocalitri:
@amir-qayyum-khan that case is just a default one and should be changed to raising a 404 as well.
That check should happen in the permission class as well.

amir-qayyum-khan:
Done @noisecapella  @giocalitri

giocalitri:
@amir-qayyum-khan : I do not think we understood each other:

The `get_serializer_class` should not be in charge of enforcing permissions. That's a duty of the permission classes.

amir-qayyum-khan:
@giocalitri you mean that we should not raise 404 error in get_serializer_class. Instead we should raise these error in permission classes?

amir-qayyum-khan:
Done @giocalitri

profiles/permissions.py
```diff
@@ -1,10 +1,13 @@
 """
 Permission classes for profiles
 """
+from django.http import Http404
+from rest_framework.generics import get_object_or_404
 from rest_framework.permissions import (
     BasePermission,
     SAFE_METHODS,
 )
+from .models import Profile
```
giocalitri: please use full paths for imports

profiles/permissions.py
```diff
@@ -20,3 +23,38 @@ def has_object_permission(self, request, view, profile):
             return True
 
         return profile.user == request.user
+
+
+class CanEditIfPrivate(BasePermission):
```
giocalitri: I think the name is misleading: it should be something like `CanSeeIfNotPrivate`

profiles/permissions.py
```diff
@@ -20,3 +23,38 @@ def has_object_permission(self, request, view, profile):
...
+        if request.user == profile.user:
+            return True
+
+        # private profiles
+        if profile.account_privacy == Profile.PRIVATE:
+            raise Http404
+        elif profile.account_privacy == Profile.PUBLIC_TO_MM:
+            # anonymous user accessing profiles.
+            if request.user.is_anonymous():
+                raise Http404
```
giocalitri: I am not sure about this case: @pdpinch do you think a "Private to MM" user should not be visible at all from anonymous users?

profiles/permissions.py
```diff
@@ -20,3 +23,38 @@ def has_object_permission(self, request, view, profile):
...
+            # user who is not a micromaster verified user is accessing profiles.
+            elif not request.user.profile.verified_micromaster_user:
+                raise Http404
```
giocalitri: @pdpinch same here for users who are not verified

profiles/permissions_test.py
```diff
@@ -2,13 +2,15 @@
 Tests for profile permissions
 """
 
-from mock import Mock
+from unittest.mock import MagicMock, patch, Mock
```
giocalitri: isn't just `from mock import ...`?

giocalitri: BTW, if you have multiple imports, please one per line

profiles/permissions_test.py
```diff
@@ -2,13 +2,15 @@
...
+from django.http import Http404
 from django.db.models.signals import post_save
 from django.test import TestCase
 from factory.django import mute_signals
 
 from profiles.factories import ProfileFactory, UserFactory
-from profiles.permissions import CanEditIfOwner
+from profiles.models import Profile
+from profiles.permissions import CanEditIfOwner, CanEditIfPrivate
```
giocalitri: One import per line

profiles/permissions_test.py
```diff
@@ -47,6 +49,31 @@ def test_cant_edit_if_not_owner(self):
...
+
+# pylint: disable=no-self-use
+class CanEditIfPrivateTests(TestCase):
+    """
+    Tests for CanEditIfPrivate permissions
+    """
+
+    def test_cant_edit_if_privacy_is_private(self):
+        """
+        Users are not to view private profiles.
```
giocalitri: `Users are not supposed to...`

profiles/permissions_test.py
```diff
@@ -47,6 +49,31 @@ def test_cant_edit_if_not_owner(self):
...
+        for method in ('POST', 'PATCH', 'PUT'):
+            with mute_signals(post_save):
+                profile_private = ProfileFactory.create(account_privacy=Profile.PRIVATE)
+
+            request = Mock(method=method, user=other_user)
+
+            with patch('rest_framework.generics.get_object_or_404', profile_private):
+                view = MagicMock(name='kwargs', return_value={'user': 'test'})
+                with self.assertRaises(Http404):
+                    perm.has_permission(request, view)
```
giocalitri: This test does not seem sufficient: how is the other permission class tested?

profiles/views.py
```diff
@@ -9,49 +9,37 @@
 from profiles.serializers import (
     ProfileSerializer,
     ProfileLimitedSerializer,
-    ProfilePrivateSerializer,
 )
-from profiles.permissions import CanEditIfOwner
+from profiles.permissions import CanEditIfOwner, CanEditIfPrivate
```
giocalitri: One import per line

giocalitri:
I think we are almost there: the only doubt I have can be answered only by @pdpinch

profiles/permissions.py
pdpinch: Yes. The use case is someone who wants to pursue a degree but doesn't want their employer to know about it. We don't want to leak the fact that they have an account at all.

profiles/permissions.py
pdpinch: Is it the requesting user who isn't verified, or is it the user of the account being viewed?

If the requesting user isn't verified, they should be treated the same as an anonymous user. 

I don't think it matters if the account being requested belongs to an unverified user.

profiles/permissions.py
giocalitri: Yes, but this case is not about the profile `Private`, but `Public to MM Users`...

pdpinch: Sorry I misunderstood. 

Still, an anonymous user hasn't agreed to the terms of service, or invested in the community. They shouldn't get to know who is a member. 

I'm trying to think of a use case where an anonymous user needs minimal access to a profile -- perhaps to see avatars in the forum -- but that won't be available to an anonymous user either.

giocalitri: OK @amir-qayyum-khan the code is fine as is

profiles/permissions.py
giocalitri: @amir-qayyum-khan the code is fine as is

amir-qayyum-khan: @pdpinch  it is requesting user who isn't verified.

amir-qayyum-khan:
Done @giocalitri

profiles/permissions.py
```diff
@@ -20,3 +23,38 @@ def has_object_permission(self, request, view, profile):
...
+    Dont allow users to view private profiles.
+    Also dont allow anonymous users or non verified mm user to view profiles with
+    account_privacy is set to public public_to_mm.
+    """
+
+    def has_permission(self, request, view):
+        """
+        Only owner of a profile can view and edit their profile.
+        Verified micromaster users can view other profiles to limited view only if
+        account_privacy of profile is set to public public_to_mm
```
giocalitri: this description is wrong and not about this permission class.
I do not think you need to put anything this long, given that the description is in the class docstring.

It will be enough something like "Implementation of the permission class"

profiles/permissions_test.py
```diff
@@ -47,6 +59,73 @@ def test_cant_edit_if_not_owner(self):
...
+        with mute_signals(post_save):
+            self.other_user = UserFactory.create()
+            username = "{}_edx".format(self.other_user.username)
+            self.other_user.social_auth.create(
+                provider=EdxOrgOAuth2.name,
+                uid=username
+            )
+            ProfileFactory.create(user=self.other_user, verified_micromaster_user=False)
+
+    def test_cant_edit_if_privacy_is_private(self):
```
giocalitri: I am confused with this test: the name says something and the description something else. What is this test supposed to do?

profiles/permissions_test.py
```diff
@@ -47,6 +59,73 @@ def test_cant_edit_if_not_owner(self):
...
+
+            with patch('rest_framework.generics.get_object_or_404', profile_private):
+                view = MagicMock(name='kwargs', return_value={'user': 'test'})
+                with self.assertRaises(Http404):
+                    perm.has_permission(request, view)
+
+    def test_cant_edit_if_anonymous_user(self):
+        """
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
```
giocalitri: same confusion as for the other test

profiles/permissions_test.py
```diff
@@ -47,6 +59,73 @@ def test_cant_edit_if_not_owner(self):
...
+
+            with patch('rest_framework.generics.get_object_or_404', profile_private):
+                view = MagicMock(name='kwargs', return_value={'user': 'test'})
+                with self.assertRaises(Http404):
+                    perm.has_permission(request, view)
+
+    def test_cant_edit_if_non_verified_mm_user(self):
+        """
+        Non verified micromaster users are not supposed to view public_to_mm profiles.
+        """
```
giocalitri: again confused, please explain also here

amir-qayyum-khan:
@giocalitri can you review

profiles/permissions.py
giocalitri: please swap the docstrings: the short one to `has_permission` and the long one to the class docstring

amir-qayyum-khan: Done,

profiles/permissions.py
```diff
@@ -1,10 +1,13 @@
 """
 Permission classes for profiles
 """
+from django.http import Http404
+from rest_framework.generics import get_object_or_404
 from rest_framework.permissions import (
     BasePermission,
     SAFE_METHODS,
 )
+from profiles.models import Profile
```
noisecapella: Can you add an empty line before here? Imports should be grouped into three groups with a empty line separating each group: https://www.python.org/dev/peps/pep-0008/#imports

profiles/permissions.py
```diff
@@ -20,3 +23,35 @@ def has_object_permission(self, request, view, profile):
...
+
+class CanSeeIfNotPrivate(BasePermission):
+    """
+    Only owner of a profile can view and edit their profile.
+    Verified micromaster users can view other profiles to limited view only if
+    account_privacy of profile is set to public public_to_mm
+    """
+
+    def has_permission(self, request, view):
+        "Implementation of the permission class"
```
noisecapella: All docstrings should have three quotes on each side: https://www.python.org/dev/peps/pep-0008/#documentation-strings

profiles/permissions.py
```diff
@@ -20,3 +23,35 @@ def has_object_permission(self, request, view, profile):
...
+
+        filter_kwargs = {'user__social_auth__uid': view.kwargs['user']}
+        profile = get_object_or_404(view.queryset, **filter_kwargs)
```
noisecapella: Why not just use `user__social_auth__uid` directly here instead of putting it in `filter_kwargs`?

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+                profile = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
             request = Mock(method=method, user=other_user)
             assert not perm.has_object_permission(request, None, profile)
+
+
+# pylint: disable=no-self-use
+class CanSeeIfNotPrivateTests(TestCase):
+    """
+    Tests for CanEditIfPrivate permissions
```
noisecapella: `CanEditIfPrivate` should be `CanSeeIfNotPrivate`

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+        Users are not supposed to view private profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+
+        for method in ('POST', 'PATCH', 'PUT'):
+            with mute_signals(post_save):
+                profile_private = ProfileFactory.create(account_privacy=Profile.PRIVATE)
+
+            request = Mock(method=method, user=self.other_user)
+            result = Mock(method='get_object_or_404', return_value=profile_private)
```
noisecapella: Why is this being mocked? It looks like we have a user and profile marked private in the database already

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+
+            with patch('rest_framework.generics._get_object_or_404', result):
+                view = MagicMock(name='kwargs', return_value={'user': 'test'})
```
noisecapella: Please reference the username from the model here. Either get the username from `self.other_user.social_auth` or store it in a field in the `setUp` method

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+        perm = CanSeeIfNotPrivate()
+        for method in ('POST', 'PATCH', 'PUT'):
+            with mute_signals(post_save):
+                profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
+            request = Mock(method=method, user=Mock(method='is_anonymous', return_value=True))
+            result = Mock(method='get_object_or_404', return_value=profile_private)
+
+            with patch('rest_framework.generics._get_object_or_404', result):
+                view = MagicMock(name='kwargs', return_value={'user': 'test'})
```
noisecapella: Please use a variable for the username as described above, here and for each instance of `'test'` below

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+            )
+            self.other_user_profile = ProfileFactory.create(user=other_user, verified_micromaster_user=False)
+
+    def test_cant_view_if_privacy_is_private(self):
+        """
+        Users are not supposed to view private profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+
+        for method in ('POST', 'PATCH', 'PUT'):
```
noisecapella: It does not look like `request.method` is used anywhere in the function, so we shouldn't need to pass it in with the mock

profiles/permissions_test.py
noisecapella: `method='get_object_or_404'` doesn't have any meaning here. Since you are mocking a function only `return_value` matters

profiles/permissions.py
noisecapella: Instead of using `view.queryset` why not just use `get_object_or_404(Profile, user__social_auth__uid=view.kwargs['user'])?

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+    def test_cant_view_if_anonymous_user(self):
+        """
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        for method in ('POST', 'PATCH', 'PUT'):
+            with mute_signals(post_save):
+                profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
+            request = Mock(method=method, user=Mock(method='is_anonymous', return_value=True))
```
noisecapella: The `user` mock is defined incorrectly here. `method` doesn't mean anything and `return_value` only matters when using the mock as a function. This only works because `Mock()` evaluates to `True`. You need a third mock for `is_anonymous` to do this correctly.

profiles/permissions_test.py
```diff
@@ -47,6 +59,127 @@ def test_cant_edit_if_not_owner(self):
...
+            result = Mock(method='get_object_or_404', return_value=profile_private)
+
+            with patch('rest_framework.generics._get_object_or_404', result):
+                view = MagicMock(name='kwargs', return_value={'user': 'test'})
+                with self.assertRaises(Http404):
+                    perm.has_permission(request, view)
+
+    def test_cant_view_if_non_verified_mm_user(self):
+        """
+        Non verified micromaster users are not supposed to view public_to_mm profiles.
```
noisecapella: Please add a test for the case where a user is a verified micromaster user

amir-qayyum-khan:
Done @giocalitri @noisecapella

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+            with self.assertRaises(Http404):
+                perm.has_permission(request, view)
+
+    def test_cant_view_if_anonymous_user(self):
+        """
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
```
giocalitri: this should be `profile_public_mm` and not `profile_private`

giocalitri: or just use a generic name

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+            with self.assertRaises(Http404):
+                perm.has_permission(request, view)
+
+    def test_cant_view_if_non_verified_mm_user(self):
+        """
+        Non verified micromaster users are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
```
giocalitri: name of the variable

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+            with self.assertRaises(Http404):
+                perm.has_permission(request, view)
+
+    def test_cant_view_if_privacy_weird(self):
+        """
+        Users can not open profiles with ambiguous account_privacy settings.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy='weird_setting')
```
giocalitri: name of the variable

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+        """
+        perm = CanSeeIfNotPrivate()
+        request = Mock(user=self.other_user)
+        view = MagicMock()
+        view.kwargs.side_effect = {'user': self.social_auth}
+        view.queryset = Profile.objects.all()
+        result = Mock(return_value=self.other_user_profile)
+
+        with patch('rest_framework.generics._get_object_or_404', result):
+            assert perm.has_permission(request, view)
```
giocalitri: this should be `assert perm.has_permission(request, view) is True`

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+            profile_public = ProfileFactory.create(account_privacy=Profile.PUBLIC)
+
+        request = Mock(user=self.other_user)
+        view = MagicMock()
+        view.kwargs.side_effect = {'user': self.social_auth}
+        view.queryset = Profile.objects.all()
+        result = Mock(return_value=profile_public)
+
+        with patch('rest_framework.generics._get_object_or_404', result):
+            assert perm.has_permission(request, view)
```
giocalitri: this should be `assert perm.has_permission(request, view) is True`

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+            ProfileFactory.create(user=verified_user, verified_micromaster_user=True)
+
+        request = Mock(user=verified_user)
+        view = MagicMock()
+        view.kwargs.side_effect = {'user': social_auth}
+        view.queryset = Profile.objects.all()
+        result = Mock(return_value=profile_public_to_mm)
+
+        with patch('rest_framework.generics._get_object_or_404', result):
+            assert perm.has_permission(request, view)
```
giocalitri: this should be `assert perm.has_permission(request, view) is True`

giocalitri:
I have some minor comments, but I would like @noisecapella to take a second look

profiles/permissions.py
```diff
@@ -20,3 +24,35 @@ def has_object_permission(self, request, view, profile):
...
+    Only owner of a profile can view and edit their profile.
+    Verified micromaster users can view other profiles to limited view only if
+    account_privacy of profile is set to public public_to_mm
+    """
+
+    def has_permission(self, request, view):
+        """
+        Implementation of the permission class.
+        """
+        profile = get_object_or_404(view.queryset, user__social_auth__uid=view.kwargs['user'])
```
noisecapella: Can you replace `view.queryset` with `Profile`? It does not seem like `view.queryset` provides anything extra here and it makes it more difficult to mock

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+        perm = CanSeeIfNotPrivate()
+
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy=Profile.PRIVATE)
+
+        request = Mock(user=self.other_user)
+        result = Mock(return_value=profile_private)
+
+        with patch('rest_framework.generics._get_object_or_404', result):
+            view = MagicMock(name='kwargs', return_value={'user': self.social_auth})
```
noisecapella: view is not a function so return_value doesn't mean anything here. Also user should not be the social_auth model, it should be a username.

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
+        request = Mock(user=Mock(return_value=Mock(method='is_anonymous', return_value=True)))
+        result = Mock(return_value=profile_private)
+
+        with patch('rest_framework.generics._get_object_or_404', result):
+            view = MagicMock(name='kwargs', return_value={'user': self.social_auth})
```
noisecapella: Same problem here regarding mock of view

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
+        request = Mock(user=Mock(return_value=Mock(method='is_anonymous', return_value=True)))
+        result = Mock(return_value=profile_private)
+
+        with patch('rest_framework.generics._get_object_or_404', result):
```
noisecapella: I feel like we shouldn't need to mock this since the objects are in the database

profiles/permissions_test.py
```diff
@@ -47,6 +59,149 @@ def test_cant_edit_if_not_owner(self):
...
+
+    def test_cant_view_if_anonymous_user(self):
+        """
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            profile_private = ProfileFactory.create(account_privacy=Profile.PUBLIC_TO_MM)
+
+        request = Mock(user=Mock(return_value=Mock(method='is_anonymous', return_value=True)))
```
noisecapella: method does not have any significance to a mock. return_value only matters if user was a function, and it's not. The attribute itself must be called is_anonymous

profiles/permissions_test.py
amir-qayyum-khan: i am mocking this because one user is viewing profile of the other user and here we are testing permission not view it self. So we dont have reference of view

amir-qayyum-khan:
Done @noisecapella  @giocalitri

profiles/permissions_test.py
noisecapella: But you're not mocking view here, you're mocking `_get_object_or_404`. You shouldn't need to mock this since the objects are in the database.

noisecapella:
Just in case it gets hidden I made a comment regarding mocking `_get_object_or_404`: https://github.com/mitodl/micromasters/pull/546#discussion_r68061263

profiles/permissions_test.py
```diff
@@ -47,6 +58,141 @@ def test_cant_edit_if_not_owner(self):
...
+        Users are not supposed to view private profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+
+        with mute_signals(post_save):
+            profile = ProfileFactory.create(account_privacy=Profile.PRIVATE)
+
+        request = Mock(user=self.other_user)
+        result = Mock(return_value=profile)
+        view = Mock(kwargs={'user': self.other_user.username})
```
noisecapella: This is the wrong username, you must use the one in the edX social_auth object. You may want to define `self.username` using this line: https://github.com/mitodl/micromasters/pull/546/files#diff-2a92c5ce6fea0098a372c4f54ea70d03R77

noisecapella: Actually it looks like `self.other_user` is the wrong user in this case. If the request user is `self.other_user` then this user must be someone else. You will need to create an edx social auth for the users you create and use the username from that, here and below.

amir-qayyum-khan:
done @noisecapella

profiles/permissions_test.py
```diff
@@ -47,6 +55,135 @@ def test_cant_edit_if_not_owner(self):
...
+
+    def test_cant_view_if_anonymous_user(self):
+        """
+        Anonymous are not supposed to view public_to_mm profiles.
+        """
+        perm = CanSeeIfNotPrivate()
+        with mute_signals(post_save):
+            ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PUBLIC_TO_MM)
+
+        request = Mock(user=Mock(return_value=Mock(return_value=True)))
```
noisecapella: Note that if you changed `True` to `False` here the test would still pass. This is due to mocks in Python evaluating to true by default. 

The first `return_value` should be `is_anonymous` because the first mock returns a mocked object with `user`, and the second mock should be a mocked object with `is_anonymous`. `return_value` should only be used for when we want to mock a function.

amir-qayyum-khan: @noisecapella Done

noisecapella:
Ok I'm satisfied now, handing it back to @giocalitri

giocalitri:
Just checked functionality and it seems to work fine.
Please squash and merge after tests pass again 👍

PR #548, opened 2016-06-15
Removed popover from EducationDisplay on /users

### LINKED ISSUE #527: On /users/ remove degree dropdown from FOB and add it to education pop-up
Using the drop-down on the education card's FOB makes it inconsistent with other buttons. We can move it to the education pop-up, although that will make it harder to re-use the education pop-up from `/profile/education` on `/users/`

aliceriot:

this closes #527 

#### What's this PR do?

This removes the little popover menu for adding education entries on the `/users` page. Instead, now we can pass a prop (`showLevelForm`) in to the `EducationDialog` component. When that's `true`, we should a menu that lets the user pick the degree level (high school and so on). When false, we hide it.

Look over the changes to `EducationDisplay` and `EducationDialog`.

Make sure you can add any kind of degree type on `/users`. Make sure you can add all degree types (and they don't show the degree type dropdown) on `/profile/education`.

### DIFF
--- a/static/js/components/EducationDialog.js
+++ b/static/js/components/EducationDialog.js
@@ -2,8 +2,9 @@ import React from 'react';
 import Button from 'react-mdl/lib/Button';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 import Dialog from 'material-ui/Dialog';
+import _ from 'lodash';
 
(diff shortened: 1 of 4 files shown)


noisecapella:
:+1:

PR #556, opened 2016-06-15
Made enroll and upgrade buttons accessible

### LINKED ISSUE #473: make "enroll" and "upgrade to verified" buttons more accessible
As a screen reader user who navigates the dashboard by links, there's no way to distinguish multiple "enroll" buttons or multiple "upgrade to verified" buttons. 

We can follow the approach in https://github.com/edx/edx-platform/pull/12269 (which includes a custom `sr` CSS class

noisecapella:
Fixes #473

#### What's this PR do?
Uses the `sr-only` CSS class to show some text only for screen readers, specifically which run the button applies to. Also refactored the `DASHBOARD_RESPONSE` constant to have all possible course status cases so we can see them at a glance.

Also switches the bootstrap chevron icon with the menu one

static/js/util/courseList.js

Stub in `DASHBOARD_RESPONSE` in the `getDashboard()` function in `api.js`. View the dashboard. You should see `ENROLL` and `UPGRADE TO VERIFIED` buttons, one of each. If you inspect each button and look at the text inside the element there should also be the run title.

http://stackoverflow.com/questions/19758598/what-is-sr-only-in-bootstrap-3

### DIFF
--- a/static/js/constants_test.js
+++ b/static/js/constants_test.js
@@ -1,0 +1,40 @@
+import _ from 'lodash';
+import { assert } from 'chai';
+
+import { DASHBOARD_RESPONSE } from './constants';
+
+describe('constants', () => {
+  it("doesn't duplicate any id numbers within the same type of information", () => {
+    let programIds : Set<number> = new Set();
+    let courseIds : Set<number> = new Set();
+    let runIds : Set<number> = new Set();
+    for (let program of DASHBOARD_RESPONSE) {
+      assert(!_.isNil(program.id), 'Missing program id');
+      assert(!programIds.has(program.id), `Duplicate program id ${program.id}`);
+      programIds.add(program.id);
+
+      let positionInProgram : Set<number> = new Set();
+
+      for (let course of program.courses) {
+        assert(!_.isNil(course.id), `Missing course id for program ${program.id}`);
+        assert(!courseIds.has(course.id), `Duplicate course id ${course.id}`);
+        courseIds.add(course.id);
+
+        assert(!_.isNil(course.position_in_program), `Missing position_in_program for course ${course.id}`);
+        assert(!positionInProgram.has(course.position_in_program), `Duplicate position for course ${course.id}`);
+        positionInProgram.add(course.position_in_program);
+
+        let positionInCourse : Set<number> = new Set();
+        for (let run of course.runs) {
+          assert(!_.isNil(run.id), `Missing run id for course ${course.id}`);
+          assert(!runIds.has(run.id), `Duplicate run id ${run.id}`);
+          runIds.add(run.id);
+
+          assert(!_.isNil(run.position), `Missing position for run ${run.id}`);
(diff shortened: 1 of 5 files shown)


aliceriot:
the change to `DASHBOARD_RESPONSE` is :100:, trying to debug all the different states can get really confusing.

aliceriot:
functionality looks good, after changing the return value of `getDashboard` to `Promise.resolve(DASHBOARD_RESPONSE)` I see an 'ENROLL' button with the correct `.sr-only` content, and I also see an 'UPGRADE TO VERIFIED' button with the correct content.

static/js/constants_test.js
```diff
@@ -0,0 +1,40 @@
+import _ from 'lodash';
+import { assert } from 'chai';
+
+import { DASHBOARD_RESPONSE } from './constants';
+
+describe('constants', () => {
+  it("doesn't duplicate any id numbers within the same type of information", () => {
+    let programIds : Set<number> = new Set();
+    let courseIds : Set<number> = new Set();
+    let runIds : Set<number> = new Set();
```
aliceriot: We don't have flow typechecking turned on for any test files, do you think this is something we should do? I had avoided it because I thought it might be a little onerous.

noisecapella: Might be good to turn it on eventually. Since we're not defining functions or variables for outside use in these files the type checking should hopefully happen without additional annotation needed

static/js/util/courseList.js
```diff
@@ -84,7 +87,10 @@ export function makeCourseStatusDisplay(course: Object, now: moment = moment()):
     } else {
       if (firstRun.course_id) {
         let courseInfoUrl = `${SETTINGS.edx_base_url}/courses/${firstRun.course_id}/about`;
-        return <Button bsStyle="success" href={courseInfoUrl} target="_blank">ENROLL</Button>;
+        return <Button bsStyle="success" href={courseInfoUrl} target="_blank">
+          ENROLL
+          <span className="sr-only"> in {firstRun.title}</span>
+        </Button>;
```
aliceriot: When I actually inspect the markup it looks something like this:

react-silliness

I imagine any screenreader will ignore comments, but do you think if we did this:

```js
<span className="sr-only">{` in ${firstRun.title}`}</span>
```

we'd have cleaner markup on the page?

static/js/constants_test.js
aliceriot: True, maybe it would be pretty painless then.

aliceriot:
One question, other than that I think the code and functionality looks good.

static/js/util/courseList.js
noisecapella: It looks like the extra comment tags are only visible in the inspector on my local machine, I don't see it on micromasters-ci for example

aliceriot: word, not a problem them.

aliceriot:
merge away! :+1:

PR #563, opened 2016-06-16
Fixed spinner positioning

### LINKED ISSUE #465: [bug] user page: spinner appears below footer
this is what the user sees on the user page if something takes a while to load:

spinner

noisecapella:
Fixes #465 

#### What's this PR do?
Sets the container div for the loader to use relative positioning so the inner spinner will be centered within that, rather than centered inside the entire page.

In `api.js` replace `getUserProfile` with this:

    export function getUserProfile(username) {
        setTimeout(() => {
          resolve(mockableFetchJSONWithCSRF(`/api/v0/profiles/${username}/`));
        }, 100000);
      });
    }

This will delay the load of the user page for 100 seconds. Go to `/users/` and see that the spinner is where it should be. You should be able to resize the browser without affecting the position of the spinner.

screenshot from 2016-06-16 14-30-00

### DIFF
--- a/static/scss/layout.scss
+++ b/static/scss/layout.scss
@@ -47,3 +47,10 @@ body {
     width: 100%;
     height: 100%;
 }
+
+.loader {
+  position: relative;
+  // arbitrarily chosen height
+  height: 300px;
+  width: 100%;
+}


gsidebo:
do we care about EOF linebreaks? everything looks good here other than that possible comment

gsidebo:
👍

aliceriot:
@gsidebo it seems like we don't in general. I dislike seeing the little :no_entry_sign: symbol, but ¯\\\_(ツ)\_/¯

PR #382, opened 2016-05-23
Removed AutoComplete onBlur handling when user has clicked a menu item

noisecapella:
#### What are the relevant tickets?
Fixes #380 

#### What's this PR do?
When the user selected an item the onBlur handler cleared the item that was selected immediately afterwards. This change causes the onBlur handler not to be invoked if the user just selected a menu item.

#### Where should the reviewer start?
`this.timerTouchTapCloseId` is set after the user selects a menu item, so we are using this in the onFocus callback to determine if the loss of focus was caused by clicking a menu item

#### How should this be manually tested?
We should be thorough in testing all different scenarios. For each item below notice these things:
 - unless you are actively typing there should never be text in the text field that doesn't match the currently selected item.
 - If an item should be selected, the country text field should have the text of the item, and when you click Save, it should pass validation. You should be able to click the state field and see a list of states that match that country.
 - If an item should not be selected, the country and state fields should be blank. When you click Save it should show a validation error under the Country. When you click the state field you should see no states to be selected.

If you're testing this locally you should also be able to verify the redux state matches what the field indicates.

For each item below start by opening a new employment form with the plus button. Then:
 - Click the country field, then click a country in the list. Verify that you selected that country.
 - Click the country field, type some text such that you get a country in the list, then press Enter to select the first country in the list. Verify that you selected the first country in the list.
 - Click on the country field, then type some text such that it doesn't match any countries. Press enter. Verify that no country is selected.
 - Click on the country field, then type text that matches the name of a country, but click elsewhere rather than selecting the country. Verify that no country is selected.
 - Click on the country field, then press down until you get to a country you want to select. Press Enter on that country. Verify that this country is selected.
 - First, select a country. Then click on a previous field, then TAB until you get to the country field. The menu list should not open up and the text should be selected. Press TAB again, then verify that the country is still selected.
 - First, select a country. Click on the field, then click on another field. Verify that the country is still selected.

Also:
 - verify that you see a complete list of countries when you click the field, but when you type text you see only countries whose prefix matches the text
 - edit another work history and verify that the expected country shows up in the select field

Hopefully that covers all the scenarios we care about!

PR #568, opened 2016-06-17
Used virtualized scrolling for AutoComplete component

### LINKED ISSUE #515: Investigate react-virtualized to fix AutoComplete performance issues
https://github.com/bvaughn/react-virtualized

This is used in https://github.com/bvaughn/react-virtualized-select to render only a portion of the option list for react-select. I'm wondering if we can do the same thing for AutoComplete

noisecapella:
Fixes #515 

#### What's this PR do?
Adds `Menu` and `List` from `material-ui` and modifies them and `AutoComplete` to increase performance and use virtualized scrolling.

In general things should act exactly as they did before, except faster hopefully or at least not slower. Open up micromasters-ci in a separate tab and for each case compare functionality between it and your local instance which is on this branch.

First, go through all of the cases in the manual testing section of https://github.com/mitodl/micromasters/pull/382 and verify that it all works. Additionally check these cases (make sure to confirm the redux state for each selection):
 - You should be able to press down to select the first item in the menu. If you press enter you should select that item in the list. 
 - Type text to filter the results, then press down and select some other item in the list. If you press enter you should select that item in the list.
 - Type text to filter the results, press down, then click an item on the list. It should be selected successfully.
 - Type text then press enter. The first item in the results should be selected.

Other things to check for:
 - scrollbars should show up if and only if appropriate. You should be able to use them like any other list.
 - When you click on a autocomplete field with many items, like the country field, it should open about as quickly as any other autocomplete field on the page. It should not be slower than it was before.
 - You should be able to scroll to the bottom of the list and select the last item without any problem.
 - The popup height should be at most 300 pixels, and if less space is needed it should shrink to fit the items in the list.
 - The width should match the text field's width.

Bugs:
 - if you press the page down key a few times and then press down, it will continue scrolling. This is different from the previous behavior where the down key would move to the next selected item in the list, no matter how far down in the list you go.

http://i.giphy.com/zf3skxc4BFBqo.gif


static/js/components/AutoComplete.js
```diff
@@ -393,7 +405,8 @@ class AutoComplete extends Component {
       triggerUpdateOnFocus, // eslint-disable-line no-unused-vars
       openOnFocus, // eslint-disable-line no-unused-vars
       maxSearchResults,
-      dataSource,
+      dataSource, // eslint-disable-line no-unused-vars
+      menuHeight,
```
gsidebo: seems odd that `menuHeight` wouldn't be part of `menuStyle`. is there a reason for this? either way, this should be moved up to be proximal to `menuStyle` and `menuProps` in this list

static/js/components/AutoComplete.js
```diff
@@ -407,84 +420,43 @@ class AutoComplete extends Component {
...
-    });
+    let renderItem = option => (
+      <MenuItem
+        key={option.value}
+        primaryText={option.label}
+        value={option.value}
+        disableFocusRipple={disableFocusRipple}
+        innerDivStyle={styles.innerDiv}
+      />
+    );
```
gsidebo: is it necessary/beneficial to define this inside of the `render` method? i think this would be cleaner if it was defined on the class and accepted a `divStyle` argument, et al. we would also avoid redefining this function every time `render` was called

static/js/components/AutoComplete.js
```diff
@@ -407,84 +420,43 @@ class AutoComplete extends Component {
...
         onItemTouchTap={this.handleItemTouchTap}
         onMouseDown={this.handleMouseDown}
-        style={Object.assign(styles.menu, menuStyle)}
+        style={Object.assign(styles.menu, menuStyleWithHeight)}
+        menuHeight={menuHeight}
         listStyle={Object.assign(styles.list, listStyle)}
-      >
-        {requestsList.map((i) => i.value)}
-      </Menu>
+        requestsList={requestsList}
```
gsidebo: `requestsList` strikes me as an odd name. i would think something like `selectableOptions` would describe this more accurately

noisecapella: This name comes from the original source

static/js/components/AutoComplete.js
noisecapella: I added `menuHeight` because we need to pass this number to the `VirtualScroll` method

gsidebo:
just checked out the branch and ran through the profile forms. this is performing really nicely. should we think about manually testing this in other browsers?

noisecapella:
Yes, we generally support the same browsers as edX: http://edx.readthedocs.io/projects/edx-guide-for-students/en/latest/front_matter/browsers.html

gsidebo:
ss 2016-06-21 at 12 14 49 pm

trying to test in an Win7/IE11 VM. happening on -ci too. ill file an issue for this. as it stands, i can't test this in IE11, soooo

👍

noisecapella:
I filed an issue to fix this: #609

gsidebo:
hold on a sec - jumped the gun on the +1. i just want to review the changes you made once more

noisecapella:
That's fine, we have the merge freeze anyway

static/js/components/AutoComplete.js
gsidebo: that's a concern of the child component. i still think height should be part of menuStyle (as 'maxHeight'). Menu.js can worry about passing the correct value for 'menuHeight'

static/js/components/AutoComplete.js
gsidebo: this is a nitpick, but another libraries unfortunate naming doesn't necessarily mean that ours needs to be the same. we can name this `selectableOptions` or something and pass it to `Menu` as `requestsList`. at the very least our naming in this file would be more understandable. you can take or leave this change - i don't feel too strong about it either way

gsidebo:
works in IE11

noisecapella:
I rebased on master so this PR should now include the polyfill

static/js/components/AutoComplete.js
noisecapella: If we wanted to merge in future changes from material-ui it would be helpful to keep the existing variable names the same. I agree that it can be a tricky balance

static/js/components/AutoComplete.js
noisecapella: I'll add it back to `menuStyle` but we still need to pass it as a separate prop since `maxHeight` is not the same thing as `VirtualScroll`'s height, and it seems clunky to extract that information from a CSS style object

noisecapella:
Any other comments?

gsidebo:
tested in IE11. looks great! 👍

PR #601, opened 2016-06-20
Added 'confirm delete all entries' when closing switches

### LINKED ISSUE #482: Clarify behavior when turning switches to 'off' on education and work history forms
Currently turning off the switch doesn't remove any previously entered jobs or degrees, so a user could enter two jobs, flip the switch to 'off', and then click 'save and continue', and currently the entered jobs would be saved.

We should clarify what we intend for the switches to do. Here's an idea:

- If a user has already entered some input, a confirmation is shown ('really delete all jobs?')
- If a user has not entered any input, the switch goes to 'off' and they are allowed past the validation
- If the switch is 'on' we validate for the presence of at least one entry (one job, one degree, etc)

aliceriot:

closes #482 

#### What's this PR do?

This adds a pop-up dialog that asks you to confirm deleting entries when toggling switches on the work history and education forms. Basically, this means that if you have entries saved and you try to turn the switch off you have to either affirm your intention to delete all the entries, or cancel turning off the switch. It shouldn't be possible get into a state where there are invisible history entries which will be saved upon clicking 'save and continue'.

I guess start with the changes to `EmploymentForm`, `EducationForm`, and `ConfirmDeletion`.

Visit `/profile/professional` and `/profile/education`. On the former

- confirm that you can freely toggle the switch when no work history is entered
- confirm that entering a work history entry causes the pop-up to appear upon clicking the switch
- confirm that cancelling the popup is non-destructive
- confirm that confirming the popup deletes all education entries

on the latter:

- confirm that you can freely toggle the switch for a degree level when no history is entered *for that particular level*
- confirm that entering a degree causes the popup to appear for that particular level, but not for any other.
- confirm that cancelling the popup is non-destructive
- confirm that confirming the popup deletes all education entries for a particular degree level

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -1,3 +1,4 @@
+// @flow
 import React from 'react';
 import IconButton from 'react-mdl/lib/IconButton';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
@@ -16,6 +17,9 @@ import {
   openNewEducationForm,
   deleteEducationEntry,
 } from '../util/editEducation';
+import { educationValidation } from '../util/validation';
+import type { Option } from '../flow/generalTypes';
+import type { EducationEntry } from '../flow/profileTypes';
 
 class EducationForm extends ProfileFormFields {
   static propTypes = {
@@ -30,19 +34,19 @@ class EducationForm extends ProfileFormFields {
     setEducationDegreeInclusions:   React.PropTypes.func,
   };
 
-  openEditEducationForm = index => {
+  openEditEducationForm: Function = (index: number): void => {
     openEditEducationForm.call(this, index);
   };
 
-  openNewEducationForm = (level, index) => {
+  openNewEducationForm: Function = (level: Option, index: number): void => {
     openNewEducationForm.call(this, level, index);
   };
 
-  deleteEducationEntry = () => {
+  deleteEducationEntry: Function = (): void => {
     deleteEducationEntry.call(this);
   };
 
-  renderEducationLevel(level){
+  renderEducationLevel(level: Option): ?React$Element[] {
     const {
       ui: { educationDegreeInclusions },
       profile: { education },
@@ -51,7 +55,7 @@ class EducationForm extends ProfileFormFields {
       let rows = [];
       if (education !== undefined) {
         rows = Object.entries(education).
-          filter(([, education]) => education.degree_name === level.value).
+          filter(([, education]: [string, EducationEntry]) => education.degree_name === level.value).
           map(([index, education]) => this.educationRow(education, index));
       }
       rows.push(
@@ -68,7 +72,7 @@ class EducationForm extends ProfileFormFields {
     }
   }
 
-  educationRow = (education, index) => {
+  educationRow: Function = (education, index) => {
     const { errors } = this.props;
     if (!('id' in education)) {
       // don't show new educations, wait until we saved on the server before showing them
@@ -98,14 +102,51 @@ class EducationForm extends ProfileFormFields {
     </Grid>;
   };
 
-  handleSwitchClick(level){
+  closeDeleteAllEducationDialog: Function = (): void => {
+    const {
+      setEducationDegreeLevel,
+      setShowEducationDeleteAllDialog,
+    } = this.props;
+    setEducationDegreeLevel('');
+    setShowEducationDeleteAllDialog(false);
+  };
+
+  deleteAllEducationEntriesForLevel: Function = (): void => {
+    const {
+      profile,
+      saveProfile,
+      ui,
+      ui: { educationDegreeLevel, educationDegreeInclusions },
+      setEducationDegreeInclusions,
+    } = this.props;
+    let clone = _.cloneDeep(profile);
+    clone.education = clone.education.filter(entry => (
+      entry.degree_name !== educationDegreeLevel
+    ));
+    let newState = Object.assign({}, educationDegreeInclusions, {
+      [educationDegreeLevel]: false
+    });
+    setEducationDegreeInclusions(newState);
+    saveProfile(educationValidation, clone, ui);
+  };
+
+  handleSwitchClick(level: string): void {
     const {
       ui: { educationDegreeInclusions },
       setEducationDegreeInclusions,
+      setEducationDegreeLevel,
+      setShowEducationDeleteAllDialog,
+      profile: { education },
     } = this.props;
-    let newState = Object.assign({}, educationDegreeInclusions);
-    newState[level] = !educationDegreeInclusions[level];
-    setEducationDegreeInclusions(newState);
+    if ( !education.find(entry => entry.degree_name === level) ) {
+      let newState = Object.assign({}, educationDegreeInclusions, {
+        [level]: !educationDegreeInclusions[level]
+      });
+      setEducationDegreeInclusions(newState);
+    } else {
+      setEducationDegreeLevel(level);
+      setShowEducationDeleteAllDialog(true);
+    }
   }
 
   render() {
@@ -115,6 +156,8 @@ class EducationForm extends ProfileFormFields {
       ui: {
         educationDegreeInclusions,
         showEducationDeleteDialog,
+        showEducationDeleteAllDialog,
+        educationDegreeLevel,
       }
     } = this.props;
 
@@ -138,6 +181,7 @@ class EducationForm extends ProfileFormFields {
           <Cell col={7} />
           <Cell col={1}>
             <Switch
+              id={`profile-tab-education-switch-${level.value}`}
               onChange={()=>{this.handleSwitchClick(level.value);}}
               checked={educationDegreeInclusions[level.value]}
             />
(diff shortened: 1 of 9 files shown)


noisecapella:
Functionality looks good

static/js/components/EducationForm.js
```diff
@@ -138,7 +189,9 @@ class EducationForm extends ProfileFormFields {
           <Cell col={7} />
           <Cell col={1}>
             <Switch
-              onChange={()=>{this.handleSwitchClick(level.value);}}
+              id={`profile-tab-education-switch-${level.value}`}
+              onClick={()=>{this.handleSwitchClick(level.value);}}
+              readOnly={true}
```
noisecapella: It doesn't look like this prop exists within react-mdl's Switch component

aliceriot: without it I get a console warning. It's not a prop on the mdl component, but on an html `input` tag:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

static/js/components/EducationForm.js
```diff
@@ -98,14 +102,59 @@ class EducationForm extends ProfileFormFields {
...
+    let clone = _.cloneDeep(profile);
+    toDelete.forEach(index => clone.education.splice(index, 1));
     let newState = Object.assign({}, educationDegreeInclusions);
-    newState[level] = !educationDegreeInclusions[level];
+    newState[educationDegreeLevel] = false;
     setEducationDegreeInclusions(newState);
+    saveProfile(educationValidation, clone, ui);
+  };
+
+  handleSwitchClick(level: string){
```
noisecapella: Can you add a return type of `void`?

static/js/components/EducationForm.js
noisecapella: I see, it looks like it's expecting us to set `onChange` instead of `onClick` here. Can you make this change here and for the employment tab?

aliceriot: I probably should have just explained this, haha.

I set an `onClick` handler because it greatly simplifies testing and the user experience is the same - I had a lot of difficulty getting the change event to fire, but simulating a click was relatively straightforward. If we don't set `onChange` then the component becomes a controlled component, so it doesn't have any value that changes on itself, but instead just broadcasts changes to the store and subscribes to updates.

Does that make sense? It might be that there's an easy way to get the `onChange` callback to fire when testing, I spent a lot time puzzling over it and didn't get it work :(

static/js/components/EducationForm.js
```diff
@@ -98,14 +102,59 @@ class EducationForm extends ProfileFormFields {
...
+    const {
+      profile,
+      saveProfile,
+      ui,
+      ui: { educationDegreeLevel, educationDegreeInclusions },
       setEducationDegreeInclusions,
     } = this.props;
+    let toDelete = this.entriesForLevel(educationDegreeLevel);
+    let clone = _.cloneDeep(profile);
+    toDelete.forEach(index => clone.education.splice(index, 1));
```
noisecapella: The `splice` in the loop here may have bad performance (though probably not noticeable) since all of the elements after the item being spliced are copied and this happens for each index being deleted. Why not do a `filter` here instead to remove the levels?

static/js/components/EducationForm.js
noisecapella: I experimented with this and `TestUtils.Simulate.change` should work fine without a second argument. This does work for me as is but if this fixes the testing issue I'd prefer to use `onChange` since it's documented in the `react-mdl` docs

aliceriot: aha, I should really read the docs for things sometimes...

I'll change it back to use `onChange`.

static/js/components/EducationForm.js
aliceriot: I think there's actually a bigger problem with what I have here, which is that `splice` will cause changes in which entries indexes point to....

consider:

```js
let entries = [a, b, c, d];
let toDelete = [1,2]; // entries[1] = b, entries[2] = c
toDelete.forEach(index => entries.splice(index, 1))
```

in the above in the first iteration of the loop index `1` will point to b, as we intend. After we splice that out, though, we'll have:

```js
entries = [a, c, d];
```

Now we're going to splice out the entry at index `2`, which is now `d`, and not `b`, which we had actually intended to delete. Oopsie.

static/js/containers/ProfilePage_test.js
```diff
@@ -70,6 +83,182 @@ describe("ProfilePage", function() {
...
+      });
+    });
+
+
+    let educationSwitchSelectors = EDUCATION_LEVELS.map(level => (
+      { value: level.value, label: level.label, selector: `#profile-tab-education-switch-${level.value}` }
+    ));
+
+    let educationEntries = (level, profile) => {
+      let out = [];
```
noisecapella: Can you use `filter` here?

static/js/containers/ProfilePage_test.js
```diff
@@ -70,6 +83,182 @@ describe("ProfilePage", function() {
...
+        "school_name": "Harvard",
+        "school_city": "Cambridge",
+        "school_state_or_territory": "US-MA",
+        "school_country": "US",
+        "online_degree": false
+      }));
+      return clone;
+    };
+
+    educationSwitchSelectors.forEach( ({label: label, value: value, selector: selector}) => {
```
noisecapella: This line and below can be shortened to `{ label, value, selector }`

static/js/components/EducationForm.js
noisecapella: I see, good catch. Can you add a test for this case?

aliceriot: yah

static/js/containers/ProfilePage_test.js
```diff
@@ -70,6 +83,182 @@ describe("ProfilePage", function() {
...
+            SET_SHOW_EDUCATION_DELETE_ALL_DIALOG,
+            SET_EDUCATION_DEGREE_INCLUSIONS,
+            START_PROFILE_EDIT,
+            UPDATE_PROFILE_VALIDATION,
+            REQUEST_PATCH_USER_PROFILE,
+          ], () => {
+            TestUtils.Simulate.change(toggle);
+            let dialog = openDialog();
+            let deleteButton = dialog.querySelector('.delete-button');
+            TestUtils.Simulate.click(deleteButton);
```
aliceriot: @noisecapella would you be able to help me debug some test issues? I'm not sure exactly what's going on, but any code I put after this line doesn't execute. The callback on `deleteButton` is being called, as is the `saveProfile` method on `ProfileFormFields` - it seems like I'm not doing the `PATCH` mocking correctly.

Any thoughts?

noisecapella: Not sure, I'll run it in the debugger and look into it

noisecapella: It looks like the `id`=2 for all new elements defined in `fullEducation`, not sure if that's causing the problem yet

noisecapella: Also `username` is being set in the test for the profile but it's missing in the `USER_PROFILE_RESPONSE` constant.

noisecapella: Also I just realized that when the last expected action is received the promise is resolved, which means that any errors which occur afterwards, or new actions that are dispatched before the callback completes, aren't detected. I'll look into fixing that in redux-asserts

aliceriot: I'm setting it on the `.returns` for the `PATCH` mock above, should I not have that?

noisecapella: I think the only problem with that is that it doesn't match what was passed to `receivedGetUserProfileSuccess` earlier

aliceriot: oh! yes, I added that to `fullEducation` and it's working now!

aliceriot: this is worrying though, b/c I can put `assert(false)` after this line and the test still passes...

aliceriot: I'm going to push up what this looks like now so you can take a look

static/js/components/EducationForm.js
```diff
@@ -98,14 +102,60 @@ class EducationForm extends ProfileFormFields {
...
+      saveProfile,
+      ui,
+      ui: { educationDegreeLevel, educationDegreeInclusions },
       setEducationDegreeInclusions,
     } = this.props;
+    let clone = _.cloneDeep(profile);
+    clone.education = clone.education.filter(entry => (
+      entry.degree_name !== educationDegreeLevel
+    ));
     let newState = Object.assign({}, educationDegreeInclusions);
```
noisecapella: Can you update this using ES6 computed keys, similar to `handleSwitchClick` below?

static/js/components/EducationForm.js
```diff
@@ -98,14 +102,60 @@ class EducationForm extends ProfileFormFields {
...
+  };
+
+  handleSwitchClick(level: string): void {
+    const {
+      ui: { educationDegreeInclusions },
+      setEducationDegreeInclusions,
+      setEducationDegreeLevel,
+      setShowEducationDeleteAllDialog,
+    } = this.props;
+    if ( _.isEmpty(this.entriesForLevel(level)) ) {
```
noisecapella: Can we change this to use `profile.education.find(...)`? It seems like we aren't using `entriesForLevel` for anything except detecting the presence of the level so it would be nice to remove it or refactor/rename for that purpose

static/js/containers/ProfilePage_test.js
```diff
@@ -70,6 +91,181 @@ describe("ProfilePage", function() {
...
+      { value: level.value, label: level.label, selector: `#profile-tab-education-switch-${level.value}` }
+    ));
+
+    let educationEntries = (level, profile) => (
+      profile.education.filter(entry => entry.degree_name === level)
+    );
+
+    let fullEducation = () => {
+      let clone = _.cloneDeep(USER_PROFILE_RESPONSE);
+      [ASSOCIATE, MASTERS, DOCTORATE, MASTERS].forEach((level, index) => clone.education.push({
```
noisecapella: Can you add two of each instead of just one, and add them such that they aren't right next to each other in the array? It would help testing that deletion works properly

static/js/components/EducationForm.js
aliceriot: Yeah, I think computed properties are a cleaner pattern.

static/js/components/EducationForm.js
aliceriot: Yes, I was previously using `entriesForLevel` for deleting entries when removing a degree level, but now that we're not using it there it can probably be deleted.

aliceriot:
@noisecapella comments should be addressed, and I think all the testing issues are sorted (whew)

PR #613, opened 2016-06-21
Sorted employment entries in resume order

### LINKED ISSUE #584: Sort employment entries on user and profile pages
"resume order" 

If in doubt, look at LinkedIn's sorting. 

same component for user page and profile page.

aliceriot:

closes #584 

#### What's this PR do?

This adds a function to `util.js` that takes an `Array<EmploymentEntry>` and returns a sorted array (in resume order). Then we can just call this on `profile.work_history` before we draw the UI to get things in the right order.

Check out the changes to `util/util.js` and make sure the tests all look legit.

Add a variety of employment entries, with and without end dates, and check that the following sort order is displayed on `/users` and on `/profile/professional`:

- 'current' entries (w/out an end date) should be displayed first, as a group
- 'current' entries should be ordered by start date
- 'non-current' entries should appear after all 'current' entries
- 'non-current' entries should be ordered by end date

current

### DIFF
--- a/static/js/components/EmploymentForm.js
+++ b/static/js/components/EmploymentForm.js
@@ -12,6 +12,7 @@ import _ from 'lodash';
 import moment from 'moment';
 
 import { generateNewWorkHistory, userPrivilegeCheck } from '../util/util';
+import { sortWorkEntriesByDate } from '../util/sorting';
 import { employmentValidation } from '../util/validation';
 import ProfileFormFields from '../util/ProfileFormFields';
 import ConfirmDeletion from './ConfirmDeletion';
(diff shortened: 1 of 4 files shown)


static/js/util/util.js
```diff
@@ -239,3 +240,26 @@ export function calculateDegreeInclusions(profile: Profile) {
   }
   return inclusions;
 }
+
+type SortableEntry = WorkHistoryEntry|EducationEntry;
+export function resumeOrder(entries: SortableEntry[], dateFieldName: string): SortableEntry[] {
+  let sortFunc = (a, b) => {
```
noisecapella: Can you rename `a` and `b` to `start` and `end`, and also the same with `adate` and `bdate`?

noisecapella: Oh nevermind, I misread this function

static/js/util/util_test.js
```diff
@@ -217,4 +220,57 @@ describe('utility functions', () => {
       });
     });
   });
+
+  describe('profile sort functions', () => {
+    it('should sort by resume order', () => {
+      let entries = ['1969', '1997', '1992', '1934'].map(year => (
+        { 'date': moment(year, 'YYYY') }
```
noisecapella: We shouldn't pass in a moment object here since it's not how we're using `resumeOrder` elsewhere. Can you format this moment object to ISO-8601?

noisecapella:
Functionality looks great

aliceriot:
@noisecapella do you think it makes sense to put these functions in `util/util.js`? I thought about adding a `util/sorting.js` or something similar.

noisecapella:
Either way works for me

aliceriot:
k, I might go for another file in the interest of not letting `util.js` get too big.

noisecapella:
Looks good :+1:

Issue #532, opened 2016-06-10
Fix validation copy


noisecapella:
Forgot to fix this before #504 was merged. From https://github.com/mitodl/micromasters/pull/504#discussion_r66634253

> For the first one, I think it would be good to give the user some guidance on what to do, e.g.

> "A Master's Degree is required if the Master's Degree switch is on. Click the + button to add a degree or switch it off"

> ... that's probably too long though.

> What's the user supposed to do in the 2nd case, on L103?

noisecapella:
Here's what I have so far. It needs some work, any comments so far?

![screenshot from 2016-06-23 11-50-52](https://cloud.githubusercontent.com/assets/863262/16310034/ea51f6c8-3938-11e6-8ab2-b07d5c728642.png)

roberthouse54:
To be honest, I wonder if we can hold off on this. I feel that I want to probably redesign this sign up flow in a way to get rid of the switches completely. The switches UI paradigm makes no sense to me.

Ferdi:
OK to redesign but can we put some text there until that happens?

roberthouse54:
OK, go with this then:

eg.

"Master's Degree is required if the switch is on. Please add a degree or switch it off."

Also, please make sure the text is aligned with the header above it. It is a bit too far left.

PR #618, opened 2016-06-22
Removed UI validators from PrivacyTab validator callback

### LINKED ISSUE #576: profile: If user refreshes page while editing profile, education switch state is lost
This is causing confusing validation errors the user can't see. To reproduce:
 - fill out education profile, click Save and continue
 - on employment page refresh the browser
 - fill out employment profile, click Save and continue
 - fill out Privacy, click "I'm Done"

Since the UI state is lost and the validation when "I'm Done" is clicked includes UI state validation, the user will be prevented from moving on. Ideas:
 - validate only the UI of the page the user is currently on
 - persist education switch state in database
 - direct user back to the profile page where they had the error

aliceriot:

closes #576 

#### What's this PR do?

removes validation of the education and work history UI state when saving on the `PrivacyTab`.

Follow the description of the repro for the problem on the issue: #576, and confirm that you can successfully save on `/profile/privacy`.

### DIFF
--- a/static/js/components/PrivacyTab.js
+++ b/static/js/components/PrivacyTab.js
@@ -8,9 +8,7 @@ import {
   combineValidators,
   personalValidation,
   educationValidation,
-  educationUiValidation,
   employmentValidation,
-  employmentUiValidation,
   privacyValidation,
 } from '../util/validation';
 import type { Profile } from '../flow/profileTypes';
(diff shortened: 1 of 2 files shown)


gsidebo:
are we also working on an issue this sprint to prevent users from navigating directly to a profile page without filling out the previous steps? we're going to need both before we go live

aliceriot:
I think there is an issue filed, not sure it's in the sprint.

aliceriot:
hrm maybe not, I just took a quick look and couldn't find it

gsidebo:
@pdpinch see above. i couldv'e sworn we created an issue during sprint planning to prevent a user from navigating directly to a profile page without filling out the previous steps

noisecapella:
Is this still being reviewed? It may collide with the work for #532 so it would be nice to get it merged first

noisecapella:
Functionality looks good :+1:

PR #619, opened 2016-06-22
Added Course and CourseRun types, among others

### LINKED ISSUE #555: Add flow types for course and run

aliceriot:

closes #555 

#### What's this PR do?

Add new `Course` and `CourseRun` types, and adds type annotations and flow checking to a bunch of files we weren't checking before.

Look over the changes, make sure the type annotations are accurate and that none of the changes made to get the typechecker to pass will change any behavior (or, in any bad ways).

Well, I guess there shouldn't be any regressions, so editing profiles and so on should all work as normal.

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -19,26 +19,33 @@ import {
 } from '../util/editEducation';
 import { educationValidation } from '../util/validation';
 import type { Option } from '../flow/generalTypes';
-import type { EducationEntry } from '../flow/profileTypes';
+import type {
+  EducationEntry,
+  Profile,
+  ValidationErrors,
+} from '../flow/profileTypes';
+import type { UIState } from '../reducers/ui';
 
 class EducationForm extends ProfileFormFields {
-  static propTypes = {
-    profile:                        React.PropTypes.object,
-    ui:                             React.PropTypes.object,
-    updateProfile:                  React.PropTypes.func,
-    clearProfileEdit:               React.PropTypes.func,
-    errors:                         React.PropTypes.object,
-    setEducationDialogVisibility:   React.PropTypes.func,
-    setEducationDialogIndex:        React.PropTypes.func,
-    setEducationDegreeLevel:        React.PropTypes.func,
-    setEducationDegreeInclusions:   React.PropTypes.func,
+  props: {
+    profile:                          Profile,
+    ui:                               UIState;
+    updateProfile:                    () => void,
+    saveProfile:                      () => void,
+    clearProfileEdit:                 () => void,
+    errors:                           ValidationErrors,
+    setEducationDialogVisibility:     () => void,
+    setEducationDialogIndex:          () => void,
+    setEducationDegreeLevel:          () => void,
+    setEducationDegreeInclusions:     () => void,
+    setShowEducationDeleteAllDialog:  (bool: boolean) => void,
   };
 
   openEditEducationForm: Function = (index: number): void => {
     openEditEducationForm.call(this, index);
   };
 
-  openNewEducationForm: Function = (level: Option, index: number): void => {
+  openNewEducationForm: Function = (level: string, index: number): void => {
     openNewEducationForm.call(this, level, index);
   };
 
@@ -46,17 +53,17 @@ class EducationForm extends ProfileFormFields {
     deleteEducationEntry.call(this);
   };
 
-  renderEducationLevel(level: Option): ?React$Element[] {
+  renderEducationLevel(level: Option): Array<React$Element|void>|void {
     const {
       ui: { educationDegreeInclusions },
       profile: { education },
     } = this.props;
     if (educationDegreeInclusions[level.value]) {
-      let rows = [];
+      let rows: Array<React$Element|void> = [];
       if (education !== undefined) {
-        rows = Object.entries(education).
-          filter(([, education]: [string, EducationEntry]) => education.degree_name === level.value).
-          map(([index, education]) => this.educationRow(education, index));
+        rows = education.map((entry, index) => (
+          entry.degree_name === level.value ? this.educationRow(entry, index) : undefined
+        ));
       }
       rows.push(
         <FABButton
@@ -72,7 +79,7 @@ class EducationForm extends ProfileFormFields {
     }
   }
 
-  educationRow: Function = (education, index) => {
+  educationRow: Function = (education: EducationEntry, index: number) => {
     const { errors } = this.props;
     if (!('id' in education)) {
       // don't show new educations, wait until we saved on the server before showing them
--- a/static/js/components/EmploymentForm.js
+++ b/static/js/components/EmploymentForm.js
@@ -133,15 +133,15 @@ class EmploymentForm extends ProfileFormFields {
     );
   }
 
-  renderWorkHistory(): ?React$Element[] {
+  renderWorkHistory(): Array<React$Element|void>|void {
     const { ui, profile, profile: { work_history } } = this.props;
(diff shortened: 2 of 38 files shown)


static/js/util/validation.js
```diff
@@ -277,8 +288,8 @@ export function validateYear(string: string): ?number {
 /**
  * Returns a function which merges the results of the given functions on a set of arguments
  */
-export function combineValidators(...validators: Array<UIValidator|Validator>): UIValidator|Validator {
-  return (...args: Array<Profile|UIState>) => _.merge({}, ...validators.map(
+export function combineValidators(...validators: Array<Function>): Function {
+  return (...args) => _.merge({}, ...validators.map(
     validator => validator(...args)
   ));
 }
```
aliceriot: Unfortunately I haven't been able to get any more specific annotation to type check for this function ; _ ;

static/js/components/EducationForm.js
```diff
@@ -30,29 +32,29 @@ class EducationForm extends ProfileFormFields {
...
     openNewEducationForm.call(this, level, index);
   };
 
-  deleteEducationEntry = () => {
+  deleteEducationEntry: Function = (): void => {
     deleteEducationEntry.call(this);
   };
 
-  renderEducationLevel(level){
+  renderEducationLevel(level: {value: string; label: string}) {
```
noisecapella: Is the semicolon here intentional?

static/js/components/EmploymentForm.js
```diff
@@ -120,9 +120,9 @@ class EmploymentForm extends ProfileFormFields {
     if ( ui.workHistoryEdit === true ) {
       let workHistoryRows = [];
       if ( !_.isUndefined(work_history) ) {
-        workHistoryRows = Object.entries(work_history).filter(([,entry]: [string, WorkHistoryEntry]) =>
-          entry.id !== undefined
-        ).map(([i, entry]) => this.jobRow(entry, i));
+        workHistoryRows = work_history.map((entry, index) => (
```
noisecapella: Can you annotate `entry` and `index` here? It looks like flow isn't doing a good job infering the types for these

static/js/components/EducationForm.js
```diff
@@ -30,29 +32,29 @@ class EducationForm extends ProfileFormFields {
...
       profile: { education },
     } = this.props;
     if (educationDegreeInclusions[level.value]) {
-      let rows = [];
+      let rows: Array<React$Element|void> = [];
       if (education !== undefined) {
-        rows = Object.entries(education).
-          filter(([, education]) => education.degree_name === level.value).
-          map(([index, education]) => this.educationRow(education, index));
+        rows = education.map((entry, index) => (
```
noisecapella: Can you annotate `entry` and `index` here?

static/js/components/EducationForm.js
aliceriot: yeah, for some reason you separate the key/value pairs in a object type declaration with semicolons?

static/js/components/EducationForm.js
aliceriot: Do you think it's insufficiently inferrable? `education` is `Array<EducationEntry>` and `index` is `number`.

static/js/util/api.js
```diff
@@ -1,9 +1,11 @@
+// @flow
 /* global SETTINGS:false, fetch: false */
 // For mocking purposes we need to use 'fetch' defined as a global instead of importing as a local.
 import 'isomorphic-fetch';
 import _ from 'lodash';
+import type { Profile } from '../flow/profileTypes';
 
-export function getCookie(name) {
+export function getCookie(name: string) {
```
noisecapella: I think this should always return either a string or null

static/js/util/courseList.js
```diff
@@ -25,7 +25,18 @@ function asPercent(number) {
 /**
  * Determine React elements for the UI given a course status
  */
-export function makeCourseStatusDisplay(course: Object, now: moment = moment()): string|React$Element {
+export type Course = {
+  runs: Array<CourseRun>;
+  status: string;
+};
+export type CourseRun = {
```
noisecapella: I think most of these should be optional. Can you enable `// @flow` on `courseList_test.js` and fix the warnings for that file?

static/js/components/EmploymentForm.js
aliceriot: It infers the types for `entry` and `index`, note the change to the annotation for `jobRow` below (which was caught by flow detecting a mismatch between `index: number` and `index: string`).

static/js/util/validation.js
```diff
@@ -11,14 +11,16 @@ import type { UIState } from '../reducers/ui';
 import { filterPositiveInt } from './util';
 import { HIGH_SCHOOL, EDUCATION_LEVELS } from '../constants';
 
-let handleNestedValidation = (profile, keys, nestedKey) => {
+let handleNestedValidation = (profile: any, keys, nestedKey) => {
```
noisecapella: Does `profile: Profile` work here?

static/js/components/EducationForm.js
noisecapella: Yeah, it seems to work just fine. Where did you find out about that? It seems like the flow docs have many nooks and crannies I haven't discovered

static/js/components/EducationForm.js
noisecapella: I think the problem stems from `profile` not being inferable. Flow doesn't use `propTypes`, we have to define it separately using `props: { ... }` on the class

static/js/util/validation.js
aliceriot: I wasn't able to get it to work with anything I tried except `any` here - I think the confusing is because it doesn't know if at runtime we'll be accessing the keys on an `EducationEntry` or on a `WorkHistoryEntry', if that makes sense.

static/js/components/EducationForm.js
aliceriot: They have some examples in the object type section of the docs: https://flowtype.org/docs/objects.html#_

mainly trial and error though :P

static/js/components/EducationForm.js
aliceriot: Flow does support `propTypes`: https://flowtype.org/docs/react.html I think it's just a bit limited, we're limited to the things defined by `React.PropTypes`.

aliceriot: I'll add a `props` annotation here.

static/js/components/EmploymentForm.js
noisecapella: Hmm, flow doesn't seem to detect any problem if I change `index` from `number` to `string` for `jobRow()`. Anyway you're right that it should be able to infer this

static/js/util/validation.js
noisecapella: Yeah, I see what you mean

static/js/components/EmploymentForm.js
aliceriot: huh, that's odd, if I change `index: number` to `index: string` on `jobRow` I get a type mismatch warning...

aliceriot: why flow why

static/js/util/validation.js
aliceriot: actually got it to work! adding `nestedKey: string` gets `profile: Profile` to work correctly.

noisecapella:
Rebase please :)

aliceriot:
So I just pushed up a bunch of additional annotations, and we're now doing proper `props` validations on a bunch of things.

noisecapella:
Functionality still functions :+1:

PR #624, opened 2016-06-22
Fixed style regression

### LINKED ISSUE #620: Bug: program page styling regression
As of 6/22 the program page is looking like this:

https://drive.google.com/file/d/0B9nbgzSIkR7mUHJZcWxid3VYNEE/view?usp=drivesdk

Yesterday it looked like this (the way it should):

https://drive.google.com/file/d/0B9nbgzSIkR7mUHdUNGVQS1Rzanc/view?usp=drivesdk

noisecapella:
Fixes #620 

#### What's this PR do?
Moves the react-mdl CSS back into dashboard.js. It overrides many common style rules which is why we're seeing different styles with it included. We're only using bootstrap CSS on the 404 page currently so this shouldn't cause any problems.

First, turn on DEBUG mode so we can view the 404 page. You can do this by setting `DEBUG=False` in `settings.py` and adding your host name to the `ALLOWED_HOSTS` list.

Then run `NODE_ENV=production ./webpack_if_prod.sh` in your terminal to generate static assets. Next, run `docker-compose run web ./manage.py collectstatic` to copy static assets into `staticfiles/`.

After that, look at the two screenshots for the program page in #620. Go to a program page and verify that it looks like it did before, more or less. Go to `/xyz/` and verify the 404 page still works as expected, and also click the dropdown and click 'logout' to make sure that works.

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -9,6 +9,7 @@ import { makeDashboardRoutes } from './dashboard_routes';
 
 // requirements for react-mdl which uses a modified version of material-design-lite
 import 'react-mdl/extra/material.js';
+import 'style!css!react-mdl/extra/material.css';
 
 // Object.entries polyfill
 import entries from 'object.entries';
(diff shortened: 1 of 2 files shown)


giocalitri:
@noisecapella just FYI, at the addresses `/404/` and `/500/` you can see the error pages with `DEBUG=True`

giocalitri:
👍

PR #626, opened 2016-06-23
Added resumeOrder to education entries

### LINKED ISSUE #583: Sort education entries on user and profile pages

aliceriot:

closes #583 

#### What's this PR do?

Adds a little function to sort education entries by `graduation_date` in resume order, which we can use in `EducationDisplay` and `EducationForm`.

Make sure that:

- on the user page all education entries are presented in a single list in resume order, and you can add/delete/edit entries as usual.
- on `/profile/education` make that that, for each degree level, the degree entries within that level are sorted by resume order, and, again, that you can do all the normal actions on entries.

### DIFF
--- a/static/js/containers/UserPage_test.js
+++ b/static/js/containers/UserPage_test.js
@@ -30,7 +30,8 @@ import {
 } from '../util/util';
 import IntegrationTestHelper from '../util/integration_test_helper';
 import * as api from '../util/api';
-import { USER_PROFILE_RESPONSE, HIGH_SCHOOL } from '../constants';
+import { USER_PROFILE_RESPONSE, HIGH_SCHOOL, DOCTORATE } from '../constants';
+import { workEntriesByDate, educationEntriesByDate } from '../util/sorting';
 
 describe("UserPage", function() {
   this.timeout(5000);
@@ -48,6 +49,42 @@ describe("UserPage", function() {
     return [...document.getElementsByClassName('deletion-confirmation')].find(dialog => (
       dialog.style["left"] === "0px"
     ));
+  };
+
+  const confirmResumeOrder = (
+    editButton,
+    profileProperty,
+    sortFunc,
+    editActions,
+    dialogIndexProperty
+  ) => {
+    let state = helper.store.getState();
+    let sorted = sortFunc(state.profiles[SETTINGS.username].profile[profileProperty]);
+
+    // sorted entries should not equal unsorted entries
+    assert.notDeepEqual(
+      state.profiles[SETTINGS.username].profile[profileProperty],
+      sorted.map(([,entry]) => entry)
+    );
+
+    return listenForActions(editActions, () => {
+      TestUtils.Simulate.click(editButton);
+    }).then(() => {
+      state = helper.store.getState();
+      let stateIndex = state.ui[dialogIndexProperty];
+      let [sortIndex, sortEntry] = sorted[0];
+      // the dialog index dispatched to the store should be the same index
+      // as the index (into the unsorted list) of the first element in our sorted list
+      // since we clicked on the first entry in the UI
+      assert.equal(stateIndex, sortIndex);
+      // this index should not be equal to 0, since the first element in the sorted list
+      // should not be the first item in the unsorted list
+      assert.notEqual(stateIndex, 0);
+      // the entry the index in the state points to should be the same element
+      // in the unsorted array that we have in the sorted array
+      let entries = state.profiles[SETTINGS.username].profile[profileProperty];
+      assert.deepEqual(sortEntry, entries[stateIndex]);
+    });
   };
 
   describe ("Authenticated user page", () => {
@@ -77,10 +114,49 @@ describe("UserPage", function() {
           getElementsByClassName('delete-button')[0];
       };
 
+      beforeEach(() => {
+        let userProfile = Object.assign({}, USER_PROFILE_RESPONSE, {
+          username: SETTINGS.username
+        });
+        userProfile.education.push({
+          "id": 3,
+          "degree_name": DOCTORATE,
+          "graduation_date": "2015-12-01",
+          "field_of_study": "Philosophy",
+          "school_name": "Harvard",
+          "school_city": "Cambridge",
+          "school_state_or_territory": "US-MA",
+          "school_country": "US",
+          "online_degree": false
+        });
+        helper.profileGetStub.
+          withArgs(SETTINGS.username).
+          returns(Promise.resolve(userProfile));
+      });
+
       it('shows the education component', () => {
         return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
           let title = div.getElementsByClassName('profile-card-title')[1];
           assert.equal(title.textContent, 'Education');
+        });
+      });
+
+      it('should show the entries in resume order', () => {
+        return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
+          let editButton = div.querySelector('#education-card').
+            querySelector('.edit-button');
+
+          return confirmResumeOrder(
+            editButton,
+            'education',
+            educationEntriesByDate,
+            [
+              SET_EDUCATION_DIALOG_INDEX,
+              SET_EDUCATION_DIALOG_VISIBILITY,
+              SET_EDUCATION_DEGREE_LEVEL,
+            ],
+            'educationDialogIndex'
+          );
         });
       });
 
@@ -230,6 +306,25 @@ describe("UserPage", function() {
         return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
           let title = div.getElementsByClassName('profile-card-title')[0];
           assert.equal(title.textContent, 'Employment');
+        });
+      });
+
+      it('should show the entries in resume order', () => {
+        return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
+          let editButton = div.getElementsByClassName('profile-tab-card')[0].
+            getElementsByClassName('profile-row-icons')[0].
+            getElementsByClassName('mdl-button')[0];
+
+          return confirmResumeOrder(
+            editButton,
+            'work_history',
+            workEntriesByDate,
+            [
+              SET_WORK_DIALOG_INDEX,
+              SET_WORK_DIALOG_VISIBILITY
+            ],
+            'workDialogIndex'
+          );
         });
       });
 
--- a/static/js/util/sorting.js
+++ b/static/js/util/sorting.js
@@ -1,26 +1,35 @@
 // @flow
 import moment from 'moment';
+import _ from 'lodash';
 
-import type { WorkHistoryEntry } from '../flow/profileTypes';
+import type { WorkHistoryEntry, EducationEntry } from '../flow/profileTypes';
 
-export function resumeOrder(entries: WorkHistoryEntry[], dateFieldName: string): WorkHistoryEntry[] {
-  let sortFunc = (a, b) => {
-    let adate = moment(a[dateFieldName]);
-    let bdate = moment(b[dateFieldName]);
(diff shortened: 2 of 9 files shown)


static/js/util/sorting.js
```diff
@@ -19,9 +20,13 @@ export function resumeOrder(entries: SortableEntry[], dateFieldName: string): So
...
-export function sortWorkEntriesByDate(entries: Array<WorkHistoryEntry>): Array<WorkHistoryEntry> {
+export function workEntriesByDate(entries: Array<WorkHistoryEntry>): Array<WorkHistoryEntry> {
   let sorted = [];
   sorted.push(...resumeOrder(entries.filter(entry => entry.end_date === null), 'start_date'));
   sorted.push(...resumeOrder(entries.filter(entry => entry.end_date !== null), 'end_date'));
   return sorted;
 }
+
+export function educationEntriesByDate(entries: Array<EducationEntry>): Array<EducationEntry> {
+  return resumeOrder(_.cloneDeep(entries), 'graduation_date');
```
noisecapella: It's generally bad practice to modify arguments passed into a function unless explicitly documented somehow. Can you change `resumeOrder` to do the list copy there and remove the `_.cloneDeep` here?

static/js/util/sorting.js
```diff
@@ -1,9 +1,10 @@
 // @flow
 import moment from 'moment';
+import _ from 'lodash';
 
-import type { WorkHistoryEntry } from '../flow/profileTypes';
+import type { WorkHistoryEntry, EducationEntry } from '../flow/profileTypes';
 
-export function resumeOrder(entries: WorkHistoryEntry[], dateFieldName: string): WorkHistoryEntry[] {
+export function resumeOrder(entries: any, dateFieldName: string): any {
```
noisecapella: Does `Array<WorkHistoryEntry|EducationEntry>` work as a syntax for flow?

aliceriot: I think you can declare an array of union types, but I wasn't able to get that to work because then flow can't make guarantees about the type of the return value, and it gives all these errors about not finding the properties on `WorkHistoryEntry` on an `EducationEntry`.

static/js/util/sorting.js
aliceriot: yah that makes sense to me

static/js/util/sorting.js
```diff
@@ -1,9 +1,11 @@
...
+  let clone = _.cloneDeep(entries);
```
noisecapella: Can you change this to a `_.clone` instead? It looks like we don't need to do a deep clone since we don't modify any of the items in the array, just the array itself.

noisecapella:
I found a bug with both employment and education sorting. It looks like the edit and delete buttons do not correspond to the items being shown anymore. Can you look into this as part of this PR? To reproduce, add an education item for a year, then add another education item for some later period of time. When you edit one of these two education items you get the other one's data

aliceriot:
Ah, yeah, I see the problem - we're using the sort when we display things, and taking the indices for all our editing work from the sorted indices, but that isn't necessarily (and probably isn't most of the time) the order that they'll be in on the profile object, so we're indexing in to the wrong object.

aliceriot:
@noisecapella I fixed it, I changed `resumeOrder` so it now returns an array of `[index, entry]` tuples, where `entry` is an entry in the original array of objects, passed in, `index` is the index of that object in the original array before sorting, and the tuples are sorted based on the date fields on `entry`.

does that make sense? I need to fix the tests but I think it should work.

noisecapella:
Yes, feels like that should work

aliceriot:
cool, I just fixed the tests.

noisecapella:
Can you also add to or update one of the tests in `UserPage_test.js` to cover this case, to assert that the index being sent via `SET_WORK_DIALOG_INDEX` for example is the same as the index of the item being clicked?

aliceriot:
Yes, I think that's a good idea. I added a few more unit tests to `util/sorting_test.js`, I may spend a bit of time trying to think of more edge cases there.

aliceriot:
@noisecapella I added some more tests

static/js/util/sorting.js
```diff
@@ -1,10 +1,12 @@
 // @flow
```
gsidebo: a couple general comment: 

1. i realize this is a somewhat common practice in the codebase right now, but i'm not a fan of us putting specific functionality in a 'util/' directory. if you ask me, these aren't utility functions - they're used by work history and education specifically. any reason that `workEntriesByDate` and `educationEntriesByDate` wouldn't be defined in EmploymentForm.js and EducationDisplay.js respectively? looks like they're the only clients of those functions and i can't see why there would be more
1. can you explain the name `resumeOrder`? it looks like a pretty general sort-by-date function. is it just a legacy name?

static/js/util/sorting.js
```diff
@@ -15,12 +17,17 @@ export function resumeOrder(entries: WorkHistoryEntry[], dateFieldName: string):
     }
     return 0;
   };
-  return entries.sort(sortFunc);
+  return clone.sort(sortFunc);
```
gsidebo: this can be abstracted further, eg:

```javascript
dateCompare (a, b) {
  if ( a.isBefore(b) ) {
    return 1;
  }
  if ( a.isAfter(b) ) {
    return -1;
  }
  return 0;
}

export function resumeOrder(entries: [number, Object][], dateFieldName: string): any {
  let clone = _.clone(entries);
  let sortFunc = ([, a], [, b]) => {
    return dateCompare(moment(a[dateFieldName]), moment(b[dateFieldName]));
  }
  return clone.sort(sortFunc);
}
```

static/js/util/sorting.js
aliceriot: I think that ease of testing / separation from the react end of things is one reason to put functions like this in a separate file. w/r/t to sorting the work history entries that would only live on one component, but we use two different components to display the education stuff (`EducationDisplay` and `EducationForm`) so I'd need to have some way of sharing the function between those two components in any case.

aliceriot: `resumeOrder` is what this ordering was referred to on the issue, I just went with that.

static/js/containers/UserPage_test.js
```diff
@@ -77,13 +78,64 @@ describe("UserPage", function() {
...
+      });
+
       it('shows the education component', () => {
         return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
           let title = div.getElementsByClassName('profile-card-title')[1];
           assert.equal(title.textContent, 'Education');
         });
       });
 
+      it('should show the entries in resume order', () => {
```
gsidebo: - these two tests are very repetitious. basically identical structure with a few variables swapped here and there. wondering if there is something we can abstract here. 
- i'm seeing these tests for the first time, and it's not easy from the outset to figure out what they're actually doing. "education dialog index should be the first entry in sorted" is particularly vague. can you think of some comments to add/edit that would make it clearer to a first-timer?

static/js/util/sorting.js
gsidebo: doesn't seem like that name applies anymore. `sortEntriesByDate` or something would be more on the mark, i think

aliceriot: kk, I'm going to do `dateOrderDesc` or something along those lines.

static/js/containers/UserPage_test.js
aliceriot: yeah, we can probably abstract some of this - I tend not to do that for stuff like this because I tend to think over-abstraction in testing can make tests much harder to read.

I'll add some more comments to explain what we're testing more.

gsidebo: in my experience, repetition in test code carries the exact same problems as repetition in 'real' app code. if nothing else, splitting some of these lines into functions gives us the opportunity to express what a block of code is doing via the function name. if done right, that would make it easier to read for the uninitiated. as this reads right now, i would need to run this with console.log lines/breakpoints to figure out what it's doing and why. probably not essential for all of our test code to be immediately transparent upon first read, but i think we move a little more in that direction here

aliceriot: hm ok I think I see what you mean, I'll give it a go

gsidebo: yeah the `then` handler function can be refactored to something like this:

```javascript
        let testDialogIndexStuff = (sortedEntries, unsortedEntries, dialogIndexKey) => {
          state = helper.store.getState();
          let stateIndex = state.ui[dialogIndexKey];
          let [sortIndex, sortEntry] = sortedEntries[0];
          // the work dialog index dispatched to the store should be the same index
          // as the index (into the unsorted list) of the first element in our sorted list
          // since we clicked on the first work entry in the UI
          assert.equal(stateIndex, sortIndex);
          // this index should not be equal to 0, since the first element in the sorted list
          // should not be the first item in the unsorted list
          assert.notEqual(stateIndex, 0);

          // the entry the index in the state points to should be the same element 
          // in the unsorted education that we have in the sorted array\
          assert.deepEqual(sortEntry, unsortedEntries[stateIndex]);
        };
```

You could throw the `assert.notDeepEqual` stuff in there too if you wanted. the body of `renderComponent` would then look something like this:

```javascript
let state = helper.store.getState();
let unsortedEntries = state.profiles[SETTINGS.username].profile.education;
let sortedEntries = educationEntriesByDate(unsortedEntries);
let dialogActionHandler = _.partial(testDialogIndexStuff, sortedEntries, unsortedEntries, 'educationDialogIndex')

return listenForActions([
  SET_EDUCATION_DIALOG_INDEX,
  SET_EDUCATION_DIALOG_VISIBILITY,
  SET_EDUCATION_DEGREE_LEVEL,
], () => {
  TestUtils.Simulate.click(editButton);
}).then(dialogActionHandler);

```

maybe a bit quirky with the `_.partial`. you could also wrap the entire `listenForActions` bit and pass in the expected actions list in addition to the other arguments. either way, i think those are definitely preferable to two distinct test cases that test for the same type of functionality and are essentially copy/pasted

static/js/containers/UserPage_test.js
```diff
@@ -233,6 +291,43 @@ describe("UserPage", function() {
...
```
gsidebo: please change the references to 'resume order' in the `it()` statements here and above

static/js/util/sorting_test.js
```diff
@@ -2,27 +2,47 @@
...
+  educationEntriesByDate,
+} from './sorting';
+import { generateNewWorkHistory, generateNewEducation } from './util';
+import { HIGH_SCHOOL } from '../constants';
 
 let format = 'YYYY-MM';
 
 describe('profile sort functions', () => {
-  it('should sort by resume order', () => {
+  describe('resume order', () => {
```
gsidebo: 'resume order' again

static/js/containers/UserPage_test.js
aliceriot: I think it's actually correct here - the unit tests for just the sorting function below are testing whether we can sort just based on date order, descending.

But at least for work history we're doing something a little more complicated, with the 'current' entries first, sorted by date descending, and then the non-current entries, sorted by date descending. It makes sense to me to keep the 'resume order' nomenclature here.

aliceriot:
@gsidebo @noisecapella this is updated, I refactored some of how the sorting is happening and the tests.

static/js/util/sorting.js
```diff
@@ -1,26 +1,35 @@
...
-    let adate = moment(a[dateFieldName]);
-    let bdate = moment(b[dateFieldName]);
-    if ( adate.isBefore(bdate) ) {
-      return 1;
-    }
-    if ( adate.isAfter(bdate) ) {
-      return -1;
-    }
-    return 0;
+export function momentSortDesc(a: moment$Moment, b: moment$Moment): number {
```
gsidebo: this function isn't actually sorting. it acts as a comparator. it's a small enough function to read, but all the same i would prefer not to have a misleading name. `momentDescComparator` et al

gsidebo:
LGTM! 👍

PR #628, opened 2016-06-23
Fixed filtering text to remain if textbox clicked

### LINKED ISSUE #547: Select field - clicking field during partial entry ignores filter
To reproduce:

1. Focus on any select field (eg: the Country field under 'Currently Living' in the Personal profile tab)
1. Type a partial entry to bring up autocomplete results (eg: 'Al' to suggest Albania, Algeria)
1. Without selecting an option, click on the active select field input

**Expected:** Autocomplete results would still be filtered
**Actual:** The partial entry is ignored in terms of option filtering - all possible options are shown

noisecapella:
Fixes #547 

#### What's this PR do?
When the user clicks an autocomplete field, it should show all options. If the user types text it filtered based on that text, but when they click again it shows all options. This PR changes the behavior to continue to show filtered results in this case.

Go to the /users/ page and find an AutoComplete field. Then look for these scenarios:
 - Clear the select field, focus away from field. Click on field again and make sure you see all entries. Type text and verify that it shows only entries whose prefix matches the text. Click on the text field and verify that the options you see haven't changed.
 - Select an option and focus away from field. Click on the text field and verify that you see all options. Click again and you should see only the filtered fields.
 - Select an option and focus away from the field. Click on the text field and delete all text, verify that you see all options. Click on the text field again and verify that you still see all options.

### DIFF
--- a/static/js/components/AutoComplete.js
+++ b/static/js/components/AutoComplete.js
@@ -370,9 +370,16 @@ class AutoComplete extends Component {
       });
     }
 
+    let skipFilter = this.props.showOptionsWhenBlank;
+    // We want to open the popup with all results if available, else
+    // if the popup is already open we want to filter as usual
+    if (this.state.searchText !== '' && this.state.searchText !== undefined && this.state.open) {
+      skipFilter = false;
+    }
+
     this.setState({
       focusTextField: true,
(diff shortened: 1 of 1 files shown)


aliceriot:
functionality looks good!

aliceriot:
code too! :+1:

PR #629, opened 2016-06-23
Redirected to 404 if user goes to a missing user page

### LINKED ISSUE #481: handle URLs that react-router doesn't have a route for in a better way
Currently if you mistakenly visit e.g. `/profiles/:username}` this is basically what you see:

error_react

there's a console warning that says `Warning: [react-router] Location "/profile/alicepote" did not match any routes`, which is the only user-facing indication of what's going on. We should figure out a better way to handle this (showing a `404` page, redirecting to a legit url, etc).

In this case, we just redirect the user to the 404 page. But we need a 404 page (see issue #440)

### LINKED ISSUE #630: Anonymous users cannot see the the /users/<username> page
They get redirected to the home page

noisecapella:
Fixes #481
Fixes #630 

#### What's this PR do?
Fixes two problems:
 - If a user goes to `/user/missing` (or some user who is invisible due to permissions), they will get a 404 page
 - Anonymous users may now look at the users page for a particular user

ProfileFormContainer.js

 - Go to `/users/<name>` with a name that doesn't match an account on the server, and verify that it shows the 404 page
 - Log out, then view that users page again. You should see a users page with a sign in button, with no ability to edit fields.

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -7,12 +7,16 @@ import logging
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.contrib.staticfiles.templatetags.staticfiles import static
+from django.views.generic import View
 from django.shortcuts import (
     render,
+    Http404,
 )
+from django.utils.decorators import method_decorator
 
 from backends.edxorg import EdxOrgOAuth2
 from micromasters.utils import webpack_dev_server_host, webpack_dev_server_url
+from profiles.permissions import CanSeeIfNotPrivate
 from ui.decorators import (
     require_mandatory_urls,
 )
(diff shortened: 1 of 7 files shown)


giocalitri:
Is there a way to return a 404 page keeping the failing URL? IMHO we are using different patterns for this: either we always redirect to the 404 page or we always keep the url and show the 404 content...

If we think this behavior is acceptable, then 👍

giocalitri:
cc @roberthouse54

noisecapella:
We could grab the HTML from the /404/ page with an AJAX request and display that to the user. That would fix the URL issue but the user wouldn't get the 404 status code anymore when they visit `/users/...` from outside the application. To fix that we could add a check for the user when serving the HTML for the users page

giocalitri:
I am not sure about what's the best solution

@pdpinch ?

noisecapella:
I talked with @giocalitri during lunch and we will just handle this case when the server does the initial page load, and require all links into user pages to be regular links, not react-router links. Thoughts @aliceriot?

aliceriot:
That sounds fine to me, but does that mean that we'll have a full page refresh if I click on the link to my profile from elsewhere in the dashboard?

noisecapella:
Yes. We could also handle react router links by listening for the 404 from the rest API and fetching the /404/ page to display its HTML but that feels clumsier to me

aliceriot:
Yeah, that sounds like it could be a bit fragile and edge case-ey.

noisecapella:
Comments are addressed, PR is pretty much a rewrite at this point so you should look over all of it again

noisecapella:
Actually it looks like I neglected the UI for the anonymous user page

noisecapella:
Added some JS tests and a sign in button

ui/views.py
```diff
@@ -33,22 +36,25 @@ def get_bundle_url(request, bundle_name):
         return static("bundles/{bundle}".format(bundle=bundle_name))
 
 
-@require_mandatory_urls
-@login_required()
-def dashboard(request, *args):  # pylint: disable=unused-argument
+def _dashboard(request, *args):  # pylint: disable=unused-argument
```
giocalitri: considering this view is serving several sub-views, we might want to rename it

giocalitri: and maybe refactoring to a class-based view so that the others can inherit this and add their own changes (but feel free to pospone this big refactoring)

giocalitri:
one minor comment. checking functionality

giocalitri:
functionality looks good: 👍

PR #641, opened 2016-06-27
prevented non-numbers from being entered in date fields

### LINKED ISSUE #585: Date fields should prevent non-numbers from being typed
Also limit number of characters allowed

### LINKED ISSUE #586: Graduation date field should be formatted consistently
Since we pad the month with a 0 in the graduation date text, the field should show the same padding

Graduation date text:
screenshot from 2016-06-23 17-48-52

Field:
screenshot from 2016-06-23 17-49-40

aliceriot:

closes #585 and closes #586 

#### What's this PR do?

This changes the text fields in `boundDateField` so that they only accept characters that can comprise valid inputs. So, no non-numerical characters, no characters beyond the number required for the field in question, etc.

Review the changes to `profile_edit.js` and `validation.js`. 

Make sure you can edit and save valid dates for all date fields we have.

Also, check for the following behavior, which this PR implements:

month field:

- should accept only 2 characters
- should not accept non-numerical characters
- should only allow input of 1-12
- input of numbers larger than 12 should correct to 12

day field:

- should accept only 2 characters
- should not accept non-numerical characters
- should only allow input of 1-31
- input of numbers larger than 31 should correct to 31

year field:

- should accept 4 characters max
- should not accept non-numerical characters
- should accept numbers below 1800 as input, but they should be marked invalid on save
- should accept any number `x | 1800 <= x <= 2100` as valid input
- input of numbers larger than 2100 should correct to 2100

### DIFF
--- a/static/js/util/validation.js
+++ b/static/js/util/validation.js
@@ -246,28 +246,49 @@ export function validateProfileComplete(profile: Profile): ProfileComplete {
 /**
  * Validate a day of month
  */
-export function validateDay(string: string): ?number {
-  let date = filterPositiveInt(string);
+export function validateDay(input: string): ?number {
+  let sanitized = sanitizeDate(input, 2);
+  let date = filterPositiveInt(sanitized);
   if (date === undefined) {
     return undefined;
   }
   // More complicated cases like Feb 29 are handled in moment.js isValid
-  if (date < 1 || date > 31) {
-    return undefined;
+  if (date > 31) {
+    return 31;
   }
   return date;
 }
 
 /**
+ * Removes non-numeric characters and truncates output string
+ */
+export function sanitizeDate(input: string|number, length: number): string {
+  if ( typeof input === 'string' ) {
+    let out = input.replace(/[^\d]+/g, '');
+    if ( out.match(/^0+/) ) {
+      if ( out.length <= length ) {
+        return out.slice(0, length);
+      } else {
+        return out.replace(/^0+/, "").slice(0, length);
+      }
+    } else {
+      return out.slice(0, length);
+    }
+  } else {
+    return String(input).slice(0, length);
+  }
+}
+/**
  * Validate a month number
  */
-export function validateMonth(string: ?string): ?number {
-  let month = filterPositiveInt(string);
+export function validateMonth(input: string|number): number|void {
+  let sanitized = sanitizeDate(input, 2);
+  let month = filterPositiveInt(sanitized);
   if (month === undefined) {
     return undefined;
   }
-  if (month < 1 || month > 12) {
-    return undefined;
+  if (month > 12) {
+    return 12;
   }
   return month;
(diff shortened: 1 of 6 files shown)


noisecapella:
Can you change the PR title to the past tense?

noisecapella:
If the month field is empty I can enter in any number of non-numeric characters. It only seems to do the check if there is a number in the field already

aliceriot:
@noisecapella I fixed the bug with empty fields and added left-padding on month and day values. I don't think there's a good way to add left-padding on values as the user is typing, so it will only pad when there's no `edit` object on the profile for the field in question (meaning the field will be left padded when you open the edit form for, say, work history, but editing the field will erase the left padding and allow you to edit the number freely.

noisecapella:
I'm seeing an edge case during testing: 
 - Go to a field where the date was already populated. Let's say the year is '1985'.
 - Move the cursor to after the '9' and hit backspace. The cursor will be moved to the end of the string. The string will now be '185' which is as expected

noisecapella:
It looks like that edge case is due to `1985` being passed to the component as a number but the edited values being strings

noisecapella:
Also can you add a `closes #xxx` for the other issue that's also fixed?

aliceriot:
oh huh, I edited the pr comment yesterday to add that but it didn't save, strange.

aliceriot:
@noisecapella it looks like the bug with the cursor jumping to the end is present on master, and on all of our text fields (not just date fields).

So big :( there, do you think we should try to fix it here or open a new issue?

noisecapella:
Ah, let's just open a new issue

aliceriot:
kk will do

static/js/util/validation.js
```diff
@@ -246,44 +246,71 @@ export function validateProfileComplete(profile: Profile): ProfileComplete {
...
   // More complicated cases like Feb 29 are handled in moment.js isValid
-  if (date < 1 || date > 31) {
-    return undefined;
+  if (date > 31) {
+    return 31;
   }
   return date;
 }
 
+export function sanitizeDate(input: string|number, length: number): string {
```
noisecapella: Can you add a JSDoc comment here?

static/js/util/validation_test.js
```diff
@@ -320,29 +321,92 @@ describe('Profile validation functions', () => {
...
+      });
+
+      it('should remove leading zeros when over the length', () => {
+        assert.equal(sanitizeDate('01999', 4), '1999');
+      });
+    });
+
+    describe('numerical input', () => {
+      it('should return a string', () => {
+        assert.deepEqual(sanitizeDate(3, 1), '3');
```
noisecapella: Feel free to leave it but we don't need `deepEqual` for strings

noisecapella:
:+1: after adding the JSDoc

PR #649, opened 2016-06-27
Added user page link to dropdown

### LINKED ISSUE #571: Add users page link in user dropdown in header in navbar
This will need to be done on the 404 page as well. (base_error.html)

gsidebo:

Closes #571 

#### What's this PR do?

Adds a link in the top right menu to the user's profile page

Log in, click the top-right menu arrow, use the Profile link

### DIFF
--- a/static/js/containers/UserMenu.js
+++ b/static/js/containers/UserMenu.js
@@ -8,7 +8,7 @@ import MenuItem from 'react-bootstrap/lib/MenuItem';
 
 import { getPreferredName } from '../util/util';
 
-class LoginButton extends React.Component {
+class UserMenu extends React.Component {
   static propTypes = {
     dispatch: React.PropTypes.func.isRequired,
     profile:  React.PropTypes.object.isRequired,
@@ -30,7 +30,12 @@ class LoginButton extends React.Component {
           <SplitButton
             title={title}
             bsStyle="danger"
-            id="logout-button">
+            id="user-menu">
+            <LinkContainer to={{ pathname: `/users/${profile.username}` }} active={false}>
+              <MenuItem>
+                Profile
+              </MenuItem>
+            </LinkContainer>
             <LinkContainer to={{ pathname: '/settings' }} active={false}>
               <MenuItem>
                 Settings
@@ -63,4 +68,4 @@ const mapStateToProps = (state) => {
   };
 };
 
-export default connect(mapStateToProps)(LoginButton);
+export default connect(mapStateToProps)(UserMenu);
--- a/static/js/containers/UserMenu.js
+++ b/static/js/containers/UserMenu.js
@@ -43,7 +43,8 @@ class UserMenu extends React.Component {
             </LinkContainer>
             <MenuItem
               href="/logout"
-              eventKey="logout">
+              eventKey="logout"
+              id="logout-link">
               Logout
             </MenuItem>
           </SplitButton>
--- a/ui/templates/base_error.html
+++ b/ui/templates/base_error.html
@@ -7,7 +7,7 @@
 {% block extrahead %}
 <script src="https://code.jquery.com/jquery-2.2.0.min.js"></script>
 <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js"
-	integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous"></script>
+    integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous"></script>
 {% endblock %}
 
 {% block content %}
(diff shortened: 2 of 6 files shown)


static/js/containers/UserMenu.js
```diff
@@ -28,8 +28,12 @@ class LoginButton extends React.Component {
       <LinkContainer to={{ pathname: '/dashboard' }} active={false}>
         <SplitButton
           title={title}
-          bsStyle="danger"
-          id="logout-button">
+          bsStyle="danger">
+          <MenuItem
+            href={`/users/${profile.username}`}
+            eventKey="users_page">
```
gsidebo: i don't really know what this `eventKey` property is intended for. the logout MenuItem has it but I didn't see what its purpose was

amir-qayyum-khan:
please rebase

static/js/containers/UserMenu.js
```diff
@@ -28,8 +28,12 @@ class LoginButton extends React.Component {
       <LinkContainer to={{ pathname: '/dashboard' }} active={false}>
         <SplitButton
           title={title}
-          bsStyle="danger"
-          id="logout-button">
+          bsStyle="danger">
+          <MenuItem
+            href={`/users/${profile.username}`}
```
amir-qayyum-khan: @gsidebo  do not need href

```js
 <LinkContainer to={{ pathname: `/users/${profile.username}` }} active={false}>
      <MenuItem>
           {profile.username}
      </MenuItem>
 </LinkContainer>
````

File: docker-compose.yml [hunk 3]
```diff
@@ -13,7 +13,7 @@ web:
   volumes:
     - .:/src
   environment:
-    DEBUG: 'True'
+    DEBUG: 'False'
```
amir-qayyum-khan: why DEBUG flag value is change?

static/js/containers/UserMenu.js
```diff
@@ -29,7 +29,12 @@ class LoginButton extends React.Component {
         <SplitButton
           title={title}
           bsStyle="danger"
-          id="logout-button">
+          id="user-menu">
+          <MenuItem
+            href={`/users/${profile.username}`}
+            eventKey="users_page">
```
amir-qayyum-khan: You dont need to use href 
```js
<LinkContainer to={{ pathname: `/users/${profile.username}` }} active={false}>
    <MenuItem>
         Profile
     </MenuItem>
</LinkContainer>
```

ui/templates/base_error.html
```diff
@@ -22,11 +22,12 @@
 	  		{% if authenticated %}
 	  		<div class="dropdown btn-group">
               <a class="btn btn-danger" href="/dashboard"><span>{{ name }}</span></a>
```
amir-qayyum-khan: what is the purpose base_error.html

gsidebo:
major issues with my machine right now. will need to make these changes later

static/js/containers/UserMenu.js
gsidebo: can i ask why this approach is preferred? the logout link uses href. the `LinkContainer` adds a good amount of code. what benefit do we get?

amir-qayyum-khan:
@gsidebo when i use DEBUG=false then ``docker-compose up`` was not working. Then i changed it to ``true`` to test ur work manually

gsidebo:
@amir-qayyum-khan that DEBUG change was pushed by accident. i reverted it

ui/templates/base_error.html
gsidebo: this file existed before the PR. it seems that it is here to serve an error page without running any react code. i'm not 100% certain why it's necessary. @aliceriot or @noisecapella might be able to explain it

ui/templates/base_error.html
```diff
@@ -7,7 +7,7 @@
 {% block extrahead %}
```
gsidebo: just noticed while i was making changes that this file used a mix of tabs and spaces. i changed it to spaces-only

static/js/containers/UserMenu.js
aliceriot: the `href` attribute on `MenuItem` will set a normal `<a>` html tag, which we actually don't want. Since we're using react-router on the frontend, we want to use something that will trigger a react-router route transition when clicked, instead of a full page refresh. `LinkContainer` is a little hack for doing just this with react-bootstrap UI components.

ui/templates/base_error.html
aliceriot: I...have no idea, haha. `git blame` suggests @giocalitri might though!

static/js/containers/UserMenu.js
aliceriot: the logout link uses `href` because we want that to hit the server.

ui/templates/base_error.html
giocalitri: the errors are served without using REACT because they are common to the CMS and made with templates only.
The `base_error.html` is the common base template that extends `base.html` and it is used by `404.html` and `500.html`

gsidebo:
@amir-qayyum-khan this is ready for review again whenever you can get to it

amir-qayyum-khan:
👍

PR #655, opened 2016-06-28
Removed logic for "upgrade" from the front end code

giocalitri:
#### What are the relevant tickets?
fixes #648 

#### What's this PR do?
Removes the logic of the upgrade to verified button from the JS code. This logic was based on a field never supplied by the backend. We will implement the correct logic in the backend with #654 

#### Where should the reviewer start?
`static/js/util/courseList.js`

#### How should this be manually tested?
if the dashboard rest API returns a `enrolled-not-verified` code, now the dashboard should show the UPGRADE button

PR #659, opened 2016-06-29
Added deadline for upgrade

### LINKED ISSUE #654: Honor upgrade_deadline when display upgrade to verified button on dashboard
Don't show the "upgrade to verified" button if the learner won't actually be able to upgrade. 

There is a deadline for upgrades that the course team will set in edX based on a date, after which, the user can no longer do well enough to get a verified certificate. When that date is passed, we should encourage learners to enroll in another run of the course, because they can't get micromasters credit any more. 

In v1 of the enrollment api, this field appears to be named `expiration_datetime`

In the catalog api, this field is named `upgrade_deadline` 

In either case, it appears to set to `null` frequently for audit tracks (understandably) but have a datetime for verified tracks.

giocalitri:
closes #654 

#### What's this PR do?
Introduces a new field in the Course run model to put a deadline to the upgrade to verified

`courses/models.py`

Enroll in a course without upgrading. Verify that the `enrolled-not-verified` in the dashboard REST API appears with a `upgrade_deadline` only in the future or Null

This will not work in the front-end until #655 is merged


dashboard/api_test.py
```diff
@@ -252,15 +255,15 @@ def setUpTestData(cls):
             cls.enrollments_json = json.loads(file_obj.read())
 
         cls.enrollments = Enrollments(cls.enrollments_json)
+        cls.now = datetime.now(pytz.utc)
 
     def test_status_for_run_not_enrolled(self):
         """test for get_status_for_courserun"""
```
noisecapella: I just noticed that all of these functions have the same docstring... it's outside the scope of this PR but would you mind updating them?

noisecapella:
Code looks good, testing functionality

noisecapella:
:+1:

PR #661, opened 2016-06-29
Added spinner and error message for profile page

### LINKED ISSUE #588: Profile page should handle in progress and error states
We should should show a spinner when the profile API is loading and an error box when profile API load has failed

noisecapella:
Fixes #588 

#### What's this PR do?
If the profile page is loading it should show a spinner while it loads. If the page is errored you should see an error alert box with the appropriate text.

static/js/containers/ProfilePage.js

To test the spinner, replace the function in `api.js` with this function:

    export function getUserProfile(username: string) {
        setTimeout(() => {
          resolve(mockableFetchJSONWithCSRF(`/api/v0/profiles/${username}/`));
        }, 3000);
      });
    }

This will cause a 3 second delay before any loading is done giving you time to see the spinner.

To test the error page, replace the same function with this code:

    import { ERROR_RESPONSE } from '../constants';
    export function getUserProfile(username: string) {
      return Promise.reject(ERROR_RESPONSE);
    }

This will cause the page to always fail loading.

screenshot from 2016-06-30 09-53-54
screenshot from 2016-06-30 09-55-37

### DIFF
--- a/static/js/containers/ProfilePage_test.js
+++ b/static/js/containers/ProfilePage_test.js
@@ -4,8 +4,11 @@ import { assert } from 'chai';
 import _ from 'lodash';
 
 import {
+  REQUEST_GET_USER_PROFILE,
+  RECEIVE_GET_USER_PROFILE_FAILURE,
   REQUEST_PATCH_USER_PROFILE,
   RECEIVE_PATCH_USER_PROFILE_SUCCESS,
+  RECEIVE_DASHBOARD_SUCCESS,
   START_PROFILE_EDIT,
   UPDATE_PROFILE_VALIDATION,
   CLEAR_PROFILE_EDIT,
@@ -25,6 +28,7 @@ import {
 } from '../actions/ui';
 import {
   USER_PROFILE_RESPONSE,
+  ERROR_RESPONSE,
   EDUCATION_LEVELS,
   ASSOCIATE,
   DOCTORATE,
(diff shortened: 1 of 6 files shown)


static/js/containers/ProfilePage_test.js
```diff
@@ -447,4 +451,41 @@ describe("ProfilePage", function() {
...
+  });
+
+  it('renders errors when there is an error receiving the profile', () => {
+    helper.profileGetStub.returns(Promise.reject(ERROR_RESPONSE));
+
+    const types = [RECEIVE_DASHBOARD_SUCCESS, RECEIVE_GET_USER_PROFILE_FAILURE];
+    return renderComponent("/profile", types, false).then(([, div]) => {
+      assert(div.getElementsByClassName('alert-message').length > 0, 'no alert message found');
+    });
+  });
```
giocalitri: Is there any value in adding tests to check the different error cases we have for the dashboard? (error code, user error message and so on)

giocalitri:
code looks fine: checking functionality

static/js/containers/ProfilePage_test.js
noisecapella: Ideally I would mock `ErrorInfo` and see that the right `props` were being passed to it, then move the `DashboardPage` tests which test the different error cases into `ErrorInfo` tests. Unfortunately we can't just mock `ErrorInfo` like we can in Python, JS mocks have some major limitations due to differences in how the languages work.

But we do get the component tree here so maybe there's a way forward. I'll investigate

PR #662, opened 2016-06-30
Added `ErrorMessage` to /users page

### LINKED ISSUE #591: Put error message on users page like on dashboard

aliceriot:

closes #591


aliceriot:
@noisecapella I didn't realize you were also working on a similar change, so I thought I would put this up now (although it's still a work-in-progress) to see if you had any feedback or if this conflicts with any changes you're making.

static/js/actions/index.js
```diff
@@ -29,9 +29,9 @@ export const receiveGetUserProfileSuccess = (username, profile) => ({
   payload: { profile, username }
 });
 
-const receiveGetUserProfileFailure = username => ({
+const receiveGetUserProfileFailure = (username, error) => ({
   type: RECEIVE_GET_USER_PROFILE_FAILURE,
-  payload: { username }
+  payload: { username, error }
```
noisecapella: Can you rename to `errorInfo` just for consistency with `receiveDashboardFailure`?

aliceriot: sure

static/js/components/Header.js
```diff
@@ -25,6 +26,7 @@ class Header extends React.Component {
               <img src="/static/images/logo-micromasters.png" width="215" height="40" alt="MIT MicroMasters" />
             </a>
           </Navbar.Brand>
+          <Link to="/users/foobar">FOOBAR</Link>
```
aliceriot: this is just for testing, to make it easier to visit a page where I'll get a 404

noisecapella:
Looks great!

Regarding tests I was originally hoping I could mock `ErrorInfo` and just verify that it was used in the render method but I couldn't figure out how to do that. Instead I just tested for the presence of the alert div and left the tests for `ErrorInfo` functionality in `DashboardPage_test.js`.

aliceriot:
OK, that sounds like a good approach to me. I was planning to just have the api mocking throw and error and check that the right error code and message are dumped onto the page.

aliceriot:
@noisecapella this is ready for review!

static/js/flow/profileTypes.js
```diff
@@ -52,4 +51,22 @@ export type Profile = {
...
+  profile?: Profile,
+  errorInfo?: APIErrorInfo,
+  getStatus: string,
+  edit?: Object;
+};
+
 export type BoundSaveProfile = (validator: Validator|UIValidator, profile: Profile, ui: UIState) => Promise<Profile>;
+
+export type APIErrorInfo = {
+  error_code: string,
```
noisecapella: I think everything except `errorStatusCode` should be optional

static/js/reducers/index_test.js
```diff
@@ -84,11 +84,16 @@ describe('reducers', () => {
...
+      let errorInfo = {
+        errorStatusCode: 404,
+        detail: "not found"
+      };
+      getUserProfileStub.returns(Promise.reject(errorInfo));
 
       return dispatchThen(fetchUserProfile('jane'), [REQUEST_GET_USER_PROFILE, RECEIVE_GET_USER_PROFILE_FAILURE]).
       then(profileState => {
         assert.equal(profileState['jane'].getStatus, FETCH_FAILURE);
+        assert.equal(profileState['jane'].errorInfo, errorInfo);
```
noisecapella: Should this be `deepEqual`? Not sure how chai asserts work for objects

aliceriot: oops, that was just a copypasta from the line above

static/js/util/util.js
```diff
@@ -249,6 +249,6 @@ export function calculateDegreeInclusions(profile: Profile) {
 /**
  * Calls an array of functions in series with a given argument and returns an array of the results
  */
-export function callFunctionArray(functionArray, arg) {
+export function callFunctionArray(functionArray: Function[], arg: any): any[] {
```
noisecapella: Is it possible to specify the type of `functionArray`? I'm thinking about something with type `A => R` with arg: `A` and the return value `R[]`

static/js/reducers/index_test.js
```diff
@@ -111,13 +116,15 @@ describe('reducers', () => {
...
-      patchUserProfileStub.returns(Promise.reject());
+      let errorInfo = {errorStatusCode: 500};
+      patchUserProfileStub.returns(Promise.reject(errorInfo));
 
       return dispatchThen(
         saveProfile('jane', USER_PROFILE_RESPONSE),
         [REQUEST_PATCH_USER_PROFILE, RECEIVE_PATCH_USER_PROFILE_FAILURE]
       ).then(profileState => {
         assert.equal(profileState['jane'].patchStatus, FETCH_FAILURE);
+        assert.equal(profileState['jane'].errorInfo, errorInfo);
```
noisecapella: Can you change this to `deepEqual` as well?

static/js/util/util.js
aliceriot: hmm maybe, I think we can do something like:

```js
callfunctionArray<T>(functionArray: Array<(): => T>, arg: T): T[]
```

maybe, I'm not super clear on how flow's type polymorphism works

static/js/util/util.js
```diff
@@ -249,6 +249,6 @@ export function calculateDegreeInclusions(profile: Profile) {
 /**
  * Calls an array of functions in series with a given argument and returns an array of the results
  */
-export function callFunctionArray(functionArray, arg) {
+export function callFunctionArray<T,R>(functionArray: Array<(t: T) => R>, arg: any): R[] {
```
aliceriot: @noisecapella this annotation type checks but I'm not sure it's quite correct, since this means that the `callFunctionArray` has type polymorphism, but every function passed in in the `functionArray` has the same argument and return types...

noisecapella: Each function needs to share a common argument type to work with `arg`. I guess they don't need to have the same return type necessarily but they do for the functions we're currently using. What do you think?

noisecapella:
Looks like there's a linting error on line 0? Other than that this looks good :+1:

noisecapella:
It looks like the flow declaration on `callFunctionArray` is confusing eslint...

noisecapella:
This looks to be fixed in the latest babel-eslint

aliceriot:
Cool, my bus missed a connection so I won't have an internet connection
probably for another hour, but when I do I can bump our eslint version.

noisecapella:
No worries, I'll take care of it

aliceriot:
I found a problem in our package.json which I think was interfering with the shrinkwrap working properly - we specific redux-asserts@0.0.8 in package.json but @0.0.7 in the shrinkwrap. I suspect this is the result of editing the package.json version instead of doing `npm install --save redux-asserts@latest` or similar.

I tried manually editing the shrinkwrap, but couldn't get that to work because of a error, I think I'm going to try to make a fresh npm shrinkwrap after installing all our dependencies only from the package.json, and see if I can get that to work.

noisecapella:
Sorry, I force pushed the changes already but they're what you described anyway. I also deleted npm-shrinkwrap.json and recreated it from package.json, hopefully this won't cause problems

aliceriot:
Ok, in general it would be good to have a heads up before force pushing to a remote branch someone else is also working on.

aliceriot:
looks like it breaks a few tests, I'll try to fix them

aliceriot:
@noisecapella I fixed the tests that were failing after the npm shrinkwrap change

static/js/containers/ProfilePage_test.js
```diff
@@ -73,23 +73,25 @@ describe("ProfilePage", function() {
...
-  let confirmSaveButtonBehavior = (updatedProfile, pageElements, validationFailure=false) => {
+  let confirmSaveButtonBehavior = (updatedProfile, pageElements, validationFailure=false, actions = []) => {
     let { div, button } = pageElements;
     button = button || div.querySelector(nextButtonSelector);
     patchUserProfileStub.throws("Invalid arguments");
     patchUserProfileStub.withArgs(SETTINGS.username, updatedProfile).returns(Promise.resolve(updatedProfile));
 
-    let actions = [];
-    if (!validationFailure) {
+    if ( actions === [] ) {
```
noisecapella: This won't compare since the `[]` and `actions` will always be different objects. Can you check the `length` instead?

aliceriot: oh oops yeah haha

static/js/containers/ProfilePage_test.js
```diff
@@ -79,7 +79,7 @@ describe("ProfilePage", function() {
     patchUserProfileStub.throws("Invalid arguments");
     patchUserProfileStub.withArgs(SETTINGS.username, updatedProfile).returns(Promise.resolve(updatedProfile));
 
-    if ( actions === [] ) {
+    if ( actions.lenth === 0 ) {
```
noisecapella: Typo here

noisecapella:
Looks good! :+1:

PR #689, opened 2016-07-05
Changed preferred name display

### LINKED ISSUE #529: Use preferred name and last name on dashboard header
The MicroMaster's site needs to feel official, not casual. We should use full names, not just first (given) names or preferred name. 

Requested by Erdin.

aliceriot:

this closes #529 

#### What's this PR do?

This changes the `getPreferredName` function from `util.js` so that it accepts a second positional argument (a boolean) that, if true, causes it to return the first and last name (provided the profile passed in has a last name on it).

Make sure you get first and last name on `/dashboard`, `/terms_of_service`, `/settings`, and `/profile`. If you set `last_name` to be `None` on your profile you should still see your first name in those places.

### DIFF
--- a/static/js/containers/UserMenu.js
+++ b/static/js/containers/UserMenu.js
@@ -21,7 +21,7 @@ class UserMenu extends React.Component {
     // and React 15. React 15 removed span tags but react-bootstrap still expects
     // them.
     let title = <span>
-      {getPreferredName(profile)}
+      {getPreferredName(profile, false)}
     </span>;
 
     if (SETTINGS.authenticated) {
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -225,8 +225,12 @@ export function makeProfileImageUrl(profile: Profile): string {
 /**
  * Returns the preferred name or else the username
  */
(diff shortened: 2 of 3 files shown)


giocalitri:
I have only one question: is there any place where you call `getPreferredName` with `last=false`?
If not I would remove the `last` parameter.

aliceriot:
yes, in the user menu in the top-right.

aliceriot:
haha, maybe that's an indication that `last` should default to `true`...

giocalitri:
yep :-)

aliceriot:
@giocalitri updated!

giocalitri:
👍 , but please check why the build on travis is failing

aliceriot:
ah I think npm is having some availability issues today.

aliceriot:
it looks like some of the JS dependencies didn't get pulled in. I'm experiencing the same issue locally when trying to install a package (the server returns a 502).

aliceriot:
I'll hold off on merging until it's sorted, I opened a help ticket with them but I think they may not yet be open for business.

aliceriot:
https://npmjs.statuspage.io/

PR #691, opened 2016-07-05
Fixed bug where validation errors showed up on first login

### LINKED ISSUE #634: After the first login the user gets validate errors in the profile
screen shot 2016-06-24 at 11 52 32 am

aliceriot:

closes #634 

#### What's this PR do?

This adds an additional check to the `requireCompleteProfile` method on `App.js`, which ensures that we don't get a race condition relating to the terms of service page.

The race condition emerges when a user is signing in to micromasters for the first time. They get shown the terms of service page, and when they agree the following happens:

- we dispatch an asynchronous action to `PATCH` the updated user profile (with `filled_out === true`)
- the `profiles.username` object is updated (`patchStatus: FETCH_PROCESSING`)
- the store change triggers `componentDidUpdate` on `App.js`
- we call the `requireCompleteProfile` method on `App.js` from the `componentDidUpdate` method
- the pathname is still `/terms_of_service` (remember that the asynchronous `PATCH` response hasn't yet come back)
- the `!PROFILE_REGEX.test(pathname)` test returns `true`, because `pathname === /terms_of_service`
- we validate the profile, and update the profile validation
- `PATCH` response comes back, and we execute the `.then` callback, which pushes `/profile` to the location
- since we already validated the profile, validation errors show up.

We can get around this by simply checking that we're not on `/terms_of_service` before we check and update the validation in the `requireCompleteProfile` method.

### DIFF
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -109,6 +109,7 @@ class App extends React.Component {
     if (
       userProfile.getStatus === FETCH_SUCCESS &&
       profile.agreed_to_terms_of_service &&
+      !TERMS_OF_SERVICE_REGEX.test(pathname) &&
       !PROFILE_REGEX.test(pathname) &&
       !complete
     ) {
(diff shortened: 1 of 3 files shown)


noisecapella:
:+1:

PR #703, opened 2016-07-05
Moved ErrorMessage tests into separate file

### LINKED ISSUE #663: Add unit tests for ErrorInfo, refactor other tests to test ErrorInfo props
We should move the unit tests from DashboardPage_test.js which test for different error cases into ErrorInfo. Then we just need to assert that DashboardPage, UserPage, and ProfilePage use ErrorInfo correctly by iterating through the component tree returned by `renderComponent` and asserting its props.

aliceriot:

this closes #663 

#### What's this PR do?

This moves existing tests for the `ErrorMessage` component into `static/js/components/ErrorMessage_test.js` and also adds some simple unit tests for the component.

Review the changes, make sure no tests got dropped in the shuffle.

### DIFF
--- a/static/js/components/ErrorMessage_test.js
+++ b/static/js/components/ErrorMessage_test.js
@@ -1,0 +1,220 @@
+/* global SETTINGS: false */
+import React from 'react';
+import TestUtils from 'react-addons-test-utils';
+import { assert } from 'chai';
+import _ from 'lodash';
+
+import {
+  RECEIVE_DASHBOARD_FAILURE,
+  RECEIVE_GET_USER_PROFILE_SUCCESS,
+  REQUEST_GET_USER_PROFILE,
+  RECEIVE_DASHBOARD_SUCCESS,
+  RECEIVE_GET_USER_PROFILE_FAILURE,
+  SET_USER_PAGE_DIALOG_VISIBILITY,
+  START_PROFILE_EDIT,
+  UPDATE_PROFILE_VALIDATION,
+  REQUEST_PATCH_USER_PROFILE,
+  RECEIVE_PATCH_USER_PROFILE_FAILURE,
+  CLEAR_PROFILE_EDIT,
+} from '../actions';
+import {
+  DASHBOARD_RESPONSE,
+  ERROR_RESPONSE,
+  USER_PROFILE_RESPONSE,
+} from '../constants';
+import IntegrationTestHelper from '../util/integration_test_helper';
+import { makeStrippedHtml } from '../util/util';
+import * as api from '../util/api';
+import ErrorMessage from './ErrorMessage';
+
+describe("ErrorInfo", () => {
+  let errorString = `Sorry, we were unable to load the data necessary
+      to process your request. Please reload the page.`;
+  errorString = errorString.replace(/\s\s+/g, ' ');
+
+  let contactExpectation = `If the error persists, please contact
+      mitx-support@mit.edu specifying this entire error message.`;
+  contactExpectation = contactExpectation.replace(/\s\s+/g, ' ');
+
+  describe('unit tests', () => {
+    let renderErrorMessage = props => {
+      return makeStrippedHtml(<ErrorMessage {...props} />);
+    };
+    let codeAttributes = [
+      ['error_code', '500'],
+      ['errorStatusCode', 404],
+    ];
+    let messageAttributes = [
+      ['user_message', 'A message'],
+      ['detail', 'Some details'],
+    ];
+
+    codeAttributes.forEach(([codeAttribute, code]) => {
+      messageAttributes.forEach(([msgAttribute, message]) => {
+        it(`should render an error and message on the ${codeAttribute} and ${msgAttribute} attributes`, () => {
+          let that = {
+            errorInfo: {
+              [codeAttribute]: code,
+              [msgAttribute]: message,
+            }
+          };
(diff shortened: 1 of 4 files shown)


aliceriot:
@noisecapella I picked up the issue you had opened, not sure if this is what you had in mind. If you were thinking of adding different sorts of tests LMK!

static/js/components/ErrorMessage_test.js
```diff
@@ -0,0 +1,214 @@
...
+          error_code: '500',
+          user_message: 'A message',
+          detail: 'Some details',
+          errorStatusCode: 404,
+        }
+      };
+    });
+
+    it('should take an errorInfo prop', () => {
+      let errorInfo = <ErrorMessage {...that}/>;
```
noisecapella: It looks like we're just asserting that the props we passed in were passed in, since there's no render step here.

aliceriot: yeah, I wasn't sure what you meant on the issue w/r/t checking props.

noisecapella:
@aliceriot The ErrorMessage tests look good but I was thinking of pulling the `ErrorMessage` from the component tree returned by `renderComponent` and asserting the props for the `ErrorMessage` for the profile, user, and dashboard pages. I'm fine with this solution though

aliceriot:
ok, do you think this is good as-is? I'm fine with switching to set of `renderComponent` based tests if you think there's value there.

aliceriot:
@noisecapella updated!

static/js/components/ErrorMessage_test.js
```diff
@@ -0,0 +1,214 @@
...
+        delete response.user_message;
+        helper.dashboardStub.returns(Promise.reject(response));
+
+        return renderComponent("/dashboard", dashboardErrorActions, false).then(([, div]) => {
+          confirmErrorMessage(div, [`AB123 ${errorString}`, '']);
+        });
+      });
+
+      it('the error from the backend does not need to exist at all as long as there is an http error', () => {
+        helper.dashboardStub.returns(Promise.reject({}));
```
noisecapella: The error response should have `errorStatusCode` at least. This is inserted in `api.js`, all other fields would come from the server

static/js/components/ErrorMessage_test.js
```diff
@@ -0,0 +1,214 @@
...
+        errorInfo: {
+          error_code: '500',
+          user_message: 'A message',
+          detail: 'Some details',
+          errorStatusCode: 404,
+        }
+      };
+    });
+
+    it('should take an errorInfo prop', () => {
```
noisecapella: Can you remove this test?

static/js/components/ErrorMessage_test.js
```diff
@@ -0,0 +1,214 @@
...
+        });
+      });
+    });
+  });
+
+  describe('showing errors on pages', () => {
+    let renderComponent, helper, patchUserProfileStub, listenForActions;
+    let dashboardErrorActions;
+
+    const confirmErrorMessage = (div, expectedMessageText) => {
```
noisecapella: Can you break `expectedMessageText` into two separate arguments instead of an array of two arguments? This should also allow you to use a default argument for the second argument

noisecapella:
Can you also move the test from `ProfilePage_test` for consistency?

static/js/components/ErrorMessage_test.js
aliceriot: ok, this test was just copypasta from what we had in `DashboardPage_test.js`. I'll add a status code.

aliceriot:
Ah yeah, somehow I missed `ProfilePage_test`.

aliceriot:
@noisecapella tests moved over

static/js/components/ErrorMessage_test.js
```diff
@@ -0,0 +1,233 @@
...
+
+        return renderComponent("/dashboard").then(([, div]) => {
+          let message = div.getElementsByClassName('alert-message')[0];
+          assert.equal(message, undefined);
+        });
+      });
+    });
+
+    describe('profile page', () => {
+      it('shows an error when profile get has errored', () => {
```
noisecapella: It looks like this test title is wrong, this is actually testing the spinner. Can you change the title? Not sure if it should be moved back or not, I'm fine either way

static/js/components/ErrorMessage_test.js
```diff
@@ -144,7 +144,7 @@ describe("ErrorInfo", () => {
     });
 
     describe('profile page', () => {
-      it('shows an error when profile get has errored', () => {
+      it('shows an error when profile get has errored instead of leaving the spinner up', () => {
```
noisecapella: It doesn't show an error at all, it only starts the request for the profile and asserts that the spinner is shown

aliceriot: oh ok sorry I didn't understand what the test was testing. I'll move it back to `ProfilePage_test` since it's super related.

aliceriot: hm so it looks like this test was actually a duplicate of the test above it in `ProfilePage_test.js`. I'm going to just delete it.

noisecapella: Oh yeah, that's a duplicate all right

noisecapella:
LGTM :+1:

PR #705, opened 2016-07-06
Added get_social_username, updated existing code to use it

### LINKED ISSUE #704: Create get_social_username function to avoid duplicating code

noisecapella:
Fixes #704 

#### What's this PR do?
Refactors code getting the username.

No changes should be noticed

### DIFF
--- a/profiles/api.py
+++ b/profiles/api.py
@@ -1,0 +1,30 @@
+"""
+Functions for profiles
+"""
+import logging
+
+from django.core.exceptions import ObjectDoesNotExist
+
+from backends.edxorg import EdxOrgOAuth2
+
+log = logging.getLogger(__name__)
+
+
+def get_social_username(user):
+    """
+    Get social auth edX username for a user, or else return None.
+
+    Args:
+        user (django.contrib.auth.models.User):
+            A Django user
+    """
+    if user.is_anonymous():
+        return None
+
+    try:
+        return user.social_auth.get(provider=EdxOrgOAuth2.name).uid
+    except ObjectDoesNotExist:
+        return None
+    except Exception as ex:  # pylint: disable=broad-except
+        log.error("Unexpected error retrieving social auth username: %s", ex)
+        return None
--- a/profiles/api_test.py
+++ b/profiles/api_test.py
@@ -1,0 +1,83 @@
+"""
+Tests for profile functions
+"""
+
+from mock import Mock
+
+from django.db.models.signals import post_save
+from django.test import TestCase
+from factory.django import mute_signals
+from testfixtures import LogCapture
+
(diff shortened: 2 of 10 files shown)


profiles/api_test.py
```diff
@@ -0,0 +1,73 @@
...
+from profiles.api import get_social_username
+from profiles.factories import ProfileFactory
+
+
+class SocialTests(TestCase):
+    """
+    Tests for profile functions
+    """
+
+    def setUp(self):
```
aliceriot: I'm guessing this function is run before each test?

profiles/api.py
```diff
@@ -0,0 +1,30 @@
...
+    if user.is_anonymous():
+        return None
+
+    try:
+        return user.social_auth.get(provider=EdxOrgOAuth2.name).uid
+    except ObjectDoesNotExist:
+        return None
+    except Exception as ex:  # pylint: disable=broad-except
+        log.error("Unexpected error retrieving social auth username: %s", ex)
+        return None
```
aliceriot: Is it worth trying to test the logging functionality?

profiles/api_test.py
noisecapella: Yes, just like `beforeEach`

profiles/api_test.py
```diff
@@ -0,0 +1,73 @@
+"""
+Tests for profile functions
+"""
+
+from mock import Mock
+
+from django.db.models.signals import post_save
+from django.test import TestCase
```
noisecapella: FYI you will see `from unittest import TestCase` in certain tests. Django's `TestCase` adds on some useful things like database setup and teardown and a test `Client` to do fake requests.

aliceriot: cool, thanks for the info!

aliceriot:
functionality looks good!

aliceriot:
I think the code generally looks good to me, and the tests all seem to be passing.

I'm wondering though, what is the reason for implementing this as a separate function, as opposed to a method on the user class? My guess is increased unit-test-ability and generally preferring composition over building big classes with lots of methods, but I was just curious.

profiles/api.py
noisecapella: We generally don't test logging, but it seems like we should here since we have no information about the problem otherwise. I'll add this to the test

noisecapella:
The `User` class comes from Django and it's nontrivial to add methods to it or to replace it with our own (though it can be done).

aliceriot:
ah, gotcha. this seems like the correct approach then.

noisecapella:
Error logging should be checked now

aliceriot:
LGTM :+1:

PR #706, opened 2016-07-06
Added Elasticsearch index

### LINKED ISSUE #670: search: re-index when user profile is created / updated

### LINKED ISSUE #699: Implement ElasticSearch index
we'll use ElasticSearch DSL

Configure the index schema to match the user profile, user education, user employment

We will be adding new fields to the models in other issues. #677 #672

noisecapella:
Fixes #699 
Fixes #670 

#### What's this PR do?
Sets up Elasticsearch indexing to happen:
 - when the `recreate_index` management command is run.
 - when a `Profile`, `Education`, or `Employment` are updated or created. In this case a celery task is executed

search/api.py

There are changes to `docker-compose.yml` so you'll need to restart your containers, no rebuilding should be necessary though.

First, run `docker-compose run web ./manage.py recreate_index`. Then run this command: `docker-compose run web curl http://elastic:9200/micromasters/_search | python -m json.tool`. Verify that you see profile data in the search results.

Go into the users page and make an update to the profile. Run the curl command again and verify that you see the updated value in the profile.

#### Background context
From now on, people creating new instances of this app will need to create an Elasticsearch index and mapping via management command: `./manage.py recreate_index`. This command may also be run at any time to delete the existing index and reindex all relevant data (this may take a while).

### DIFF
--- a/courses/views_tests.py
+++ b/courses/views_tests.py
@@ -1,12 +1,12 @@
 """Tests for the API"""
 # pylint: disable=no-self-use
 from django.core.urlresolvers import reverse
-from django.test import TestCase
 
 from .factories import ProgramFactory, CourseFactory
+from search.base import ESTestCase
 
 
-class ProgramTests(TestCase):
+class ProgramTests(ESTestCase):
     """Tests for the Program API"""
     def test_lists_live_programs(self):
         """Live programs should show up"""
@@ -26,7 +26,7 @@ class ProgramTests(TestCase):
         assert len(resp.json) == 0
 
 
-class CourseTests(TestCase):
+class CourseTests(ESTestCase):
     """Tests for the Course API"""
     def test_list_course_if_program_live(self):
         """
--- a/search/api.py
+++ b/search/api.py
@@ -1,0 +1,259 @@
+"""
+Functions for search
+"""
+from itertools import islice
+import logging
+
+from django.conf import settings
+from django.contrib.auth.models import User
+from elasticsearch.helpers import bulk
+from elasticsearch.exceptions import NotFoundError
+from elasticsearch_dsl import Mapping
+from elasticsearch_dsl.connections import connections
+
+from profiles.models import Profile
+from profiles.serializers import ProfileSerializer
+from search.exceptions import ReindexException
+
+log = logging.getLogger(__name__)
+
+USER_DOC_TYPE = 'user'
+DOC_TYPES = (USER_DOC_TYPE, )
+_CONN = None
+# When we create the connection, check to make sure all appropriate mappings exist
+_CONN_VERIFIED = False
+# A string which must be indexed exactly as is and not stemmed for full text search (the default)
+NOT_ANALYZED_STRING_TYPE = {
+    'type': 'string',
+    'index': 'not_analyzed',
+}
+BOOL_TYPE = {'type': 'boolean'}
+DATE_TYPE = {'type': 'date', 'format': 'date'}
+LONG_TYPE = {'type': 'long'}
+
+
+def get_conn(verify=True):
+    """
+    Lazily create the connection.
+    """
+    # pylint: disable=global-statement
+    global _CONN
+    global _CONN_VERIFIED
+
+    do_verify = False
+    if _CONN is None:
+        _CONN = connections.create_connection(hosts=[settings.ELASTICSEARCH_URL])
+        # Verify connection on first connect if verify=True.
+        do_verify = verify
+
+    if verify and not _CONN_VERIFIED:
+        # If we have a connection but haven't verified before, do it now.
+        do_verify = True
+
+    if not do_verify:
+        if not verify:
+            # We only skip verification if we're reindexing or
+            # deleting the index. Make sure we verify next time we connect.
+            _CONN_VERIFIED = False
+        return _CONN
+
+    # Make sure everything exists.
+    index_name = settings.ELASTICSEARCH_INDEX
+    if not _CONN.indices.exists(index_name):
+        raise ReindexException("Unable to find index {index_name}".format(
+            index_name=index_name
+        ))
+
+    mappings = _CONN.indices.get_mapping()[index_name]["mappings"]
+    for doc_type in DOC_TYPES:
+        if doc_type not in mappings.keys():
+            raise ReindexException("Mapping {doc_type} not found".format(
+                doc_type=doc_type
+            ))
+
+    _CONN_VERIFIED = True
+    return _CONN
+
+
+def _index_users_chunk(users):
+    """
+    Add/update a small number of user records in Elasticsearch
+
+    Args:
+        users (list of User):
+            List of users
+
+    Returns:
+        int: Number of items inserted into Elasticsearch
+    """
+
+    conn = get_conn()
+    insert_count, errors = bulk(
+        conn,
+        (serialize_user(user) for user in users),
+        index=settings.ELASTICSEARCH_INDEX,
+        doc_type=USER_DOC_TYPE,
+    )
+
+    if len(errors) > 0:
+        raise ReindexException("Error during bulk insert: {errors}".format(
+            errors=errors
+        ))
+    refresh_index()
+
+    return insert_count
+
+
+def index_users(users, chunk_size=100):
+    """
+    Add/update profile records in Elasticsearch.
+
+    Args:
+        users (iterable of User):
+            Iterable of users
+        chunk_size (int):
+            How many users to index at once.
+
+    Returns:
+        int: Number of indexed items
+    """
+    # Use an iterator so we can keep track of what's been indexed already
+    users = iter(users)
+
+    count = 0
+    chunk = list(islice(users, chunk_size))
+    while len(chunk) > 0:
+        count += _index_users_chunk(chunk)
+        chunk = list(islice(users, chunk_size))
+
+    refresh_index()
+
+    return count
+
+
+def remove_user(user):
+    """
+    Remove a user from Elasticsearch.
+    """
+    conn = get_conn()
+    try:
+        conn.delete(index=settings.ELASTICSEARCH_INDEX, doc_type=USER_DOC_TYPE, id=user.id)
+    except NotFoundError:
+        # Item is already gone
+        pass
+
+
+def serialize_user(user):
+    """
+    Serializes user for use with Elasticsearch.
+
+    Args:
+        user (User): A user to serialize
+    Returns:
+        dict: The data to be sent to Elasticsearch
+    """
+    serialized = {
+        'id': user.id,
+        '_id': user.id,
+    }
+    try:
+        serialized['profile'] = ProfileSerializer().to_representation(user.profile)
+    except Profile.DoesNotExist:
+        # Just in case
+        pass
+    return serialized
+
+
+def refresh_index():
+    """
+    Refresh the Elasticsearch index
+    """
+    get_conn().indices.refresh(index=settings.ELASTICSEARCH_INDEX)
+
+
+def create_user_mapping():
+    """
+    Create a mapping for profiles. If one already exists, delete it first.
+    """
+    conn = get_conn(verify=False)
+
+    index_name = settings.ELASTICSEARCH_INDEX
+    if conn.indices.exists_type(index=index_name, doc_type=USER_DOC_TYPE):
+        conn.indices.delete_mapping(index=index_name, doc_type=USER_DOC_TYPE)
+
+    mapping = Mapping(USER_DOC_TYPE)
+    mapping.field("id", "long")
+    mapping.field("profile", "nested", properties={
+        'account_privacy': NOT_ANALYZED_STRING_TYPE,
+        'agreed_to_terms_of_service': BOOL_TYPE,
+        'birth_city': NOT_ANALYZED_STRING_TYPE,
+        'birth_country': NOT_ANALYZED_STRING_TYPE,
+        'birth_state_or_territory': NOT_ANALYZED_STRING_TYPE,
+        'city': NOT_ANALYZED_STRING_TYPE,
+        'country': NOT_ANALYZED_STRING_TYPE,
+        'date_of_birth': DATE_TYPE,
+        'education': {'type': 'nested', 'properties': {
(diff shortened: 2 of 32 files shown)


giocalitri:
@gsidebo Feel free to review, but I would like to make a pass at the very end to see if I spot something odd

aliceriot:
I ran into an issue with CORS headers when trying to hit the ES index directly from the JS client:

cors

We probably will solve this issue when we build out the search API.

gsidebo:
> From now on, people creating new instances of this app will need to create an Elasticsearch index and mapping via management command:

does this need to be added to the readme?

search/base.py
```diff
@@ -0,0 +1,18 @@
+"""
```
gsidebo: i would not expect to find our base test case class for ElasticSearch-related functionality, or any test case class for that matter, in `search/base.py`. i don't know why it's in the search/ directory, and base.py is misleading. out of the directories we have, i would think a new file (`test_classes.py`, et al) in `micromasters/tests` would make more sense, but maybe we want a new directory

search/base.py
```diff
@@ -0,0 +1,18 @@
...
+Base test classes for search
+"""
+
+from django.test import TestCase
+
```
gsidebo: any reason for the empty line?

search/apps.py
```diff
@@ -0,0 +1,17 @@
+"""
+AppConfig
+"""
+from django.apps import AppConfig
```
gsidebo: let's be consistent about whether or not we're inserting an empty line before the imports

search/api_test.py
```diff
@@ -0,0 +1,285 @@
...
+
+class SerializerTests(ESTestCase):
+    """
+    Tests for serializer
+    """
+
+    def test_profile_serializer(self):  # pylint: disable=no-self-use
+        """
+        Tests the profile serializer
+        """
```
gsidebo: are docstrings a hard requirement for the linter? if so, this is an example of when that kinda hurts us. this docstring is really just taking up space and doesn't add any information. not that it _needs_ more information - `test_profile_serializer` describes itself just fine

search/api_test.py
```diff
@@ -0,0 +1,285 @@
...
+        # Confirm nothing in index
+        assert_search(search(), [])
+        index_profiles(Profile.objects.iterator(), chunk_size=4)
+        assert_search(search(), list(Profile.objects.all()))
+
+
+class SerializerTests(ESTestCase):
+    """
+    Tests for serializer
+    """
```
gsidebo: ditto for the comment below?

search/api_test.py
```diff
@@ -0,0 +1,285 @@
...
+    def test_profile_update(self):
+        """
+        Test that Profile is reindexed after being updated
+        """
+        user = UserFactory.create()
+        assert search()['total'] == 1
+        profile = user.profile
+        profile.first_name = 'updated'
+        profile.save()
+
```
gsidebo: can we get rid of these empty lines before the assertion?

search/api.py
```diff
@@ -0,0 +1,207 @@
...
+log = logging.getLogger(__name__)
+
+
+PROFILE_DOC_TYPE = 'profile'
+
+DOC_TYPES = (PROFILE_DOC_TYPE, )
+
+_CONN = None
+# When we create the connection, check to make sure all appropriate mappings exist
+_CONN_VERIFIED = False
```
gsidebo: can we get rid of the blank lines from line 20 to here?

search/base.py
```diff
@@ -0,0 +1,18 @@
...
+class ESTestCase(TestCase):
+    """
+    Set up ES index on setup
+    """
+
+    def setUp(self):
+        super(ESTestCase, self).setUp()
+
+        clear_index()
```
gsidebo: empty line

search/api.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+    Wipe and recreate index and mapping. No indexing is done.
+    """
+    conn = get_conn(verify=False)
+    index_name = settings.ELASTICSEARCH_INDEX
+    if conn.indices.exists(index_name):
+        conn.indices.delete(index_name)
+    conn.indices.create(index_name)
+    conn.indices.refresh()
+
```
gsidebo: sorry to harp on the empty lines across these files, but this one and line 206 don't seem to add any value

search/api.py
gsidebo: i'm ok with the usage of `global` in the function(s) below, but i think it would help to encapsulate `_CONN` and `_CONN_VERIFIED` in a simple class, eg: `ESConnection`?  lines 24 & 26 would become something like `es_connection = ESConnection()`. some other possible changes as a result:

- line 38: `if _CONN is None` --> `if not es_connection.created():`
- line 39: `_CONN = connections.create_connection...` --> `es_connection.set(connections.create_connection...)`
- line 43: `if verify and not _CONN_VERIFIED:` --> `if verify and not es_connection.verified():`
- create helper methods for `_CONN.indices.exists...`, `_CONN.indices.get_mapping()...` etc

since `get_conn` is used in a bunch of places, maybe it should still return that object (ie: `return es_connection.conn`)

gsidebo:
functionality looks good

noisecapella:
I was just thinking that this requires #707 before it gets merged

search/base.py
noisecapella: The `_test.py` files are supposed to be located next to the files they test so that they show up next to each other in code review. I'm not sure it makes sense to move `search/base.py` into a `tests` directory by itself

search/api_test.py
noisecapella: I still think it's a good idea for them to be mandatory, though this function is small enough to not need it. I'll reword this docstring to describe the test better

search/api.py
noisecapella: I'm not sure this gives us much. We'll still need some global to store the `ESConnection`, and the helper function `verify` would only be called in one place. `get_conn` already encapsulates the verification so that it happens when a connection is requested

noisecapella:
Per my conversation with @giocalitri I'm also going to change the indexing to be on the user rather than profile

search/base.py
gsidebo: there are many suggestions for structuring a python project and placement of unit tests, and [a](http://docs.python-guide.org/en/latest/writing/structure/#sample-repository) [few](http://learnpythonthehardway.org/book/ex46.html) of them contradict the structure of this codebase. i don't really agree with using code review convenience as a guide to structuring a project, and i can't say that i've ever found it really beneficial to have the two files next to each other, though others' results may vary. on top of that, if each project folder had a tests/ directory (where necessary), the file with the actual logic and the test file would be reasonably proximal anyway.

given the structure of the project elsewhere, i'm ok with leaving this as-is, but i think this is a pretty good illustration of the downfalls of mixing our test modules in with 'real' modules. we're putting a test class into a module whose name does not suggest that it has anything to do with tests. we might have other functionality (perhaps the ESConnection class i suggested elsewhere in the PR) that makes sense in `search/base.py`. then we have 'real' code mixed in with test 'code'. not the end of the world, but i think we should be avoiding that

noisecapella:
Ready for review again. I also added an explicit ES mapping which I forgot in the earlier revision

@giocalitri Can you take a look when you have time?

search/api.py
```diff
@@ -0,0 +1,261 @@
...
+from search.exceptions import ReindexException
+
+log = logging.getLogger(__name__)
+
+
+USER_DOC_TYPE = 'user'
+DOC_TYPES = (USER_DOC_TYPE, )
+_CONN = None
+# When we create the connection, check to make sure all appropriate mappings exist
+_CONN_VERIFIED = False
```
gsidebo: any opinion on the change to `_CONN` and `_CONN_VERIFIED` that i mentioned earlier?

noisecapella: I don't think it's worth encapsulating into `ESConnection` since we still need one global variable to store the `ESConnection`. Both variables are only used by `get_conn`

search/api.py
```diff
@@ -0,0 +1,261 @@
...
+        }},
+    })
+    mapping.save(index_name)
+
+
+def create_mappings():
+    """
+    Create all mappings, deleting existing mappings if they exist.
+    """
+    create_user_mapping()
```
gsidebo: do we have a reason for both `create_mappings` and `create_user_mapping` to exist?

noisecapella: It's there in case we want to add other mappings in the future

search/api.py
```diff
@@ -0,0 +1,261 @@
...
+        conn.indices.delete_mapping(index=index_name, doc_type=USER_DOC_TYPE)
+
+    # A string which must be indexed exactly as is and not stemmed for full text search (the default)
+    not_analyzed_string_type = {
+        'type': 'string',
+        'index': 'not_analyzed',
+    }
+    bool_type = {'type': 'boolean'}
+    date_type = {'type': 'date', 'format': 'date'}
+    long_type = {'type': 'long'}
```
gsidebo: if we're setting `USER_DOC_TYPE` as a module-level variable, i think it makes sense to do the same for the above 4 values

search/apps.py
gsidebo: fair enough, im not too insistent on that one

search/api.py
```diff
@@ -0,0 +1,261 @@
...
+from elasticsearch.helpers import bulk
+from elasticsearch.exceptions import NotFoundError
+from elasticsearch_dsl import Mapping
+from elasticsearch_dsl.connections import connections
+
+from profiles.models import Profile
+from profiles.serializers import ProfileSerializer
+from search.exceptions import ReindexException
+
+log = logging.getLogger(__name__)
```
gsidebo: since this isn't an import, shouldn't this line have 2 blank lines preceding it and, if preferred, just 1 blank line succeeding it?

gsidebo:
looks good aside from the two small comments. testing functionality again

search/api.py
```diff
@@ -0,0 +1,261 @@
+"""
+Functions for search
+"""
+from itertools import islice
+import logging
```
giocalitri: `import ` imports come before `from` imports

search/api.py
giocalitri: nope: Class and function definition have to

search/api.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+USER_DOC_TYPE = 'user'
+DOC_TYPES = (USER_DOC_TYPE, )
+_CONN = None
+# When we create the connection, check to make sure all appropriate mappings exist
+_CONN_VERIFIED = False
+
+
+def get_conn(verify=True):
```
giocalitri: this does come directly from LORE, right?

search/api.py
```diff
@@ -0,0 +1,261 @@
...
+            List of users
+
+    Returns:
+        int: Number of items inserted into Elasticsearch
+    """
+
+    conn = get_conn()
+    insert_count, errors = bulk(
+        conn,
+        (serialize_user(user) for user in users),
```
giocalitri: nit: why did you put parentheses here?

search/api.py
gsidebo: then i would suggest we get rid of the double line break under it

search/api.py
noisecapella: In general yes. Here's the LORE code if you want to compare: https://github.com/mitodl/lore/blob/master/search/utils.py#L49

search/api.py
noisecapella: In generators parentheses are required: https://docs.python.org/3/reference/expressions.html#generator-expressions

search/api.py
noisecapella: Is this our convention or a PEP 8 requirement? I'm happy to make the change, just haven't heard of this before

search/api.py
gsidebo: `get_conn` could be made shorter and more readable if we created a class that took care of some of these concerns, but i don't feel too insistent on this one. carry on

search/api.py
giocalitri: More like a general convention 
http://stackoverflow.com/a/20763446/257092
But feel free to leave it as is

courses/views_tests.py
```diff
@@ -1,12 +1,12 @@
...
 # pylint: disable=no-self-use
 from django.core.urlresolvers import reverse
-from django.test import TestCase
 
 from .factories import ProgramFactory, CourseFactory
+from search.base import ESTestCase
 
 
-class ProgramTests(TestCase):
+class ProgramTests(ESTestCase):
```
giocalitri: why this change?

giocalitri: wait, I have still to read the new mixin

giocalitri: ok... I remember now

noisecapella: We need to run `recreate_index()` before each test to allow signals sent when the models change to connect to Elasticsearch correctly

search/api_test.py
```diff
@@ -0,0 +1,284 @@
...
+def assert_search(results, users):
+    """
+    Assert that search results match users
+    """
+    assert results['total'] == len(users)
+    sources = sorted([hit['_source'] for hit in results['hits']], key=lambda hit: hit['id'])
+    sorted_users = sorted(users, key=lambda user: user.id)
+    serialized = [remove_key(serialize_user(user), "_id") for user in sorted_users]
+    assert serialized == sources
```
giocalitri: I would put these two functions in a mixin, but feel free to leave as is

search/base.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ESTestCase(TestCase):
+    """
+    Set up ES index on setup
+    """
+
+    def setUp(self):
+        super(ESTestCase, self).setUp()
+        clear_index()
```
giocalitri: we might want to add an extra function to disable the indexing while we perform some actions in a test: I would prefer to avoid the indexing at all in some cases where I do not need it.
But feel free to ignore for now (and maybe add another issue)

giocalitri:
I think this code is fine

search/base.py
noisecapella: Filed #721 for this

noisecapella:
Comments addressed, any other comments @gsidebo?

gsidebo:
looks great 👍

PR #709, opened 2016-07-07
Added custom roles definition.

### LINKED ISSUE #674: Add roles to user model
3 roles currently:

admin
MM team (can have this role on multiple programs) 
learner (can have this role on multiple programs)

Do we need an instructor role, distinct from MM team?

MM team permissions includes things like:
- editing their pages in the CMS
- search private learner profiles in their MicroMasters
- possible future features like creating new forums and messaging learners in their MicroMasters

giocalitri:
closes #674 

#### What's this PR do?
Implements a basic infrastructure for roles, permissions and object based permissions.
It does not enforce any of them

`micromasters/models.py` then
`micromasters/roles.py` then
`micromasters/signals.py`

The easiest way is to assign an user a role in the new model and then check that the user has a group with the same name of the role assigned to her.

Until I work on the `django-role-permissions` library, we will support only one role per user.


micromasters/models.py
```diff
@@ -0,0 +1,37 @@
...
+General models for the micromasters app
+"""
+
+from django.contrib.auth.models import User
+from django.db import models
+
+from courses.models import Program
+
+
+class MicroMastersRole(models.Model):
```
noisecapella: Why not just `Role` here?

micromasters/admin.py
```diff
@@ -0,0 +1,14 @@
+"""
+Admin views for Courses & Programs
```
noisecapella: Can you update the docstring?

noisecapella:
Why add these models to the `micromasters` app instead of creating a new `roles` app for example?

micromasters/permissions.py
```diff
@@ -0,0 +1,30 @@
...
+from micromasters.models import MicroMastersRole
+
+
+@register_object_checker()
+def can_advance_search(role, user, program):
+    """
+    Specific permission checker on a program
+    """
+    return (
+        has_permission(user, 'can_advance_search') and MicroMastersRole.objects.filter(
```
noisecapella: Just to clarify, the permissions are Django user permissions? And the permissions are not attached to a program?

noisecapella: Nevermind, reading the docs now

noisecapella: Can we use the constants defined in `MicroMastersPermissions` below?

noisecapella:
Heading to lunch, will resume after

micromasters/signals.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Signal handler that happens before a role assignment is done.
+    If the the save happens for a modification, the previous role must be removed
+    if not correspondent to other programs.
+
+    Theoretically this is not necessary with the current implementation of the
+    django-role-permission library.
+    """
+    try:
+        old_instance = MicroMastersRole.objects.get(pk=instance.pk)
```
noisecapella: Should these all be wrapped in a transaction?

micromasters/signals.py
```diff
@@ -0,0 +1,80 @@
...
+    # Note this check is made BEFORE the save
+    if MicroMastersRole.objects.filter(role=old_instance.role).count() > 1:
+        return
+
+    log.debug(
+        'removing role % for user %s',
+        instance.role,
+        instance.user.username,
+    )
+    remove_role(instance.user)
```
noisecapella: Does this function change `MicroMastersRole`?

micromasters/signals.py
```diff
@@ -0,0 +1,80 @@
...
+@receiver(post_delete, sender=MicroMastersRole, dispatch_uid="delete_remove_role_from_user")
+def delete_remove_role_from_user(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Signal handler that happens before a role removal is done.
+    The role must be removed only if not correspondent to other programs.
+    """
+    # if the user has the same role assigned for other programs, then the role must be kept
+    # (it is NOT specific to a program at this level)
+    # Note this check is made AFTER the delete
+    if MicroMastersRole.objects.filter(role=instance.role).count() > 0:
```
noisecapella: I feel like we should have a transaction here too

micromasters/models.py
giocalitri: I wanted a name slightly different from the logical roles... I can change it

micromasters/admin.py
giocalitri: sure

giocalitri:
> Why add these models to the micromasters app instead of creating a new roles app for example?

Because the definition of the logical Roles and Permissions must be in the micromasters app (the library works in this way if I understood well), so I did not want to put half of the stuff in another app.

micromasters/permissions.py
giocalitri: need to check for circular imports, but in theory yes

micromasters/signals.py
giocalitri: the problem here is that this stuff run in different part of the code, meaning the insert happens in the model and this stuff happens in the signal handlers: not sure how to make a big transaction here...

giocalitri: but I will look into it

micromasters/signals.py
giocalitri: no: this is an helper function that basically removes the logical role assignment from the user

noisecapella:
I think the instructions are awkwardly written. The quick start for the roles library says this just as a suggestion, but the other parts of the docs don't require this

micromasters/signals.py
giocalitri: same answer as before, I guess.

giocalitri:
I will look into it then and in case create another app

roles/roles.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class Permissions:
+    """
+    All MicroMasters custom permissions
+    """
+    CAN_ADVANCE_SEARCH = 'can_advance_search'
+    CAN_EDIT_CMS = 'can_edit_cms'
+    CAN_MESSAGE_LEARNERS = 'can_message_learners'
+    CAN_CREATE_FORUMS = 'can_create_forums'
```
gsidebo: why isn't this class defined in `roles/permissions.py`? that's where i would expect to find it, at least

roles/signals.py
```diff
@@ -0,0 +1,80 @@
...
+        instance.role,
+        instance.user.username,
+    )
+    assign_role(instance.user, instance.role)
+
+
+@receiver(post_delete, sender=Role, dispatch_uid="delete_remove_role_from_user")
+def delete_remove_role_from_user(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Signal handler that happens before a role removal is done.
```
gsidebo: this handler executes _after_ a role removal is done, right?

roles/roles.py
giocalitri: Because `roles/permissions.py` is something else: it is the definition of the per-object permissions function checkers

roles/roles_test.py
```diff
@@ -0,0 +1,29 @@
...
+    def test_staff_role_id(self):
+        """
+        Test for the Staff role ID
+        """
+        assert hasattr(roles.Staff, 'ROLE_ID')
+        assert roles.Staff.ROLE_ID == Role.STAFF
+
+    def test_instructor_role_id(self):
+        """
+        Test for the Staff role ID
```
gsidebo: these two test cases can be combined into one pretty easily, but i'll leave that up to you. if you decide not to do that, you should update this docstring 😄

roles/signals.py
```diff
@@ -0,0 +1,80 @@
...
+    The role must be removed only if not correspondent to other programs.
+    """
+    # if the user has the same role assigned for other programs, then the role must be kept
+    # (it is NOT specific to a program at this level)
+    # Note this check is made AFTER the delete
```
gsidebo: these first line of these comments seems to echo exactly what the docstring is saying

roles/signals_test.py
```diff
@@ -0,0 +1,85 @@
...
+        signal that removes the previous role from the user.
+        """
+        mm_role = Role.objects.create(
+            program=self.program,
+            user=self.user,
+            role='staff',
+        )
+        assert has_role(self.user, 'staff') is True
+        assert has_permission(self.user, 'can_advance_search') is True
+        assert has_object_permission('can_advance_search', self.user, self.program) is True
```
gsidebo: these 3 assertions are doing the same thing as the last 3 assertions of the previous test case. we can probably get rid of these

roles/signals_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+        with mute_signals(post_save):
+            mm_role.role = 'instructor'
```
gsidebo: nitpick: the _actual_ Instructor role has the 'can_advance_search' permission, so this is a bit confusing. can you change this to something like 'learner' or 'test_role'?

roles/signals_test.py
```diff
@@ -0,0 +1,85 @@
...
+            role='staff',
+        )
+        assert has_role(self.user, 'staff') is True
+        assert has_permission(self.user, 'can_advance_search') is True
+        assert has_object_permission('can_advance_search', self.user, self.program) is True
+
+        mm_role.delete()
+        assert has_role(self.user, 'staff') is False
+        assert has_permission(self.user, 'can_advance_search') is False
+        assert has_object_permission('can_advance_search', self.user, self.program) is False
```
gsidebo: i think it's probably worth it to include a test case that removes a user's role for one program, but they still have that role for a different program

roles/signals.py
```diff
@@ -0,0 +1,80 @@
...
+    Theoretically this is not necessary with the current implementation of the
+    django-role-permission library.
+    """
+    try:
+        old_instance = Role.objects.get(pk=instance.pk)
+    except Role.DoesNotExist:
+        return
+    # if the user has the same role assigned for other programs, then the role must be kept
+    # (it is NOT specific to a program at this level)
+    # Note this check is made BEFORE the save
```
gsidebo: same comment as below re: comments mimicking docstring

gsidebo:
when you say 'logical role' in these comments, are you just talking about roles that we've defined (as opposed to roles that Django recognizes)?

roles/roles.py
gsidebo: the django-role-permissions docs specify that object permission checkers should go in `permissions.py`, but from what i've read, they don't say (a) that nothing else should go in there, or (b) that an internal vocabulary of permissions (which we have here) should go into `roles.py`. it's up to us, and to me it makes more sense for that vocabulary of permissions to be defined in `permissions.py`

roles/permissions.py
```diff
@@ -0,0 +1,31 @@
...
+    return (
+        has_permission(user, Permissions.CAN_ADVANCE_SEARCH) and Role.objects.filter(
+            user=user, role=role.ROLE_ID, program=program).exists()
+    )
+
+
+@register_object_checker()
+def can_message_learners(role, user, program):
+    """
+    Specific permission checker on a program
```
gsidebo: can you make these docstrings more specific? eg: "Determines whether a user can perform an advanced search" and "Determines whether a user can send a message to learners".

gsidebo:
i was just thinking about the use of `STAFF` and `INSTRUCTOR` in models.py and the setting of `ROLE_ID` in the actual role classes. it seemed a little clunky to me, so i tried this: https://gist.github.com/gsidebo/5df9670b9694d26da44a0b9555ad4552

cuts out some boilerplate code, sets the Role model choices 'dynamically', defines the `role_id` as part of the Role class rather than the model, which made it necessary for you to write that comment about proper naming. setting the default choice is a little opaque (just uses the first `assignable_roles` item) but i think this is an improvement. let me know what you think

roles/permissions.py
giocalitri: sure

roles/roles.py
giocalitri: I think this logically belongs here, so I think I will leave them here.

roles/roles_test.py
giocalitri: sure

roles/signals.py
giocalitri: fixing

roles/signals.py
giocalitri: yep, will fix

roles/signals.py
giocalitri: fixing

roles/signals_test.py
giocalitri: not sure I understand what you are trying to say here

roles/signals_test.py
giocalitri: this is a good idea

roles/signals_test.py
giocalitri: good idea

roles/signals_test.py
gsidebo: the actual Instructor role ([here](https://github.com/mitodl/micromasters/pull/709/files#diff-1ce2c7710978a1b0265fa651f240dde0R37)) is set to have the 'can_advance_search' = True by default. below, you are asserting that 'can_advance_search' is False. i don't think there is actually a Role associated with 'instructor' in the context of this test case, so it seems like you can set any value to `mm_role.role`. it's just confusing to set it to 'instructor' and expect 'can_advance_search' to be false when in the _actual_ instructor role, that should be true

giocalitri: This is testing the `pre_save` disabling the `post_save`. It is pretty clear what is doing IMHO.
The other problem is that I cannot use a different fake role, because the model requires `staff` or `instructor`

giocalitri:
>i was just thinking about the use of `STAFF` and `INSTRUCTOR` in models.py and the setting of `ROLE_ID` in the actual Role classes. it seemed a little clunky to me, so i tried this: https://gist.github.com/gsidebo/5df9670b9694d26da44a0b9555ad4552

>cuts out some boilerplate code, sets the Role model choices 'dynamically', defines the `role_id` as part of the Role class rather than the model, which made it necessary for you to write that comment about proper naming. setting the default choice is a little opaque (just uses the first `assignable_roles` item) but i think this is an improvement. let me know what you think

I do not think this is a preferable pattern: the actual django documentation [suggests the same approach I used](https://docs.djangoproject.com/en/1.9/ref/models/fields/#choices).

Plus the way you implement does not enforce a direct connection between the model and the roles definition in `roles.py` (this is important to me, but not sure it is to everybody).

roles/signals_test.py
```diff
@@ -0,0 +1,106 @@
...
+            user=self.user,
+            role='staff',
+        )
+        Role.objects.create(
+            program=program,
+            user=self.user,
+            role='staff',
+        )
+        self.assert_standard_role_permissions(True)
+        self.assert_standard_role_permissions(True, program)
```
gsidebo: can you change the above to something like this?

```python
        existing_program = self.program
        new_program = ProgramFactory.create()
        mm_role_1 = Role.objects.create(
            program=existing_program,
            user=self.user,
            role='staff',
        )
        Role.objects.create(
            program=new_program,
            user=self.user,
            role='staff',
        )
        self.assert_standard_role_permissions(True, existing_program)
        self.assert_standard_role_permissions(True, new_program)
```

would help a lot with clarity and showing the intent of the test. reading all these `program`s and `self.program`s is a bit confusing

roles/models.py
```diff
@@ -0,0 +1,37 @@
...
+        (STAFF, STAFF),
+        (INSTRUCTOR, INSTRUCTOR),
+    ]
+    user = models.ForeignKey(User)
+    program = models.ForeignKey(Program)
+    role = models.CharField(
+        choices=ALL_ROLES_CHOICES,
+        default=STAFF,
+        max_length=30,
+    )
```
gsidebo: (continuing the conversation from the main thread where I linked to the gist...)

that example from django documentation is a scenario where the choices are simple strings, and there aren't actual classes associated with the choices. we have specific classes associated with these choices, and the choices should correspond to them directly.

how about this instead:

```python
# roles/roles.py
class Staff(AbstractUserRole):
    # ...
    role_id = 'staff'

class Instructor(AbstractUserRole):
    # ...
    role_id = 'instructor'

# roles/models.py
class Role(models.Model):
    # ...
    ASSIGNABLE_ROLES = [Staff, Instructor]
    DEFAULT_ROLE = Staff
    # ...
    role = models.CharField(
        choices=[(role.role_id, role.role_id) for role in ASSIGNABLE_ROLES],
        default=DEFAULT_ROLE.role_id,
        max_length=30,
    )

```

similar to what i posted with a couple key differences. there is a formal relationship between the model and the actual Role classes - `ASSIGNABLE_ROLES` is defined directly on the model, it references Roles directly, and builds the choices from that list. i just think it's very awkward to have the role choice name ('staff', etc) defined outside of the actual Role class when it's basically Role class metadata. any new Role we create should have an associated choice available for the model, and i don't see a good reason why we should need to define data for that Role in two different places. added benefit: the above example doesn't require a comment in the docstring to guide the creation of these choices

roles/signals_test.py
giocalitri: sure

roles/models.py
giocalitri: OK, I clearly misunderstood the gist.
This makes sense.

giocalitri:
@gsidebo I think I am done with the comments.

gsidebo:
looks great! 👍

PR #710, opened 2016-07-11
Updated validation state when editing fields

### LINKED ISSUE #549: the error message should disappear if the user fixes the issue
I noticed this in the validation of the different profile tabs, but we should fix this everywhere.

screen shot 2016-06-15 at 11 40 00 am

aliceriot:

this closes #549 

#### What's this PR do?

This changes the validation setup so that when the user fixes a validation error (fills out a required field, enters a valid date, etc) the error will be removed automatically. Really, we're just re-validating the profile.

Review the changes to `actions/ui.js` and the changes to `setWorkHistoryEdit` on `EmploymentForm.js`, and the changes to `setEducationDegreeInclusions` on `EducationForm.js`.

There are also some changes comprising adding types (esp `props` declarations) to files or classes which didn't previously have them.

Visit `/users` and `/profile`. For every input on those pages (and in all stages of filling out the profile) the following should hold:

- make an invalid input (empty text field, invalid selection for 'preferred language', etc)
- click save
- validation errors should appear
- correct the invalid input
- validation errors should disappear, without the user having to click on 'save' again

additionally, correcting validation errors involving the switches on the education and work history profile steps should immediately remove the messages as well.

### DIFF
--- a/static/js/actions/ui.js
+++ b/static/js/actions/ui.js
@@ -1,6 +1,6 @@
 // @flow
 // general UI actions
-import type { Action } from '../flow/generalTypes';
+import type { Action, Dispatcher } from '../flow/generalTypes';
 
 export const CLEAR_UI = 'CLEAR_UI';
 export const clearUI = () => ({ type: CLEAR_UI });
@@ -22,9 +22,12 @@ export const setDialogVisibility = (bool: boolean): Action => (
 
 // work history actions
 export const SET_WORK_HISTORY_EDIT = 'SET_WORK_HISTORY_EDIT';
-export const setWorkHistoryEdit = (bool: boolean): Action => (
-  { type: SET_WORK_HISTORY_EDIT, payload: bool }
-);
+export const setWorkHistoryEdit = (bool: boolean): Dispatcher => {
+  return dispatch => {
+    dispatch({ type: SET_WORK_HISTORY_EDIT, payload: bool});
+    return Promise.resolve();
+  };
+};
 
 export const SET_WORK_DIALOG_VISIBILITY = 'SET_WORK_DIALOG_VISIBILITY';
 export const setWorkDialogVisibility = (bool: boolean): Action => (
@@ -60,9 +63,12 @@ export const setEducationDegreeLevel = (level: string): Action => (
 );
 
 export const SET_EDUCATION_DEGREE_INCLUSIONS = 'SET_EDUCATION_DEGREE_INCLUSIONS';
-export const setEducationDegreeInclusions = (degreeInclusions: Object): Action => (
-  { type: SET_EDUCATION_DEGREE_INCLUSIONS, payload: degreeInclusions }
-);
+export const setEducationDegreeInclusions = (degreeInclusions: Object): Dispatcher => {
+  return dispatch => {
+    dispatch({ type: SET_EDUCATION_DEGREE_INCLUSIONS, payload: degreeInclusions });
+    return Promise.resolve();
+  };
+};
 
 export const SET_USER_PAGE_DIALOG_VISIBILITY = 'SET_USER_PAGE_DIALOG_VISIBILITY';
 export const setUserPageDialogVisibility = (bool: boolean): Action => (
--- a/static/js/components/EducationDialog.js
+++ b/static/js/components/EducationDialog.js
@@ -12,6 +12,9 @@ import SelectField from './inputs/SelectField';
 import CountrySelectField from './inputs/CountrySelectField';
 import StateSelectField from './inputs/StateSelectField';
 import FieldsOfStudySelectField from './inputs/FieldsOfStudySelectField';
+import type { UIState } from '../reducers/ui';
+import type { Profile, SaveProfileFunc } from '../flow/profileTypes';
+import type { Validator, UIValidator } from '../util/validation';
 
 export default class EducationDialog extends ProfileFormFields {
   constructor(props: Object) {
@@ -23,11 +26,16 @@ export default class EducationDialog extends ProfileFormFields {
   }
   educationLevelLabels: Object;
 
-  static propTypes = {
-    open:           React.PropTypes.bool,
-    onClose:        React.PropTypes.func,
-    onSave:         React.PropTypes.func,
-    showLevelForm:  React.PropTypes.bool,
+  props: {
+    ui:                           UIState,
+    setEducationDialogVisibility: () => void,
+    setEducationDegreeLevel:      () => void,
+    setEducationDialogIndex:      () => void,
+    clearProfileEdit:             () => void,
+    saveProfile:                  SaveProfileFunc,
+    profile:                      Profile,
+    showLevelForm:                boolean,
+    validator:                    Validator|UIValidator,
   };
 
   clearEducationEdit: Function = (): void => {
@@ -57,7 +65,7 @@ export default class EducationDialog extends ProfileFormFields {
       profile: { education },
     } = this.props;
 
-    let educationDegreeLevel = _.get(education, [educationDialogIndex, "degree_name"]) || BACHELORS;
+    let educationDegreeLevel = _.get(education[educationDialogIndex], "degree_name") || BACHELORS;
     let keySet = (key) => ['education', educationDialogIndex, key];
 
     let fieldOfStudy = () => {
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -24,23 +24,27 @@ import type {
   EducationEntry,
   Profile,
   ValidationErrors,
-  BoundSaveProfile,
+  SaveProfileFunc,
+  UpdateProfileFunc,
 } from '../flow/profileTypes';
 import type { UIState } from '../reducers/ui';
+import type { Validator, UIValidator } from '../util/validation';
+import type { AsyncActionHelper } from '../flow/generalTypes';
 
 class EducationForm extends ProfileFormFields {
   props: {
     profile:                          Profile,
     ui:                               UIState;
-    updateProfile:                    () => void,
-    saveProfile:                      BoundSaveProfile,
+    updateProfile:                    UpdateProfileFunc,
+    saveProfile:                      SaveProfileFunc,
     clearProfileEdit:                 () => void,
     errors:                           ValidationErrors,
     setEducationDialogVisibility:     () => void,
     setEducationDialogIndex:          () => void,
     setEducationDegreeLevel:          () => void,
-    setEducationDegreeInclusions:     () => void,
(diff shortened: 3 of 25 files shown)


static/js/actions/ui.js
```diff
@@ -22,9 +23,12 @@ export const setDialogVisibility = (bool: boolean): Action => (
 
 // work history actions
 export const SET_WORK_HISTORY_EDIT = 'SET_WORK_HISTORY_EDIT';
-export const setWorkHistoryEdit = (bool: boolean): Action => (
-  { type: SET_WORK_HISTORY_EDIT, payload: bool }
-);
+export const setWorkHistoryEdit: Function = (bool: boolean): (d: Dispatch) => Promise => {
```
gsidebo: 2 things:

1. can we change the name of `bool` to `flag` or something? it's a nitpick for such a small function, but i think it's a little more descriptive and 'flag' isn't a type declaration in other languages
1. this declaration syntax is a little mind-bending. can you explain in english just for my benefit? some sources of confusion: `d` in `(d: Dispatch)` doesn't appear to be used, `(d: Dispatch)` seems like it might be a return type, but then `Promise` looks like the actual return type.

aliceriot: the declaration reads (in English) something like:

`setWorkHistoryEdit` is a function. it accepts an argument `bool`, of type `boolean`, and returns a function which takes an argument `d`, of type `Dispatch`, and returns a promise.

it's a little confusing because the annotation for function types looks like `(arg: argType) => returnType`, but if you're annotating an arrow function literal you do `(arg: argType): returnType => { ... }`.

static/js/components/EducationDialog.js
```diff
@@ -23,11 +26,16 @@ export default class EducationDialog extends ProfileFormFields {
...
-    onClose:        React.PropTypes.func,
-    onSave:         React.PropTypes.func,
-    showLevelForm:  React.PropTypes.bool,
+  props: {
+    ui:                           UIState,
+    setEducationDialogVisibility: () => void,
+    setEducationDegreeLevel:      () => void,
+    setEducationDialogIndex:      () => void,
+    clearProfileEdit:             () => void,
+    saveProfile:                  BoundSaveProfile,
```
gsidebo: you've added some types here, which is awesome, but a few of them have sort of jarring names (`BoundSaveProfile`, `BoundUpdateProfile`, `SetWorkHistoryEdit` etc). let's say i'm reading through these props and trying to get a sense of what everything is.

> what is this 'profile' prop? oh, it's a Profile. what is the 'showLevelForm' variable supposed to be? oh, it's a 'boolean'. what is 'saveProfile'? it's a... BoundSaveProfile. wat?

i think that type should be called `ProfileSaveFunction` or something like that. "what am i going to be getting in the 'saveProfile' prop? ah, i'll be getting a 'ProfileSaveFunction'." 

that is, of course, assuming that we need a declared type. my examples above (`BoundSaveProfile`, `BoundUpdateProfile`, `SetWorkHistoryEdit`) seem to be single functions, ie: there aren't several functions that conform to that 'type'. i'll defer to you to decide whether we need to keep these around as types

aliceriot: it's a declared type because we need to annotate it in several places, flow enforces type declarations at module boundaries so there's a little bit of redundancy there. So we need to have the declared type (it's declared in `flow/profileTypes.js`) in order to share the annotation (and not have `(validator: Validator|UIValidator, profile: Profile, ui: UIState) => Promise<Profile>`, which is a bit verbose, all over the place).


I called it `BoundSaveProfile` because it's the result of taking the `saveProfile` method on `ProfileFormContainer` and binding `this` and the first positional argument (`isEdit`).

static/js/components/EmploymentForm.js
```diff
@@ -20,8 +20,35 @@ import SelectField from './inputs/SelectField';
...
+    saveProfile:                      BoundSaveProfile,
+    clearProfileEdit:                 () => void,
+    errors:                           ValidationErrors,
+    setWorkDialogVisibility:          () => void,
+    setWorkDialogIndex:               () => void,
+    setWorkDegreeLevel:               () => void,
+    setWorkDegreeInclusions:          () => void,
+    setWorkHistoryEdit:               SetWorkHistoryEdit,
+    deletionIndex:                    number,
+    setShowWorkDeleteAllDialog:       (bool: boolean) => void,
```
gsidebo: is there a way that we can avoid specifying the actual name of the argument in these type declarations? this seems like it would add extra pain to renaming arguments with no apparent benefit

aliceriot: this annotation (`(bool: boolean) => void`) just tells flow to check that the first positional argument to the function is of type `boolean`, it doesn't actually check that the variable name in the function declaration is the same.

static/js/components/PersonalTab.js
```diff
@@ -30,7 +35,7 @@ class PersonalTab extends React.Component {
       <Grid className="profile-tab-grid">
         <Cell col={1} />
         <Cell col={10}>
-          <PersonalForm {...this.props} />
+          <PersonalForm {...this.props} validator={personalValidation} />
```
gsidebo: any reason that we can't just import `personalValidation` directly in `PersonalForm` and get rid of the import here?

static/js/components/EmploymentForm.js
aliceriot: to me it's a bit weird that you need to give them names at all in function type declarations, but maybe it's a difficulty they ran into when designing the syntax.

static/js/components/PersonalTab.js
aliceriot: we're pulling `validator` out of `this.props` in the `onChange methods in the various input field methods - so we could import `personalValidation` in `PersonalForm` and set `defaultProps`. I originally was using this pattern for the components where we want the same validation in all contexts, but I had a lot of trouble getting this to type-check (I think because `PersonalForm` and the other form components subclass `ProfileFormFields`), so I went with this pattern instead.

static/js/actions/ui.js
aliceriot: `bool` is the argument name we use a lot elsewhere for other action creators, so I'd either want to stick with it or change all of them, and overall I'd probably prefer to just stick with the convention we've been using.

static/js/components/EducationDialog.js
gsidebo: to the first paragraph, is it possible to be more vague and tell flow that the type is just `Function` and be done with it? maybe it

to the second paragraph, my comments still hold. `boolean` makes sense as a type. `Object` makes sense as a type. `UIValidator` makes sense as a type, in that it suggests that i'm probably expecting a function for this variable that performs some validation. `BoundSaveProfile` makes some sense as the name of a _function_, but IMO it does not make sense as the name of a _type_. `ProfileSaveFunction`, et al, tells me what to expect here more accurately. 

as for `Bound`, we can't have any expectation at the _type_ level that this function has been bound to something before it got here. all the _type_ can say is that it's a function that accepts certain arguments and returns a value of a certain type. the "bound" part of the name is an implementation detail, and i don't think we should be insisting on that in the name of the type

gsidebo: addendum to the above: i think we should get rid of the `SetWorkHistoryEdit` and `SetEducationDegreeInclusions` as named types. they're each used to describe one function, then they're used in one place each to indicate that a prop is of that type. this pattern suggests a world where every single function we declare needs an accompanying type declaration that exists alongside it, whether other functions will ever share that type or not. i would really not want to live in that world

gsidebo: re: 'BoundSaveProfile', sounds good 👍 

re: `SetWorkHistoryEdit` and `SetEducationDegreeInclusions`, i see how it's valuable to insist that those props will be functions that return a Promise. that way we can invoke `setEducationDegreeInclusions` as a function, and we know that we can use `.then()` after the invocation. can we declare a type called `Dispatcher` or something that's vague enough to say that it's a function that returns a Promise and not worry about the exact arguments? i really don't agree with declaring a custom named type for each of these functions that are used in exactly one place. it bloats the codebase with relatively little added value, and like i suggested above it sets a precedent for creating a custom type for every single function we create. what if we change other functions like `setEducationDialogIndex` or `setEducationDegreeLevel` to return the Promise from `dispatch`? will we declare two more types then? if you have more questions about this, let's chat in person.

static/js/actions/ui.js
gsidebo: sounds good

static/js/components/PersonalTab.js
gsidebo: ok i think i see what's going on. looks like `<EmploymentForm>` (just as an example) gets a different validator depending on the parent component. makes sense to keep this as is

aliceriot: Yeah exactly, when we're on the profile we want to validate the UI stuff (the switches and all that) but we don't use the switches on the users page.

static/js/components/EducationDialog.js
aliceriot: yeah, I agree with making a more generic type - a lot of our async actions arent annotated but it would also be handy if we add annotation to them.

aliceriot:
@gsidebo comments should be addressed!

gsidebo:
functionality looks good. let me run through the code again

static/js/components/EducationDialog.js
```diff
@@ -23,11 +26,16 @@ export default class EducationDialog extends ProfileFormFields {
...
-    onClose:        React.PropTypes.func,
-    onSave:         React.PropTypes.func,
-    showLevelForm:  React.PropTypes.bool,
+  props: {
+    ui:                           UIState,
+    setEducationDialogVisibility: () => void,
+    setEducationDegreeLevel:      () => void,
+    setEducationDialogIndex:      () => void,
+    clearProfileEdit:             () => void,
+    saveProfile:                  SaveProfile,
```
gsidebo: again, i think it would be a lot more helpful if this type name was `SaveProfileFunction`. it's a more accurate description of the type we're expecting, and it avoids potential confusion with the `Profile` type. same comment goes for `UpdateProfile`

static/js/components/EducationForm.js
```diff
@@ -24,23 +24,27 @@ import type {
   EducationEntry,
   Profile,
   ValidationErrors,
-  BoundSaveProfile,
+  SaveProfile,
+  UpdateProfile,
 } from '../flow/profileTypes';
 import type { UIState } from '../reducers/ui';
+import type { Validator, UIValidator } from '../util/validation';
+import type { AsyncActionHelper } from '../flow/generalTypes';
```
gsidebo: yeah, i like this a lot more

gsidebo:
👍  once CI checks finish! thanks for bearing with me 😄

PR #711, opened 2016-07-12
Added models for Enrollment and Certificate and code to populate them

### LINKED ISSUE #677: add model(s) for users' enrollments, grades, certificates
To start with, we will update this data (synchronously) whenever the user looks at his dashboard, if it's older than a configured amount of time. 

When the user makes an API request, we check the date of last update. If it's been longer than a set amount (say 6 hours), we make an upstream request to edX's API and store the result in our database. 

We may need another story for doing batch updates.

noisecapella:
Fixes #677
Rebased on #706

#### What's this PR do?
Adds four models: Certificate, CertificateAccess, Enrollment, EnrollmentAccess. They cache data coming back from edX. After six hours for certificates or 5 minutes for enrollments, the cache is cleared for that user and new data is retrieved.

dashboard/api.py

Look at your devstack console. Access the dashboard. You should see some requests to the enrollments and possibly certificates APIs. Immediately afterwards access the dashboard again. You should see the same data in the dashboard but no requests to the edX APIs in the console.


dashboard/api.py
```diff
@@ -367,3 +376,107 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+        edx_client (EdxApi): EdX client to retrieve enrollments
+        course_ids (list of str): List of course keys
+    Returns:
+        Enrollments: an Enrollments object from edx_api. This may contain more enrollments than correspond to
+            course_ids if more exist from edX, or it may contain fewer enrollments if they
+            don't exist for the course id.
+    """
+    # Data in database is refreshed after 5 minutes
+    now = datetime.datetime.now(tz=pytz.utc)
+    five_minutes_ago = now - datetime.timedelta(0, 0, 0, 0, 5)
```
giocalitri: I would totally put this number as a global variable in this module and make the variable name less specific.

dashboard/api.py
```diff
@@ -367,3 +376,107 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+    enrollments_list = list(models.Enrollment.objects.filter(
```
giocalitri: coercing this to a `list` forces the ORM to load the date from the DB. 
I do not think you need to have the data in a list, looking briefly at how you use it.

dashboard/api.py
```diff
@@ -367,3 +376,107 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+        models.Enrollment.objects.filter(user=user, course_run__edx_course_key__in=course_ids).delete()
+
+        for enrollment in enrollments.enrollments.values():
+            if enrollment.course_id in course_ids:
+                models.Enrollment.objects.create(
+                    user=user,
+                    course_run=CourseRun.objects.get(edx_course_key=enrollment.course_id),
+                    data=enrollment.json,
+                    last_request=now,
+                )
```
giocalitri: why don't you just update?

giocalitri: I meant instead of delete/insert

dashboard/api.py
giocalitri: plus `timedelta` accepts kwargs parameters, so you can write 
`timedelta(minutes=5)` that looks much more readable

dashboard/api.py
```diff
@@ -367,3 +376,107 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+        edx_client (EdxApi): EdX client to retrieve enrollments
+        course_ids (list of str): A list of course keys
+    Returns:
+        Certificates: a Certificates object from edx_api. This may contain more certificates than correspond to
+            course_ids if more exist from edX, or it may contain fewer certificates if they
+            don't exist for the course id.
+    """
+    # Certificates in database are refreshed after 6 hours
+    now = datetime.datetime.now(tz=pytz.utc)
+    six_hours_ago = now - datetime.timedelta(0, 0, 0, 0, 0, 6)
```
giocalitri: same comment about timedelta and making the variable global to this module

dashboard/api.py
```diff
@@ -367,3 +376,107 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+                    user=user,
+                    course_run=CourseRun.objects.get(edx_course_key=enrollment.course_id),
+                    data=enrollment.json,
+                    last_request=now,
+                )
+
+        return enrollments
+
+
+def get_student_certificates(user, edx_client, course_ids):
```
giocalitri: same comments as the previous function

dashboard/api_test.py
```diff
@@ -799,3 +800,357 @@ def test_program_no_courses(self, mock_info_course):
...
+        cls.user = UserFactory.create()
+
+    def assert_certificates_in_db(self, expected_timestamp, course_ids, certificates=None):
+        """
+        Check that self.certificates matches what's in the database
+        """
+        if certificates is None:
+            certificates = self.certificates
+
+        one_minute_before = expected_timestamp - timedelta(0, 0, 0, 0, 1)
```
giocalitri: please use kwargs

dashboard/api_test.py
```diff
@@ -799,3 +800,357 @@ def test_program_no_courses(self, mock_info_course):
...
+        one_minute_after = expected_timestamp + timedelta(0, 0, 0, 0, 1)
+
+        certificate_list = list(models.Certificate.objects.filter(
+            user=self.user,
+            course_run__edx_course_key__in=course_ids,
+        ))
+        assert len(certificate_list) == len(course_ids)
```
giocalitri: better to use the ORM `.count()`

giocalitri:
few changes asked for the code, but nothing structural

dashboard/views.py
```diff
@@ -52,11 +55,10 @@ def get(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-s
         # create an instance of the client to query edX
         edx_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL)
         # get an enrollments client for the student
-        enrollments = edx_client.enrollments.get_student_enrollments()
+        course_keys = CourseRun.objects.filter(course__program__live=True).values_list("edx_course_key", flat=True)
```
giocalitri: you need to filter by `edx_course_key` not null or not empty:

I found a bug in my local instance where 
```
course_keys == ['course-v1:odl+MX+2015', 'course-v1:edX+DemoX+Demo_Course', 'course-v1:odl+GIO101+CR-FALL15', 'course-v1:odl+GIO101+FALL13', 'course-v1:odl+GIO101+FALL14', '']
```
note the empty string at the end of the list

giocalitri:
found a bug that prevents to test full functionality

dashboard/api.py
noisecapella: I was worried about race conditions, but I could just wrap the two calls with a transaction instead.

dashboard/api.py
noisecapella: We would need to keep track of enrollments which are updated and then delete the other ones, though that solution probably performs slightly better

dashboard/models.py
```diff
@@ -0,0 +1,40 @@
...
+        unique_together = (('user', 'course_run'), )
+
+
+class Certificate(Model):
+    """
+    Model for certificate data from edX
+    """
+    user = ForeignKey(User)
+    course_run = ForeignKey(CourseRun)
+    data = JSONField()
```
pdpinch: Does this include the final grade?

dashboard/models.py
```diff
@@ -0,0 +1,40 @@
+"""
```
gsidebo: don't know if this was your decision initially, but why are these models being defined in the 'dashboard' app? it makes logical sense for Profile/Education/etc to be defined in the 'profiles' app. Same for Course/CourseRun/Program being defined in the 'courses' app. i don't understand why these are defined in 'dashboard' when it seems like 'courses' would make a lot more sense

dashboard/api.py
```diff
@@ -367,3 +380,127 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+            or it may contain fewer enrollments if they don't exist for the course id in edX
+    """
+    # Data in database is refreshed after 5 minutes
+    now = datetime.datetime.now(tz=pytz.utc)
+    refresh_delta = now - datetime.timedelta(minutes=REFRESH_ENROLLMENT_CACHE_MINUTES)
+
+    with transaction.atomic():
+        course_ids = CourseRun.objects.filter(course__program__live=True).exclude(
+            Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
+        ).values_list("edx_course_key", flat=True)
```
gsidebo: this same query is used @ line 464. my first thought is that we should add a helper method on the CourseRun model (`get_live_course_ids` or something). if that isn't preferred, we should at least add a helper method in this module

dashboard/api.py
```diff
@@ -367,3 +380,127 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
         # TODO: here goes the logic to pull the current grade  # pylint: disable=fixme
         pass
     return formatted_run
+
+
+def get_student_enrollments(user, edx_client):
```
gsidebo: `get_student_enrollments` and `get_student_certificates` are basically the same thing with some minor changes, and the models for `Enrollment` and `Certificate` are identical. we can create a new parent class for those two models (`EdxCachedModel` or something?) and abstract away a ton of the logic for these two `get_` methods. that's a bigger change, of course. given the urgency with this issue, if we want to open up a separate issue to make that change, i'm cool with that. what do you think?

dashboard/api_test.py
```diff
@@ -799,3 +803,461 @@ def test_program_no_courses(self, mock_info_course):
             "courses": []
         }
         self.assertEqual(res, expected_data)
+
+
+class CertificatesTests(ESTestCase):
```
gsidebo: `CertificatesTests` and `EnrollmentsTests` are 200+ lines each of nearly identical code. if we want to hold off on refactoring this and do it as part of the issue i suggested in my comment on `dashboard/api.py`

dashboard/views_test.py
```diff
@@ -175,6 +177,22 @@ def test_with_runs(self, mocked_refresh):
                     if course_data['runs'][0]['course_id'] == "course-v1:MITx+8.MechCX+2014_T1":
                         assert course_data['runs'][0]['status'] == CourseStatus.UPGRADE
 
+    @patch('backends.utils.refresh_user_token', autospec=True)
+    def test_certificates_enrollments(self, mocked_refresh):
+        """
+        Make sure we call the proper api functions
```
gsidebo: More accurate: "Make sure we retrieve student certificates and enrollments whenever user dashboard data is fetched"

dashboard/views_test.py
```diff
@@ -175,6 +177,22 @@ def test_with_runs(self, mocked_refresh):
...
+        """
+        certificates = Certificates([])
+        enrollments = Enrollments([])
+
+        with patch('dashboard.views.get_student_certificates', autospec=True, return_value=certificates) as cert:
+            with patch('dashboard.views.get_student_enrollments', autospec=True, return_value=enrollments) as enroll:
```
gsidebo: i would prefer:
```python
with patch('dashboard.views.get_student_certificates', autospec=True, return_value=certificates) as cert, \
        patch('dashboard.views.get_student_enrollments', autospec=True, return_value=enrollments) as enroll:
    resp = self.client.get(self.url)
    assert resp.status_code == 200, resp.content.decode('utf-8')
```

dashboard/models.py
```diff
@@ -0,0 +1,40 @@
...
+    user = ForeignKey(User)
+    course_run = ForeignKey(CourseRun)
+    data = JSONField(null=True)
+    last_request = DateTimeField()
+
+    class Meta:
+        unique_together = (('user', 'course_run'), )
+
+
+class Certificate(Model):
```
gsidebo: what do you think about renaming these to `CachedEnrollment` and `CachedCertificate`? more accurately describes the data, and it would avoid naming collisions with `edx_api` classes

dashboard/models.py
giocalitri: I have not a strong opinion about this: I can move it if you feel strong about it. Let me know

dashboard/api.py
giocalitri: it is not that simple: for the certificate we pull the verified ones that are a subset of the certificates. I actually thought about turn these in calls to a generic function, but this generic function may turn to be pretty complicated.

dashboard/api_test.py
giocalitri: again, the tests look similar but some small part are pretty different, so move them in a single class can result in more work than this.

dashboard/views_test.py
giocalitri: the `\` should be avoided, but I can wrap with parentheses

dashboard/views_test.py
giocalitri: well, here we actually mock the calls, so we the only thing we do is checking that the mocked functions are properly called

dashboard/models.py
giocalitri: I do not have strong opinions: I can rename them

dashboard/models.py
giocalitri: Actually I was trying to do this right now and I realized that the APIs logically belong here, because they implement a cached interface to edX.
Now I could move only the models to `Profile`, but I am not sure it makes much sense.

dashboard/views_test.py
giocalitri: actually I was wrong: the context managers are one of the few cases the `\` is the only option...

dashboard/models.py
giocalitri: We discussed this at standup and decided to leave them in the dashboard

dashboard/api.py
gsidebo: right, so if Enrollment and Certificate inherit from the same parent class, then that would be a function that we would implement differently for Enrollment and Certificate. these two classes follow extraordinarily similar patterns, and the differences between them would be covered by implementing some specific function differently in each class. a lot of this code is copy/pasted (comments from lines 425 to 427 indicate 'certificate' and the function is pulling enrollments). that by itself is a pretty big indicator that we are missing some abstraction. again, i'm happy to push this to a different story since that is a bigger change and this issue is time-sensitive

dashboard/models.py
gsidebo: i don't agree with this, but i'll go with the crowd on this one. 

or what it's worth: a 'dashboard' is a _front-end_ concept. our data modeling should not have any idea what a 'dashboard' is. these are two models that relate directly to courses and course runs. if i'm looking for a final grade that a student got in a course run, i would expect to find that model in `courses/models.py`. same thing if i'm looking for a student's enrollment in a course. it just makes sense. the fact that we _show_ the grades in the dashboard on the front-end is, in my opinion, irrelevant to where we put these models. any API could ask for enrollment/certificate information. the fact that we have an api module in dashboard, and that happens to ask for enrollment/certificate information, shouldn't be a guide to where we put the models.

but again, i'll go with the crowd

dashboard/views_test.py
gsidebo: ok, then _why_ are we testing that the functions are called? what's the purpose? from my understanding, this test is here to make sure that we're retrieving student certificates and enrollments whenever user dashboard data is fetched. if that's not true, and the test has a different purpose, then i would update the docstring to indicate that. 'Make sure we call the proper api functions' is vague and it doesn't tell me why the test is here

giocalitri: the reason is because we test how the called functions work separately. So assuming they work properly the only thing to test here is that they are called in the right way.

gsidebo:
looks good to me 👍

dashboard/api.py
giocalitri: you are probably right about this, but I think we spent too much time on this PR already.
I am willing to revisit the code if I have to touch again or if we need to cache something else coming from edX

PR #713, opened 2016-07-12
Refactored ProfileFormContainer to remove boilerplate

### LINKED ISSUE #712: Refactor action helpers on `ProfileFormContainer`
it's messy and there is a lot of boilerplate.

aliceriot:

this closes #712 

#### What's this PR do?

Just abstracts out a little bit of boilerplate. We had a lot of helper methods on `ProfileFormFields` like the following:

```js
dispatchSomeAction: Function = (arg: someType): void => {
  dispatch(someAction(arg));
};
```

this abstracts that pattern, so instead we do:

```js
createActionHelper: Function = (actionCreator: Function): (...args: any) => void  => {
  return (...args) => dispatch(actionCreator(...args));
};
createActionHelper(someAction);
```

We can do this for most of the action creator helpers we had on `ProfileFormFields`, although anything with more specialized logic than the example above (e.g. `saveProfile`) needs to be left alone.

Read over the changes.

Make sure there are no regressions on any of the profile components (on `/profile` and `/users`).

### DIFF
--- a/static/js/containers/ProfileFormContainer.js
+++ b/static/js/containers/ProfileFormContainer.js
@@ -26,12 +26,12 @@ import {
   setDeletionIndex,
   setShowWorkDeleteAllDialog,
   setShowEducationDeleteAllDialog,
-  setProfileStep,
 } from '../actions/ui';
+import { createSimpleActionHelpers, createAsyncActionHelpers } from '../util/redux';
+import type { ActionHelpers, AsyncActionHelpers } from '../util/redux';
 import type { Validator, UIValidator } from '../util/validation';
 import type { Profile, Profiles, ProfileGetResult } from '../flow/profileTypes';
 import type { UIState } from '../reducers/ui';
-import type { AsyncActionHelper } from '../flow/generalTypes';
 
 type UpdateProfile = (isEdit: boolean, profile: Profile, validator: Validator|UIValidator) => void;
 
@@ -81,81 +81,6 @@ class ProfileFormContainer extends React.Component {
     }
     dispatch(updateProfile(username, profile));
     this.updateProfileValidation(this.props, profile, validator);
-  }
-
-  setProfileStep: Function = (step: string): void => {
-    const { dispatch } = this.props;
-    dispatch(setProfileStep(step));
-  };
-
-  setDeletionIndex: Function = (index: number): void => {
-    const { dispatch } = this.props;
-    dispatch(setDeletionIndex(index));
-  }
-
-  setShowEducationDeleteDialog: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setShowEducationDeleteDialog(bool));
-  }
-
-  setShowWorkDeleteDialog: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setShowWorkDeleteDialog(bool));
-  }
-
-  setShowWorkDeleteAllDialog: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setShowWorkDeleteAllDialog(bool));
-  };
-
-  setShowEducationDeleteAllDialog: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setShowEducationDeleteAllDialog(bool));
-  };
-
-  setUserPageDialogVisibility: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setUserPageDialogVisibility(bool));
-  };
-
-  setWorkHistoryEdit: AsyncActionHelper = (bool: boolean) => {
-    const { dispatch } = this.props;
-    return dispatch(setWorkHistoryEdit(bool));
-  }
-
-  setWorkDialogVisibility: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setWorkDialogVisibility(bool));
-  }
-
-  setWorkDialogIndex: Function = (index: number): void => {
-    const { dispatch } = this.props;
-    dispatch(setWorkDialogIndex(index));
-  }
-
-  clearProfileEdit: Function = (): void => {
-    const { dispatch } = this.props;
-    dispatch(clearProfileEdit(SETTINGS.username));
-  }
-
-  setEducationDialogVisibility: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setEducationDialogVisibility(bool));
-  }
-
-  setEducationDialogIndex: Function = (index: number): void => {
-    const { dispatch } = this.props;
-    dispatch(setEducationDialogIndex(index));
-  }
-
-  setEducationDegreeLevel: Function = (level: string): void => {
-    const { dispatch } = this.props;
-    dispatch(setEducationDegreeLevel(level));
-  };
-
-  setEducationDegreeInclusions: AsyncActionHelper = (inclusions: Object) => {
-    const { dispatch } = this.props;
-    return dispatch(setEducationDegreeInclusions(inclusions));
   };
 
   saveProfile(isEdit: boolean, validator: Validator|UIValidator, profile: Profile, ui: UIState) {
@@ -176,6 +101,32 @@ class ProfileFormContainer extends React.Component {
       return Promise.reject(errors);
     }
   }
+
+  simpleActionHelpers: Function = (): ActionHelpers => {
+    const { dispatch } = this.props;
+    return createSimpleActionHelpers(dispatch, [
+      setWorkDialogVisibility,
+      setWorkDialogIndex,
+      clearProfileEdit,
+      setEducationDialogVisibility,
+      setEducationDialogIndex,
+      setEducationDegreeLevel,
+      setUserPageDialogVisibility,
+      setShowEducationDeleteDialog,
+      setShowWorkDeleteDialog,
+      setDeletionIndex,
+      setShowWorkDeleteAllDialog,
+      setShowEducationDeleteAllDialog,
+    ]);
+  };
+
+  asyncActionHelpers: Function = (): AsyncActionHelpers => {
+    const { dispatch } = this.props;
+    return createAsyncActionHelpers(dispatch, [
+      setEducationDegreeInclusions,
+      setWorkHistoryEdit
+    ]);
+  };
 
   profileProps: Function = (profileFromStore: ProfileGetResult) => {
     let { ui } = this.props;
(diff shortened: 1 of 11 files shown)


static/js/containers/ProfileFormContainer.js
```diff
@@ -81,78 +81,18 @@ class ProfileFormContainer extends React.Component {
     }
     dispatch(updateProfile(username, profile));
     this.updateProfileValidation(this.props, profile, validator);
-  }
+  };
 
   setProfileStep: Function = (step: string): void => {
     const { dispatch } = this.props;
     dispatch(setProfileStep(step));
   };
```
aliceriot: I left this as-is because we use `this.setProfileStep` on `ProfilePage`, but now that I think about it we could just use `createActionHelper` there too...

static/js/containers/ProfileFormContainer.js
```diff
@@ -177,6 +111,28 @@ class ProfileFormContainer extends React.Component {
     }
   }
 
+  createActionHelper: Function = (actionCreator: Function): (...args: any) => void  => {
+    const { dispatch } = this.props;
+    return (...args) => dispatch(actionCreator(...args));
+  };
```
aliceriot: another question is if it makes sense to put this in another file, so that we could re-use a similar pattern on other container components (same for the method below). then we'd just do:

```js
import { createActionHelper } from '../util/redux';

createActionHelper.call(this, someAction);
```

aliceriot: or `createActionHelper` could take dispatch as a parameter, hmmm

aliceriot: I think I'm going to do that so we can write unit tests

gsidebo:
this has been irking me for a while now. thanks for doing something about it! i'll try to take a look later today

static/js/containers/ProfileFormContainer.js
gsidebo: just wondering: why can't we just import dispatch where it's needed? why does it need to be passed down as a prop?

aliceriot:
haha, me too, every time I had to add another one of those boilerplate functions it felt dirty and wrong.

aliceriot:
@gsidebo I think this is ready for review whenever you're free

static/js/containers/ProfileFormContainer.js
aliceriot: Because dispatch is defined at runtime by the store, so we wouldn't be able to import something.

I think the only alternative to passing it down as a prop would be sticking it on the global object (or on `window` or something).

gsidebo:
(continuing conversation from outdated diff re: dispatch)

sticking dispatch on `window` or something would indeed be yucky, but i consider the current approach (pass as a prop to All The Things) to be pretty yucky as well. i stumbled across this thread while looking into this: https://github.com/reactjs/redux/issues/916. any reason we aren't just using `connect()` for the components that need to use `dispatch`? we do it in static/js/containers/App.js

aliceriot:
Right now we're following basically the container component approach, where we have container components (like `App.js` and `ProfilePage.js` and so forth) which are connected to the store, and they pass data from the store and the `dispatch` function to presentational components, which are never connected to the store and pretty much just receive props and render UI with them. I think there's some value to retaining the separation of concerns there, where the container components have all the logic around creating helper methods for dispatching actions, doing higher level data fiddling (figuring out which profile to pass to a presentational component, for instance), and so on, and the presentational components just worry about being given, say, a profile and some callbacks and drawing the right UI with those inputs.

Does that make sense? We do have quite a number of different components which are connected to the store (basically everything in `static/js/containers`) right now, I'm not exactly sure what would be different about what you're suggesting.

static/js/containers/ProfileFormContainer.js
```diff
@@ -177,6 +113,24 @@ class ProfileFormContainer extends React.Component {
...
+      [clearProfileEdit, "clearProfileEdit"],
+      [setEducationDialogVisibility, "setEducationDialogVisibility"],
+      [setEducationDialogIndex, "setEducationDialogIndex"],
+      [setEducationDegreeLevel, "setEducationDegreeLevel"],
+      [setUserPageDialogVisibility, "setUserPageDialogVisibility"],
+      [setShowEducationDeleteDialog, "setShowEducationDeleteDialog"],
+      [setShowWorkDeleteDialog, "setShowWorkDeleteDialog"],
+      [setDeletionIndex, "setDeletionIndex"],
+      [setShowWorkDeleteAllDialog, "setShowWorkDeleteAllDialog"],
+      [setShowEducationDeleteAllDialog, "setShowEducationDeleteAllDialog"],
```
gsidebo: perhaps a little controversial, but one way to avoid this key-value repetition would be to use `import * as ui_actions from 'path/to/actions'`. might make the most sense to put that in the util file rather than here. that way you could just pass a list of strings, eg: `["setWorkDialogVisibility", "setWorkDialogIndex", "clearProfileEdit"]` and we wouldn't have this slightly awkward list-of-lists-with-repeated-names, nor the large import statement above. yes, that would open up the possibility of passing a function name that doesn't exist in actions/ui, but we could make the list of action names a property of this class and test its validity in a unit test

gsidebo: another (cleaner?) option would be to pass in a list of the action functions an use `.name`. the util function could just use `actionCreator.name` to get the string

static/js/containers/ProfileFormContainer.js
```diff
@@ -81,78 +82,13 @@ class ProfileFormContainer extends React.Component {
...
-
-  setUserPageDialogVisibility: Function = (bool: boolean): void => {
-    const { dispatch } = this.props;
-    dispatch(setUserPageDialogVisibility(bool));
   };
 
   setWorkHistoryEdit: AsyncActionHelper = (bool: boolean) => {
     const { dispatch } = this.props;
     return dispatch(setWorkHistoryEdit(bool));
   }
```
gsidebo: should be pretty easy to follow that createSimpleActionHelpers pattern with the `AsyncActionHelper` type functions, no?

gsidebo:
makes sense. i was confusing myself a bit when i wrote that comment. my main issue was that we're passing this mound of functions down the entire component tree as props, many of whom would never need those functions. i think i understand the reasoning now, though

side comment: i'm not sure why we're doing this `export default connect(ProfileFormContainer.mapStateToProps)(ProfilePage)` on every class that inherits from `ProfileFormContainer` instead of just doing that on the actual `ProfileFormContainer` class. no need to change that for this PR, though

static/js/containers/ProfileFormContainer.js
aliceriot: yes I think so, I had just left them alone since there's only two. Probably makes sense just to do it though.

static/js/containers/ProfileFormContainer.js
aliceriot: when I was initially writing this I ran into some issue with `func.name`, can't figure out what it was now though! haha

I'll rewrite it to use that.

gsidebo: if that doesn't work for whatever reason, i think you can also use `func.prototype.constructor.name` or something gross like that

aliceriot:
I just tried that, and it does not work. It seems that subclassing a class connected with to the store with `connect` does not make the child class connected to the store, so unfortunately we need that on each class subclassing `ProfileFormFields`.

We could write a mix-in to add that I think, but those are always a bit messy.

gsidebo:
LGTM pending CI build 👍

aliceriot:
haha hold your horses, haven't written tests yet :P

aliceriot:
Ok tests are done!

static/js/util/redux.js
```diff
@@ -0,0 +1,25 @@
...
+    { [actionCreator.name]: createActionHelper(dispatch, actionCreator) }
+  ));
+}
+
+export type AsyncActionHelpers = Array<{[k: string]: AsyncActionHelper}>;
+export function createAsyncActionHelpers(dispatch: Dispatch, actionManifest: AsyncActionCreator[]): AsyncActionHelpers {
+  return actionManifest.map(actionCreator => (
+    { [actionCreator.name]: createActionHelper(dispatch, actionCreator) }
+  ));
+}
```
gsidebo: wait... these are the same function `createSimpleActionHelpers ` and `createAsyncActionHelpers`. i wondered this before but forgot about it: why not just return the promise from `dispatch` in all cases? if the client doesn't do anything with the return value, that's on them. don't think it has any performance penalty. then we could combine these two into one called `createActionHelpers`.

also, i think we should rename `actionManifest` to `actions` or `actionList`

aliceriot: We could turn all the action creators into async ones, but we'd have to refactor all of the synchronous ones to make that work (which seems out of scope for this pr). I think we also rely on synchronicity for a lot of them, so I imagine that would be a really big refactor.

agree re: actionManifest, it's not really a manifest any more, haha.

gsidebo: i'm missing something here. the body of `createSimpleActionHelpers` and `createAsyncActionHelpers` is literally the same thing. i added some confusion with my comment about returning the `dispatch` value. in brief: either we're missing something that should be differentiating these two functions or we should combine them, right?

aliceriot: they're differentiated because they have different return types, `createSimpleActionHelpers` returns this:

```js
Array<{[k: string]: (...args: any) => void}>
```

an array of objects with a string key and a function that takes some arguments and returns `undefined`. these are the simple synchronous action creators that just take a value and dispatch a change to the store.

`createAsyncActionHelpers` has the following return type:

```js
Array<{[k: string]: (...a: any) => Promise}>
```

so an array of objects mapping a string to a function that takes some arguments and returns a promise. these are the async action creators.

note: we're using the [redux-thunk](https://open.spotify.com/album/16i5KnBjWgUtwOO7sVMnJB) middleware for asynchronous actions.

gsidebo: so specifying the return type actually has an affect on the functionality? i thought it was just a sort of 'wrapper' to validate types going in and out. i'm so confused - it looks like we have 2 blocks of the same exact javascript, but we're getting something different from each block because we've added flow typing. i'm willing to thumbs-up this, but i'd like to understand this first

aliceriot: the flow type annotations are all stripped out before the code is parsed by Babel for transpilation, so this is a case where we have to declare another function to get the typechecker to pass. Alternatives would require checking the return type in a lot of places to see if it is a `Promise` before calling `.then` on it, I think this is a better solution since the actual functionality is in the shared `createActionHelper` function anyway and these two functions are basically just wrappers around that.

gsidebo: can you leave a brief comment for possibly-confused devs (read: me) looking at this in the future? aside from that, this looks good

gsidebo:
👍

static/js/util/redux.js
aliceriot: we'll get "can't call method `.then` on possibly undefined value" errors, I had that setup initially but couldn't find an easy way to get it to pass.

Issue #682, opened 2016-07-01
filter by average grade (back end)


pdpinch:
As a MicroMaster team member, I need to filter learner profile search results by grade range. 

The proposed UI includes both summary data from the search results and a control for specifying the grade:

![image](https://cloud.githubusercontent.com/assets/430126/16533477/edf0bcd2-3fa6-11e6-896a-e1586a65930d.png)

- [ ] When certificates & final grades are updated, update the index with an average of all the learner's final grades
- [ ] add the average final grade as a facet in the search results (API)
- [ ] How will we deal with learners with an average of Zero? Add another facet for number of courses completed.

gsidebo:
we can definitely look into front-end components for this, and we can always mock out data as needed, but i spoke with @noisecapella last week, and it seemed like there was some uncertainty about how we're going to fetch or store grade data

pdpinch:
The only data that we have at this time is the final grade for each certificate. It's stored in the certificate model (data field).

pdpinch:
Can Elasticsearch calculate the average on the search results, to find an "overall grade" or will we need to add that to the index explicitly? 

A quick read of the docs suggests yes.

gsidebo:
@pdpinch 

> The only data that we have at this time is the final grade for each certificate. It's stored in the certificate model (data field).

do we? isn't this model being added as part of #677, which is still in progress? i don't think this issue can be started until that's done. i'm more than happy to do some ES study in the meantime

pdpinch:
#677/#678 are closed/merged, so we have the necessary component data. 

Further discussion and evaluation of elastic search has lead us to conclude, however, that we need to calculate and add the average grade (per program) to the elastic search document, instead of relying on elasticsearch to aggregate the grades at query-time.

pdpinch:
Blocked by #774

gsidebo:
### Requirements update

- We will be calculating a user's overall grade on a *per program basis* in our app before sending it to be indexed in ES. Filtering by grade will act on these per-program values. Adding this calculation and serializing it should be done as part of 
- As a result, the filter-by-grade widget *should only be available when a single Program is selected as a filter/facet*. Grade-filtering paramaters should not be added to the ES query unless a single program is selected

Issue #672, opened 2016-07-01
Add MicroMasters program enrollment to user model 


pdpinch:
As a learner and as a MicroMasters team member, I want the portal to recognize what specific MicroMasters I am part of. 

We need to add this to the user model (foreign key(s) to programs), and also determine how best to populate it. For now, we will assume that if the user is enrolled in any courserun associated with a particular MM program, they qualify as enrolled. 

We'll need to update this whenever we update the user's enrollments. 

depends on the story about enrollment models #677

pdpinch:
Or we can calculate this from the user's enrollments

gsidebo:
if we're adding a reference to a user's Enrollments, we should do the same for Certificates. both of them have foreign keys to a User. pretty sure they would just need a 'related_name' like Education or Employment have in their ForeignKey relationship with Profile

gsidebo:
@annagav are you back on this or is @giocalitri working on it? this has some implications for #682. i think it make sense to add a user's overall program grade to this model, which we would use when filtering users by grade. we could push that to a different issue, but #682 depends on having a pre-calculated average grade for a program and this issue seems like a good place to address that

giocalitri:
the PR is open for review

PR #720, opened 2016-07-13
Added indexing for Certificate and Enrollment

### LINKED ISSUE #678: update profile index when user's enrollments, grades and certificates are updated
When a MM learner views their dashboard, we may discover new enrollments. If that is the case, we need to update the search index.

noisecapella:
Rebased on #711
Fixes #678

#### What's this PR do?
Indexes the Certificate and Enrollment models when they change (the user accesses the dashboard).

Wait 5 minutes if you have accessed the dashboard recently. Then watch the elasticsearch console. Go to the dashboard and refresh the browser. You should see some activity in the console.

### DIFF
--- a/dashboard/factories.py
+++ b/dashboard/factories.py
@@ -15,8 +15,8 @@ from factory.fuzzy import (
 import faker
 
 from dashboard.models import (
-    Certificate,
-    Enrollment,
+    CachedCertificate,
+    CachedEnrollment,
 )
 from courses.factories import CourseRunFactory
 from profiles.factories import UserFactory
@@ -25,25 +25,25 @@ from profiles.factories import UserFactory
 FAKE = faker.Factory.create()
 
 
-class CertificateFactory(DjangoModelFactory):
+class CachedCertificateFactory(DjangoModelFactory):
     """Factory for Certificate"""
     user = SubFactory(UserFactory)
     data = FuzzyAttribute(lambda: {'key{}'.format(i): FAKE.text() for i in range(3)})
     course_run = SubFactory(CourseRunFactory)
     # certificates expire after 6 hours, this generates a last request between 6:15 hours ago and now
-    last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(0, 0, 0, 0, 6, 15))
+    last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(hours=6, minutes=15))
 
     class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,old-style-class
-        model = Certificate
+        model = CachedCertificate
 
 
-class EnrollmentFactory(DjangoModelFactory):
+class CachedEnrollmentFactory(DjangoModelFactory):
     """Factory for Enrollment"""
     user = SubFactory(UserFactory)
     data = FuzzyAttribute(lambda: {'key{}'.format(i): FAKE.text() for i in range(3)})
     course_run = SubFactory(CourseRunFactory)
     # enrollments expire after 5 minutes, this generates a last request between 10 minutes ago and now
-    last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(0, 0, 0, 0, 0, 10))
+    last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(minutes=10))
 
     class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,old-style-class
-        model = Enrollment
+        model = CachedEnrollment
--- a/dashboard/factories.py
+++ b/dashboard/factories.py
@@ -1,0 +1,49 @@
+"""
+Factories for dashboard
+"""
+from datetime import datetime, timedelta
+import pytz
+
+from factory import (
+    SubFactory,
+)
+from factory.django import DjangoModelFactory
+from factory.fuzzy import (
+    FuzzyAttribute,
+    FuzzyDateTime,
+)
+import faker
+
+from dashboard.models import (
+    Certificate,
+    Enrollment,
+)
+from courses.factories import CourseRunFactory
+from profiles.factories import UserFactory
+
+
+FAKE = faker.Factory.create()
+
+
+class CertificateFactory(DjangoModelFactory):
+    """Factory for Certificate"""
+    user = SubFactory(UserFactory)
+    data = FuzzyAttribute(lambda: {'key{}'.format(i): FAKE.text() for i in range(3)})
+    course_run = SubFactory(CourseRunFactory)
+    # certificates expire after 6 hours, this generates a last request between 6:15 hours ago and now
+    last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(0, 0, 0, 0, 6, 15))
+
+    class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,old-style-class
+        model = Certificate
+
+
+class EnrollmentFactory(DjangoModelFactory):
+    """Factory for Enrollment"""
+    user = SubFactory(UserFactory)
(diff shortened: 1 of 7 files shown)


dashboard/factories.py
```diff
@@ -0,0 +1,49 @@
...
+FAKE = faker.Factory.create()
+
+
+class CertificateFactory(DjangoModelFactory):
+    """Factory for Certificate"""
+    user = SubFactory(UserFactory)
+    data = FuzzyAttribute(lambda: {'key{}'.format(i): FAKE.text() for i in range(3)})
+    course_run = SubFactory(CourseRunFactory)
+    # certificates expire after 6 hours, this generates a last request between 6:15 hours ago and now
+    last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(0, 0, 0, 0, 6, 15))
```
giocalitri: please use kwargs for timedelta (only for the kwargs != 0)

search/signals.py
```diff
@@ -12,26 +12,32 @@
...
+from dashboard.models import (
+    Certificate,
+    Enrollment,
+)
 from search.tasks import index_users, remove_user
 
 log = logging.getLogger(__name__)
 
 
 @receiver(post_save)
```
giocalitri: I know this is old stuff, but unless I miss something I think this is wrong: you should attach the signal handler to the proper sender, otherwise this will be triggered for each `post_save` signal fired from any class.

search/signals.py
```diff
@@ -12,26 +12,32 @@
...
-def handle_profile_update(sender, **kwargs):  # pylint: disable=unused-argument
-    """Update index when a Profile, Education, or Employment is updated."""
+def handle_update(sender, **kwargs):  # pylint: disable=unused-argument
+    """Update index when a relevant model is updated."""
     instance = kwargs["instance"]
```
giocalitri: if you define the signal handler with an arg called `instance`, it will be unpacked from `kwargs` automatically

search/signals.py
giocalitri: oh... I see what's going on here: you filter afterwards. Still I think we should make a generic handler function and use it in specific signal handlers (one per sender).

noisecapella: I agree

search/signals.py
```diff
@@ -12,26 +12,32 @@
...
+    if isinstance(instance, (Certificate, Enrollment, Profile)):
         index_users.delay([instance.user])
     elif isinstance(instance, (Education, Employment)):
         index_users.delay([instance.profile.user])
 
 
 @receiver(post_delete)
-def handle_profile_delete(sender, **kwargs):  # pylint: disable=unused-argument
-    """Update index when a Profile, Education, or Employment is deleted."""
+def handle_delete(sender, **kwargs):  # pylint: disable=unused-argument
```
giocalitri: same point here: actually here it makes more sense to have multiple signal handlers given that you run different code

giocalitri:
some comments on the signals, but overall not bad

noisecapella:
Feel free to take over @giocalitri, I agree with all proposed changes

giocalitri:
I think it is ready for a review again

search/signals.py
```diff
@@ -12,26 +12,75 @@
...
-    """Update index when a Profile, Education, or Employment is deleted."""
-    instance = kwargs['instance']
-    if isinstance(instance, Profile):
-        remove_user.delay(instance.user)
-    elif isinstance(instance, (Education, Employment)):
-        index_users.delay([instance.profile.user])
+# all the following signal handlers do basically the same.
+# The reason why there is one function per sender is
+# because each signal handler needs to be hooked to a single sender
+# otherwise it would run for any `post_save`/`post_delete` coming from any model
```
gsidebo: we can probably generate these functions with less repetition using a syntax like this: `handle_<action>_<model> = receiver(...)`. i don't think that's too important though, so i won't hold up the PR for it. maybe i'll open up a separate issue

search/api_test.py
```diff
@@ -60,6 +65,24 @@ def search():
     return get(url).json()['hits']
 
 
+def mappings():
```
gsidebo: can you rename this to `current_es_mappings()` or `get_current_es_mappings()`? this snippet below is just a little confusing as it is:

    mapping = mappings()
    nodes = list(traverse_mapping(mapping))

search/api.py
```diff
@@ -228,6 +235,27 @@ def create_user_mapping():
             'state_or_territory': NOT_ANALYZED_STRING_TYPE,
         }},
     })
+    mapping.field('enrollments', 'nested', properties={
+        'course_details': {
```
gsidebo: just wondering: what do we gain from defining these specific properties in this mapping (as opposed to just defining it as `mapping.field('enrollments', 'nested')`)

search/api_test.py
```diff
@@ -191,6 +214,83 @@ def test_index_users(self):
...
+        assert_search(search(), [user])
+
+    def test_update_certificate(self):
+        """
+        Test that Certificate is reindexed after being updated
+        """
+        user = UserFactory.create()
+        assert search()['total'] == 1
+        certificate = CachedCertificateFactory.create(user=user)
+        certificate.data = [{'new': 'data'}]
```
gsidebo: is there a reason that the dummy data here is a list and the same dummy data for enrollments is a dict (line 262)?

search/api_test.py
```diff
@@ -191,6 +214,83 @@ def test_index_users(self):
         index_users(User.objects.iterator(), chunk_size=4)
         assert_search(search(), list(User.objects.all()))
 
+    def test_add_certificate(self):
```
gsidebo: what do you think about defining a list of cached model factories (`CachedFactories = [CachedCertificateFactory, CachedEnrollmentFactory]`) and looping through those to perform these add/update/delete tests? this can potentially be done as part of the refactor story ( #743 ) if we ever get around to it. this is just a lot of repetition and it seems to make sense to treat these cases in a more generic way

search/util.py
```diff
@@ -0,0 +1,16 @@
+"""Utility functions for search"""
+
+
+def traverse_mapping(mapping):
+    """
+    Traverse a mapping, yielding each node
+
+    Args:
+        mapping (dict): A dictionary within the mapping
```
gsidebo: a couple notes for this docstring:

1. I think the first line should read "Traverse a mapping, yielding each nested dict"
1. I think this line should read "mapping (dict): The mapping itself, or a nested dict"

gsidebo:
just some minor comments

gsidebo:
related to #682: we _could_ include some Course/Program info for a certificate in the index, right? pretty sure we would just need to add some extra logic into `serialize_user`

search/signals.py
giocalitri: mmmmm this might make our life a nightmare in case we find bugs... I agree we should discuss this outside this PR

search/api_test.py
giocalitri: reasonable

search/api.py
giocalitri: I think what George did was extracting the actual enrollment information from the nested part of the JSON: I actually did not pay much attention to this specific part.

search/api_test.py
giocalitri: need to check: I did not write these tests but I want to be sure there is a reason

search/api_test.py
giocalitri: I think they are more or less fine as they are.

giocalitri:
>related to #682: we could include some Course/Program info for a certificate in the index, right? pretty sure we would just need to add some extra logic into `serialize_user`

I think we need to wait for have whatever @annagav is doing in #672 to add this

search/util.py
giocalitri: ok

search/api_test.py
giocalitri: there is no reason and this should be a `dict`. It does not make any difference for the test, but I will change it anyway

giocalitri:
I think I replied and/or addressed all your comments.

gsidebo:
👍

PR #722, opened 2016-07-15
Pinned html5lib to fix build

noisecapella:
#### What's this PR do?
Fixes the broken build caused by a change in html5lib


giocalitri:
seriously it has that version number?
anyway 👍

noisecapella:
Yes, I just kept removing `9` from the current version number until the build passed

PR #738, opened 2016-07-19
Created management command to generated realistic-looking fake users

### LINKED ISSUE #675: create realistic fake users for testing learner profile search
We'll want to cover at least two values in each facet.

gsidebo:

Closes #675 

#### What's this PR do?

Creates a new Django management command that creates a set of realistic-looking users from a json file that I've added to the codebase.

    docker-compose run web ./manage.py gen_realistic_users

If you'd like to delete these test users..

    docker-compose run web ./manage.py delete_realistic_users

I have some scripts in a Jupyter notebook that generate the json files that you're seeing here. The scripts are reasonably parametrized/configurable so we can tweak the resulting set of users as we want. It will be easy to generate enrollment/grade data and profile pictures once those are ready to go. Some details about the user data and the parameters I went with:

- randomuser.me was used to generate the basic data for these users
- 60 users in total
- Nationality is split evenly among US, Canada, and Spain (those are just 3 of the nationalities that randomuser.me allows)
- Gender is split evenly among users within those nationalities (10 women & men in each group)
- I created some basic, real-looking education and work history choices that are randomly assigned

I kept the actual user data generation code out of the MM codebase on purpose.

Some things I can do if people prefer:

- Put the generation scripts up in a Github snippet
- Minify the json file

### DIFF
--- a/micromasters/utils.py
+++ b/micromasters/utils.py
@@ -1,6 +1,7 @@
 """
 General micromasters utility functions
 """
+import json
 from django.conf import settings
 
 
@@ -16,3 +17,12 @@ def webpack_dev_server_url(request):
     Get the full URL where the webpack dev server should be running
     """
     return 'http://{}:{}'.format(webpack_dev_server_host(request), settings.WEBPACK_DEV_SERVER_PORT)
+
+
+def load_json_from_file(project_rel_filepath):
+    """
+    Loads JSON data from a file
+    """
+    path = '{}/{}'.format(settings.BASE_DIR, project_rel_filepath)
+    with open(path, 'r') as f:
+        return json.load(f)
--- a/profiles/management/commands/gen_realistic_search_data.py
+++ b/profiles/management/commands/gen_realistic_search_data.py
@@ -1,0 +1,257 @@
+"""
+Generates a set of realistic users/programs to help us test search functionality
+"""
+from datetime import datetime, timedelta
+from django.core.management import BaseCommand
+from django.contrib.auth.models import User
+from profiles.models import Employment, Education
+from courses.models import Program, Course, CourseRun
+from dashboard.models import CachedCertificate, CachedEnrollment
+from micromasters.utils import load_json_from_file
+
+
+USER_DATA_PATH = 'profiles/management/realistic_user_data.json'
+PROGRAM_DATA_PATH = 'profiles/management/realistic_program_data.json'
+FAKE_USER_USERNAME_PREFIX = 'fake.'
+FAKE_PROGRAM_DESC_PREFIX = '[FAKE] '
+CACHED_MODEL_LAST_REQUEST = datetime.now() + timedelta(days=365)
+
+
+# Util functions
+
+def first_matching_item(iterable, predicate):
+    """Returns the first item in an iterable that matches a given predicate"""
+    return next(x for x in iterable if predicate(x))
+
+
+def filter_dict_by_key_set(dict_to_filter, key_set):
+    """
+    Takes a dictionary and returns a copy without keys that don't exist in a given set
+    """
+    return {key: dict_to_filter[key] for key in dict_to_filter.keys() if key in key_set}
+
+
+def get_non_relation_field_names(model_cls):
+    """
+    Gets a set of field names from a model class that aren't relation fields
+    """
+    return set(field.name for field in model_cls._meta.get_fields()  # pylint: disable=protected-access
+               if not field.is_relation)
+
+
+def deserialize_model_data_on_object(model_obj, data, save=True):
+    """
+    Sets field values on an existing model object using some supplied data
+    """
+    non_relation_field_names = get_non_relation_field_names(model_obj.__class__)
+    for k, v in filter_dict_by_key_set(data, non_relation_field_names).items():
+        setattr(model_obj, k, v)
+    if save:
+        model_obj.save()
+    return model_obj
+
+
+def deserialize_model_data(model_cls, data, relational_data=None):
+    """
+    Creates a new instance of a model class and fills in field values using some supplied data
+    """
+    non_relation_field_names = get_non_relation_field_names(model_cls)
+    model_data = filter_dict_by_key_set(data, non_relation_field_names)
+    if relational_data:
+        model_data.update(relational_data)
+    return model_cls.objects.create(**model_data)
+
+
+# User data deserialization
+
+class CachedModelDeserializer:
+    """
+    Base class for deserializing a model used for edX data caching (eg: Enrollments, Certificates)
+    """
+    model_cls = None
+
+    @classmethod
+    def _get_edx_course_key(cls, data):   # pylint: disable=unused-argument
+        """Gets the edx_course_key value from the data dictionary"""
+        raise NotImplementedError
+
+    @classmethod
+    def _fill_in_missing_data(cls, data, user, course_run):   # pylint: disable=unused-argument
+        """Fills in data using the associated User and CourseRun objects"""
+        raise NotImplementedError
+
+    @classmethod
+    def deserialize(cls, user, data, course_runs):
+        """
+        Creates a new cached model object and returns the associated CourseRun
+        """
+        edx_course_key = cls._get_edx_course_key(data)
+        course_run = first_matching_item(course_runs, lambda cr: cr.edx_course_key == edx_course_key)
+        data = cls._fill_in_missing_data(data, user, course_run)
+        cls.model_cls.objects.create(
+            user=user,
+            course_run=course_run,
+            data=data,
+            last_request=CACHED_MODEL_LAST_REQUEST
+        )
+        return course_run
+
+
+class CertificateDeserializer(CachedModelDeserializer):
+    """
+    CachedCertificate deserializer
+    """
+
+    model_cls = CachedCertificate
+    data_key = 'certificates'
+
+    @classmethod
+    def _get_edx_course_key(cls, data):
+        return data['course_id']
+
+    @classmethod
+    def _fill_in_missing_data(cls, data, user, course_run):
+        data['username'] = user.username
+        return data
+
+
+class EnrollmentDeserializer(CachedModelDeserializer):
+    """
+    CachedEnrollment deserializer
+    """
+
+    model_cls = CachedEnrollment
+    data_key = 'enrollments'
+
+    @classmethod
+    def _get_edx_course_key(cls, data):
+        return data['course_details']['course_id']
+
+    @classmethod
+    def _fill_in_missing_data(cls, data, user, course_run):
+        data['user'] = user.username
+        data['course_details'].update({
+            'course_start': course_run.start_date.isoformat(),
+            'course_end': course_run.end_date.isoformat(),
+            'enrollment_start': course_run.enrollment_start.isoformat(),
+            'enrollment_end': course_run.enrollment_end.isoformat()
+        })
+        return data
+
+
+CACHED_MODEL_DESERIALIZERS = [EnrollmentDeserializer, CertificateDeserializer]
+
+
+def deserialize_user_data(user_data, course_runs):
+    """
+    Deserializes a dict of mixed User/Profile data and returns the newly-inserted User
+    """
+    user_model_data = user_data.copy()
+    user_model_data.update(
+        dict(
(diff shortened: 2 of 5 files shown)


gsidebo:
this is set as 'work in progress' right now b/c we're waiting on enrollments/grades/etc models to be added to MM

Ferdi:
Can this PR focus on just the fields that we already have ?   Open another PR when other fields are ready. Or make it a requirement that the PR for adding other fields should come with support for fake user data.

gsidebo:
@Ferdi it does focus on fields we have (profiles, education, employment). i'm not sure we get any value from pushing this PR forward and merging it independently of the enrollments/grades data. i could be wrong on that. either way, anyone can feel free to test this for the creation of profile/education/employment data

Ferdi:
The benefit is that we can have something to test sooner. Also, we are adding/removing/changing fields every day. It will slow you down if you keep waiting for every field. And you can let the developer that changes/add fields be responsible for adding "fake user" support.

gsidebo:
@Ferdi i was under the impression that users wouldn't even show up in search results if they didn't have any enrollment data. that may just be an incorrect assumption. at any rate, 2 other things to consider:

- learner search is not yet ready, so this is cannot yet be tested for our intended use anyway
- not waiting for every field, just a couple specific models. the story to add those models is in progress ( #677 ), and it's likely that it will be merged before learner search

gsidebo:
this is ready for review

micromasters/utils.py
```diff
@@ -16,3 +17,12 @@ def webpack_dev_server_url(request):
...
     return 'http://{}:{}'.format(webpack_dev_server_host(request), settings.WEBPACK_DEV_SERVER_PORT)
+
+
+def load_json_from_file(project_rel_filepath):
+    """
+    Loads JSON data from a file
+    """
+    path = '{}/{}'.format(settings.BASE_DIR, project_rel_filepath)
+    with open(path, 'r') as f:
+        return json.loads(f.read())
```
aliceriot: I think you can do

```python
json.load(f)
```

here instead, I do very little python stuff these days so I don't know if there's a reason not to pass the file handle around.

aliceriot:
deleting and adding users works well, gonna read through the code now.

profiles/management/commands/delete_realistic_users.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Command(BaseCommand):
+    """
+    Deletes a set of realistic users and programs/courses that were generated to help us test search
+    """
+    help = "Deletes a set of realistic users and programs/courses that were generated to help us test search"
+
+    def handle(self, *args, **options):
+        User.objects.filter(username__startswith=FAKE_USER_USERNAME_PREFIX).delete()
+        Program.objects.filter(description__startswith=FAKE_PROGRAM_DESC_PREFIX).delete()
```
aliceriot: do you think it might make sense to rename this to reflect that it's also deleting the fake program data? (ditto for creating it).

profiles/management/commands/gen_realistic_users.py
```diff
@@ -0,0 +1,254 @@
...
+def first_matching_item(iterable, predicate):
+    """Returns the first item in an iterable that matches a given predicate"""
+    return next(x for x in iterable if predicate(x))
+
+
+def is_relation_field(model_obj, attr_name):
+    """
+    Checks if a given model attribute is a relation rather than a single value
+    """
+    return 'RelatedManager' in str(type(getattr(model_obj, attr_name)))
```
aliceriot: I'm a little confused about this. If I'm looking at a user and I do:

```python
str(type(getattr(user, 'profile')))
```

I get `"<class 'profiles.models.Profile'>"`. Is this what this is supposed to capture? The Profile object is a different model, but `RelatedManager` doesn't appear in that result.

aliceriot: Oh maybe I get it, a `RelatedManager` is for relationships other than one-to-one?

profiles/management/commands/gen_realistic_users.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @classmethod
+    def _get_edx_course_key(cls, data):   # pylint: disable=unused-argument
+        """Gets the edx_course_key value from the data dictionary"""
+        return ''
+
+    @classmethod
+    def _fill_in_missing_data(cls, data, user, course_run):   # pylint: disable=unused-argument
+        """Fills in data using the associated User and CourseRun objects"""
+        return data
```
aliceriot: are this method and the above method complete? is this functionality we aren't yet adding?

profiles/management/commands/gen_realistic_users.py
gsidebo: this exists for the benefit of `deserialize_model_data`. in that method, i have a model object and some data (key-value pairs). all it's trying to do is set the value of a given attribute if that attribute is a scalar value (int/string/boolean/etc). if that `data` dict has a key that refers to a 'related' field - eg: `work_history` in a Profile object - i don't want to attempt to set it. this method works in the cases that i'm using it. that User.profile case is an interesting one. i'll have a look at that and see if i can make this function more 'correct'

profiles/management/commands/gen_realistic_users.py
gsidebo: these are supposed to be implemented by the child classes. i could do `raise NotImplementedError`, but i think i didn't do it because PyCharm and/or the linter were complaining and i just wanted it/them to shut up

aliceriot: Oh, ok so does the parent class need to implement all methods the child class implements?

profiles/management/commands/gen_realistic_users.py
```diff
@@ -0,0 +1,254 @@
...
+        if cached_model_deserializer.data_key in user_data:
+            accounted_course_runs = [
+                cached_model_deserializer.deserialize(user, data, course_runs)
+                for data in user_data[cached_model_deserializer.data_key]
+            ]
+            unaccounted_course_runs = set(course_runs) - set(accounted_course_runs)
+        else:
+            unaccounted_course_runs = course_runs
+        # For each course run that didn't have associated cached edX data (for Enrollments, Certificates, etc),
+        # create an 'empty' record (data=None)
```
gsidebo: @giocalitri just want to confirm that this is what we're supposed to be doing. before this happens, a couple of fake programs/courses and a few course runs are created. right here, this code is making sure that every user gets an Enrollment and Certificate record for every single course run. if our deserialized data doesn't indicate anything for a particular course run, then the `data` field for the Enrollment or Certificate models are not set (meaning they are set to None via the default)

profiles/management/commands/delete_realistic_users.py
gsidebo: `gen_realistic_search_data` and `delete_realistic_search_data` maybe?

profiles/management/commands/gen_realistic_users.py
aliceriot: Ok cool, gotcha. I did this:

```python
type(getattr(u.profile, 'work_history'))
```

and got a result that makes sense. Maybe for the profile case you could also check for the `'class'` keyword?

gsidebo:
@aliceriot FYI i just rebased on master, so you can check out all of this fake user data in the ES index if you run the recreate_index command: `docker-compose run web ./manage.py recreate_index`

aliceriot:
cool :)

profiles/management/commands/delete_realistic_users.py
aliceriot: yeah that sounds good to me, just something to make clear that if you want program data and so on this command actually takes care of all of it.

profiles/management/commands/gen_realistic_users.py
aliceriot: Or is it because the linter complains about calling a non-existent method on `self` in the method below?

profiles/management/commands/gen_realistic_users.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def deserialize_user_data(user_data, course_runs):
+    """
+    Deserializes a dict of mixed User/Profile data and returns the newly-inserted User
+    """
+    user = User.objects.create(
+        email=user_data['email'],
+        first_name=user_data['first_name'],
+        last_name=user_data['last_name'],
```
aliceriot: can we not use the `deserialize_model_data` method here?

profiles/management/commands/gen_realistic_users.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def deserialize_profile_data(profile, user_data):
+    """
+    Deserializes a dict of mixed User/Profile data and creates a new Profile
+    """
+    for k, v in user_data.items():
+        if hasattr(profile, k) and not is_relation_field(profile, k):
+            setattr(profile, k, v)
+    profile.save()
```
aliceriot: this looks very similar to the `deserialize_model_data` method above, is there a reason we couldn't do:

```python
deserialize_model_data(profile, user_data)
```

instead?

aliceriot: oh actually nvm, looks like this isn't called anywhere, did you intend to delete it?

aliceriot:
some questions, back to you!

profiles/management/commands/gen_realistic_users.py
gsidebo: i think the reason i didn't do it here was b/c `deserialize_model_data` wants a model _object_ rather than a class. at the time, i didn't see a clean way to pass in a model class and figure out all the attributes that could be filled in, and i think User was complaining about required fields when i tried to instantiate an object then fill in attributes. i just took another look in the shell and discovered `<ModelClass>._meta.get_fields()`. might be able to leverage that. i'll take a look

profiles/management/commands/gen_realistic_users.py
gsidebo: correct :)

profiles/management/commands/gen_realistic_users.py
giocalitri: If I understood what you mean, i think this is fine.

profiles/management/commands/gen_realistic_users.py
gsidebo: yes to the last one. the IDE/linter complain about the non-existent method. i just changed these to `raise NotImplementedError` with no issue. not sure what i was seeing before...

profiles/management/commands/gen_realistic_users.py
```diff
@@ -0,0 +1,256 @@
...
+
+def deserialize_model_data(model_cls, data, relational_data=None):
+    """
+    Creates a new instance of a model class and fills in field values using some supplied data
+    """
+    non_relation_field_names = get_non_relation_field_names(model_cls)
+    model_data = filter_dict_by_key_set(data, non_relation_field_names)
+    if relational_data:
+        model_data.update(relational_data)
+    return model_cls.objects.create(**model_data)
```
gsidebo: @aliceriot i split this deserialization into 2 functions. `deserialize_model_data_on_object` is only needed for deserializing profiles. we need this b/c when a User is created, a Profile object is automatically created as well, so i needed a way to deserialize data onto that existing object. changed everything else to use `deserialize_model_data` and pass in a class

gsidebo:
sorry, forgot to rename the files...

micromasters/utils.py
gsidebo: ill look into this too

profiles/management/commands/gen_realistic_search_data.py
```diff
@@ -0,0 +1,256 @@
...
+# Util functions
+
+def first_matching_item(iterable, predicate):
+    """Returns the first item in an iterable that matches a given predicate"""
+    return next(x for x in iterable if predicate(x))
+
+
+def filter_dict_by_key_set(dict_to_filter, key_set):
+    """
+    Takes a dictionary and returns the same dictionary without keys that don't exist in a given set
```
aliceriot: it's not the same dictionary though, right? we're creating a new one, not mutating the one passed in?

micromasters/utils.py
aliceriot: probably not a crucial optimization :)

aliceriot:
LGTM :+1: :shipit:

Issue #733, opened 2016-07-18
Add e-mail opt-in to user settings 


pdpinch:
There is already a field for opt-in in the user profile. We want to expose this to users via the user settings page. 

See https://github.com/mitodl/micromasters/blob/master/profiles/models.py#L93

For form text we can use something like:

```
<h4 class="privacy-form-heading">Who can see your profile?</h4>

<input type="radio">
<label>Yes, opt-in to e-mail</label>
<p>Send me occasional e-mails from MIT and MIT MicroMasters faculty and staff.</p>
<input type="radio">
<label>No, don't send me any e-mail<label>
<p>Don't send me any e-mail.</p>
``` 

We need to look into how to synchronize this setting with mailgun (in a separate issue)

pdpinch:
Updated the description to include some pseudo-code including the form text. Make it look like the rest of the form. Once we have a screenshot, we can run it by Rob.

amir-qayyum-khan:
<img width="1177" alt="screen shot 2016-07-20 at 6 41 10 pm" src="https://cloud.githubusercontent.com/assets/10431250/16988342/959f69fe-4ea9-11e6-8e01-cbefb21a1ef2.png">

roberthouse54:
I am not sure that the copy is right for this. Basically, what we are doing (for V1) is to allow faculty and staff to send emails to users. In V2 we are probably going to allow students to send email to other students (maybe). I think even for V1 we should change the copy slightly. The way it is worded sounds linke you are opting in to spam/marketing emails. I am not sure what the right solution is here, but I think this needs a bit of discussion/clarification.

Ferdi:
@roberthouse54 , I agree we can work with the wording a bit. Any ideas ?

> Basically, what we are doing (for V1) is to allow faculty and staff to send emails to users

And this is a setting for them to opt-out of these e-mails.  OK ?

amir-qayyum-khan:
So you want to add checklist here
- list of email opt-in categories for example
 - Staff to students
 - Student to students

@pdpinch

pdpinch:
Our [model for opt-in right now is a boolean](https://github.com/mitodl/micromasters/blob/master/profiles/models.py#L93) -- you're either in or out. In my draft of the copy, I was trying to make it work for just the V1 use case (faculty to students).

I'd prefer to not update the model at this time (as Amir proposed above) if we don't have to. Is there a way to finesse the copy to make that work?

roberthouse54:
[...]

PR #740, opened 2016-07-20
Added e-mail preference settings

amir-qayyum-khan:
closes https://github.com/mitodl/micromasters/issues/733

#### What's this PR do?
- Add email preference settings to profile settings form.

- open settings page 
- open profile/settings

@pdpinch 
screen shot 2016-07-22 at 4 06 41 pm

### DIFF
--- a/static/js/components/PrivacyForm.js
+++ b/static/js/components/PrivacyForm.js
@@ -15,10 +15,16 @@ class PrivacyForm extends ProfileFormFields {
   };
 
   render() {
+    const emailOptions = [
+      { value: "true", label: "Faculty and staff can send me emails"},
+      { value: "false", label: "I don't want to receive any emails" }
+    ];
     return (
       <div>
         <h4 className="privacy-form-heading">Who can see your profile?</h4>
         { this.boundRadioGroupField(['account_privacy'], '', this.privacyOptions) }
+        <h4 className="privacy-form-heading">Email Preferences</h4>
+        { this.boundRadioGroupField(['email_optin'], '', emailOptions) }
       </div>
     );
   }
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -35,7 +35,13 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
 
   let onChange = e => {
     let clone = _.cloneDeep(profile);
-    _.set(clone, keySet, e.target.value);
+    let value = e.target.value;
+    if (value === "true") {
+      value = true;
+    } else if (value === "false") {
+      value = false;
+    }
+    _.set(clone, keySet, value);
     updateProfile(clone, validator);
   };
 
@@ -62,7 +68,7 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
     );
   });
 
-  const value = _.get(profile, keySet);
+  const value = String(_.get(profile, keySet));
   return (
     <div>
       <span className="profile-radio-group-label">
@@ -282,8 +288,8 @@ export function saveProfileStep(validator: Validator|UIValidator, isLastStep: bo
   let clone = Object.assign({}, profile, {
     filled_out: profile.filled_out || isLastStep
   });
-  if (isLastStep) {
-    // user has also seen email consent message at this point
+
+  if (isLastStep && !profile.filled_out) {
     clone.email_optin = true;
   }
   return saveProfile(validator, clone, ui);
--- a/static/js/containers/ProfilePage_test.js
+++ b/static/js/containers/ProfilePage_test.js
@@ -337,7 +337,8 @@ describe("ProfilePage", function() {
     return renderComponent('/profile').then(([, div]) => {
       // close all switches and remove all education so we don't get validation errors
       let receivedProfile = Object.assign({}, USER_PROFILE_RESPONSE, {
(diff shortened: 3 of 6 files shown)


aliceriot:
@amir-qayyum-khan is this ready for review? you've got both 'needs review' and 'work in progress' tags on it.

amir-qayyum-khan:
@aliceriot this is under development and discussion https://github.com/mitodl/micromasters/issues/733#issuecomment-233957803

aliceriot:
cool :)

amir-qayyum-khan:
@aliceriot I am done with working on this. If you want you can review it. 

Thanks,

aliceriot:
@amir-qayyum-khan sure!

static/js/components/PrivacyForm.js
```diff
@@ -15,10 +15,16 @@ class PrivacyForm extends ProfileFormFields {
   };
 
   render() {
+    const emailOptions = [
+      { value: "true", label: `Faculty and staff can send me emails about the program`},
+      { value: "false", label: `I don't want to receive any emails` }
```
aliceriot: is there a reason you're declaring the labels as template literals? I think strings are preferred unless you need to do interpolation.

static/js/util/profile_edit.js
```diff
@@ -35,7 +35,14 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
 
   let onChange = e => {
     let clone = _.cloneDeep(profile);
-    _.set(clone, keySet, e.target.value);
+    let value = e.target.value;
+    // string to boolean conversion.
```
aliceriot: this comment isn't necessary, this code is very straightforward and understandable on it's own

static/js/util/profile_edit.js
```diff
@@ -62,7 +69,7 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
     );
   });
 
-  const value = _.get(profile, keySet);
+  const value = String(_.get(profile, keySet));
```
aliceriot: I'm curious about this overall, is there a reason that you can't use a boolean as the value of the options we provide to the radio button select? why do we need to convert it to a string?

static/js/util/profile_edit.js
```diff
@@ -282,9 +289,5 @@ export function saveProfileStep(validator: Validator|UIValidator, isLastStep: bo
   let clone = Object.assign({}, profile, {
     filled_out: profile.filled_out || isLastStep
   });
-  if (isLastStep) {
-    // user has also seen email consent message at this point
-    clone.email_optin = true;
-  }
```
aliceriot: I'm not sure about this deletion - we probably do want to set 'email_option' to true unless a user opts out, but we only want to do so if we're saving and the user hasn't yet completed the profile onboarding process.

maybe we want something like

```
if (isLastStep && !profile.filled_out) {
  ...
}
```

so that we only set the default value when the user is saving for the first time (i.e. when `filled_out` hasn't yet be set to `true`).

aliceriot:
some questions! back to you

static/js/components/PrivacyForm.js
amir-qayyum-khan: removed

static/js/util/profile_edit.js
amir-qayyum-khan: it is giving me 3 warning messages if I use boolean. here

- https://github.com/mitodl/micromasters/blob/a235e8d3292c708f88d82637e9bc38f14fa5dc9f/static/js/util/profile_edit.js#L64
- https://github.com/mitodl/micromasters/blob/a235e8d3292c708f88d82637e9bc38f14fa5dc9f/static/js/util/profile_edit.js#L66
- https://github.com/mitodl/micromasters/blob/a235e8d3292c708f88d82637e9bc38f14fa5dc9f/static/js/util/profile_edit.js#L82

static/js/util/profile_edit.js
amir-qayyum-khan: hmm, sounds ok, let me add this check.

amir-qayyum-khan: @aliceriot only issue with this logic is if profile is not filled_out then if user selects ``email_optin = false`` on the ``isLastStep == true`` last step the this info with not save, because we are making it ``clone.email_optin = true;`` always

static/js/util/profile_edit.js
aliceriot: huh, I guess that doesn't work! this looks like generally a good solution then.

static/js/util/profile_edit.js
```diff
@@ -35,7 +35,14 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
 
   let onChange = e => {
     let clone = _.cloneDeep(profile);
-    _.set(clone, keySet, e.target.value);
+    let value = e.target.value;
+
```
aliceriot: I'd get rid of this blank line

aliceriot:
This LGTM! One comment about whitespace, other than that it's good to go

:+1: :100: :fireworks: :confetti_ball:

static/js/util/profile_edit.js
amir-qayyum-khan: Done, I am going to merge it once test get pass.

PR #742, opened 2016-07-20
Removed box-shadow from Navbar

### LINKED ISSUE #741: remove navbar drop shadow
I oopsed a bit and we don't want the drop shadow

aliceriot:

closes #741 

#### What's this PR do?

just sets `box-shadow: none !important` on the navbar.

Make sure the navbar doesn't have a shadow. I couldn't get the styling to apply without `!important`, if you know of a sneaky way to avoid using the important keyword that would be :100: (I think some of the difficulty is because some styles are applied via JS by material design lite).

### DIFF
--- a/static/scss/navbar.scss
+++ b/static/scss/navbar.scss
@@ -11,6 +11,8 @@
 }
 
 .micromasters-nav {
+  box-shadow: none !important;
+
   .micromasters-title {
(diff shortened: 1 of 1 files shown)


gsidebo:
i'll allow it 😄  👍

Issue #734, opened 2016-07-18
dashboard/views_test contact use edx backend instead of mocking


giocalitri:
If edx is down (or cannot be contacted) 3 tests fail.
`docker-compose run web tox -- dashboard/views_test.py`

EDX should be completely mocked.

PR #745, opened 2016-07-25
Mocked certificate api in test suit, So that i can run even if edX instance is shutdown

amir-qayyum-khan:
closes https://github.com/mitodl/micromasters/issues/734

#### What's this PR do?
- Alter test suit ``dashboard/views_test.py`` and mocked certificate api so that tests can run event of edX instance is not working. Previously one test gets fail if edX platform instance is close.

- see ``dashboard/views_test.py``

- Close edX platform (shutdown server or vagrant box)
- Run test ``docker-compose run web tox -- dashboard/views_test.py``

@pdpinch @giocalitri

### DIFF
--- a/dashboard/views_test.py
+++ b/dashboard/views_test.py
@@ -14,7 +14,10 @@ from django.core.urlresolvers import reverse
 from requests.exceptions import HTTPError
 from rest_framework import status
 from rest_framework.test import APITestCase
-from edx_api.certificates.models import Certificates
+from edx_api.certificates.models import (
+    Certificate,
+    Certificates
+)
 from edx_api.enrollments.models import Enrollments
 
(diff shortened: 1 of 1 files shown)


dashboard/views_test.py
```diff
@@ -99,9 +108,8 @@ def create_run(self, course, start=None, end=None,  # pylint: disable=no-self-us
 
     def get_with_mocked_enrollment(self):
```
giocalitri: please rename this to `get_with_mocked_enrollment_certificates`

giocalitri:
a very small comment, then 👍

dashboard/views_test.py
amir-qayyum-khan: Done @giocalitri , I will merge it once test get pass

PR #747, opened 2016-07-26
Added ProgramEnrollments to dashboard

### LINKED ISSUE #672: Add MicroMasters program enrollment to user model

annagav:
Fix #672 
#### What's this PR do?
Create ProgramEnrollment model. Add signal to enroll student in a program (create ProgramEnrollment) when the student enrolls in a course (CachedEnrollment is updated with valid data).


dashboard/apps.py
```diff
@@ -0,0 +1,17 @@
+"""
+AppConfig
```
giocalitri: better to specify for which app is this config

dashboard/apps.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+from django.apps import AppConfig
+
+
+class ProfilesConfig(AppConfig):
```
giocalitri: you may want to rename it

dashboard/signals.py
```diff
@@ -0,0 +1,15 @@
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from dashboard.models import CachedEnrollment,ProgramEnrollment
+
+
+@receiver(post_save, sender=CachedEnrollment)
+def create_program_enrollment(sender, instance, created, **kwargs):
+
```
giocalitri: missing docstring

dashboard/signals.py
```diff
@@ -0,0 +1,15 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from dashboard.models import CachedEnrollment,ProgramEnrollment
+
+
+@receiver(post_save, sender=CachedEnrollment)
```
giocalitri: please add the the argument `dispatch_uid`

dashboard/signals_test.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Test class for signals that creates a profile whenever a user is created
+    """
+    def test_program_enrollment_creation(self):
+        """
+        Tests that if a user is created, the profile for such user is created
+        """
+        # there are no users
+        assert len(ProgramEnrollment.objects.all()) == 0
+        assert len(CachedEnrollment.objects.all()) == 0
```
giocalitri: better to use `CachedEnrollment.objects.all().count()`

gsidebo: even better: `CachedEnrollment.objects.count()`

dashboard/signals_test.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # the creation of a user triggers the creation of a profile
+        user = User.objects.create(username='test', password='test')
+        program = Program.objects.create(title='New Program', live=True)
+        course = Course.objects.create(program=program, title="Test Course")
+        course_run = CourseRun.objects.create(course=course, title="Test Run")
+        enrollment = CachedEnrollment.objects.create(user=user, course_run=course_run, data=None)
+        assert len(ProgramEnrollment.objects.all()) == 0
+        enrollment.data = {}
+        enrollment.save()
```
gsidebo: does this work? line 11 in `dashboard/signals.py` (`if instance.data:`) makes it look like this shouldn't work

```python
# Doesn't print anything
if {}:
     print('hi')

# Prints 'hi'
if {'a':1}:
     print('hi')
```

dashboard/signals.py
```diff
@@ -0,0 +1,19 @@
...
+@receiver(post_save, sender=CachedEnrollment, dispatch_uid="update_programenrollment")
+def create_programenrollment(sender, instance, created, **kwargs):  # pylint: disable=unused-argument
+    """
+    Signal handler to create Program enrollment when the CachedEnrollment table is updated
+    """
+    if instance.data is not None:
+        ProgramEnrollment.objects.get_or_create(
+            user=instance.user,
+            program=instance.course_run.course.program
+        )
```
gsidebo: if a student enrolls in a single course in a program then withdraws, will we want to keep the ProgramEnrollment record?

gsidebo:
i just have the one comment. everything else looks good

dashboard/signals.py
giocalitri: you are right, we need to make the logic better

dashboard/signals.py
```diff
@@ -0,0 +1,30 @@
...
+        # check if there is any other entry with data != None
+        # if not delete the enrollment
+        if CachedEnrollment.objects.filter(
+            user=instance.user,
+            course_run__course__program=instance.course_run.course.program
+        ).exclude(data__isnull=True).count() == 0:
+            ProgramEnrollment.objects.filter(
+                user=instance.user,
+                program=instance.course_run.course.program
+            ).delete()
```
gsidebo: a few things

1. this else statement seems problematic. we add a _lot_ of CachedEnrollment records with empty data, right? as I understand it, it's (total # of users) x (total # of course runs) - (actual user enrollments). this else clause would be triggered on the insertion of _every one_ of those, and the 'if' statement would evaluate to true for almost every one. my understanding is that we only want to delete the ProgramEnrollment if the CachedEnrollment had some data previously. could you use `if kwargs.get('created', False):` or something like that to make sure that this is an update rather than an insertion? 
1. i would prefer the `CachedEnrollment.objects.filter...` query to be stored in a local variable (`active_enrollment_count` or something) for readability
1. since `instance.user` and `instance.course_run.course.program` are used multiple times in this function, and in both branches of logic, i think it makes sense to store them in local variables

dashboard/signals_test.py
```diff
@@ -0,0 +1,92 @@
...
+            course_run=self.course_run,
+            data={'data': 'data'},
+            last_request=self.now
+        )
+        assert ProgramEnrollment.objects.count() == 1
+        enrollment.data = None
+        enrollment.save()
+        assert ProgramEnrollment.objects.count() == 0
+
+    def test_program_enrollment_No_deletion(self):
```
gsidebo: all lowercase please

dashboard/signals.py
giocalitri: about 2 and 3: ok

about 1: I understand your point but I am not convinced.
it is true that if I check for `created` I will limit the queries at insertion time, but the big amount of queries will be at "update" time, because it will be more likely that the entry for `CachedEnrollment` will be already in the DB (consider that there will be at minimum an update of the timestamp every 6 hours). So I do not see a significant increase of performance here.
Let me know if I am missing something

dashboard/signals_test.py
giocalitri: sure

dashboard/signals.py
```diff
@@ -0,0 +1,33 @@
...
+        # if not delete the enrollment
+        active_enrollment_count = CachedEnrollment.objects.filter(
+            user=user,
+            course_run__course__program=program
+        ).exclude(data__isnull=True).count()
+        if active_enrollment_count == 0:
+            ProgramEnrollment.objects.filter(
+                user=user,
+                program=program
+            ).delete()
```
gsidebo: what about doing this `else:` logic in a `pre_save` signal instead? something like:

```python
# might need to use something other than 'instance.pk' to let us know this is an update...
if not instance.data and instance.id is not None:
  instance_in_db = CachedEnrollment.objects.filter(id=instance.id).first()
  if instance_in_db.data:
    active_enrollment_count = CachedEnrollment.objects.filter(...)
    if active_enrollment_count == 0:
      ProgramEnrollment.objects.filter(
          user=user,
          program=program
      ).delete()
```

you can do the delete safely before the save since ProgramEnrollment has no formal relationship with the CachedEnrollment. this adds some extra code, and it still executes a query most times against `CachedEnrollment` but that query is only on the primary key so it's cheap. the main benefit is that this avoids excessive querying that would happen in the vast majority of cases

giocalitri: I think this is a good approach: I like the idea. 
I thought about it and I implemented an even more (I hope) optimized version

giocalitri:
@gsidebo I think this should be fine now

gsidebo:
looks great! 👍

PR #749, opened 2016-07-26
pass the username to clearProfileEdit

### LINKED ISSUE #748: Profile errors not being cleared successfully
This is just a small bug with the `clearProfileEdit` function, we should be passing in a username but we kinda aren't.

aliceriot:

closes #748 

#### What's this PR do?

We need to pass a username to the `clearProfileEdit` action helper because redux has to know the username for which it should clear the `edit` object. We weren't doing so, which meant that validation errors that should have been removed by `clearProfileEdit` were hanging around and doing some strange things.

Read over the code!

Currently, on master, if you do the following:

1. go to `/users`
2. create a validation error on the personal info form
3. click 'save'
4. click 'cancel'

you should see the `clearProfileEdit` action being dispatched, but if you inspect the redux state you'll note that the `edit` object is actually still there (`profiles -> ${username} -> edit`). The same holds for editing a work history or education history entry.

Basically, these changes should ensure that the correct `edit` object is actually cleared.

### DIFF
--- a/static/js/components/EducationDialog.js
+++ b/static/js/components/EducationDialog.js
@@ -44,11 +44,12 @@ export default class EducationDialog extends ProfileFormFields {
       setEducationDegreeLevel,
       setEducationDialogIndex,
       clearProfileEdit,
+      profile: { username },
     } = this.props;
     setEducationDialogVisibility(false);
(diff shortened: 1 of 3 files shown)


gsidebo:
👍

PR #750, opened 2016-07-26
implemented ValidationAlert to alert user to problems in form dialogs

### LINKED ISSUE #635: Profile validation errors sometimes have no visual indicator
To reproduce:

1. Open up the /profile/personal tab
1. Fill out the entire form, and enter something invalid for DOB (eg: "00/00/0000")
1. Scroll down to the bottom of the page and, if necessary, shrink the window so you can't see the DOB field
1. Click "Save and Continue"

Expected: some visual indication that the form validation failed
Actual: visually, nothing happens

A couple options to consider:

- when validation fails, change the form focus to the first field that needs to be fixed
- when validation fails, put some generic error message above/aside/below the "Save and Continue" button, eg: "Please fix the errors above"

aliceriot:

this closes #635 

#### What's this PR do?

This adds a visual indicator to the profile dialogs which lives right above the 'cancel' and 'save' buttons. If there's a validation error it will look like this:

error_alert

Otherwise it looks as it currently does on master.

Read the changes!

Go to `/users` and do the following:

1. open the personal info dialog
2. enter something invalid (like deleting your last name)
3. click 'save'
4. confirm the message shows up
5. fix the error (add a last name)
6. confirm the error message goes away

then do the same thing for the education and work history dialogs as well.

### DIFF
--- a/static/js/components/ValidationAlert.js
+++ b/static/js/components/ValidationAlert.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import _ from 'lodash';
+
+import type { ValidationErrors } from '../flow/profileTypes';
+
+export default class ValidationAlert extends React.Component {
+  props: {
+    errors:   ValidationErrors,
+    children: React$Element[],
+  };
+
+  static message = "Please fix the errors above and try again";
+
+  alertMessage: Function = (): React$Element|void => {
+    const { errors } = this.props;
+    if ( !_.isEmpty(errors) ) {
+      return (
+        <span className="message">
+          { this.constructor.message }
+        </span>
+      );
+    }
+  };
+
+  render () {
+    const { children } = this.props;
+    return (
+      <div className="validation-alert">
+        { this.alertMessage() }
+        <div className="actions">
+          { children }
+        </div>
+      </div>
+    );
+  }
+}
--- a/static/js/containers/UserPage_test.js
+++ b/static/js/containers/UserPage_test.js
@@ -33,6 +33,7 @@ import IntegrationTestHelper from '../util/integration_test_helper';
 import * as api from '../util/api';
 import { USER_PROFILE_RESPONSE, HIGH_SCHOOL, DOCTORATE } from '../constants';
 import { workEntriesByDate, educationEntriesByDate } from '../util/sorting';
+import ValidationAlert from '../components/ValidationAlert';
 
 describe("UserPage", function() {
   this.timeout(5000);
@@ -135,6 +136,9 @@ describe("UserPage", function() {
             TestUtils.Simulate.click(save);
             let state = helper.store.getState();
             assert.deepEqual(state.profiles.jane.edit.errors, validationExpectation);
+            let validationInfoText = document.querySelector('.validation-alert').
+              querySelector('.message').textContent;
+            assert.equal(validationInfoText, ValidationAlert.message);
             modifyTextField(input, removeErrorValue);
           }).then(() => {
             let state = helper.store.getState();
--- a/static/js/components/EducationDialog.js
+++ b/static/js/components/EducationDialog.js
@@ -12,6 +12,8 @@ import SelectField from './inputs/SelectField';
 import CountrySelectField from './inputs/CountrySelectField';
 import StateSelectField from './inputs/StateSelectField';
 import FieldsOfStudySelectField from './inputs/FieldsOfStudySelectField';
+import ValidationAlert from './ValidationAlert';
+
 import type { UIState } from '../reducers/ui';
 import type { Profile, SaveProfileFunc } from '../flow/profileTypes';
 import type { Validator, UIValidator } from '../util/validation';
@@ -130,22 +132,20 @@ export default class EducationDialog extends ProfileFormFields {
   render () {
     const { ui: {educationDialogVisibility } } = this.props;
 
-    let actions = [
+    let actions = <ValidationAlert {...this.props}>
       <Button
         type='button'
-        key='cancel'
         className="cancel-button"
         onClick={this.clearEducationEdit}>
         Cancel
-      </Button>,
+      </Button>
       <Button
-        key='save'
         type='button'
         className="save-button"
         onClick={this.saveEducationForm}>
         Save
-      </Button>,
-    ];
+      </Button>
+    </ValidationAlert>;
 
     return (
       <Dialog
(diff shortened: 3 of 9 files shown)


static/js/components/ValidationAlert.js
```diff
@@ -0,0 +1,34 @@
...
+
+  render () {
+    const { children } = this.props;
+    return (
+      <div className="validation-alert">
+        { this.alertMessage() }
+        <div className="actions">
+          { children }
+        </div>
+      </div>
```
gsidebo: hmmm. i guess this is material-ui forcing us to wrap the buttons in this new component since you can only render content in the bottom bar via the `actions` property. i think it makes sense to create our own component (`ValidatingDialog` or something) that encapsulates these changes. we could leave `actions` alone in all of these files, and in `EducationDialog.js` (just as an example) the API would change to something like this.

```javascript

# In the render method...
const { errors } = this.props;
<ValidatingDialog
           open={workDialogVisibility}
           className="dashboard-dialog employment-dashboard-dialog"
           onRequestClose={this.closeWorkDialog}
           actions={actions}
           autoScrollBodyContent={true}
           errors={errors}
         >
```

`ValidatingDialog` would wrap the actions with `<div className="validation-alert">` and show the message only when validation errors exist. it would then just pass everything along to `Dialog`

aliceriot: I don't really see why that would be more simple than what we have here. I chose to build this as a component that wraps it's children because that was the easiest way to get things to display in the layout we wanted, which is this:

```
------------
|  message |
------------
--------- ---------
| box 1 | | box 2 |
--------- ---------
```

the `actions` prop accepts either a single element or an array of elements, so if we wanted the message to appear in front of the buttons we could have done that too. I'd prefer to keep what I have here instead of building a utility class like `ValidatedDialog`.

static/js/containers/UserPage_test.js
```diff
@@ -135,6 +135,9 @@ describe("UserPage", function() {
             TestUtils.Simulate.click(save);
             let state = helper.store.getState();
             assert.deepEqual(state.profiles.jane.edit.errors, validationExpectation);
+            let validationInfoText = document.querySelector('.validation-alert').
+              querySelector('.message').textContent;
+            assert.deepEqual(validationInfoText, "Please fix the errors above and try again");
```
gsidebo: nitpick: can you make this error message a property of the actual component and assert the text against that? also, this is just a string compare so i don't think you need deepEqual

aliceriot: Not totally sure what you mean, do you mean make the error message a property of the `ValidationAlert` class?

static/js/components/ValidationAlert.js
gsidebo: i wasn't very clear there. `ValidatedDialog` as i'm describing would be a component, not a utility class, and it would replace `ValidationAlert` (while keeping some of its functionality). it would look something like this (forgive the pseudo-code):

```javascript
class ValidatedDialog extends React.Component {
	ERROR_MESSAGE: "error"
	
	dialogActions() {
		const { actions, errors } = this.props
		if errors
			return <div className='validation-error'>{this.ERROR_MESSAGE}<div className='actions'>{actions}</div></div>
		else
			return actions
	}
	
	dialogProps() {
		return dialogActions() + (this.props without 'actions')
	}
	
	render() {
		return <Dialog {...this.dialogProps()} />
	}
}
```

benefits:

1. no unnecessary `<div className="validation-alert">` wrapper element where it's not necessary
1. wrapping the `actions` property is a hack that we need to get around a library limitation. better to keep that in one place
1. simple code reuse/easier API. you need to add a component to wrap the `actions` property in 3 different places in this PR. my suggestion replaces a component instead of adding a new component

static/js/containers/UserPage_test.js
gsidebo: yup. if you're on board with my suggestion to create a ValidatedDialog component, it would go there

static/js/components/ValidationAlert.js
aliceriot: OK cool, that gives me a better idea of what you're suggesting.

Something occurs to me though, which is that we might want to use the validation alert in places other than in dialogs, in particular on the personal info page. If someone is on a device without much vertical room (say, a smartphone in landscape mode or a netbook) they might not see the validation errors up top when they click save, the `ValidationAlert` component as written can be just dropped into the personal tab component.

Thoughts?

Here's an example of what it looks like:

validation

aliceriot: @roberthouse54 do you think we should have this error message on the personal form on the profile as well?

roberthouse54: Yeah, I think we should do the same thing on all form pages.

aliceriot: @gsidebo that pushes me towards leaving the `ValidationAlert` class as-is, because we can add it to those other non-dialog forms with one line. What do you think?

gsidebo: (currently pondering...)

aliceriot: A key thing is that the `ValidationAlert` component (as written) can also render just find without children. So what I'm doing to get it to render on the `PersonalTab` and `PrivacyTab` is just (more or less) doing this:

```js
<ValidationAlert {...this.props} />
```

so it's not wrapping anything at all. I actually think that this approach (of a component that optionally takes children) is very idiomatic for React.

Also the reason that I'm wrapping the buttons to pass in to the `actions` prop for `Dialog` with the `ValidationAlert` component is because it makes it easier to get the layout to look the way I want, not so much because of a limitation of the dialog component. The dialog component is kind flexible in this regard, actually, since it can take a single component or an array of components.

gsidebo: 3 main problems i still have

1. i don't think it makes semantic sense for us to pass `Button`s or any other component as children to a component called `ValidationAlert`. _semantically_ i don't know why i would be wrapping `Button`s in that component
1. this still outputs an unnecessary `<div className="validation-alert">` wrapper element when there are no errors in a Dialog
1. not terribly important, but it's easier to reason about the error message component and the general component tree if it's not invoking `{children}`

**with that said**, i'm getting the sense that we're at an impasse, and this isn't going to introduce some kind of show-stopping tech debt. you're good with your approach if i haven't convinced you 😄

aliceriot: Haha, ok. I'm going to incorporate your suggestions below, I'll ping when it's ready for review again.

aliceriot:
@gsidebo forgot I had already done that on friday >_<

anyway, ready for review

gsidebo:
👍

PR #755, opened 2016-07-28
refactor `actions/index.js`

### LINKED ISSUE #754: reorganize `static/js/actions/index.js`
In `actions/ui.js` we've been using the following style for exporting an action type and it's corresponding action creator:

```js
export const MY_ACTION_TYPE = 'MY_ACTION_TYPE';
export const myActionType = (arg: type): Action => (
  { type: MY_ACTION_TYPE, payload: arg }
);
```

So the declaration of the action type and it's corresponding action creator are right next to each other. I think this is nice and readable. In `actions/index.js` currently we declare all of the action types at the top in a big block and then all the action creators down below. I'd like to reorganize `actions/index.js` along the lines that the `ui` actions are currently organized.

While I'm at it I'm going to add type annotations.

aliceriot:

this closes #754 

#### What's this PR do?

Small refactor of `static/js/actions/index.js`. Basically just moving some code around and adding type annotations.

Read the code!

Functionality should be totally the same, so probably click around and make sure all the profile and dashboard related functionality is working correctly.

### DIFF
--- a/static/js/actions/index.js
+++ b/static/js/actions/index.js
@@ -1,17 +1,9 @@
-import * as api from '../util/api';
+// @flow
+import type { Dispatch } from 'redux';
 
-// user profile actions
-export const REQUEST_GET_USER_PROFILE = 'REQUEST_GET_USER_PROFILE';
-export const RECEIVE_GET_USER_PROFILE_SUCCESS = 'RECEIVE_GET_USER_PROFILE_SUCCESS';
-export const RECEIVE_GET_USER_PROFILE_FAILURE = 'RECEIVE_GET_USER_PROFILE_FAILURE';
-export const CLEAR_PROFILE = 'CLEAR_PROFILE';
-export const UPDATE_PROFILE = 'UPDATE_PROFILE';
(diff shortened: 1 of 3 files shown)


static/js/actions/index.js
```diff
@@ -79,12 +82,14 @@ export const saveProfile = (username, profile) => {
...
+
+export const UPDATE_PROFILE_VALIDATION = 'UPDATE_PROFILE_VALIDATION';
+export const updateProfileValidation = (username: string, errors: ValidationErrors): Action => ({
   type: UPDATE_PROFILE_VALIDATION,
   payload: { errors, username }
 });
 
-export function fetchUserProfile(username) {
+export function fetchUserProfile(username: string): Dispatcher {
   return dispatch => {
```
gsidebo: does it make sense to add the same type checking here/`saveProfile`/etc as you did in `fetchDashboard` below? (referring to `return (dispatch: Dispatch) => {`)

aliceriot: yes! good catch

aliceriot:
@gsidebo I added that annotation (for `dispatch`)

gsidebo:
looks great! 👍

PR #777, opened 2016-08-01
added 'jump to error'

### LINKED ISSUE #561: Dialog forms - when window is too small, some fields are invisible w/ no indication that they exist
ss 2016-06-16 at 12 10 07 pm

If a dialog window is too small vertically, required fields can be obscured with no visual indication that they exist. The scrollbar to access these fields in invisible by default. As a result, a user might be informed about form errors on fields they cannot see. Clicking 'Save' produces no apparent visual change whatsoever in these cases

Opinion: This one should be considered very high priority

aliceriot:

closes #561 

#### What's this PR do?

This adds a little bit of code that basically does two things.

1. Sets a class (`.invalid-input`) on a profile input field that currently has a validation error.
2. Whenever we save the profile we find the first such input (`document.querySelector` to the rescue) and call `.scrollIntoView` on it.

On a bigger screen this has no visual effect. However, on a shorter screen (e.g. one where the content in the dialog box might overflow and have an invisible scrollbar) it will cause that div to scroll to the location of the invalid field, prompting the user for their input.

1. go to `/users`
2. open dev tools or otherwise make your viewport really short
3. open the 'personal' dialog
4. delete your last name
5. scroll down to the bottom of the dialog content
6. click 'save'

when you click save the dialog content should scroll up to show you the 'last name' field

The same basic thing should happen on work_history and education as well, and also on the personal step of the profile.

### DIFF
--- a/static/js/containers/ProfileFormContainer.js
+++ b/static/js/containers/ProfileFormContainer.js
@@ -98,6 +98,15 @@ class ProfileFormContainer extends React.Component {
         dispatch(clearProfileEdit(username));
       });
     } else {
+      // `setState` is being called here because we want to guarantee that
+      // the callback executes after the `dispatch` call above. A callback
+      // passed to `setState` executes when the component next re-renders.
+      this.setState({}, () => {
+        let invalidField = document.querySelector('.invalid-input');
+        if ( invalidField !== null ) {
+          invalidField.scrollIntoView();
+        }
+      });
       return Promise.reject(errors);
     }
   }
--- a/static/js/containers/UserPage_test.js
+++ b/static/js/containers/UserPage_test.js
@@ -4,6 +4,7 @@ import TestUtils from 'react-addons-test-utils';
 import { assert } from 'chai';
 import _ from 'lodash';
 import moment from 'moment';
+import sinon from 'sinon';
 
 import { 
   RECEIVE_GET_USER_PROFILE_SUCCESS,
@@ -13,6 +14,9 @@ import {
   UPDATE_PROFILE_VALIDATION,
   REQUEST_PATCH_USER_PROFILE,
   RECEIVE_PATCH_USER_PROFILE_SUCCESS,
+
+  startProfileEdit,
+  updateProfile,
 } from '../actions';
 import {
   SET_WORK_DIALOG_VISIBILITY,
@@ -120,32 +124,13 @@ describe("UserPage", function() {
 
     describe("validation", () => {
       const inputs = dialog => [...dialog.getElementsByTagName('input')];
-      const confirmClearValidation = (actions, getInput, validationExpectation, removeErrorValue) => {
-        return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
-          let editPersonalButton = div.
-            querySelector('.main-content').
-            querySelector('.material-icons');
+      const getEditPersonalButton = div => div.querySelector('.main-content .material-icons');
+      const getDialog = () => document.querySelector('.personal-dialog');
+      const getSave = () => getDialog().querySelector('.save-button');
 
-          return listenForActions(actions, () => {
-            TestUtils.Simulate.click(editPersonalButton);
-
-            let dialog = document.querySelector('.personal-dialog');
-            let save = dialog.querySelector('.save-button');
-            let input = getInput(dialog);
-            modifyTextField(input, "");
-            TestUtils.Simulate.click(save);
-            let state = helper.store.getState();
-            assert.deepEqual(state.profiles.jane.edit.errors, validationExpectation);
-            let validationInfoText = document.querySelector('.validation-alert').
-              querySelector('.message').textContent;
-            assert.equal(validationInfoText, ValidationAlert.message);
-            modifyTextField(input, removeErrorValue);
-          }).then(() => {
-            let state = helper.store.getState();
-            assert.deepEqual(state.profiles.jane.edit.errors, {});
-          });
-        });
-      };
+      beforeEach(() => {
+        HTMLDivElement.prototype.scrollIntoView = sinon.stub();
+      });
 
       let userProfileActions = [
         SET_USER_PAGE_DIALOG_VISIBILITY,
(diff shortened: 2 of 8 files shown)


noisecapella:
Functionality looks good

static/js/util/util_test.js
```diff
@@ -259,4 +260,27 @@ describe('utility functions', () => {
       ]);
     });
   });
+
+  describe('validationErrorSelector', () => {
+    const invalid = "invalid-input";
+    let errors, keySet;
```
noisecapella: Can we move the `let` into each of the tests below, unless there's some data to share between the tests?

aliceriot: sure, probably best not to leave the door open for leakage between tests

static/js/containers/ProfileFormContainer.js
```diff
@@ -98,7 +98,13 @@ class ProfileFormContainer extends React.Component {
         dispatch(clearProfileEdit(username));
       });
     } else {
-      return Promise.reject(errors);
+      return Promise.reject(errors).catch(() => {
```
noisecapella: Can we replace this `else` block with everything inside the `catch` here? This line seems redundant

static/js/containers/UserPage_test.js
```diff
@@ -120,20 +121,19 @@ describe("UserPage", function() {
 
     describe("validation", () => {
       const inputs = dialog => [...dialog.getElementsByTagName('input')];
-      const confirmClearValidation = (actions, getInput, validationExpectation, removeErrorValue) => {
-        return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
-          let editPersonalButton = div.
-            querySelector('.main-content').
-            querySelector('.material-icons');
+      const getEditPersonalButton = div => (
+        div.querySelector('.main-content').querySelector('.material-icons')
```
noisecapella: Can you combine the two `querySelector` here?

static/js/containers/ProfileFormContainer.js
aliceriot: I tried that but I couldn't get it to work, I think there's a race condition between dispatching the validation errors above and react re-rendering the page, so if we look for the `.invalid-input` selector on the page it hasn't shown up yet. Doing this (putting the `scrollIntoView` code in the `.catch` block) is the only way I found to ensure that the code to find the selector executed after the DOM had actually been updated.

Do you have any ideas for how to organize it better?

static/js/containers/UserPage_test.js
aliceriot: ...yes, haha

static/js/containers/ProfileFormContainer.js
noisecapella: Hmm... one hacky way to solve this is to use `setState`: `setState({}, () => { // dom manipulation here })`. According to [the docs](https://facebook.github.io/react/docs/component-api.html) the `setState` callback triggers after component rerender. Right now there's no guarantee the `catch` callback will execute after rerender, but if the code is moved inside the `setState` callback we should have that guarantee.

Ideally we should do all DOM manipulation inside `componentDidMount` or `componentDidUpdate` or other relevant lifecycle methods, but doing it this way means we don't need to keep track of the state of whether we need to call `scrollIntoView`

noisecapella: Also, can you replace `document` with `ReactDOM.findDOMNode(this)`?

aliceriot: ah, `setState` works! that's a nice trick!

static/js/containers/UserPage_test.js
```diff
@@ -156,37 +177,39 @@ describe("UserPage", function() {
         UPDATE_PROFILE_VALIDATION,
       ];
```
noisecapella: Can we add a test for `boundRadioGroupField`?

aliceriot: yeah, that's probably a good idea

aliceriot: I'm a little unsure of how to go about that though, is it even possible for the user to enter invalid input in a radio field? I think we always supply a default selection.

aliceriot: wait nvm

aliceriot:
@noisecapella finally fixed all the tests!

static/js/containers/ProfileFormContainer.js
```diff
@@ -98,6 +98,12 @@ class ProfileFormContainer extends React.Component {
         dispatch(clearProfileEdit(username));
       });
     } else {
+      this.setState({}, () => {
```
noisecapella: Can you add a comment here to explain what's going on and that we want to make sure `scrollIntoView` is executed after rerender?

static/js/containers/UserPage_test.js
```diff
@@ -120,76 +124,155 @@ describe("UserPage", function() {
...
+      const scrollIntoView = (actions, getInput) => {
+        return renderComponent(`/users/${SETTINGS.username}`, userActions).then(([, div]) => {
+          return listenForActions(scrollActions, () => {
+            TestUtils.Simulate.click(getEditPersonalButton(div));
+            let input = getInput(getDialog());
+            modifyTextField(input, "");
+            TestUtils.Simulate.click(getSave());
+          }).then(() => {
+            return new Promise(resolve => {
+              setTimeout(() => {
```
noisecapella: Can you add a comment to explain the purpose of `setTimeout` here?

static/js/util/integration_test_helper.js
```diff
@@ -67,7 +68,9 @@ class IntegrationTestHelper {
       this.browserHistory.push(url);
       div = document.createElement("div");
       component = ReactDOM.render(
-        makeDashboardRoutes(this.browserHistory, this.store, () => null),
+        <div ref={component => window.component = component}>
```
noisecapella: Can you undo this change?

static/js/containers/ProfileFormContainer.js
aliceriot: Good idea, we are, after all, using a side-effect of `setState` and not actually, you know, setting any state, haha.

static/js/containers/UserPage_test.js
aliceriot: yeah

static/js/util/integration_test_helper.js
aliceriot: oh oop, totally missed that.

noisecapella:
:+1:

PR #783, opened 2016-08-02
Added roles to SETTINGS

### LINKED ISSUE #708: Include the current user's permissions in the SETTINGS variable
Although the front-end is not responsible for enforcing permissions to API endpoints, it does need to be aware of permissions so it can display the appropriate UI.

We will add the user's permissions and role per program(s) to the SETTING variable in javascript.

noisecapella:
Fixes #708 

#### What's this PR do?
Adds the roles assigned for a user to the `SETTINGS` variable

First, add a role to the user. You can do this using something like `Role.objects.create(user=user, program=program, role='instructor')`

Go to the dashboard, then open the browser console. Type `SETTINGS.roles` and verify that there's one role assigned to the right program.

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -13,6 +13,7 @@ from django.shortcuts import (
     Http404,
 )
 from django.utils.decorators import method_decorator
+from rolepermissions.shortcuts import available_perm_status
 
 from micromasters.utils import webpack_dev_server_host, webpack_dev_server_url
 from profiles.api import get_social_username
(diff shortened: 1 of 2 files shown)


aliceriot:
huh, so I don't seem to have `SETTINGS` available as global in the JS console.

But I can check it out in a debugger breakpoint, and it all looks good!

roles

aliceriot:
I don't see any issues with the code!

:+1:

PR #790, opened 2016-08-03
Enforced permissions on profiles REST API

### LINKED ISSUE #89: Add roles to profiles REST API
MIT instructors should be able to see more information about a user's profile than other users.

Staff can view all profiles, regardless of privacy settings. 

Learners can only view profiles where:
- privacy == `PUBLIC_TO_MM` AND the learner is in the same MM program as the profile
- OR privacy == `PUBLIC` 

Anonymous users can only view profiles where:
- privacy == `PUBLIC`

noisecapella:
Fixes #89 

#### What's this PR do?
Enforces permissions on the profiles API

profiles/permissions.py

Create two users and a program if necessary. One user should have `account_privacy` set to private and be assigned to the program in the `ProgramEnrollment` model. The other user should have an instructor or staff role added for the same program using the `Role` model.

Log in as the instructor or staff, then view the profile of the other user by going to `/api/v0/profiles/username/`. You should see everything that that user can see for their own profile. For example you should see `agreed_to_terms_of_service` which is not shown even in public profiles to other students.

### DIFF
--- a/profiles/permissions_test.py
+++ b/profiles/permissions_test.py
@@ -7,6 +7,8 @@ from django.db.models.signals import post_save
 from factory.django import mute_signals
 
 from backends.edxorg import EdxOrgOAuth2
+from courses.factories import ProgramFactory
+from dashboard.models import ProgramEnrollment
 from profiles.factories import (
     ProfileFactory,
     UserFactory,
@@ -15,6 +17,11 @@ from profiles.models import Profile
 from profiles.permissions import (
     CanEditIfOwner,
     CanSeeIfNotPrivate,
+)
+from roles.models import Role
+from roles.roles import (
+    Instructor,
+    Staff,
 )
 from search.base import ESTestCase
 
@@ -103,9 +110,9 @@ class CanSeeIfNotPrivateTests(ESTestCase):
         with self.assertRaises(Http404):
             perm.has_permission(request, view)
 
-    def test_cant_view_if_anonymous_user(self):
+    def test_cant_view_public_to_mm_if_anonymous_user(self):
         """
-        Anonymous are not supposed to view public_to_mm profiles.
+        Anonymous are not supposed to view public_to_mm or private profiles.
         """
         perm = CanSeeIfNotPrivate()
         with mute_signals(post_save):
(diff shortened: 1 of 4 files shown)


profiles/permissions_test.py
```diff
@@ -187,3 +221,106 @@ def test_can_view_if_verified_mm_user(self):
...
+        instructor_request = Mock(user=self.instructor, is_anonymous=Mock(return_value=False))
+        view = Mock(kwargs={'user': self.profile_user_id})
+
+        ProgramEnrollment.objects.create(
+            user=self.profile_user,
+            program=ProgramFactory.create(),
+        )
+        assert perm.has_permission(staff_request, view) is True
+        with self.assertRaises(Http404):
+            perm.has_permission(instructor_request, view)
```
giocalitri: We talked about this via chat, but I think you should at least rename the variables or put some comments to make explicit that the instructor is not actually instructor of the course.
I would actually go with a third user that is not even staff or instructor.

noisecapella: @pdpinch 'instructor' and 'staff' are the two roles we created in #709. We can change it but it would be out of the scope of this PR I think

profiles/views_test.py
```diff
@@ -14,14 +14,24 @@
...
 from profiles.serializers import (
     ProfileLimitedSerializer,
     ProfileSerializer,
 )
 from profiles.views import ProfileViewSet
+from roles.models import (
+    Instructor,
+    Role,
+    Staff,
+)
```
giocalitri: please import `Instructor` and `Staff` from the proper module
`from roles.roles import Staff, Instructor`

giocalitri:
A couple of minor comments but overall good

profiles/permissions_test.py
pdpinch: You're right. Out of scope. I'll follow up with Giovanni later. 
> @pdpinch <https://github.com/pdpinch> 'instructor' and 'staff' are the two roles we created in #709 <https://github.com/mitodl/micromasters/pull/709>. We can change it but it would be out of the scope of this PR I think
>

noisecapella:
Comments addressed

giocalitri:
👍

PR #794, opened 2016-08-04
upgrade flow to 0.30

### LINKED ISSUE #784: Upgrade flow
When I upgrade flow locally there are many extra warnings which need fixing

aliceriot:

closes #784 

#### What's this PR do?

well, it upgrades flow.

also the library we were previously using for flow definitions for redux hasn't been updated to work with 0.30 yet. I switched us to using [flow-typed](https://github.com/flowtype/flow-typed), which is a centralized repository of interface files for third-party packages (lots of them are supported).

There's a CLI client that we can use to install and update definitions. I installed the definitions for moment and redux (and removed iflow-moment and iflow-redux).

Read over the changes, make sure they make sense. Anywhere we should be more specific about our annotation would be :100:

There shouldn't be any regressions. Some small changes to some of the validation functions and the `boundDateField` function were necessary to get the typechecker to pass, as well as a couple of small edits here and there.


static/js/util/profile_edit.js
```diff
@@ -196,19 +196,19 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
     let validatedDay = 1;
     if (!omitDay) {
       validatedDay = validateDay(newEdit.day);
-      newEdit.day = validatedDay === undefined ? '' : String(validatedDay);
+      newEdit.day = String(validatedDay);
```
noisecapella: What does this change do?

noisecapella: Nevermind, it looks like the validate functions are changed to never return undefined

static/js/flow/profileTypes.js
```diff
@@ -22,12 +22,12 @@ export type DateEdit = {
...
-  position?: ?string;
-  industry?: ?string;
-  company_name?: ?string;
-  start_date?: ?string|DateEdit;
-  start_date_edit?: ?string|DateEdit;
-  end_date?: ?string|DateEdit;
+  position: string;
+  industry: string;
+  company_name: string;
+  start_date: string;
```
noisecapella: We should make this change to `graduation_date` too

static/js/flow/profileTypes.js
```diff
@@ -22,12 +22,12 @@ export type DateEdit = {
 
 export type WorkHistoryEntry = {
   id?: ?number;
-  position?: ?string;
-  industry?: ?string;
-  company_name?: ?string;
-  start_date?: ?string|DateEdit;
-  start_date_edit?: ?string|DateEdit;
-  end_date?: ?string|DateEdit;
+  position: string;
```
noisecapella: These all will still be `null` when a user first creates a new work history entry

noisecapella:
`end_date` needs to support null because the REST API doesn't allow empty strings, only strings that are valid date strings. Can we restrict this PR to only have type changes, and open another PR for other refactoring which needs further testing?

static/js/util/profile_edit.js
aliceriot: yeah, flow now doesn't allow `${undefined}`, which I think is the error that necessitated that change.

static/js/flow/profileTypes.js
aliceriot: agreed

static/js/flow/profileTypes.js
aliceriot: I modified `generateNewWorkHistoryEntry`, otherwise flow complains about type mismatches.

aliceriot:
Looking through the code in `boundDateField` (esp. in the `setNewDate` function it looks like if the date entered in a field isn't a valid date we'll still set `null` in the date field. I'm able to save an empty-end-date work history entry currently.

aliceriot:
ah nvm, it works on an old entry because it already has `end_date === null`, but on a new entry it fails.

aliceriot:
@noisecapella could you take a look at what I just pushed? it's a little bit of a workaround but it lets us post new work history entries without an end date.

package.json
```diff
@@ -27,10 +27,9 @@
     "eslint-plugin-react": "5.0.1",
     "fetch-mock": "4.4.0",
     "file-loader": "0.8.5",
-    "flow-bin": "^0.27.0",
+    "flow-bin": "^0.30.0",
     "history": "2.1.1",
     "iflow-lodash": "^1.1.16",
-    "iflow-moment": "^1.1.13",
     "iflow-react-router": "^1.1.16",
     "iflow-redux": "^1.0.37",
```
noisecapella: Can you remove this package since we don't use it anymore?

static/js/util/api.js
```diff
@@ -1,4 +1,4 @@
-// @flow
+// @flow weak
```
noisecapella: What does this do?

package.json
aliceriot: Yes, I meant to but missed it.

static/js/util/api.js
```diff
@@ -109,9 +109,15 @@ export function getUserProfile(username: string) {
 }
 
 export function patchUserProfile(username: string, profile: Profile) {
+  let clone = _.cloneDeep(profile);
+  clone.work_history.forEach(entry => {
```
noisecapella: Can you add a comment explaining this?

static/js/util/api.js
aliceriot: I believe it makes the file not throw errors when the type checker doesn't work out. I put it there temporarily because I just wanted to get your thoughts about the change below

aliceriot: (i.e. I think it should be removed before merging, but I'm not quite sure how to handle assigning the value we've type-annotated as `string` to be `null below)

static/js/util/api.js
aliceriot: yeah

static/js/util/api.js
noisecapella: I'm not sure if changing from `?string` to `string` is a good idea. Anyone can still PATCH the API to make these fields `null` so for some user these fields may end up being `null`. I'm not sure if we should be using flow to describe our types as they should be or to describe all possible edge cases. What do you think?

aliceriot: Not sure exactly what you mean, I agree that we might want to have `string|null` instead of just `string` for some of these fields, but then it's a question of how much work we put in to getting the typechecker to pass. I changed `end_date` on work history to be `string|null` and that seems to be passing.

static/js/util/validation.js
```diff
@@ -160,9 +160,14 @@ export function employmentValidation(profile: Profile): ValidationErrors {
     let errors = checkFieldPresence(profile, requiredFields, messages);
 
     profile.work_history.forEach((workHistory, index) => {
-      if (!isNilOrEmptyString(workHistory.end_date) && workHistory.end_date !== undefined &&
-        moment(workHistory.end_date).isBefore(workHistory.start_date, 'month')) {
-        _.set(errors, ['work_history', String(index), 'end_date'], "End date cannot be before start date");
+      if ( workHistory.start_date !== undefined ) {
```
noisecapella: Is `start_date` ever undefined?

aliceriot: ah no it is not, I forgot to change this.

static/js/util/api.js
noisecapella: My main concern is that these values can be null and if there are type errors due to a null value, they might be important to handle. I agree that it complicates type checking. Maybe we can convert null values to empty strings when we receive the profile from the API, then we should be safe in assuming no null values

aliceriot: That seems like a decent plan to me, I can give that a try.

aliceriot:
@noisecapella I added some code to change all empty strings in the profile to be `null` before we issue the `PATCH` request. I wrote a function called `cloneAndReplace` which makes this a little easier (we can also use it when we want to go from `null -> ""`.

lmk what you think!

static/js/reducers/index.js
```diff
@@ -54,9 +55,10 @@ export const profiles = (state: ProfileState = INITIAL_PROFILES_STATE, action: A
       getStatus: FETCH_PROCESSING
     });
   case RECEIVE_GET_USER_PROFILE_SUCCESS:
+    profile = cloneAndReplace(action.payload.profile, v => v === null ? "" : v);
```
noisecapella: I feel like this may have unintended consequences because each null does not necessarily replace a string type, though there are no profile fields like that right now. We are telling the type checker that these fields are not null, but we're just replacing the null with an empty string sentinel value.

aliceriot:
@noisecapella I looked around and there are a few libraries that add fancy functional stuff to JS (which would give us access to a `Maybe` Monad).

This one seems like the most mature and interesting: https://github.com/sanctuary-js/sanctuary
it's designed to work with a popular functional utility library: http://ramdajs.com/0.22.1/index.html

aliceriot:
there's also these two: 
https://github.com/ramda/ramda-fantasy
https://github.com/cwmyers/monet.js

aliceriot:
@noisecapella this is ready for review again.

static/js/util/api.js
```diff
@@ -109,6 +109,7 @@ export function getUserProfile(username: string) {
 }
 
 export function patchUserProfile(username: string, profile: Profile) {
+  // change "" entries to null because the API doesn't like empty strings
```
noisecapella: If we're not still doing this can you remove the comment?

static/js/util/profile_edit.js
```diff
@@ -191,37 +193,44 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
...
+    newEdit.year = str(validatedYear);
 
     // keep text up to date
     _.set(clone, editKeySet, newEdit);
 
-    if ([validatedDay, validatedMonth, validatedYear].includes(undefined)) {
-      // store the edit value and make the formatted date undefined so it doesn't pass validation
+    let padYear = s => _.padStart(s, 4, '0');
+
+    let allJust = R.reduce((acc, m)=> acc && m.toBoolean(), true);
```
noisecapella: Can we use `R.all` here instead?

aliceriot: oh yeah I think we can!

static/js/util/profile_edit.js
```diff
@@ -191,37 +193,44 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
       day: day !== undefined ? day : edit.day
     });
 
-    // these functions return undefined if a month, day, or year is invalid, and converts the value to a number
-    // except if the input value is an empty string, in which case an empty string is returned
-    let validatedDay = 1;
+    let str = S.maybe("", String); // returns String(Maybe.value) || ""
```
noisecapella: Can you add a comment here and for certain steps below? It would be helpful to other people working on this to get up to speed what's going on here since it's not typical JS

static/js/util/profile_edit.js
aliceriot:
```js
let allJust = R.all(S.isJust);
```

should do it!

static/js/util/profile_edit.js
aliceriot: here do you think more explanation than what I have is necessary?

noisecapella: For this one how about `str converts Maybe<String> to String`

static/js/util/profile_edit.js
```diff
@@ -191,37 +193,42 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
...
+    newEdit.year = str(validatedYear);
 
     // keep text up to date
     _.set(clone, editKeySet, newEdit);
 
-    if ([validatedDay, validatedMonth, validatedYear].includes(undefined)) {
-      // store the edit value and make the formatted date undefined so it doesn't pass validation
+    let padYear = s => _.padStart(s, 4, '0');
+
+    let validCheck = dates => Maybe.of(R.all(S.isJust)(dates)).filter(R.equals(true));
```
noisecapella: I feel like `validCheck` should be renamed `allJust` here, since there's nothing date specific about it. I would also add a comment: `allJust returns Just(true) if all items are of type Just, or else Nothing`

static/js/util/profile_edit.js
aliceriot: sounds good

static/js/util/profile_edit.js
aliceriot: Do you think some of these should be pulled out into a util file? I think you're right about the name, and others we may want the same functionality in the future.

static/js/util/sanctuary.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import R from 'ramda';
+import S, { Maybe } from 'sanctuary';
+
+/*
+ * returns Just(true) if all items are Just, else Nothing
+ */
+export const allJust = (items: Maybe[]) => (
```
noisecapella: Is it possible to change this to return `Just(items)` instead of `Just(true)`?

aliceriot: yes I think that's preferable

static/js/util/profile_edit.js
```diff
@@ -191,37 +193,38 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
...
 
-    if ([validatedDay, validatedMonth, validatedYear].includes(undefined)) {
-      // store the edit value and make the formatted date undefined so it doesn't pass validation
+    let padYear = s => _.padStart(s, 4, '0');
+
+    let stringifyDates = () => `${mstr(validatedYear.map(padYear))}-${mstr(validatedMonth)}-${mstr(validatedDay)}`;
+
+    let dateList = [validatedYear, validatedMonth, validatedDay];
+
+    let dateString = S.maybe("", stringifyDates, allJust(dateList));
```
noisecapella: If we can change `allJust` to return `Just(dateList)`, we should change `stringifyDates` to take year, month, and day as parameters instead of using the values from the context here

noisecapella:
Can you rebase please?

aliceriot:
@noisecapella i made some changes, and pulled some of the utility functions into `util/sanctuary.js` so they can be unit tests.

noisecapella:
:+1:

Issue #778, opened 2016-08-01
Change index to store users based on Program enrollments


gsidebo:
_This is replacing #774. For more details on why this indexing change is needed, read the description of that issue_

To get around some Elasticsearch limitations, we will be changing the structure of our index. We will be creating one indexed document per Program enrollment rather than per User. This means that a User enrolled in 2 Programs would have 2 documents in the index. In terms of Program-related data (enrollments, grades, etc), only the information that is directly linked to the given Program would be a part of the document.

pdpinch:
Can this be closed?

(I was reminded recently that we if we write "fixes" before the issue number in the PR description, the issue will close automatically.)

Issue #743, opened 2016-07-21
Refactor Enrollment/Certificate models and tests


gsidebo:
#711 introduced the CachedEnrollment and CachedCertificate to our codebase to provide a way for us to cache edx data in our own database. The model definitions are identical, the `get_` functions to retrieve their data in `dashboard/api.py` are very similar, and the test cases for these models in `dashboard/api_test.py` are also very similar. We should be able to abstract a lot of this logic into a parent class that both models inherit from and eliminate this repetition.

PR #811, opened 2016-08-05
Changed indexing structure and logic to use a user's program enrollments

gsidebo:

#778 

#### What's this PR do?

Changes our indexing to index one record per user _per program_. A user enrolled in two programs, for example, will now have two records in the index. The enrollment/certificate/etc data for each of those records would only be those that are associated with the appropriate program.

Probably search/api.py. This might be one that you want to call me over to co-review at some point

Load the realistic users if you haven't already, the run the recreate_index manage command. Once you've done that, you should be able to query ES and see the new indexing format


search/commands_test.py
```diff
@@ -1,54 +0,0 @@
-"""
```
gsidebo: moved these test cases to search/api_test.py since it's testing search/api.py functionality

dashboard/fixtures/certificates.json
```diff
@@ -5,14 +5,14 @@
 	    "certificate_type": "verified",
 	    "status": "downloadable",
 	    "download_url": "http://www.example.com/demo.pdf",
-	    "grade": "0.98"
+	    "grade": 98
```
noisecapella: Why is this being changed? This is how the data comes from edX

search/tasks.py
```diff
@@ -4,30 +4,52 @@
...
+
+@async.task
+def index_program_enrolled_users(program_enrollments):
     """
     Index profiles
 
     Args:
-        users (iterable of User):
-            Iterable of Users
+        program_enrollments: Iterable of ProgramEnrollments
```
noisecapella: Can you add a type here for `program_enrollments`? It should go in parentheses after the name according to the sphinx-napoleon convention: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#docstrings

dashboard/fixtures/certificates.json
gsidebo: i don't like this either, but the method where we post-process data coming from edx is being mocked out in multiple places: https://github.com/mitodl/micromasters/blob/master/dashboard/api_test.py#L878. `get_student_certificates` and `get_student_enrollments` are doing _way_ too many things right now. i have an [issue open](https://github.com/mitodl/micromasters/issues/743) to refactor the code for CachedEnrollments/Certificates, which would include those methods. I can make a note to revert this JSON file to mimic edx data again

search/tasks.py
gsidebo: will do

dashboard/fixtures/certificates.json
noisecapella: It looks like we are still passing in this JSON in the `Certificates` objects in our tests like we did before. Wouldn't this cause a difference of behavior between our tests and the edX API?

If we need `grade` to be an int in the Elasticsearch index for some reason, we should make that change in the serializer function in `search/api.py`. But I also still don't understand why grade needs to be an int here anyway?

gsidebo: Elasticsearch limitation: https://discuss.elastic.co/t/less-than-1-interval-values/55201

we can't use decimal/float values for a histogram, which we need to construct the value ranges for the filter-by-grade component. 

as far as making it an int in the model vs. performing that in the serialization, i'm trying to remember exactly why i/we made that decision. can't find the hipchat convo right now, but off the top of my head, this would cause a disagreement between the data in our models and the data in Elasticsearch. i think we would need a compelling reason to introduce that disparity. i feel like i'm missing the real reason though - i'll dig through my notes and in hipchat

giocalitri: I might be wrong, but it looks like the decimal histogram has been implemented 
https://github.com/elastic/elasticsearch/issues/4847

gsidebo: that fix looks like it's part of an unreleased version. here's their releases page: https://github.com/elastic/elasticsearch/releases. i'm not quite sure what to make of the v2.x.x versions mixed in with the v.5.0.0-x versions (elasticsearch version vs. lucene version??), but it looks like they haven't released that [histogram fix](https://github.com/elastic/elasticsearch/pull/19551) yet (it's tagged as 'v5.0.0-alpha5'). 

i can create an issue to try that version when it releases, make sure it works with decimal values for grades, and revert our code as necessary. would that address your concerns @noisecapella ?

noisecapella: I think it's fine to have our Elasticsearch data be different than what we store in the JSONField in the model. For example we may want to create an extra field for customized sorting in Elasticsearch. Since this JSON is supposed to represent data coming from edX we shouldn't make any change here. I'm fine with changing `grade` to an int in the serializer function, I think it's less effort than upgrading Elasticsearch, and I'm not sure it's wise to point to an alpha version. Alternatively we can also leave the field alone and create some other field like `grade_histogram` which represents the grade as an int.

gsidebo: @noisecapella - actually, im noticing something pretty weird here. that edx post-processing code that i mentioned earlier doesn't even seem to be in this branch. i'll dig into this a little more

dashboard/factories.py
```diff
@@ -47,3 +50,27 @@ class CachedEnrollmentFactory(DjangoModelFactory):
...
+    class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,old-style-class
+        model = ProgramEnrollment
+
+    @classmethod
+    def create(cls, **kwargs):
+        # ProgramEnrollments should only exist if there is a CachedEnrollment associated with
+        # the given User and Program. Instead of creating a new record with the factory, we
+        # will create the necessary objects to trigger its creation.
+        user = kwargs['user'] = kwargs.get('user', UserFactory.create())
+        program = kwargs['program'] = kwargs.get('program', ProgramFactory.create())
```
giocalitri: why do you assign a copy of the objects to `kwargs`?

dashboard/factories.py
```diff
@@ -47,3 +50,27 @@ class CachedEnrollmentFactory(DjangoModelFactory):
...
+        course = CourseFactory.create(program=program)
+        course_run = CourseRunFactory.create(course=course)
+        CachedEnrollmentFactory.create(user=user, course_run=course_run)
+        CachedCertificateFactory.create(user=user, course_run=course_run)
```
giocalitri: technically you do not need to create a certificate for the user.

dashboard/signals.py
```diff
@@ -25,27 +38,61 @@ def precreate_programenrollment(sender, instance, **kwargs):  # pylint: disable=
...
                 ProgramEnrollment.objects.filter(
                     user=user,
                     program=program
                 ).delete()
 
 
-@receiver(post_save, sender=CachedEnrollment, dispatch_uid="update_programenrollment")
-def create_programenrollment(sender, instance, **kwargs):  # pylint: disable=unused-argument
+@receiver(post_save, sender=CachedEnrollment, dispatch_uid="cachedenrollment_post_save")
+def handle_update_enrollment(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
giocalitri: I do not understand why you changed the name of this signal

dashboard/signals.py
```diff
@@ -25,27 +38,61 @@ def precreate_programenrollment(sender, instance, **kwargs):  # pylint: disable=
...
+
+
+@receiver(post_save, sender=CachedCertificate, dispatch_uid="cachedcertificate_post_save")
+def handle_update_certificate(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """When a CachedCertificate model is updated, update index """
+    if instance.data is not None:
+        program_enrollment = ProgramEnrollment.objects.get_or_create(
+            user=instance.user,
+            program=instance.course_run.course.program
+        )[0]
```
giocalitri: I would write this in the following way 
```
program_enrollment, _ = ProgramEnrollment.objects.get_or_create(
    user=instance.user,
    program=instance.course_run.course.program
)
```

dashboard/signals.py
```diff
@@ -25,27 +38,61 @@ def precreate_programenrollment(sender, instance, **kwargs):  # pylint: disable=
...
             user=instance.user,
             program=instance.course_run.course.program
-        )
+        )[0]
+        index_program_enrolled_users.delay([program_enrollment])
+
+
+@receiver(post_save, sender=CachedCertificate, dispatch_uid="cachedcertificate_post_save")
+def handle_update_certificate(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """When a CachedCertificate model is updated, update index """
```
giocalitri: very very minor nit
"""When a CachedCertificate model is updated, update index"""
(no space and the end of the string)

dashboard/signals.py
```diff
@@ -25,27 +38,61 @@ def precreate_programenrollment(sender, instance, **kwargs):  # pylint: disable=
...
+        index_program_enrolled_users.delay([program_enrollment])
+
+
+@receiver(post_delete, sender=CachedEnrollment, dispatch_uid="cachedenrollment_post_delete")
+def handle_delete_enrollment(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """Update index when CachedEnrollment model instance is deleted."""
+    user = instance.user
+    program = instance.course_run.course.program
+    program_enrollment = ProgramEnrollment.objects.filter(user=user, program=program).first()
+    if program_enrollment and CachedEnrollment.active_count(user, program) == 0:
```
giocalitri: if the code `if program_enrollment` means `if program_enrollment is not None` you should be explicit

micromasters/utils.py
```diff
@@ -19,6 +19,14 @@ def webpack_dev_server_url(request):
     return 'http://{}:{}'.format(webpack_dev_server_host(request), settings.WEBPACK_DEV_SERVER_PORT)
 
 
+def remove_key(dictionary, key):
+    """
+    Helper method to remove a key from a dict and return the dict
+    """
+    del dictionary[key]
+    return dictionary
```
giocalitri: what's the benefit of doing this instead of just calling `del dictionary[key]` wherever you need it?

giocalitri: actually, returning the dictionary here is useless, because the `del` is working in the same object passed via arg from the parent. And this means that the function itself does not have a real meaning... Unless I am missing something

giocalitri:
I have some comments. 

I also think all the refactoring looks good, but you could have postponed at least some parts of it.

dashboard/factories.py
gsidebo: part of an earlier refactor idea. i'll remove

dashboard/factories.py
gsidebo: technically no, but it seems that in other test cases, we prefer to have an enrollment and a certificate. i put this in for simple convenience

dashboard/signals.py
gsidebo: it's doing a few different things now. it's creating (or getting) a program enrollment, and it's reindexing based on that enrollment. it's not just creating a ProgramEnrollment anymore, so it makes sense to change the name

dashboard/signals.py
gsidebo: will change

dashboard/factories.py
giocalitri: ok

dashboard/signals.py
giocalitri: ok

giocalitri: please make sure that all the `dispatch_uid` are unique

micromasters/utils.py
gsidebo: this used to be a part of search/api_test.py ([line](https://github.com/mitodl/micromasters/pull/811/files#diff-63f4279d1f66269772e1f02ca030fda9L45)). its being used in a list comprehension, so the function has to return the actual dict after removing the key for it to work correctly. this definitely could use a better name - `remove_key` is (obviously) misleading

giocalitri: oh, I understand now... maybe try to add what you just said in the docstring.

gsidebo: docstring looks good to me ("Helper method to remove a key from a dict and return the dict"). describes exactly what it's doing. function name is now `dict_without_key` - does that help?

giocalitri: What I meant is to put some description of why someone would use this function instead of calling directly `del`

gsidebo:
i'm getting an error in searchkit. might be data-related. i'll look into it

noisecapella:
Searchkit works fine for me

gsidebo:
yep, i had bad data in the index

dashboard/signals.py
```diff
@@ -25,27 +38,61 @@ def precreate_programenrollment(sender, instance, **kwargs):  # pylint: disable=
...
+
+
+@receiver(post_delete, sender=CachedCertificate, dispatch_uid="cachedcertificate_post_delete")
+def handle_delete_certificate(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """Update index when CachedCertificate model instance is deleted."""
+    user = instance.user
+    program = instance.course_run.course.program
+    program_enrollment = ProgramEnrollment.objects.filter(user=user, program=program).first()
+    if not program_enrollment:
+        index_users.delay([user])
```
noisecapella: Should we just get rid of `index_users`? It seems like if there is no `ProgramEnrollment` for that user, the index should not need to know about that user at all

search/api.py
```diff
@@ -75,7 +75,7 @@ def get_conn(verify=True):
     return _CONN
 
 
-def _index_users_chunk(users):
+def _index_program_enrolled_users_chunk(program_enrollments):
```
noisecapella: Can you update arguments in the docstring?

search/api.py
```diff
@@ -90,21 +90,19 @@ def _index_users_chunk(users):
...
         ))
     refresh_index()
-
     return insert_count
 
 
-def index_users(users, chunk_size=100):
+def index_program_enrolled_users(program_enrollments, chunk_size=100):
     """
     Add/update profile records in Elasticsearch.
```
noisecapella: Update this docstring too please

dashboard/signals.py
gsidebo: keep in mind that this is looking for a `ProgramEnrollment` for a User _and_ a `Program`. A User could have other ProgramEnrollments. your comment does point out an issue, though. i had this code in place to handle cases where the ProgramEnrollment had already been deleted by an earlier signal, but that in itself should trigger a reindexing. ill get rid of this and change up the enrollment signal as well

search/api_test.py
```diff
@@ -32,65 +35,57 @@
...
-    elasticsearch_url = settings.ELASTICSEARCH_URL
-    if not elasticsearch_url.startswith("http"):
-        elasticsearch_url = "http://{}".format(elasticsearch_url)
-    url = urljoin(
-        elasticsearch_url,
-        "{}/{}".format(settings.ELASTICSEARCH_INDEX, "_search")
-    )
-    return get(url).json()['hits']
+    def search(self):
+        """Gets full index data from the _search endpoint"""
```
noisecapella: This actually just gets the first 20 or so results, though I'm not sure we run tests with more results than that

search/api_test.py
```diff
@@ -104,188 +99,154 @@ def test_user_add(self):
         """
         Test that a newly created User is indexed properly
```
noisecapella: Can you update the docstrings for each of these tests?

search/api_test.py
```diff
@@ -100,192 +95,160 @@ class IndexTests(ESTestCase):
...
-        certificate = CachedCertificateFactory.create(user=user)
-        assert_search(search(), [user])
-        certificate.delete()
-        assert_search(search(), [user])
+        with mute_signals(post_save):
+            program_enrollments = [ProgramEnrollmentFactory.build() for _ in range(10)]
+        with patch('search.api._index_program_enrolled_users_chunk', autospec=True, return_value=0) as index_chunk:
+            index_program_enrolled_users(program_enrollments, chunk_size=4)
+            assert index_chunk.call_count == 3
+            index_chunk.assert_any_call(program_enrollments[0:4])
```
gsidebo: @noisecapella @giocalitri i refactored this test case (`test_index_program_enrolled_users`, formerly `test_index_users`). this was taking 2.5 sec with calls to ES and to the database, and all it was uniquely testing was that the 'index_' function could handle an iterable. let me know if you have a problem with this refactor

dashboard/factories.py
```diff
@@ -47,3 +50,28 @@ class CachedEnrollmentFactory(DjangoModelFactory):
...
+    program = SubFactory(ProgramFactory)
+
+    class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,old-style-class
+        model = ProgramEnrollment
+
+    @classmethod
+    def create(cls, **kwargs):
+        # ProgramEnrollments should only exist if there is a CachedEnrollment associated with
+        # the given User and Program. Instead of creating a new record with the factory, we
+        # will create the necessary objects to trigger its creation.
```
giocalitri: nit: please move this comment to a docstring

search/api.py
```diff
@@ -90,83 +90,106 @@ def _index_users_chunk(users):
...
 
 
-def serialize_user(user):
+def remove_user(user):
+    """
+    Remove a user from Elasticsearch.
+    """
+    program_enrollments = ProgramEnrollment.objects.filter(user=user).select_related('user', 'program').all()
+    for program_enrollment in program_enrollments:
+        remove_program_enrolled_user(program_enrollment)
```
noisecapella: From the code coverage report it looks like this line is not triggered by tests. Can you add or adjust a test to cover this line?

gsidebo: good catch. just added a test

giocalitri:
the functionality looks good to me 👍 for my part of the review

noisecapella:
I'm happy too :+1:

Issue #737, opened 2016-07-19
Sign-up: change "where are you from" to country of birth & nationality


pdpinch:
We will ask users for country of birth and nationality, instead of the full country, region and city of birth. This fits better with some future use cases for this profile data. 

![image](https://cloud.githubusercontent.com/assets/430126/18098538/a505fa4e-6eb0-11e6-814b-d4351f8cd02d.png)

The text and style for the tooltip is below: 

![image](https://cloud.githubusercontent.com/assets/20047260/17441844/f3c9be2c-5b00-11e6-98f7-d938f4334820.png)

# Acceptance Criteria

- [ ] update profile model to include nationality (and remove birth_region and birth_city) (#775)
- [ ] update serialization for elastic search (#775)
- [ ] update profile API (#775); nationality is an optional field
- [ ] update react component
- [ ] use a simple country list for nationalities
- [ ] implement tool tip on click, consistent with other tool tips UX and styling

# Out of scope

- updating search results

aliceriot:
is nationality a drop-down or a free form text box?

aliceriot:
if it's the former, do we have a list to work with?

aliceriot:
we might be able to use this: https://gist.github.com/didats/8154290

although we'd probably want to reformat it as JSON

aliceriot:
@pdpinch do we have an idea of what we want to do with this yet?

noisecapella:
Just a thought that I had this morning, but national origin is a protected category for employment. I'm not sure what the regulations are regarding admissions but I'm wondering if it sends the wrong message to ask for a person's birth country. I can't think of other forms which ask for this information. In general if we ask for pieces of information just for analytics we should communicate that explicitly, else they will assume the worst.

It also feels like we're invalidating their nationality by asking this question. If they immigrate to the US and become a citizen, they are equal to natural born citizens under the law. If we have a good reason for this question we should be explicit about expressing that.

aliceriot:
I agree, I think in general we should be a little more explicit about how information that users provide is going to be used.

When I'm signing up for a service, if they ask me for a whole bunch of info but it's unclear what it's for (or it's unclear they're going to use it in my interest) I'll often input fake data.

roberthouse54:
[...]

PR #815, opened 2016-08-08
Replaced filtering with hierarchical filtering

### LINKED ISSUE #785: Search: Implement Facet UI for filtering on learner's location
We can use the [SearchKit taxonomy component](http://demo.searchkit.co/taxonomy) to present a hierarchical facet.

current_residence_facet

The top level will be country, 2nd level is region. These both come from a controlled vocabulary that's currently implemented in JavaScript, and are required fields in the user profile. The 3rd level is city, which is free text currently. 

At the moment, we are storing acronyms for city and region in the database and in the search index. We will have to map them in the facet UI, or expand them when indexing. In planning, the group was leaning towards indexing the expanded country and region names.

noisecapella:
Fixes #785

#### What's this PR do?
Replaces the country facets with a hierarchical facets

LearnerSearch.js

Run `./manage.py recreate_index`. Then go to the learner search and try faceting on a state or territory

`profile` and `program` don't need to be nested because there is only one of them, so these are now indexed as `object`. We don't have checkboxes for now because there's no equivalent `itemComponent` prop available for `HierarchicalMenuFilter`. Also, countries and states are translated in the Javascript, but we may want to change our indexing to store with their full names in the future.

`birth_country` was left alone because we're planning on removing `birth_state_or_territory` in the near future: #737 

screenshot from 2016-08-08 15-37-49

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import {
   SearchkitComponent,
+  HierarchicalMenuFilter,
   Hits,
   NoHits,
   SelectedFilters,
@@ -10,12 +11,25 @@ import {
 } from 'searchkit';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 import Card from 'react-mdl/lib/Card/Card';
+import iso3166 from 'iso-3166-2';
 
 import LearnerResult from './search/LearnerResult';
 import CountryRefinementOption from './search/CountryRefinementOption';
 import FilterVisibilityToggle from './search/FilterVisibilityToggle';
 import HitsCount from './search/HitsCount';
 import type { Option } from '../flow/generalTypes';
+
+let makeSearchkitTranslations: () => Object = () => {
+  let translations = {};
+  for (let code of Object.keys(iso3166.data)) {
+    translations[code] = iso3166.data[code].name;
+    for (let stateCode of Object.keys(iso3166.data[code].sub)) {
+      translations[stateCode] = iso3166.data[code].sub[stateCode].name;
+    }
+  }
+
+  return translations;
+};
 
 export default class LearnerSearch extends SearchkitComponent {
   props: {
@@ -28,12 +42,7 @@ export default class LearnerSearch extends SearchkitComponent {
     { value: 'dob', label: "Sort: dob" },
   ];
 
-  profileFieldOptions: Function = (): Object => ({
-    type: 'nested',
-    options: {
-      path: 'profile'
-    }
-  });
+  searchkitTranslations: Object = makeSearchkitTranslations();
 
   render () {
     return (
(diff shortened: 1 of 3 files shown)


static/js/components/LearnerSearch.js
```diff
@@ -10,13 +11,26 @@ import {
...
 import Card from 'react-mdl/lib/Card/Card';
+import iso3166 from 'iso-3166-2';
 
 import LearnerResult from './search/LearnerResult';
 import CountryRefinementOption from './search/CountryRefinementOption';
 import FilterVisibilityToggle from './search/FilterVisibilityToggle';
 import HitsCount from './search/HitsCount';
 import type { Option } from '../flow/generalTypes';
 
+let makeTranslations: () => Object = () => {
```
gsidebo: since 'translations' is a searchkit-specific term/concept, i think it would help to name this `makeSearchkitTranslations` and the associated variable `searchkitTranslations`

gsidebo:
a couple small design quirks:

ss 2016-08-09 at 11 35 50

ss 2016-08-09 at 11 35 29

A long region name seems to squeeze the display of the count, and a fantastically long name exaggerates that

static/js/components/LearnerSearch.js
```diff
@@ -10,13 +11,26 @@ import {
...
+  let translations = {};
+  for (let code of Object.keys(iso3166.data)) {
+    translations[code] = iso3166.data[code].name;
+    for (let stateCode of Object.keys(iso3166.data[code].sub)) {
+      translations[stateCode] = iso3166.data[code].sub[stateCode].name;
+    }
+  }
+
+  return translations;
```
gsidebo: this object has about 5200 keys when all is said and done. probably not a _terribly_ significant memory/performance hit, but i'm wondering if there's an easy-enough way to build this object on demand and/or cache the values that we've already seen

noisecapella: Once we implement #828 we should not need this function anymore so I'll leave it for the time being

static/js/components/LearnerSearch.js
noisecapella: Will do

static/js/components/LearnerSearch.js
gsidebo: sounds good

noisecapella:
How's this? 
screenshot from 2016-08-10 11-10-28

gsidebo:
still a couple design issues:

**1) 23 character region name pushes counts out of the container.** there might not _be_ a 23 character region name, but i figured it was enough of a possibility to mention

ss 2016-08-10 at 11 24 12

**2) slim window size pushes counts out of the container.** this might be one where we say we don't care about the counts, but i think the container should have enough padding to cut off the text so the text isn't running all the way up to the border

ss 2016-08-10 at 11 18 41

ss 2016-08-10 at 11 29 46

gsidebo:
looks good besides those two design issues

noisecapella:
I adjusted the padding to fix the count hiding issue, at least for certain browser widths.

gsidebo:
ok looks good enough, i'd say. if we want to get _really_ nitpicky about the, we can fix it as part of a separate issue

👍

PR #816, opened 2016-08-08
Updated realistic user data to include users with current country != …

### LINKED ISSUE #813: Change birth country for fake users to not match current country
The birth location seems to match the current location exactly for all fake users, but these should deviate at least some of the time

gsidebo:

Closes #813 

#### What's this PR do?

Adds 6 users with a current country that is different from their birth country.

Here are the usernames for those 6 users:

['fake.terri.pena',
 'fake.ted.sullivan',
 'fake.camille.jones',
 'fake.ana.cabrera',
 'fake.alvaro.santana',
 'fake.hugo.suarez']

Delete your current set of fake users and re-generate. Check the generated set for the users mentioned above

### DIFF
--- a/profiles/management/realistic_program_data.json
+++ b/profiles/management/realistic_program_data.json
@@ -1,218 +1,218 @@
 [
     {
+        "description": "Learn stuff about digital learning.",
+        "title": "Digital Learning",
         "courses": [
             {
+                "position_in_program": 1,
+                "description": "An introductory course to Digital Learning",
+                "title": "Digital Learning 100",
                 "course_runs": [
                     {
-                        "end_date": "2015-05-15T00:00:00", 
-                        "title": "Digital Learning 100 - January 2015", 
-                        "enrollment_start": "2015-01-15T00:00:00", 
-                        "edx_course_key": "course-v1:MITx+Digital+Learning+100+Jan_2015", 
-                        "upgrade_deadline": "2015-01-22T00:00:00", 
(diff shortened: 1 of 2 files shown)


profiles/management/realistic_program_data.json
```diff
@@ -1,218 +1,218 @@
 [
```
gsidebo: no actual changes here. this is just a regenerated copy of the same data

noisecapella:
Can you add your generation script too? Maybe create a `tools` directory for that purpose?

gsidebo:
@noisecapella generation script: https://github.com/mitodl/realistic-mm-users

i think we should keep the generation scripts out of this codebase. its extra code that doesn't have anything to do with actual MM functionality. i can add a link in the README if that's something we want

noisecapella:
I'm getting a 404 on that link, can you give me access to it? Is it a private repository?

gsidebo:
oops. sorry about that. just added the ODL engineering team as admins. does it work now?

noisecapella:
Yes, looks good. I feel like this should be public with an open source license, since there's no reason to keep it private (any opinion @pdpinch?)

pdpinch:
Unless there's a good reason, it should be open source. Use the same license (BSD 3 clause?) as micromasters.

noisecapella:
:+1: 

I found a bug when deleting existing users: #818

gsidebo:
script repo has been updated

PR #823, opened 2016-08-09
Search REST API Proxy

### LINKED ISSUE #701: Search API for staff users
wrap the standard elastic search API with access controls, including privacy restrictions

only signed in users can use the API.

Admin can search all users and all user data. 

Staff can search users in their program, and all user data (including grades)

(Learners can only search users who share their profile, and they can never see grades. We'll implement learner-to-learner search as another API) 

requires roles issues to be completed. Blocked by #709

giocalitri:
closes #701 

#### What's this PR do?
Introduces a REST API to proxy requests to elasticsearch, enforce permissions and limit results to allowed ones

`search.views`

In your `.env` you should set the variable
`ELASTICSEARCH_URL=http://elastic:9200` (not sure about the actual value for not Mac users) 

To check that the rest API actually works, access its url `/api/v0/search/`
To verify that searchkit works, edit your `.env` file and include/edit the entry
`CLIENT_ELASTICSEARCH_URL=http://<your_ip>:8079/api/v0/search/`

You should also verify that the user cannot see students from programs he is not supposed to.
To do that, load fake users with the proper management command and make your user `Staff` of only one of these programs.

This PR does not take care of setting a basic filter for the program at the SearchKit level: I think that should go together with the refactoring of the UI in general.

### DIFF
--- a/roles/api.py
+++ b/roles/api.py
@@ -1,0 +1,23 @@
+"""
+API for roles
+"""
+from rolepermissions.verifications import has_object_permission
+
+from roles.models import Role
+
+
+def get_advance_searchable_programs(user):
+    """
+    Helper function to retrieve all the programs where the user is allowed to search
+
+    Args:
+        user (User): Django user instance
+    Returns:
+        list: list of courses.models.Program instances
+    """
+    user_role_program = Role.objects.filter(user=user)
+    programs = [
+        role.program for role in user_role_program
+        if has_object_permission('can_advance_search', user, role.program)
+    ]
+    return programs
--- a/search/permissions_test.py
+++ b/search/permissions_test.py
@@ -1,0 +1,47 @@
+"""
+Tests for permissions classes
+"""
+from django.contrib.auth.models import AnonymousUser
+from django.test import RequestFactory
+
+from courses.factories import ProgramFactory
+from profiles.factories import UserFactory
+from roles.models import Role
+from roles.roles import Staff
+from search import permissions
+from search.base import ESTestCase
+
+
+class PermissionsTests(ESTestCase):
+    """Tests for the search view permissions"""
+
+    @classmethod
+    def setUpTestData(cls):
+        super(PermissionsTests, cls).setUpTestData()
+        # create an user
+        cls.user = UserFactory.create()
+        # create the program
+        cls.program = ProgramFactory.create(live=True)
+
+    def setUp(self):
+        super(PermissionsTests, self).setUp()
+        self.request = RequestFactory().get('/')
+        self.request.user = AnonymousUser()
+
+    def test_user_can_search(self):
+        """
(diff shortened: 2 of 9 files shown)


gsidebo:
whoops. just saw that this is WIP. carry on!

search/permissions_test.py
```diff
@@ -0,0 +1,47 @@
...
+        cls.program = ProgramFactory.create(live=True)
+
+    def setUp(self):
+        super(PermissionsTests, self).setUp()
+        self.request = RequestFactory().get('/')
+        self.request.user = AnonymousUser()
+
+    def test_user_can_search(self):
+        """
+        Checks that an user can search only if she shas the right permissions
```
noisecapella: `shas` -> `has`

roles/api.py
```diff
@@ -0,0 +1,23 @@
+"""
+API for roles
+"""
+from rolepermissions.verifications import has_object_permission
+
+from roles.models import Role
+
+
+def get_advance_searchable_programs(user):
```
noisecapella: Does it matter whether a program is live or not?

search/views.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class ElasticProxyView(APIView):
+    """
+    Elasticsearch proxy needed to enforce authentication and permissions
+    """
+    authentication_classes = (SessionAuthentication, )
+    permission_classes = (IsAuthenticated, UserCanSearchPermission, )
+
+    def _search_elasicsearch(self, request):  # pylint: disable=no-self-use
```
noisecapella: Typo: `_search_elasticsearch`

roles/api.py
giocalitri: Honestly right now it exposes users also for not live programs, but I do not think this is a decision that should be done at this level

giocalitri: Meaning, this might be solved also by not exposing the dashboard for not live programs, but I do not know if is is useful to return these users.

Actually we might even solve this by filtering the users at index time...

search/views_test.py
```diff
@@ -0,0 +1,164 @@
...
+
+class SearchTests(ESTestCase, APITestCase):
+    """Tests for the search api view"""
+
+    @classmethod
+    def setUpTestData(cls):
+        super(SearchTests, cls).setUpTestData()
+        # create some students
+        with mute_signals(post_save):
+            cls.students = [(ProfileFactory.create()).user for _ in range(30)]
```
noisecapella: Just FYI the search queries below will only get the first 20 results. I don't think that matters in any of these tests though

giocalitri: it does not really matter

roles/api.py
noisecapella: Ok, I filed #837 to deal with this later

noisecapella:
Looks good :+1:

Issue #731, opened 2016-07-18
front-end for filtering by average grade


pdpinch:
Add the appropriate searchkit component to the search page to work with the new facet from #682

pdpinch:
Blocked by #682

PR #826, opened 2016-08-10
Added program grade to ES index to support filtering by grade

### LINKED ISSUE #682: filter by average grade (back end)

gsidebo:

Closes #682 

#### What's this PR do?

Adds 'grade_average' to the ES index (calculated at index time). With this, we can work on #731 (filtering users by program grade)

Probably dashboard/serializers.py since that's where the program grade is calculated. Probably search/api.py after that

If you'd like, you can reindex and check the actual ES results using this query (Postman is very useful for this). This will filter by average grade >= 75 and build a histogram in 10 pt intervals:

```
{
    "query": {
        "nested": {
            "path": "program",
            "query": {
                "bool": {
                    "must": [
                        { "range": { "program.grade_average": { "gte": 75 } } }
                    ]
                }
            }
        }
    },
    "aggs": {
        "overall_grade": {
            "nested" : {
                "path" : "program"
            },
            "aggs": {
                "grade_hist" : {
                    "histogram" : {
                        "field" : "program.grade_average",
                        "interval" : 10
                    }
                }
            }
        }
    }
}
```

### DIFF
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -1,0 +1,51 @@
+"""
+Provides functionality for serializing a ProgramEnrollment for the ES index
+"""
+from decimal import Decimal
+
+
+class UserProgramSerializer:
+    """
+    Provides functions for serializing a ProgramEnrollment for the ES index
+    """
+    @staticmethod
+    def calculate_certificate_grade_average(certificates):
+        """
+        Calculates an average grade (integer) from a list of certificates with <1 decimal grades
+        """
+        return None if len(certificates) == 0 else round(
+            (sum((Decimal(certificate['grade']) for certificate in certificates)) / len(certificates)) * 100
+        )
+
+    @classmethod
+    def serialize_valid_edx_data(cls, related_edx_obj_set, program):
+        """
+        Args:
+            related_edx_obj_set (RelatedManager): RelatedManager for set of cached edX model objects
+                (eg: CachedEnrollment)
+            program (Program): Program object
+
+        Returns:
+            list of dict: List of serialized objects
+        """
+        return [
+            edx_obj.data for edx_obj in related_edx_obj_set.filter(
+                course_run__course__program=program
+            ).exclude(data__isnull=True)
+        ]
+
+    @classmethod
+    def serialize(cls, program_enrollment):
+        """
+        Serializes a ProgramEnrollment object
+        """
+        user = program_enrollment.user
+        program = program_enrollment.program
+        program_cached_enrollments = cls.serialize_valid_edx_data(user.cachedenrollment_set, program)
+        program_cached_certificates = cls.serialize_valid_edx_data(user.cachedcertificate_set, program)
+        return {
+            'id': program.id,
+            'enrollments': program_cached_enrollments,
+            'certificates': program_cached_certificates,
+            'grade_average': cls.calculate_certificate_grade_average(program_cached_certificates)
+        }
--- a/dashboard/serializers_test.py
+++ b/dashboard/serializers_test.py
@@ -1,0 +1,101 @@
+"""
+Test cases for the UserProgramSerializer
+"""
+from django.test import TestCase
+from django.db.models.signals import post_save
+from factory.django import mute_signals
+from profiles.factories import (
+    EducationFactory,
+    EmploymentFactory,
+    ProfileFactory,
+)
+from courses.factories import (
+    ProgramFactory,
+    CourseFactory,
+    CourseRunFactory,
+)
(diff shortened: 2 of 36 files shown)


noisecapella:
Is this ready for review?

gsidebo:
it is. sorry, forgot the label

dashboard/serializers.py
```diff
@@ -0,0 +1,52 @@
...
+        return None if len(certificates) == 0 else round(
+            (sum([Decimal(certificate['grade']) for certificate in certificates]) / len(certificates)) * 100
+        )
+
+    @classmethod
+    def serialize_valid_edx_data(cls, related_edx_obj_set, program):
+        """
+
+        Args:
+            related_edx_obj_set: Related QuerySet of cached edX model objects (eg: CachedEnrollment)
```
noisecapella: Can you add types for these arguments and the return value?

dashboard/serializers.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Provides functions for serializing a ProgramEnrollment for the ES index
+    """
+    @staticmethod
+    def calculate_certificate_grade_average(certificates):
+        """
+        Calculates an average grade (integer) from a list of certificates with <1 decimal grades
+        """
+        return None if len(certificates) == 0 else round(
+            (sum([Decimal(certificate['grade']) for certificate in certificates]) / len(certificates)) * 100
```
noisecapella: Can you change the square brackets in `[Decimal(..) for cert in certificates]` to parentheses, since `sum` will evaluate on all iterables?

dashboard/serializers.py
```diff
@@ -0,0 +1,52 @@
...
+class UserProgramSerializer:
+    """
+    Provides functions for serializing a ProgramEnrollment for the ES index
+    """
+    @staticmethod
+    def calculate_certificate_grade_average(certificates):
+        """
+        Calculates an average grade (integer) from a list of certificates with <1 decimal grades
+        """
+        return None if len(certificates) == 0 else round(
```
noisecapella: I think we should keep the average as a decimal value like the source data is. We should only convert it to a percent format when we're about to display the value

dashboard/serializers_test.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    @staticmethod
+    def _generate_cached_enrollments(user, program, num_course_runs=1, data=None):
+        """
+        Helper method to generate CachedEnrollments for test cases
+        """
+        course = CourseFactory.create(program=program)
+        course_runs = [CourseRunFactory.create(course=course) for _ in range(num_course_runs)]
+        factory_kwargs = dict(user=user)
+        if data:
```
noisecapella: Can you change this to `if data is not None`?

dashboard/serializers.py
gsidebo: 1. `related_edx_obj_set` is kind of awkward... would something like this suffice?
`related_edx_obj_set (Many-to-one RelatedManager): QuerySet for cached edX model objects (eg: CachedEnrollment)

2. i think `program: Program object` should either stay how it is, or become `program (Program):` with nothing after the colon. anything more would be redundant and i don't see the use

3. this just returns a list of dictionaries, really. can you link me to that edx style guide again? i don't really know the best way to express that

noisecapella: The types within the parentheses are read by PyCharm and possibly other editors, so it's useful to include them even if they read as redundant in English

The edX guide: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#docstrings

I'm fine with just

    Returns:
        list of dict: List of serialized objects

There isn't really a standard for generic types in sphinx-napoleon docstrings AFAIK so it seems like the best we can do here. For 1 I'm fine with `RelatedManager` there, it would let the editors infer certain manager functions at least

dashboard/serializers.py
gsidebo: i don't think we can do this yet b/c of the elasticsearch histogram limitation (can't use decimal values/ranges)

noisecapella: Right I forgot about that. Feel free to leave it as is

noisecapella:
:+1:

Issue #795, opened 2016-08-04
Update styling for user profile view


pdpinch:
For example, the URL at /users/pdpinch should look like

![user_profile 2x](https://cloud.githubusercontent.com/assets/20047260/17299343/fd1e1b6c-57db-11e6-87e9-c60d619f3179.jpg)

instead of 

![/users/pdpinch](https://box.everhelper.me/attachment/540634/5b9e34bf-3230-4d23-90b2-5cf4f94c3cb7/623857-1KnJn2tsWZStP14w/screen.png)

pdpinch:
@amir-qayyum-khan if you get a chance, let us know if there's a branch for this.

amir-qayyum-khan:
I did some work into it but not  complete https://github.com/mitodl/micromasters/pull/864

amir-qayyum-khan:
@roberthouse54 @pdpinch 

- I need an email icon just like in screen shot with email.
- I need a Camera Icon, just like in screenshot above user image.
- What about the ``About section``. Do i need to add it or take it from some where?

<img width="652" alt="screen shot 2016-08-23 at 6 37 31 pm" src="https://cloud.githubusercontent.com/assets/10431250/17894026/4ab0675e-6961-11e6-9f5e-84470fa0e68b.png">

pdpinch:
Skip the about for now. That will be another issue (probably after our 10/15 release)

amir-qayyum-khan:
OK,

change in design, removed border line for last row.

<img width="652" alt="screen shot 2016-08-23 at 6 37 31 pm" src="https://cloud.githubusercontent.com/assets/10431250/17894772/02f548d2-6964-11e6-9195-29d0aacb79c5.png">

Ferdi:
What does "Edit" do ?

amir-qayyum-khan:
@Ferdi I asked the same question from @roberthouse54  https://github.com/mitodl/micromasters/issues/795#issuecomment-241734593, From mockups it seems they disable row wise editing and enabled editing for whole section i.e education or work history.

roberthouse54:
Here is a revised design with inline editing.
![profile 2x](https://cloud.githubusercontent.com/assets/20047260/17941451/5ec0bf78-6a01-11e6-9f15-28328beb3bb0.png)

roberthouse54:
[...]

PR #864, opened 2016-08-16
Style changes for user page

amir-qayyum-khan:
https://github.com/mitodl/micromasters/issues/795

#### What's this PR do?
- Changed the design of user page

screen shot 2016-09-05 at 8 14 04 pm


static/scss/profile.scss
```diff
@@ -50,6 +69,9 @@
   .profile-form-grid {
     width: 100%;
 
+    &.profile-form-grid-tranparent {
```
noisecapella: Should be `transparent`

static/js/components/ProfileImage.js
```diff
@@ -11,14 +11,34 @@ export default class ProfileImage extends React.Component {
...
       />
     );
+    if (editable) {
+      img = (
+        <div className="avatar">
+          <img
+            src={imageUrl}
+            alt={`Profile image for ${getPreferredName(profile, false)}`}
+            className="card-image"
+          />
```
aliceriot: there's a component called `ProfileImage` you can use here instead

aliceriot: oh gosh I need to read better >_< disregard that comment haha

amir-qayyum-khan: :)

aliceriot:
in the mockup there are lines between education entries, this is what happens when I add two education entries on your branch:

lines_education_1

aliceriot:
Can you use the new layout classes we're using on the `profile` and `dashboard` pages? Basically, you should be wrapping the content you want to size with a `<div class="single-column">`.

aliceriot:
the spacing and overall layout doesn't really match the mockup very closely:

spacing

vs

mockup

aliceriot:
email icon is way too big

email_icon

profiles/models.py
```diff
@@ -166,6 +166,11 @@ def pretty_printed_student_id(self):
         """pretty prints the student id for easy display"""
         return "MMM{0:06}".format(self.student_id) if self.student_id else ""
 
+    @property
+    def email(self):
+        """pretty prints the student id for easy display"""
```
aliceriot: This looks like the same comment as the method above.

aliceriot:
Why did you need to check in those icons? Did you take a look at Google's [material icons](https://design.google.com/icons/)?

we use them in several other places in the app, and I think they're preferable over checking in a PNG.

static/js/components/UserInfoCard.js
```diff
@@ -0,0 +1,57 @@
...
+
+  getCurrentWorkPlace: Function = (profile: Profile): string => {
+    let company_name = "";
+    for (let work_history of profile.work_history) {
+      if (work_history.end_date === null) {
+        company_name = work_history.company_name;
+        break;
+      }
+    }
+    return company_name;
```
aliceriot: there's a `getEmployer` function in `util/util.js`

aliceriot:
OK I looked through the code quickly, but I think the UI still needs a fair amount of work before it makes sense to read the code.

Overall though, try to write shorter class names and reuse patterns you see us using elsewhere on the app, if possible - for instance, this page is very similar to the profile page, so you should take a close look at how that is implemented in JS and styled, and try to re-use as much code as possible.

For instance, putting a class name like `"basic-info basic-info-full-width basic-info-border-bottom"` is really verbose. Consider instead defining classes like this:

```css
basic-info {
  // set fonts and so on for that component
}

full-width {
  width: 100%
}

border-bottom {
  // set whatever you want for a border
}
```

then on your actual react element you can put `className = "basic-info full-width border-bottom"`, and if you need to write styles specific to combinations of those selectors you can do that in css. Does that make sense? I'm not saying this is exactly how you should refactor this, but I think this is the overall pattern we want to move towards.

Take a look in `page-layout.scss`. We define a bunch of classes in there that can be used to lay out a whole page, either in a single column or two column layout. Then we can share that css across all of our pages, instead of having a big pile of css for each page that inconsistently styles them.

aliceriot:
Feel free to ping me with any questions you have as you're working through these questions.

static/js/util/util.js
```diff
@@ -238,7 +238,7 @@ export function makeStrippedHtml(textOrElement: any): string {
 }
 
 export function makeProfileImageUrl(profile: Profile): string {
-  let imageUrl = urljoin(SETTINGS.edx_base_url, 'static/images/profiles/default_120.png');
+  let imageUrl = '/static/images/default_120.png';
```
aliceriot: what's the reason for this change?

static/scss/profile.scss
```diff
@@ -58,16 +112,55 @@
...
+      float: right;
+      display: inline;
+      > a {
+        color: #1B7AE3;
+        font-size: 19px;
+        font-weight: 400;
+        cursor: pointer
       }
     }
   }
```
aliceriot: If you write css that's specific to the user page it should live in a separate file.

aliceriot: (e.g. `user_page.scss` or something along those lines)

amir-qayyum-khan:
Rob has given me these icon,  but if you think i should use other
metrilistic,  let me search

static/js/util/util.js
amir-qayyum-khan: @roberthouse54  provided me new icon

amir-qayyum-khan: @roberthouse54 provided me new default image

amir-qayyum-khan:
Done with changes @aliceriot.

screen shot 2016-08-29 at 5 34 52 pm 2

aliceriot:
@amir-qayyum-khan great, I'll take a look shortly!

aliceriot:
looks like a bunch of `UserPage` tests are broken

aliceriot:
spacing on the profile is still off:

profile_spacing_2

aliceriot:
it's too wide, the titles of the education and employment cards are too close to the top, and the margin on the top of the user name / image card is too big

aliceriot:
Couple other things:

1. can you please use the `single-column` class for the outer-most `div` on the user page? This will give you the appropriate width automatically.

2. The lines on the mockup only appear between education and work history entries, not after the last entry.

3. please move css that is specific to the user page into a separate file, and scope it to a top-level selector for all content on that page (e.g. `.user-page` or `#user-page`).

static/js/components/UserInfoCard.js
```diff
@@ -0,0 +1,53 @@
...
+            {userPrivilegeCheck(profile, () => <IconButton name="edit" onClick={toggleShowPersonalDialog}/>)}
+          </Cell>
+          <Cell col={4}>
+            <ProfileImage profile={profile} editable={true} />
+          </Cell>
+          <Cell col={8}>
+            <div className="profile-title">{getPreferredName(profile)}</div>
+            <div className="profile-company-name">{this.getCurrentWorkPlace(profile)}</div>
+            <div className="profile-email">
+              <img src='/static/images/email.png' alt="Email image" className="email-icon"/>
```
aliceriot: Instead of these `.png`'s please use a material design icon. You can look at them here: https://design.google.com/icons/

find the one you want (probably called `email` or something, not sure), and then do the following where you want to place the icon:

```js
import Icon from 'react-mdl/lib/Icon';

render () {
  ...
  <Icon name="email" />
  ...
}
```

let me know if that's not clear.

aliceriot: You can look at `UserChip.js` for an example.

static/js/components/UserInfoCard.js
```diff
@@ -0,0 +1,53 @@
...
+    toggleShowPersonalDialog: () => void
+  };
+
+  getCurrentWorkPlace: Function = (profile: Profile): string => {
+    let companyName = "";
+    if (!getEmployer(profile).isNothing) {
+      companyName = getEmployer(profile).value;
+    }
+    return companyName;
+  }
```
aliceriot: There's a utility function you should use here instead, `getEmployer`, from `util.js`.

aliceriot: Oh wait sorry, I didn't read closely enough >_<

aliceriot: But this function isn't necessary anyway, `getEmployer` returns a `Maybe<string>`. There's a utility function in `util/sanctuary` called `mstr` that replicates this:

```js
const mstr = S.maybe("", String);
```

Basically, if you have a `Maybe` you can pass it to `mstr` and if the maybe is `Nothing` you'll get `""`, and otherwise you'll get `String(val)`, where `val` is the value wrapped up in the `Maybe`.

Here's an example:

```js
import { Just, Nothing } from 'sanctuary';
import { mstr } from '../util/sanctuary';

const justName = Just("Alice");

const nothing = Nothing();

mstr(justName) // -> "Alice"
mstr(nothing)  // -> ""
```

So you can call `mstr(getEmployer(profile))` inline in the JSX, and that should take care of everything!

static/js/components/ProfileImage.js
```diff
@@ -7,18 +7,39 @@ import type { Profile } from '../flow/profileTypes';
...
+            <img
+              src='/static/images/camera.png'
+              alt="Camera image for"
+            />
+          </span>
+        </div>
+      );
+    }
+
+    return img;
```
aliceriot: What do you think about refactoring this a little? Instead of declaring the `img` item twice you could write a method that checks for `editable` and, if it's set to `true`, returns the camera icon (which, again, should be an MDL `Icon`). Then you can just call that method inline.

static/js/components/User.js
```diff
@@ -34,43 +29,13 @@ export default class User extends React.Component {
...
-    let getStateName = () => {
-      if ( profile.state_or_territory ) {
-        return iso3166.subdivision(profile.state_or_territory).name;
-      } else {
-        return '';
-      }
-    };
-
-    return <div className="card">
+    return <div>
```
aliceriot: please put `className="single-column"` on either this div or a wrapping `div` in `UserPage`. You should be able to delete most of the classes you have (like `profile-form-center-card`) when you do that.

static/js/components/User.js
```diff
@@ -34,43 +29,13 @@ export default class User extends React.Component {
...
-
-      <Grid className="user-cards-grid">
-        <Cell col={6}>
-          <EmploymentForm {...this.props} showSwitch={false} validator={employmentValidation} />
-        </Cell>
-        <Cell col={6}>
-          <EducationDisplay {...this.props} />
-        </Cell>
-      </Grid>
+      <div className="card-copy">
```
aliceriot: We're trying to get rid of the `.card-copy` class, please delete it.

aliceriot:
There's some issues with what you've got here still - let me know if you have any questions.

static/js/components/ProfileImage.js
```diff
@@ -7,18 +7,39 @@ import type { Profile } from '../flow/profileTypes';
...
+        return (
+          <span className="img">
+            <img
+              src='/static/images/camera.png'
+              alt="Camera image for profile image"
+            />
+          </span>
+        );
+      }
+    };
```
aliceriot: I would move this function to a module level or class-level function / method. As is, the function will be redeclared on every re-render. sometimes we need to generate functions based on runtime values, so that makes sense to do, but in this case you should be able to write a function like this:

```js
const cameraIcon =  editable => {
  if ( editable ) {
    return <img
      src='/static/images/camera.png'
      alt="Camera image for profile image"
    />;
  }
};
```

Then in the `render` method you would just pull `editable` out of `this.props` and call `cameraIcon(editable)` inline in the JSX.

static/scss/profile.scss
```diff
@@ -58,16 +73,44 @@
...
+        }
+      }
       .school-type {
-        font-size: 12px;
-        font-weight: 400;
-        text-transform: lowercase;
+        margin-bottom: -10px;
+        font-size: 16px;
+        font-weight: 600;
+        text-transform: capitalize;
```
aliceriot: this changes the styling on the `profile` page in a perhaps unintended way.

static/scss/profile.scss
```diff
@@ -58,16 +73,44 @@
...
+      .profile-row-new-line {
+        display: block;
+        width: 100%;
       }
     }
 
     .profile-card-header {
       > span {
-        font-size: 18px;
+        font-size: 22px;
```
aliceriot: can you leave this as-is? it changes the styling of the profile during the signup flow.

aliceriot: I think it should be 18px on the userpage anyway, that makes it look closer to the mockup

aliceriot:
spacing on the profile still is not correct:

profile_head

aliceriot:
for that element I think you should do something like this:

```js
<div className="profile-form-row profile-card-header">
  <ProfileImage ... />
  <div className="user-info">
    ... // show the name and so on in here
    <EditButton />
  </div>
</div>
```

`profile-form-row` already (on master) has `display: flex` and `flex-direction: row` so that should correctly lay out it's children. Then basically all you need to do is correctly position the elements in `.user-info`, and all the css you add for that (for getting the name and so on right) will be scoped to just this one element, and won't mess with the styling of the profile pages either. You should be able to get it with something like this:


```css
.user-info {
  display: flex;
  flex-direction: column;
  
  .edit-button {
    // do some absolute positioning to get this in the right spot
  }
}
```

does that make sense? I'd like to try to re-use as many of the existing classes and CSS as possible, and it would be good to try to avoid using `Cell` elements for layout (we're sort of moving away from this).

static/js/components/ProfileImage.js
aliceriot: Also did you try looking for material design icons that would work for this? they have a couple that look similar to what Rob has in the mockups.

aliceriot:
can you remove `edit.png`, since we're not using it?

static/js/components/EducationDisplay.js
```diff
@@ -117,7 +119,7 @@ export default class EducationDisplay extends ProfileFormFields {
         <Card shadow={1} className="profile-form" id="education-card">
           <Grid className="profile-form-grid">
             <Cell col={12} className="profile-form-row profile-card-header">
-              <span>
+              <span className="profile-form-card-header">
```
aliceriot: why not just change this to `.title`? then you can style it by doing:

```css
.profile-card-header {
  .title {
    font-weight: 600;
  }
}
```

or any other styles you want to add.

static/js/components/ProfileImage.js
aliceriot: with material-design the above function would look something like:

```js
const cameraIcon = editable => editable ? <Icon name="camera" /> : null;
```

assuming the name of the appropriate icon is `camera`.

aliceriot:
some comments, back to you

roberthouse54:
Also, can we use the new default avatar image I sent you?

amir-qayyum-khan:
@roberthouse54 i am using the same image

amir-qayyum-khan:
Done @aliceriot need to upload default image to https://www.gravatar.com/

screen shot 2016-08-31 at 5 24 38 pm

aliceriot:
the camera icon is different than the one used in the mockup:

camera_icon

amir-qayyum-khan:
this icon is used in materialistic design

aliceriot:
yeah, but we should use the one that Rob put in the mockup

aliceriot:
maybe this one: https://design.google.com/icons/#ic_camera_alt

aliceriot:
again, the lines after the last entry are still there: 
lines_lines_lines

aliceriot:
the icon for the camera also isn't blue in the mockup

static/js/containers/UserPage_test.js
```diff
@@ -41,7 +42,7 @@ import ValidationAlert from '../components/ValidationAlert';
 import { modifyTextField } from '../util/test_utils';
 
 describe("UserPage", function() {
-  this.timeout(5000);
+  this.timeout(12000);
```
aliceriot: oof if we need to do this to get tests to pass we should re-write the tests.

static/js/components/User.js
```diff
@@ -34,43 +29,13 @@ export default class User extends React.Component {
...
-
-      <Grid className="user-cards-grid">
-        <Cell col={6}>
-          <EmploymentForm {...this.props} showSwitch={false} validator={employmentValidation} />
-        </Cell>
-        <Cell col={6}>
-          <EducationDisplay {...this.props} />
-        </Cell>
-      </Grid>
+      <div>
```
aliceriot: what's the purpose of this div?

static/js/components/ProfileImage.js
```diff
@@ -1,24 +1,37 @@
...
+    editable: false
   };
 
   render () {
-    const { profile } = this.props;
+    const { profile, editable } = this.props;
     const imageUrl = makeProfileImageUrl(profile);
+    const cameraIcon = editable => (
+      editable ? <span className="img"><Icon name="camera" /></span> : null
+    );
```
aliceriot: can you move the declaration of this function out of the render method? it's unnecessary to re-declare it every time.

static/js/components/EducationDisplay.js
```diff
@@ -61,18 +61,20 @@ export default class EducationDisplay extends ProfileFormFields {
...
           <div className="profile-row-name">
-            <span className="school-type">
+            <div className="school-type">
               { degree }
-            </span><br/>
+            </div><br/>
             { entry.school_name }
           </div>
+        </div>
+        <div className="basic-info full-width border-bottom education-top-spacing">
```
aliceriot: I think here you should just put another class on this div to indicate it's a userpage row (like `.user-page-row` or something).

then all the styling for adding the border at the bottom and doing layout can be just on that selector.

aliceriot: Basically overall I think this should look more or less like this:

```js
<Cell col={12} className="profile-form-row user-page-row" key={index}>
  <div className="basic-info education">
    <span className="school-type">{ degree }</span>
    <span className="school-name">{ entry.school_name }</span>
  </div>
  <div className="basic-info">
    { userPrivilegeCheck(profile, icons, () => <Cell col={2} />) }
  </div>
</Cell>
```

then for the layout particular to the user page all you need to do is write some CSS targeting `.profile-form-row.user-page-row`. You can do something like this:

```css
.profile-form-row.user-page-row {
  > div {
    width: 50%;
  }
  
  .basic-info.education {
    // education specific styling
  }
  
  .basic-info.work_history {
    // any work-history specific styling
  }
}
```

does that make sense? what you have here is I think just more markup than is necessary.

static/js/components/EducationDisplay.js
```diff
@@ -117,7 +119,7 @@ export default class EducationDisplay extends ProfileFormFields {
         <Card shadow={1} className="profile-form" id="education-card">
           <Grid className="profile-form-grid">
             <Cell col={12} className="profile-form-row profile-card-header">
-              <span>
+              <span className="profile-form-card-header title">
```
aliceriot: can you change this to just `.title`?

static/js/components/EmploymentForm.js
```diff
@@ -318,7 +320,7 @@ class EmploymentForm extends ProfileFormFields {
         <Card shadow={1} className={`profile-form ${cardClass()}`}>
           <Grid className="profile-form-grid">
             <Cell col={12} className="profile-form-row profile-card-header">
-              <span>
+              <span className="profile-form-card-header title">
                 Employment
               </span>
               { workSwitch() }
```
aliceriot: similar comments to the ones about education above

static/js/components/ProfileImage.js
aliceriot: also can you remove the `<span>` tag rapping the `Icon`? If you want to style the icon it would be better to just use a selector that targets it, i.e. you could put `.camera-icon` on it or something.

static/js/components/UserInfoCard.js
```diff
@@ -0,0 +1,49 @@
...
+              <Icon name="email" className="email-icon" />
+              <span className="profile-email">{profile.email}</span>
+            </div>
+          </Cell>
+          <Cell col={1}>
+            <div className="edit-profile-holder">
+              {userPrivilegeCheck(profile, () => <IconButton name="edit" onClick={toggleShowPersonalDialog}/>)}
+            </div>
+          </Cell>
+        </Grid>
```
aliceriot: What do you think about moving away from using a lot of `Cell` elements here? I think you could simplify this by putting a class on the `Card` that does `display: flex; flex-direction: row`, and then wrap the middle elements (currently in `Cell col={8}`) in a div that has `display: flex; flex-direction: column;` on it.

aliceriot:
comments and questions, back to you

amir-qayyum-khan:
Done @aliceriot

aliceriot:
cool, taking a look now

aliceriot:
I believe this is the most recent mockup: https://projects.invisionapp.com/share/9B804AMXA#/screens/183655916

( @roberthouse54 can you confirm that? )

on the mockup education is above employment.

aliceriot:
other than that things look generally good! taking a look at the code.

static/js/components/EducationDisplay.js
```diff
@@ -61,18 +61,18 @@ export default class EducationDisplay extends ProfileFormFields {
...
           <div className="profile-row-name">
-            <span className="school-type">
-              { degree }
-            </span><br/>
-            { entry.school_name }
+            <div className="school-type">{ degree }</div>
+            <div className="school-name">{ entry.school_name }</div>
           </div>
+        </div>
+        <div className="profile-form-col education-top-spacing">
```
aliceriot: what do you think about renaming `.profile-form-col` to just `.col`?

aliceriot: also maybe `.education-top-spacing` -> `.education` and `.employment-top-spacing` -> `.employment`

static/js/components/EducationDisplay.js
```diff
@@ -86,7 +86,7 @@ export default class EducationDisplay extends ProfileFormFields {
     }
     userPrivilegeCheck(profile, () => {
       rows.push(
-        <Cell col={12} className="profile-form-row" key={"I'm unique!"}>
+        <Cell col={12} className="profile-form-add-row" key={"I'm unique!"}>
```
aliceriot: can you change this className to be `.profile-form-row.add` or `.profile-form-row.add-item`? you should then be able to apply the styling you want by doing:

```scss
.profile-form-row {
  // styles
  &.add {
    // styles
  }
}
```

instead of adding another class that receives the same styles as `.profile-form-row`

static/js/components/EmploymentForm.js
```diff
@@ -236,15 +236,17 @@ class EmploymentForm extends ProfileFormFields {
...
+        <div className="profile-form-col">
           <div className="profile-row-name">
             {`${position.company_name}, ${position.position}`}
           </div>
+        </div>
+        <div className="profile-form-col employment-top-spacing">
           <div className="profile-row-date-range">
             {`${dateFormat(position.start_date)} - ${endDateText()}`}
           </div>
+          { userPrivilegeCheck(profile, icons, () => <Cell col={2} />) }
```
aliceriot: same comments apply here that I left about the education component above

static/js/components/ProfileImage.js
```diff
@@ -7,18 +7,41 @@ import type { Profile } from '../flow/profileTypes';
...
+  }
+
   render () {
-    const { profile } = this.props;
+    const { profile, editable } = this.props;
     const imageUrl = makeProfileImageUrl(profile);
+    const cameraIcon = editable => (
+      editable ? <span className="img">{this.cameraSvgImage()}</span> : null
+    );
```
aliceriot: can you move this function definition out of the render method? it doesn't have any runtime dependencies, there's no reason to re-define it on every re-render.

static/js/components/ProfileImage.js
```diff
@@ -7,18 +7,41 @@ import type { Profile } from '../flow/profileTypes';
...
+  cameraSvgImage: Function = (): React$Element<*> => {
+    return (
+      <svg fill="#FFFFFF" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
+        <circle cx="12" cy="12" r="3.2"/>
+        <path d={`M9 2L7.17 4H4c-1.1 0-2 .9-2 2v12c0 1.1.9 2 2 2h16c1.1 0 2-.9 
+          2-2V6c0-1.1-.9-2-2-2h-3.17L15 2H9zm3 15c-2.76 0-5-2.24-5-5s2.24-5 5-5 5 2.24 5 5-2.24 5-5 5z`}/>
+        <path d="M0 0h24v24H0z" fill="none"/>
+      </svg>
+    );
+  }
```
aliceriot: were you unable to get the standard `<Icon>` component to do what you needed?

aliceriot: just wondering about the inline `svg`

aliceriot: also style nit: a JS arrow function that returns on item can surround the return value with parentheses, which is a little more compact. This works for multi-line JSX statements, so you could do something like:

```js
cameraImage = () => (
  <svg>
    ...
  </svg>
);
```

just avoids two extra lines

static/js/components/UserInfoCard.js
```diff
@@ -0,0 +1,44 @@
...
+    profile: Profile,
+    toggleShowPersonalDialog: () => void
+  };
+
+  render() {
+    const { profile, toggleShowPersonalDialog } = this.props;
+
+    return (
+      <Card shadow={1} className="profile-form user-page">
+        <div className="user-form-grid">
```
aliceriot: it seems like you should be able to re-use `.profile-form-row` here, and, if you refactor the column class along the lines I suggested above, you should also be able to re-use the `.col` class, then you could make the class for this `.profile-form-row.user-card`, and use that selector to override any styles you needed to on the user card (e.g. for font sizes)

static/scss/profile.scss
```diff
@@ -50,26 +57,66 @@
     width: 100%;
     padding: 0;
 
-    .profile-form-row, .profile-form-row > .basic-info {
+    .title {
+      font-weight: 600;
+      margin-top: 15px;
+    }
+    .profile-form-row, .profile-form-add-row, .profile-form-row > .profile-form-col {
```
aliceriot: see my comments above, but let's try to change this to just:

```scss
.profile-form-row, .profile-form-row > .col {
 //
}
```

static/scss/user_page.scss
```diff
@@ -0,0 +1,56 @@
...
+    display: flex;
+    flex-direction: row;
+    .column {
+      display: flex;
+      flex-direction: column;
+      width: 250% !important;
+      margin-left: 15px;
+    }
+  }
+}
```
aliceriot: general scss style stuff:

we generally try to leave a blank line after closing a block and before opening a new one. So instead of

```scss
.class {
  display: none;
  .other-class {
    color: purple;
  }
  .third-class {
    text-decoration: underline;
  }
}
```

we generally prefer:

```scss
.class {
  display: none;
  
  .other-class {
    color: purple;
  }
  
  .third-class {
    text-decoration: underline;
  }
}
```

when closing multiple blocks (like in the example when we close the `.third-class` and `.class` blocks) the curly braces don't need a blank line in between.

check out `page-layout.scss` and `user-chip.scss` for examples of this formatting.

aliceriot:
@amir-qayyum-khan back to you! some questions about css details and tweaks to the profile image changes.

static/js/components/EmploymentForm.js
amir-qayyum-khan: @aliceriot are you talking abt .col and .employment

aliceriot: yeah exactly

aliceriot: and `.add` above

static/js/components/ProfileImage.js
amir-qayyum-khan: @aliceriot I was not able to get white Icon (camer-alt), It was giving me blue, may be because of theme settings?

aliceriot: we set `i.material-icons` to have `color: $mm-brand-blue`. you should be able to override that for a particular icon by using a more specific selector (like `i.material-icons.user-info-edit` or something). In that case I think your `cameraIcon` function would directly return the `<Icon>` component.

amir-qayyum-khan:
Done @aliceriot

static/js/components/ProfileImage.js
```diff
@@ -1,24 +1,38 @@
...
+    profile: Profile,
+    editable: boolean
   };
 
+  static defaultProps = {
+    editable: false
+  };
+
+  cameraIcon: Function = (editable: bool): React$Element<*>|null => {
+    return editable ? <span className="img"><Icon name="camera_alt" /></span> : null;
```
aliceriot: can you just put a className on the `Icon` and use that to position it?

static/scss/profile.scss
```diff
@@ -50,26 +57,77 @@
...
+      &:nth-last-child(2) {
+        .col {
+          border-bottom: none !important;
+        }
+      }
+    }
+
+    &.add {
+      margin-bottom: 17px;
+    }
```
aliceriot: should this be inside `.profile-for-row`?

static/js/components/ProfileImage.js
amir-qayyum-khan: @aliceriot span is a transparent circle, inside which we have icon.

aliceriot: ohh gotcha, that makes sense then.

static/js/components/UserInfoCard.js
aliceriot: any thoughts on this @amir-qayyum-khan ?

aliceriot:
code's looking good, and the user page is good to go I think.

the one issue is that it seems something has been messed up on the profile sign up flow:

profile_regression

profile_regression_employ

do you think you can figure out what changed, and then scope your changes to the user page more?

aliceriot:
looks like a bunch of the styling for `.profile-card-header` was deleted and that's probably the culprit!

amir-qayyum-khan:
yeah, i am looking into this,

aliceriot:
I think a quick fix is putting:

```scss
justify-content: space-between;
```

on `.profile-card-header` and bumping the font size to `18px` for `.profile-card-header > span`.

amir-qayyum-khan:
Fixed @aliceriot

static/scss/profile.scss
```diff
@@ -32,6 +32,9 @@
     flex-direction: row;
     color: $lightgrey !important;
     text-transform: uppercase;
+    position: absolute;
+    right: 0;
+    padding-right: 50px;
```
aliceriot: if you get rid of these three lines the layout seems to be the same.

with:

with_lines

without:

without_lines

let's delete them

aliceriot: Can you set the font for `.profile-card-header > span` to be 18px or put `.title` on the `<span>` tags in `.profile-card-header` on the education and work history elements?

static/scss/profile.scss
```diff
@@ -63,6 +66,7 @@
     }
 
     .profile-card-header {
+      justify-content: space-between;
       .title {
```
aliceriot: you could change this line to `.title, > span`, for instance

aliceriot: (also plz leave a blank line above this line)

static/scss/profile.scss
```diff
@@ -89,6 +100,18 @@
...
+      text-align: right;
+      color: #898989;
+
+      &.alignment {
+        float: right;
+        position: absolute;
+        right: 0px;
+        padding-right: 37px;
+      }
+    }
```
aliceriot: here I think you should be able to just set:

```scss
.profile-row-icons {
  display: flex;
  flex-direction: row;
  justify-content: flex-end;
  width: 50%;
}
```

something along those lines.

amir-qayyum-khan:
Done @aliceriot

static/scss/profile.scss
```diff
@@ -32,6 +32,9 @@
     flex-direction: row;
     color: $lightgrey !important;
     text-transform: uppercase;
+    position: absolute;
+    right: 0;
+    padding-right: 41px;
```
aliceriot: again, I don't think these three lines are necessary, I can delete them and the layout is the same

aliceriot:
one comment about three lines I don't think are necessary, other than that this looks good.

static/scss/profile.scss
aliceriot: @amir-qayyum-khan do you have a reason to keep them around? if not let's delete.

aliceriot:
:+1: 🚢 :shipit: 🎆 🎊

PR #871, opened 2016-08-16
Added UserChip to search results

### LINKED ISSUE #757: Implement user chip in search results
When the user hovers over a search result, the "user chip" appears to reveal more information about the learner found in the result, and to present links to the learner's profile. 

For now, we will only implement the "View Profile" button on the user chip. Omit the user's e-mail address and the "Send Email" button that are in the design. 

From https://invis.io/9B804AMXA#/172058723_Instructor_User_Search : 

instructor_user_search 2x

- [x] @roberthouse54 please confirm this description

aliceriot:

closes #757 

#### What's this PR do?

This adds a little pop-up thingy that shows up when you hover over a result in the search.

Take a look over the code! If you have ideas about how to manage the CSS cascade better (so we can remove some `!important` styles) that would be :100:.

Go to `/learners`. Hovering over any result should cause a little popup to display.

### DIFF
--- a/static/js/components/UserChip.js
+++ b/static/js/components/UserChip.js
@@ -1,0 +1,29 @@
+// @flow
+import React from 'react';
+import { Card } from 'react-mdl/lib/Card';
+import Icon from 'react-mdl/lib/Icon';
+
+import { getPreferredName, getEmployer } from '../util/util';
+import ProfileImage from './ProfileImage';
+import { mstr } from '../util/sanctuary';
+import type { Profile } from '../flow/profileTypes';
+
+const UserChip = ({ profile }: {profile: Profile}): React$Element<*> => (
+  <Card className="user-chip" shadow={2}>
+    <div className="profile-info">
+      <span className="name">
+        { getPreferredName(profile) }
+      </span>
+      <span className="employer">
+        { mstr(getEmployer(profile)) }
+      </span>
+      <a href={`/users/${profile.username}`} className="mm-minor-action">
+        <Icon name="person" />
+        <span>View profile</span>
+      </a>
+      </div>
+    <ProfileImage profile={profile} />
+  </Card>
+);
+
+export default UserChip;
--- a/static/js/components/UserChip_test.js
+++ b/static/js/components/UserChip_test.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { assert } from 'chai';
+
+import { USER_PROFILE_RESPONSE } from '../constants';
+import R from 'ramda';
+import UserChip from './UserChip';
+import { getPreferredName } from '../util/util';
+import ProfileImage from './ProfileImage';
+
+describe('UserChip', () => {
+  let clone;
+
+  const renderChip = profile => (
+    shallow(<UserChip profile={profile} />)
+  );
+
+  beforeEach(() => {
+    clone = R.clone(USER_PROFILE_RESPONSE);
+  });
+
+  it('should show the preferred name', () => {
+    let chip = renderChip(clone);
+    assert.equal(chip.find('.name').text(), getPreferredName(clone));
+  });
+
+  it('should show the employer, if present', () => {
+    let chip = renderChip(clone);
+    assert.equal(chip.find('.employer').text(), "Planet Express");
+  });
+
+  it('should leave employer blank, if absent', () => {
+    clone.work_history = [];
+    let chip = renderChip(clone);
+    assert.equal(chip.find('.employer').text(), "");
+  });
+
+  it('should link to the profile', () => {
+    let chip = renderChip(clone);
+    let url = chip.find('a').props().href;
+    assert.equal(url, `/users/${clone.username}`);
+  });
+
+  it('should include the profile image', () => {
+    let chip = renderChip(clone);
+    assert(
+      chip.containsMatchingElement(<ProfileImage />),
+      "chip should contain a ProfileImage component"
+    );
+  });
+});
--- a/static/js/util/sanctuary_test.js
+++ b/static/js/util/sanctuary_test.js
@@ -4,10 +4,13 @@ import { assert } from 'chai';
 
 import { allJust, mstr } from './sanctuary';
 
-describe('sanctuary util functions', () => {
-  let maybeEquality = (j1, j2) => assert(j1.equals(j2), "just equality");
+export const assertMaybeEquality = (m1: Maybe, m2: Maybe) => {
+  assert(m1.equals(m2), "Maybe's should be equal");
+};
 
-  let isNothing = m => assert(m.isNothing, "should be nothing");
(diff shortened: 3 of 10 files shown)


static/js/components/UserChip.js
```diff
@@ -0,0 +1,29 @@
...
+  <Card className="user-chip" shadow={2}>
+    <div className="profile-info">
+      <span className="name">
+        { getPreferredName(profile) }
+      </span>
+      <span className="employer">
+        { mstr(getEmployer(profile)) }
+      </span>
+      <a href={`/users/${profile.username}`} className="mm-minor-action">
+        <Icon name="person" />
```
gsidebo: when you hover over this link, an underline appears under this icon. looks kinda wonky. might be able to be solved in CSS?

static/js/components/UserChip_test.js
```diff
@@ -0,0 +1,47 @@
...
+
+describe('UserChip', () => {
+  let clone;
+
+  beforeEach(() => {
+    clone = R.clone(USER_PROFILE_RESPONSE);
+  });
+
+  it('should show the preferred name', () => {
+    const chip = shallow(<UserChip profile={clone} />);
```
gsidebo: a helper method would be nice here since we're shallow rendering a chip in every single test case. something like:

```javascript
let renderChip = (profile) -> (
    shallow(<UserChip profile={profile} />)
)
```

static/js/components/UserChip_test.js
```diff
@@ -0,0 +1,47 @@
...
+import R from 'ramda';
+import UserChip from './UserChip';
+import { getPreferredName } from '../util/util';
+import ProfileImage from './ProfileImage';
+
+describe('UserChip', () => {
+  let clone;
+
+  beforeEach(() => {
+    clone = R.clone(USER_PROFILE_RESPONSE);
```
gsidebo: i think i've seen this pattern elsewhere in the codebase, but why is this called `clone`? why not `profile`?

static/js/util/util.js
```diff
@@ -256,6 +258,21 @@ export function getLocation(profile: Profile): string {
...
+  if ( _.isEmpty(entries) ) {
+    return Nothing();
+  }
+  let [, entry] = entries[0];
+  if ( entry.company_name ) {
+    return Just(entry.company_name);
+  }
+  return Nothing();
+}
+
```
gsidebo: general comment, not something that's needed for this PR: from what i can see, `static/js/util/util.js` doesn't seem like it exists to provide utility functions. it seems like the place where we just dump any functionality that we don't have a logical place for. there are a bunch of functions in here that we use to extract/calculate information about a Profile (add this one to the list). at the very least, i feel like we should create a `static/js/util/profile.js` file and move those functions into there. i would say that's still a little iffy since it's still in a 'util/' folder and it isn't really 'utility' functionality, but probably a step in the right direction. what do you think?

static/js/util/sanctuary_test.js
```diff
@@ -4,10 +4,13 @@ import { assert } from 'chai';
 
 import { allJust, mstr } from './sanctuary';
 
-describe('sanctuary util functions', () => {
-  let maybeEquality = (j1, j2) => assert(j1.equals(j2), "just equality");
+export const maybeEquality = (j1: Maybe, j2: Maybe) => {
+  assert(j1.equals(j2), "just equality");
+};
+
+export const isNothing = (m: Maybe) => assert(m.isNothing, "should be nothing");
```
gsidebo: good refactor here. the naming and messaging could be clearer, though. 

1. i think it would help to put 'assert' at the beginning of both of these functions since they exist to make an assertion. makes the test cases a bit clearer when you see the invocation of something called `assert*`
1. instead of `maybeEquality`, maybe something like `assertJustValuesEqual`? 
1. `isNothing ` --> `assertMaybeIsNothing`? 
1. instead of "just equality", maybe something like "Just values should be equal". maybe something as simple as  "should be equal" would do here

static/scss/user_chip.scss
```diff
@@ -0,0 +1,47 @@
+.user-chip {
```
gsidebo: im pretty out of practice w/ css. ill try to tinker with this locally a bit

static/js/components/UserChip.js
aliceriot: I think I fixed it? I did:

```css
a:hover > i {
  text-decoration: none;
}
```

I think it worked...

static/js/components/UserChip_test.js
aliceriot: For a lot of the redux stuff we end up modifying fields deep in an object, which led us to use the lodash `cloneDeep` method a lot, I think this is where the clone name originated.

static/js/util/util.js
aliceriot: I don't have an opinion about the `util` folder. I know you feel strongly about it, so perhaps you should open a separate issue to discuss renaming it / moving things around.

static/js/util/sanctuary_test.js
aliceriot: ah yeah I agree, I'll do `isNothing -> assertIsNothing` and `maybeEquality -> assertMaybeEquality`.

static/js/components/UserChip_test.js
```diff
@@ -0,0 +1,52 @@
...
+    chip = shallow(<UserChip profile={profile} />)
+  );
+
+  beforeEach(() => {
+    clone = R.clone(USER_PROFILE_RESPONSE);
+  });
+  afterEach(() => chip = undefined);
+
+  it('should show the preferred name', () => {
+    renderChip(clone);
```
gsidebo: i think it's preferable to define `chip` inside each of these functions (`const chip = renderChip(clone);`) instead of relying on a side effect and cleaning up that side effect in the `afterEach` method

static/js/components/UserChip.js
gsidebo: doesn't for me, but i'll try a couple things as i tinker with the css locally

gsidebo:
aside from the `renderChip` comment, everything looks good. ill look into the css a little now

static/js/components/UserChip.js
gsidebo: this works:

```
// in UserChip.js
      <a href={`/users/${profile.username}`} className="mm-minor-action">
        <Icon name="person" />
        <span>View profile</span>
      </a>

// in user_chip.scss
    a:hover {
      text-decoration: none;
    }

    a:hover > span {
      text-decoration: underline;
    }
```

just a bit hacky as it wraps the text in an otherwise-unnecessary <span>

aliceriot: that doesn't make much sense to me, did you pull down the latest commits on the branch? it was working on my machine.

aliceriot:
@gsidebo I think all comments are addressed, with the css patch.

gsidebo:
looks great! 👍

PR #878, opened 2016-08-17
Made LearnerResult avatars round

### LINKED ISSUE #855: Avatars in search results should be round instead of square
Instead of 

image

it should be:

image

aliceriot:

closes #855 

#### What's this PR do?

This makes the profile images in the results round instead of square.

### DIFF
--- a/static/scss/search_page.scss
+++ b/static/scss/search_page.scss
@@ -203,6 +203,10 @@
       font-size: 12px;
       position: relative;
 
+      img {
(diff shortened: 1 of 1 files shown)


pdpinch:
screenshot?

aliceriot:
background is black so you can see the borders of the placeholder image (otherwise it's not super visible against the background):

round

gsidebo:
looks good! 👍

PR #882, opened 2016-08-17
Added program grade to search result UI

### LINKED ISSUE #681: display average grade in search results
As an MicroMaster team member, I need a single number that represents the performance of a student across all courses in a MicroMaster. This will be used to give a quick indication of a learner's status and also help ranking and clustering learners. 

Proposal: we simply average the final grade in completed courses, per MicroMasters program. Leaners who haven't successfully completed a course get a null value, which will be a signal to the UI to not display an average grade at all. 

from [MM2 on Invision](https://invis.io/9B804AMXA#/172058723_Instructor_User_Search) (note that the label should be "Average grade" not "Overall grade"

image

gsidebo:

Closes #681 

#### What's this PR do?

Adds the program grade average to search results (when it exists). Also made a minor refactor to a few assertions in other JS test cases

static/js/components/search/LearnerResult.js

Go to the search tab and observe the grades. It will help to have users in your DB with actual grades, so the realistic user set is useful here.

w/r/t the assertion refactor: i just saw test cases that could use a more accurate assertion in the same file that i was editing, so i fixed those. it was easy enough to search for other examples in the codebase, so i fixed those too. when these test fail, it now prints out a more helpful message

ss 2016-08-17 at 17 30 12

https://66.media.tumblr.com/tumblr_megu6kysDx1rg0w3to1_500.gif

### DIFF
--- a/static/js/components/ErrorMessage_test.js
+++ b/static/js/components/ErrorMessage_test.js
@@ -59,10 +59,10 @@ describe("ErrorInfo", () => {
             }
           };
           let errorMessage = renderErrorMessage(that);
-          assert.equal(true, errorMessage.includes(String(code)));
-          assert.equal(true, errorMessage.includes(errorString));
-          assert.equal(true, errorMessage.includes(contactExpectation));
-          assert.equal(true, errorMessage.includes(message));
+          assert.include(errorMessage, String(code));
+          assert.include(errorMessage, errorString);
+          assert.include(errorMessage, contactExpectation);
+          assert.include(errorMessage, message);
         });
       });
     });
--- a/static/js/components/search/LearnerResult.js
+++ b/static/js/components/search/LearnerResult.js
@@ -1,7 +1,7 @@
 // @flow
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
-
+import _ from 'lodash';
 import ProfileImage from '../ProfileImage';
 import UserChip from '../UserChip';
 import { getPreferredName, getLocation } from '../../util/util';
@@ -12,8 +12,12 @@ export default class LearnerResult extends React.Component {
     result: { _source: SearchResult }
   };
 
+  static hasGrade = program => (
+    _.has(program, 'grade_average') && _.isNumber(program.grade_average)
+  );
+
   render () {
-    const { result: { _source: { profile } } } = this.props;
+    const { result: { _source: { profile, program } } } = this.props;
     return (
       <Grid className="search-grid learner-result">
         <Cell col={2}>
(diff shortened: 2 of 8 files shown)


static/js/components/ErrorMessage_test.js
```diff
@@ -59,10 +59,10 @@ describe("ErrorInfo", () => {
             }
```
gsidebo: (as discussed in the PR description) this is just a refactor to some existing assertions. not directly related to this PR, but it was an easy enough fix and i did the same thing in a few different files

aliceriot:
functionality looks good! taking a look at the code now.

static/js/components/search/LearnerResult.js
```diff
@@ -12,8 +12,14 @@ export default class LearnerResult extends React.Component {
...
 
+  static gradeAverageDisplay (program) {
+    return _.has(program, 'grade_average') && _.isNumber(program.grade_average) ?
+      `${program.grade_average}%` :
+      '-';
+  }
+
   render () {
-    const { result: { _source: { profile } } } = this.props;
+    const { result: { _source: { profile, program } } } = this.props;
```
aliceriot: can you add `program` to the type declaration for `this.props`? flow complains about this line otherwise.

static/js/components/search/LearnerResult.js
```diff
@@ -12,8 +12,14 @@ export default class LearnerResult extends React.Component {
     result: { _source: SearchResult }
   };
 
+  static gradeAverageDisplay (program) {
+    return _.has(program, 'grade_average') && _.isNumber(program.grade_average) ?
+      `${program.grade_average}%` :
+      '-';
```
aliceriot: I'm not sure about this method. What do you think about refactoring this along the following lines:

```js
const hasGrade = program => (
  _.has(program, 'grade_average') && _.isNumber(program.grade_average)
);
```

this could be either at the module level or an instance method. Then in the render call we'd do:

```js
<span className="percent">
  { hasGrade(program) ? `${program.grade_average}` : '-' }
</span>
```

Just an idea, we may not want the `hasGrade` function for anything else but the code is a little more modular this way.

static/js/components/ErrorMessage_test.js
```diff
@@ -59,10 +59,10 @@ describe("ErrorInfo", () => {
...
           };
           let errorMessage = renderErrorMessage(that);
-          assert.equal(true, errorMessage.includes(String(code)));
-          assert.equal(true, errorMessage.includes(errorString));
-          assert.equal(true, errorMessage.includes(contactExpectation));
-          assert.equal(true, errorMessage.includes(message));
+          assert.include(errorMessage, String(code));
+          assert.include(errorMessage, errorString);
+          assert.include(errorMessage, contactExpectation);
+          assert.include(errorMessage, message);
```
aliceriot: this is much cleaner, thanks for spotting it!

aliceriot:
Generally looks good, just a couple questions.

static/js/components/search/LearnerResult.js
gsidebo: i like it. will do

static/js/components/search/LearnerResult.js
gsidebo: will do

aliceriot:
LGTM :+1:

PR #883, opened 2016-08-18
Fixed program and social username creation in realistic search data g…

### LINKED ISSUE #879: Update realistic user generation command for programs & social username
Two updates are needed in the gen_realistic_search_data command:

1. Fake programs should be set to live=True
1. A social username should be created for each fake user (ie: we need to make sure that profiles.api.get_social_username returns a value for these users)

gsidebo:

Closes #879 

#### What's this PR do?

_(copied from issue)_
Two updates are needed in the gen_realistic_search_data command:

1. Fake programs should be set to live=True
2. A social username should be created for each fake user (ie: we need to make sure that profiles.api.get_social_username returns a value for these users)

I also added an argument that can be used to assign the 'staff' role to a user of a given username

Run `delete_realistic_search_data`, then `gen_realistic_search_data`

- Check the index to see that users have a `username` value in their profile
- Check that the programs created have live=True

### DIFF
--- a/profiles/management/commands/gen_realistic_search_data.py
+++ b/profiles/management/commands/gen_realistic_search_data.py
@@ -7,7 +7,11 @@ from django.contrib.auth.models import User
 from profiles.models import Employment, Education
 from courses.models import Program, Course, CourseRun
 from dashboard.models import CachedCertificate, CachedEnrollment
+from roles.models import Role
+from roles.roles import Staff
 from micromasters.utils import load_json_from_file
+from backends.edxorg import EdxOrgOAuth2
+from search.api import recreate_index
 
 
 USER_DATA_PATH = 'profiles/management/realistic_user_data.json'
@@ -156,6 +160,11 @@ def deserialize_user_data(user_data, course_runs):
         )
     )
     user = deserialize_model_data(User, user_model_data)
+    # Create social username
+    user.social_auth.create(
+        provider=EdxOrgOAuth2.name,
+        uid=user.username,
+    )
     # Create new cached edX data records for each type we care about and associate them with a User and CourseRun
     for cached_model_deserializer in CACHED_MODEL_DESERIALIZERS:
         if cached_model_deserializer.data_key in user_data:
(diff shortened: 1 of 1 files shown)


aliceriot:
the fake programs are both live, and the users all have usernames.

aliceriot:
looking at code now

profiles/management/commands/gen_realistic_search_data.py
```diff
@@ -248,10 +275,13 @@ def handle(self, *args, **options):
         if len(user_data_list) == existing_fake_user_count and len(program_data_list) == existing_fake_program_count:
             self.stdout.write("Realistic users and programs appear to exist already.")
         else:
-            new_program_count = deserialize_program_data_list(program_data_list)
+            new_programs = deserialize_program_data_list(program_data_list)
+            if options.get('staff_user', None):
```
aliceriot: I think the default 2nd argument for `.get` is already `None`, is there a python subtlety I'm missing?

aliceriot:
I think I'm running into an issue where the `social_auth` records aren't all being deleted, does that seem like something that's possible? I edited some of the users to try to figure out why I couldn't see their profiles on `/users/${username}`, which might have mucked something up.

pdpinch:
@gsidebo you didn't answer my questions on this issue (#879). I don't understand why this needed to be worked on right now?

gsidebo:
@pdpinch i knocked this one out really quickly. sorry, didn't see those questions. i'll answer them on the issue

gsidebo:
@aliceriot did you happen to edit User.username for those users you edited?

aliceriot:
@gsidebo yeah I did for one user. I ran the delete command, and then when I tried to generate the fake users again I got an error from the one user I had edited. I deleted her and deleted all the fake users, and then ran it again.

aliceriot:
nvm I don't think this is actually a problem with your code.

aliceriot:
@gsidebo This looks good, I just had one question, back to you!

aliceriot:
looks like one linting error, after that's fixed it's a 👍 from me!

PR #894, opened 2016-08-19
Added ecommerce models

### LINKED ISSUE #801: create ecommerce models
We need models, at least for products and orders. Each program may have its own pricing scheme, possibly with discounts. We also need a way to reflect any scholarships the learner has earned. 

Questions:

- [ ] When we worked on Teacher's Portal, we initially built our ecommerce system with Oscar, but then we ultimately removed it because it was too complex. Do we need to consider it again, or can we start from scratch?

noisecapella:
Fixes #801 

#### What's this PR do?
Creates `Order` and `Line` models to store information about orders made

ecommerce/models.py


ecommerce/models.py
```diff
@@ -0,0 +1,47 @@
...
+    created_at = DateTimeField(auto_now_add=True)
+    modified_at = DateTimeField(auto_now=True)
+    total_price_paid = DecimalField(decimal_places=2, max_digits=20)
+
+
+class Line(Model):
+    """
+    A line in an order
+    """
+    course_key = TextField()
```
giocalitri: this can probably be a char field, but it's not important

giocalitri:
👍

PR #895, opened 2016-08-19
Added price field to CourseRun

### LINKED ISSUE #846: Add a base price to the course run model

noisecapella:
Fixes #846 

#### What's this PR do?
Adds a price field to the CourseRun model


giocalitri:
I do not think this should go in the course run, but in a separated model with 

- a foreign key to the course run,
- a creation time
- (a way to support multiple prices but with only one valid: maybe the last one)

I am also in favor to postpone the implementation of this, but at some point we want to be able to track why a price was X at the time Y and if we loose context, this is going to be a nightmare.

noisecapella:
That's a good idea, I'll make the change

noisecapella:
How is this? The `is_valid` field seems like a good idea but I think there should only be one that's valid at any given time. We could write an api function to verify that. Choosing the last one in a list of prices seems error prone

ecommerce/models.py
```diff
@@ -45,3 +48,15 @@ class Line(Model):
...
+class CoursePrice(Model):
+    """
+    Information about a course run's price and other ecommerce info
+    """
+    course_run = ForeignKey(CourseRun)
+    price = DecimalField(decimal_places=2, max_digits=20)
+    is_valid = BooleanField(default=False)
+
+    created_at = DateTimeField(auto_now_add=True)
+    modified_at = DateTimeField(auto_now=True)
```
giocalitri: I would put a constraint (not sure how) to enforce that there can be only one `course_run` with `is_valid = True`

noisecapella: I'm not sure that kind of constraint exists on the database side, but I can write an API function to enforce this

giocalitri: maybe in the `.save()` method of the model?

noisecapella: Sure, I'll take a look

noisecapella:
Ok, added the constraint

giocalitri:
why there are the migrations 
`courses/migrations/0012_courserun_price.py`
and 
`courses/migrations/0013_remove_courserun_price.py`
?

noisecapella:
I'll remove them

noisecapella:
I also fixed a bug with the check for is_valid

giocalitri:
I think this is fine 👍

PR #897, opened 2016-08-22
Added API to create Order and populate CyberSource parameters

### LINKED ISSUE #891: Add API to handle order fulfillment

noisecapella:
Fixes #891 

#### What's this PR do?
Adds `/api/v0/checkout/` to create a new `Order` and provide parameters for CyberSource

 - Create a verified course on your devstack.
 - Set up a `CourseRun` object which matches the course on your devstack. Set the `start_date` to be before today and the other three date fields should be `None`. You should see an 'Enroll' button on the dashboard.
 - Create a `CoursePrice` object which points to the `CourseRun`. `is_valid` must be true, choose any positive value for the price.
 - Fill in the new environment variables in `.env`. Talk to me in private for the keys needed here
 - Click `Enroll`. You should be directed to a CyberSource page.
 - Enter fake information for the address component. For the credit card, use `4111111111111111`, the card type is `Visa`, the CVN is any three digits, and the expiration month is some time in the future. Enter a legit email address so you can see the receipt generated.
 - Pay for the course. You should be redirected to a request bin with the word `ok`.
 - Ask me for the purchase data email. It should report a successful purchase and include all the information entered (but only the last 4 digits of the credit card).

This workflow should also be the same for course runs with the button `Upgrade`.


noisecapella:
@giocalitri When you have some free time can you start the review of this PR? I will be working on tests over the next day but the code should be ready for review and functionality should work

static/js/components/dashboard/CourseAction.js
```diff
@@ -54,7 +54,9 @@ export default class CourseAction extends React.Component {
       break;
     case STATUS_ENROLLED_NOT_VERIFIED: {
       let courseUpgradeUrl = urljoin(SETTINGS.edx_base_url, '/course_modes/choose/', firstRun.course_id);
-      action = this.makeActionButton("Upgrade", firstRun, false, courseUpgradeUrl);
+      action = this.makeActionButton("Upgrade", firstRun, false, () => {
```
noisecapella: This is still in progress

static/js/components/dashboard/CourseAction.js
```diff
@@ -71,8 +73,29 @@ export default class CourseAction extends React.Component {
         }
       }
 
-      let courseInfoUrl = urljoin(SETTINGS.edx_base_url, '/courses/', firstRun.course_id, 'about');
-      action = this.makeActionButton("Enroll", firstRun, disabled, courseInfoUrl);
+      //let courseInfoUrl = urljoin(SETTINGS.edx_base_url, '/courses/', firstRun.course_id, 'about');
+
+      action = this.makeActionButton("Enroll", firstRun, disabled, () => {
+        checkout(firstRun.course_id).then(result => {
```
noisecapella: This will be cleaned up too

ecommerce/api.py
```diff
@@ -0,0 +1,149 @@
...
+def create_unfulfilled_order(course_id, user):
+    """
+    Create a new Order which is not fulfilled for a purchasable course run. If course run is not purchasable,
+    it raises an Http404
+
+    Args:
+        course_id (str):
+            A course key
+        user (User):
+            The purchaser of the course run
```
giocalitri: Missing `Returns` section in docstring

ecommerce/api.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    course_run = get_purchasable_course_run(course_id, user)
+    price = course_run.courseprice_set.get(is_valid=True).price
```
giocalitri: Question: this is assumed to be always valid because we enforce the check at insertion time?

ecommerce/api.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    # http://apps.cybersource.com/library/documentation/dev_guides/Secure_Acceptance_WM/Secure_Acceptance_WM.pdf
+    # Section: API Fields
+    payload = {
+        'access_key': settings.CYBERSOURCE_ACCESS_KEY,
+        'amount': str(order.total_price_paid),
+        'consumer_id': get_social_username(order.user),
+        'currency': 'USD',
+        'locale': 'en-us',
+        # TODO
```
giocalitri: left back?

ecommerce/api.py
noisecapella: Yes, we shouldn't have more than one `is_valid` price. We check that there is at least one in `get_purchasable_course_run`

ecommerce/api.py
noisecapella: Will add

ecommerce/api.py
```diff
@@ -0,0 +1,149 @@
...
+        'override_custom_cancel_page': 'https://micromasters.mit.edu?cancel',
+        'override_custom_receipt_page': "https://micromasters.mit.edu?receipt",
```
giocalitri: is it ok if we hardcode these URLs?

ecommerce/api.py
```diff
@@ -0,0 +1,149 @@
...
+        'reference_number': "MM-{}".format(order.id),
```
giocalitri: Maybe we should ask how this should look like. For now it is ok
@pdpinch ?

ecommerce/api.py
noisecapella: I'll remove the TODO. It's to replace the two next lines with urls with valid order receipt (#904) and order cancellation pages (#905)

ecommerce/api.py
noisecapella: They will be replaced in a future PR with URLs going to our own server

ecommerce/api.py
noisecapella: Yes, I'm also wondering if the order id number is unique enough or if we should add the server in here somewhere

ecommerce/api.py
giocalitri: should not be configurable?

ecommerce/api.py
```diff
@@ -0,0 +1,151 @@
...
+        'profile_id': settings.CYBERSOURCE_PROFILE_ID,
+        'signed_date_time': datetime.utcnow().strftime(ISO_8601_FORMAT),
+        'transaction_type': 'sale',
+        'transaction_uuid': uuid.uuid4().hex,
+        'unsigned_field_names': '',
+    }
+
+    field_names = sorted(list(payload.keys()) + ['signed_field_names'])
+    payload['signed_field_names'] = ','.join(sorted(field_names))
```
giocalitri: cannot you remove this `sorted` and just sort before? Or am I missing something?

ecommerce/api_test.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        Create common objects
+        """
+        super(PurchasableTests, self).setUp()
+
+        self.course_run = CourseRunFactory.create(course__program__live=True)
+        # Also creates a user
+        self.program_enrollment = ProgramEnrollmentFactory.create(program=self.course_run.course.program)
+        self.user = self.program_enrollment.user
+        CoursePriceFactory.create(course_run=self.course_run, is_valid=True)
```
giocalitri: Please move the creation of these data to the `setUpTestData`: we do not need to recreate the entries every time.

If you have some specific parts that get changed by a test, you can use `setUp` to restore the default status

ecommerce/api.py
noisecapella: Yes, `field_names` is already sorted so it's not necessary here too

ecommerce/api_test.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_not_live(self):
+        """
+        Make sure get_purchasable_course_run raises an Http404 if the program is not live
+        """
+        program = self.course_run.course.program
+        program.live = False
+        program.save()
+
+        with self.assertRaises(Http404):
+            get_purchasable_course_run(self.course_run.edx_course_key, self.user)
```
giocalitri: I think a test for `Line.objects.filter(...).exists()` is missing

giocalitri: also, we should assume that the other functions work? (I am fine with that if you say so)

ecommerce/api.py
noisecapella: In a future PR this will be pointing to our own server, using the host in `request` as a base. We won't need it to be configured

ecommerce/api_test.py
noisecapella: There will be negative tests to adjust each of these models. I'll move this into a separate function called by each test so it's more explicit

ecommerce/api_test.py
noisecapella: The tests are still in progress

giocalitri:
the PR is overall in a good shape

noisecapella:
Ok, finally all tests are in place I think

ecommerce/api.py
```diff
@@ -0,0 +1,164 @@
...
+        status=Order.CREATED,
+        total_price_paid=price,
+        user=user,
+    )
+    Line.objects.create(
+        order=order,
+        course_key=course_run.edx_course_key,
+        description='Seat for {}'.format(course_run.title),
+        price=price,
+    )
```
giocalitri: More a general comment than a request for change: I noticed that edX makes extensive use of factories even in the normal workflow and not only in the tests. This would be the perfect usage: you will call the factory for `Line` and then you override the `.create` method to take care of creating the order first.
Again, only food for thoughts.

ecommerce/views_test.py
```diff
@@ -0,0 +1,77 @@
...
+    patch,
+)
+
+from django.core.urlresolvers import reverse
+from django.test import override_settings
+from rest_framework.status import (
+    HTTP_200_OK,
+    HTTP_400_BAD_REQUEST,
+    HTTP_403_FORBIDDEN,
+)
```
giocalitri: nit: I think it's better to import the `status` module and not any single code we need

ecommerce/views_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_valid_course_id(self):
+        """
+        If course_id is not present in payload a ValidationError is raised
+        """
+        user = UserFactory.create()
+        self.client.force_login(user)
+        resp = self.client.post(reverse('checkout'), {}, format='json')
+        assert resp.status_code == HTTP_400_BAD_REQUEST
+        assert resp.json() == ['Missing course_id']
```
giocalitri: we should probably implement our own mixin to handle these errors, but not for this PR

ecommerce/views.py
```diff
@@ -0,0 +1,38 @@
...
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, *args, **kwargs):
+        """
+        Creates a new unfulfilled Order and returns information used to submit to CyberSource
+        """
+        try:
+            course_id = request.data['course_id']
+        except KeyError:
+            raise ValidationError("Missing course_id")
```
giocalitri: we should probably implement our own mixin to handle these errors, but not for this PR

ecommerce/views_test.py
```diff
@@ -0,0 +1,77 @@
...
+        course_key = 'course_key'
+
+        user = UserFactory.create()
+        self.client.force_login(user)
+        order = MagicMock()
+        payload = {
+            'a': 'payload'
+        }
+        with patch('ecommerce.views.create_unfulfilled_order', autospec=True, return_value=order) as create_mock:
+            with patch(
```
giocalitri: nit: you do not need two `with`: it's enough to put the different patch one after the other comma separated

giocalitri:
I think the code is fine: only few very minor comments
Not sure if we need a second pair of eyes on the JS

ecommerce/views_test.py
noisecapella: Can you do that if two different `return_value`s are being defined?

giocalitri:
first problem I found:
I think you need to handle the "Upgrade" button as well.
screen shot 2016-08-25 at 4 58 35 pm

noisecapella:
The Upgrade button should be handled the same as the Enroll button. What's not working about it?

ecommerce/views_test.py
giocalitri: sure it's something like: 
```
with patch('ecommerce.views.create_unfulfilled_order', autospec=True, return_value=order) as create_mock, patch(....,  return_value=payload) as generate_mock:
```
the two mocks keep being separated

giocalitri: they just go on the same logical line

giocalitri:
I think I misunderstood the problem: I am debugging now

ecommerce/views_test.py
noisecapella: Oh, I see, I'll make the change

ecommerce/api.py
noisecapella: Wouldn't factories populate with a bunch of nonsense values?

ecommerce/views_test.py
noisecapella: Will do

ecommerce/api.py
giocalitri: they do not use factory_boy :
see https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/new/course_grade.py#L171-L202

giocalitri:
We need an admin for the course price

giocalitri:
👍

PR #900, opened 2016-08-23
UI for sending learner email

### LINKED ISSUE #719: Allow instructors to compose an email to learners in search results
_This issue depends on #669 (Learner Search) being done and is part of the actual implementation of #694. The issue to implement the actual sending of the email is #718_ 

After an instructor performs a learner search, they need a way to select any number of learners from the search results and compose a message to them. The following should be done for this issue:

- Add the "Email Selected" button (can be seen in the mockup in #669) to the search results page
- Keep in mind that these messages will be text-only for the time being
- When the message is submitted, post the message body and recipient list to the endpoint created is #718

@roberthouse54: Your input would be appreciated for the message composition part of this. The composition interface probably doesn't need much more than a simple text box, a submit button, and some simple display of the recipients. We could present this interface in a completely separate view, in a lightbox, or in some other manner that someone with actual design talent might think of 😄

aliceriot:

this closes #719 

#### What's this PR do?

This adds the ui for sending email to users based on a search query. The API isn't there yet, so for now we just dump the email on the floor when you click send.

Read over the new changes, especially the `EmailCompositionDialog` component and the new email reducer.

Go to the `learners` page. Click 'Email Selected', then you should see a dialog box. The 'recipient count' in the box should match what is shown in the normal search interface. You should be able to edit the 'subject' and 'body' fields in the email. Editing these fields should dispatch redux actions. You should be able to click the 'send' button, but it will just `console.log` the email for now.

Also, if you inspect the redux store, you should see the search query stringified in there, so we'll have it ready to `PATCH` later.

email_compose

### DIFF
--- a/static/js/actions/email.js
+++ b/static/js/actions/email.js
@@ -1,0 +1,13 @@
+// @flow
+export const actionCreatorGenerator = (type: string) => (
+  (args: any) => args === undefined ? { type: type } : { type: type, payload: args }
+);
+
+export const START_EMAIL_EDIT = 'START_EMAIL_EDIT';
+export const startEmailEdit = actionCreatorGenerator(START_EMAIL_EDIT);
+
+export const UPDATE_EMAIL_EDIT = 'UPDATE_EMAIL_EDIT';
+export const updateEmailEdit = actionCreatorGenerator(UPDATE_EMAIL_EDIT);
+
+export const CLEAR_EMAIL_EDIT = 'CLEAR_EMAIL_EDIT';
+export const clearEmailEdit = actionCreatorGenerator(CLEAR_EMAIL_EDIT);
--- a/static/js/actions/email_test.js
+++ b/static/js/actions/email_test.js
@@ -1,0 +1,50 @@
+// @flow
+import { assert } from 'chai';
+
+import {
+  actionCreatorGenerator,
+  startEmailEdit,
+  updateEmailEdit,
+  clearEmailEdit,
+
+  START_EMAIL_EDIT,
+  UPDATE_EMAIL_EDIT,
+  CLEAR_EMAIL_EDIT,
+} from './email';
+
+describe('email actions', () => {
+  describe('actionCreatorGenerator', () => {
+    it('should return a function', () => {
+      assert.isFunction(actionCreatorGenerator(START_EMAIL_EDIT));
+    });
+
+    it('returned function should return a simple action when no args are passed', () => {
+      let creator = actionCreatorGenerator(START_EMAIL_EDIT);
+      assert.deepEqual({ type: START_EMAIL_EDIT }, creator());
+    });
+
+    it('returned function should return an action w/ payload when args are passed', () => {
+      let creator = actionCreatorGenerator(START_EMAIL_EDIT);
+      let expected = { type: START_EMAIL_EDIT, payload: "HI" };
+      assert.deepEqual(expected, creator('HI'));
+    });
+  });
+
+  describe('email action helpers', () => {
+    const assertCreatedActionHelper = ([actionHelper, actionType]) => {
+      it(`should create the ${actionType} simple action helper correctly`, () => {
+        assert.deepEqual(actionHelper(), {type: actionType});
+      });
+
+      it(`should create the ${actionType} action helper with args correctly`, () => {
+        assert.deepEqual(actionHelper({foo: "bar"}), { type: actionType, payload: { foo: "bar" } });
+      });
+    };
+
+    [
+      [startEmailEdit, START_EMAIL_EDIT],
+      [updateEmailEdit, UPDATE_EMAIL_EDIT],
+      [clearEmailEdit, CLEAR_EMAIL_EDIT],
+    ].forEach(assertCreatedActionHelper);
+  });
+});
--- a/static/js/components/EmailCompositionDialog.js
+++ b/static/js/components/EmailCompositionDialog.js
@@ -1,0 +1,77 @@
+// @flow
+import React from 'react';
+import Dialog from 'material-ui/Dialog';
+import Button from 'react-mdl/lib/Button';
+import R from 'ramda';
+import type { EmailEditState } from '../reducers/email';
+
+const createDialogActions = (close, send) => ([
+  <Button
+    type="button"
+    className="dialog-button cancel-button"
+    key="first"
+    onClick={close}
+  >
+    Cancel
+  </Button>,
+  <Button
+    type="button"
+    className="dialog-button save-button"
+    key="second"
+    onClick={send}
+  >
+    Send
+  </Button>
+]);
+
+const hitsCount = searchkit => R.isNil(searchkit) ? 0 : searchkit.getHitsCount();
+
+type EmailDialogProps = {
+  closeEmailDialog: () => void,
+  updateEmailEdit:  Function,
+  open:             boolean,
+  email:            EmailEditState,
+  searchkit:        Object,
+  sendEmail:        (e: EmailEditState) => void,
+};
+
+const EmailCompositionDialog = (props: EmailDialogProps) => {
+  const {
+    closeEmailDialog,
+    updateEmailEdit,
+    open,
+    email,
+    searchkit,
+    sendEmail,
+  } = props;
+
+  return <Dialog
+    open={open}
+    className="email-composition-dialog"
+    actions={createDialogActions(closeEmailDialog, sendEmail)}
+    onRequestClose={closeEmailDialog}
+    title="New Email"
+  >
+    <div className="email-composition-contents">
+      <span className="user-count">
+        {hitsCount(searchkit)} recipients selected
+      </span>
+      <textarea
+        rows="1"
+        className="email-subject"
+        placeholder="Subject"
+        value={email.subject || ""}
+        onChange={updateEmailEdit('subject')}
+      />
+      <textarea
+        rows="7"
+        className="email-body"
+        placeholder="Type a message"
+        value={email.body || ""}
+        onChange={updateEmailEdit('body')}
+      />
+    </div>
+  </Dialog>;
+};
+
+export default EmailCompositionDialog;
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -1,3 +1,4 @@
+// @flow
 import React from 'react';
 import {
   SearchkitComponent,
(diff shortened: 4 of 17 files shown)


static/js/components/EmailCompositionDialog.js
```diff
@@ -0,0 +1,77 @@
...
+    type="button"
+    className="dialog-button save-button"
+    key="second"
+    onClick={send}
+  >
+    Send
+  </Button>
+]);
+
+const hitsCount = sk => R.isNil(sk) ? 0 : sk.getHitsCount();
```
gsidebo: 1. nitpick: can we just change `sk` to `searchkit` or `searchkit_prop`?
1. general question: i'm seeing a lot of `R` use lately. it seems like theres a good amount of overlap between R and lodash, this `isNil` method being one of them (https://lodash.com/docs#isNil). does R have any specific advantages? i ask b/c i think it's safe to say that lodash/underscore are more widely used/understood than R. there's a bit more cognitive load when an uninitiated dev (let's call him 'davin') sees a method being invoked on R

aliceriot: I'm fine with `s/sk/searchkit/g`, I had used `sk` to make the function body a little shorter.

I prefer Ramda because it's designed from the ground up for a functional style, so all the functions Ramda defines are autocurried, data-last, etc. It has a lot of handy methods for writing code in a functional style too, like `R.compose`, `R.map`, etc.

Lodash has a functional module (`lodash/fp`) which also does data-last, autocurried functions, but Ramda is more 'functional-nnative-, if that makes sense.

I've been using ramda where it makes sense to do so - I don't think we should really make a concerted effort to purge lodash, but I've been preferring writing in a more functional style lately and ramda is nice for that :)

static/js/components/EmailCompositionDialog.js
```diff
@@ -0,0 +1,77 @@
...
+  return <Dialog
+    open={open}
+    className="email-composition-dialog"
+    actions={createDialogActions(closeEmailDialog, sendEmail)}
+    onRequestClose={closeEmailDialog}
+    title="New Email"
+  >
+    <div className="email-composition-contents">
+      <span className="user-count">
+        { `${hitsCount(searchkit)} recipients selected` }
```
gsidebo: i think this can be rewritten as: `{hitsCount(searchkit)} recipients selected`

static/js/components/LearnerSearch.js
```diff
@@ -142,6 +173,7 @@ export default class LearnerSearch extends SearchkitComponent {
                     role="button"
                     id="email-selected"
                     className="mm-button"
+                    onClick={this.openEmailComposer}
```
gsidebo: one way to get rid of the `openEmailComposer()` definition above is to add `openEmailComposer` to the list of `const` variables and rewrite this line as `_.bind(openEmailComposer, this, this.searchkit)` or `_.partial(openEmailComposer, this.searchkit`. ramda might have some equivalent as well

static/js/components/LearnerSearch.js
```diff
@@ -73,18 +76,46 @@ export default class LearnerSearch extends SearchkitComponent {
...
       <div className="learners-search">
+        <EmailCompositionDialog
+          open={emailDialogVisibility}
+          closeEmailDialog={closeEmailDialog}
+          updateEmailEdit={updateEmailEdit}
+          email={email}
+          sendEmail={sendEmail}
+          searchkit={this.searchkit}
+        />
+       { this.props.children }
```
gsidebo: any specific reason that this was added?

static/js/reducers/email.js
```diff
@@ -0,0 +1,34 @@
...
+
+export const INITIAL_EMAIL_STATE = {};
+
+export const NEW_EMAIL_EDIT = {
+  subject:    null,
+  body:       null,
+  query:      null,
+};
+
+export type EMAIL_EDIT_STATE = {
```
gsidebo: it seems that in other parts of our code, we name state _types_ with camel case (eg: `UIState`, `ProfileState`), and _examples_ of those states with upper-snake-case (eg: `INITIAL_UI_STATE`). shouldn't `EMAIL_EDIT_STATE` be named `EmailEditState`? also, going by the example in static/js/reducers/ui.js, it seems that `NEW_EMAIL_EDIT` and `INITIAL_EMAIL_STATE ` should be declared after `EmailEditState`, and should be defined like so:

```javascript
export type EmailEditState { ... }

export const INITIAL_EMAIL_STATE: EmailEditState = { }

export const NEW_EMAIL_EDIT: EmailEditState = { ... }
```

let me know if i'm missing something here

static/js/actions/email.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+export const actionCreatorGenerator = (type: string) => (
+  (args: any) => args === undefined ? { type: type } : { type: type, payload: args }
+);
```
gsidebo: seems like this could be useful elsewhere. does it make sense to create `static/js/actions/util.js` (et al) and define this in there?

static/js/actions/email_test.js
```diff
@@ -0,0 +1,47 @@
...
+  CLEAR_EMAIL_EDIT,
+} from './email';
+
+describe('email actions', () => {
+  describe('actionCreatorGenerator', () => {
+    it('should return a function', () => {
+      assert.isFunction(actionCreatorGenerator(START_EMAIL_EDIT));
+    });
+
+    it('returned function should be a simple action creator with no args', () => {
```
gsidebo: some possibly-clearer alternatives to this description and the one on line 26

'should return a simple action creator function when no args are provided'

'should return an action creator function with payload when args are provided'

static/js/actions/email_test.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    it('returned function should be an action creator with payload with args', () => {
+      let creator = actionCreatorGenerator(START_EMAIL_EDIT);
+      let expected = { type: START_EMAIL_EDIT, payload: "HI" };
+      assert.deepEqual(expected, creator('HI'));
+    });
+  });
+
+  describe('created action helpers', () => {
```
gsidebo: it doesn't seem like we're actually testing these 3 actions for anything apart from the fact that they're generic actions created by this helper method (which you've already tested above). if that's all we want to assert, maybe we should write something like this:

```javascript
const assertGenericAction(func, type_str) {
  assert.deepEqual(func(), {type: type_str})
  assert.deepEqual(func({a: 1, b: 2}), {type: type_str, payload: {a: 1, b: 2}})
}

describe('email actions', () => {
  it('should be generic', () => {
      assertGenericAction(startEmailEdit, START_EMAIL_EDIT)
      assertGenericAction(updateEmailEdit, UPDATE_EMAIL_EDIT)
      assertGenericAction(clearEmailEdit, CLEAR_EMAIL_EDIT)
  });
});
```

static/js/actions/email.js
aliceriot: take a look [here](http://redux.js.org/docs/basics/Actions.html)

I agree it's probs useful elsewhere, just wanted to get feedback on it before I rewrote both of the other action files! I think if you agree that's a good idea I might open an issue and open a follow-up PR to make that change.

static/js/actions/email_test.js
aliceriot: works for me!

static/js/reducers/email_test.js
```diff
@@ -0,0 +1,56 @@
...
+  afterEach(() => {
+    sandbox.restore();
+
+    store = null;
+    dispatchThen = null;
+  });
+
+  let query = { query: "Hi, I'm a query!" };
+
+  let initialExpectation = { ...NEW_EMAIL_EDIT, ...{ query: query } };
```
gsidebo: would this work?: `let initialExpectation = { ...NEW_EMAIL_EDIT, query: query };`

static/js/reducers/email_test.js
```diff
@@ -0,0 +1,56 @@
...
+
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+    store = configureTestStore(rootReducer);
+    dispatchThen = store.createDispatchThen(state => state.email);
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+
```
gsidebo: unnecessary blank line

static/js/reducers/ui_test.js
```diff
@@ -325,4 +327,17 @@ describe('ui reducers', () => {
...
     });
   });
+
+  describe('Email dialog visibility', () => {
+    [true, false].forEach(bool => {
+      it(`should let you set email dialog visibility to ${bool}`, () => {
+        return dispatchThen(setEmailDialogVisibility(bool), [
+          SET_EMAIL_DIALOG_VISIBILITY
+        ]).then(state => {
+          assert.deepEqual(state.emailDialogVisibility, bool);
```
gsidebo: this can just be `assert.equal` i think

static/js/actions/email.js
gsidebo: yeah, i think it will definitely be useful elsewhere. there should also be a clean-enough way to extend it for actions that require specific number/types of arguments. if you want to move this (and the tests) into a util file as part of the follow-up issue, that's cool w/ me

static/js/actions/email_test.js
```diff
@@ -0,0 +1,50 @@
...
+  describe('email action helpers', () => {
+    const assertCreatedActionHelper = ([actionHelper, actionType]) => {
+      it(`should create the ${actionType} simple action helper correctly`, () => {
+        assert.deepEqual(actionHelper(), {type: actionType});
+      });
+
+      it(`should create the ${actionType} action helper with args correctly`, () => {
+        assert.deepEqual(actionHelper({foo: "bar"}), { type: actionType, payload: { foo: "bar" } });
+      });
+    };
```
gsidebo: this'll work. if/when the generic action helper code is moved into a util file, i think this assert function, and the ''actionCreatorGenerator' block above, should move with it

static/js/actions/email.js
aliceriot: yeah, I was planning to do something like that. We already have `util/redux`.

static/js/actions/email_test.js
aliceriot: agreed :)

static/js/components/LearnerSearch.js
aliceriot: oh oops, debugging

static/js/components/LearnerSearch.js
aliceriot: ah yeah we might be able to do something like that, I was trying to get away from re-declaring the function on every re-render. Before I was passing in a function that took `searchkit` and an argument and then returns an `onClick` handler, but I realized it was being called constantly, which was creating some strange effects with a variable I was trying to keep in closure (the current query object at the time of opening the dialog).

I'll give that a try!

static/js/reducers/email.js
aliceriot: yeah it should be `EmailEditState`, I'll change it around

static/js/reducers/email_test.js
aliceriot: indeed it does!

aliceriot:
@gsidebo all comments should be addressed!

gsidebo:
👍

PR #901, opened 2016-08-23
Remove 'new group from selected' button

### LINKED ISSUE #852: Remove "New Group from Selected" button in search results
We don't have a feature that uses this yet. 

image

aliceriot:

closes #852 

#### What's this PR do?

removes an unimplemented button

Just make sure the button's absence looks ok!

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -157,15 +157,6 @@ export default class LearnerSearch extends SearchkitComponent {
                 <Cell col={6} className="result-info">
                   <div
                     role="button"
-                    id="new-group"
-                    className="mm-button"
-                  >
-                    <span>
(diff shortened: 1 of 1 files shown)


noisecapella:
The learner page looks a little odd. On my laptop:

screenshot from 2016-08-25 18-05-50

Not sure if that's caused by removing the button though. If we have an issue to address learner page style issues then :+1:

noisecapella:
Specifically the sort and the pagination buttons don't quite line up with each other, and I'm not sure where the Email Selected button wants to be, it's odd to see it just floating there not aligned with anything

aliceriot:
I think the email button will be sorted out in Rob's pr, not sure about the
pagination/sort buttons.

amir-qayyum-khan:
👍

PR #902, opened 2016-08-23
Added test case to mock out Elasticsearch to speed up tests

### LINKED ISSUE #721: Add two test case base classes and remove ESTestCase
One of the reasons tests were slow in LORE is because many tests needlessly interacted with Elasticsearch, either via setup and teardown or with signals for commonly used models. To only use Elasticsearch where we need to we should have two test cases:
 - One for database tests which don't care about Elasticsearch. We should be able to mock the elasticsearch module so that it ignores all interaction with it.
 - One for search tests. This should call `clear_index()` like we do in `ESTestCase` currently

### LINKED ISSUE #898: Unable to delete programs
To reproduce, run the `gen_realistic_search_data` command, then run the `delete_realistic_search_data` command (or just run `Program.objects.all().delete()`)

noisecapella:
Fixes #721

#### What's this PR do?
 - Adds `MockedESTestCase` to mock Elasticsearch to speed up tests which don't use it

There shouldn't be any non-testing code changed

### DIFF
--- a/search/base.py
+++ b/search/base.py
@@ -1,7 +1,12 @@
 """
 Base test classes for search
 """
-from django.test import TestCase
+from mock import patch
+
+from django.test import (
+    override_settings,
+    TestCase,
+)
 
 from search.api import recreate_index
 
@@ -14,3 +19,32 @@ class ESTestCase(TestCase):
     def setUp(self):
         super(ESTestCase, self).setUp()
         recreate_index()
+
+
+@override_settings(ELASTICSEARCH_URL="fake")
+class MockedESTestCase(TestCase):
+    """
+    Mock ES signals
+    """
+
+    def setUp(self):
+        self.patchers = [
+            patch('search.api.get_conn', autospec=True),
+            patch('search.api.bulk', autospec=True, return_value=(0, [])),
+            patch('search.api.Mapping', autospec=True),
+        ]
+        for patcher in self.patchers:
+            patcher.start()
+        try:
+            super(MockedESTestCase, self).setUp()
+        except:
+            for patcher in self.patchers:
+                patcher.stop()
+            raise
+
+    def tearDown(self):
+        try:
+            super(MockedESTestCase, self).tearDown()
+        finally:
+            for patcher in self.patchers:
+                patcher.stop()


profiles/management/tests.py
```diff
@@ -0,0 +1,82 @@
...
+        assert CourseRun.objects.count() == 18
+        # No staff option provided
+        assert Role.objects.count() == 0
+        assert ProgramEnrollment.objects.count() == 60
+        assert User.objects.count() == 60
+        assert User.objects.filter(
+            username__startswith=delete_realistic_search_data.FAKE_USER_USERNAME_PREFIX
+        ).count() == 60
+        assert CachedEnrollment.objects.count() == 60
+        assert CachedCertificate.objects.count() == 60
```
gsidebo: these two assertions pass for you?? they certainly don't for me, and i don't think they should. our current approach is to create a CachedEnrollment and CachedCertificate for every single CourseRun times every single User (most of which just have a null `data` value). this count is 1080 when i run this test suite (60 users * 18 course runs). even if you're just looking for non-'empty' Certificates/Enrollments, i'm pretty sure this logic is wrong. some users have several Certificates/Enrollments, some have none at all, and i doubt that they both add up to 60 total

gsidebo:
in it's current state, i really don't think we should be merging this. here's some basic profiling info for these tests:

```
11.60s call     profiles/management/tests.py::FakeDataTests::test_can_delete
8.12s setup    profiles/management/tests.py::FakeDataTests::test_can_delete
2.46s call     profiles/management/tests.py::FakeDataTests::test_can_generate_fake_data
2.20s call     profiles/management/tests.py::FakeDataTests::test_gen_staff_option
```

these tests take over 20 seconds to complete, and it's pretty easy to see why. using a _live_ DB, we are creating these Users/Programs and their web of relationships three times, then deleting all of them once. we should also consider the fact that we're adding tests to our test suite that do not exist to test the actual functionality of micromasters. travis would be running these tests for every merge/deploy, and i'm really doubtful of the value that we'd be adding in that case.

some possible alternatives:

- instead of adding this logic as unit tests, we could just make a series of assertions after the commands execute (making sure the # of Users/whatever is what we expect, making sure that all fake users have been removed after a delete command, etc)
- keep this as-is, but mock out all DB communication

what do you think?

noisecapella:
I'm going to remove the tests for the management commands here and just leave the new test cases which mock out elasticsearch

gsidebo:
looks great! 👍

PR #907, opened 2016-08-23
Replaced var with let

noisecapella:
#### What's this PR do?
Fixes the webpack build for node 5.11 which heroku uses

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,9 @@
     "webpack-dev-server": "1.14.1",
     "wowjs": "^1.1.3"
   },
+  "engines": {
+    "node": "6.2.0"
+  },
   "scripts": {
     "postinstall": "./webpack_if_prod.sh",
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",


noisecapella:
Can you review this @blarghmatey?

blarghmatey:
This looks good. Another option is to set the version as specified here https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

aliceriot:
I think we should try to maintain dev / ci / rc / prod parity for versions, otherwise I think we risk some stranger, less noisy, and more subtle edge case showing up later.

noisecapella:
Yeah, I think setting the node version is probably a better long term solution

noisecapella:
Can you look at the deploy failure @blarghmatey? Tomorrow is fine

blarghmatey:
@noisecapella The `engine` section needs to go into the `package.json`. Move what you've got from `app.json` and it should work.

noisecapella:
That makes sense, I'll try that

blarghmatey:
👍

PR #911, opened 2016-08-24
Switched to use enzyme in IntegrationTestHelper

noisecapella:
#### What's this PR do?
Passes an enzyme wrapper as the first argument of `renderComponent` instead of the React component. We don't actually use the component anywhere at the moment so this shouldn't cause any problems

### DIFF
--- a/static/js/util/integration_test_helper.js
+++ b/static/js/util/integration_test_helper.js
@@ -1,5 +1,5 @@
-import ReactDOM from 'react-dom';
 import React from 'react';
+import { mount } from 'enzyme';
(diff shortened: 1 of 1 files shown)


aliceriot:
execution time for tests seems about the same:

master:

userpage_master

this branch:

userpage_enzyme

almost all the numbers actually get smaller, not sure how that makes sense!

anyway, 👍

noisecapella:
Yeah, I would bet random fluxtuation in travis instances

aliceriot:
oh that's running locally, but same thing probably.

Issue #905, opened 2016-08-23
Create an order cancelled page


noisecapella:
If they cancel on CyberSource the user will be directed to this page

pdpinch:
Have you discussed this with @roberthouse54 ?

noisecapella:
This will be a toast message similar to #904

Issue #904, opened 2016-08-23
Create an order receipt page


noisecapella:
This is the page which the user will be redirected to after a successful payment on cybersource

pdpinch:
@roberthouse54 has a design for this, but it's just the dashboard with a banner. Is that OK?

https://invis.io/9B804AMXA#/182566039_Mm_Dashboard_After_Ecommerce

noisecapella:
This is fine. We can send the receipt url '.../dashboard/?order=MM-xx' and use the order number to show the receipt

PR #913, opened 2016-08-25
Added order fulfillment API for CyberSource

### LINKED ISSUE #891: Add API to handle order fulfillment

noisecapella:
Fixes #891

#### What's this PR do?
Creates an API which CyberSource can use to trigger order fulfillment. 

ecommerce/views.py

Talk to me about this


noisecapella:
Closing this since we're going in a different direction

ecommerce/views.py
```diff
@@ -36,3 +46,46 @@ def post(self, request, *args, **kwargs):
...
+            if request.data['decision'] != 'ACCEPT':
+                # This may happen if the user clicks 'Cancel Order'
+
+                order.status = Order.FAILED
+                order.save()
+            else:
+                # Do the verified enrollment with edX here
+
+                order.status = Order.FULFILLED
+                order.save()
```
gsidebo: line breaks at 77 and 82 don't seem necessary, and `order.save()` can be moved outside of the if/else instead of repeated in both conditions

ecommerce/views_test.py
```diff
@@ -74,3 +86,78 @@ def test_creates_order(self):
...
+        If CyberSource POSTs with fields missing, we should at least save it in a receipt
+        """
+        data = {}
+        for _ in range(5):
+            data[FAKE.text()] = FAKE.text()
+        with patch('ecommerce.views.IsSignedByCyberSource.has_permission', return_value=True):
+            try:
+                self.client.post(reverse('order-fulfillment'), data=data)
+            except KeyError:
+                pass
```
gsidebo: can you explain why this except/pass is here?

gsidebo:
left some code comments. let's get together to discuss functionality testing

ecommerce/views_test.py
noisecapella: If there is a key missing from the Cybersource response a `KeyError` is raised. The intent of the test here is to check that even if an exception is raised, we still have the `Receipt` object saved with the data so we can analyze it later

gsidebo:
functionality looks good

ecommerce/api.py
```diff
@@ -162,3 +166,40 @@ def make_reference_id(order):
             A reference number for use with CyberSource to keep track of orders
     """
     return "MM-{}-{}".format(settings.CYBERSOURCE_REFERENCE_PREFIX, order.id)
+
```
noisecapella: Yes, that's for #904 and #905

ecommerce/api.py
```diff
@@ -162,3 +166,40 @@ def make_reference_id(order):
...
+    is already fulfilled an EcommerceException is raised.
+
+    Args:
+        reference_number (str):
+            A string which contains the order id and the instance which generated it
+    Returns:
+        Order:
+            An order
+    """
+    if not reference_number.startswith("MM-"):
```
noisecapella: Will do

ecommerce/views_test.py
gsidebo: ok. can you add a brief comment above the except line? something like 'Cybersource response should raise an exception here - ignore it and make sure we still have a Receipt'. also, should we have a custom exception in these cases?

ecommerce/api.py
```diff
@@ -162,3 +166,40 @@ def make_reference_id(order):
             A reference number for use with CyberSource to keep track of orders
     """
     return "MM-{}-{}".format(settings.CYBERSOURCE_REFERENCE_PREFIX, order.id)
+
+
+def get_new_order_by_reference_number(reference_number):
```
gsidebo: the name `get_new_order_by_reference_number` seems inaccurate. it's supposed to refer to an existing order, isn't it?

ecommerce/views_test.py
noisecapella: Will do. Regarding a custom exception, this case should never happen except for a coding error on our part so I feel like a 500 error would be appropriate here.

ecommerce/api.py
noisecapella: Yes, it's referring to a newly created order, one with a `status` of `CREATED`. What would you rename it to?

gsidebo: i meant that the `Order` object is assumed to exist when this method is called. otherwise line 203 would fail every time. after reading your comment, though, this seems like a good name

gsidebo:
looks good 👍  once tests pass

PR #914, opened 2016-08-25
added validation to the email composition dialog

### LINKED ISSUE #888: add validation to email composition
depends on #719

aliceriot:

this closes #888 

#### What's this PR do?

This adds a new function to `util/validation.js` to check the email fields, and then adds some UI to show the validation errors if they are present.

Read over the changes to `util/validation.js`. I had to refactor the email reducer a bit as well, probably important to look that over closely.

Go to the learners search. Click the 'Email Selected' button. If you then click 'send' you should see two validation errors (no subject and no body).

These errors should go away as you fix the problems. You shouldn't be able to 'send' the email (currently `send === console.log`) until the validation errors are cleared away.

email_validation

### DIFF
--- a/static/js/components/EmailCompositionDialog_test.js
+++ b/static/js/components/EmailCompositionDialog_test.js
@@ -5,6 +5,7 @@ import sinon from 'sinon';
 import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import getMuiTheme from 'material-ui/styles/getMuiTheme';
 import R from 'ramda';
+import TestUtils from 'react-addons-test-utils';
 
 import { NEW_EMAIL_EDIT } from '../reducers/email';
 import { modifyTextField } from '../util/test_utils';
@@ -19,10 +20,14 @@ describe('EmailCompositionDialog', () => {
     updateEmailEdit: sinon.stub().returns(updateStub),
     open: true,
     email: Object.assign({}, NEW_EMAIL_EDIT),
+    errors: {},
     searchkit: {
       getHitsCount: sinon.stub().returns(20)
-    }
+    },
+    sendEmail: sinon.stub(),
   };
+
+  const getDialog = () => document.querySelector('.email-composition-dialog');
 
   const renderDialog = (props = defaultProps) => (
     mount (
@@ -47,6 +52,12 @@ describe('EmailCompositionDialog', () => {
     );
   });
 
+  it('should fire sendEmail when the "send" button is clicked', () => {
+    renderDialog();
+    TestUtils.Simulate.click(getDialog().querySelector('.save-button'));
+    assert(defaultProps.sendEmail.called, "called sendEmail method");
+  });
+
   ['subject', 'body'].forEach(field => {
     describe(`editing ${field}`, () => {
       let getField = () => document.querySelector(`.email-${field}`);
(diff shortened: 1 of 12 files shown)


static/js/components/EmailCompositionDialog_test.js
```diff
@@ -18,12 +19,16 @@ describe('EmailCompositionDialog', () => {
     closeEmailDialog: sinon.stub().returns(updateStub),
     updateEmailEdit: sinon.stub().returns(updateStub),
     open: true,
-    email: Object.assign({}, NEW_EMAIL_EDIT),
+    email: Object.assign({}, NEW_EMAIL_EDIT.email),
```
noisecapella: I'm not sure what this is supposed to do, it looks like `email` doesn't exist on `NEW_EMAIL_EDIT`

static/js/containers/LearnerSearchPage.js
```diff
@@ -51,23 +57,31 @@ class LearnerSearchPage extends React.Component {
...
   closeEmailComposeAndSend: Function = () => {
-    const { dispatch, email } = this.props;
-    console.log(email); // eslint-disable-line no-console
-    dispatch(clearEmailEdit());
-    dispatch(setEmailDialogVisibility(false));
+    const { dispatch, email: { email } } = this.props;
+    let errors = emailValidation(email);
+    dispatch(updateEmailValidation(errors));
+    if ( R.isEmpty(errors) ) {
+      console.log(email); // eslint-disable-line no-console
```
noisecapella: Are we just leaving this here until an API for mail is in place?

static/js/reducers/email.js
```diff
@@ -3,18 +3,33 @@ import {
...
+export type Email = {
+    subject?:   ?string;
+    body?:      ?string;
+    query?:     ?Object;
+};
+
+export type EmailValidationErrors = {
+    subject?:   ?string;
+    body?:      ?string;
+    query?:     ?string;
```
noisecapella: How would I trigger a query error?

static/js/util/validation.js
```diff
@@ -204,6 +204,18 @@ export function privacyValidation(profile: Profile): ValidationErrors {
   return checkFieldPresence(profile, requiredFields, messages);
 }
 
+/**
+ * validate an email for presence of the 'subject' and 'body' fields
+ */
+export function emailValidation(email: Object): ValidationErrors {
```
noisecapella: Can we use the `Email` type instead of `Object`?

noisecapella:
Functionality looks great!

static/js/util/validation.js
aliceriot: Yes! I wrote this before I'd declared the type and forgot to update it.

static/js/reducers/email.js
aliceriot: As of now there isn't a way to, I had thought we might want to use this for reporting if the user is trying to send an email to 0 users.

static/js/components/EmailCompositionDialog_test.js
aliceriot: Oh yeah I think this is supposed to be just `NEW_EMAIL_EDIT`, oops

static/js/containers/LearnerSearchPage.js
aliceriot: yeah exactly, I merged it like this on Wednesday because I wanted to move forward on the UI work (and it's not so dependent on having the API in place anyway).

static/js/reducers/email.js
aliceriot: Do you think we should leave it for that reason, or take it out and add it back in if / when we implement that?

aliceriot:

static/js/reducers/email.js
noisecapella: It's fine, it doesn't look like it's causing any harm leaving it there

noisecapella:
:+1:

Issue #889, opened 2016-08-19
implement frontend interface for email API


aliceriot:
depends on #719

pdpinch:
Is this closed by #900 ?

aliceriot:
No, this is for writing the actual code to make the API requests.

pdpinch:
So it's blocked by #718 -- is that far enough along yet that this can be started?

@gsidebo

gsidebo:
not yet. PR will be up today

PR #916, opened 2016-08-26
Added API endpoint to send text email to a list of recipients

### LINKED ISSUE #718: Create API endpoint to send a text email to a list of recipients
_This issue is part of the actual implementation of #694_ 

An endpoint should be created in MM that takes an email message body (text only for now - no HTML) and a list of recipient email addresses, and sends an email.

We'll be using [Mailgun](https://mailgun.com/) as our email service. The actual send should be performed by [sending a POST request to Mailgun](https://documentation.mailgun.com/api-sending.html#sending). Configuring API communication with Mailgun will be part of this story. Devops will need to perform a few tasks to [verify our domain](https://documentation.mailgun.com/quickstart-sending.html#verify-your-domain) for sending, and we should also talk to them about the necessity of creating a new subdomain for these emails (as Mailgun [recommends](https://documentation.mailgun.com/best_practices.html#dns)).

These emails should no be reply-able, so you'll need to figure out what to do about the From: address. Mailgun may require it, and if so, we should use some sensible "no-reply" MIT address if one exists.

inputs: message subject, message body, elasticsearch query/parameters

- [ ] create and enforce permissions
- [ ] use a no-reply address for the return address
- [x] ~~use a basic template, with footer (this can be a hardcoded string)~~ _This can be done as part of #805_
- [ ] retrieve the set of users' email addresses from elasticsearch (hopefully this is just a matter of setting the results per page on elastic search to the max)
- [ ] post to Mailgun
- [ ] In our models (#830), save some reliable reference to the message as it was sent in Mailgun _(see below)_

### Referring to Mailgun messages from our data

Mailgun has plenty of documentation around querying their API for email data (like open rate) at the _account_ level, but it was surprisingly difficult to figure out how to use the API to get the same data for a single email. We should look into this more, but we may be able to accomplish this by using [Mailgun tags](https://documentation.mailgun.com/api-tags.html#tags). Basically we would assign a custom tag to every mail we send, then query this Tag API endpoint with one of those custom tags to get data for a single email

gsidebo:

Closes #718 

#### What's this PR do?

PR title says it all. This will be used by the "Email selected learners" button on the search results page. One of the key parts of the new mail API endpoint is that it fetches all of the email addresses that match a given search query.

Probably search/api.py, followed by mail/api.py and mail/views.py

This isn't hooked up to the front end yet, so it can't be tested from the web app. The way I tested this was very rudimentary - I'll try to think of an easier way

I moved search/api.py and search/api_test.py to search/indexing_api.py and search/indexing_api_test.py. This PR adds functions that execute actual searches, and search/api.py only contained functions that related to setting up the ES index. **search/api.py and search/api_test.py should probably be viewed in the raw view (not diff), and search/indexing_api.py and search/indexing_api_test.py can be ignored since that functionality was just moved from those /api* files**


gsidebo:
FYI: I still need to add a couple tests to mail/views_test.py and make the linter happy. I wanted to get this PR up in the meantime to get some eyes on it.

search/api.py
```diff
@@ -1,306 +1,90 @@
-"""
```
gsidebo: probably best to view this on its own. its contents were moved to search/indexing_api.py. the contents here are new and unrelated

search/api_test.py
```diff
@@ -1,365 +1,95 @@
-"""
```
gsidebo: probably best to view this on its own. its contents were moved to search/indexing_api_test.py. the contents here are new and unrelated

gsidebo:
tests added, linting fixed, rebased

noisecapella:
Can you update .env.example with the new environment variables?

mail/api.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Provides functions for communicating with the Mailgun REST API.
+    """
+    _basic_auth_credentials = ('api', settings.MAILGUN_KEY)
+    _base_params = {'from': settings.MAILGUN_FROM_EMAIL}
+
+    @classmethod
+    def _mailgun_request(cls, request_func, endpoint, params):
+        """
+        Sends a request to the Mailgun API
```
noisecapella: Can you add information about each argument and the return value?

mail/api_test.py
```diff
@@ -0,0 +1,44 @@
...
+class MailAPITests(TestCase):
+    def test_send_bcc(self, mock_post):  # pylint: disable=no-self-use
+        """
+        Test that MailgunClient.send_bcc sends expected parameters to the Mailgun API
+        """
+        MailgunClient.send_bcc('email subject', 'email body', 'a@example.com,b@example.com')
+        assert mock_post.called
+        called_args, called_kwargs = mock_post.call_args  # pylint: disable=unused-variable
+        assert list(called_args)[0] == '{}/{}'.format(settings.MAILGUN_URL, 'messages')
+        assert 'auth' in called_kwargs
```
noisecapella: Can we override the setting `MAILGUN_KEY` and then assert this exact value?

mail/api_test.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+from unittest.mock import patch
+from django.test import TestCase
+from django.conf import settings
+from django.test.utils import override_settings
+
+from mail.api import MailgunClient
+
+
+@patch('requests.post')  # pylint: disable=missing-docstring
```
noisecapella: Can we assert `to` and `from` in one of these tests?

pdpinch:
this isn't passing, but Travis isn't updating the PR for some reason

micromasters/urls.py
```diff
@@ -27,8 +28,8 @@
     url(r'^api/v0/dashboard/$', UserDashboard.as_view(), name='dashboard_api'),
     url(r'^api/v0/search/(?P<elastic_url>.*)', ElasticProxyView.as_view(), name='search_api'),
     url(r'^api/v0/checkout/$', CheckoutView.as_view(), name='checkout'),
+    url(r'^api/v0/mail$', MailView.as_view(), name='mail_api'),
```
noisecapella: Can you include a trailing slash here?

search/api.py
```diff
@@ -1,306 +1,142 @@
...
-    """
-    Remove a user from Elasticsearch.
+    Returns:
+        Search: elasticsearch_dsl Search object
     """
-    program_enrollments = ProgramEnrollment.objects.filter(user=user).select_related('user', 'program').all()
-    for program_enrollment in program_enrollments:
-        remove_program_enrolled_user(program_enrollment)
+    search_obj = Search(index=settings.ELASTICSEARCH_INDEX, doc_type=DOC_TYPES)
+    if search_param_dict:
```
noisecapella: Can you make this `if search_param_dict is not None`?

search/api.py
```diff
@@ -1,306 +1,142 @@
...
-            "match_mapping_type": "string",
-            "mapping": NOT_ANALYZED_STRING_TYPE
-        }
-    }])
-    return mapping
+    def add_to_results(self, es_search_hits):
+        """
+        Adds a set of search results ('hits') to the combined results
+        """
+        self.results = self.results + list(es_search_hits)
```
noisecapella: The `x = x + y` pattern has performance problems here because `x + y` creates a new list before assigning back to `x`. You may want to do `self.results += list(es_search_hits)` instead.

search/api.py
```diff
@@ -1,306 +1,142 @@
...
     """
-    Create all mappings, deleting existing mappings if they exist.
+    Provides a function to return all unique emails for documents that match
+    an ES query
     """
-    create_program_enrolled_user_mapping()
+    def init_combined_results(self):
+        self.results = set()
 
+    def add_to_results(self, es_search_hits):
```
noisecapella: What does this do differently than in the parent class?

search/api.py
```diff
@@ -1,306 +1,142 @@
...
+        Adds a set of search results ('hits') to the combined results
+        """
+        self.results = self.results + list(es_search_hits)
 
+    def execute(self, search_obj, page_size=DEFAULT_ES_LOOP_PAGE_SIZE):
+        """
+        Executes a search query, pages through all of its results, maintains
+        a record of all results found, and returns those combined results.
+        """
+        self.init_combined_results()
```
noisecapella: What's the use of having `init_combined_results` and `add_to_results` defined as separate methods, if they're only used in this method?

search/api_test.py
```diff
@@ -1,365 +1,122 @@
...
-        Test that Employment is removed from index after being deleted
+        Test that Search objects are created with program-limiting query parameters
         """
-        program_enrollment = ProgramEnrollmentFactory.create()
-        employment = EmploymentFactory.create(profile=program_enrollment.user.profile)
-        assert_search(es.search(), [program_enrollment])
-        employment.delete()
-        assert_search(es.search(), [program_enrollment])
+        search_obj = create_search_obj(self.user)
+        # assert set(search_obj._doc_type) == set(DOC_TYPES)
```
noisecapella: Can you remove these two lines?

static/js/util/api.js
```diff
@@ -128,3 +128,14 @@ export function checkout(courseId: string): Promise<CheckoutResponse> {
     })
   });
 }
+
+export function sendMail() {
```
noisecapella: This needs tests. Also, is this used anywhere?

gsidebo: not used anywhere as part of this PR. i added this as a stub to make the actual integration easier. the actual issue (which i'm working on now): https://github.com/mitodl/micromasters/issues/889

mail/views_test.py
```diff
@@ -0,0 +1,79 @@
...
+        super(MailViewsTests, self).setUp()
+        self.client.force_login(self.staff)
+
+    def test_send_view(self, mock_mailgun_client, mock_prepare_exec_search):
+        """
+        Test that the email send view will accept and return expected values
+        """
+        email_results = ['a@example.com', 'b@example.com']
+        mock_prepare_exec_search.return_value = email_results
+        mock_mailgun_client.send_bcc.return_value = Mock(spec=Response, status_code=200)
```
noisecapella: Can you use the enums in `rest_framework.status` to replace the HTTP status codes in this file?

mail/views.py
```diff
@@ -0,0 +1,44 @@
...
+    authentication_classes = (authentication.SessionAuthentication, )
+    permission_classes = (permissions.IsAuthenticated, UserCanMessageLearnersPermission, )
+
+    def post(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-self-use, missing-docstring
+        emails = prepare_and_execute_search(
+            request.user,
+            search_param_dict=request.data['search_request'],
+            search_func=get_all_query_matching_emails
+        )
+        mailgun_resp = MailgunClient.send_bcc(
```
noisecapella: Does mailgun ever have anything of value in the response content?

micromasters/urls.py
gsidebo: here's why i didn't add the slash:

```
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 123, in get_response
    response = middleware_method(request)
  File "/usr/local/lib/python3.5/site-packages/django/middleware/common.py", line 62, in process_request
    path = self.get_full_path_with_slash(request)
  File "/usr/local/lib/python3.5/site-packages/django/middleware/common.py", line 100, in get_full_path_with_slash
    'url': request.get_host() + new_path,
RuntimeError: You called this URL via POST, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining POST data. Change your form to point to 192.168.99.100:8079/api/v0/mail/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.
```

i'm happy to add the `APPEND_SLASH` setting if that's what we want to go with. i went with the path of least resistance for this PR

mail/views_test.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Test that the status code from Mailgun's API response will be returned in our own
+         response
+        """
+        mock_prepare_exec_search.return_value = []
+        mock_mailgun_client.send_bcc.return_value = Mock(spec=Response, status_code=400)
+        resp_post = self.client.post(self.mail_url, data=self.request_data, format='json')
+        assert resp_post.status_code == 400
+
+    def test_no_program_user_response(self, *args):  # pylint: disable=unused-argument
```
noisecapella: Can you also test that a user can be staff of a different program and not have access to this program?

search/api.py
gsidebo: `self.results` is a list in the parent class. it's a set in this class. there's a different API for adding to a set vs. adding to a list, hence the overridden method

search/api.py
gsidebo: `init_combined_results` and `add_to_results` are only called within `execute` but they're implemented differently in the two classes. `execute` is almost identical in both classes

mail/views.py
gsidebo: for our purposes, i haven't seen anything compelling. we haven't looked into or decided what we want in terms of alerting the user when they send an email. if we want to do some specific notifications, we can look deeper into the mailgun response. right now we dont tell the user _anything_ about their attempted send

mail/views_test.py
gsidebo: will do

micromasters/urls.py
noisecapella: If you add the slash here and to the JS at the same time I don't think you should get this error. All of our other URLs have a trailing slash so it would be nice to be consistent

search/api.py
noisecapella: Why not remove duplicates in `execute` instead?

search/api_test.py
```diff
@@ -1,365 +1,122 @@
...
-        program_enrollment = ProgramEnrollmentFactory.create()
-        assert es.search()['total'] == 1
-        employment = EmploymentFactory.create(profile=program_enrollment.user.profile)
-        employment.city = 'city'
-        employment.save()
-        assert_search(es.search(), [program_enrollment])
+        search_obj = Search()
+        search_obj.execute = Mock(name='execute')
+        execute_search(search_obj)
+        assert search_obj.execute.called
```
noisecapella: Can you assert that `get_conn` is called too?

search/api_test.py
```diff
@@ -1,365 +1,122 @@
...
         """
-        Test that recreate_index will create an index and let search successfully
+        Test that a Search object is properly prepared and executed
         """
-        recreate_index()
-        assert es.search()['total'] == 0
+        mock_search_func = Mock(name='execute', return_value=['result1', 'result2'])
+        with patch('search.api.create_search_obj', autospec=True, return_value=None) as mock_create_search_obj:
+            results = prepare_and_execute_search(self.user, search_func=mock_search_func)
+            assert mock_create_search_obj.called
```
noisecapella: Can you assert the call args too?

gsidebo:
@noisecapella all of the comments should be addressed

search/api_test.py
```diff
@@ -1,365 +1,122 @@
...
         """
-        Test that Employment is removed from index after being deleted
+        Test that Search objects are created with program-limiting query parameters
         """
-        program_enrollment = ProgramEnrollmentFactory.create()
-        employment = EmploymentFactory.create(profile=program_enrollment.user.profile)
-        assert_search(es.search(), [program_enrollment])
-        employment.delete()
-        assert_search(es.search(), [program_enrollment])
+        search_obj = create_search_obj(self.user)
```
noisecapella: It doesn't look like there are any `create_search_obj` tests using either `search_param_dict` or `search_func`. Can you add tests or update existing tests to cover these parameters?

search/api_test.py
```diff
@@ -1,365 +1,122 @@
...
-        recreate_index()
-        assert es.search()['total'] == 0
+        mock_search_func = Mock(name='execute', return_value=['result1', 'result2'])
+        with patch('search.api.create_search_obj', autospec=True, return_value=None) as mock_create_search_obj:
+            results = prepare_and_execute_search(self.user, search_func=mock_search_func)
+            assert mock_create_search_obj.called
+            assert results == ['result1', 'result2']
 
-    def test_update_index(self):  # pylint: disable=no-self-use
+    def test_all_query_matching_emails(self):  # pylint: disable=no-self-use
```
noisecapella: Do we use `get_all_query_matching_results` anywhere? Or `ESFullResultSearch`?

search/api.py
```diff
@@ -1,306 +1,142 @@
...
-    """
-    conn = get_conn(verify=False)
-    index_name = settings.ELASTICSEARCH_INDEX
-    if conn.indices.exists(index_name):
-        conn.indices.delete(index_name)
-    conn.indices.create(index_name)
-    conn.indices.refresh()
-    create_mappings()
+    def execute(self, search_obj, page_size=DEFAULT_ES_LOOP_PAGE_SIZE):
+        search_obj = search_obj.fields('email')
```
noisecapella: The code coverage doesn't show these two lines being executed during the tests... I think your `__call__` assignment is not inherited here

search/api_test.py
```diff
@@ -1,365 +1,122 @@
...
-        # recreate_index should index the program-enrolled user
-        recreate_index()
-        assert_search(es.search(), [program_enrollment])
+        fake_results = FakeSearchResults()
+        mock_execute_search = Mock(spec=execute_search, return_value=fake_results)
+        test_es_page_size = 2
+        with patch('search.api.execute_search', mock_execute_search):
+            results = get_all_query_matching_emails(Search(), test_es_page_size)
+            assert mock_execute_search.call_count == math.ceil(fake_results.hits.total/test_es_page_size)
+            assert results == set([result.email for result in fake_results.results])
```
noisecapella: Can you add some assertion to make sure the `search_obj = search_obj.fields('email')` from `ESFullResultEmailSearch` was executed correctly?

search/api_test.py
gsidebo: not used yet, no. i initially went about this trying to solve the general problem of taking a search query and paging through all of its results. i adapted that to the problem of paging through results and just retrieving the unique emails. having the 'base' class (`ESFullResultSearch`) is going to be beneficial in future issues for micromasters email. i'm happy to add a test case or two to assert that its working correctly

search/api.py
noisecapella: My main question here is, why not just have `get_all_query_matching_emails` be a function? The two classes here and the inheritance relationship doesn't seem to add anything

search/api.py
gsidebo: i'll look into it

noisecapella:
We should also consider moving the bulk of the work to Celery as @giocalitri mentioned, but I'm happy if we dealt with that in a separate PR

giocalitri:
I agree: let's open an issue and postpone the  celery part.

On Aug 29, 2016, at 2:40 PM, George Schneeloch <notifications@github.com<mailto:notifications@github.com>> wrote:


We should also consider moving the bulk of the work to Celery as @giocalitri<https://github.com/giocalitri> mentioned, but I'm happy if we dealt with that in a separate PR

mail/views_test.py
gsidebo: actually, we're already asserting this in `test_create_search_obj_program_limit`. that asserts that the query parameters to limit a user to their programs will be applied. if we test that the actual response from ES comes back only with data for the given programs, then we're testing ES/elasticsearch_dsl and i don't think we need to be doing that

gsidebo:
Created an issue for the celery task #949

gsidebo:
@noisecapella let me know if you spot anything else

.env.example
```diff
@@ -10,3 +10,4 @@ CYBERSOURCE_TRANSACTION_KEY=
 CYBERSOURCE_SECURE_ACCEPTANCE_URL=
 CYBERSOURCE_PROFILE_ID=
 CYBERSOURCE_REFERENCE_PREFIX=
+MAILGUN_RECIPIENT_OVERRIDE=
```
noisecapella: Can you add the other mailgun settings too?

micromasters/urls.py
noisecapella: Any response on this?

search/api.py
noisecapella: Any response here?

noisecapella:
Sorry for the delay here, got bogged down with server side validation

mail/views_test.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        Test that a 403 will be returned when a user with inadequate permissions attempts
+        to send an email through the email send view
+        """
+        with mute_signals(post_save):
+            no_permissions_profile = ProfileFactory.create()
+        self.client.force_login(no_permissions_profile.user)
+        resp_post = self.client.post(self.mail_url, data=self.request_data, format='json')
+        assert resp_post.status_code == 403
```
noisecapella: Can you replace this status code too?

search/api.py
gsidebo: sorry, didn't see the comments in response. `get_all_query_matching_emails` _is_ effectively a function. you can call it, and indeed it is being called in mail/views.py. what we get from the inheritance relationship is code reuse and, in general, the representation of a concept in code. whether we want to retrieve (a) the full ES documents that match a query, or (b) the full set of email addresses in those documents that match a query, that process is very similar. in both cases (to quote the docstring), we "execute a search query, page through all of its results, maintain a record of results found, and return those combined results". what's different between a and b? (1) the data structure that is most appropriate to store the results, and (2) the way that results are added when a page of ES documents is retrieved. hence, the two methods that are implemented differently in those two classes. in the case of `get_all_query_matching_emails`, duplicates could indeed be removed as part of `execute`, but that duplicate removal is irrelevant to `get_all_query_matching_results`. `execute` is largely the same process for both, so inheritance/code reuse is appropriate

if you still don't agree with or understand what i'm going for, i'm happy to chat in person

.env.example
```diff
@@ -10,3 +10,5 @@ CYBERSOURCE_TRANSACTION_KEY=
 CYBERSOURCE_SECURE_ACCEPTANCE_URL=
 CYBERSOURCE_PROFILE_ID=
 CYBERSOURCE_REFERENCE_PREFIX=
+MAILGUN_KEY=
+MAILGUN_RECIPIENT_OVERRIDE=
```
gsidebo: @noisecapella these are the only two Mailgun settings values that really make sense to expect from someone running MM locally. let me know if you think something should be added

search/api.py
gsidebo: is `__call__ = execute` causing the confusion here by any chance? that's what is making `get_all_query_matching_results` and `get_all_query_matching_emails` callable. after this declaration - `get_all_query_matching_emails = ESFullResultEmailSearch()` - you can invoke `get_all_query_matching_emails` like you were invoking `ESFullResultEmailSearch.execute`.

search/api.py
```diff
@@ -1,306 +1,143 @@
...
+    Provides a function to return all unique emails for documents that match
+    an ES query
     """
-    conn = get_conn(verify=False)
-    index_name = settings.ELASTICSEARCH_INDEX
-    if conn.indices.exists_type(index=index_name, doc_type=USER_DOC_TYPE):
-        conn.indices.delete_mapping(index=index_name, doc_type=USER_DOC_TYPE)
-    mapping = program_enrolled_user_mapping()
-    mapping.save(index_name)
+    def init_combined_results(self):
```
noisecapella: Can you change this and `add_to_results` to have an underscore before them if they're not meant to be used outside the class?

gsidebo:
@noisecapella you convinced me 😄 . future needs for paging through ES results are likely to be very similar to the email use case. i saved that snippet if we need it down the road, but given the known use cases in the future, i think its appropriate to keep it simpler for now

noisecapella:
Sounds good, I'll take a one more quick look through

noisecapella:
I'm getting a problem in manual testing. When I send this JSON to `/api/v0/mail/` as a POST:

    {"email_subject": "Subject", "search_request": {}, "email_body": "Body"}

I get this exception:

    web_1     | [31/Aug/2016 17:30:59] "POST /api/v0/mail/ HTTP/1.1" 500 23857
    web_1     | [31/Aug/2016 17:30:59] "GET /api/v0/mail/ HTTP/1.1" 405 5053
    web_1     | [2016-08-31 17:34:19] ERROR 20 [django.request] base.py:284 - [cd21f573fcde] - Internal Server Error: /api/v0/mail/
    web_1     | Traceback (most recent call last):
    web_1     |   File "/usr/local/lib/python3.5/site-packages/elasticsearch_dsl/utils.py", line 120, in __getattr__
    web_1     |     return _wrap(self._d_[attr_name])
    web_1     | KeyError: 'email'
    web_1     | 
    web_1     | During handling of the above exception, another exception occurred:
    web_1     | 
    web_1     | Traceback (most recent call last):
    web_1     |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 149, in get_response
    web_1     |     response = self.process_exception_by_middleware(e, request)
    web_1     |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 147, in get_response
    web_1     |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
    web_1     |   File "/usr/local/lib/python3.5/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    web_1     |     return view_func(*args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/base.py", line 68, in view
    web_1     |     return self.dispatch(request, *args, **kwargs)
    web_1     |   File "/usr/local/lib/python3.5/site-packages/rest_framework/views.py", line 466, in dispatch
    web_1     |     response = self.handle_exception(exc)
    web_1     |   File "/usr/local/lib/python3.5/site-packages/rest_framework/views.py", line 463, in dispatch
    web_1     |     response = handler(request, *args, **kwargs)
    web_1     |   File "/src/mail/views.py", line 35, in post
    web_1     |     search_func=get_all_query_matching_emails
    web_1     |   File "/src/search/api.py", line 78, in prepare_and_execute_search
    web_1     |     return search_func(search_obj)
    web_1     |   File "/src/search/api.py", line 105, in get_all_query_matching_emails
    web_1     |     results.add(hit.email[0])
    web_1     |   File "/usr/local/lib/python3.5/site-packages/elasticsearch_dsl/utils.py", line 123, in __getattr__
    web_1     |     '%r object has no attribute %r' % (self.__class__.__name__, attr_name))
    web_1     | AttributeError: 'Result' object has no attribute 'email'

gsidebo:
looking into this now. can you tell me how you manually tested this? i'm trying to repro this myself but ran into a different, very WTF error

mail/api.py
```diff
@@ -0,0 +1,66 @@
+"""
+Provides functions for sending and retrieving data about in-app email
+"""
+import requests
+
+from micromasters.settings import (
```
noisecapella: `settings` isn't a package and we shouldn't import directly from `micromasters`. Instead this should be:

https://docs.djangoproject.com/en/1.10/topics/settings/#using-settings-in-python-code

mail/api_test.py
```diff
@@ -0,0 +1,55 @@
...
+class MailAPITests(TestCase):
+    def test_from_address(self, mock_post):  # pylint: disable=no-self-use
+        """
+        Test that the 'from' address for our emails is set correctly
+        """
+        # NOTE: Using patch.multiple to override settings values because Django's
+        # override_settings decorator fails to work for mysterious reasons
+        with patch.multiple('mail.api', MAILGUN_FROM_EMAIL='mailgun_from_email@example.com'):
```
noisecapella: Instead of using `patch.multiple` you should use [override_settings](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.override_settings) here and below. You can use it as a decorator or as a context manager like here

mail/api_test.py
```diff
@@ -0,0 +1,55 @@
+"""
+Test cases for email API
+"""
+from unittest.mock import patch
+from django.test import TestCase
+
+from micromasters import settings
```
noisecapella: Same comment on `from django.conf import settings`

mail/api_test.py
gsidebo: read the comment directly above. it simply isn't working. if you want to try it locally, please do. i dumped at least an hour trying to get that decorator to work

noisecapella: Did you try replacing the `from micromasters import settings` above first?

gsidebo: welp... just got it to work. i just dont know anymore...

noisecapella:
Can you rebase? The changes you made look fine, if the tests pass feel free to merge :+1:

PR #952, opened 2016-08-30
add celerybeat-schedule to gitignore

aliceriot:
I believe there are plans to fix this issue, but I think we should add this until we sort out the fix.


noisecapella:
:+1:

PR #956, opened 2016-08-31
Added some server side verification for profiles

### LINKED ISSUE #847: In profile API, validate that profile
Implement server side validation based on the value of filled_out

noisecapella:
Fixes #847

#### What's this PR do?
If `filled_out` is true or is being set to true, it does some extra validation on the profile data. See the issue to see what is being validated and what is skipped.

In the user page for your own user, try clearing the end date for some field of employment. You should still be able to.

Then, go to `/api/v0/profiles/your_user_name_here/`. Try to PATCH a profile to clear a field, eg

    {
      "country": ""
    }

You should see 'This profile may not be blank'

### DIFF
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -1,7 +1,6 @@
 """
 Serializers for user profiles
 """
-
 from django.db import transaction
 
 from rest_framework.exceptions import ValidationError
@@ -96,6 +95,38 @@ class EmploymentSerializer(ModelSerializer):
         )
 
 
+def set_fields_to_required(serializer, ignore_fields=None):
+    """
+    Iterate through fields in serializer and set all to required except ignore_fields
+
+    Args:
+        serializer (rest_framework.serializers.Serializer):
+            A serializer
+        ignore_fields (list of str):
+            If not none, a list of field names to skip
+    Returns:
+        None
+    """
+    if ignore_fields is None:
+        ignore_fields = []
+    for field in serializer.fields.values():
+        if field.field_name not in ignore_fields:
+            field.required = True
+            field.allow_null = False
+            field.allow_blank = False
+
+
+class EmploymentFilledOutSerializer(EmploymentSerializer):
+    """Serializer for Employment objects in filled out Profiles"""
+
+    def __init__(self, *args, **kwargs):
+        """
+        Update serializer_field_mapping to use fields setting required=True
+        """
+        set_fields_to_required(self, ['end_date'])
+        super(EmploymentFilledOutSerializer, self).__init__(*args, **kwargs)
+
+
 class EducationSerializer(ModelSerializer):
     """Serializer for Education objects"""
     id = IntegerField(required=False)  # override the read_only flag so we can edit it
@@ -112,6 +143,17 @@ class EducationSerializer(ModelSerializer):
             'school_city',
             'school_state_or_territory',
             'school_country')
+
+
+class EducationFilledOutSerializer(EducationSerializer):
+    """Serializer for Education objects in filled out Profiles"""
+
+    def __init__(self, *args, **kwargs):
+        """
+        Update serializer_field_mapping to use fields setting required=True
+        """
+        set_fields_to_required(self, ['field_of_study'])
+        super(EducationFilledOutSerializer, self).__init__(*args, **kwargs)
 
 
 class ProfileBaseSerializer(ModelSerializer):
@@ -226,3 +268,28 @@ class ProfileLimitedSerializer(ProfileBaseSerializer):
             'edx_level_of_education',
             'education'
         )
+
+
+class ProfileFilledOutSerializer(ProfileSerializer):
+    """Serializer for Profile objects which require filled_out = True"""
+    work_history = EmploymentFilledOutSerializer(many=True)
+    education = EducationFilledOutSerializer(many=True)
+
+    def __init__(self, *args, **kwargs):
+        """
+        Update serializer_field_mapping to use fields setting required=True
+        """
+        set_fields_to_required(self)
+        super(ProfileFilledOutSerializer, self).__init__(*args, **kwargs)
+
+    def validate(self, attrs):
+        """
+        Assert that filled_out can't be turned off and that agreed_to_terms_of_service is true
+        """
+        if 'filled_out' in attrs and not attrs['filled_out']:
+            raise ValidationError("filled_out cannot be set to false")
+
+        if 'agreed_to_terms_of_service' in attrs and not attrs['agreed_to_terms_of_service']:
+            raise ValidationError("agreed_to_terms_of_service cannot be set to false")
+
+        return super(ProfileFilledOutSerializer, self).validate(attrs)
--- a/profiles/serializers_test.py
+++ b/profiles/serializers_test.py
@@ -1,10 +1,17 @@
 """
 Tests for profile serializers
 """
+from copy import deepcopy
 
 from django.db.models.signals import post_save
 from factory.django import mute_signals
-from rest_framework.fields import DateTimeField
+from rest_framework.fields import (
+    CharField,
+    DateTimeField,
+    ReadOnlyField,
+    SerializerMethodField,
+)
+from rest_framework.serializers import ListSerializer
 from rest_framework.exceptions import ValidationError
 
 from backends.edxorg import EdxOrgOAuth2
@@ -24,6 +31,7 @@ from profiles.serializers import (
     EmploymentSerializer,
     ProfileLimitedSerializer,
     ProfileSerializer,
+    ProfileFilledOutSerializer,
 )
 from profiles.util import (
     GravatarImgSize,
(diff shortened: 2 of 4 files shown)


profiles/serializers.py
```diff
@@ -227,3 +260,32 @@ class Meta:  # pylint: disable=missing-docstring
...
+    education = EducationFilledOutSerializer(many=True)
+
+    def __init__(self, *args, **kwargs):
+        """
+        Update serializer_field_mapping to use fields setting required=True
+        """
+        for field in self.fields.values():
+            field.required = True
+            field.allow_null = False
+            field.allow_blank = False
```
gsidebo: we really should put this loop into a separate method. it's repeated 3 times. this method could do the trick.

```python
def set_fields_to_required(self, ignore_fields=None):
    if ignore_fields:
        fields_to_require = list(field for name, field in serializer.fields.items() if name not in ignore_fields)
    else:
        fields_to_require = serializer.fields.values()
    for field in fields_to_require:
        field.required = True
        field.allow_null = False
        field.allow_blank = False
```

one option would be to add a mixin with this function and inherit that in each of these filled out serializers

profiles/serializers_test.py
```diff
@@ -304,3 +309,158 @@ def test_delete_employment(self):
...
+        # No exception raised
+
+    def test_education(self):
+        """
+        All fields are required after a profile is marked filled_out
+        """
+        profile = self.create_profile()
+
+        serializer = ProfileFilledOutSerializer(profile)
+        data = serializer.to_representation(profile)
```
gsidebo: every single test case in this class opens with these 3 lines. at the very least, these should be added to setUp

profiles/serializers_test.py
```diff
@@ -304,3 +309,158 @@ def test_delete_employment(self):
 
         # Other profile is unaffected
         assert employment3.profile.work_history.count() == 1
+
+
+class ProfileFilledOutTests(ESTestCase):
```
gsidebo: this question surely applies to other test classes in the codebase, but why is this inheriting from ESTestCase? is it testing any ES functionality? does it require indexing to be done?

noisecapella: Yes, whenever a profile is created it is indexed. There is no ES functionality being tested

profiles/serializers.py
noisecapella: Sure, I'll investigate this

profiles/serializers_test.py
noisecapella: Will do

profiles/serializers_test.py
```diff
@@ -304,3 +309,158 @@ def test_delete_employment(self):
...
+            'username',
+            'profile_url_full',
+            'profile_url_large',
+            'profile_url_medium',
+            'profile_url_small',
+            'pretty_printed_student_id',
+            'work_history',
+            'education',
+        }
+        for key in keys:
```
gsidebo: this loop is repeated with some minor changes in test_work_history and test_education. would strongly prefer that we abstract it - they are testing the same general idea, after all. i'll think about a good way to go about it

profiles/serializers_test.py
gsidebo: then this could inherit from MockedESTestCase, right?

noisecapella: Yes, after that PR gets merged

profiles/serializers.py
```diff
@@ -96,6 +95,39 @@ class Meta:
...
+
+class EmploymentFilledOutSerializer(EmploymentSerializer):
+    """Serializer for Employment objects in filled out Profiles"""
+
+    def __init__(self, *args, **kwargs):
+        """
+        Update serializer_field_mapping to use fields setting required=True
+        """
+        set_fields_to_required(self, ['end_date'])
+
```
gsidebo: unnecessary blank line=

profiles/serializers_test.py
```diff
@@ -304,3 +309,134 @@ def test_delete_employment(self):
...
+        keys = set(ProfileFilledOutSerializer.Meta.fields) - {
+            'username',
+            'profile_url_full',
+            'profile_url_large',
+            'profile_url_medium',
+            'profile_url_small',
+            'pretty_printed_student_id',
+            'work_history',
+            'education',
+        }
```
gsidebo: this set of readonly field names seems to be copied in ProfilePATCHTests.test_patch_own_profile. we should maintain this in some common class. does it make sense to define it in ProfileSerializer or ProfileFilledOutSerializer?

profiles/serializers_test.py
```diff
@@ -304,3 +309,134 @@ def test_delete_employment(self):
...
+                    ProfileFilledOutSerializer(data=clone).is_valid(raise_exception=True)
+                assert ex.exception.args[0] == {"work_history": [{key: ['This field may not be blank.']}]}
+
+    def test_work_history_end_date(self):
+        """
+        Make sure end_date can be set to null on filled out profiles
+        """
+        self.data['work_history'][0]['end_date'] = None
+        ProfileFilledOutSerializer(data=self.data).is_valid(raise_exception=True)
+        # No exception raised
```
gsidebo: probably should be deleted

profiles/views_test.py
```diff
@@ -64,6 +65,27 @@ def setUp(self):
             )
         self.url2 = reverse('profile-detail', kwargs={'user': username})
 
+        # Monkey patch get_serializer_class so its return value can be asserted. This is reverted in tearDown
```
gsidebo: can you help me understand the intention here? the comment indicates that we want to assert the return value for `get_serializer_class`, but i don't see where we're actually making that assertion

noisecapella: `self.returned_serializer` is asserted in a couple places, for example https://github.com/mitodl/micromasters/pull/956/files/cef11cc3a8a70363495f85b838ac26e0d3cce0d1#diff-fe503725d8f59174a69abe2613648b44R326

profiles/serializers_test.py
noisecapella: I'll look into cleaning this up

profiles/views_test.py
gsidebo: sorry, missed that

profiles/serializers_test.py
gsidebo: whipped this together:

```python
class ReqProfileFieldTester:
    def __init__(self, test_obj):
        self.test = test_obj
        self.clone = {}

    def expected_req_field_exception(self, field_key, message):
        return {field_key: [message]}

    def set_value_at_key(self, field_key, value):
        self.clone[field_key] = value

    def get_field_from_key(self, field_key):
        return ProfileFilledOutSerializer(data=self.clone).fields[field_key]

    def check_required_fields(self, keys, data):
        for key in keys:
            self.clone = deepcopy(data)
            self.set_value_at_key(key, None)
            with self.test.assertRaises(ValidationError) as ex:
                ProfileFilledOutSerializer(data=self.clone).is_valid(raise_exception=True)
            assert ex.exception.args[0] == self.expected_req_field_exception(
                key, 'This field may not be null.'
            )

            field = self.get_field_from_key(key)
            if isinstance(field, CharField):
                # test blank string too
                self.set_value_at_key(key, '')
                with self.test.assertRaises(ValidationError) as ex:
                    ProfileFilledOutSerializer(data=self.clone).is_valid(raise_exception=True)
                assert ex.exception.args[0] == self.expected_req_field_exception(
                    key, 'This field may not be blank.'
                )


class ReqProfileNestedFieldTester(ReqProfileFieldTester):
    def __init__(self, test_obj, nested_key=''):
        self.nested_key = nested_key
        super(ReqProfileNestedFieldTester, self).__init__(test_obj)

    def expected_req_field_exception(self, field_key, message):
        return {self.nested_key: [{field_key: [message]}]}

    def set_value_at_key(self, field_key, value):
        self.clone[self.nested_key][0][field_key] = value

    def get_field_from_key(self, field_key):
        return ProfileFilledOutSerializer(data=self.clone).fields[self.nested_key].child.fields[field_key]
```

invocation changes to:

```python
req_field_tester = ReqProfileFieldTester(self)
req_field_tester.check_required_fields(keys, self.data)
# or...
req_field_tester = ReqProfileNestedFieldTester(self, nested_key='work_history')
req_field_tester.check_required_fields(keys, self.data)
```

i'm sort of on the line as to whether or not this is overkill. all i know is that this functionality (iterating through fields, asserting the right validation errors with a none value or blank value for character fields) is general enough and complex enough that it's probably worth some abstraction and not preferable to repeat it 3 times over

profiles/views_test.py
```diff
@@ -35,16 +42,16 @@
...
     """
 
     def setUp(self):
         """
         Create a user and profile
         """
-        super(ProfileTests, self).setUp()
+        super(ProfileBaseTests, self).setUp()
 
         with mute_signals(post_save):
```
gsidebo: just curious: shouldn't the creation of `user1` and `user2` etc live inside of `setUpTestData` rather than `setUp`?

noisecapella: Yes, I'll move them

profiles/views_test.py
```diff
@@ -292,3 +280,62 @@ def test_staff_sees_entire_profile(self):
...
+            new_profile = ProfileFactory.create(filled_out=False)
+        new_profile.user.social_auth.create(
+            provider=EdxOrgOAuth2.name,
+            uid="{}_edx".format(new_profile.user.username)
+        )
+        patch_data = ProfileSerializer().to_representation(new_profile)
+
+        resp = self.client.patch(self.url1, content_type="application/json", data=json.dumps(patch_data))
+        assert resp.status_code == 200
+        assert self.returned_serializer is ProfileSerializer
```
gsidebo: these test for the serializer used are a little convoluted, and in this case, sort of shoe-horned into a case that is testing something else. i think we should get rid of the monkey patch and create test cases that focus directly on the serializer use. you could start with something like:

```python
class ProfileSerializerTests(ProfileBaseTests):
    def set_profile_and_login(self, **profile_kwargs):
        with mute_signals(post_save):
            ProfileFactory.create(user=self.user1, **profile_kwargs)
        self.client.force_login(self.user1)
    
    @patch('profiles.serializers.ProfileSerializer', autospec=True)
    def test_not_filled_out_serialization(self, mock_serializer):
        self.set_profile_and_login(self, filled_out=False)
        self.client.patch(self.url1, content_type="application/json", data=json.dumps(patch_data))
        assert mock_serializer.called
```
then add tests for the other conditions. the monkey-patching adds confusion, and the returned_serializer assertion in this test case sort of obscures its intention and makes it less readable

noisecapella: Sure, I'll investigate

profiles/serializers_test.py
noisecapella: Yeah I'm not sure either. I'll investigate

noisecapella:
Comments should be addressed

noisecapella:
field_of_study bug should be fixed

gsidebo:
is MockedESTestCase not working?

profiles/views_test.py
```diff
@@ -292,3 +261,72 @@ def test_staff_sees_entire_profile(self):
...
+
+            if isinstance(field, (ListSerializer, SerializerMethodField, ReadOnlyField)):
+                # these fields are readonly
+                continue
+            elif isinstance(field, DateField):
+                assert getattr(old_profile, key) == parse(value).date()
+            else:
+                assert getattr(old_profile, key) == value
+
+    def test_validate_filled_out(self):
```
gsidebo: 1. doesn't seem like an accurate name for this test case. maybe `test_filled_out_serializer`?
1. can you move the similar `profiles.views.ProfileSerializer` block of testing code above into its own test case?

profiles/serializers_test.py
```diff
@@ -303,3 +311,117 @@ def test_delete_employment(self):
...
+            if isinstance(field, (ListSerializer, SerializerMethodField, ReadOnlyField)):
+                continue
+
+            clone = deepcopy(self.data)
+            parent_getter(clone)[key] = None
+            with self.assertRaises(ValidationError) as ex:
+                ProfileFilledOutSerializer(data=clone).is_valid(raise_exception=True)
+            assert parent_getter(ex.exception.args[0]) == {key: ['This field may not be null.']}
+
+            field = field_parent_getter(ProfileFilledOutSerializer(data=clone).fields)[key]
```
gsidebo: 1. can you move this function under `setUp` rather than mixed in among the test cases?
1. i don't think this line is necessary. the definition on line 372 seems to be taking care of this. i commented this line out and the tests succeeded

noisecapella:
@gsidebo Yes there was a problem using MockedESTestCase, it seems out of the scope of this PR to fix it

noisecapella:
Comments addressed

gsidebo:
looks great! 👍

Issue #938, opened 2016-08-29
Dashboard: Display course prices


Ferdi:
![image](https://cloud.githubusercontent.com/assets/430126/18055660/bf1754d0-6dd6-11e6-9e2b-a28d1f3b66ad.png)

# Acceptance Criteria

- [x] Update the API to return course price
- [x] Display the price for each course in the dashboard

This does not include the (?) tooltip. 

# Related Issues:

- Financial Aid

Issue #925, opened 2016-08-26
Dashboard: update to match design


pdpinch:
Should look like this:

![mm_dashboard 2x](https://cloud.githubusercontent.com/assets/430126/18012133/2d7d4694-6b86-11e6-93d2-c4add1304746.jpg)

Not like this:

![image](https://box.everhelper.me/attachment/559111/eb68aeb9-7366-4cd9-9ed9-edae1a1b5a5a/623857-O2nYtnYzIAku0OV3/screen.png)

Here is a complete set of course states:

![course_statuses 2x](https://cloud.githubusercontent.com/assets/430126/18178388/1caecd22-704c-11e6-9328-1c7fab19b421.png)

# Acceptance Criteria:

- [x] Match design
- [x] Remove histogram widget
- [x] Remove tab navigation
- [x] Fix Firefox CSS issue (#780) 
- [ ] each course state implemented

# Related Issues:

- Learners near me widget (#600)
- Context Switch (#796)
- Financial Aid links, if program has financial aid option #937
- Display course prices (requires API update) #938
- Dashboard: Display multiple runs of the same course #939
- Dashboard: Update progress widget to work with program context #941

pdpinch:
I updated the description with a screenshot of all course states. It seems likely this is more than 2 points now.

pdpinch:
Amir is going to start to work on the following issues:

- [ ] Remove histogram widget
- [ ] Remove tab navigation
- [ ] Fix Firefox CSS issue (#780)

amir-qayyum-khan:
sure

noisecapella:
I made issues for the remaining work to be done and assigned the stories to myself. See #1015 for the issues

PR #960, opened 2016-09-01
Added course price in listing

amir-qayyum-khan:
Fixes https://github.com/mitodl/micromasters/issues/938

#### What's this PR do?
- Adds course price on dashboard.

- see ```CourseRow.js``` and ```dashboard/api.py```

- Open MM dashboard,  add price in ecom table and see price with course on which you are not enroll

@pdpinch @giocalitri @aliceriot 
Ignore upgrade button.
screen shot 2016-09-01 at 5 58 58 pm

screen shot 2016-09-02 at 11 11 51 am


static/js/components/dashboard/CourseRow.js
```diff
@@ -14,6 +18,30 @@ export default class CourseRow extends React.Component {
...
+  coursePriceOrGrade: Function = (course: Course, now: moment$Moment): React$Element<*> => {
+    let firstRun = {};
+    let price;
+    if (course.runs.length > 0) {
+      firstRun = course.runs[0];
+      if (course.status === STATUS_OFFERED_NOT_ENROLLED) {
+        if (firstRun.price) {
+          price = <span className="course-price-percent">{firstRun.price}$</span>;
+        } else {
+          price = <span className="course-price-percent">Free</span>;
```
amir-qayyum-khan: @pdpinch what should i write if a course has no price? Is this OK.

pdpinch: I would leave it blank. @roberthouse54 ?

pdpinch:
If you look at the design, you'll see that the $ sign goes before the digits (American-style). 

Also, prices are typically going to be whole numbers. I don't think we want to display a decimal mark "." in that case.

static/js/components/dashboard/CourseRow.js
```diff
@@ -14,6 +18,30 @@ export default class CourseRow extends React.Component {
     now: moment$Moment,
   };
 
+  coursePriceOrGrade: Function = (course: Course, now: moment$Moment): React$Element<*> => {
+    let firstRun = {};
+    let price;
+    if (course.runs.length > 0) {
+      firstRun = course.runs[0];
+      if (course.status === STATUS_OFFERED_NOT_ENROLLED) {
```
pdpinch: We should also display the price for status 'enrolled_not_verified'

static/js/components/dashboard/CourseRow.js
```diff
@@ -14,6 +18,30 @@ export default class CourseRow extends React.Component {
     now: moment$Moment,
   };
 
+  coursePriceOrGrade: Function = (course: Course, now: moment$Moment): React$Element<*> => {
```
pdpinch: I'm confused why this is "price or grade"

static/scss/dashboard.scss
```diff
@@ -165,6 +165,20 @@ c.validation-wrapper {
         }
       }
 
+      .course-price {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+
+        .course-price-percent {
```
pdpinch: This doesn't seem like the right class name to me -- there are no percents in prices.

pdpinch:
Can you talk to @noisecapella about how to implement the (?) button? He's doing something similar for another issue.

dashboard/api.py
```diff
@@ -14,6 +14,7 @@
 from django.db.models import Q
 import pytz
 
+from ecommerce.models import CoursePrice
```
noisecapella: Yes, his logic looks correct here

amir-qayyum-khan:
@pdpinch done with fixes

static/js/components/dashboard/CourseRow.js
```diff
@@ -14,6 +19,33 @@ export default class CourseRow extends React.Component {
...
+        return (
+          <div className="course-price">
+            {price}
+            {enrollmentStatus}
+          </div>
+        );
+      }
+    }
+    return (
+      <CourseGrade course={course} now={now} />
```
pdpinch: Grades aren't going to go in this column anymore. It's probably simplest to just omit this logic for now and it will get addressed in #925

static/js/components/dashboard/CourseRow.js
```diff
@@ -14,6 +19,33 @@ export default class CourseRow extends React.Component {
...
 
+  courseMiddleSection: Function = (course: Course, now: moment$Moment): React$Element<*> => {
+    let firstRun = {};
+    let price;
+    let enrollmentStatus;
+    if (course.runs.length > 0) {
+      firstRun = course.runs[0];
+      if (course.status === STATUS_OFFERED_NOT_ENROLLED || course.status === STATUS_ENROLLED_NOT_VERIFIED) {
+        if (firstRun.price) {
+          price = <span className="course-price">${firstRun.price}</span>;
```
noisecapella: Can you create a function `formatPrice` to format the price and put it in `static/js/util/util.js`?

static/js/components/dashboard/CourseRow.js
```diff
@@ -14,6 +19,33 @@ export default class CourseRow extends React.Component {
     now: moment$Moment,
   };
 
+  courseMiddleSection: Function = (course: Course, now: moment$Moment): React$Element<*> => {
```
noisecapella: Please move this logic into `CourseGrade`, and also rename `CourseGrade` and its file and the test file to `CoursePrice`

static/scss/dashboard.scss
```diff
@@ -165,6 +165,20 @@ c.validation-wrapper {
         }
       }
 
+      .course-price {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+
+        .course-price {
```
noisecapella: We shouldn't have the inner and outer elements have the same CSS class name

static/scss/dashboard.scss
```diff
@@ -165,6 +165,20 @@ c.validation-wrapper {
         }
       }
 
+      .course-price {
```
noisecapella: Can you delete the `.course-grade` block above since you are replacing it?

noisecapella:
There should be tests added to cover both the added Python code and refactored JS

amir-qayyum-khan:
Done @nicolaeherrera

static/js/components/dashboard/CoursePrice.js
```diff
@@ -0,0 +1,39 @@
...
+    let enrollmentStatus = null;
+
+    if (course.runs.length > 0) {
+      firstRun = course.runs[0];
+      if (course.status === STATUS_OFFERED_NOT_ENROLLED || course.status === STATUS_ENROLLED_NOT_VERIFIED) {
+        if (firstRun.price) {
+          price = <span className="course-price-display">${firstRun.price}</span>;
+        }
+
+        if (course.status === STATUS_OFFERED_NOT_ENROLLED) {
```
amir-qayyum-khan: @pdpinch is this logic correct. Only show ```Enrollment open``` in case of status ```STATUS_OFFERED_NOT_ENROLLED```

pdpinch:
Did you mean to tag @noisecapella ?

amir-qayyum-khan:
@pdpinch i tag you in specific question, should i add ```Enrollment open``` for status ```STATUS_ENROLLED_NOT_VERIFIED``` where button title is ```upgrade``` option

cc @nicolaeherrera

pdpinch:
Yes. See #925 and https://cloud.githubusercontent.com/assets/430126/18178388/1caecd22-704c-11e6-9328-1c7fab19b421.png

amir-qayyum-khan:
got it thanks

dashboard/api_test.py
```diff
@@ -145,6 +146,81 @@ def test_format_run_no_run(self):
...
+            {
+                'title': crun.title,
+                'status': api.CourseStatus.UPGRADE,
+                'id': crun.pk,
+                'course_id': crun.edx_course_key,
+                'position': 1,
+                'price': 50.0
+            }
+        )
+        self.assertNotEqual(
```
noisecapella: `assertNotEqual` is not very useful here since any number of things can cause two dictionaries to be different. Can you change the below four assertions to assert that `price` does not exist in the returned dictionary?

dashboard/api_test.py
```diff
@@ -145,6 +146,81 @@ def test_format_run_no_run(self):
             api.format_courserun_for_dashboard(None, api.CourseStatus.PASSED)
         )
 
+    def test_price_in_course_run(self):
+        """Assert that price appears in course run for offered and upgrade."""
+        crun = self.create_run()
+        CoursePriceFactory.create(course_run=crun, is_valid=True, price=50)
```
noisecapella: Can you create a variable for `price` and use it in your assertions below?

static/js/components/dashboard/CoursePrice.js
```diff
@@ -0,0 +1,37 @@
...
+  render() {
+    const { course } = this.props;
+    let firstRun = {};
+    let price = null;
+    let enrollmentStatus = null;
+
+    if (course.runs.length > 0) {
+      firstRun = course.runs[0];
+      if (course.status === STATUS_OFFERED_NOT_ENROLLED || course.status === STATUS_ENROLLED_NOT_VERIFIED) {
+        if (firstRun.price) {
```
noisecapella: Can you change this to `!_.isNil(firstRun.price)` so we handle `$0` correctly?

static/js/components/dashboard/CoursePrice.js
```diff
@@ -0,0 +1,37 @@
...
+          price = <span className="course-price-display">{formatPrice(firstRun.price)}</span>;
+        }
+        enrollmentStatus = <span className="course-enrollment-description">Enrollment open</span>;
```
noisecapella: Enrollment is not always open, it depends on the `enrollment_start_date`. We want to display it only if the date is equal or later than the current day (right @pdpinch?). See `CourseDescription` for an example of handling dates. Also, if the user is already enrolled it doesn't make sense to show `Enrollment open`.

pdpinch: This is out of scope for this PR, but the goal is to support all these cases in depicted in https://cloud.githubusercontent.com/assets/430126/18178388/1caecd22-704c-11e6-9328-1c7fab19b421.png

George is correct that the enrollmentStatus string depends on the date. It should probably be handled separately from the price display logic.

amir-qayyum-khan: this message will only appear for course statuses ```STATUS_OFFERED_NOT_ENROLLED```,  ```STATUS_ENROLLED_NOT_VERIFIED```

static/js/components/dashboard/CoursePrice.js
```diff
@@ -0,0 +1,38 @@
...
+    let price = null;
+    let enrollmentStatus = null;
+
+    if (course.runs.length > 0) {
+      firstRun = course.runs[0];
+      if (course.status === STATUS_OFFERED_NOT_ENROLLED || course.status === STATUS_ENROLLED_NOT_VERIFIED) {
+        if (!_.isNil(firstRun.price)) {
+          price = <span className="course-price-display">{formatPrice(firstRun.price)}</span>;
+        }
+        enrollmentStatus = <span className="course-enrollment-description">Enrollment open</span>;
```
noisecapella: It looks like there's more to #925 than just showing this text. Can you remove it altogether and it will be added as part of #925?

amir-qayyum-khan: @noisecapella this PR address https://github.com/mitodl/micromasters/issues/938

amir-qayyum-khan: you mean remove 'open enrollment'  text only?

noisecapella: Yes

amir-qayyum-khan: Done @noisecapella

static/js/components/dashboard/CoursePrice_test.js
```diff
@@ -0,0 +1,57 @@
...
+
+  it('shows price of course with status enrolled-not-verified', () => {
+    let course = findCourse(course => course.status === STATUS_ENROLLED_NOT_VERIFIED);
+    assert.equal(course.runs[0].price, 50.00);
+
+    const wrapper = shallow(<CoursePrice course={course}/>);
+    assert.equal(wrapper.find(".course-price-display").text(), "$50");
+  });
+
+  it('shows price of course with status passed', () => {
```
noisecapella: Can you change this to use a for loop here? For example:

    for (let status of [STATUS_PASSED, STATUS_NOT_OFFERED, STATUS_VERIFIED_NOT_COMPLETED]) {
      it(`doesn't show the price of course with status ${status}`, () => {
        let course = findCourse(course => course.status === status);
        assert.isNotOk(course.runs[0].price);

        const wrapper = shallow(<CoursePrice course={course} />);
        assert.isNotOk(wrapper.find(".course-price-display"));
      });
    }

static/js/components/dashboard/CoursePrice_test.js
```diff
@@ -0,0 +1,57 @@
...
+    let course = findCourse(course => course.status === STATUS_PASSED);
+    assert.isNotOk(course.runs[0].price);
+
+    const wrapper = shallow(<CoursePrice course={course}/>);
+    assert.isNaN(wrapper.find(".course-price-display"));
```
noisecapella: We shouldn't use `isNaN` unless we are actually looking at a numerical result. Please use `isNotOk` instead

amir-qayyum-khan: @noisecapella I this [``is```](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/is.md) is valid method here, let me add it

amir-qayyum-khan:
Done @noisecapella ,
```js
 assert.isNotOk(wrapper.find(".course-price-display"));
```
was giving exception:
```AssertionError: expected { Object (root, unrendered, ...) } to be falsy```
So  I used this line 
```js
assert.equal(wrapper.find(".course-price-display").length, 0);
```

noisecapella:
Looks good to me :+1:

Issue #796, opened 2016-08-04
Learner Dashboard: allow learner to switch between MM programs (program context)


pdpinch:
As a learner, I need to be able to see my progress in all micromasters that I'm a member of. 

We will add a component to the top nav of learner dashboard with a drop-down list of all the MM programs the current user is registered for. 

![contexts 2x](https://cloud.githubusercontent.com/assets/430126/18054853/98db074c-6dd3-11e6-8a86-dbb188966614.jpg)

# Acceptance Criteria / Tasks

- [x] 1. Update the dashboard API to return data only for the current program. 
- [x] 2. API for updating the user's programs (#953)
- [x] 3. All live programs appear in the drop-down list
- [ ] 4. modal to allow user to enroll in additional programs
- [x] 5. Remove the signals handlers that update the program enrollment based on the CachedEnrollment model: the Program Enrollment will be updated only via REST API.

# Related issues

- #917 Sign Up / Get Started Modal 
- #833 Refactor dashboard API

pdpinch:
@giocalitri Does this cover the dashboard context, or is there more?

giocalitri:
We need to update the dashboard rest API to take in input the id of the program and return data about only that one.

This means that we need to somehow pass the list of the programs to the UI. This is important also to build the dropdown menu itself. Probably it makes sense to build a new REST API specific to this purpose.

If there are other parts that need to be modified, I will add another comment

pdpinch:
Ok. I updated the description.

pdpinch:
Updated screenshots

giocalitri:
If this story includes also the enrollment in the program (it seems so from the latest screenshots), then, in addition to what I wrote before, we need also to:

1. create a rest API to enroll the user in a program (the program enrollment model already exists).
2. create something (a view?) where we can send via parameter and preselect the program where to enroll and just enroll the user (use case is a user that clicks "enroll in this program" from the program marketing page: we do not want to ask her again).
3. Remove the signals handlers that update the program enrollment based on the `CachedEnrollment` model: the Program Enrollment will be updated only via REST API.

giocalitri:
I think I need clarification about 

>  If the learner is in only one MicroMasters, we will omit the component:

from the description

pdpinch:
@roberthouse54 that was an older UX that we've scrapped. The drop-down will always be present.

pdpinch:
[...]

PR #964, opened 2016-09-01
Removed signals creating or deleting a ProgramEnrollment

noisecapella:
Part 5 of #796

#### What's this PR do?
`CachedEnrollment` and `CachedCertificate` no longer automatically create `ProgramEnrollment`s when they are created. To create a `ProgramEnrollment` the user must now use the REST API

Delete all `ProgramEnrollment` using the Django shell or admin. Go to the dashboard. Wait five minutes, refresh the dashboard. Verify that there are still no `ProgramEnrollment`s in the database. Then add back the `ProgramEnrollment`s you created in the first place using the shell.

### DIFF
--- a/dashboard/signals.py
+++ b/dashboard/signals.py
@@ -1,11 +1,22 @@
 """
 Signals for user profiles
 """
-from django.db.models.signals import pre_save, post_save, post_delete
+from django.db.models.signals import post_save, post_delete
 from django.dispatch import receiver
 
 from dashboard.models import CachedEnrollment, CachedCertificate, ProgramEnrollment
-from search.tasks import index_program_enrolled_users, index_users, remove_program_enrolled_user
+from search.tasks import index_program_enrolled_users, remove_program_enrolled_user
+
+
+def _index_enrollment(user, program):
+    """
+    Helper function to index a ProgramEnrollment if one exists for a user and program
+    """
+    try:
+        program_enrollment = ProgramEnrollment.objects.get(user=user, program=program)
+        index_program_enrolled_users.delay([program_enrollment])
+    except ProgramEnrollment.DoesNotExist:
+        pass
 
 
 @receiver(post_save, sender=ProgramEnrollment, dispatch_uid="programenrollment_post_save")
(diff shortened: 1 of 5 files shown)


dashboard/signals.py
```diff
@@ -101,6 +75,8 @@ def handle_delete_certificate(sender, instance, **kwargs):  # pylint: disable=un
     """
     user = instance.user
     program = instance.course_run.course.program
-    program_enrollment = ProgramEnrollment.objects.filter(user=user, program=program).first()
-    if program_enrollment is not None:
+    try:
+        program_enrollment = ProgramEnrollment.objects.get(user=user, program=program)
         index_program_enrolled_users.delay([program_enrollment])
+    except ProgramEnrollment.DoesNotExist:
+        pass
```
gsidebo: this try/except block appears 4 times. can you create a helper method (`index_enrollment_if_exists` et al) that takes a user and program and runs these lines?

noisecapella: I'm not sure that would help with clarity here. The try except is a standard way of testing if objects exist in Django. The helper function would not be adding much and it wouldn't be as obvious to people who were familiar with Django but not our code base

gsidebo: i'm not suggesting we change the try/except. i'm saying we move the whole block of code to a new helper method since it's repeated 4 times in this file 

```python
def fetch_and_index_enrollment(user, program):
    try:
        program_enrollment = ProgramEnrollment.objects.get(
            user=user,
            program=program
        )
        index_program_enrolled_users.delay([program_enrollment])
    except ProgramEnrollment.DoesNotExist:
        pass
```

on top of that, all 4 of them get `user` and `program` from `instance` in the exact same way. perhaps the method above could be added to search.tasks, and another helper method could be added here to extract `user` and `program` from `instance` and invoke it

noisecapella: I see, I'll make the change

dashboard/signals_test.py
```diff
@@ -2,91 +2,205 @@
...
+
+    def test_delete_unlinked(self):
+        """
+        The database is not reindexed when a CachedCertificate is deleted when not linked to a ProgramEnrollment
+        """
+        with mute_signals(post_save):
+            certificate = CachedCertificateFactory.create(user=self.user, course_run=self.separate_course_run)
+        with patch('search.tasks._index_program_enrolled_users', autospec=True) as mocked:
+            certificate.delete()
+        assert not mocked.called
```
gsidebo: this class is another 65 lines of code that we've copied/pasted because we don't parameterize our tests. i won't hold up this PR for it, but it's worth pointing out

gsidebo:
code is good, testing functionality now

gsidebo:
the `gen_realistic_search_data` won't create program enrollments anymore. don't think that's in scope for this PR. i'll create an issue

gsidebo:
looks good 👍

PR #972, opened 2016-09-02
Made edx_level_of_education read only

### LINKED ISSUE #971: Profile validation shouldn't check edx_level_of_education
This field may be null when passed from edX. I'm going to fix this by making it readonly

noisecapella:
Fixes #971

#### What's this PR do?
Makes edx_level_of_education readonly to prevent users from overwriting its value, and to fix a `filled_out` validation error when it was blank

In the Django shell set your `Profile`'s `edx_level_of_education` to `None`. Then try to edit your profile. No validation errors should be seen and you should get a 200 back from the server.

### DIFF
--- a/profiles/factories.py
+++ b/profiles/factories.py
@@ -64,7 +64,7 @@ class ProfileFactory(DjangoModelFactory):
     edx_requires_parental_consent = FuzzyAttribute(FAKE.boolean)
     date_of_birth = FuzzyDate(date(1850, 1, 1))
     edx_level_of_education = FuzzyChoice(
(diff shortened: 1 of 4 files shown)


aliceriot:
LGTM and functionality seems to work, but it seems like it also works on master?

aliceriot:
Meaning, I ran through the testing routine you described but it looks like I also get a 200 on master

noisecapella:
Oh, you need to have `filled_out` set to true first before the validation error will happen, forgot to put that in the instructions

aliceriot:
oh ok I'll try that

aliceriot:
turns out it was a sillier error, I forgot to call `.save()` in the console >_<

anyway :+1:

PR #974, opened 2016-09-02
Fixed key name for search request API param

### LINKED ISSUE #973: Mailing a filtered set of users actually mails all users in a program
Steps to Reproduce:

1. Run MicroMasters with a set of fake users and set `MAILGUN_RECIPIENT_OVERRIDE` to a convenient e-mail address (i.e. not mit.edu)
2. Log in to MicroMasters as an instructor, navigate to the learner search and apply a filter. 
3. Click on Email Selected. Note the number of recipients displayed in the modal. 
4. Send a message and look at the results. 

Expected Result:

In the e-mail that is sent, there should be a list of e-mail address (below `[overridden recipients]`) that has the same number of e-mails as reported in the modal in step 3 above. 

Actual Result:

The list of e-mails listed below `[overridden recipients]` contains e-mails for all the users in the program.

gsidebo:

Fixes #973 

#### What's this PR do?

Fixes a bug where the search request API param was named incorrectly when passed to the mail API endpoint.

Go to the search page, apply a filter or two, and send an email with MAILGUN_RECIPIENT_OVERRIDE set in your .env. You should see the correct intended recipients in the email body

### DIFF
--- a/static/js/util/api.js
+++ b/static/js/util/api.js
@@ -138,7 +138,7 @@ export function sendSearchResultMail(subject: string, body: string, searchReques
     body: JSON.stringify({
       email_subject: subject,
       email_body: body,
-      searchRequest: searchRequest
+      search_request: searchRequest
     })
   });
 }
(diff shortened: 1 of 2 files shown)


noisecapella:
Works for me :+1:

PR #978, opened 2016-09-06
Fixed course tests

### LINKED ISSUE #977: Some Python tests aren't getting run
It looks like at least `courses/views_test.py` aren't getting run except for the linting tests. When you run the tests for that specific file the tests run and result in test failures

noisecapella:
Fixes #977

#### What's this PR do?
Renames `courses/views_tests.py` to `courses/views_test.py` and fixes the test failures in the test

Look in the Travis output and make sure that there are tests being executed under `courses/views_test.py` and that they all pass.

I think we should get rid of the courses and programs API altogether since we don't use them: #237

### DIFF
--- a/courses/views_test.py
+++ b/courses/views_test.py
@@ -5,7 +5,7 @@ from django.core.urlresolvers import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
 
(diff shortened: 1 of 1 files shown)


gsidebo:
looks good 👍

Issue #969, opened 2016-09-02
Add UI for adding and choosing program enrollments


noisecapella:
Part of #796

PR #979, opened 2016-09-06
Filtered out programs which are not live from program enrollments API

noisecapella:
Part of #969

#### What's this PR do?
Filters out program enrollments for programs that aren't live. We filter this in the REST API when the user POSTs to create the program enrollment but there's still the possibility a program may stop being live at some point in the future, so we should filter on GET too.

Go to `/api/v0/enrolledprograms/`. Make sure there is at least one enrollment there. In the Django shell set `live=False` for that program. Go to `/api/v0/enrolledprograms/` and see that this program enrollment no longer appears in the API.

### DIFF
--- a/courses/views.py
+++ b/courses/views.py
@@ -58,7 +58,7 @@ class ProgramEnrollmentListView(ListCreateAPIView):
         """
         Filter programs by the user enrollment
         """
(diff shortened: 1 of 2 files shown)


gsidebo:
looks good! 👍

PR #992, opened 2016-09-07
Added home page signup popup

### LINKED ISSUE #917: Home: Sign Up / Get started modal
Implement Sign Up workflow:

popup

Acceptance criteria:

- [x] Add pop-up modal form to home page, including drop-down for user to choose
Program. On home page, user chooses program. 
- [x] Add pop-up modal to program page; determine learner's program from the page. On Program page, we can set program to the current page's program. 
- [x] Open TOS in a new window
- [x] Associate user with the chosen program.
- [x] Set "agreed to TOS" field to true when the user returns from logging in to edX
- [x] Clicking outside the modal dismisses it

TOS Design:

tos 2x

aliceriot:

closes #917 

#### What's this PR do?

This adds a little dialog to the home page and to the signup pages which lets you pick a program. We persist this to `localStorage`. In the dashboard, when the `App` container updates it checks for this value in `localStorage` and, if present, it uses it to issue an `addProgramEnrollment` api request.

right now:

- go to the homepage
- run `window.openDialog()` in the JS console
- pick a program your user is not currently enrolled in
- go to the dashboard
- confirm that, after the dashboard loads, `localStorage.programId` is empty, and that your user has a new program enrollment.


static/js/signup_dialog.js
```diff
@@ -0,0 +1,40 @@
...
+
+import injectTapEventPlugin from 'react-tap-event-plugin';
+injectTapEventPlugin();
+
+const store = configureStore();
+
+const dialogDiv = document.querySelector('#signup-dialog');
+
+const openDialog = () => store.dispatch(setSignupDialogVisibility(true));
+window.openDialog = openDialog;
```
noisecapella: Is this just for debugging?

static/js/util/test_utils.js
```diff
@@ -1,7 +1,36 @@
...
+  const setItem = sinon.spy((key, value) => {
+    storage[key] = value || "";
+  });
+
+  const removeItem = sinon.spy(key => {
+    delete storage[key];
+  });
+
+  const reset = () => {
+    [getItem, setItem, removeItem].forEach(func => {
```
aliceriot: Can you give an example of how I could make that work? I tried playing around with a sandbox and I couldn't get it to play nicely with the spy functions.

static/js/signup_dialog.js
```diff
@@ -0,0 +1,40 @@
...
+
+// find the DOM element and attach openDialog to onClick
+const logoLink = document.querySelector('a.navbar-brand');
+
+logoLink.onClick = openDialog;
+
+if ( typeof SETTINGS.program === 'number' ) {
+  store.dispatch(setSignupProgram(SETTINGS.program));
```
noisecapella: Since the reducers have access to `SETTINGS` I feel like it would be cleaner to set the initial value there instead.

static/js/signup_dialog.js
aliceriot: yes

cms/models.py
```diff
@@ -17,6 +17,15 @@
 from profiles.api import get_social_username
 from ui.views import get_bundle_url
 
+def program_info(programs):
```
noisecapella: Can we use the existing `ProgramSerializer` for this? It provides the same two fields. You should be able to use it like `ProgramSerializer().to_representation(program)`

backends/pipeline_api.py
```diff
@@ -72,6 +72,7 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
         pass
     user_profile.gender = user_profile_edx.get('gender')
     user_profile.edx_mailing_address = user_profile_edx.get('mailing_address')
+    user_profile.agreed_to_terms_of_service = True
```
noisecapella: Currently the server side validation will allow you to switch this back to `False` as long as `filled_out` is not `True`. Since it doesn't seem like the user has a reason to set this anymore via PATCH, can we add it to the list of read only fields in `ProfileSerializer` and `ProfileLimitedSerializer`?

static/js/containers/App.js
```diff
@@ -60,6 +64,7 @@ class App extends React.Component {
     this.fetchEnrollments();
     this.requireProfileFilledOut();
     this.requireCompleteProfile();
+    this.updateProgramEnrollments();
```
noisecapella: This should go in `componentDidMount` too

static/js/util/test_utils.js
noisecapella: At the top of `localStorageMock` you would put `const sandbox = sinon.sandbox.create()`. Then for example `getItem` would become `const getItem = sandbox.spy(...)`. I've never actually used it with spies, just stubs, but I feel like that should work.

noisecapella:
It looks like localStorage is used entirely within React classes. What do you think about using redux-localstorage instead and putting the new program id in its own reducer?

static/js/containers/App.js
```diff
@@ -91,6 +96,30 @@ class App extends React.Component {
...
+    ) {
+      if ( !enrollments.programEnrollments.find(e => e.id === programId) ) {
+        dispatch(addProgramEnrollment(programId)).then(() => {
+          window.localStorage.removeItem("programId");
+        }, e => {
+          window.localStorage.removeItem("programId");
+          if ( e.errorStatusCode === 404 ) {
+            window.localStorage.removeItem("programId");
+          } else {
+            window.localStorage.setItem("programIdFailed", programId);
```
noisecapella: I'm not sure if it makes sense to store this if we're not going to use it again. If we want to alert the user to a failure we should do a `console.error` instead or something more visible (cc @roberthouse54)

static/js/util/test_utils.js
aliceriot: I tried the following in the nodejs repl:

```js
var sinon = require('sinon')

const sandbox = sinon.sandbox;

const spyFunc = x => x + 2

spyFunc(2) // 4
let spy = sandbox.spy(spyFunc)

spy.called // false

spy(2) // 4

spy.called // true

sandbox.restore()

spy.called // still true

sandbox.reset()

spy.called // still true
```

so it seems like the `restore` and `reset` methods on the sandbox don't clear and reset the spys and mocks created by the sandbox, it looks like it just clears the sandbox out so that new mocks may be added (`sandbox.fakes === []`).

aliceriot:
I was thinking about that, I think it would make sense to pull out the code we need into a separate reducer both for simplicity and also to keep the bundle size down.

static/js/util/test_utils.js
noisecapella: Oh that's interesting. I guess there's no way around this

aliceriot: yeah I was kinda bummed, seemed like not exactly the way the sandbox should work...

backends/pipeline_api.py
aliceriot: yes that's a good idea

static/js/util/test_utils.js
aliceriot: nvm! I forgot to called `sinon.sandbox.create()` above >_<

it actually does work, haha.

cms/models.py
aliceriot: yeah that makes sense to me.

static/js/containers/App.js
aliceriot: Yeah I agree, I think we need more clarity on what sort of error handling behavior to use.

aliceriot:
@noisecapella I just pushed the changes to use `redux-localstorage` on the home page. I also moved the signup-dialog reducers and stuff into separate files (so we don't need to include the rest of the redux code in the bundle) and refactored `configureStore.js` to support the changes.

noisecapella:
Ok, let me know when it's ready for review

aliceriot:
will do!

aliceriot:
@noisecapella do you have any thoughts about testing the `SignupDialog` component? It's a container component (connected to a store) but it's not rendered inside of react-router. I thought about just making a store and creating some test infrastructure using `<Provider>` for it.

noisecapella:
Makes sense to me

aliceriot:
word, I'll give that a go and ping you when it's working.

aliceriot:
@noisecapella this is ready for review again

noisecapella:
This doesn't currently pick the newly added program as the current program, though that's coming as part of #1000 so no need to do anything here

static/js/actions/signup_dialog.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+// actions for the signup dialog on the home and program pages
+import { createAction } from 'redux-actions';
+
+export const SET_DIALOG_VISIBILITY = 'SET_SIGNUP_DIALOG_VISIBILITY';
```
noisecapella: Can you change the text for this and `SET_PROGRAM` to match the variable name?

static/js/containers/App.js
```diff
@@ -102,6 +107,32 @@ class App extends React.Component {
...
+      let programId = filterPositiveInt(R.prop('program', JSON.parse(programData)));
+      if ( programId && !enrollments.programEnrollments.find(e => e.id === programId) ) {
+        dispatch(addProgramEnrollment(programId)).then(
+          () => {
+            window.localStorage.removeItem("signupDialogRedux");
+          },
+          e => {
+            window.localStorage.removeItem("signupDialogRedux");
+            if ( e.errorStatusCode !== 404 ) {
+              window.localStorage.setItem("programIdFailed", programId);
```
noisecapella: Can you change this to a `console.error` unless we have a use case for leaving it in localStorage?

static/js/containers/SignupDialog.js
```diff
@@ -0,0 +1,104 @@
...
+        Which MicroMasters program are you signing up for?
+        { selectProgram(setProgram, program) }
+      </div>
+      <a className="mm-button signup" href="/login/edxorg">
+        Continue with edX
+      </a>
+      <div className="terms-of-service-text">
+        By clicking "Continue with edX" I certify that I agree with
+        <a href="/terms_of_service" target="_blank">
+          {" MIT MicroMasters Terms of Service. "}
```
noisecapella: Just nit picking here but the leading space will show up in the link here and below

static/js/containers/SignupDialog.js
```diff
@@ -0,0 +1,104 @@
...
+  </Dialog>;
+};
+
+const mapStateToProps = state => ({
+  open: state.dialogVisibility,
+  program: state.program,
+});
+
+const mapDispatchToProps = dispatch => {
+  const setProgramHelper = (e,k,v) => {
```
noisecapella: Can you change `e,k,v` to `event, key, value` just for clarity?

static/js/containers/SignupDialog.js
```diff
@@ -0,0 +1,104 @@
...
+    </div>
+  </Dialog>;
+};
+
+const mapStateToProps = state => ({
+  open: state.dialogVisibility,
+  program: state.program,
+});
+
+const mapDispatchToProps = dispatch => {
```
noisecapella: This is a useful function, I keep forgetting `connect` takes a second argument

static/js/reducers/signup_dialog_test.js
```diff
@@ -0,0 +1,52 @@
...
+  SET_PROGRAM
+} from '../actions/signup_dialog';
+
+describe('signup dialog reducer', () => {
+  let sandbox, store, dispatchThen;
+
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+    store = configureTestStore(signupDialog);
+    dispatchThen = store.createDispatchThen(state => state);
```
noisecapella: I think `state => state` is the default argument so you can leave it out here

static/js/containers/App.js
aliceriot: oh yeah, forgot to change that.

static/js/store/configureStore.js
```diff
@@ -4,29 +4,52 @@ import { compose, createStore, applyMiddleware } from 'redux';
...
+
+const devTools = () => (
+  notProd() && window.devToolsExtension ? window.devToolsExtension() : f => f
+);
+
+const storage = path => (
+  compose(filter([path]))(adapter(window.localStorage))
+);
+
+const createPersistantStore = persistance => compose(
```
noisecapella: `persistant` => `persistent`, same for `persistance`

static/js/store/configureStore.js
```diff
@@ -4,29 +4,52 @@ import { compose, createStore, applyMiddleware } from 'redux';
...
+  persistance,
+  middleware(),
+  devTools()
+)(createStore);
+
+const createPersistantTestStore = persistance => compose(
```
noisecapella: Also here

static/js/containers/SignupDialog.js
aliceriot: Yeah that's intentional, so that there's a space between `with` from the line above and `MIT`.

noisecapella:
Looks good overall. I think if we can make use of the same program signup reducer in both bundles we won't need to mock localStorage or parse its JSON anymore

static/js/actions/signup_dialog.js
aliceriot: oops, that was a half-completed small refactor

static/js/store/configureStore.js
aliceriot: oops, spelling :D

aliceriot:
Cool. I like the idea of refactoring to use the same reducer in both bundles, but I want to keep the localStorage mocking (and use it in unit tests) to test the behavior of the code in `App.js` to update the API if there's a saved program id.

aliceriot:
@noisecapella I made those changes, let me know what you think!

noisecapella:
The visibility of the dialog is stored in localStorage which is a little quirky to me, but feel free to leave it

static/js/containers/SignupDialog_test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  it('should pull programs from SETTINGS.programs', () => {
+    openDialog();
+
+    let renderedSelectOptions = findReact(programSelect()).props.children.map(child => {
+      let { value, primaryText } = child.props;
+      return { id: value, title: primaryText };
+    });
+
+    assert.deepEqual(renderedSelectOptions, SETTINGS.programs);
```
noisecapella: Should we sort the programs listed in the dialog?

noisecapella:
I just merged #1000, can you rebase?

static/js/store/configureStore.js
```diff
@@ -3,30 +3,59 @@
...
+  compose(filter(paths))(adapter(window.localStorage))
+);
+
+const createPersistentStore = persistence => compose(
+  persistence,
+  middleware(),
+  devTools()
+)(createStore);
+
+const createPersistantTestStore = persistence => compose(
```
noisecapella: Should be `createPersistentTestStore`

noisecapella:
Functionality works great

static/js/store/configureStore.js
aliceriot: lol idk why I have so much trouble with that word! haha

static/js/containers/SignupDialog_test.js
aliceriot: we could, alphabetically by the title?

noisecapella: Sounds good to me, that's how I did it in the other PR

aliceriot: word, I'll add that6

aliceriot:

noisecapella:
:+1:

PR #995, opened 2016-09-08
Various css changes to colors, fonts, margins

roberthouse54:

These are general style changes not necessarily for any ticket. I made changes to styles in the dashboard, the user page, and the learner search.

Look at the changes to the css and js files and see that nothing looks broken.

This is an ongoing process of tweaking css styles that will continue.

### DIFF
--- a/static/js/components/UserChip.js
+++ b/static/js/components/UserChip.js
@@ -18,6 +18,7 @@ const UserChip = ({ profile }: {profile: Profile}): React$Element<*> => (
         { mstr(getEmployer(profile)) }
       </span>
       <a href={`/users/${profile.username}`} className="mm-minor-action">
+        <Icon name="person" />
         <span>View profile</span>
       </a>
       </div>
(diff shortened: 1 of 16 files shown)


static/js/components/UserChip.js
```diff
@@ -18,7 +18,6 @@ const UserChip = ({ profile }: {profile: Profile}): React$Element<*> => (
         { mstr(getEmployer(profile)) }
       </span>
       <a href={`/users/${profile.username}`} className="mm-minor-action">
-        <Icon name="person" />
```
noisecapella: Can you remove line 4 in this file since we no longer need `Icon` after your change?

amir-qayyum-khan:
```log
/src/static/js/components/UserChip.js
  4:8  error  'Icon' is defined but never used  no-unused-vars\
```

amir-qayyum-khan:
👍 💯 🎉

noisecapella:
Why was the `Icon` change reverted?

roberthouse54:
@noisecapella I made a mistake deleting it and asked Amir to revert the change

PR #993, opened 2016-09-07
Refactored dashboard page

amir-qayyum-khan:
#### What are the relevant tickets?
- Fix some issues from https://github.com/mitodl/micromasters/issues/925 
- Fixes https://github.com/mitodl/micromasters/issues/780

#### What's this PR do?
- [x] Remove histogram widget
- [x] Remove tab navigation
- [x] Fix Firefox CSS issue (#780) 

#### Where should the reviewer start?
- See ``App.js, Nav.js``

#### How should this be manually tested?
- see ``/dashboard/`` page

@pdpinch @noisecapella @aliceriot 

#### Screenshots (if appropriate)
screen shot 2016-09-07 at 11 18 18 pm

Issue #940, opened 2016-08-29
Dashboard: Add tooltip with personal pricing info 


Ferdi:
Depends on #937 

<img width="485" alt="screen shot 2016-09-09 at 9 42 43 am" src="https://cloud.githubusercontent.com/assets/430126/18389011/d6ad5146-7671-11e6-96d2-2161d85dbe63.png">

# Acceptance Criteria

- [ ] tool tip icon is only displayed for courses that offer personalized pricing
- [ ] Need to get text from DEDP program

pdpinch:
@roberthouse54 do we still have a need for this tooltip?

PR #997, opened 2016-09-09
Fixed course display on dashboard

amir-qayyum-khan:
- Fixes https://github.com/mitodl/micromasters/issues/925
- Fixes https://github.com/mitodl/micromasters/issues/1009

#### What's this PR do?
- [x] Match design

- see course row

- see ``/dashboard/`` page

@noisecapella @pdpinch @aliceriot 

Continuation of https://github.com/mitodl/micromasters/pull/993

screen shot 2016-09-09 at 4 21 23 pm

### DIFF
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -22,19 +22,27 @@ export default class CourseAction extends React.Component {
   makeEnrollButton = (text: string, run: CourseRun, disabled: boolean) => {
     const { checkout } = this.props;
     let onClick;
+    let button;
+
     if (!disabled) {
       onClick = () => {
         checkout(run.course_id);
       };
+      button = (
+        <Button type='button' className="dashboard-button" onClick={onClick}>
+          {text}
+        </Button>
+      );
+    } else {
+      button = (
+        <Button disabled type='button' className="dashboard-button">
+          {text}
+        </Button>
+      );
     }
+
     return <span>
-      <Button
-        className="dashboard-button"
-        disabled={disabled}
-        onClick={onClick}
-      >
-        {text}
-      </Button>
+      {button}
       <span className="sr-only"> in {run.title}</span>
     </span>;
   };
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -2,6 +2,7 @@
 // @flow
 import React from 'react';
 import moment from 'moment';
+import Button from 'react-mdl/lib/Button';
 
 import type { Course, CourseRun } from '../../flow/programTypes';
 import {
@@ -27,13 +28,13 @@ export default class CourseAction extends React.Component {
       };
     }
     return <span>
-      <button
+      <Button
         className="dashboard-button"
         disabled={disabled}
         onClick={onClick}
       >
         {text}
-      </button>
+      </Button>
       <span className="sr-only"> in {run.title}</span>
     </span>;
   };
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -21,30 +21,24 @@ export default class CourseAction extends React.Component {
 
   makeEnrollButton = (text: string, run: CourseRun, disabled: boolean) => {
     const { checkout } = this.props;
-    let onClick;
-    let button;
+    let buttonProps = {};
 
     if (!disabled) {
-      onClick = () => {
+      buttonProps.onClick = () => {
         checkout(run.course_id);
       };
-      button = (
-        <Button type='button' className="dashboard-button" onClick={onClick}>
-          {text}
-        </Button>
-      );
     } else {
-      button = (
-        <Button disabled type='button' className="dashboard-button">
-          {text}
-        </Button>
-      );
+      buttonProps.disabled = true;
     }
 
-    return <span>
-      {button}
-      <span className="sr-only"> in {run.title}</span>
-    </span>;
+    return (
+      <span>
+        <Button className="dashboard-button" {...buttonProps}>
+          {text}
+        </Button>
+        <span className="sr-only"> in {run.title}</span>
+      </span>
+    );
   };
 
   render() {
--- a/static/js/components/dashboard/CourseAction_test.js
+++ b/static/js/components/dashboard/CourseAction_test.js
@@ -63,11 +63,13 @@ describe('CourseAction', () => {
     const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
     let description = wrapper.find(".course-action-description");
-    let buttonProps = buttonContainer.find("button").props();
+    let buttonProps = buttonContainer.find(".dashboard-button").props();
+    let spanText = buttonContainer.find(".sr-only").text();
+    let buttonText = buttonContainer.find(".dashboard-button").children().text();
     let firstRun = course.runs[0];
 
     assert(!buttonProps.disabled);
-    assert.equal(buttonContainer.text(), `Upgrade in ${firstRun.title}`);
+    assert.equal(`${buttonText}${spanText}`, `Upgrade in ${firstRun.title}`);
     assert.equal(description.text(), "");
     assertCheckoutButton(buttonProps, firstRun.course_id);
   });
@@ -81,11 +83,13 @@ describe('CourseAction', () => {
     const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
     let description = wrapper.find(".course-action-description");
-    let buttonProps = buttonContainer.find("button").props();
+    let buttonProps = buttonContainer.find(".dashboard-button").props();
+    let spanText = buttonContainer.find(".sr-only").text();
+    let buttonText = buttonContainer.find(".dashboard-button").children().text();
     let firstRun = course.runs[0];
 
     assert(buttonProps.disabled);
-    assert.equal(buttonContainer.text(), `Enroll in ${firstRun.title}`);
+    assert.equal(`${buttonText}${spanText}`, `Enroll in ${firstRun.title}`);
     assert.equal(description.text(), `Enrollment begins ${firstRun.fuzzy_enrollment_start_date}`);
   });
 
@@ -100,10 +104,12 @@ describe('CourseAction', () => {
     const wrapper = shallow(<CourseAction course={course} now={yesterday} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
     let description = wrapper.find(".course-action-description");
-    let buttonProps = buttonContainer.find("button").props();
+    let buttonProps = buttonContainer.find(".dashboard-button").props();
+    let spanText = buttonContainer.find(".sr-only").text();
+    let buttonText = buttonContainer.find(".dashboard-button").children().text();
 
     assert(buttonProps.disabled);
-    assert.equal(buttonContainer.text(), `Enroll in ${firstRun.title}`);
+    assert.equal(`${buttonText}${spanText}`, `Enroll in ${firstRun.title}`);
     let formattedDate = moment(firstRun.enrollment_start_date).format(DASHBOARD_FORMAT);
     assert.equal(description.text(), `Enrollment begins ${formattedDate}`);
   });
@@ -118,10 +124,12 @@ describe('CourseAction', () => {
     const wrapper = shallow(<CourseAction course={course} now={today} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
     let description = wrapper.find(".course-action-description");
-    let buttonProps = buttonContainer.find("button").props();
+    let buttonProps = buttonContainer.find(".dashboard-button").props();
+    let spanText = buttonContainer.find(".sr-only").text();
+    let buttonText = buttonContainer.find(".dashboard-button").children().text();
 
     assert(!buttonProps.disabled);
-    assert.equal(buttonContainer.text(), `Enroll in ${firstRun.title}`);
+    assert.equal(`${buttonText}${spanText}`, `Enroll in ${firstRun.title}`);
     assert.equal(description.text(), ``);
     assertCheckoutButton(buttonProps, firstRun.course_id);
   });
@@ -136,10 +144,12 @@ describe('CourseAction', () => {
     const wrapper = shallow(<CourseAction course={course} now={tomorrow} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
     let description = wrapper.find(".course-action-description");
-    let buttonProps = buttonContainer.find("button").props();
+    let buttonProps = buttonContainer.find(".dashboard-button").props();
+    let spanText = buttonContainer.find(".sr-only").text();
+    let buttonText = buttonContainer.find(".dashboard-button").children().text();
 
     assert(!buttonProps.disabled);
-    assert.equal(buttonContainer.text(), `Enroll in ${firstRun.title}`);
+    assert.equal(`${buttonText}${spanText}`, `Enroll in ${firstRun.title}`);
     assert.equal(description.text(), ``);
     assertCheckoutButton(buttonProps, firstRun.course_id);
   });
--- a/static/js/components/dashboard/CourseAction_test.js
+++ b/static/js/components/dashboard/CourseAction_test.js
@@ -58,18 +58,26 @@ describe('CourseAction', () => {
     assert.equal(wrapper.text(), '');
   });
 
+  [STATUS_OFFERED_NOT_ENROLLED, STATUS_ENROLLED_NOT_VERIFIED].forEach((status) => {
+    it(`shows the enroll button followed by course title when status is ${status}`, () => {
+      let course = findCourse(course => course.status === status);
+      let firstRun = course.runs[0];
+      const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
+      let buttonContainer = wrapper.find(".course-action-action");
+
+      assert.include(buttonContainer.text(), `<Button /> in ${firstRun.title}`);
+    });
+  });
+
   it('shows an upgrade button if user is not verified but is enrolled', () => {
     let course = findCourse(course => course.status === STATUS_ENROLLED_NOT_VERIFIED);
     const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
-    let spanText = buttonContainer.find(".sr-only").text();
     let button = buttonContainer.find(".dashboard-button");
-    let buttonText = button.children().text();
     let description = wrapper.find(".course-action-description");
     let firstRun = course.runs[0];
 
     assert.isUndefined(button.props().disabled);
-    assert.equal(`${buttonText}${spanText}`, `Upgrade in ${firstRun.title}`);
     assert.equal(description.text(), "");
     assertCheckoutButton(button, firstRun.course_id);
   });
@@ -82,14 +90,11 @@ describe('CourseAction', () => {
     ));
     const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
-    let spanText = buttonContainer.find(".sr-only").text();
     let button = buttonContainer.find(".dashboard-button");
-    let buttonText = button.children().text();
     let description = wrapper.find(".course-action-description");
     let firstRun = course.runs[0];
 
     assert.isTrue(button.props().disabled);
-    assert.equal(`${buttonText}${spanText}`, `Enroll in ${firstRun.title}`);
     assert.equal(description.text(), `Enrollment begins ${firstRun.fuzzy_enrollment_start_date}`);
   });
 
@@ -103,13 +108,10 @@ describe('CourseAction', () => {
     let yesterday = moment(firstRun.enrollment_start_date).add(-1, 'days');
     const wrapper = shallow(<CourseAction course={course} now={yesterday} checkout={checkoutStub}/>);
     let buttonContainer = wrapper.find(".course-action-action");
-    let spanText = buttonContainer.find(".sr-only").text();
     let button = buttonContainer.find(".dashboard-button");
-    let buttonText = button.children().text();
     let description = wrapper.find(".course-action-description");
 
     assert.isTrue(button.props().disabled);
-    assert.equal(`${buttonText}${spanText}`, `Enroll in ${firstRun.title}`);
     let formattedDate = moment(firstRun.enrollment_start_date).format(DASHBOARD_FORMAT);
     assert.equal(description.text(), `Enrollment begins ${formattedDate}`);
   });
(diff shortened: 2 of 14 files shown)


gsidebo:
hey @amir-qayyum-khan. i'm going to start reviewing this. i know this is a work in progress, so let me know if there are any files/sections that i should stay away from

amir-qayyum-khan:
@gsidebo  Right now working on ``courseAction.js`` and js tests. Other then that it is good to review.

gsidebo:
@amir-qayyum-khan my data isn't good for testing this. the fastest way to get me to a good state would probably be for you to give me an export of your Course and CourseRun tables (or just a copy/paste of the output when you select * from those tables). would you mind doing that?

static/js/components/dashboard/CoursePrice.js
```diff
@@ -1,21 +1,69 @@
...
+      break;
+    }
+    return (
+      <div>
+        <span className="tooltip-link"
+          data-tip
+          data-for='why-we-ask-this'>
+          <IconButton name="help" className="help"/>
+        </span>
+        <ReactTooltip id="why-we-ask-this" effect="solid" event="click" globalEventOff="click">
```
gsidebo: please change this `id` attribute (and the `data-for` attribute at line 55) to something like 'course-detail' or something else that describes this tooltip more accurately

static/js/components/dashboard/CoursePrice.js
```diff
@@ -1,21 +1,69 @@
...
 } from '../../constants';
 import { formatPrice } from '../../util/util';
 
 export default class CoursePrice extends React.Component {
   props: {
     course: Course,
+    now: moment$Moment,
   };
 
+  help(course: Course, now: moment$Moment, firstRun: CourseRun): React$Element<*> {
```
gsidebo: something like `renderTooltip` would probably be a better name for this function

static/js/components/dashboard/CourseDescription.js
```diff
@@ -19,49 +18,69 @@ export default class CourseDescription extends React.Component {
...
     case STATUS_NOT_OFFERED:
-      if (firstRun.status === STATUS_NOT_PASSED) {
-        text = 'You failed this course';
+      if (firstRun.status === STATUS_NOT_PASSED && firstRun.end_date) {
+       let courseEndDate = moment(firstRun.end_date);
+        if (courseEndDate.isAfter(now, 'day')) {
+          text = this.courseDate('Ended', courseEndDate);
+        }
+      } else if (!_.isNil(firstRun.fuzzy_start_date)) {
+        text = `Comming ${firstRun.fuzzy_start_date}`;
```
gsidebo: typo: ~~Comming~~ -> Coming

noisecapella:
Until the dashboard API refactor is done much of the information necessary to display certain states is missing

static/js/components/dashboard/CourseDescription.js
```diff
@@ -19,49 +18,69 @@ export default class CourseDescription extends React.Component {
...
-    let text = "", enrolled = "";
-    let firstRun: CourseRun = {};
-    if (course.runs.length > 0) {
-      firstRun = course.runs[0];
+  courseDate(label: string, date: Date): string {
+    let formattedDate = date.format(DASHBOARD_FORMAT);
+    return `${label}: ${formattedDate}`;
+  }
+
+  courseDateMessage(courseStatus: string, firstRun: CourseRun): string {
```
gsidebo: i think it might be better if the logic to produce the course description text and the course tooltip text (from CoursePrice.js) was moved into CourseRow.js. both this file and CoursePrice.js have logic that is built around the value of `courseStatus`, and there is some similar logic with specific values of `courseStatus` (take `STATUS_VERIFIED_NOT_COMPLETED` for example - the logic to produce the description text and tooltip text is very similar). if you combined this code to produce the course description and tooltip text in CourseRow.js, you could pass the text as a prop to the necessary components. 

it's worth getting other opinions on this - @noisecapella @aliceriot what do you think of the idea above?

static/js/components/dashboard/CoursePrice.js
gsidebo: also, i don't think this should be part of this component (`CoursePrice`). the tooltip is not related to the course price in any way, and i don't think there is a good reason for it to be rendered inside of the same element as the price

static/js/components/dashboard/CoursePrice.js
```diff
@@ -1,21 +1,69 @@
...
+    case STATUS_VERIFIED_NOT_COMPLETED:
+      if (firstRun.course_start_date) {
+        let courseStartDate = moment(firstRun.course_start_date);
+        if (courseStartDate.isAfter(now, 'day')) {
+          let formattedDate = courseStartDate.format(DASHBOARD_FORMAT);
+          text = `Begins ${formattedDate}`;
+        }
+      }
+      break;
+    }
```
gsidebo: we either need a default value for `text`, or we need to hide the tooltip when `text` is blank. right now, this is possible (i don't know why the output is 'true', to be honest...)

ss 2016-09-09 at 10 45 14

logically it just doesn't make sense to render a tooltip component if there is no text

static/js/components/dashboard/CoursePrice.js
pdpinch: fwiw, in practice that tooltip is going to be used to explain personalized pricing -- for programs that offer personalized pricing. 

The original text about verified courses is going to be replaced.

static/js/components/dashboard/CoursePrice.js
pdpinch: I agree with @gsidebo -- this would give use an easy way for the tooltip to be conditional on whether the program offer personalized pricing (see #940)

static/js/components/dashboard/CoursePrice.js
```diff
@@ -1,21 +1,69 @@
...
   props: {
     course: Course,
+    now: moment$Moment,
   };
 
+  renderTooltip(course: Course, now: moment$Moment, firstRun: CourseRun): React$Element<*> {
+    let text = "";
+
+    switch (course.status) {
+    case STATUS_PASSED:
```
amir-qayyum-khan: @pdpinch @roberthouse54 

On clicking help icon (?) on course, I am showing this information in tooltip. Can please review it and tell me if I am wrong?
I am not clear what to show on help click.

static/js/components/dashboard/CoursePrice.js
amir-qayyum-khan: Fixed

static/js/components/dashboard/CoursePrice.js
gsidebo: @amir-qayyum-khan please look at this design: https://cloud.githubusercontent.com/assets/430126/18178388/1caecd22-704c-11e6-9328-1c7fab19b421.png

not every state puts text in a tooltip. look at the "Failed" state (near the bottom). there is no tooltip and the price isn't displayed

gsidebo:
_pasting a previous comment on an outdated diff for visibility. @amir-qayyum-khan please look into this_ 

_(comment on CoursePrice.js)_ i don't think this code to produce the tooltip should be part of this component (CoursePrice). the tooltip is not related to the course price in any way, and i don't think there is a good reason for it to be rendered inside of the same element as the price. this should be moved to CourseRow.js or its own component

EDIT: sorry, i just saw that there is a tooltip that is related to price, and that is "You need to upgrade to the Verified course to get MicroMasters credit". the other bits of text ("Complete!", "You failed this course", "Begins [date]") don't seem like they should be in a tooltip at all. i'm going to review the latest design from @roberthouse54 and come back to this. if the 3rd screenshot in #925 is accurate, then we have some major issues with this PR.

1. there is text in the tooltip when it isn't supposed to be in a tooltip
1. missing a few text outputs ("You are enrolled in this course", "You are enrolled in this course")
1. it looks like the 'Enrollment starts/begins [date]' text needs to move away from the Enroll button in CourseAction.js to be closer to this output

there are some other issues, but let's start there

amir-qayyum-khan:
@gsidebo got it, I am waiting for @roberthouse54 feedback on what to display on tooltip. I am not sure what to display. After that i will refactor price

roberthouse54:
Gavin, your comments above are correct, but to be honest, I am not clear what the goal is for this issue vs. #925. The third mockup on that other ticket does contain all the course status designs... or at least the mock on Invision does. It seems like this should be the ticket to fix all the course statuses to match the design.

The tooltip next to price should only contain information about the verified course. It should say "You need to enroll in the Verified Course to get MicroMasters credit."

amir-qayyum-khan:
- @gsidebo Fixed tooltip, now tooltip reflects price or upgrade as @roberthouse54 suggested. 
- @roberthouse54 All statues was not part of this PR. This is for style Fix. IF you want me to add statues as well then please help me to understand what status for what flag.

cc @pdpinch

gsidebo:
@amir-qayyum-khan thanks for letting me know. i'm still just a little confused, but in the meantime, it would help if you could tell me how you altered your course run/enrollment data to test the changes in this PR

static/js/components/dashboard/CoursePrice.js
```diff
@@ -29,7 +58,7 @@ export default class CoursePrice extends React.Component {
     }
 
     return <div className="course-price">
-      {price}
+      {price} {this.renderTooltip(course)}
```
gsidebo: i think you should create a new function `courseTooltipText` which just returns the text for a course (or no text at all) and change `renderTooltip` to accept a `text` parameter. this `render` function (starting at line 46) would call `courseTooltipText`, then call `renderTooltip` if the text isn't blank. it's confusing to have `renderTooltip` being called in all cases, but in many cases it's not actually rendering anything

static/js/components/dashboard/CourseAction.js
```diff
@@ -21,19 +22,27 @@ export default class CourseAction extends React.Component {
...
+
     return <span>
-      <button
-        className="dashboard-button"
-        disabled={disabled}
-        onClick={onClick}
-      >
-        {text}
-      </button>
+      {button}
```
gsidebo: i think this function can be rewritten like this:

```javascript
  makeEnrollButton = (text: string, run: CourseRun, disabled: boolean) => {
    const { checkout } = this.props;
    let buttonProps = {};

    if (!disabled) {
      buttonProps.onClick = () => {
        checkout(run.course_id);
      };
    } else {
      buttonProps.disabled = true;
    }

    return <span>
      <Button className="dashboard-button" {...buttonProps}>
        {text}
      </Button>
      <span className="sr-only"> in {run.title}</span>
    </span>;
  };
```

side note: i don't think the <Button /> component needs type='button'. not sure about that, though

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -31,80 +34,58 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
+  for (let status of statuses) {
+    it(`does show date with status ${status}`, () => {
+      let course = findCourse(course => course.status === status);
+      const wrapper = shallow(<CourseDescription course={course} />);
+      let firstRun: CourseRun = {};
 
-    // course starts today
-    let wrapper = shallow(<CourseDescription course={course} now={courseStart}/>);
-    assert.equal(wrapper.find(".course-description-result").text(), "");
+      if (course.runs.length > 0) {
```
gsidebo: in unit tests, we should not be loading some test data, then doing something with that data if it happens to match some conditions. we should be using exact data that should result in some kind of behavior. that means the data should be created (or loaded) in its exact desired state as part of the test.

example: line 68 makes an assertion that an element has the text `Ended: ${formattedDate}`. the data that is intended to result in that label should be created or loaded exactly as part of the test. here's what i'm thinking in terms of code.

```javascript
export type CourseRunDateValues = {
  course_start_date?:     string;
  fuzzy_start_date?:     string;
  course_end_date?:     string;
};

let createCourseRun = (course: Course, status: string, dateValues: CourseRunDateValues) => {
  return Object.assign({
    "position": 1,
    "title": "title",
    "course_id": "course-v1:odl+GIO101+FALL14",
    "status": status,
    "id": 1
  }, dateValues);
};

it(`should show date with status ${status}`) {
  let course = _cloneDeep(findCourse(course => course.status === STATUS_PASSED));
  course.runs = [
    createCourseRun(course, STATUS_PASSED, {
      'course_end_date': '2016-09-09T10:20:10Z'
    })
  ];
  let firstRun = course.runs[0];
  const wrapper = shallow(<CourseDescription course={course} />);
  let courseEndDate = moment(firstRun.course_end_date);
  let formattedDate = courseEndDate.format(DASHBOARD_FORMAT);
  assert.equal(wrapper.find(".course-description-result").text(), `Ended: ${formattedDate}`);
}
```

we can load the course from `DASHBOARD_RESPONSE`, but we create our own course run as part of the test since the course run data is what influences the description text. i think your `switch` statement should be broken up into individual test cases (`it()` blocks) and data should be created explicitly like above

gsidebo:
can you also look into updating the `CourseRun` type in static/js/flow/programTypes.js? it seems like that type has been updated as part of this PR (you added "course_start_date", "fuzzy_start_date", "course_end_date" to course run data in DASHBOARD_RESPONSE), but it also seems like this type definition was out-of-date before this PR

amir-qayyum-khan:
@gsidebo to create data here is tips

- Create course as admin and add start dates for course and enrollment before today and end date after today then you will see enroll button
 - screen shot 2016-09-12 at 10 40 27 pm
- Create a course with enrollment start date before today and course start date after today, i.e ``30/9/2016`` then You will se upgrade button.
 - screen shot 2016-09-12 at 10 43 48 pm
- Make a course and add end date before today you will see ``not-offered`` course and course run ``not-passed``
 - screen shot 2016-09-12 at 10 47 47 pm
- I dont know how to produce course passed status.

amir-qayyum-khan:
@gsidebo Done with changes you suggested.

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -31,80 +35,102 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
 }
 
 describe('CourseDescription', () => {
-  const now = moment();
+  let createCourseRun = (status: string, courseEndDate: string, courseStartDate: string) => {
```
gsidebo: nitpick: it makes more sense to position the `courseStartDate` before the `courseEndDate` argument

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -31,80 +35,102 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
-        assert.equal(text, "(enrolled)");
-      } else {
-        assert.equal(text, "");
-      }
-    }
-  });
+  it(`does show date with status passed`, () => {
+    let course = _.cloneDeep(findCourse(course => course.status === STATUS_PASSED));
+    course.runs = [
+      createCourseRun(STATUS_PASSED, '2016-10-09T10:20:10Z', '2016-10-01T10:20:10Z')
```
gsidebo: a few things: 

1. do these `courseStartDate` and `courseEndDate` arguments ever change in the 6 times you call this function? it looks like the same values are repeated many times. it would help to have default values for these arguments. that way we can easily identify which test cases use a 'special' date range
1. are these values sensitive to the actual current datetime? in other words, will anything change when the current datetime is inside this range, or beyond this range? just making sure

static/js/components/dashboard/CoursePrice.js
```diff
@@ -11,25 +16,79 @@ import { formatPrice } from '../../util/util';
...
 export default class CoursePrice extends React.Component {
   props: {
-    course: Course,
+    course: Course
   };
 
+  renderTooltip(text: string): React$Element<*>|void {
+    let tooltip;
+
+    if (text) {
```
gsidebo: i think you should just return the `<div>` directly in this function and not worry about whether the `text` value is blank

static/js/components/dashboard/CoursePrice.js
```diff
@@ -11,25 +16,79 @@ import { formatPrice } from '../../util/util';
...
+
+    return tooltip;
+  }
+
+  courseTooltipText(courseStatus: string): string {
+    let text = "";
+
+    if (courseStatus === STATUS_ENROLLED_NOT_VERIFIED) {
+      text = "You need to enroll in the Verified Course to get MicroMasters credit.";
+    }
```
gsidebo: this would be more concise

```javascript
if (courseStatus === STATUS_ENROLLED_NOT_VERIFIED) {
  return "You need to enroll in the Verified Course to get MicroMasters credit.";
}
return '';
```

static/js/components/dashboard/CoursePrice.js
```diff
@@ -11,25 +16,79 @@ import { formatPrice } from '../../util/util';
...
+    let courseHasPrice = (
+      !_.isNil(firstRun.price) &&
+      (courseStatus === STATUS_OFFERED_NOT_ENROLLED || courseStatus === STATUS_ENROLLED_NOT_VERIFIED)
+    );
+
+    if (courseHasPrice) {
+      price = formatPrice(firstRun.price);
+    }
+
+    return price;
```
gsidebo: same comment as above. you can just `return` once you match the appropriate logic to calculate the price. the local `price` variable isn't necessary

static/js/components/dashboard/CourseDescription.js
```diff
@@ -10,58 +10,72 @@ import {
...
       }
       break;
     }
 
+    return text;
+  }
+
+  render() {
+    const { course } = this.props;
+    let enrolled = "";
```
gsidebo: i don't understand why this is here when it can only be a blank string

static/js/components/dashboard/CourseAction_test.js
```diff
@@ -63,11 +63,13 @@ describe('CourseAction', () => {
...
     let description = wrapper.find(".course-action-description");
-    let buttonProps = buttonContainer.find("button").props();
+    let buttonProps = buttonContainer.find(".dashboard-button").props();
+    let spanText = buttonContainer.find(".sr-only").text();
+    let buttonText = buttonContainer.find(".dashboard-button").children().text();
     let firstRun = course.runs[0];
 
     assert(!buttonProps.disabled);
-    assert.equal(buttonContainer.text(), `Upgrade in ${firstRun.title}`);
+    assert.equal(`${buttonText}${spanText}`, `Upgrade in ${firstRun.title}`);
```
gsidebo: a few changes for these unit tests:

1) change `assertCheckoutButton` to accept a wrapped button object
```javascript
  let assertCheckoutButton = (button, courseId) => {
    button.simulate('click');
    assert.equal(checkoutStub.callCount, 1);
    assert.deepEqual(checkoutStub.args[0], [courseId]);
  };
```

2) create a new test case for the course title so we don't need to repeat it 6 times (the text doesn't change for any of these cases)
```javascript
it('shows the course title near the enroll button', () => {
    let course = findCourse(course => course.status === STATUS_ENROLLED_NOT_VERIFIED);
    const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
    let buttonContainer = wrapper.find(".course-action-action");
    let firstRun = course.runs[0];
    assert.include(buttonContainer.text(), firstRun.title);
  });
```

3) change the pattern of these unit tests to match this (i think it's cleaner, clearer, and slightly less repetitive)
```javascript
  it('shows an upgrade button if user is not verified but is enrolled', () => {
    let course = findCourse(course => course.status === STATUS_ENROLLED_NOT_VERIFIED);
    const wrapper = shallow(<CourseAction course={course} now={now} checkout={checkoutStub}/>);
    let buttonContainer = wrapper.find(".course-action-action");
    let description = wrapper.find(".course-action-description");
    let button = buttonContainer.find(".dashboard-button");
    let firstRun = course.runs[0];

    assert(!button.is(':disabled'));
    assert.equal(button.children().text(), 'Upgrade');
    assert.equal(description.text(), "");
    assertCheckoutButton(button, firstRun.course_id);
  });
```

static/js/components/dashboard/CourseDescription.js
```diff
@@ -10,58 +10,72 @@ import {
...
-      if (firstRun.status === STATUS_NOT_PASSED) {
-        text = 'You failed this course';
+      if (firstRun.status === STATUS_NOT_PASSED && firstRun.course_end_date) {
+        let courseEndDate = moment(firstRun.course_end_date);
+        text = this.courseDate('Ended', courseEndDate);
+      } else if (!_.isNil(firstRun.fuzzy_start_date)) {
+        text = `Coming ${firstRun.fuzzy_start_date}`;
       }
       break;
+    case STATUS_ENROLLED_NOT_VERIFIED:
```
gsidebo: you don't need to include `case` statements for cases that don't do anything

amir-qayyum-khan: @gsidebo these 3 consecutive statements are equal to 
```js
if (course.status === STATUS_ENROLLED_NOT_VERIFIED || course.status === STATUS_VERIFIED_NOT_COMPLETED || course.status === STATUS_OFFERED_NOT_ENROLLED )  
** same functionality**
```

Which mean these 3 case have same functionality, please see http://www.w3schools.com/js/js_switch.asp

static/js/components/dashboard/CourseAction_test.js
amir-qayyum-khan: @gsidebo 

2. Text is changing  buttonContainer.text() === (Enroll + course title) or (upgrade + course title) or (empty string)

amir-qayyum-khan:
Done @gsidebo

static/scss/dashboard.scss
```diff
@@ -184,14 +188,15 @@ c.validation-wrapper {
           }
 
           .dashboard-button {
-            background-color: $mm-brand-bg;
+            background-color: $mm-brand-bg !important;
```
gsidebo: just wondering: why is `!important` necessary here and below

amir-qayyum-khan: @gsidebo here we are using this css on a  button which was imported from ``react-mdl/lib/Button``. This is to override basic style of mentioned lib.

static/js/components/dashboard/CourseDescription.js
gsidebo: oh, interesting! somehow ive never seen that before. thanks for the lesson 😄

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -31,80 +35,102 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
+    assert.equal(wrapper.find(".course-description-result").text(), `Start date: ${formattedDate}`);
+  });
 
-    // course starts today
-    let wrapper = shallow(<CourseDescription course={course} now={courseStart}/>);
-    assert.equal(wrapper.find(".course-description-result").text(), "");
+  it(`does show date with status offered-not-enrolled`, () => {
+    let course = _.cloneDeep(findCourse(course => course.status === STATUS_OFFERED_NOT_ENROLLED));
+    course.runs = [
+      createCourseRun(STATUS_OFFERED_NOT_ENROLLED, '2016-05-09T10:20:10Z', '2016-08-01T10:20:10Z')
```
gsidebo: i see that you changed the date arguments you're passing into `createCourseRun` for all of these test cases. are these different values important for their test cases? it doesn't look like they are. my earlier suggestion was to set a default value for `courseStartDate` and `courseEndDate` in `createCourseRun`. that way, if the test case doesn't need some specific values for those dates, you can just call `createCourseRun(SOME_STATUS)` and not worry about passing in those values. `createCourseRun` would fill them in if they are undefined

gsidebo:
@amir-qayyum-khan re: CourseAction_test.js, i created a patch to show you the changes i was suggesting: https://gist.github.com/gsidebo/c19e5de6abdee263a0de777ff89b282e

my point was that we were redundantly testing the span text in multiple test cases. we only need to make sure that the two specific statuses produce that span text. i also made a couple other refactors. let me know what you think

you're right about the disabled attribute. i forgot that we're dealing with material-ui, not just HTML elements

static/scss/dashboard.scss
gsidebo: ok. i'll take a quick look at this but i won't delay the PR for it

gsidebo:
@amir-qayyum-khan once you take a look at that patch i linked, this looks good

amir-qayyum-khan:
@gsidebo Done with changes. Thanks

gsidebo:
if the build passes, this looks great! 👍

PR #999, opened 2016-09-09
Added elastisearch auth

### LINKED ISSUE #782: Add configuration to authenticate against Elasticsearch
In order to provide some measure of access control to our Elasticsearch cluster we will be using the elasticsearch-readonly-reset plugin (https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/wiki/Supported-Rules). This will require handling authentication in the proxy API based on environment variables.

/cc @giocalitri @noisecapella

giocalitri:
closes #782

#### What's this PR do?
Adds authentication key for elasticsearch.
It also adds a new admin (I needed to do some tests and I left it), but this is not necessary for this PR.
The key is not mandatory and is used only if set. 
If the backend is not set up to handle authentication, even if the key is set, everything will keep working.

`search/indexing_api.py`

This part is tricky: the docker elasticsearch should keep working. The actual authentication should be tested against the elasticsearch cluster when @blarghmatey finishes to set it up

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -1,0 +1,15 @@
+"""
+Admin views for Dashboard
+"""
+
+from django.contrib import admin
+
+from dashboard.models import ProgramEnrollment
+
+
+class ProgramEnrollmentAdmin(admin.ModelAdmin):
+    """ModelAdmin for ProgramEnrollment"""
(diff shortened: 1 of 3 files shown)


blarghmatey:
👍

PR #1000, opened 2016-09-09
Added enrollment dialog

### LINKED ISSUE #796: Learner Dashboard: allow learner to switch between MM programs (program context)

noisecapella:
Fixes #796 

#### What's this PR do?
Adds a dialog box which opens when the user clicks 'Enroll in a new program'. This dialog is used to add a new `ProgramEnrollment` for the user.

Make sure you are not enrolled in every live program. Then you should see the link in the program selector. Note that:

 - If you select the 'Enroll in a new program' option, it should not change the currently selected program displayed in the dropdown.
 - If you click 'Cancel' the dialog should close and nothing should change.
 - If you click 'Enroll' without a program selected you should get a validation warning. (The warning should disappear if you close and open the dialog again.)
 - You should see only programs you have not yet enrolled in.
 - When you select a program and click 'Enroll', it should close the dialog immediately and POST to the API. If it succeeds you should see a success message with a check mark like in the screenshot for #796. If it fails you should see an error icon with an error message. In both cases the toast will disappear after five seconds.
 - After a successful enrollment you will automatically be switched to the new program in the context.

The click outside the dialog to close it is not supported due to complications with material-ui and react-select: https://github.com/JedWatson/react-select/issues/532
There may be some workaround for that, I haven't spent much time on it. Made an issue: #1002

Also, program page URLs are not available in the dashboard at the moment so they won't be displayed in the UI. Made an issue: #1003 

screenshot from 2016-09-09 12-39-58

### DIFF
--- a/static/js/components/NewEnrollmentDialog.js
+++ b/static/js/components/NewEnrollmentDialog.js
@@ -1,0 +1,102 @@
+import React from 'react';
+import Dialog from 'material-ui/Dialog';
+import Button from 'react-mdl/lib/Button';
+import _ from 'lodash';
+
+import SelectField from 'material-ui/SelectField';
+import MenuItem from 'material-ui/MenuItem';
+import type { DashboardState } from '../flow/dashboardTypes';
+import type { ProgramEnrollmentsState } from '../flow/enrollmentTypes';
+
+export default class NewEnrollmentDialog extends React.Component {
+  props: {
+    addProgramEnrollment:        (programId: number) => void,
+    dashboard:                   DashboardState,
+    enrollments:                 ProgramEnrollmentsState,
+    enrollDialogError:           ?string,
+    enrollDialogVisibility:      boolean,
+    enrollSelectedProgram:       ?number,
+    setEnrollDialogError:        (error: string) => void,
+    setEnrollDialogVisibility:   (open: boolean) => void,
+    setEnrollSelectedProgram:    (programId: number) => void,
+  };
+
+  closeDialog = () => {
+    const { setEnrollDialogVisibility } = this.props;
+    setEnrollDialogVisibility(false);
+  };
+
+  addEnrollment = () => {
+    const {
+      addProgramEnrollment,
+      enrollSelectedProgram,
+      setEnrollDialogError,
+      setEnrollDialogVisibility,
+    } = this.props;
+
+    if (_.isNil(enrollSelectedProgram)) {
+      setEnrollDialogError("No program selected");
+    } else {
+      addProgramEnrollment(enrollSelectedProgram);
+      setEnrollDialogVisibility(false);
+    }
+  };
+
+  createDialogActions = () => {
+    return [
+      <Button
+        className="dialog-button cancel-button"
+        key="cancel"
+        onClick={this.closeDialog}
+      >
+        Cancel
+      </Button>,
+      <Button
+        className="dialog-button enroll-button"
+        key="enroll"
+        onClick={this.addEnrollment}
+      >
+        Enroll
+      </Button>,
+    ]
+  };
+
+  handleEnrollSelectedProgram = (event, index, value) => {
+    const { setEnrollSelectedProgram } = this.props;
+    setEnrollSelectedProgram(value);
+  };
+
+  render() {
+    const {
+      dashboard: { programs },
+      enrollDialogError,
+      enrollDialogVisibility,
+      enrollSelectedProgram,
+      enrollments: { programEnrollments },
+    } = this.props;
+
+    let enrollmentLookup = new Map(programEnrollments.map(enrollment => [enrollment.id, null]));
+    let unenrolledPrograms = programs.filter(program => !enrollmentLookup.has(program.id));
+    unenrolledPrograms = _.sortBy(unenrolledPrograms, 'title');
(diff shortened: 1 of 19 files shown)


static/js/components/NewEnrollmentDialog.js
```diff
@@ -0,0 +1,102 @@
...
+        className="dialog-button enroll-button"
+        key="enroll"
+        onClick={this.addEnrollment}
+      >
+        Enroll
+      </Button>,
+    ];
+  };
+
+  handleEnrollSelectedProgram = (event, index, value) => {
```
gsidebo: `handleSelectedProgramChange` is probably a better name for this

static/js/components/NewEnrollmentDialog.js
```diff
@@ -0,0 +1,102 @@
...
+  render() {
+    const {
+      dashboard: { programs },
+      enrollDialogError,
+      enrollDialogVisibility,
+      enrollSelectedProgram,
+      enrollments: { programEnrollments },
+    } = this.props;
+
+    let enrollmentLookup = new Map(programEnrollments.map(enrollment => [enrollment.id, null]));
```
gsidebo: `Map` doesn't seem to be supported fully in IE11 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

static/js/reducers/ui_test.js
```diff
@@ -266,6 +270,48 @@ describe('ui reducers', () => {
...
+      it('should let you set the dialog error', () => {
+        return dispatchThen(setEnrollDialogError("value"), [
+          SET_ENROLL_DIALOG_ERROR
+        ]).then(state => {
+          assert.equal(state.enrollDialogError, "value");
+        });
+      });
+    });
+
+    describe('enrollment success/failure message', () => {
```
gsidebo: the body of this describe block is exactly the same as the block at line 273 (apart from the 'it' output)

static/js/reducers/ui_test.js
```diff
@@ -266,6 +270,48 @@ describe('ui reducers', () => {
...
+      });
+
+      it('should let you set the message', () => {
+        return dispatchThen(setEnrollMessage("value"), [
+          SET_ENROLL_MESSAGE
+        ]).then(state => {
+          assert.equal(state.enrollMessage, "value");
+        });
+      });
+    });
```
gsidebo: we're using this exact same test pattern a few dozen times in this file now. would be nice to have a helper method

```javascript
export const assertReducerResultState = (action, actionType, stateParamName, value) => {
  return dispatchThen(action(value), [actionType]).then(state => {
    assert.equal(state[stateParamName], value, `$(action) should set state.$(stateParamName) to $(value)`);
  });
};

assertReducerResultState( setEnrollMessage, SET_ENROLL_MESSAGE, 'enrollMessage', 'new_value' );
```

readability with 4 arguments isn't the greatest, but at this point i think it's well worth it. your default value check could also be implemented with a helper method but that's just 1 line anyway - you can take or leave that one

static/js/components/Toast.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+
+export default class Toast extends React.Component {
```
gsidebo: is this just a component to pop up a notification to the user? whatever it is, it could definitely use a more descriptive name

noisecapella: Toast is a standard name for a popup notification to a user: https://material.google.com/components/snackbars-toasts.html

gsidebo: fair enough 🍞

static/js/components/NewEnrollmentDialog.js
noisecapella: Ugh...

noisecapella: Actually it looks like babel-polyfill takes care of this for us. I tested this out on IE11 and the `new Map(entries)` constructor is supported when on micromasters-ci

static/js/components/NewEnrollmentDialog.js
noisecapella: Sure

static/js/reducers/ui_test.js
noisecapella: Yeah, it seems worth refactoring this

noisecapella:
Comments addressed, I'm going back to working on tests now

noisecapella:
Ok, tests are in place. I skipped testing Toast in the interest of saving time, and there's not too much going on there anyway

gsidebo:
looks great! 👍

PR #1011, opened 2016-09-12
avoid violating the GPL

### LINKED ISSUE #1008: GPL License violation
Currently you use wowjs it's under the GPL. I maintain a fork (wow.js) see https://wowjs.uk

This version is completely UMD compatible and doesn't need any hacks.

You could also use AOS instead

graingert:
fixes #1008


graingert:
note, this PR does not exempt you from the MIT license. You should still include the license in your end product.

pdpinch:
Thanks @graingert. We're going to be updating the design of our home page this sprint or next and we may remove this library altogether. If we don't, we'll review your PR.

noisecapella:
We just got rid of this package with #1062 so we shouldn't need this PR anymore.

PR #1012, opened 2016-09-12
Disabled SanctuaryJS run-time type checking in production

### LINKED ISSUE #985: (Bug) Graduation date year field in education form is slow
On the Education page of the signup flow, editing the "Graduation Date" year field is very slow. The response to input is significantly slower than other text inputs. Possibly significant: each character entered appears to trigger an UPDATE_PROFILE action

year_field_slowness

aliceriot:

closes #985 

#### What's this PR do?

This PR disables the runtime type checking (see [this](https://github.com/sanctuary-js/sanctuary-def)) for Sanctuary in production. Optionally, it could be disabled in development as well, by editing a file. This significantly speeds up the date fields (which are using sanctuary's `Maybe` type constructor).

I think we want to keep the runtime type-checking in development, because if you're using the sanctuary stuff at all it's actually super helpful. However, it's pretty slow, so I think we should definitely turn it off for production.

Make sure the changes look legit. I was following along with what the sanctuary repo recommends here: https://github.com/sanctuary-js/sanctuary#type-checking

Pull down the code. Go to a profile editing page (either `/users` or `/profile`) and edit a date field. Observe how slow it is (it's really slow).

Then, open `util/sanctuary.js`. Change line 6 from this:

```js
export const S = create({ checkTypes: checkTypes, env: env });
```

to this:

```js
export const S = create({ checkTypes: false, env: env });
```

Reload the app. editing date fields should be significantly faster now. This is how the date field should behave in production.

### DIFF
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -5,7 +5,8 @@ import _ from 'lodash';
 import TextField from 'material-ui/TextField';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 import Checkbox from 'material-ui/Checkbox';
-import S, { Maybe } from 'sanctuary';
+import { S } from './sanctuary';
+const { Maybe } = S;
 import R from 'ramda';
 
 import { ISO_8601_FORMAT } from '../constants';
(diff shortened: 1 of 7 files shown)


noisecapella:
:+1:

PR #1019, opened 2016-09-13
Added Financial Aid basic models

giocalitri:
Not a real ticket: more a baseline for others to work on

#### What's this PR do?
Introduces the basic models for the Financial Aid

`financialaid/models.py`
`courses/models.py`

I guess just verify that the admin contains the new models.

based on the discussion that produced this:
img_5822

Does not contain the models for the algorithm


financialaid/admin.py
```diff
@@ -0,0 +1,57 @@
...
+    """Admin for Coupons"""
+    model = Coupons
+
+
+class DocumentsAdmin(admin.ModelAdmin):
+    """Admin for Documents"""
+    model = Documents
+
+
+class FinancialAidAdmin(admin.ModelAdmin):
```
zags: There is a possible data privacy question as to whether or not we should expose this model in the django admin.  We haven't finalized auditing requirements yet, but it is possible we will have to track any time someone views the name and income of a person.  If this is exposed through the admin, someone will be able to view this information without us being able to track it

financialaid/models.py
```diff
@@ -0,0 +1,110 @@
...
+
+class TierProgram(models.Model):
+    """
+    The tiers for discounted pricing assigned to a program
+    """
+    program = models.ForeignKey(Program, null=False)
+    tier = models.ForeignKey(Tiers, null=False)
+    value = models.IntegerField(null=False)
+    current = models.BooleanField(null=False, default=False)
+
```
zags: Given what we know about the algorithm, financial aid tiers are based purely on income (country only comes into play for auto-approval, and is not used in determining the tier for which a user goes into review).  Given this, I think we should add an income threshold to the TierProgram model.  Then, we can assign tiers to users by selecting the tier for a program with the lowest threshold that is still greater than that user's income

financialaid/models.py
```diff
@@ -0,0 +1,110 @@
...
+        per program and tier
+        """
+        if self.current:
+            TierProgram.objects.filter(program=self.program, tier=self.tier, current=True).update(current=False)
+        return super(TierProgram, self).save(*args, **kwargs)
+
+
+class FinalcialAidStatus:
+    """Statuses for the Financial Aid model"""
+    CREATED = 'created'
```
zags: When does someone have the "created" status?

I'm pretty sure we can always start someone on auto_approved or pending_docs

financialaid/models.py
```diff
@@ -0,0 +1,110 @@
...
+    financial_aid = models.ForeignKey(FinancialAid, null=True)
+
+
+class Documents(models.Model):
+    """
+    Documents to attach to a financial aid application
+    """
+    financial_aid = models.ForeignKey(FinancialAid, null=True)
+    name = models.TextField()
+    url = models.TextField()
```
zags: If we're going to use Django Forms, we should make this a `models.FileField`

financialaid/models.py
```diff
@@ -0,0 +1,110 @@
...
+    exchange_rate = models.FloatField(null=True)
+    date_exchange_rate = models.DateTimeField(null=True)
+
+
+class Coupons(models.Model):
+    """
+    Coupons stored to be used by the users who are approved to use financial aid
+    """
+    course_run = models.ForeignKey(CourseRun, null=False)
+    code = models.CharField(null=False, max_length=50)
```
zags: I think we need a foreign key to ProgramTier on coupons as well, given a coupon has a specific discount value.

financialaid/models.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class FinancialAidAudit(models.Model):
+    """
+    Audit table for the Financial Aid
+    """
+    user = models.ForeignKey(User, null=False)
+    table_changed = models.CharField(null=False, max_length=50)
+    data_before = JSONField(blank=True, null=False)
+    date = models.DateTimeField(null=False)
```
zags: This looks great for tracking data modifications.  Just adding a note that we will have to expand this if we also have to track data views

zags:
Just a couple tweaks.  Otherwise looks good to me.

financialaid/admin.py
giocalitri: I agree: I put it now just because it is probably useful while we write the code. We can remove it when we realize what goes in this table.

giocalitri: Actually, I got confused: I think we should get clarification about this. The Admin should not be accessible to many people anyway

financialaid/models.py
giocalitri: my idea is that when the entry is created the first time it gets `created`, then it changes to `auto_approved` or `pending_docs ` according to the algorithm.
This is mainly because I want the `FinancialAidAudit` to have all the log of the changes and the timestamps.

financialaid/models.py
giocalitri: that's right: I am going to change it

financialaid/models.py
giocalitri: I agree in principle, but I see this as a potential nightmare when we are going to import the coupons.
I will add it anyway.

financialaid/models.py
giocalitri: I agree.

financialaid/models.py
giocalitri: I am confused by what you are saying, but I think this change can be made when we implement the algorithm.

financialaid/models.py
giocalitri: Actually, there is a foreign key to the `FinancialAid` model that has a foreign key to `TierProgram`.
The only reason to have a direct foreign key is if we think that we could assign a coupon (that belongs to a tier "x") for a financial aid application that has an assigned tier "y", but I see this extremely dangerous.

giocalitri:
@zags I think that I answered all your comments. If you think that the status of the PR is fine, please give a `thumbs up` icon

zags:
:+1:
We'll change a couple fields as we hit them

financialaid/admin.py
zags: Update from peter: we don't need to store document views, so we're good on this

financialaid/models.py
zags: I was just thinking about unallocated coupons; we need to know which tier they belong to so that we can give a user the right coupon upon the user being approved

financialaid/models.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from jsonfield import JSONField
+
+from courses.models import (
+    CourseRun,
+    Program,
+)
+
+
+class Tiers(models.Model):
```
gsidebo: these model classes should be singular, not plural (~~Tiers~~ -> Tier, ~~Coupons~~ -> Coupon, ~~Documents~~ -> Document)

giocalitri: good point: changing

financialaid/models.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        Override the save to enforce the existence of only one `current` = True
+        per program and tier
+        """
+        if self.current:
+            TierProgram.objects.filter(program=self.program, tier=self.tier, current=True).update(current=False)
+        return super(TierProgram, self).save(*args, **kwargs)
+
+
+class FinalcialAidStatus:
```
noisecapella: You have a typo here FYI

PR #1029, opened 2016-09-14
Removed privacy page from signup flow

### LINKED ISSUE #899: Remove Profile Policy screen from profile sign-up
This functionality has been moved to the user settings pane

gsidebo:

Closes #899 

#### What's this PR do?

Removes the privacy page (4th in order) from the signup flow

Set filled_out to `false` for your user and run through the signup flow. Make sure validation still works. The privacy form in the settings page should also be tested.

I also fixed up some js tests that used `Object.assign({}, USER_PROFILE_RESPONSE, {...})`. Object.assign can produce some [unexpected behavior with deep cloning](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) (see "Warning for Deep Clone") and I had a very nasty time with a js test last week as a result of it


aliceriot:
functionality looks good! reading code now.

static/js/components/EmploymentTab.js
```diff
@@ -22,9 +28,16 @@ class EmploymentTab extends React.Component {
...
-          isLastTab={false}
-          validator={employmentValidation}
+          nextBtnLabel="I'm Done!"
+          isLastTab={true}
+          validator={
+            combineValidators(
+              personalValidation,
+              educationValidation,
+              employmentValidation,
+              privacyValidation
```
aliceriot: do we still want the privacy validator here? what would the user see if there were an error?

gsidebo: whoopsie daisy

aliceriot:
w/r/t the `Object.assign` / `_.cloneDeep` thing, we've got a lot of defensive programming around this around the codebase. `Object.assign` just copies values over, it doesn't do any kind of recursive clone, so you could have the following:

```js
let obj = { arr: [1,2] }

let other = Object.assign({}, obj);

obj.arr === other.arr // -> true
```

is that more or less what you ran into?

there's a good reason for this, namely performance considerations, but yeah, I agree the behavior can be a little unexpected. You shouldn't need to avoid `Object.assign` if you're just modifying top-level properties, but it's a good idea to use `_.cloneDeep` or `R.clone` if you're going to mutate any nested objects on the resulting object.

static/js/components/EmploymentTab.js
aliceriot: :D

aliceriot:
except for the privacy validator thing above this looks good!

looks like there's a failing test though.

gsidebo:
@aliceriot ill dig up the Object.assign example i found from that test last week

gsidebo:
@aliceriot here's one way to reproduce the `Object.assign` behavior.

1. open static/js/containers/UserPage_test.js (in this branch or master - doesn't matter)
1. insert these console.log lines at line 289
    ```javascript
    console.log(USER_PROFILE_RESPONSE.education.length);
    console.log('');
    ```

1. run the 'Education history' test cases in that file: `node ./node_modules/mocha/bin/_mocha --compilers js:babel-register static/js/global_init.js static/js/containers/UserPage_test.js --grep 'Education history'`
1. observe that the length of `USER_PROFILE_RESPONSE.education` remains the same
1. change line 274 to remove the `_.cloneDeep` --> `userProfile = Object.assign({}, USER_PROFILE_RESPONSE, {`
1. run the step 3 command again
1. observe that `USER_PROFILE_RESPONSE.education` grows with every subsequent test. this is due to line 277 (`userProfile.education.push`). we assume that line won't alter `USER_PROFILE_RESPONSE`, but it does

aliceriot:
Yup, same problem, `Object.assign` is just copying a reference to the `education` array, so when we call `.push` it's getting bigger everytime.

aliceriot:
wee new github features! i guess the above is equivalent to a 👍 ?

PR #1031, opened 2016-09-14
change travis.yml around a little

### LINKED ISSUE #1030: move setup commands in `travis.yml` into the `install` section
Our travis build output is super long, and a good chunk of it is `apt`, `pip` and `npm` spewing out package install info.

If we move these commands out of `script` and put them into `install` we can automatically collapse all that stuff - it's still there to look at if we want, but it's collapsed by default.

@noisecapella @giocalitri @blarghmatey is there any reason not to do this?

aliceriot:

closes #1030 

#### What's this PR do?

this basically adds an `install` section to `travis.yml`, and puts a command to build the docker containers in there, and the command to install JS packages. This basically results in the output of these commands being collapsed by default when you look at the build output, which means it's a lot shorter to scroll through (and it's easier to see the tests!).

Make sure the build is still passing and all that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 # Set Ruby as the language so it doesn't download the pip things. Instead, let docker do that.
 language: ruby
 cache: bundler
+install:
+  - docker-compose -f travis-docker-compose.yml build
+  - docker-compose -f travis-docker-compose.yml run watch npm install --quiet
+
 script:
   - docker-compose -f travis-docker-compose.yml run web tox
-  - docker-compose -f travis-docker-compose.yml run watch npm install --quiet
   - docker-compose -f travis-docker-compose.yml run watch npm run-script coverage
   - docker-compose -f travis-docker-compose.yml run watch npm run-script lint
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,12 @@
 # Set Ruby as the language so it doesn't download the pip things. Instead, let docker do that.
 language: ruby
 cache: bundler
+install:
+  - docker-compose -f travis-docker-compose.yml build
```
noisecapella: What does this do? Is it done automatically before the script section but not before the install section?

aliceriot: I think if you ask docker to run a container that is hasn't yet built it will build it first automatically, so that:

```
docker-compose run web tox
```

is more or less equivalent to

```
docker-compose build && docker-compose run web tox
```

basically what adding the `build` command does is separates building the `watch` and `web` containers from running the test commands, so we don't have to look at all the output from installing packages and pulling the containers from docker hub.

noisecapella:
:+1:

PR #1033, opened 2016-09-14
Removed course status

### LINKED ISSUE #1032: Dashboard: handle courses with zero runs

noisecapella:
Fixes #1032

#### What's this PR do?
 - Removes the course status from the dashboard API. From now on the first run will have the appropriate status. If there are no runs 'Coming soon...' will be displayed, from [this mockup](https://projects.invisionapp.com/share/NB7VM1GSG#/screens/185611170)
 - The `not-offered` status is removed because it was only used with the course status, not the run status.
 - Handle case where there are no runs to display

If there are no future course runs for a given course (this is the default for the realistic fake data we have) then you should now see 'Coming soon...' similar to the screenshot. Otherwise all data should be the same as before.

screenshot from 2016-09-14 15-18-40

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -29,25 +29,24 @@ REFRESH_ENROLLMENT_CACHE_MINUTES = 5
 
 class CourseStatus:
     """
-    Possible statuses for a course for a user
+    Possible statuses for a course for a user. These are the course run statuses used in the dashboard API.
     """
     PASSED = 'passed'
     NOT_PASSED = 'not-passed'
     CURRENT_GRADE = 'verified'
     UPGRADE = 'enrolled'
-    NOT_OFFERED = 'not-offered'
     OFFERED = 'offered'
 
     @classmethod
     def all_statuses(cls):
         """Helper to get all the statuses"""
         return [cls.PASSED, cls.NOT_PASSED, cls.CURRENT_GRADE,
-                cls.UPGRADE, cls.NOT_OFFERED, cls.OFFERED]
+                cls.UPGRADE, cls.OFFERED]
 
 
 class CourseRunStatus:
     """
-    Possible statuses for a course run for a user
+    Possible statuses for a course run for a user. These are used internally.
     """
     NOT_ENROLLED = 'not-enrolled'
     GRADE = 'grade'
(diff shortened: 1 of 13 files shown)


dashboard/api.py
```diff
@@ -187,96 +182,48 @@ def get_info_for_course(user, course, user_enrollments, user_certificates):
     # remove the picked run_status from the list
     run_statuses.remove(run_status)
 
+    def add_run(next_run, status, certificate=None):
+        """Helper function to add a course run to the status dictionary"""
+        course_data['runs'].append(
+            format_courserun_for_dashboard(
+                next_run,
```
aliceriot: also just curious, is there any performance downside to declaring this inside another function? I was thinking instead it could take the `course_data` as a parameter and mutate it.

static/js/components/dashboard/CourseAction.js
```diff
@@ -75,6 +77,15 @@ export default class CourseAction extends React.Component {
...
       break;
     }
+    case STATUS_NOT_PASSED:
+    case STATUS_VERIFIED:
+      // do nothing;
+      break;
+    default: {
+      // there are no runs
+      action = this.makeEnrollButton("Enroll", firstRun, true);
+      break;
```
aliceriot: I think this isn't recessary

dashboard/api.py
```diff
@@ -350,6 +284,7 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
         status_for_user (str): a string representing the status of a course for the user
         certificate (Certificate): an object representing the
             certificate of the user for this run
+        position (int): The position of the course run within the list
```
noisecapella: It's to tell the dashboard what order the runs are in within a course. Though having them in a list in that order anyway seems to make that unnecessary. We don't actually use it anywhere

noisecapella:
Comments addressed (not sure how to do that in the new UI)

aliceriot:
cool, taking a look now

dashboard/api.py
aliceriot: gotcha - well, I'm fine with leaving it for now, but if we're really not using it for anything we should probs delete it.

dashboard/api.py
```diff
@@ -138,18 +136,29 @@ def get_info_for_program(program, user, enrollments, certificates):
...
         data['courses'].append(
-            get_info_for_course(user, course, enrollments, certificates)
+            get_info_for_course(course, enrollments, certificates)
         )
     data['courses'].sort(key=lambda x: x['position_in_program'])
     return data
 
 
-def get_info_for_course(user, course, user_enrollments, user_certificates):
+def _add_run(course_data, next_run, status, certificate=None):
```
aliceriot: I still feel like this parameter should be called `run`, and not `next_run`, what do you think?

aliceriot:
:+1: :shipit:

PR #1036, opened 2016-09-14
Only one role per User can be assigned

### LINKED ISSUE #1034: One role type per user
The user should not be allowed to have different roles, not even in different programs.

This is a limitation of the library we are using right now.

giocalitri:
closes #1034

#### What's this PR do?
Enforces the existence of only one role type per user

`roles/models.py`

try to assign `Staff` and `Admin` to the same user in the django Admin

This is a limitation of the library we are using and given that there are thoughts to create a new `reviewer` role, it makes sense to be sure there is not confusion on this front.

There is a known issue with this code: even if in the Admin you get an error, there will still be a confirmation message. AFAIK this cannot be removed because the confirmation message is generated after the error message. I do not think this is a problem given that only people who access the Admin can see it.
screen shot 2016-09-14 at 4 36 01 pm

### DIFF
--- a/roles/models.py
+++ b/roles/models.py
@@ -3,7 +3,8 @@ General models for the micromasters app
 """
 
 from django.contrib.auth.models import User
-from django.db import models
+from django.core.exceptions import ValidationError
+from django.db import models, transaction
 
 from courses.models import Program
 from roles.roles import Staff, Instructor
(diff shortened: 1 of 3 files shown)


noisecapella:
Code looks fine overall, testing functionality now...

noisecapella:
Just to clarify, a user can be 'staff' of many programs, so the thing that this PR prevents is a user being staff of any program and instructor of any program at the same time?

noisecapella:
When I add a new 'instructor' role to a user who is staff, I get this error:
screenshot from 2016-09-15 10-41-47

giocalitri:
> Just to clarify, a user can be 'staff' of many programs, so the thing that this PR prevents is a user being staff of any program and instructor of any program at the same time?

yes

giocalitri:
I see the error as well: I will try to deal with it.

giocalitri:
Ok I fixed the problem and in the process I implemented a better solution that prevents also the problem of the previous screenshot from happening
screen shot 2016-09-15 at 2 38 38 pm

noisecapella:
Looks great :+1:

giocalitri:
@noisecapella done

noisecapella:
:+1:

PR #1046, opened 2016-09-15
Set search facets to be open by default

### LINKED ISSUE #931: Instructor search: facets should be open by default
On loading the page, it should look like:

image

Not like:

image

gsidebo:

closes #931 

#### What's this PR do?

Sets search facets to be open by default

Check the learner search page. Also ensure that minimizing these facets will persist


noisecapella:
:+1:

PR #1053, opened 2016-09-15
Financial Aid: creating new requests

### LINKED ISSUE #1041: Financial Aid: creating new requests
# Acceptance criteria:

- [x] API endpoint for new financial aid request (takes income and currency)
  we can focus on USD for now
- [x] Calculation of pending financial aid tier
- [x] Auto acceptance criteria for requests
- [x] Import of per-country thresholds for auto acceptance; these will be global for all programs

alvinsiu:
closes #1041 - tests are included

The endpoint (currently `reverse("financialaid_api")`) takes "original_currency", "original_income", and "program_id".  Currenly only "USD" is valid for "original_currency".

> Calculation of pending financial aid tier

This is in financialaid.backend: `determine_tier_program(program, income)`

> Auto acceptance criteria for requests

This is in financialaid.backend: `determine_auto_approval(financial_aid)`.

> Import of per-country thresholds for auto acceptance; these will be global for all programs

The data for country to income threshold is loaded in financialaid.constants into a constant `COUNTRY_INCOME_THRESHOLDS`, which reads the a csv file from data.

The crux of the functionality is in financialaid.serializers and financialaid.backend.


cms/migrations/0015_auto_20160914_2033.py
```diff
@@ -0,0 +1,42 @@
...
+            model_name='programpage',
+            name='program',
+            field=models.OneToOneField(help_text='The program for this page', null=True, on_delete=django.db.models.deletion.SET_NULL, to='courses.Program'),
+        ),
+        migrations.AlterField(
+            model_name='programpage',
+            name='thumbnail_image',
+            field=models.ForeignKey(blank=True, help_text='Thumbnail size must be at least 690x530 pixels. Thumbnails are cropped down to this size, preserving aspect ratio.', null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='+', to='wagtailimages.Image'),
+        ),
+    ]
```
giocalitri: Why there is this file?
I do not see any changes to the CMS models: why there is a set of migrations?

giocalitri:
also, the content of such files gets loaded from disk every time the process starts: I have doubts this is a good idea. @noisecapella ?

giocalitri:
please rename `financialaid/backend.py` to `financialaid/api.py`

giocalitri:
Pease add `Args` and `Returns` section for any docstring of any (main) function

financialaid/backend.py
```diff
@@ -0,0 +1,35 @@
...
+
+def determine_tier_program(program, income):
+    """
+    Determines and returns the TierProgram for a given income.
+    """
+    # To determine the tier for a user, find the set of every tier whose income threshold is
+    # less than or equal to the income of the user. The highest tier out of that set will
+    # be the tier assigned to the user.
+    tier_programs_set = program.tier_programs.filter(current=True, income_threshold__lte=income)
+    tier_program = tier_programs_set.order_by("-income_threshold")[0]
```
giocalitri: you can use `.first()` instead of `[0]`

giocalitri: also, why two rows here?

financialaid/backend.py
```diff
@@ -0,0 +1,35 @@
...
+    return tier_program
+
+
+def determine_auto_approval(financial_aid):
+    """
+    Takes income and country code and returns a boolean if auto-approved.
+    """
+    income_threshold = COUNTRY_INCOME_THRESHOLDS.get(financial_aid.country_of_income, DEFAULT_INCOME_THRESHOLD)
+    is_above_threshold = financial_aid.income_usd >= income_threshold
```
giocalitri: I think it is above if it is `>` not `>=`

financialaid/backend.py
```diff
@@ -0,0 +1,35 @@
...
+
+    if is_above_threshold:
+        financial_aid.status = FinancialAidStatus.AUTO_APPROVED
+    else:
+        financial_aid.status = FinancialAidStatus.PENDING_DOCS
+    financial_aid.save()
```
giocalitri: I do not think that saving the change in this function is a good idea:

this should just return a boolean, not changing the DB

financialaid/backend_tests.py
```diff
@@ -0,0 +1,91 @@
+"""
+Tests for financialaid view
+"""
+from django.db.models.signals import post_save
+
```
giocalitri: nit, please remove space here

financialaid/backend_tests.py
```diff
@@ -0,0 +1,91 @@
...
+from dashboard.models import ProgramEnrollment
+from financialaid.backend import determine_tier_program, determine_auto_approval
+from financialaid.factories import TierProgramFactory, FinancialAidFactory
+from profiles.factories import ProfileFactory
+from search.base import ESTestCase
+
+
+class FinancialAidBackendTests(ESTestCase):
+    """
+    Tests for financialaid views
```
giocalitri: this should be `Tests for financialaid backend` or `api` like I asked before

financialaid/backend_tests.py
```diff
@@ -0,0 +1,91 @@
...
+        self.tiers = {
+            "0k": TierProgramFactory.create(program=self.program, income_threshold=0, current=True),
+            "15k": TierProgramFactory.create(program=self.program, income_threshold=15000, current=True),
+            "50k": TierProgramFactory.create(program=self.program, income_threshold=50000, current=True),
+            "100k": TierProgramFactory.create(program=self.program, income_threshold=100000, current=True)
+        }
+        self.program_enrollment = ProgramEnrollment.objects.create(
+            user=self.profile.user,
+            program=self.program
+        )
```
giocalitri: If all this code does not need to run once per test, please move it to `setUpTestData`

financialaid/backend_tests.py
```diff
@@ -0,0 +1,91 @@
...
+    def test_determine_auto_approval(self):  # pylint: disable=no-self-use
+        """
+        Tests determine_auto_approval()
+        """
+        # Assumes US threshold is 100000
+        financial_aid = FinancialAidFactory.create(
+            income_usd=150000,
+            country_of_income="US"
+        )
+        assert determine_auto_approval(financial_aid)
```
giocalitri: please assert `is True` or `is False` for any of the following tests

financialaid/constants.py
```diff
@@ -0,0 +1,14 @@
...
+import csv
+import os
+
+from django.conf import settings
+
+
+DEFAULT_INCOME_THRESHOLD = 100000
+with open(os.path.join(settings.BASE_DIR, "data/country_income_thresholds.csv"), mode="r") as file:
+    reader = csv.DictReader(file)
+    COUNTRY_INCOME_THRESHOLDS = {row["country"]: int(row["income"]) for row in reader}
```
giocalitri: I think this should be run once offline and have the dictionary in this module instead of loading the file each time

financialaid/factories.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+import datetime
+from factory import SubFactory
+
+from factory.django import DjangoModelFactory
+from factory.fuzzy import (
+    FuzzyAttribute,
+    FuzzyInteger,
+    FuzzyText,
+    FuzzyChoice, FuzzyFloat, FuzzyDateTime)
```
giocalitri: I would import `fuzzy` and not the single imports, but if you want to keep in this way, please one import per line

financialaid/factories.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TierFactory(DjangoModelFactory):
+    """
+    Factory for Tier
+    """
+    name = FuzzyText()
+    description = FuzzyText()
+
+    class Meta:  # pylint: disable=missing-docstring, no-init, too-few-public-methods, old-style-class
```
giocalitri: is it really failing the pylint for `missing-docstring, no-init, too-few-public-methods, old-style-class` ?

financialaid/factories.py
```diff
@@ -0,0 +1,63 @@
...
+from factory.fuzzy import (
+    FuzzyAttribute,
+    FuzzyInteger,
+    FuzzyText,
+    FuzzyChoice, FuzzyFloat, FuzzyDateTime)
+import faker
+from pytz import UTC
+
+from courses.factories import ProgramFactory
+from financialaid.models import TierProgram, Tier, FinancialAidStatus, FinancialAid
```
giocalitri: one import per line and sorted

giocalitri:
I do not see any reason for having multiple migrations for the financialaid app in this PR: please remove them and re-create only one

financialaid/models.py
```diff
@@ -27,10 +27,11 @@ class TierProgram(models.Model):
     """
     The tiers for discounted pricing assigned to a program
     """
-    program = models.ForeignKey(Program, null=False)
+    program = models.ForeignKey(Program, null=False, related_name="tier_programs")
     tier = models.ForeignKey(Tier, null=False)
-    value = models.IntegerField(null=False)
+    discount_amount = models.IntegerField(null=False)
```
giocalitri: so at the end are we going with a fixed discount and not a percentage?

financialaid/models.py
```diff
@@ -46,14 +47,14 @@ def save(self, *args, **kwargs):
...
+class FinancialAidStatus:
     """Statuses for the Financial Aid model"""
     CREATED = 'created'
-    APPROVED = 'approved'
     AUTO_APPROVED = 'auto-approved'
-    REJECTED = 'rejected'
     PENDING_DOCS = 'pending_docs'
     PENDING_MANUAL_APPROVAL = 'pending_manual_approval'
+    APPROVED = 'approved'
+    REJECTED = 'rejected'
```
giocalitri: why this change?

financialaid/serializers.py
```diff
@@ -0,0 +1,54 @@
+"""
+Serializers from financialaid
+"""
+import datetime
```
giocalitri: nit: please insert a space after this

financialaid/serializers.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    original_income = FloatField(min_value=0)
+    original_currency = CharField()
+    program_id = IntegerField()
+
+    def validate(self, data):
+        """
+        Validators for this serializer
+        """
+        data["program"] = get_object_or_404(Program, pk=data["program_id"])
```
giocalitri: will this raise in case `"program_id"` has not been submitted?

giocalitri: also I would prefer to raise custom exceptions so we can intercept them in the view and return proper meaningful errors to the UI

financialaid/serializers.py
```diff
@@ -0,0 +1,54 @@
...
+        if not data["program"].financial_aid_availability:
+            raise ValidationError("Financial aid not available for this program.")
+        if not ProgramEnrollment.objects.filter(program=data["program"], user=self.context["request"].user).exists():
```
giocalitri: question: which kind of constraints gives the fact that you pull the user from the context?

financialaid/views_test.py
```diff
@@ -0,0 +1,75 @@
...
+from factory.django import mute_signals
+from rest_framework.status import HTTP_201_CREATED, HTTP_400_BAD_REQUEST
+from rest_framework.test import APIClient
+
+from financialaid.backend_tests import FinancialAidBackendTests
+from financialaid.models import FinancialAid
+from profiles.factories import ProfileFactory
+
+
+class FinancialAidViewTests(FinancialAidBackendTests, APIClient):
```
giocalitri: you should not import here another test class `FinancialAidBackendTests`

Instead create a common base class and import that one here

financialaid/views_test.py
```diff
@@ -0,0 +1,75 @@
...
+    def setUp(self):
+        super().setUp()
+        with mute_signals(post_save):
+            self.profile2 = ProfileFactory.create()
+        self.client.force_login(self.profile.user)
+        self.data = {
+            "original_currency": "USD",
+            "program_id": self.program.id,
+            "original_income": 80000
+        }
```
giocalitri: I am pretty sure that some of these statement can be moved to `setUpTestData`

giocalitri:
I just finished to give a first pass:
I think that overall it is good and many of the changes I asked are not that significant

cms/migrations/0015_auto_20160914_2033.py
alvinsiu: I merged master into this branch - not my changes

financialaid/backend.py
zags: It was a long line and we wanted to split it up logically, according to the steps described by the comment above the line.

financialaid/backend.py
zags: you're right.  They did have it as `>`

financialaid/models.py
zags: that's what peter said he preferred yesterday

financialaid/serializers.py
zags: Because the view subclasses from GenericAPIView, it converts ValidationErrors into 400 responses with the error message as the response body.  It's the same behavior as ProgramEnrollmentListView, so it should work fine on the frontend.  The validator has builtin validation for the fields as described, including not passing program_id

financialaid/serializers.py
zags: The request is passed to the context from the view (this is behavior of GenericAPIView)

financialaid/models.py
alvinsiu: It made more sense to me to be in order of status change flow.

cms/migrations/0015_auto_20160914_2033.py
giocalitri: No, the file has been introduced by this commit:
https://github.com/mitodl/micromasters/pull/1053/commits/82372ff2a6059e78f683a72cfb0b02794cca741d

And in general, even if you merged master, if this file was on taste, was not appearing as a new one on your PR.

Please figure out why it is here and if it should not be, remove it.

giocalitri:
I noticed that a bunch of files have the permission modified from 644 to 755. Please restore the permissions.

financialaid/backend.py
giocalitri: ok

financialaid/models.py
giocalitri: ok

financialaid/models.py
giocalitri: ok

financialaid/serializers.py
giocalitri: what I meant is something like [this](https://github.com/mitodl/micromasters/blob/master/search/views.py#L34-L38), but I do not have strong opinions at the moment

financialaid/serializers.py
giocalitri: ok, but does this mean I cannot use the serializer outside the context of a request?

financialaid/models.py
pdpinch: Just to make a note of the reasoning: if we uses percentage off coupons, it limits the potential values we can set for adjusted price. But if we use fixed discounts, we can potentially set any price.

financialaid/api.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def determine_tier_program(program, income):
+    """
+    Determines and returns the TierProgram for a given income.
+    Args:
+        program: Program
+        income: numeric
+    Returns:
+        TierProgram
```
giocalitri: Args and Returns should include the type and a brief description.
See [this](https://github.com/mitodl/micromasters/blob/master/mail/api.py#L47-L52) as an example

giocalitri:
There are a bunch of comments that still need to be addressed: please change the labe to "Needs review" again only if all the comments have been addressed or you have given an answer for not changing the code.

cms/migrations/0015_auto_20160914_2033.py
alvinsiu: Hmm, looks like you're right.  This was probably created accidentally when we made migrations and we didn't notice.  I'll remove it.

financialaid/data/country_to_currency.csv
```diff
@@ -0,0 +1,250 @@
...
+VU,VUV
+VA,EUR
+VE,VEF
+VN,VND
+WF,XPF
+EH,MAD
+YE,YER
+ZM,ZMW
+ZW,ZWL
+AX,EUR
```
alvinsiu: We'll do it later when it's used, given that Peter said this is lower priority now

financialaid/serializers.py
alvinsiu: We can customize later if needed when we know what the frontend wants for validation.

financialaid/serializers.py
alvinsiu: Probably not, but we don't anticipate needing to do that. If a case comes up, we can modify it then.

alvinsiu:
I think everything is now addressed, unless I missed something.

financialaid/data/country_to_currency.csv
giocalitri: ok

financialaid/migrations/0002_auto_20160916_0529.py
```diff
@@ -0,0 +1,50 @@
...
+            name='income_threshold',
+            field=models.IntegerField(default=None),
+            preserve_default=False,
+        ),
+        migrations.AlterField(
+            model_name='tierprogram',
+            name='program',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='tier_programs', to='courses.Program'),
+        ),
+    ]
```
giocalitri: please rename this migration file to something meaningful

financialaid/serializers.py
giocalitri: you missed this

financialaid/serializers.py
giocalitri: ok

financialaid/serializers.py
giocalitri: ok

financialaid/views_test.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_income_validation(self):
+        """
+        Tests IncomeValidationView post endpoint
+        """
+        # Test that a FinancialAid object is created and is not auto-approved
+        assert FinancialAid.objects.count() == 0
+        resp = self.client.post(reverse("financialaid_api"), self.data, format='json')
```
giocalitri: please define  `reverse("financialaid_api")` as a variable during the `setUpTestData`

financialaid/api_tests.py
```diff
@@ -0,0 +1,93 @@
+"""
```
noisecapella: Please rename this file to `api_test.py`. Our testing code only detects based on `*_test.py` currently

.gitignore
```diff
@@ -93,3 +93,6 @@ staticfiles/
 release-notes-checklist
 
 static/scss/material.css
+
+# Local settings
+micromasters.yml
```
noisecapella: What's this for? It's fine to leave it, just wondering if this is forthcoming in the future

financialaid/constants.py
```diff
@@ -0,0 +1,224 @@
+"""
+Constants for financialaid
+"""
+
+
+DEFAULT_INCOME_THRESHOLD = 100000
+COUNTRY_INCOME_THRESHOLDS = {
```
noisecapella: @pdpinch this data will all be exposed in the source code, is that fine with you?

giocalitri:
Something I noticed during my functional test is that the code does not handle the case when a tier is not available and raises an exception because the tier in the financial aid table is not nullable

financialaid/api_tests.py
alvinsiu: Ah, my bad on the typo. Thanks!

alvinsiu:
Good catch, @giocalitri - how do you think we should handle that case?  We could just default to the highest tier as a safe catch, but it seems like this is more of something for the course staff to fix.

giocalitri:
I think this should be handled by an exception and proper logging without assigning anything.
This should never happen if the application is configured properly.

financialaid/constants.py
noisecapella: What's the source for the threshold data here and in the CSV? I am getting some disprepancies with the CSV files I got in an email from Anna. I imported those CSVs, removed the blank lines, and converted these country codes to three letter country codes.

    In [78]: set(expected.keys()) - set(thresholds.keys())
    Out[78]: {'ADO', 'CHI', 'IMY', 'KSV', 'ROM', 'TMP', 'WBG', 'ZAR'}
    
    In [79]: set(thresholds.keys()) - set(expected.keys())
    Out[79]: {'AND', 'COD', 'IMN', 'PSE', 'ROU', 'TLS', 'XKX'}

noisecapella: For countries which exist in both lists it looks like the thresholds all match up as expected

giocalitri:
I set up my tiers like the ones in the test and submit a post with:
income 10.000
currency EUR
program id (the one that has financial aid enabled)
and I get this exception:

```
[2016-09-16 15:55:37] ERROR 81 [django.request] base.py:284 - [bb14125580fe] - Internal Server Error: /api/v0/financialaid/
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 174, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 172, in get_response
    response = response.render()
  File "/usr/local/lib/python3.5/site-packages/django/template/response.py", line 160, in render
    self.content = self.rendered_content
  File "/usr/local/lib/python3.5/site-packages/rest_framework/response.py", line 71, in rendered_content
    ret = renderer.render(self.data, media_type, context)
  File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py", line 675, in render
    context = self.get_context(data, accepted_media_type, renderer_context)
  File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py", line 654, in get_context
    'filter_form': self.get_filter_form(data, view, request),
  File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py", line 592, in get_filter_form
    queryset = view.get_queryset()
  File "/usr/local/lib/python3.5/site-packages/rest_framework/generics.py", line 67, in get_queryset
    % self.__class__.__name__
AssertionError: 'IncomeValidationView' should either include a `queryset` attribute, or override the `get_queryset()` method.
[16/Sep/2016 15:55:37] "POST /api/v0/financialaid/ HTTP/1.1" 500 116081
```

giocalitri:
Something else I noticed is that you are assuming that the currency of income corresponds to the country of income. 
This is probably the norm, but I think I might be able to be paid in USD (or at least have a contract that says so) and live in Canada (as an example)

Ferdi:
@giocalitri just as side comment.  You guys dont have to spent too much time on the details like "from canada getting paid in US dollars".  You are right it could come up but not worth it spending much time.

alvinsiu:
@giocalitri I have to run to a meeting - I'll fix the exception when I'm back.  I've pushed changes to the other comments.

giocalitri:
@Ferdi I was just running some tests (one of which returned an exception) and logging what I thought I found, not saying we should deal with it now.
BTW that was not the case and I was wrong in that specific case.

giocalitri:
Ok, I finished to look at the functionality and I cannot find anything else other than what I already did.

financialaid/api_test.py
```diff
@@ -0,0 +1,93 @@
...
+        assert determine_tier_program(self.program, 0) == self.tiers["0k"]
+        assert determine_tier_program(self.program, 1000) == self.tiers["0k"]
+        assert determine_tier_program(self.program, 15000) == self.tiers["15k"]
+        assert determine_tier_program(self.program, 23500) == self.tiers["15k"]
+        assert determine_tier_program(self.program, 50000) == self.tiers["50k"]
+        assert determine_tier_program(self.program, 72800) == self.tiers["50k"]
+        assert determine_tier_program(self.program, 100000) == self.tiers["100k"]
+        assert determine_tier_program(self.program, 34938234) == self.tiers["100k"]
+
+    def test_determine_auto_approval(self):  # pylint: disable=no-self-use
```
noisecapella: Can you add a test that a tier with `current=False` is ignored?

financialaid/api_test.py
```diff
@@ -0,0 +1,93 @@
...
+            country_of_income="US"
+        )
+        assert determine_auto_approval(financial_aid) is True
+        financial_aid = FinancialAidFactory.create(
+            income_usd=1000,
+            country_of_income="US"
+        )
+        assert not determine_auto_approval(financial_aid)
+
+        # Assumes IN threshold is 15000
```
noisecapella: Can you add `assert COUNTRY_INCOME_THRESHOLDS['IN'] == 15000` here and in the other cases being tested?

financialaid/api_test.py
```diff
@@ -0,0 +1,93 @@
...
+        assert determine_tier_program(self.program, 15000) == self.tiers["15k"]
+        assert determine_tier_program(self.program, 23500) == self.tiers["15k"]
+        assert determine_tier_program(self.program, 50000) == self.tiers["50k"]
+        assert determine_tier_program(self.program, 72800) == self.tiers["50k"]
+        assert determine_tier_program(self.program, 100000) == self.tiers["100k"]
+        assert determine_tier_program(self.program, 34938234) == self.tiers["100k"]
+
+    def test_determine_auto_approval(self):  # pylint: disable=no-self-use
+        """
+        Tests determine_auto_approval()
```
noisecapella: Can you test for a `50000` country as well?

financialaid/serializers.py
giocalitri: Actually I thought more about this and I think we should explicitly pass the user username together with the POST request and not pull it from the context.
This has the complication that we need an extra permission class on the view to enforce that the user can change only her data.
This might be easier if we modify the url to include the username.

I am open to discuss this if you do not agree, but my reason is that the serializer should not be tied to a request context.

alvinsiu:
@giocalitri The error you encountered was for returning a view to the user, which hasn't been defined/created yet (and it's outside the scope of this PR).  I've pushed some changes anyway to address this error so that you can use the DRF helper pages to test later if you'd like.  We'll modify this later as needed once the pages have been created and we know what needs to be passed to it.

@noisecapella I've updated the tests per your suggestions!

noisecapella:
@alvinsiu  Can you look at the other comments I made? I didn't mark them 'request changes' (still getting familiar with the new Github interface) but they are all part of this review

financialaid/serializers.py
noisecapella: I agree with @giocalitri, this check should be done in the view.

financialaid/views.py
```diff
@@ -0,0 +1,24 @@
+"""
+Views from financialaid
```
giocalitri: nit: `from` -> `for`?

financialaid/constants.py
zags: This was the issue I mentioned on hipchat a few days ago.  The files sent to us by the econ department used the word bank codes, not the ISO 3166-3 codes. which are mostly overlapping, except for the set you are seeing here.  We corrected these manually when we converted to ISO-31662 codes to be consistent with the rest of the system

noisecapella: Sounds good

.gitignore
zags: This file is an alternate means of specifying environment variables.  This comes from the django cookiecutter and is defined in settings.py

financialaid/views_test.py
```diff
@@ -0,0 +1,90 @@
...
+        assert FinancialAid.objects.count() == 1
+        financial_aid = FinancialAid.objects.first()
+        assert financial_aid.tier_program == self.tiers["50k"]
+        assert financial_aid.status == FinancialAidStatus.PENDING_DOCS
+        # Test that a FinancialAid object is created and is auto-approved
+        self.data["original_income"] = 200000
+        resp = self.client.post(self.url, self.data, format='json')
+        assert resp.status_code == HTTP_201_CREATED
+        assert FinancialAid.objects.count() == 2
+        financial_aid = FinancialAid.objects.last()
```
alvinsiu: Good point - I've split these up into two separate test cases since the order_by seems a little messy.

financialaid/serializers.py
```diff
@@ -0,0 +1,62 @@
...
+from rest_framework.exceptions import ValidationError
+from rest_framework.fields import IntegerField, FloatField, CharField
+
+from courses.models import Program
+from dashboard.models import ProgramEnrollment
+from financialaid.api import determine_tier_program, determine_auto_approval
+from financialaid.models import FinancialAid, FinancialAidStatus
+
+
+class FinancialAidSerializer(serializers.Serializer):
```
alvinsiu: We originally did, but we also needed to pass in a field (program) that isn't on the model, which lead to a bunch of issues with the ModelSerializer's auto validation and data type checking.  We decided it was more efficient to just change it to a regular serializer instead of trying to debug getting it to work with a ModelSerializer.

financialaid/serializers.py
```diff
@@ -0,0 +1,62 @@
...
+            user=user,
+            income_usd=self.validated_data["original_income"],
+            country_of_income=user.profile.country,
+            date_exchange_rate=datetime.datetime.now()
+        )
+
+        if determine_auto_approval(financial_aid) is True:
+            financial_aid.status = FinancialAidStatus.AUTO_APPROVED
+        else:
+            financial_aid.status = FinancialAidStatus.PENDING_DOCS
```
alvinsiu: This was done intentionally for audits. @giocalitri (correct me if I'm wrong) wanted to have it initialized with status "CREATED" before we did any status changes.

financialaid/views_test.py
alvinsiu: Ah, I now see your comment above - I'm still getting used to this review system too haha.

financialaid/serializers.py
alvinsiu: @giocalitri If you can create the permission class, we'd be happy to stick it on the serializer.  I'm also open to discussing this more at a later time, but I don't think we should let this small detail hold up the pull request in the meantime.  Would that alright?

alvinsiu:
I think the latest commit should address all of the remaining issues - let me know if I missed anything, as I'm also getting familiar with the new interface.  Thanks!

financialaid/serializers.py
giocalitri: I was not talking about the permission class in the serializer. I was saying that the post request should submit the user username and not pulling the user from the request context.

The permission class is intended for the view to control who can submit the POST request.

financialaid/serializers.py
giocalitri: Yes, I actually thought that it was better to keep these two steps to be separated.

giocalitri:
For me there is only the "user in the serializer" left, but I will let @noisecapella decide what's best.

noisecapella:
@alvinsiu Can you address this comment? https://github.com/mitodl/micromasters/pull/1053#r79202297

financialaid/serializers.py
noisecapella: But `program` is on the `FinancialAid` model though...

Anyway in the interest of speeding this up feel free to leave it, but this solution throws out a lot of free validation.

financialaid/api.py
```diff
@@ -30,4 +30,4 @@ def determine_auto_approval(financial_aid):
         boolean: True if auto-approved, False if not
     """
     income_threshold = COUNTRY_INCOME_THRESHOLDS.get(financial_aid.country_of_income, DEFAULT_INCOME_THRESHOLD)
-    return financial_aid.income_usd > income_threshold
+    return financial_aid.income_usd > income_threshold or financial_aid.income_usd == 0
```
alvinsiu: Ah, I meant to write income_threshold == 0, good catch, thanks!

financialaid/api_test.py
```diff
@@ -0,0 +1,112 @@
...
+        assert determine_tier_program(self.program, 1000) == self.tiers["0k"]
+        assert determine_tier_program(self.program, 15000) == self.tiers["15k"]
+        assert determine_tier_program(self.program, 23500) == self.tiers["15k"]
+        assert determine_tier_program(self.program, 50000) == self.tiers["50k"]
+        assert determine_tier_program(self.program, 72800) == self.tiers["50k"]
+        assert determine_tier_program(self.program, 100000) == self.tiers["100k"]
+        assert determine_tier_program(self.program, 34938234) == self.tiers["100k"]
+        assert determine_tier_program(self.program, 34938234) != self.tiers["150k_not_current"]
+
+    def test_determine_auto_approval(self):  # pylint: disable=no-self-use
```
noisecapella: Can you test for `income_usd=0` and a tier which is not `0k`?

financialaid/serializers.py
noisecapella: The way the code is now is bad design and changing it to use a permission class is a trivial change IMO. But I don't think it's worth holding up the review as long since we test this

financialaid/serializers.py
alvinsiu: Program is not on FinancialAid - it's connected via TierProgram.  We did spent a while trying to wrangle ModelSerializer to work, but decided it wasn't worth spending hours over when this current solution took us 5 minutes to make work.  I definitely agree with the interest of speeding this up.

financialaid/views_test.py
alvinsiu: The self.data is kept here because the data can be modified in the test and we don't want to encounter pass by reference issues.  The force_login was left here in case a future test case wanted to test logging in a different user for permissions reasons.  Let us know if you want us to move it to setUpTestData anyway; we're happy to do that!

alvinsiu:
@noisecapella I've added the test case as requested.

noisecapella:
The comment I made was on `test_determine_auto_approval`. Can you add a test there which tests an income tier that's not `0k` with an income of `0`?

noisecapella:
:+1:

noisecapella:
Please make sure the merge is squashed, either by Github or with a git rebase, and wait for tests to pass before merging

alvinsiu:
Ah, I've made an off by one error in where I thought the comment pertained.  I've added the test!

PR #1054, opened 2016-09-15
Set "Place of Birth" facet to accept multiple values

### LINKED ISSUE #986: Search: location facets should allow multiple selections
Users should be able to perform an "OR" search by selecting multiple locations from the "Current Residence" and "Country of Birth" facets.

for a functionality example, see the "country" facet at http://demo.searchkit.co/ 

# Acceptance Criteria

- [ ] User should be able to select more than one value for each of these facets above.
- [ ] checking a facet updates the search results (in the right pane) but not facet counts (left pane)

gsidebo:

closes #986 

#### What's this PR do?

Allows a user to select multiple values for the "Place of Birth" facet. Those values are treated in an 'OR' fashion - any user that matches any one of the values selected will be in the search results

Run the app and navigate to the learner's search page. Test with multiple values for Place of Birth

Due to some searchkit limitations, we've decided not to pursue the same change for Current Location (see issue comments for details)

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -129,6 +129,7 @@ export default class LearnerSearch extends SearchkitComponent {
                   id="birth_location"
                   title="Place of Birth"
                   field="profile.birth_country"
+                  operator="OR"
                   itemComponent={CountryRefinementOption}
                 />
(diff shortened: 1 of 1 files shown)


giocalitri:
👍

Issue #919, opened 2016-08-26
Home Page: update template to match current design


pdpinch:
![mm_lp 2x](https://cloud.githubusercontent.com/assets/430126/18008793/9863c4c4-6b77-11e6-968d-46dbd8c59ef8.jpg)

# acceptance criteria:

- [ ] match current design, including template & styles
- [ ] header consistent with the rest of the site (#730)
- [ ] Pull Program thumbnails, descriptions and links from program model (see also #653/#658, #573)
- [ ] fix CSS issue (#892)
- [ ] update styling of mail form to match [new design](https://cloud.githubusercontent.com/assets/430126/18179970/e612462a-7052-11e6-8c8b-e67d58abc293.png)
- [ ] fix https bug in mail form #965

# related issues:

- #944 allow updates to program thumbnails, etc. 
- #917 Sign Up / Get started modal

PR #1056, opened 2016-09-16
Style changes for the home page

roberthouse54:

#919 

#### What's this PR do?

This PR fixes the css styles for the Home Page

See that the code looks good, and that the home page styling is not broken.

Maybe test in a couple browsers. NOTE: The layout is not responsive for mobile.


roberthouse54:
Also NOTE: The final product ended up different than the mock. The reason for this was that Marketing requested a late color palette change to use MIT's maroon instead of the green we have had. To make these changes required other color palette changes. Now that the release got pushed back, I decided to stick with the green and use the extra time to rethink the color palette. But the preparatory changes I made in the design I am leaving in, because I think it looks good as is.

cms/templates/cms/home_page.html
```diff
@@ -145,22 +155,23 @@
 
                   {% if program.programpage %}
                     {% if program.programpage.external_program_page_url %}
-                      <a href="{{ program.programpage.external_program_page_url }}" class="program-link">Learn More</a>
+                      <p><a href="{{ program.programpage.external_program_page_url }}" class="program-link">Learn More</a></p>p>
```
noisecapella: Can you fix the `></p>p>` at the end here?

noisecapella:
It looks like there are conflicts now. Do you want me to make the changes I mentioned and then merge this? The changes are pretty minor, the PR is generally in good shape

Ferdi:
@noisecapella that would be great .

Issue #1045, opened 2016-09-15
Financial Aid: Review Page


pdpinch:
Simple list of financial aid submissions, for review. 

![image](https://cloud.githubusercontent.com/assets/430126/18748865/50a040c8-80a2-11e6-9715-0a7ce2a60c0f.png)

# Acceptance Criteria:

- [ ] Remove "reject" button  /cc @roberthouse54 
- [ ] Rename "approve" checkmark with label "Save"  /cc @roberthouse54 
- [ ] Move "adjusted cost" column close to "Save" button
- [ ] Add a dropdown with a couple of predefined values that will allow reviewer to log a justification for adjusting the cost (e.g  "Docs not notarized" , "Insufficient Docs")
- [ ] Button for documents received
- [ ] Date calculated
- [ ] Date docs sent
- [ ] Radio button for displaying income in dollars or local currency (as calculated, not real time)
- [ ] Modified `PUT` requests for `FinancialAidActionView` to `PATCH` requests to better follow REST
- [ ] Changed `FinancialAidMailView` to use a serializer and to take a `FinancialAid` object instead of email recipient for security and auto-message-generation reasons
- [ ] Linked backend functionality for emailing, approving, and marking documents as received to the frontend
- [ ] UI selectively shows/hides sections and functionality depending on the status selected to view
- [ ] Updated labels and copy
- [ ] Search by name functionality
- [x] Make some of the columns sortable 
- [x] API endpoint for requests given a status (searchable, sortable and pageable)
- [x] simple search for user name
- [x] fields for each request include income of an applicant (as submitted in local currency and as dollars as calculated), date of calculation, document status
- [x] Limit API endpoints to staff in the appropriate program
- [x] API endpoint to get dollar values of each tier (for populating drop-down)

pdpinch:
Note that we modified this UI to allow the reviewer to choose the adjusted cost and override the cost proposed by the algorithm. Allowing this reviewer discretion means we can stop debating the outcome of rejected applications.

zags:
I talked with @giocalitri and we decided that the best approach for this page and it's children was to implement them as Django templates using Django forms rather than API endpoints with react.  This is because it will be faster to build them that way, and they are only accessible to select staff

alvinsiu:
[...]

PR #1071, opened 2016-09-20
Implemented backbone for review financial aid page

alvinsiu:
#1045 

#### What's this PR do?
Implements the backbone template and view for reviewing financial aid applications.  This PR does not implement any of the functionality from the page (aside from pagination and sorting).


ui/views_test.py
```diff
@@ -208,7 +208,7 @@ def test_roles_setting(self):
             {
                 'program': role.program.id,
                 'role': role.role,
-                'permissions': [key for key, value in available_perm_status(profile.user).items()],
+                'permissions': [key for key, value in available_perm_status(profile.user).items() if value is True],
```
noisecapella: This change looks correct to me. Can you confirm @giocalitri?

micromasters/urls.py
```diff
@@ -37,6 +37,7 @@
     url(r'^api/v0/mail/$', MailView.as_view(), name='mail_api'),
     url(r'^api/v0/financialaid/$', IncomeValidationView.as_view(), name='financialaid_api'),
     url(r'^status/', include('server_status.urls')),
+    url(r'^financial-aid/', include('financialaid.urls')),
```
noisecapella: Can you change the URL to `financial_aid` with an underscore instead of a dash?

financialaid/templatetags/financial_aid_tags.py
```diff
@@ -0,0 +1,17 @@
+"""
```
noisecapella: Instead of defining these two template tags please do this calculation in the view and send the result to the template to be displayed.

financialaid/views.py
```diff
@@ -1,24 +1,205 @@
...
+    def get_context_data(self, **kwargs):
+        """
+        Gets context for view
+        """
+        context = super().get_context_data(**kwargs)
+
+        # Constants required in view
+        context["selected_status"] = self.selected_status
+        context["current_program_id"] = self.program.id
+        # Note: This implementation of retrieving a course price is a naive lookup that assumes
```
noisecapella: This is an incorrect assumption. Prices will certainly be different from program to program at least.

financialaid/models.py
```diff
@@ -10,20 +9,40 @@
...
+    Base model for create/update timestamps
+    """
+    created_on = models.DateTimeField(auto_now_add=True)  # UTC
+    updated_on = models.DateTimeField(auto_now=True)  # UTC
+
+    def update(self, **kwargs):
+        """
+        Automatically update updated_on timestamp
+        """
+        update_fields = {"updated_on"}
```
noisecapella: What does this do? Isn't the timestamp already updated on save?

financialaid/views.py
```diff
@@ -1,24 +1,205 @@
...
+        # all course runs and courses will be the same price for the foreseeable future,
+        # irrespective of the program. Therefore we can just take the price from any currently
+        # enroll-able course run.
+        course_price = CoursePrice.objects.filter(
+            is_valid=True,
+            course_run__in=CourseRun.objects.filter(CourseRun.get_active_enrollment_queryset())
+        ).first()
+        context["course_price"] = course_price.price
```
noisecapella: What if a course run doesn't have a price?

noisecapella:
@giocalitri Can you take a look?

financialaid/views.py
alvinsiu: @pdpinch confirmed this is fine for the foreseeable future "That said, it's safe to use the course prices of any run that's available for enrollment. All courses will be the same price, for the foreseeable future."

financialaid/templatetags/financial_aid_tags.py
alvinsiu: We decided to do this because it prevents the need to iterate through all the returned database objects again to attach the extra information.  Django's .annotate is sufficient to cover removing the use of subtract(), but it won't be able to replace the get_attribute().

I'm happy to redo the implementation to avoid the need for templatetags, but this is probably better in terms of performance.  Thoughts?

ui/views_test.py
giocalitri: It seems reasonable to me (even if it does not make much sense if the  a value = False is not expected in the rest of the code).

micromasters/urls.py
alvinsiu: Yup

financialaid/views.py
alvinsiu: This would be an issue with the initial configuration of the course/course runs.  If there's a particular exception you think we should raise, we can definitely add it in - we're just not familiar enough with how the initial set up process works to know what is a meaningful exception (unless you recommend creating a custom exception or just raising a generic exception with a helpful message).

financialaid/templatetags/financial_aid_tags.py
noisecapella: I would prefer moving this to the views to make the code cleaner. Templates should be relatively simple and computation and lookups are generally cleaner when written in Python rather than the template syntax

financialaid/models.py
alvinsiu: We've encountered issues in the past where the timestamp is not updated because someone used queryset.update() (which does not call .save() because .update() happens on the database level without loading the object(s) into memory).

ui/views_test.py
alvinsiu: This test failed after adding the extra permissions - it was trying to check to make sure the response was the same as expected, and the return response has `if value is True` as part of it.  The reason it wasn't failing earlier is because for both roles right now, all of the permissions are True.

financialaid/templatetags/financial_aid_tags.py
alvinsiu: Sounds good - we'll move it to the view then.

financialaid/views.py
pdpinch: Sorry @alvinsiu I meant that to be per program, not all course in the the entire application. 

I thought you were asking about variability within in the economics program. It's definitely the case that DEDP/economics courses will cost different from SCM courses.

alvinsiu: Ah, we'll change the implementation to be program-specific then, thanks!

alvinsiu:
@noisecapella I believe everything has been updated as per your requests.  Thanks for the review!

financialaid/models.py
```diff
@@ -1,7 +1,6 @@
 """
```
giocalitri: please set the permissions of this file back to 644

financialaid/views_test.py
```diff
@@ -82,13 +135,78 @@ def test_income_validation_user_not_enrolled(self):
...
+                "program_id": self.program.id,
+                "status": "aksdjfk"
+            }
+        )
+        resp = self.client.get(url_with_bad_filter)
+        assert resp.status_code == HTTP_200_OK
+        # Shouldn't break with invalid filter and sort fields
+        url_with_bad_filter_and_bad_sorting = "{url}?sort_by=-askjdf".format(url=url_with_bad_filter)
+        resp = self.client.get(url_with_bad_filter_and_bad_sorting)
+        assert resp.status_code == HTTP_200_OK
```
giocalitri: this test is way too big: is there any way to break it in multiple tests?

ui/views_test.py
giocalitri: I actually read the rest of the code just now

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+                sort_direction=self.sort_direction,
+                sort_field=self.sort_field_mappings.get(self.sort_field, self.sort_field)
+            )
+        )
+
+        # Add adjusted cost and TierPrograms valid for selection for each FinancialAid object
+        # Note: iterating through this queryset, if large, can be very costly in terms of memory/time,
+        # but it is done here to simplify the templates.
+        # Get program ids and associated tier programs
+        program_ids = Program.objects.filter(
```
noisecapella: Why are we making this list here? `self.program.id` is the only id which matters here

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+        ).values_list(
+            "id",
+            flat=True
+        )
+        # Get course price to calculate adjusted cost
+        # Note: This implementation of retrieving a course price is a naive lookup that assumes
+        # all course runs and courses will be the same price for the foreseeable future,
+        # irrespective of the program. Therefore we can just take the price from any currently
+        # enroll-able course run.
+        course_price = CoursePrice.objects.filter(
```
noisecapella: We need to handle the case where the CoursePrice is not set for whatever reason. Can you return an empty queryset if the `course_price` is `None` here?

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+            is_valid=True,
+            course_run__in=CourseRun.objects.filter(
```
noisecapella: Can you combine these lines into `course_run__course__program=self.program`?

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+                course__program_id=self.program.id
+            ).filter(
+                CourseRun.get_active_enrollment_queryset()
+            )
+        ).first()
+        tier_programs = {
```
noisecapella: We should only be looking for tier programs connected to `self.program`. Also, we should move this code into `get_context_data` above since the list of tier programs here is not dependent on any particular `FinancialAid`.

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+        "adjusted_cost": "adjusted_price",
+        "reported_income": "income_usd",
+    }
+    default_sort_field = "first_name"
+
+    def test_func(self):
+        """
+        Validate user permissions (Analogous to permissions_classes for DRF)
+        """
+        self.program = get_object_or_404(Program, id=self.kwargs.get("program_id", None))
```
noisecapella: Can you filter on `live=True` here?

financialaid/views_test.py
```diff
@@ -82,13 +135,78 @@ def test_income_validation_user_not_enrolled(self):
...
         """
-        Tests IncomeValidationView post  Only takes USD
+        Tests IncomeValidationView post; only takes USD
         """
         self.data["original_currency"] = "NOTUSD"
-        resp = self.client.post(self.url, self.data)
+        resp = self.client.post(self.income_validation_url, self.data)
         assert resp.status_code == HTTP_400_BAD_REQUEST
+
+    def test_review_financial_aid_view_not_allowed(self):
```
noisecapella: Can you add a test for the case where the program does not exist? Where the program exists but is not live?

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+        # Note: This implementation of retrieving a course price is a naive lookup that assumes
+        # all course runs and courses will be the same price for the foreseeable future,
+        # irrespective of the program. Therefore we can just take the price from any currently
+        # enroll-able course run.
+        course_price = CoursePrice.objects.filter(
+            is_valid=True,
+            course_run__in=CourseRun.objects.filter(
+                course__program_id=self.program.id
+            ).filter(
+                CourseRun.get_active_enrollment_queryset()
```
noisecapella: I don't think we need to filter by active enrollments here

financialaid/views_test.py
```diff
@@ -82,13 +135,78 @@ def test_income_validation_user_not_enrolled(self):
...
+        """
+        Tests ReviewFinancialAidView that are not allowed
```
noisecapella: Can you test for the case where `CoursePrice` with `is_valid=True` does not exist within the program?

financialaid/views_test.py
noisecapella: Also please test for `financial_aid_availability` and add this check in `test_func`

financialaid/models.py
alvinsiu: Yup!

financialaid/views_test.py
alvinsiu: I can break it up by type - I'll go ahead and do that

financialaid/views.py
```diff
@@ -1,24 +1,229 @@
...
+        Gets queryset for ListView to return to view
+        """
+        # Filter by availability
+        financial_aids = FinancialAid.objects.filter(
+            tier_program__program__live=True,
+            tier_program__program__financial_aid_availability=True
+        )
+
+        # Filter by program (self.program set in test_func())
+        financial_aids = financial_aids.filter(
```
alvinsiu: Ah - I had them separate for readibility, but I forgot about the potential issues with chaining filters on foreign keys.  I'll combine these.

financialaid/views.py
alvinsiu: You're absolutely right - I built this originally for all financial aid objects, and this slipped through the crack when I then limited it to within the program.

financialaid/views.py
alvinsiu: Good idea

financialaid/views.py
alvinsiu: That makes more sense

financialaid/views.py
alvinsiu: Wow not sure how I missed that, good catch

financialaid/views.py
alvinsiu: Yup

financialaid/views_test.py
alvinsiu: Will do

financialaid/views.py
alvinsiu: I took it out for now  - let me know if you want me to put it back in.

noisecapella: Thanks. I think since we are operating under the assumption that prices are the same for each course run in a program, they should be the same no matter whether the run is under active enrollment or not. At some point in the future we'll look into simplifying this.

noisecapella: Also, in general do not worry about the little performance gains here and there. We should write clean code first and optimize only if there's an immediate need for it. There will probably never be greater than say, 10000 course runs over the life of this system, and at that level it doesn't really matter how that query is filtered as long as we avoid things like n+1 selects.

I know, famous last words, but if performance problems come up it will be a long time in the future and we can deal with optimizing then.

financialaid/models.py
```diff
@@ -10,20 +9,42 @@
...
 
-class Tier(models.Model):
+class TimeStampedModel(models.Model):
+    """
+    Base model for create/update timestamps
+    """
+    created_on = models.DateTimeField(auto_now_add=True)  # UTC
+    updated_on = models.DateTimeField(auto_now=True)  # UTC
+
+    def update(self, **kwargs):
```
noisecapella: I'm having trouble getting this to work. My code:

    In [24]: t = Tier.objects.create()
    
    In [25]: t.id
    Out[25]: 4
    
    In [26]: t.updated_on
    Out[26]: datetime.datetime(2016, 9, 21, 13, 0, 6, 931005, tzinfo=<UTC>)
    
    In [27]: t.save()
    
    In [28]: t.updated_on
    Out[28]: datetime.datetime(2016, 9, 21, 13, 0, 16, 165478, tzinfo=<UTC>)
    
    In [29]: Tier.objects.filter(id=t.id).update(name='new name')
    Out[29]: 1
    
    In [30]: Tier.objects.get(id=t.id).updated_on
    Out[30]: datetime.datetime(2016, 9, 21, 13, 0, 16, 165478, tzinfo=<UTC>)

How is this function supposed to be used?

financialaid/views.py
```diff
@@ -1,24 +1,217 @@
...
+    }
+    default_sort_field = "first_name"
+
+    def test_func(self):
+        """
+        Validate user permissions (Analogous to permissions_classes for DRF)
+        """
+        self.program = get_object_or_404(
+            Program,
+            id=self.kwargs.get("program_id", None),
```
noisecapella: You should be able to use `self.kwargs['program_id']` here instead of `get`, the `program_id` argument should always exist since it's set in `urls.py`

financialaid/views.py
```diff
@@ -1,24 +1,217 @@
...
+
+    def get_queryset(self):
+        """
+        Gets queryset for ListView to return to view
+        """
+        # Get course price to calculate adjusted cost - we put this first so that we can return
+        # an empty queryset if no valid CoursePrice is found.
+        # Note: This implementation of retrieving a course price is a naive lookup that assumes
+        # all course runs and courses will be the same price for the foreseeable future,
+        # irrespective of the program. Therefore we can just take the price from any currently
```
noisecapella: Can you remove `irrespective of the program`

financialaid/views.py
```diff
@@ -1,24 +1,217 @@
...
+        ).first()
+        if course_price_object is None:
+            # If course price is not set, we can't meaningfully display any financial aid requests
+            return []
+        else:
+            self.course_price = course_price_object.price
+
+        # Filter by availability and program (self.program set in test_func())
+        financial_aids = FinancialAid.objects.filter(
+            tier_program__program__live=True,
```
noisecapella: We should be able to remove these two filters since we already know `self.program` was filtered this way in `test_func`

financialaid/templates/review_financial_aid.html
```diff
@@ -0,0 +1,112 @@
...
+              <th>Income/Yr.</th>
+              <th>Date Calculated</th>
+              <th></th>
+            </tr>
+          </thead>
+          <tbody>
+          {% for financial_aid in financial_aid_objects %}
+            <tr>
+              <td>
+                  <img src="https://s3.amazonaws.com/odl-micromasters-production/avatar_default.png"
```
noisecapella: Please use `{% static 'images/avatar_default.png' %}` instead of linking to AWS directly

financialaid/templates/review_financial_aid.html
```diff
@@ -0,0 +1,112 @@
...
+            </option>
+          {% endfor %}
+          </select>​
+        </div>
+        <!-- /Filters -->
+        <!-- Pagination buttons -->
+        <div>
+          <span class="step-links">
+          {% if page_obj.has_previous %}
+            <a href="?page={{ page_obj.previous_page_number }}" class="btn btn-default"><</a>
```
noisecapella: Fix the duplicate `<<` here please

financialaid/templates/review_financial_aid.html
```diff
@@ -0,0 +1,112 @@
...
+          {% else %}
+            <button class="btn btn-default disabled"><</button>
```
noisecapella: Fix the duplicate `<<` here

noisecapella:
When sorting by adjusted_cost I got this exception: `Cannot resolve keyword 'adjusted_price' into field`

screenshot from 2016-09-21 09-58-32

courses/models.py
```diff
@@ -125,3 +116,26 @@ def is_upgradable(self):
...
+            # there is only the start date
+            (models.Q(enrollment_end=None) &
+             models.Q(end_date=None) & models.Q(start_date__lte=now)) |
+            # there is no enrollment date but the course has not ended yet
+            (models.Q(enrollment_end=None) & models.Q(end_date__gte=now)) |
+            # the enrollment end date is simply greater than now
+            models.Q(enrollment_end__gte=now)
+        )
+
+    def get_course_price(self):
```
alvinsiu: Yup, removed.

financialaid/models.py
alvinsiu: Ah, I got some things mixed up - this update function is a helper function for saving several field changes via .update() rather than having to explicitly set attributes and save.  It's not currently used so I can take it out.

Updating the updated_on timestamp via QuerySet is implemented by subclassing the Manager and QuerySet.  Given that these models are ones that require an auditing trail, unless you disagree, I'll go ahead and add that in as a safety measure for future development.

financialaid/views.py
alvinsiu: Changed

financialaid/views.py
alvinsiu: Yup

financialaid/views.py
alvinsiu: Good idea

financialaid/templates/review_financial_aid.html
alvinsiu: Yup

financialaid/templates/review_financial_aid.html
alvinsiu: That's an intentional < for the arrow on the button - I can change it to an html code instead?

financialaid/templates/review_financial_aid.html
alvinsiu: Same as above

financialaid/views.py
```diff
@@ -1,24 +1,217 @@
...
         """
         return None
+
+
+class ReviewFinancialAidView(UserPassesTestMixin, ListView):
+    """
+    View for reviewing financial aid requests.
+    Note: In the future, it may be worth factoring out the code for sorting into its own subclass of ListView
+    """
+    paginate_by = 10
```
noisecapella: Can you change this to 50 per @roberthouse54's request?

alvinsiu: Ah, I didn't see this request.  Done

financialaid/templates/review_financial_aid.html
noisecapella: I see, yes change it to an HTML entity please

financialaid/models.py
noisecapella: Ok, take out the `update` method for now, feel free to leave the abstract base class for timestamps

financialaid/views.py
noisecapella: It's just an informal request he made to me a few minutes ago

noisecapella:
When the user goes to the next page all sorting is removed. Can we preserve the other query parameters when going back and forth between pages?

alvinsiu:
Changes requested have been updated, thanks!

alvinsiu:
Note: financialaid/views.py line 95 (# pylint: disable=unsubscriptable-object) is necessary because of a bug in django pylint.  https://github.com/landscapeio/pylint-django/pull/67

financialaid/views.py
```diff
@@ -48,7 +48,7 @@ class ReviewFinancialAidView(UserPassesTestMixin, ListView):
     View for reviewing financial aid requests.
     Note: In the future, it may be worth factoring out the code for sorting into its own subclass of ListView
     """
-    paginate_by = 10
+    paginate_by = 2
```
noisecapella: This should be `50`

financialaid/templates/review_financial_aid.html
```diff
@@ -0,0 +1,116 @@
...
+        <!-- /Filters -->
+        <!-- Pagination buttons -->
+        <div>
+          <span class="step-links">
+          {% if page_obj.has_previous %}
+            <a href="?page={{ page_obj.previous_page_number }}&sort_by={{ current_sort_field }}" class="btn btn-default">
+              &lt;
+            </a>
+          {% else %}
+            <button class="btn btn-default disabled"><</button>
```
noisecapella: Please change `>` to an html entity, here and below

noisecapella:
Functionality looks great! I still need to go over the tests to make sure everything is covered but otherwise the code is in good shape, except the two comments I just made

financialaid/views_test.py
```diff
@@ -64,31 +120,130 @@ def test_income_validation_missing_args(self):
...
+    def test_review_financial_aid_view_allowed(self):
+        """
+        Tests ReviewFinancialAidView that are allowed
+        """
+        # Allowed for staff of program
+        self.client.force_login(self.staff_user_profile.user)
+        resp = self.client.get(self.review_url)
+        assert resp.status_code == status.HTTP_200_OK
+
+    def test_review_financial_aid_view_with_filter_and_sorting(self):
```
noisecapella: Can we test that the sorting and filtering is reflected in the queryset? The queryset output looks to be available in `resp.context_data['object_list']`

financialaid/views.py
alvinsiu: Ah, I changed it to 50, then to 2 to test sorting with pagination, then forgot to change it back.  Thanks!

financialaid/templates/review_financial_aid.html
alvinsiu: Ah missed these two, thanks

financialaid/views_test.py
alvinsiu: I've added those tests

financialaid/views_test.py
```diff
@@ -211,18 +212,39 @@ def test_review_financial_aid_view_with_filter_and_sorting(self):
...
         self.client.force_login(self.staff_user_profile.user)
         # Should work a filter
         resp = self.client.get(self.review_url_with_filter)
         assert resp.status_code == status.HTTP_200_OK
+        resp_obj_id_list = resp.context_data["financial_aid_objects"].values_list("id", flat=True)
+        expected_obj_id_list = FinancialAid.objects.filter(
+            tier_program__program_id=self.program.id,
+            status=FinancialAidStatus.AUTO_APPROVED
+        ).order_by("user__profile__first_name").values_list("id", flat=True)  # Default sort field
+        self.assertListEqual(list(resp_obj_id_list), list(expected_obj_id_list))
```
noisecapella: `assert` should work here too but feel free to leave it

noisecapella:
Looks good :+1:

PR #1074, opened 2016-09-20
Standardize on dash separators for Sass

singingwolfboy:
The Sass community seems to have standardized on dash-separators instead of underscore_separators. It seems odd to have MicroMasters use underscore separators when the rest of the community, including Sass in npm packages that we use, use dash separators. This pull request switches all our Sass to use dash separators.

See also: http://stackoverflow.com/questions/7560813/why-are-dashes-preferred-for-css-selectors-html-attributes


static/scss/layout.scss
```diff
@@ -1,27 +1,27 @@
...
+@import "welcome-banner";
+@import "profile-progress-controls";
+@import "progress-widget";
+@import "bound-check-box";
+@import "terms-of-service-dialog";
+@import "user-chip";
+@import "email-composition-dialog";
+@import "terms-of-service";
+@import "user-page";
+@import "responsive-styles";
```
aliceriot: it looks like this was accidentally renamed to `response-styles.scss`

singingwolfboy: Done!

PR #1079, opened 2016-09-20
Faculty carousel

### LINKED ISSUE #923: Program Pages: Faculty profile widget
Update the program page template to include a carousel of faculty profiles

image
[InVisionapp mockup](https://projects.invisionapp.com/share/NB7VM1GSG#/screens/182565886)
# Depends on:

- #924 : Allow program owners to add/update faculty profiles

# Acceptance criteria 

- [ ] Display faculty name/title, bio and image
- [ ] no autoplay
- [ ] test and plan for at least 2 faculty members

singingwolfboy:
Fixes #923. Pulls faculty information from database and renders it in Django Templates. Then that information is parsed and passed to [react-slick](https://github.com/akiran/react-slick) for a nice image carousel effect.


cms/templates/cms/program_page.html
```diff
@@ -104,11 +104,37 @@
     </div>
   </div>
 
+  {# if page.faculty_members.count > 0 #}
```
noisecapella: Just curious, what's the difference between `{%` and `{#` here?

static/js/faculty_carousel.js
```diff
@@ -0,0 +1,64 @@
...
+
+class FacultyTile extends React.Component {
+  render() {
+    let nameStr;
+    if (this.props.title) {
+      nameStr = this.props.name + ", " + this.props.title
+    } else {
+      nameStr = this.props.name
+    }
+    let style = {'backgroundImage': 'url('+this.props.imageUrl+')'};
```
noisecapella: Do we need to escape `imageUrl` here?

noisecapella: Nevermind, not ready for review :)

singingwolfboy: I _believe_ that React will take care of the escaping here, but I'm not sure. Can you give me an example of a string that will cause an error, if not properly handled/escaped?

webpack.config.shared.js
```diff
@@ -25,8 +26,7 @@ module.exports = {
         loaders: ['react-hot', 'babel-loader?' + JSON.stringify(babelQuerySettings)],
       },  // to transform JSX into JS
       {
-        test: /\.(svg|ttf|woff|woff2|eot)$/,
-        exclude: /node_modules/,
```
singingwolfboy: I needed to remove this `exclude` line to get the Slick theme to load properly with webpack. I talked about this with @aliceriot, and we both think that _all_ of these `exclude:` lines are unnecessary. Should we maybe remove them in a future pull request?

cms/templates/cms/program_page.html
singingwolfboy: `{#` is a comment in Django templates -- this logic was commented-out for testing. Now that the PR is ready for review, it is no longer commented out.

static/js/faculty_carousel.js
noisecapella: If the url had a `)` in it, you might get `url())`. I think we can just assume the URL is encoded properly before we get it in this component.

static/js/faculty_carousel.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
```
noisecapella: Can you add `// @flow` as the first line of this file to enable flow on it?

cms/templates/cms/program_page.html
```diff
@@ -104,11 +104,37 @@
     </div>
   </div>
 
+  {% if page.faculty_members.count > 0 %}
+    <div class="faculty-section">
+      <h3 class="title">Faculty</h3>
+      {% if page.faculty_description|length > 0 %}
+        <p class="description">{{ page.faculty_description }}</p>
+      {% endif %}
+      <div id="faculty-carousel" class="single-column-wide">
```
noisecapella: A future version of `ReactDOM.render` may leave the HTML below in place so the behavior of replacing the contents of the container is not guaranteed. See the red box: https://facebook.github.io/react/docs/top-level-api.html#reactdom.render

Instead can we populate a JS array with the faculty data? This would also make `FacultyCarousel` easier to test

webpack.config.shared.js
noisecapella: Sounds good with me

static/js/faculty_carousel.js
```diff
@@ -0,0 +1,89 @@
...
+class FacultyTile extends React.Component {
+  props: {
+    name:     string,
+    title:    string,
+    imageUrl: string,
+    bio:      string,
+  }
+  render() {
+    let nameStr;
+    if (this.props.title) {
```
noisecapella: Can you use destructuring here: `const { name, title, bio } = this.props;` for example

static/js/faculty_carousel.js
```diff
@@ -0,0 +1,89 @@
...
+        <p>{this.props.bio}</p>
+      </div>
+    );
+  }
+}
+
+
+class FacultyCarousel extends React.Component {
+  props: {
+    faculty:  array,
```
noisecapella: Please change `array` to `Array<Object>`

static/js/faculty_carousel.js
```diff
@@ -0,0 +1,89 @@
...
+      <div className="faculty-tile" style={style}>
+        <h4>{nameStr}</h4>
+        <p>{this.props.bio}</p>
+      </div>
+    );
+  }
+}
+
+
+class FacultyCarousel extends React.Component {
```
noisecapella: Can we move this class into a separate file in `static/js/components`?

static/js/faculty_carousel.js
```diff
@@ -0,0 +1,89 @@
...
+      <Slider {...settings}>
+        {tiles}
+      </Slider>
+    );
+  }
+}
+
+
+// define a generic `map` function, that can be used on a NodeList
+let map = (...fnargs) => {
```
noisecapella: Why not `Array.from(nodelist).map`?

noisecapella:
Is there a way to get the dots beneath the carousel to show up?

noisecapella:
Can you change the background color of the carousel to match the mockups? My color picker says it's `#0c0c0c`

noisecapella:
With two items in the carousel I'm seeing some weird behavior. If I swipe left twice then there is nothing being displayed anymore. If I then swipe left again it goes back to the first item. Can you investigate this?

static/js/faculty_carousel.js
singingwolfboy: Done

static/js/faculty_carousel.js
singingwolfboy: Done

static/js/faculty_carousel.js
singingwolfboy: Done

static/js/faculty_carousel.js
singingwolfboy: Done

static/js/faculty_carousel.js
singingwolfboy: Switched to `Array.from(nodelist).map`

singingwolfboy:
Alright, I've switched over from passing the data via HTML to passing the data via JSON. However, there are a few unsolved problems here:

* The JSON only provides the URL to the full-size image, so if the image is HUGE, the browser will have to download the whole thing and then scale it down. I would like to make the image serializer display multiple URLs for multiple sizes, but I can't figure out the API that Wagtail uses to generate those URLs.
* I switched from a CSS `background-image` attribute to an HTML `` tag, which is better for accessibility since the image contains useful information rather than just being meaningless decoration. However, this `` tag needs to have an `alt` attribute to be accessible. We'll need to add a field in the database for this, and allow CMS users to define it.

Nonetheless, these seem like things that can be cleaned up in the future.

noisecapella:
Sounds good, let me know when you want me to take another look

cms/models.py
```diff
@@ -26,6 +26,12 @@ def programs_for_sign_up(programs):
     return [ProgramSerializer().to_representation(p) for p in programs]
 
 
+def faculty_for_carousel(faculty):
+    """formats faculty info for the carousel"""
+    from cms.serializers import FacultySerializer
```
noisecapella: Can you move this import to the top of the file?

cms/serializers.py
```diff
@@ -0,0 +1,28 @@
+from rest_framework import serializers
+from wagtail.wagtailimages.models import Image
+from cms.models import ProgramPage, ProgramFaculty
+
+
+class ImageSerializer(serializers.ModelSerializer):
+    "Serializer for Wagtail Image objects."
```
noisecapella: Can you change the docstrings to triple `"` here and below?

cms/serializers_test.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+
+    def test_faculty_serializer(self):  # pylint: disable=no-self-use
+        """
+        Make sure program id appears correctly
+        """
+        faculty = FacultyFactory.create()
+        result = FacultySerializer().to_representation(faculty)
+        assert result['name'] == faculty.name
+        assert result['short_bio'] == faculty.short_bio
```
noisecapella: Can you assert `image` and `title` too?

static/js/components/FacultyCarousel.js
```diff
@@ -0,0 +1,69 @@
+// @flow
+import React from 'react';
+import Slider from 'react-slick';
+
+
+class FacultyTile extends React.Component {
```
noisecapella: Can you move this to its own file too?

cms/models.py
singingwolfboy: Unfortunately, no. That would cause a circular import problem. (`serializers.py` importing `models.py` importing `serializers.py`...)

noisecapella: Sounds good

cms/serializers.py
singingwolfboy: Done

cms/serializers_test.py
singingwolfboy: Done

static/js/components/FacultyCarousel.js
singingwolfboy: Done

noisecapella:
Can you update the help text of `ProgramFaculty.image` to specify a minimum pixel resolution for people to upload? Wagtail will not upscale images, only downscale or crop, so they can't upload an image that's too small.

After that feel free to merge if tests pass :+1:

PR #1080, opened 2016-09-20
Display courses with enrollment status on ProgramPage

### LINKED ISSUE #922: Program Pages: Courses widget
programpagecourses

image

Acceptance criteria:

- [ ] Style should match the mockup
- [ ] Data in this widget is not editable via CMS. They should come from MM app. 
- [ ] don't display prices for now

Related issues:

- Update course and course_run data on a regular schedule (#679)

annagav:
Fix #922 
#### What's this PR do?
Update the courses widget on program page to display start/end date, and enrollment information.
Create a program with few courses. For each course create a course run with the following possible parameters:
1. start date in the future, enrollment start date in the future
2. start date in the future, enrollment start date has past, enrollment end date in the future
etc.

screen shot 2016-09-20 at 5 37 38 pm

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -1,0 +1,37 @@
+
+"""
+apis for program page
+"""
+
+
+def get_course_enrollment_text(course):
+    """
+    Return text that contains start and enrollment
+    information about the course.
+    """
+    course_run = course.get_next_run()
+    text = 'Not available'
+
+    if course_run:
+        if course_run.is_current:
+            text = 'Ongoing - '
+            if course_run.enrollment_end:
+                text += 'Enrollment Ends {:%D}'.format(
+                    course_run.enrollment_end
+                )
+            else:
+                text += 'Enrollment Open'
+        elif course_run.is_future:
+            text = 'Starts {:%D}'.format(course_run.start_date)
+            if course_run.is_future_enrollment_open:
+                text += ' - Enrollment Open'
+            elif course_run.enrollment_start:
+                text += ' - Enrollment {:%m/%Y}'.format(
+                    course_run.enrollment_start
+                )
+    else:
+        course_run = course.get_promised_run()
+        if course_run:
+            text = 'Coming ' + course_run.fuzzy_start_date
+
+    return text
--- a/cms/api_test.py
+++ b/cms/api_test.py
@@ -1,0 +1,116 @@
+"""
+Tests for the program page api functions
+"""
+from datetime import datetime, timedelta
+import pytz
+
+from search.base import ESTestCase
+from courses.factories import (
+    CourseFactory,
+    CourseRunFactory,
+)
+from cms.api import get_course_enrollment_text
+
+
+# pylint: disable=no-self-use
+class CourseEnrollmentInfoTest(ESTestCase):
+    """Test for get_course_enrollment_text"""
+
+    def setUp(self):
+        super(CourseEnrollmentInfoTest, self).setUp()
+        self.now = datetime.now(pytz.utc)
+
+    def test_future_course_enr_future(self):
+        """Test Course that starts in the future, enrollment in future"""
+        course = CourseFactory.create(title="Course in the future")
+        future_date = self.now + timedelta(weeks=1)
+        start_date = 'Starts {:%D} - '.format(future_date)
+        enr_start = 'Enrollment {:%m/%Y}'.format(future_date)
+        # create a run that starts soon, enrollment_start in the future
+        CourseRunFactory.create(
+            course=course,
+            start_date=future_date,
+            end_date=self.now + timedelta(weeks=10),
+            enrollment_start=future_date,
+            enrollment_end=self.now + timedelta(weeks=2),
+        )
+        assert get_course_enrollment_text(course) == start_date + enr_start
+
+    def test_future_course_enr_open(self):
+        """Test course in the future, enrollment open"""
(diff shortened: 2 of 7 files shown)


cms/models.py
```diff
@@ -153,10 +153,40 @@ def get_context(self, request):
         context["username"] = username
         context["js_settings_json"] = json.dumps(js_settings)
         context["title"] = self.title
+        context["courses_info"] = get_course_run_enrollment_state(self)
 
         return context
 
 
+def get_course_run_enrollment_state(program):
```
noisecapella: Since there's no decision being made using information from the program, just the course, can you change this function to accept a course instead of a program? Then you can construct the `state` dict in `get_context` instead

noisecapella:
It looks like this still needs tests for `get_course_run_enrollment_state` and the view. I can show you how to patch the function in the view to assert it was called

cms/models.py
noisecapella: Also can you rename to `get_course_enrollment_text`?

ui/views_test.py
```diff
@@ -361,6 +361,20 @@ def test_program_thumbnail(self):
...
+        course = CourseFactory(program=self.program_page.program)
+        course.save()
+        with patch('cms.models.get_course_enrollment_text') as get_enrollment_text:
+
+            self.client.get(self.program_page.url)
+            for course in self.program_page.program.course_set.all():
+                get_enrollment_text.assert_called_with(course)
+
+            get_enrollment_text.return_value = 'some text'
+            response = self.client.get(self.program_page.url)
```
noisecapella: We don't need this second `get` since we have the one above. Can you use the response coming from that `get` instead?

ui/views_test.py
```diff
@@ -361,6 +361,20 @@ def test_program_thumbnail(self):
...
         self.assertContains(resp, image.get_rendition('fill-345x265').url)
 
+    def test_courses_enrollment_text(self):
+        """Verify get_course_enrollment_test is called with each course in the program"""
+        course = CourseFactory(program=self.program_page.program)
+        course.save()
+        with patch('cms.models.get_course_enrollment_text') as get_enrollment_text:
+
+            self.client.get(self.program_page.url)
+            for course in self.program_page.program.course_set.all():
```
noisecapella: Can we assert the order of the courses here using [call_args_list](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args_list)? We should also order it by `position_in_program` here and in `cms/models.py`

cms/models.py
```diff
@@ -146,7 +146,9 @@ def get_context(self, request):
         context = super(ProgramPage, self).get_context(request)
 
         courses_info = {}
-        for course in self.program.course_set.all():
+        for course in self.program.course_set.all().order_by(
```
noisecapella: Python won't store the order of items in the dictionary (at least, it's not guaranteed that this will be true in the future). Can you change `courses_info` to be an `OrderedDict` above?

noisecapella:
Looks great! This just needs unit tests for `get_course_enrollment_text`

courses/models.py
```diff
@@ -63,6 +63,18 @@ def get_next_run(self):
...
 
+    def get_promised_run(self):
+        """
+        Get a run that does not have start_date,
+        only fuzzy_start_date
+        """
+        run_set = self.courserun_set.filter(
+            models.Q(start_date=None) & models.Q(fuzzy_start_date__isnull=False)
+        )
+        if not run_set.count():
```
noisecapella: Can you change this to `return run_set.first()`?

cms/api.py
```diff
@@ -28,8 +28,9 @@ def get_course_enrollment_text(course):
                 text += ' - Enrollment {:%m/%Y}'.format(
                     course_run.enrollment_start
                 )
-        elif course_run.fuzzy_start_date:
-            text = 'Coming ' + course_run.fuzzy_start_date
+    else:
+        course_run = course.get_promised_run()
```
noisecapella: This is not a guarantee, we need to handle the case where the return value is `None` here

annagav:
@noisecapella, All comments should be addressed now.

courses/models_test.py
```diff
@@ -343,6 +343,49 @@ def test_is_future(self):
...
+        course_run = self.create_run()
+        assert course_run.is_future_enrollment_open is False
+
+        # with start in the future, no enrollment_start
+        course_run = self.create_run(
+            start=self.now + timedelta(weeks=2)
+        )
+        assert course_run.is_future_enrollment_open is False
+
+        # enrollment_start in the past, no enrollment_end
```
noisecapella: Can you break this into separate test functions?

noisecapella: Nevermind, we should be doing separate functions but this way it's consistent with the rest of the file

cms/api_test.py
```diff
@@ -0,0 +1,104 @@
...
+from courses.factories import (
+    CourseFactory,
+    CourseRunFactory,
+)
+from cms.api import get_course_enrollment_text
+
+
+# pylint: disable=no-self-use
+class CourseEnrollmentInfoTest(ESTestCase):
+    """Test for get_course_enrollment_text"""
```
noisecapella: Can you add a test that causes the `Not available` text to occur?

noisecapella:
Just one more comment!

noisecapella:
Good to merge once tests pass :+1:

PR #1092, opened 2016-09-22
Implemented enrollment after course purchase

### LINKED ISSUE #918: Update order fulfillment API to create audit course enrollment on edX
The existing PR #913 will leave a blank space for this functionality

noisecapella:
Fixes #918 

#### What's this PR do?
After a successful course purchase, the user is enrolled in the audit track for that course on edX. We also update `CachedEnrollment` locally so the user sees the new state in the dashboard.

Talk to me, we will need to use the PR build here. Also we should test and merge #1085 first so that we can verify that the dashboard button updates are immediate.

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -336,6 +336,38 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
     return formatted_run
 
 
+def update_cached_enrollment(user, enrollment, course_id, now):
+    """
+    Updates the cached enrollment based on an Enrollment object
+
+    Args:
+        user (User): A user
+        enrollment (Enrollment):
+            An Enrollment object from edx_api_client
+        course_id (str): A course key
+        now (datetime.datetime): The datetime value used as now
+
+    Returns:
+        None
+    """
+    # get the certificate data or None
+    # None means we will cache the fact that the student
+    # does not have an enrollment for the given course
+    enrollment_data = enrollment.json if enrollment is not None else None
+    course_run = CourseRun.objects.get(edx_course_key=course_id)
+    updated_values = {
+        'user': user,
+        'course_run': course_run,
+        'data': enrollment_data,
+        'last_request': now,
+    }
+    models.CachedEnrollment.objects.update_or_create(
+        user=user,
+        course_run=course_run,
+        defaults=updated_values
+    )
+
+
 def _check_if_refresh(user, cached_model, refresh_delta):
     """
     Helper function to check if cached data in a model need to be refreshed.
@@ -398,22 +430,7 @@ def get_student_enrollments(user, edx_client):
     with transaction.atomic():
         for course_id in course_ids:
             enrollment = enrollments.get_enrollment_for_course(course_id)
-            # get the certificate data or None
-            # None means we will cache the fact that the student
-            # does not have an enrollment for the given course
-            enrollment_data = enrollment.json if enrollment is not None else None
-            course_run = CourseRun.objects.get(edx_course_key=course_id)
-            updated_values = {
-                'user': user,
-                'course_run': course_run,
-                'data': enrollment_data,
-                'last_request': now,
-            }
-            models.CachedEnrollment.objects.update_or_create(
-                user=user,
-                course_run=course_run,
-                defaults=updated_values
-            )
+            update_cached_enrollment(user, enrollment, course_id, now)
 
     return enrollments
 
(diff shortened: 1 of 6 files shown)


dashboard/api.py
```diff
@@ -331,6 +331,38 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
     return formatted_run
 
 
+def update_cached_enrollment(user, enrollment, course_id, now):
```
giocalitri: do you think it is worth decorate with a `transaction.atomic`? I mean, if it is used by itself I still would like to change the DB atomically

noisecapella: Will do

ecommerce/api.py
```diff
@@ -204,3 +209,38 @@ def get_new_order_by_reference_number(reference_number):
...
+
+    Returns:
+         None
+    """
+    user_social = order.user.social_auth.get(provider=EdxOrgOAuth2.name)
+    enrollments_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL).enrollments
+
+    exceptions = []
+    enrollments = {}
+    for line in order.line_set.all():
```
giocalitri: question: is this because I could buy multiple courses?

noisecapella: Theoretically an order can contain multiple course runs so I thought it was best to handle the case. But yes, users should not be able to purchase more than one run at a time

ecommerce/api.py
```diff
@@ -204,3 +209,38 @@ def get_new_order_by_reference_number(reference_number):
...
+        course_key = line.course_key
+        try:
+            enrollments[course_key] = enrollments_client.create_audit_student_enrollment(course_key)
+        except Exception as ex:  # pylint: disable=broad-except
```
giocalitri: theoretically all the exceptions raised here should be subclass of some basic exception from the `requests` library, but I am fine with this if you don't want to change it

ecommerce/api.py
```diff
@@ -204,3 +209,38 @@ def get_new_order_by_reference_number(reference_number):
...
+    """
+    user_social = order.user.social_auth.get(provider=EdxOrgOAuth2.name)
+    enrollments_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL).enrollments
+
+    exceptions = []
+    enrollments = {}
+    for line in order.line_set.all():
+        course_key = line.course_key
+        try:
+            enrollments[course_key] = enrollments_client.create_audit_student_enrollment(course_key)
```
giocalitri: you could use the enrollment object (that has a property `course_id`) directly here and just have a list (but still feel free to leave as is)

ecommerce/api.py
```diff
@@ -204,3 +209,38 @@ def get_new_order_by_reference_number(reference_number):
...
+                course_key,
+                get_social_username(order.user),
+            )
+            exceptions.append(ex)
+
+    now = datetime.now(pytz.UTC)
+    for course_key, enrollment in enrollments.items():
+        update_cached_enrollment(order.user, enrollment, course_key, now)
+
+    if len(exceptions) > 0:
```
giocalitri: nit: this can be just 
`if exceptions:`

dashboard/api.py
```diff
@@ -331,6 +331,39 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
...
+        user (User): A user
+        enrollment (Enrollment):
+            An Enrollment object from edx_api_client
+        course_id (str): A course key
+        now (datetime.datetime): The datetime value used as now
+
+    Returns:
+        None
+    """
+    # get the certificate data or None
```
giocalitri: enrollment?

ecommerce/api.py
```diff
@@ -218,3 +223,38 @@ def get_new_order_by_reference_number(reference_number):
         return Order.objects.get(id=order_id, status=Order.CREATED)
     except Order.DoesNotExist:
         raise EcommerceException("Order {} is expected to have status 'created'".format(order_id))
+
+
+def enroll_user(order):
```
giocalitri: nit: maybe the name should be modified to something that implies that this is given a successful order

giocalitri:
mostly minor stuff

ecommerce/api.py
noisecapella: You're right, I'd like to leave it in case the function implementation changes in the future and they forget to update this part

noisecapella:
Comments addressed

giocalitri:
👍

Issue #1077, opened 2016-09-20
Abstract "MicroMasters" track to include financial aid case


pdpinch:
A user is enrolled in a MicroMasters track for a course if one of the two following conditions are met:
1) for programs without financial aid, they have an enrolled in a verified track
2) for programs with financial aid, they have enrolled in an audit track and they have paid for the course

Learner dashboard statuses should be updated so that they depend on this abstract track enrollment, instead of depending directly on the verified track enrollment.

pdpinch:
@giocalitri does this still make sense?

giocalitri:
Wasn't this what we discussed to do?


On Sep 22, 2016, at 5:08 PM, Peter Pinch <notifications@github.com<mailto:notifications@github.com>> wrote:


@giocalitri<https://github.com/giocalitri> does this still make sense?

pdpinch:
It's my interpretation of what we discussed, but I wanted to check with you. Sounds like a "yes"

PR #1094, opened 2016-09-22
Added support for current grades fetching and caching

giocalitri:
this is probably part of #1077

#### What's this PR do?
Adds support for current grades. It also takes care of properly cache the data and retrieve it during the call to the dashboard rest API

`dashboard/api.py`

check if the CachedCurrentGrade model will be updated with some data

The logic implemented here is the same used for Enrollments and Certificates. The cache for the grades expires after 1 hour, but this is open for discussion.


dashboard/api.py
```diff
@@ -24,6 +28,7 @@
 # pylint: disable=too-many-branches
 
 REFRESH_CERT_CACHE_HOURS = 6
+REFRESH_GRADES_CACHE_HOURS = 1
```
giocalitri: @pdpinch : what's a good value for refreshing the gates cache?

dashboard/views.py
```diff
@@ -53,10 +54,12 @@ def get(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-s
...
         # create an instance of the client to query edX
         edx_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL)
-        # get an enrollments client for the student
+        # get enrollments for the student
         enrollments = get_student_enrollments(request.user, edx_client)
-        # get a certificates client for the student
+        # get certificates for the student
         certificates = get_student_certificates(request.user, edx_client)
+        # get current_grades for the student
+        current_grades = get_student_current_grades(request.user, edx_client)  # pylint: disable=unused-variable
```
giocalitri: this has been put here on purpose: right now the data is not used, but it will be as soon as we work on the course statuses

dashboard/api.py
```diff
@@ -331,6 +336,32 @@ def format_courserun_for_dashboard(course_run, status_for_user, certificate=None
     return formatted_run
 
 
+def _check_if_refresh(user, cached_model, refresh_delta):
+    """
+    Helper function to check if cached data in a model need to be refreshed.
+    Args:
+        user (django.contrib.auth.models.User): A user
+        cached_model (dashboard.models.CachedEdxInfoModel): a model containing cached data
+        refresh_delta (datetime.datetime): time limit for refresh the data
```
noisecapella: I think this should be a `timedelta`

noisecapella: Nevermind, I see that it's a `datetime`. Can we pass in a `timedelta` instead of having the caller calculate `now - delta`?

dashboard/fixtures/current_grades.json
```diff
@@ -0,0 +1,15 @@
+[
+	{
+		"course_key": "course-v1:edX+DemoX+Demo_Course",
+		"passed": true, "percent": 0.97,
+		"letter_grade": "Pass",
```
noisecapella: Are `Pass` and `null` the only two values that this can have?

noisecapella: Nevermind, it's irrelevant for us. I skipped over the `passed` key

noisecapella:
Code and functionality look good, looking over tests now...

noisecapella:
Looks good to me :+1: feel free to skip my comment on the timedelta in the interest of moving things along

dashboard/api.py
giocalitri: The value needed is datetime because `last_request__gt=refresh_delta,` needs a datetime and not a timedelta

dashboard/fixtures/current_grades.json
giocalitri: to answer anyway your first question: no, they are not, the letter grade is actually defined in the course grading policy (by the course creators) and could be basically anything.

PR #1096, opened 2016-09-22
Financial Aid: Implemented review page backend actions

### LINKED ISSUE #1049: Financial Aid: Review Actions
# Acceptance Criteria

- [ ] API endpoint to mark documents received
  - [ ] when documents are received, an e-mail is sent to the learner
- [ ] API endpoint to approve a request (at a particular tier)
  - [ ] when a request is approved, an e-mail is sent to the learner
- [ ] API endpoint to reject a request
  - [ ] when a request is rejected, an e-mail is sent to the learner
  - [ ] On rejection, change tier to full price
- [ ] API endpoint to email an applicant
- [ ] Limit API endpoints to staff in the appropriate program

alvinsiu:
Fixes #1049 

#### What's this PR do?
Implements API endpoints for reviewing financial aid applications (approve, reject, mark documents as received).  Also implemented API endpoint for emailing an applicant relating to financial aid.


noisecapella:
If this is meant to close #1049 please change the text to `Fixes #1049`

financialaid/api.py
```diff
@@ -11,7 +12,7 @@ def determine_tier_program(program, income):
         program (Program): the Program to determine a TierProgram for
         income (numeric): the income of the User
     Returns:
-        TierProgram: the TierProgram for the Program given the User's income
+        (TierProgram): the TierProgram for the Program given the User's income
```
noisecapella: Parentheses are not used in the return value. See here: https://pypi.python.org/pypi/sphinxcontrib-napoleon

financialaid/api.py
```diff
@@ -27,8 +28,19 @@ def determine_auto_approval(financial_aid):
     Args:
         financial_aid (FinancialAid): the financial aid object to determine auto-approval
     Returns:
-        boolean: True if auto-approved, False if not
+        (boolean): True if auto-approved, False if not
```
noisecapella: Please remove parentheses here

financialaid/api.py
```diff
@@ -27,8 +28,19 @@ def determine_auto_approval(financial_aid):
...
     return financial_aid.income_usd > income_threshold or income_threshold == 0
+
+
+def get_no_discount_tier_program(program):
+    """
+    Takes a program_id and returns the no discount TierProgram for that Program
+    Args:
+        program (Program): the id of the Program object
+    Returns:
+        (TierProgram): the no discount TierProgram program associated with the Program
```
noisecapella: Remove parentheses here too please

financialaid/constants.py
```diff
@@ -222,3 +222,19 @@
     "UG": 0,
     "ZW": 0
 }
+
+FINANCIAL_AID_APPROVAL_MESSAGE_BODY = """Your application for financial aid has been approved!
+Please log on to your account for more details.
+Love,
+MIT Micromasters"""
+FINANCIAL_AID_REJECTION_MESSAGE_BODY = """Your application for financial aid has been rejected.
```
noisecapella: @pdpinch is there content recommended for these emails?

financialaid/permissions.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    def has_object_permission(self, request, view, obj):
+        """
+        Returns True if the user has the can_edit_financial_aid permission for a program.
+        Args:
+            request (Request): DRF request object
+            view (View): DRF view object
+            obj (FinancialAid): FinancialAid object
+        Returns:
+            (boolean)
```
noisecapella: Remove parentheses here

mail/api.py
```diff
@@ -109,3 +109,18 @@ def send_batch(cls, subject, body, recipients, chunk_size=settings.MAILGUN_BATCH
...
+    @classmethod
+    def send_individual_email(cls, subject, body, recipient):
+        """
+        Sends a text email to a single recipient.
+        Args:
+            subject (str): email subject
+            body (str): email body
+            recipient (str): email recipient
+        Returns:
+            (requests.Response): response from Mailgun
```
noisecapella: Remove parentheses here

financialaid/constants.py
```diff
@@ -222,3 +222,19 @@
...
+Please log on to your account for more details.
+Sympathy,
+MIT Micromasters"""
+FINANCIAL_AID_DOCUMENTS_MESSAGE_BODY = """Your financial aid documents have been received.
```
pdpinch: We'd like these e-mails to be personalized. Shall I open a separate issue for templating the e-mail?

alvinsiu: That makes sense to me.  These are placeholders since I don't think we've discussed the content of the emails yet.

financialaid/api.py
alvinsiu: Ah, thanks for the reference.  Will update.

pytest.ini
```diff
@@ -3,5 +3,6 @@ addopts = --cov . --pep8 --pylint --cov-report term --cov-report html --ds=micro
 norecursedirs = node_modules .git .tox static templates .* CVS _darcs {arch} *.egg
 pep8ignore =
    */migrations/* ALL
+   */financialaid/*.py W503
```
noisecapella: Please remove this change and update the lint errors in `financialaid/serializers.py` instead

financialaid/views_test.py
```diff
@@ -269,3 +245,167 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+    def test_financial_aid_action_view_with_invalid_data(self, *args):  # pylint: disable=unused-argument
+        """
+        Tests FinancialAidActionView when invalid data is posted
+        """
+        # Invalid action
+        self.financial_review_data["action"] = FinancialAidStatus.PENDING_DOCS
+        self.client.force_login(self.staff_user_profile.user)
+        resp = self.client.post(self.action_url, data=self.financial_review_data)
+        assert resp.status_code == status.HTTP_400_BAD_REQUEST
+        # Invalid tier
```
noisecapella: Please split these up into individual tests. As individual tests it is easier to identify which cases are failing and to prevent one case from impacting another case.

financialaid/views.py
```diff
@@ -218,3 +223,20 @@ def get_queryset(self):
...
+    """
+    View for rejecting and approving financial aid requests
+    """
+    serializer_class = FinancialAidActionSerializer
+    permission_classes = (IsAuthenticated, UserCanEditFinancialAid)
+    lookup_field = "id"
+    lookup_url_kwarg = "financial_aid_id"
+    queryset = FinancialAid.objects.all()
+
+    def post(self, request, *args, **kwargs):
```
noisecapella: Why do we need to define this method? Can't the client just send a PUT?

micromasters/urls.py
```diff
@@ -37,8 +43,11 @@
     url(r'^api/v0/search/(?P<elastic_url>.*)', ElasticProxyView.as_view(), name='search_api'),
     url(r'^api/v0/checkout/$', CheckoutView.as_view(), name='checkout'),
     url(r'^api/v0/enrolledprograms/$', ProgramEnrollmentListView.as_view(), name='user_program_enrollments'),
-    url(r'^api/v0/mail/$', MailView.as_view(), name='mail_api'),
-    url(r'^api/v0/financialaid/$', IncomeValidationView.as_view(), name='financialaid_api'),
+    url(r'^api/v0/search_result_mail/$', SearchResultMailView.as_view(), name='search_result_mail_api'),
```
noisecapella: Please change back to `api/v0/mail`. This PR hasn't updated the JS client for the changed endpoint, and in any case I'd like to not rename things outside the scope of this PR if possible

alvinsiu: Good point, will do - would you recommend renaming the financial aid mail view to `api/v0/mail/financial_aid` or `api/v0/financial_aid_mail`?

financialaid/views.py
alvinsiu: We just put it in for flexibility, but you're right it's not required.  We can take it out if you'd like?

pytest.ini
alvinsiu: W503 references line breaks before/after operators where the linter disagrees with PEP8 (https://github.com/PyCQA/pycodestyle/issues/197) and (https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).  It's kind of a controversial topic IIRC, we're happy to go with whichever style you prefer.  Would you like us to remove this change and update lint errors instead?

micromasters/urls.py
noisecapella: You can feel free to rename the financial aid endpoints, I'm just looking to keep the existing code unchanged to make it easier to review

financialaid/views.py
noisecapella: Yes, please remove it

pytest.ini
noisecapella: Yes, I don't necessarily agree or disagree but I'd like the linter rules to remain unchanged at least for the duration of the financial aid work, to keep the code consistent with everything else

financialaid/serializers.py
```diff
@@ -61,3 +83,75 @@ def save(self):
...
+                and self.instance.status != FinancialAidStatus.PENDING_MANUAL_APPROVAL):
+            raise ValidationError("Cannot approve application that is not pending manual approval.")
+        if (data['action'] == FinancialAidStatus.PENDING_MANUAL_APPROVAL
+                and self.instance.status != FinancialAidStatus.PENDING_DOCS):
+            raise ValidationError("Cannot mark documents as received for application not pending docs.")
+        # Check tier program exists
+        try:
+            data["tier_program"] = TierProgram.objects.get(
+                id=data["tier_program_id"],
+                program=self.instance.tier_program.program_id
```
noisecapella: Can you change `tier_program.program_id` to `tier_program.program` here?

financialaid/constants.py
pdpinch: My comment seems to have been lost but, yes, there is and I'll open another issue for updating it. 

I'd like it to actually be a template, so we can customize the response. See https://goo.gl/moqYQu

financialaid/serializers.py
alvinsiu: Wouldn't this have a slight performance decrease since it'll be hitting the database again to fetch `.program` instead of just using `.program_id` on `tier_program`?

noisecapella: Yes, in that case can you change `program=` to `program_id=` instead?

alvinsiu:
I believe all of the current change requests have been implemented!

financialaid/serializers.py
```diff
@@ -61,3 +83,77 @@ def save(self):
...
+                "email_recipient": self.instance.user.email
+            }
+        elif self.instance.status == FinancialAidStatus.PENDING_MANUAL_APPROVAL:
+            email_data = {
+                "email_subject": FINANCIAL_AID_DOCUMENTS_SUBJECT_TEXT,
+                "email_body": FINANCIAL_AID_DOCUMENTS_MESSAGE_BODY,
+                "email_recipient": self.instance.user.email
+            }
+        else:
+            raise Exception("Error in ")
```
noisecapella: Can you update that message?

financialaid/api.py
```diff
@@ -32,3 +33,14 @@ def determine_auto_approval(financial_aid):
...
+
+def get_no_discount_tier_program(program):
+    """
+    Takes a program_id and returns the no discount TierProgram for that Program
+    Args:
+        program (Program): the id of the Program object
+    Returns:
+        TierProgram: the no discount TierProgram program associated with the Program
+    """
+    return TierProgram.objects.get(program=program, discount_amount=0)
```
noisecapella: It seems like this needs a `current=True` condition too. Also, are we requiring every program to have one and only one zero discount tier?

financialaid/serializers.py
```diff
@@ -61,3 +83,77 @@ def save(self):
...
+        self.instance.save()
+        # Send email notification
+        FinancialAidMailView().post(self.context["request"], **email_data)
```
noisecapella: Why aren't we just using `MailgunClient.send_individual_email` directly?

financialaid/serializers.py
```diff
@@ -61,3 +83,77 @@ def save(self):
...
+        self.instance.status = self.validated_data["action"]
+        if self.instance.status == FinancialAidStatus.APPROVED:
+            self.instance.tier_program = tier_program
+            email_data = {
+                "email_subject": FINANCIAL_AID_APPROVAL_SUBJECT_TEXT,
+                "email_body": FINANCIAL_AID_APPROVAL_MESSAGE_BODY,
+                "email_recipient": self.instance.user.email
+            }
+        elif self.instance.status == FinancialAidStatus.REJECTED:
+            self.instance.tier_program = get_no_discount_tier_program(self.instance.tier_program.program_id)
```
noisecapella: It looks like you're sending `program_id` but the function expects a `Program`. Can you change one or the other to be consistent?

mail/views.py
```diff
@@ -52,3 +54,34 @@ def post(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-
...
+    """
+    authentication_classes = (authentication.SessionAuthentication, )
+    permission_classes = (permissions.IsAuthenticated, UserCanMessageLearnersPermission, UserCanEditFinancialAid)
+
+    def post(self, request, *args, **kwargs):
+        """
+        Post request to send emails to an individual learner
+        """
+        try:
+            subject = request.data.get('email_subject') or kwargs['email_subject']
```
noisecapella: Please only use `request.data` to get information here. We shouldn't use this function except to respond to requests. If there is substantial functionality which can be shared between other functions we should make a separate function to share that functionality.

mail/views_test.py
```diff
@@ -34,7 +34,8 @@ def json(*args, **kwargs):  # pylint:disable=unused-argument, missing-docstring
 class MailViewsTests(APITestCase):
```
noisecapella: Please use a separate test case for the financial aid mail tests

financialaid/serializers.py
alvinsiu: Oops, I forgot to delete this.  This exception will never bit hit because of validation.

financialaid/api.py
alvinsiu: Good catch, I'll add that in.  And yes, that's one of the requirements.

financialaid/serializers.py
alvinsiu: Switched to program_id, thanks!

noisecapella:
Functionality is looking good

mail/views.py
alvinsiu: @gsidebo felt strongly about using the views for routing all mailgun communications in our conversation on HipChat ("lots and lots of reasons to funnel all of our communication with the mailgun API through views like the one thats already there").  So this allows us to call `FinancialAidMailView().post()` from within another view to also send emails.

Gavin, I would love your input here as well.  I'm happy to go either way.

mail/views_test.py
alvinsiu: Sounds good!

financialaid/serializers.py
alvinsiu: Referenced in https://github.com/mitodl/micromasters/pull/1096#discussion_r80290727

financialaid/api.py
noisecapella: That's fine, hopefully we have a list of the `TierProgram` objects that need to be created when a new program is set up

mail/views.py
gsidebo: sorry, i can definitely see how i misled you here. i meant all communication coming from the front end. if there are other parts of the django app that need to send an individual email, you can definitely feel free to invoke `MailgunClient.send_individual_email` directly.

financialaid/api.py
alvinsiu: @pdpinch Is this something that would be helpful for us to write up for you?

mail/views.py
alvinsiu: No worries!  Thanks for clearing things up.  @noisecapella I've marked this as waiting for author, and I'll change it back once we've implemented this change.

financialaid/api.py
```diff
@@ -32,3 +33,14 @@ def determine_auto_approval(financial_aid):
     income_threshold = COUNTRY_INCOME_THRESHOLDS.get(financial_aid.country_of_income, DEFAULT_INCOME_THRESHOLD)
     # The income_threshold == 0 is because in all cases BUT threshold == 0, it's strictly > instead of >=
     return financial_aid.income_usd > income_threshold or income_threshold == 0
+
+
+def get_no_discount_tier_program(program):
+    """
+    Takes a program_id and returns the no discount TierProgram for that Program
+    Args:
+        program (Program): the id of the Program object
```
noisecapella: Can you update the description or the argument here to be consistent?

mail/views.py
gsidebo: interesting thing to note: if that mailgun call fails for whatever reason, the app will keep chugging along as normal. for that reason and for general separation of concerns, i'm wondering if it's a better idea to move the invocation of `send_individual_email` out of the Serializer classes and invoke it after the Serializer successfully saves the instance. come to think of it, that `save()` could fail, in which case i don't think we would want to sent the email

financialaid/views_test.py
```diff
@@ -269,3 +246,277 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+        )
+        assert resp.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_invalid_action(self, *args):  # pylint: disable=unused-argument
+        """
+        Tests FinancialAidActionView when invalid action is posted
+        """
+        # Invalid action
+        self.client.force_login(self.staff_user_profile.user)
+        self.financial_review_data["action"] = FinancialAidStatus.PENDING_DOCS
```
noisecapella: Can you test all the invalid actions here with a for loop?

financialaid/views_test.py
```diff
@@ -269,3 +246,277 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+        self.financialaid.save()
+        self.financial_review_data["action"] = FinancialAidStatus.REJECTED
+        self.financial_review_data["tier_program_id"] = self.tier_programs["15k"].id
+        resp = self.client.put(
+            self.action_url,
+            data=json.dumps(self.financial_review_data),
+            content_type="application/json"
+        )
+        assert resp.status_code == status.HTTP_400_BAD_REQUEST
+        # Reset FinancialAid object
```
noisecapella: If we need to reset `self.financialaid` we should move its creation into `setUp`, or run `self.financialaid.refresh_from_db()` there.

financialaid/views_test.py
```diff
@@ -269,3 +246,277 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+
+    def test_approve_invalid_status(self, *args):  # pylint: disable=unused-argument
+        """
+        Tests FinancialAidActionView when trying to approve a FinancialAid that isn't pending manual approval
+        """
+        # FinancialAid object that cannot be approved
+        self.client.force_login(self.staff_user_profile.user)
+        self.financialaid.status = FinancialAidStatus.PENDING_DOCS
+        self.financialaid.save()
+        self.financial_review_data["action"] = FinancialAidStatus.APPROVED
```
noisecapella: Can we iterate through all statuses that aren't `APPROVED` here?

financialaid/views_test.py
```diff
@@ -269,3 +246,277 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+        self.financialaid.status = FinancialAidStatus.PENDING_MANUAL_APPROVAL
+        self.financialaid.save()
+
+    def test_mark_documents_received_invalid_status(self, *args):  # pylint: disable=unused-argument
+        """
+        Tests FinancialAidActionView when trying to approve a FinancialAid that isn't pending docs
+        """
+        # FinancialAid object whose documents cannot received
+        self.client.force_login(self.staff_user_profile.user)
+        self.financialaid.status = FinancialAidStatus.REJECTED
```
noisecapella: Can we iterate through all non `PENDING_DOCS` states in this test case?

financialaid/api.py
alvinsiu: Will do!

financialaid/views_test.py
alvinsiu: Yup

financialaid/views_test.py
```diff
@@ -269,3 +246,277 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+        )
+        self.client.force_login(self.staff_user_profile.user)
+        # Application is approved for a different tier program
+        self.financial_review_data["tier_program_id"] = self.tier_programs["50k"].id
+        resp = self.client.put(
+            self.action_url,
+            data=json.dumps(self.financial_review_data),
+            content_type="application/json"
+        )
+        from pprint import pprint
```
alvinsiu: Wow, my bad

financialaid/views_test.py
alvinsiu: Moved.  (Also, would you prefer that we respond to all of your comments so you know we're acknowledging it, or would you prefer that we spare your inbox/notification queue?)

alvinsiu: Sounds good - I'll keep replying then unless you let me know not to.  It helps me keep track of what I've changed.

financialaid/serializers.py
gsidebo: (continuing from [this comment thread](https://github.com/mitodl/micromasters/pull/1096/files/208bbaaa39fec22640e5b4f88e73608ef483858b#diff-d43c1becf2eaa76305a72bd2991dcfc4)) sorry for the stream-of-consciousness comments here, but after thinking about this some more, i think we definitely need to move the email sending functionality out of the serializer. all this method is supposed to do is deserialize some data and save a `FinancialAidAction` object. if i want to use or test out that functionality on it's own, i most definitely do not want to send an email or invoke any mail-related code. that's an entirely separate concern. it seems more likely that we would only want to send this email when a certain View method is used (in this case, i guess the view class is `FinancialAidActionView`). i don't know off-hand how to do this using rest_framework code, but we should send the email as part of the appropriate view method (update/post/etc) after the request has been successfully deserialized. @noisecapella do you agree?

noisecapella:
In the interest of moving things along I'm fine with the serializer sending email here

financialaid/views_test.py
alvinsiu: Yup!

financialaid/views_test.py
alvinsiu: Yup!

alvinsiu:
Let's keep it in the serializer for now for this PR, but I'll add a new issue to further discuss this topic.

alvinsiu:
Requested changes updated as per request!

financialaid/views_test.py
```diff
@@ -263,9 +226,213 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+        self.financialaid.save()
+        self.assert_http_status(self.client.put, self.action_url, status.HTTP_400_BAD_REQUEST, data=self.data)
+
+    def test_mark_documents_received_invalid_status(self, *args):  # pylint: disable=unused-argument
+        """
+        Tests FinancialAidActionView when trying to approve a FinancialAid that isn't pending docs
+        """
+        # FinancialAid object whose documents cannot received
+        self.data["action"] = FinancialAidStatus.PENDING_MANUAL_APPROVAL
+        self.financialaid.status = FinancialAidStatus.CREATED
```
noisecapella: Use a for loop here too please

mail/views_test.py
```diff
@@ -34,7 +34,8 @@ def json(*args, **kwargs):  # pylint:disable=unused-argument, missing-docstring
...
     @classmethod
     def setUpTestData(cls):
-        cls.mail_url = reverse('mail_api')
+        cls.search_result_mail_url = reverse('search_result_mail_api')
+        cls.financial_aid_mail_url = reverse('financial_aid_mail_api')
```
noisecapella: Remove this line please since we don't use this value in this test case

financialaid/views_test.py
```diff
@@ -263,9 +226,213 @@ def test_review_financial_aid_view_with_invalid_filter_and_sorting(self):
...
+        self.data["action"] = FinancialAidStatus.REJECTED
+        self.assert_http_status(self.client.put, self.action_url, status.HTTP_400_BAD_REQUEST, data=self.data)
+
+    def test_approve_invalid_status(self, *args):  # pylint: disable=unused-argument
+        """
+        Tests FinancialAidActionView when trying to approve a FinancialAid that isn't pending manual approval
+        """
+        # FinancialAid object that cannot be approved
+        self.data["action"] = FinancialAidStatus.APPROVED
+        self.financialaid.status = FinancialAidStatus.CREATED
```
alvinsiu: Yup!

financialaid/views_test.py
alvinsiu: Yup

mail/views_test.py
alvinsiu: Good catch

noisecapella:
Can you fix the test failures?

financialaid/constants.py
```diff
@@ -222,3 +222,19 @@
...
     "ZW": 0
 }
+
+FINANCIAL_AID_APPROVAL_MESSAGE_BODY = """Your application for financial aid has been approved!
+Please log on to your account for more details.
+Love,
+MIT Micromasters"""
+FINANCIAL_AID_REJECTION_MESSAGE_BODY = """Your application for financial aid has been rejected.
+Please log on to your account for more details.
+Sympathy,
```
noisecapella: Can we remove the `Sympathy` and `Love` here? I'm just looking to guard against the case where we forget to update this in a later issue

alvinsiu: Good idea

noisecapella:
Everything looks good :+1:

alvinsiu:
@gsidebo Are you also good with this PR?

gsidebo:
yup, looks great 👍

PR #1099, opened 2016-09-22
Financial Aid: Implemented endpoint for retrieving a learner's course price

### LINKED ISSUE #1044: Financial Aid: Auto-Approved requests
When a user's financial aid request is auto-approved their dashboard updates with a message and all their prices are updated to reflect their personal price.

image

## Acceptance criteria:

- [x] API endpoint for getting a user's price for a course
- [ ] Update message
- [ ] Update course statuses to display "Pay Now - $$$" button, with appropriate price

alvinsiu:
Fixes #1044 

#### What's this PR do?
Implements the API endpoint for retrieving a learner's course price as well as information about their potential financial aid request

This PR depends on changes made in #1096, so most of the commits/files changed are duplicated in that PR.  I've left this without a "Needs Review" label for now until the other PR is merged.

You can visit api/v0/financial_aid_course_price/<user_id>/<program_id> to use DRFs built-in helper pages to check input and output.


noisecapella:
Please change `#1044` to `Fixes #1044`

noisecapella:
I'm going to take a quick look at this and make some comments, hopefully on the stuff that isn't touched by the other PR this is based on

courses/models.py
```diff
@@ -19,6 +20,24 @@ class Program(models.Model):
...
 
+    def get_course_price(self):
+        """
+        Returns a decimal course price attached to this program.
+
+        Note: This implementation of retrieving a course price is a naive lookup that assumes
+        all course runs and courses will be the same price for the foreseeable future.
+        Therefore we can just take the price from any currently enroll-able course run.
+        """
+        from ecommerce.models import CoursePrice
```
noisecapella: Please move all imports to the top level

courses/models.py
```diff
@@ -19,6 +20,24 @@ class Program(models.Model):
     def __str__(self):
         return self.title
 
+    def get_course_price(self):
+        """
+        Returns a decimal course price attached to this program.
+
+        Note: This implementation of retrieving a course price is a naive lookup that assumes
+        all course runs and courses will be the same price for the foreseeable future.
```
noisecapella: Please change `all course runs and courses will...` to `all course runs in a single program will...`

micromasters/urls.py
```diff
@@ -37,8 +44,13 @@
...
     url(r'^api/v0/checkout/$', CheckoutView.as_view(), name='checkout'),
     url(r'^api/v0/enrolledprograms/$', ProgramEnrollmentListView.as_view(), name='user_program_enrollments'),
-    url(r'^api/v0/mail/$', MailView.as_view(), name='mail_api'),
-    url(r'^api/v0/financialaid/$', IncomeValidationView.as_view(), name='financialaid_api'),
+    url(r'^api/v0/search_result_mail/$', SearchResultMailView.as_view(), name='search_result_mail_api'),
+    url(r'^api/v0/financial_aid_mail/$', FinancialAidMailView.as_view(), name='financial_aid_mail_api'),
+    url(r'^api/v0/financial_aid_request/$', FinancialAidRequestView.as_view(), name='financial_aid_request'),
+    url(r'^api/v0/financial_aid_action/(?P<financial_aid_id>[\d]+)/$', FinancialAidActionView.as_view(),
+        name='financial_aid_action'),
+    url(r'^api/v0/financial_aid_course_price/(?P<user_id>[\d]+)/(?P<program_id>[\d]+)/$',
```
noisecapella: What is the user id used for here? Wouldn't we just need the program id?

courses/models.py
alvinsiu: This would result in a circular import since `ecommerce.models` imports `courses.models`

courses/models.py
alvinsiu: Yup

micromasters/urls.py
alvinsiu: Not sure, but we built it in because it seems likely that there will be a use case for retrieving someone else's course price in the future (say, for auditing or some future staff view).

alvinsiu:
Merged master into this branch so should only include changes relevant to this PR.

courses/models.py
noisecapella: That's fine, let's leave it here then

micromasters/urls.py
noisecapella: I'd prefer to avoid building APIs for future use cases unless they're on an immediate road map. Can you change this API to remove the `user_id` parameter and change the code to use the logged in user for this purpose instead?

alvinsiu: Sure

financialaid/views.py
```diff
@@ -5,25 +5,32 @@
...
+from dashboard.models import ProgramEnrollment
 from financialaid.models import (
     FinancialAid,
     FinancialAidStatus,
     TierProgram
 )
-from financialaid.permissions import UserCanEditFinancialAid
+from financialaid.permissions import (
+    UserCanEditFinancialAid,
+    UserCanViewLearnerCoursePrice
```
noisecapella: Since we're not putting `user_id` into the API we can probably get rid of this permission for now

financialaid/views.py
```diff
@@ -234,3 +227,48 @@ class FinancialAidActionView(UpdateAPIView):
...
+            raise ValidationError("Learner not enrolled in this program.")
+        course_price = program.get_course_price()
+        has_financial_aid_request = False
+        financial_aid_adjustment = False
+        try:
+            # Check to see if learner has a financial aid request
+            financial_aid = FinancialAid.objects.get(
+                user=learner,
+                tier_program__program=program,
+                tier_program__program__financial_aid_availability=True
```
noisecapella: Can we remove this check here and add it above when we get the `program`?

financialaid/views.py
alvinsiu: Agreed

financialaid/views.py
alvinsiu: Good idea

financialaid/views.py
```diff
@@ -234,3 +227,48 @@ class FinancialAidActionView(UpdateAPIView):
...
+            )
+            has_financial_aid_request = True
+            if financial_aid.status == FinancialAidStatus.APPROVED:
+                # If the financial aid request is approved, adjust course price
+                course_price = course_price - financial_aid.tier_program.discount_amount
+                financial_aid_adjustment = True
```
giocalitri: Nit: please wrap in a `try/except` only the call that can raise an exception: in this case `FinancialAid.objects.get`

alvinsiu: That would make it a less clear in terms of readibility.  Doing so would require initializing `financial_aid = None` first in case it doesn't exist and also changing `if financial_aid.status == FinancialAidStatus.APPROVED:` to `if financial_aid is not None and financial_aid.status == FinancialAidStatus.APPROVED`.

I personally think this adds unnecessary bulk, but if you disagree then I'm happy to change it.

giocalitri: I see many other ways around this (i.e. use `.filter()` to get a query set and then use a `if .exists()`), but this is not crucial at the moment: feel free to leave as is.

alvinsiu: That's a good point, I'll change it then.

alvinsiu:
The requested changes have been made!

financialaid/views.py
```diff
@@ -234,3 +223,52 @@ class FinancialAidActionView(UpdateAPIView):
...
+class GetLearnerPriceForCourseView(APIView):
+    """
+    View for retrieving a leaner's price for a course run
+    """
+    authentication_classes = (SessionAuthentication, )
+    permission_classes = (IsAuthenticated, )
+
+    def get(self, request, *args, **kwargs):
+        """
+        Get request for GetLearnerPriceForCourseView
```
noisecapella: Can you create a function `get_price_for_user(user, program)` in `api.py` and move some of the functionality here into that function? I think it would be useful for ecommerce work in the near future.

noisecapella:
Functionality looks good

financialaid/views.py
```diff
@@ -234,3 +223,52 @@ class FinancialAidActionView(UpdateAPIView):
...
+        """
+        learner = request.user
+        program = get_object_or_404(
+            Program,
+            id=self.kwargs["program_id"],
+            live=True,
+            financial_aid_availability=True
```
noisecapella: Can we remove this restraint here? @giocalitri and I were talking and it would be nice to use this functionality for the non financial aid programs too.

financialaid/views.py
alvinsiu: Will do!

financialaid/views.py
alvinsiu: Sounds good!  It will now return an additional field "financial_aid_availability" as per @giocalitri's request.

financialaid/views_test.py
```diff
@@ -498,6 +472,20 @@ def test_get_learner_price_for_course_allowed(self):
         expected_response = {
             "has_financial_aid_request": False,
             "course_price": self.course_price.price,
-            "financial_aid_adjustment": False
+            "financial_aid_adjustment": False,
+            "financial_aid_availability": True
+        }
+        self.assertDictEqual(resp.data, expected_response)
+        # Enrolled but program has no financial aid availability
```
noisecapella: Please split this up into separate tests, one per case being tested

financialaid/views_test.py
```diff
@@ -498,6 +472,20 @@ def test_get_learner_price_for_course_allowed(self):
...
+        self.program.save()
+        resp = self.assert_http_status(self.client.get, self.course_price_url, status.HTTP_200_OK)
+        expected_response = {
+            "has_financial_aid_request": False,
+            "course_price": self.course_price.price,
+            "financial_aid_adjustment": False,
+            "financial_aid_availability": False
         }
         self.assertDictEqual(resp.data, expected_response)
+        self.program.financial_aid_availability = True
```
noisecapella: We don't need to reset this object, `setUpTestData` will do that for us.

financialaid/api_test.py
```diff
@@ -187,3 +214,63 @@ def test_determine_auto_approval(self):  # pylint: disable=no-self-use
...
+            "course_price": self.course_price.price,
+            "financial_aid_adjustment": False,
+            "financial_aid_availability": True
+        }
+        self.assertDictEqual(
+            get_course_price_for_learner(self.enrolled_profile3.user, self.program),
+            expected_response
+        )
+        # Enrolled but program has no financial aid availability
+        self.program.financial_aid_availability = False
```
noisecapella: Please split up these tests into separate cases, one per case being tested (basically, each test should end with an assert being made)

financialaid/api_test.py
```diff
@@ -187,3 +214,63 @@ def test_determine_auto_approval(self):  # pylint: disable=no-self-use
...
+            "has_financial_aid_request": False,
+            "course_price": self.course_price.price,
+            "financial_aid_adjustment": False,
+            "financial_aid_availability": False
+        }
+        self.assertDictEqual(
+            get_course_price_for_learner(self.enrolled_profile3.user, self.program),
+            expected_response
+        )
+        self.program.financial_aid_availability = True
```
noisecapella: We don't need to reset the state after the test ends

financialaid/views_test.py
alvinsiu: I thought `setUpTestData` will only run once per class, where `setUp` runs once per test.  And since a lot of stuff depends on having `self.program`, I'm explicitly keeping that in `setUpTestData`.

financialaid/views_test.py
alvinsiu: Sure thing!

financialaid/api_test.py
alvinsiu: Sounds good!

financialaid/api_test.py
alvinsiu: Related to https://github.com/mitodl/micromasters/pull/1099#discussion_r80805756, if `self.program` is only set up once per class, and someone decides to write a test at a later time after this one, wouldn't it make sense to reset the state?

financialaid/views_test.py
noisecapella: It will run once per class but it runs within a transaction which is rolled back after the test ends. You only need to run `self.program.refresh_from_db()` in `setUp()` if you are making use of the cached value of `self.program`

alvinsiu: Oh, that's neat - then what is the benefit of putting things into `setUp()` instead of `setUpTestData()`?

I'll take out the resets in the meantime.

noisecapella: The benefit is only for particular cases where you need to make changes that can't be reverted in a database transaction, for example `ESTestCase` does some Elasticsearch cleanup.

More information here: https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.TestCase.setUpTestData

alvinsiu: Thanks for the link!

noisecapella:
Looks great :+1:

PR #1104, opened 2016-09-22
Overrode save method on FinancialAid to ensure uniqueness between Use…

### LINKED ISSUE #1091: Financial Aid: add unique constraint on FinancialAid objects for user and program
This constraint will protect against bugs in other parts of the codebase in accessing FinancialAid objects.  After the 10/15 deadline, this may change when we flush out how to handle multiple financial aid requests for the same user and program.

alvinsiu:
Fixes #1091 

#### What's this PR do?
Adds a uniqueness constraint on `FinancialAid `objects between User and Program so that learners cannot have multiple financial aid applications per program.  This is done by overriding `FinancialAid.save`.

After 10/15, we'll most likely be allowing learners to apply multiple times, so this will eventually change, but for now will protect against bugs in other parts of the codebase.

### DIFF
--- a/financialaid/models.py
+++ b/financialaid/models.py
@@ -3,6 +3,7 @@ Models for the Financial Aid App
 """
 import datetime
 from django.contrib.auth.models import User
+from django.core.exceptions import ValidationError
 from django.db import (
     models,
     transaction,
@@ -137,7 +138,7 @@ class FinancialAid(TimestampedModel):
                 user=self.user,
                 tier_program__program=self.tier_program.program
         ).exclude(id=self.id).exists():
-            raise Exception("Cannot have multiple FinancialAid objects for the same User and Program.")
+            raise ValidationError("Cannot have multiple FinancialAid objects for the same User and Program.")
         super().save(*args, **kwargs)
 
 
--- a/financialaid/models.py
+++ b/financialaid/models.py
@@ -134,7 +134,7 @@ class FinancialAid(TimestampedModel):
         """
         Override save to make sure only one FinancialAid object exists for a User and the associated Program
         """
-        if self.__class__.objects.filter(
+        if FinancialAid.objects.filter(
                 user=self.user,
                 tier_program__program=self.tier_program.program
         ).exclude(id=self.id).exists():
--- a/financialaid/models.py
+++ b/financialaid/models.py
@@ -128,6 +128,18 @@ class FinancialAid(TimestampedModel):
     country_of_income = models.CharField(null=True, max_length=100)
     date_exchange_rate = models.DateTimeField(null=True)
 
+    @transaction.atomic
+    def save(self, *args, **kwargs):
+        """
+        Override save to make sure only one FinancialAid object exists for a User and the associated Program
+        """
+        if self.__class__.objects.filter(
+            user=self.user,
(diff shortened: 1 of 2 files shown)


financialaid/models.py
```diff
@@ -128,6 +128,18 @@ class FinancialAid(TimestampedModel):
     country_of_income = models.CharField(null=True, max_length=100)
     date_exchange_rate = models.DateTimeField(null=True)
 
+    @transaction.atomic
+    def save(self, *args, **kwargs):
+        """
+        Override save to make sure only one FinancialAid object exists for a User and the associated Program
+        """
+        if self.__class__.objects.filter(
```
noisecapella: Can you change `self.__class__` to `FinancialAid`, just for code clarity?

noisecapella:
Why not use `unique_together` instead?

financialaid/models.py
alvinsiu: Good idea, will do.

alvinsiu:
I don't think unique_together works across foreign keys.

noisecapella:
Ah, that's right. Can you add tests for this change?

alvinsiu:
Sure - will add tests, then update this PR.

financialaid/models.py
noisecapella: It looks like this change still needs to be made

financialaid/models_test.py
```diff
@@ -22,3 +24,20 @@ def test_timestamped_model_update(self):  # pylint: disable=no-self-use
...
+        """
+        Tests that FinancialAid objects are unique per User and Program
+        """
+        financial_aid = FinancialAidFactory.create()
+        try:
+            # Test creation of FinancialAid that isn't unique_together with "user" and "tier_program__program"
+            FinancialAidFactory.create(user=financial_aid.user)
+            FinancialAidFactory.create(tier_program=financial_aid.tier_program)
+        except ValidationError:
+            self.fail("Creation of FinancialAid objects should have been successful")
```
noisecapella: Can you remove the `except` clause here? It would be better to have the original stack trace when the test fails

financialaid/models_test.py
```diff
@@ -22,3 +24,20 @@ def test_timestamped_model_update(self):  # pylint: disable=no-self-use
         Tier.objects.filter(id=tier.id).update(name="new_tier")
         third_timestamp = Tier.objects.get(id=tier.id)  # Since we need to re-fetch the object
         assert second_timestamp != third_timestamp
+
+    def test_financial_aid_model_unique(self):
```
noisecapella: Can you also test that, if you update an existing `FinancialAid` you won't get the `ValidationError`?

financialaid/models_test.py
alvinsiu: Makes sense!

financialaid/models.py
alvinsiu: Ah, thanks!

financialaid/models_test.py
alvinsiu: Good idea!

alvinsiu:
Updated!

noisecapella:
Looks good :+1: Also please change `#1091` to `Fixes #1091` before merge

PR #1109, opened 2016-09-23
Dashboard course status refactor

gsidebo:

Not related to a specific ticket; consider it related to all of our open issues that relate to displaying courses in the dashboard

#### What's this PR do?

Refactors code around deciding and formatting course statuses

dashboard/api.py

Make sure the dashboard works!

Basically all of the code in the _test.py files is simply rewritten. The actual test case logic is basically the same


gsidebo:
@giocalitri i still need to fix the linting, but this is ready in the meantime

giocalitri:
after reading the beginning of the first file and in particular `class RunStatus:` that replaces two different kind of statuses, I think I need an overview of the idea you implemented.

gsidebo:
the two different status types were a pretty significant source of confusion, and after spending a good amount of time with this code I did not see the reason that we needed these different types. keep in mind that the same (refactored) unit tests are passing on this code

giocalitri:
I am not saying that the code was not clear, but I also wonder how you changed it if you did not fully understand the logic behind the previous implementation

dashboard/api.py
```diff
@@ -27,33 +27,23 @@
...
-    """
-    NOT_ENROLLED = 'not-enrolled'
-    GRADE = 'grade'
-    READ_CERT = 'read-cert'
-    WILL_ATTEND = 'will-attend'
-    UPGRADE = 'upgrade'
-    NOT_PASSED = 'not-passed'
+        """Helper to get all status values"""
+        return {cls.CURRENTLY_ENROLLED, cls.PASSED, cls.WILL_ATTEND, cls.CAN_UPGRADE,
+                cls.NOT_PASSED, cls.NOT_ACTIONABLE, cls.OFFERED}
```
giocalitri: just a comment: the major problem I have with this change is that from now on we will need to propagate any change to this class to the front-end

gsidebo:
i'm confident that i fully understood it. i actually kept the two classes until very recently when i noticed an easy and sensible way to combine them. i should also say that both `CourseStatus` and `CourseRunStatus` were used to express statuses for course _runs_. `CourseStatus` was never used for courses

dashboard/api.py
gsidebo: i don't think that is any different from what we currently have. if we changed a value in the old `CourseStatus`, we would need to make the same change in `static/js/constants.js` (`STATUS_PASSED`, `STATUS_NOT_PASSED`, etc). or are you talking about a different kind of change?

giocalitri: the idea behind having two sets of statuses before was about keeping CourseStatus in sync with the front end and having CourseRunStatus for internal logic

giocalitri: not saying that we actually took advantage of that though

dashboard/api.py
```diff
@@ -91,20 +81,19 @@ def get_assoc_field(cls, course_status):
...
 
-class CourseRunUserStatus:
+class UserCourseRun:
     """
-    Representation of a course run status for a specific user
+    Representation of a course run for a specific user
     """
-    def __init__(self, status, course_run=None, enrollment_for_course=None):
-        self.status = status
+    def __init__(self, course_run, status, certificate=None):
```
giocalitri: nit" why the change of sequence?

gsidebo: this ordering just made more sense to me. the CourseRun object is (in my opinion) the 'operative'/most important property of this class, and to me it's weird for the order of argument order to be `(string, complex_model_object, ...)`

noisecapella:
This is a pretty large PR. I'd prefer to avoid refactorings of this size until after the deadline unless there's an immediate need for it. When I mentioned that we should feel free to refactor whatever we want I was referring to the front end dashboard code which has a lot of outdated assumptions. IMO we should keep the server side code the same unless there's an immediate change that's necessary for some other piece of work

gsidebo:
@pdpinch @noisecapella @giocalitri in that case i'd like to open a different PR that just fixes the names of the statuses in the back end and front end. this has been a consistent source of confusion and is the likely cause of at least one bug. what do you all think?

noisecapella:
Sounds good to me

Ferdi:
That sounds like a good plan.  We make the changes to implement the correct statuses and consider the refactoring later.

giocalitri:
I am fine as long as this new PR does not explode in scope.

Ferdi:
Great,  @giocalitri and @gsidebo pls talk and come up with a plan before starting to work on this.

PR #1124, opened 2016-09-26
Switch to Django JSONField

### LINKED ISSUE #1095: Switch to Django JSONField
Currently we're using the field with the same name from the `jsonfield` package. Switching would make it easier to query within the JSON

noisecapella:
Fixes #1095 

#### What's this PR do?
Switches to Django's JSONField which would allow people to query the receipt data more flexibly

Everything should still work. These migrations should be non-destructive


aliceriot:
migrated, nothing blew up. is that all I need to test?

noisecapella:
If there was any data in any of the fields before, make sure that the data still exists afterwards

aliceriot:
I don't think I had data in any of those fields, unfortunately

noisecapella:
Try reversing the migration then creating a `Receipt` object with something in the `data` field

aliceriot:
yup, it worked!

aliceriot:
(merge conflict tho)

PR #1126, opened 2016-09-26
Set default staff page to learner search

### LINKED ISSUE #903: Home page based on user role
- The  "MIT Micromasters" should be a link that goes to users "home page"
- For learners, "home page" = dashboard
- For staff, "home page" = user search page

noisecapella:
Fixes #903

#### What's this PR do?
 - The Micromasters link in the navbar now points to either /dashboard or /learners depending on if you have a Instructor or Staff role
 - When the user logs in they are directed to /dashboard or /learners depending on if they have an instructor or staff role

Log in and out as a staff and a student and verify that the new page matches the expected one based on the description above. Also click the image on the upper left to test that the link is updated too.

### DIFF
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -3,12 +3,17 @@ APIs for extending the python social auth pipeline
 """
 import logging
 from datetime import datetime
+from rolepermissions.verifications import has_role
 from urllib.parse import urljoin
 
 from backends.edxorg import EdxOrgOAuth2
 from profiles.api import get_social_username
 from profiles.models import Profile
 from profiles.util import split_name
+from roles.models import (
+    Instructor,
+    Staff,
+)
 
 log = logging.getLogger(__name__)
 
@@ -30,7 +35,15 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):  
 
     # this function is completely skipped if the backend is not edx or
     # the user has not created now
-    if backend.name != EdxOrgOAuth2.name or not is_new:
+    if backend.name != EdxOrgOAuth2.name:
+        return
+
+    if has_role(user, [Staff.ROLE_ID, Instructor.ROLE_ID]):
+        backend.strategy.session_set('next', '/learners')
+    else:
+        backend.strategy.session_set('next', '/dashboard')
+
+    if not is_new:
         return
 
     access_token = response.get('access_token')
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -3,8 +3,9 @@ APIs for extending the python social auth pipeline
 """
 import logging
 from datetime import datetime
-from rolepermissions.verifications import has_role
 from urllib.parse import urljoin
+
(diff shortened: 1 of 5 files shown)


amir-qayyum-khan:
Functionality works as mentioned.

backends/pipeline_api.py
```diff
@@ -5,10 +5,16 @@
 from datetime import datetime
 from urllib.parse import urljoin
 
+from rolepermissions.verifications import has_role
+
```
amir-qayyum-khan: why space here? can you remove this.

backends/pipeline_api.py
```diff
@@ -30,7 +36,15 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
...
-    if backend.name != EdxOrgOAuth2.name or not is_new:
+    if backend.name != EdxOrgOAuth2.name:
+        return
+
+    if has_role(user, [Staff.ROLE_ID, Instructor.ROLE_ID]):
+        backend.strategy.session_set('next', '/learners')
+    else:
+        backend.strategy.session_set('next', '/dashboard')
+
+    if not is_new:
```
amir-qayyum-khan: why you are checking ``is_new`` at bottom? 
if user is not ``is_new`` then will he have roles?
what session_set for next do for not is_new users?
And what it do for new users?

backends/pipeline_api_test.py
```diff
@@ -209,12 +215,34 @@ def test_preferred_language(self, mocked_get_json):
...
             assert self.user_profile.preferred_language is None
             assert self.user_profile.edx_language_proficiencies == proficiencies
 
+    def test_login_redirect_dashboard(self):
+        """
+        A student should be directed to /dashboard
+        """
+        student = UserFactory.create()
+        session_set = mock.Mock()
+        backend = edxorg.EdxOrgOAuth2(strategy=mock.Mock(session_set=session_set))
```
amir-qayyum-khan: @noisecapella Is their test for [``not is_new``](https://github.com/mitodl/micromasters/blob/cb4725fc85a6632b9d7862e6875b09a60de40343/backends/pipeline_api.py#L47)  user. What behavior the get?

backends/pipeline_api.py
noisecapella: The imports are broken up into three groups, the space separates the imports from third party libraries from the imports from our own code: https://www.python.org/dev/peps/pep-0008/#imports

backends/pipeline_api_test.py
noisecapella: This test and the next one set `is_new` to be `False`

backends/pipeline_api.py
noisecapella: `session_set` will set the next page after the user logs in via edX. The code should be executed regardless of whether `is_new` is true or not so I put it above the `is_new` check.

noisecapella:
Any other comments?

amir-qayyum-khan:
👍

PR #1136, opened 2016-09-27
Used fill rule to crop faculty images

noisecapella:
None
#### What's this PR do?
Updates faculty images to crop in the same way as the program thumbnail images


roberthouse54:
👍

PR #1138, opened 2016-09-27
Financial Aid Auditing

### LINKED ISSUE #1047: Financial Aid: auditing
# Acceptance criteria 
- [ ] Track status changes to financial aid requests
- [ ] Track status changes to tier changes
- [ ] Track initial emails to financial aid recipients

alvinsiu:
Closes #1047 

#### What's this PR do?
Implements auditing for FinancialAid status changes and emails sent in relation to financial aid.


mail/api.py
```diff
@@ -29,12 +38,11 @@ def _mailgun_request(cls, request_func, endpoint, params):
             requests.Response: HTTP response
         """
         mailgun_url = '{}/{}'.format(settings.MAILGUN_URL, endpoint)
-        emails_params = {'from': settings.MAILGUN_FROM_EMAIL}
-        emails_params.update(**params)
+        params.update(cls.get_base_params())
```
noisecapella: Instead of calling `params.update` here we should make a copy first. The caller should expect that the arguments passed to the function will not modified unless the function documents so explicitly

mail/api.py
```diff
@@ -29,12 +38,11 @@ def _mailgun_request(cls, request_func, endpoint, params):
...
         return request_func(
             mailgun_url,
-            auth=('api', settings.MAILGUN_KEY),
-            data=dict(**emails_params)
+            auth=cls._basic_auth_credentials,
+            data=dict(**params)
```
noisecapella: We can also just do `data=params` here but feel free to leave it if you prefer

mail/api_test.py
```diff
@@ -122,16 +131,42 @@ class FinancialAidMailAPITests(TestCase):
     """
     Tests for the Mailgun client class for financial aid
     """
+    @classmethod
+    def setUpTestData(cls):
+        with mute_signals(post_save):
+            cls.staff_user_profile = ProfileFactory.create()
+            cls.financial_aid = FinancialAidFactory.create()
```
noisecapella: Can you move this outside the `mute_signals` block?

mail/api_test.py
```diff
@@ -122,16 +131,42 @@ class FinancialAidMailAPITests(TestCase):
...
         assert called_kwargs['data']['subject'] == 'email subject'
-        assert called_kwargs['data']['to'] == ['a@example.com']
+        assert called_kwargs['data']['to'] == [self.financial_aid.user.email]
+        # Check audit creation
+        assert FinancialAidEmailAudit.objects.count() == 1
+        audit = FinancialAidEmailAudit.objects.first()
+        assert audit.acting_user == self.staff_user_profile.user
+        assert audit.financial_aid == self.financial_aid
+        assert audit.to_email == self.financial_aid.user.email
+        assert audit.email_subject.startswith('email subject')
```
noisecapella: If we don't need `startswith` here and above can you change to a direct comparison?

mail/api_test.py
```diff
@@ -122,16 +131,42 @@ class FinancialAidMailAPITests(TestCase):
...
         assert called_kwargs['data']['text'].startswith('email body')
         assert called_kwargs['data']['subject'] == 'email subject'
-        assert called_kwargs['data']['to'] == ['a@example.com']
+        assert called_kwargs['data']['to'] == [self.financial_aid.user.email]
+        # Check audit creation
+        assert FinancialAidEmailAudit.objects.count() == 1
+        audit = FinancialAidEmailAudit.objects.first()
+        assert audit.acting_user == self.staff_user_profile.user
+        assert audit.financial_aid == self.financial_aid
+        assert audit.to_email == self.financial_aid.user.email
```
noisecapella: Can we assert `from_email` too?

mail/models.py
```diff
@@ -0,0 +1,19 @@
...
+from financialaid.models import TimestampedModel, FinancialAid
+
+
+class FinancialAidEmailAudit(TimestampedModel):
+    """
+    Audit table for the Financial Aid
+    """
+    acting_user = models.ForeignKey(User, null=False)
+    financial_aid = models.ForeignKey(FinancialAid, null=True, on_delete=models.SET_NULL)
+    to_email = models.CharField(null=False, max_length=250)
```
noisecapella: Can we use `TextField` here and below instead? AFAIK there is no performance penalty for using `TextField` when working with postgres

mail/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+from financialaid.models import TimestampedModel, FinancialAid
+
+
+class FinancialAidEmailAudit(TimestampedModel):
+    """
+    Audit table for the Financial Aid
+    """
+    acting_user = models.ForeignKey(User, null=False)
+    financial_aid = models.ForeignKey(FinancialAid, null=True, on_delete=models.SET_NULL)
```
noisecapella: It seems like this is never going to be `None`, can we change to `null=False`?

mail/models.py
```diff
@@ -0,0 +1,19 @@
...
+    to_email = models.CharField(null=False, max_length=250)
+    from_email = models.CharField(null=False, max_length=250)
+    email_subject = models.CharField(null=False, max_length=250)
```
noisecapella: I think this and `email_body` should have `blank=True` in case the email sent is missing a subject or a body. Can you also add a test for that case?

mail/views_test.py
```diff
@@ -114,28 +115,33 @@ class FinancialAidMailViewsTests(APITestCase):
...
+            'financial_aid_mail_api',
+            kwargs={'financial_aid_id': cls.financial_aid.id}
+        )
         cls.request_data = {
             'email_subject': 'email subject',
             'email_body': 'email body',
-            'email_recipient': 'a@example.com'
         }
 
     def test_send_financial_aid_view(self, mock_mailgun_client):
```
noisecapella: Can you add a test for the case where the user doesn't have permission to send the email for a given `FinancialAid`?

mail/api.py
alvinsiu: This was how it was before I edited it - I'll leave it in place in case it was done this way for a reason.

mail/api_test.py
alvinsiu: Just the `cls.financial_aid`, right?

mail/api_test.py
alvinsiu: Sure - I just copied the test structure from the existing mail api tests.

mail/api_test.py
alvinsiu: Will do!

mail/models.py
alvinsiu: Ah, I forget we're using postgre - I agree.

mail/models.py
alvinsiu: We thought it would be important to allow it to be null in case there happens to be a case where the `FinancialAid` object is deleted.  The last thing we'd want is for an audit record to be deleted.

mail/models.py
alvinsiu: Good idea!

mail/api.py
alvinsiu: Okie doke

mail/views_test.py
alvinsiu: Yup!

mail/api.py
alvinsiu: I lied - I changed it.

mail/api_test.py
noisecapella: Yes

mail/models.py
noisecapella: I feel like there would be an integrity error due to the null constraint if the user deleted only `FinancialAid`. But I'm not sure so let's leave it then.

noisecapella:
Functionality looks good, taking one more look through...

noisecapella:
Everything looks good :+1:

PR #1141, opened 2016-09-27
Abstracted "MicroMasters" track 

### LINKED ISSUE #1077: Abstract "MicroMasters" track to include financial aid case

### LINKED ISSUE #1078: Return final grade in API for financial aid-enabled courses
MicroMasters programs with financial aid won't use certificates, so we can't rely on the `certificates` API to get the user's final grade or to determine the user's passed status. 

For those programs we can determine the "final" grade and passed/not-passed status by making a call to the course_grade API after the course end_date.

For example, using my user, a call to https://courses.edx.org/api/grades/v0/course_grade/HarvardX/SW12.3x/1T2014/users/?username=pdpinch returns:

```
[
    {
        "username": "pdpinch",
        "letter_grade": "Pass",
        "percent": 0.97,
        "course_key": "HarvardX/SW12.3x/1T2014",
        "passed": true
    }
]
```

giocalitri:
closes #1077 
closes #1078

#### What's this PR do?
Abstracts the MM tract in the dashboard in order to handle `verified` programs and `financial aid`.


dashboard/utils.py
```diff
@@ -0,0 +1,173 @@
+"""
```
gsidebo: looks like you have a lot of forked logic around `self.financial_aid_available` being true or false. i think this should be implemented as 2 different classes with the common functionality abstracted to a parent class

giocalitri: We agreed this is a good idea for post deadline refactoring

courses/models.py
```diff
@@ -90,7 +90,7 @@ class CourseRun(models.Model):
       rather a specific instance of that course being taught.
     """
     title = models.CharField(max_length=255)
-    edx_course_key = models.CharField(max_length=255, blank=True, null=True)
+    edx_course_key = models.CharField(max_length=255, blank=True, null=True, db_index=True)
```
noisecapella: This needs a migration

ecommerce/models.py
```diff
@@ -35,6 +35,7 @@ class Order(Model):
         choices=[(status, status) for status in STATUSES],
         default=CREATED,
         max_length=30,
+        db_index=True,
```
noisecapella: These two changes need a migration

noisecapella: I'm not sure we need this index, this column has low cardinality so the index may not give us much: http://blog.celerity.com/how-to-design-sql-indexes

dashboard/views.py
```diff
@@ -59,9 +59,16 @@ def get(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-s
         # get certificates for the student
         certificates = get_student_certificates(request.user, edx_client)
         # get current_grades for the student
-        current_grades = get_student_current_grades(request.user, edx_client)  # pylint: disable=unused-variable
+        current_grades = get_student_current_grades(request.user, edx_client)
 
         response_data = []
-        for program in Program.objects.filter(live=True):
-            response_data.append(get_info_for_program(program, enrollments, certificates))
+        for program_enrollment in ProgramEnrollment.objects.filter(user=request.user):
```
noisecapella: Can we keep the filter on `live=True`? Just a precaution

dashboard/utils.py
```diff
@@ -0,0 +1,180 @@
...
+
+class RequiredFieldMissing(Exception):
+    """
+    Custom exception to raise in case a field that is not mandatory in the db
+    is mandatory for the implementation of some business logic.
+    Example CourseRun end date.
+    """
+
+
+class MMTrack:
```
noisecapella: Can you add descriptions for the args and return values for the public methods?

dashboard/utils.py
```diff
@@ -0,0 +1,180 @@
...
+            certificates (edx_api.certificates.models.Certificates): User's certificates
+        """
+        self.user = user
+        self.program = program
+        self.enrollments = enrollments
+        self.current_grades = current_grades
+        self.certificates = certificates
+        self.financial_aid_available = program.financial_aid_availability
+
+        self.course_ids = list(CourseRun.objects.filter(course__program=self.program).exclude(
```
noisecapella: Can we wrap the rest of this function in a `transaction.atomic`?

dashboard/utils.py
```diff
@@ -0,0 +1,180 @@
...
+            Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
+        ).values_list("edx_course_key", flat=True))
+
+        if self.financial_aid_available:
+            self.paid_course_ids = list(Line.objects.filter(
```
noisecapella: Do we have this information if financial aid is not available?

dashboard/utils.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        return self._is_course_in_program(course_id) and self.enrollments.is_enrolled_in(course_id)
+
+    def is_enrolled_mmtrack(self, course_id):
+        """
+        Returns whether an used is enrolled mmtrack in a course run.
+        This means if the user is enrolled verified for normal programs
+        or enrolled and paid on micromasters for financial aid ones.
+        """
+        if not self.is_enrolled:
```
noisecapella: I think this was meant to be a function call

dashboard/utils.py
noisecapella: I think I understand now, `paid_course_ids` is defined to only include payments on our own servers, right?

dashboard/utils.py
```diff
@@ -0,0 +1,180 @@
...
+        if not self.financial_aid_available:
+            return self.certificates.has_verified_cert(course_id)
+        # financial aid programs need to have an audit enrollment,
+        # a current grade with a passed attribute and the course should be ended
+        else:
+            cur_grade = self.current_grades.get_current_grade(course_id)
+            if cur_grade is None:
+                return False
+            course_run_qset = CourseRun.objects.filter(edx_course_key=course_id)
+            if not course_run_qset.exists():
```
noisecapella: Can we use `first()` and check it for `None` instead? There's the (tiny) possibility of a race condition since there are two SQL statements here without a transaction

dashboard/utils.py
giocalitri: that is right

dashboard/utils.py
giocalitri: sure

dashboard/utils.py
giocalitri: yep

dashboard/views.py
giocalitri: sure

dashboard/utils.py
```diff
@@ -0,0 +1,180 @@
...
+        Returns the course final grade number for the user if she passed.
+        This means the grade in the certificate for normal programs
+        or the current_grade for ended courses.
+        """
+        if not self.has_passed_course(course_id):
+            return
+        # for normal programs need to pull from the certificate
+        if not self.financial_aid_available:
+            certificate = self.certificates.get_verified_cert(course_id)
+            return certificate.grade
```
noisecapella: This is also a value from `0` to `100`?

dashboard/api.py
```diff
@@ -119,59 +117,62 @@ def __repr__(self):
...
         course (Course): a course object
-        user_enrollments (Enrollments): the user enrollments object
-        user_certificates (Certificates): the user certificates objects
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program
 
     Returns:
         dict: dictionary representing the course status for the user
     """
     # pylint: disable=too-many-statements
+    def _add_run(course_data, run, mmtrack, status):
```
noisecapella: Can you rename `mmtrack` and `course_data` to not shadow the outside variables?

noisecapella:
I'm done looking over the code other than the tests, let me know when I should look at the tests and functionality

dashboard/utils_test.py
```diff
@@ -0,0 +1,395 @@
...
+            program=cls.program_financial_aid,
+            start_date=cls.now-timedelta(weeks=52),
+            end_date=cls.end_date,
+            enrollment_start=cls.now-timedelta(weeks=62),
+            enrollment_end=cls.now-timedelta(weeks=53),
+            edx_course_key="course-v1:odl+FOO101+CR-FALL15"
+        )
+        CourseRunFactory.create(
+            program=cls.program_financial_aid,
+            edx_course_key=""
```
noisecapella: Why is this blank?

giocalitri: We do not want to handle course run with empty `edx_course_key`, meaning runs that have not a correspondent run on edX

giocalitri: Actually there is the case with `edx_course_key=None` missing

noisecapella:
Can you add tests for the uncovered cases in dashboard/util.py? Specifically to cover lines 
 - https://github.com/mitodl/micromasters/pull/1141/files#diff-bf545a2544a83f1acac6cd6d14942b19R158
 - https://github.com/mitodl/micromasters/pull/1141/files#diff-bf545a2544a83f1acac6cd6d14942b19R162
 - https://github.com/mitodl/micromasters/pull/1141/files#diff-bf545a2544a83f1acac6cd6d14942b19R205

dashboard/utils_test.py
```diff
@@ -0,0 +1,396 @@
...
+            user=self.user,
+            program=self.program,
+            enrollments=self.enrollments,
+            current_grades=self.current_grades,
+            certificates=self.certificates
+        )
+        for course_id in ["course-v1:edX+DemoX+Demo_Course", "course-v1:MITx+8.MechCX+2014_T1"]:
+            assert mmtrack.is_enrolled(course_id) is True
+
+        mmtrack_fa = MMTrack(
```
noisecapella: Can you add a comment on what is being tested here, that wasn't tested in the previous assert?

dashboard/utils_test.py
```diff
@@ -0,0 +1,396 @@
...
+            program=self.program,
+            enrollments=self.enrollments,
+            current_grades=self.current_grades,
+            certificates=self.certificates
+        )
+        for course_id in ["course-v1:edX+DemoX+Demo_Course", "course-v1:MITx+8.MechCX+2014_T1"]:
+            assert mmtrack._is_course_in_program(course_id) is True
+        assert mmtrack._is_course_in_program("course-v1:odl+FOO101+CR-FALL15") is False
+
+    def test_is_enrolled(self):
```
noisecapella: Can you also create a test case where `is_enrolled()` returns `False`?

dashboard/utils_test.py
```diff
@@ -0,0 +1,396 @@
...
+        assert mmtrack.has_passed_course(course_id) is False
+
+        # remove the end date
+        self.crun_fa.end_date = None
+        self.crun_fa.save()
+        with self.assertRaises(ImproperlyConfigured):
+            mmtrack.has_passed_course(course_id)
+
+        # restore the end date
+        self.crun_fa.end_date = self.end_date
```
noisecapella: We shouldn't need this code, `setUpTestData` will roll back the transaction after the test completes

dashboard/utils_test.py
giocalitri: ok

giocalitri:
>Can you add tests for the uncovered cases in dashboard/util.py? Specifically to cover lines
>
>https://github.com/mitodl/micromasters/pull/1141/files#diff-bf545a2544a83f1acac6cd6d14942b19R158
>https://github.com/mitodl/micromasters/pull/1141/files#diff-bf545a2544a83f1acac6cd6d14942b19R162
>https://github.com/mitodl/micromasters/pull/1141/files#diff-bf545a2544a83f1acac6cd6d14942b19R205

Done for first and last case

noisecapella:
Ok, finally made it through the tests. Everything looks good for the most part, and functionality works fine

dashboard/utils_test.py
giocalitri: ok

giocalitri:
I should have addressed the latest comments

noisecapella:
:+1:

Issue #1048, opened 2016-09-15
Financial Aid: Currency Conversion


pdpinch:
# Acceptance Criteria

- [x] Currency conversion API
- [x] Cache currency conversion results and only refetch if cache is missing or older than a day
- [ ] API endpoint to get default currency for a user (or have the data file used on frontend)

## Questions

- [x] front-end (JavaScript) API or backend?

alvinsiu:
We should probably be doing this on the backend for better auditing and to be able to cache results from the API.

alvinsiu:
Removed - will be doing backend:

> ## Questions
> 
> - [x] front-end (JavaScript) API or backend?

alvinsiu:
@pdpinch Are there any business requirements about what currencies we need to support or not support?  And do we need to display that information on the frontend somewhere?

Right now, the way we've designed it, the API endpoint will return a ValidationError if the currency isn't supported by the library we use, and the celery task will ensure all all exchange rates are up to date (and consequently add new ones and delete no-longer-supported ones).  Is that sufficient?

alvinsiu:
@aliceriot Just wanted to double check that I have things stright - the first two tasks will be closed by #1146, and the second two tasks are your domain, correct?

The "Data file of currencies per country" is currently in `financialaid/data/country_to_currency.csv`.  Let me know if there's anything we can help with other than that!

aliceriot:
Oh! I forgot about that file. Is the backend just expecting an ISO 4217 currency code? I found a JS package that provides a nice api for code -> country correspondance using that standard: https://github.com/freeall/currency-codes

alvinsiu:
Yup!  The [documentation](https://docs.openexchangerates.org/) for the exchange rate API we're using only says "3-letter ISO code), but I think ISO 4217 is the only major one for currencies.

Up to you if you want to use that file or not!

aliceriot:
I already wrote the code using that JS package, so we should probably delete the file at some point.

alvinsiu:
Oh cool!  Should we remove these two then?  Or are those still in queue?

> - [ ] Data file of currencies per country
> - [ ] API endpoint to get default currency for a user (or have the data file used on frontend)

aliceriot:
I think we can take care of both of those in JS

alvinsiu:
Should I remove those two items from this issue (if they're already on another issue) or leave them here for you?

aliceriot:
[...]

PR #1146, opened 2016-09-27
Currency Conversion

alvinsiu:
#1048 

#### What's this PR do?
Implements a celery task for updating currency exchange rates every 24 hours. Also integrates with initial financial aid request to convert non-USD currencies to USD.  This covers the first two tasks in #1048.


alvinsiu:
Will merge master into this branch at a later time - there will likely be conflicts from some other PRs that are currently in the pipelinee, so I want to wait until those are merged before trying to fix merge conflicts.

noisecapella:
I think the other relevant PRs are merged, can you rebase?

noisecapella:
Can you revert the permission changes in these files back to 644?

financialaid/exceptions.py
```diff
@@ -0,0 +1,10 @@
+"""
+Exceptions for financialaid.
+"""
+
+
+class NotSupportedException(Exception):
+    """
+    Not supported by current financial aid system.
+    """
+    pass
```
noisecapella: This is not necessary since you have the docstring. Can you remove this line please?

financialaid/migrations/0005_modified_financial_aid_audit.py
```diff
@@ -9,7 +9,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('financialaid', '0003_added_timestamps'),
+        ('financialaid', '0004_currencyexchangerate'),
```
noisecapella: Please put the new migration after all other existing ones, and leave the old migrations untouched

financialaid/api.py
```diff
@@ -45,6 +47,27 @@ def determine_auto_approval(financial_aid):
     return financial_aid.income_usd > income_threshold or income_threshold == 0
 
 
+def determine_income_usd(original_income, original_currency):
+    """
+    Take original income and original currency and converts income from the original currency
+    to USD.
+    Args:
+        original_income: original income, in original currency (for a FinancialAid object)
```
noisecapella: Can you add types within parentheses here and in the next line?

financialaid/api.py
```diff
@@ -45,6 +47,27 @@ def determine_auto_approval(financial_aid):
...
+        original_currency: original currency, a three-letter code
+    Returns:
+        income_usd: the original income converted to US dollars
```
noisecapella: The format here should be `float: the original income...` where the word before the `:` is a type

financialaid/exceptions.py
alvinsiu: Yup!

financialaid/migrations/0005_modified_financial_aid_audit.py
alvinsiu: Yup!

financialaid/api.py
alvinsiu: Yup!

financialaid/api.py
alvinsiu: Yup!

financialaid/models.py
```diff
@@ -166,3 +166,11 @@ class FinancialAidAudit(TimestampedModel):
     financial_aid = models.ForeignKey(FinancialAid, null=True, on_delete=models.SET_NULL)
     data_before = JSONField(blank=True, null=False)
     data_after = JSONField(blank=True, null=False)
+
+
+class CurrencyExchangeRate(TimestampedModel):
+    """
+    Table of currency exchange rates
```
noisecapella: Can you update this to mention that the `exchange_rate` field is the amount of foreign currency per single USD, just to be explicit?

noisecapella:
In `mail/migrations/0001_initial.py` can you update the dependency from `0005` and `0004` to match the migration name? I think I forgot to update it in my own migration PR

alvinsiu:
Ah, yeah, I can do that.

financialaid/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+    Updates all CurrencyExchangeRate objects to reflect latest exchange rates from
+    Open Exchange Rates API (https://openexchangerates.org/).
+    """
+    url = "{url}latest.json?app_id={app_id}".format(
+        url=settings.OPEN_EXCHANGE_RATES_URL,
+        app_id=settings.OPEN_EXCHANGE_RATES_APP_ID
+    )
+    resp = requests.get(url).json()
+    latest_rates = resp["rates"]
+    for currency_exchange_rate in CurrencyExchangeRate.objects.all():
```
noisecapella: Can you move this currency update code to a separate function in `api.py`? And we should also decorate that function with `@transaction.atomic` to make sure all exchange rates update at the same time

financialaid/models.py
alvinsiu: Good idea!

financialaid/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+
+@async.task
+def update_currency_exchange_rates():
+    """
+    Updates all CurrencyExchangeRate objects to reflect latest exchange rates from
+    Open Exchange Rates API (https://openexchangerates.org/).
+    """
+    url = "{url}latest.json?app_id={app_id}".format(
+        url=settings.OPEN_EXCHANGE_RATES_URL,
+        app_id=settings.OPEN_EXCHANGE_RATES_APP_ID
```
alvinsiu: Yup!

financialaid/tasks_test.py
```diff
@@ -0,0 +1,78 @@
...
+        Assert task schedule using celery beat.
+        """
+        self.assertTrue(update_currency_exchange_rates.delay())
+
+    @override_settings(CELERY_ALWAYS_EAGER=True)
+    def test_update_and_add_currency_exchange_rates(self, mocked_request):
+        """
+        Assert currency exchange rates are updated and added
+        """
+        mocked_request.return_value.json.return_value = self.data
```
noisecapella: Can we assert the `URL` passed to `get`?

financialaid/tasks.py
alvinsiu: Yup!

financialaid/tasks_test.py
```diff
@@ -0,0 +1,78 @@
...
+                "DEF": "2",
+                "MNO": "1.7",
+                "PQR": "0.4"
+            }
+        }
+
+    @override_settings(CELERY_ALWAYS_EAGER=True)
+    def test_celery_task_works(self, mocked_request):  # pylint: disable=unused-argument
+        """
+        Assert task schedule using celery beat.
```
noisecapella: I'm not sure this test is doing anything useful, and it doesn't seem to interact with celery beat at all. Can we delete it?

alvinsiu: Sure

financialaid/views_test.py
```diff
@@ -117,10 +128,27 @@ def test_income_validation_user_not_enrolled(self):
...
+        self.data["original_currency"] = "ABC"
+        assert FinancialAid.objects.count() == 0
+        resp = self.client.post(self.request_url, self.data, format='json')
+        assert resp.status_code == status.HTTP_201_CREATED
+        assert FinancialAid.objects.count() == 1
+        financial_aid = FinancialAid.objects.first()
+        exchange_rate = CurrencyExchangeRate.objects.get(currency_code="ABC").exchange_rate
+        assert financial_aid.tier_program == self.tier_programs["50k"]
+        assert financial_aid.status == FinancialAidStatus.PENDING_DOCS
+        assert round(financial_aid.income_usd, 5) == round(self.data["original_income"]/exchange_rate, 5)
```
noisecapella: Would `assertAlmostEqual` work here?

financialaid/api_test.py
```diff
@@ -219,6 +233,15 @@ def test_determine_auto_approval(self):  # pylint: disable=no-self-use
         )
         assert determine_auto_approval(financial_aid) is True
 
+    def test_determine_income_usd(self):  # pylint: disable=no-self-use
+        """
+        Tests determine_income_usd()
+        """
+        # original income is in US dollars
+        assert determine_income_usd(5000, "USD") == 5000
+        # original income is in ABC currency
```
noisecapella: Can you update `ABC` to `GHI` in the comment?

financialaid/tasks_test.py
alvinsiu: Yup!

noisecapella:
Functionality works great

financialaid/api_test.py
alvinsiu: Yup

financialaid/views_test.py
alvinsiu: Yep, changing it now

noisecapella:
I'm seeing some currency codes unaccounted for in the JS client and on the server side (which comes from the API). I'll make a separate issue for that though

alvinsiu:
Like there are more currencies on the JS client than what we get from openexchangerates?  Or the other way around?

And I'm fixing the failed build now.

noisecapella:
Both. I'm not sure how to resolve the discrepancies. For example @aliceriot filtered out `USS` and `USN` because they are special purpose currencies. The exchange rate API includes `BTC` which may not be relevant to us

financialaid/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+from micromasters.celery import async
+
+
+@async.task
+def sync_currency_exchange_rates():
+    """
+    Updates all CurrencyExchangeRate objects to reflect latest exchange rates from
+    Open Exchange Rates API (https://openexchangerates.org/).
+    """
+    url = quote_plus(
```
noisecapella: `quote_plus` is only for escaping the query parameter, not the entire URL

alvinsiu:
@pdpinch what should we do about discrepancies between the frontend and backend currencies that are supported?  We can easily whitelist or blacklist currencies, but there's no guarantee that either API's supported currencies will be static.

financialaid/tasks.py
alvinsiu: My bad, will update

pdpinch:
Presumably the API should be the source of truth, since we can't do anything useful with a currency that it doesn't support. 


However, I'm not terribly concerned about this right now and I'm willing to put off the ideal solution until sometime after 10/15. 

We should move this discussion to the issue George opened.

financialaid/tasks.py
alvinsiu: Should I do it on both parameters?

noisecapella: Only on the `GET` query parameter

financialaid/tasks.py
```diff
@@ -17,11 +17,9 @@ def sync_currency_exchange_rates():
...
     Open Exchange Rates API (https://openexchangerates.org/).
     """
-    url = quote_plus(
-        "{url}latest.json?app_id={app_id}".format(
-            url=settings.OPEN_EXCHANGE_RATES_URL,
-            app_id=settings.OPEN_EXCHANGE_RATES_APP_ID
-        )
+    url = "{url}latest.json?app_id={app_id}".format(
+        url=settings.OPEN_EXCHANGE_RATES_URL,
+        app_id=quote_plus(settings.OPEN_EXCHANGE_RATES_APP_ID)
```
noisecapella: I think you'll need to update the test too

alvinsiu: Aaaarg, stop being right! :]

noisecapella:
Looks good, feel free to merge after tests pass :+1:

alvinsiu:
Thanks for the review and being so quick on this!

PR #1152, opened 2016-09-28
Responsive program page

singingwolfboy:
Making the program page responsive. @roberthouse54, could you review this?
top of program page
middle of program page
faculty on program page

### DIFF
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -23,10 +23,10 @@
       {{ page.title_over_image|richtext }}
     </div>
     {% if not authenticated %}
-    <button class="mm-button main-action open-signup-dialog">
+    <button class="mm-button main-action open-signup-dialog mdl-cell--hide-phone">
       Sign Up Now
     </button>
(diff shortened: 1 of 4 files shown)


Ferdi:
@singingwolfboy is there anyway  to keep "Faculty" as carousel with one image per slide ?

roberthouse54:
@Ferdi I requested that on mobile, we just switch to a stacked vertical layout, since users will be vertically scrolling anyway. My concern is that it will be an awkward layout on mobile, and users may not realize that there is more than one faculty.

roberthouse54:
I am thinking I would like to merge this as is, since it is mostly all working. I will want to go through both this and the home page and make some tweaks for mobile layout, but I'd like to do that as a separate PR.

singingwolfboy:
@roberthouse54 Sounds good to me. Let's wait for all the tests to pass on Travis CI, and then you're clear to hit the merge button, as far as I'm concerned.

roberthouse54:
👍

PR #1156, opened 2016-09-28
refactor getPreferredName

### LINKED ISSUE #1068: Sign-up page: greeting doesn't dynamically change name correctly
The sign-up page at `/profile/` has a nice feature where the greeting will dynamically change as the user types. However, it doesn't do so _correctly_, which leads to a confusing user experience.

# Actual behavior
When the user first visits the page, the greeting contains the first part of their email address, like so:
initial signup

However, after the user enters their given and family names, only the family name appears:
typing in first and last name

After the user types in their preferred name, the given name is updated with their preferred name, but the family name still appears:
typing in given, family, and preferred names

# Expected behavior
When the user first visits the page, the greeting should contain the first part of their email address. The application does this correctly.

When the user types in any text into the "given name" and "family name" fields, they should _both_ override the email address fragment. If the user types "First" into the given name file and "Last" into the family name field, the result should be "First Last" in the greeting.

When the user types any text into the "Preferred name" field, that should override all other fields. If the user types "Preferred" into this field, the result should be "Preferred" in the greeting (not "Preferred Last").

aliceriot:

closes #1068 

#### What's this PR do?

This refactors the `getPreferredName` function to be a little smarter. Basically, it now has the correct precedence for first names (`preferred_name > first_name > username`).

Read over the changes!

1. go to `/profile`
2. delete `preferred_name`, `first_name`, and `last_name` in the fields.
3. you should see `Hello ${username}`
4. add a first name
5. you should see `Hello ${first_name}`
6. add a last name
7. you should see `Hello ${first_name} ${last_name}`
8. add a preferred name
9. you should see `Hello ${preferred_name}`

I think that's it!

preferred


static/js/util/util.js
```diff
@@ -251,16 +251,8 @@ export function makeProfileImageUrl(profile: Profile): string {
...
-    first = profile.preferred_name || SETTINGS.name || SETTINGS.username;
-  } else {
-    first = profile.preferred_name || profile.first_name;
-  }
-  if ( last ) {
-    return profile.last_name ? `${first} ${profile.last_name}` : first;
-  }
-  return first;
+  let first = profile.preferred_name || profile.first_name || profile.username;
+  return last && profile.last_name && !profile.preferred_name ? `${first} ${profile.last_name}` : first;
```
annagav: Should we remove `name` or `username` form `js_settings` in ui/views.py?

aliceriot: We use `SETTINGS.username` in a bunch of places so I'd be inclined to leave that, not sure about `SETTINGS.name` though

annagav: if it is not used elsewhere...

aliceriot: I'm happy to delete it, @noisecapella what do you think? do we need to keep `SETTINGS.name` for anything?

noisecapella: No, I'm fine removing it

aliceriot:
@annagav I removed `SETTINGS.name`

annagav:
When all three fields are clear, it shows the username, as expected. But if I start typing last name, it combines the username with last name. Is that the correct behavior?

screen shot 2016-09-28 at 5 05 26 pm

aliceriot:
hmm not sure about that, it's a bit weird to be sure. what do you think?

annagav:
I think it is ok as long as the username resembles some sort of first name. @pdpinch, what do you think?

roberthouse54:
This seems needlessly complex to me. I dont think having this weird typeahead feature adds anything to the user experience. Can we just get rid of it?

aliceriot:
The change as-you-edit feature isn't being added in this PR, this is just
changing the function we use to get a user's name for display (from their
profile).

I think it wouldn't be that hard to remove it if that's what we want but I
think it's a separate issue.

annagav:
Let's leave it as it is. Functionality looks good!

annagav:
Looks great! 👍

Issue #1058, opened 2016-09-17
ecommerce for DEDP


pdpinch:
Tracks tasks for DEDP e-commerce

# ToDos

- [x] enable ecommerce conditionally (DEDP needs it. SCM does not)
- [x] Integration with Cybersource, MIT's preferred ecommerce provider
  - [x] #904 order receipt page / message 
  - [x] #905 order cancellation page
- [x] Fulfillment through edX, by enrolling the user in an audit track
- [x] use personalized prices when checking out
- [x] track the verified status of the user in the micromasters portal

pdpinch:
@noisecapella can you take a look at this and add anything that's missing to the description?

pdpinch:
remove #1043 file storage configuration as we're not taking files anymore.

pdpinch:
@noisecapella now that #913 has been merged, can we check off any of these boxes?

noisecapella:
Just the Cybersource box. We store the result of the payment but we don't do anything with that information yet.

pdpinch:
Can you work on #904?

Are you blocked by mitodl/edx-api-client#26 (I believe Giovanni is working on it)?

noisecapella:
No, once the enrollment API function is in edx-api-client it should be a small change to cause the order fulfillment to actually enroll the user. I'll start on the receipt page. By the way there's been no work done yet to make use of the financial aid calculated price, or to provide different checkout flows for SCM and DEDP

Ferdi:
@noisecapella @pdpinch can you update  the ticket description with all dependent issues ?

PR #1159, opened 2016-09-28
Implemented personal pricing for ecommerce

noisecapella:
Part of #1058 

#### What's this PR do?
Uses personal price when charging user.

Talk to me first to set up the Cybersource test environment variables

- Create a `TierProgram` for some program with some discount amount if you don't have one already. 
- Create a `FinancialAid` with status `approved` referencing that `TierProgram`.
-  Create a `CoursePrice` on some `CourseRun` which has `is_valid=True`. Make sure that the price is more than the discount you set. Set up `CourseRun` such that the enroll button shows up in the dashboard (soon to be replaced with a Pay button)
 - Click the enroll button. You should be directed to Cybersource. The total being requested should be equal to the price in `CoursePrice` minus the discount amount.
 - Go into the Django shell and verify that an `Order` is created for this course run, that it has `total_price_paid` equal to the discounted price, and that there is one `Line` in the `Order` whose `price` is also equal to the discounted price.

Then, test that negative or zero prices are rejected:
- Delete the `Order` that was created in the Django shell.
 - Set `CoursePrice.price` to a value lower or equal to the discount amount
 - Click the enroll button again. You should get a 500 error, and an `ImproperlyConfigured` exception in the Django console

7ktihtl

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -21,6 +21,7 @@ from rest_framework.exceptions import ValidationError
 from backends.edxorg import EdxOrgOAuth2
 from courses.models import CourseRun
 from dashboard.api import update_cached_enrollment
+from dashboard.models import ProgramEnrollment
 from ecommerce.exceptions import (
     EcommerceEdxApiException,
     EcommerceException,
@@ -30,7 +31,7 @@ from ecommerce.models import (
     Line,
     Order,
 )
-from financialaid.api import get_course_price_for_learner
+from financialaid.api import get_formatted_course_price
 from financialaid.models import (
     FinancialAid,
     FinancialAidStatus,
@@ -108,8 +109,8 @@ def create_unfulfilled_order(course_id, user):
         Order: A newly created Order for the CourseRun with the given course_id
     """
     course_run = get_purchasable_course_run(course_id, user)
-
-    price_dict = get_course_price_for_learner(user, course_run.course.program)
+    enrollment = get_object_or_404(ProgramEnrollment, program=course_run.course.program, user=user)
+    price_dict = get_formatted_course_price(enrollment)
     price = price_dict['course_price']
     if price <= 0:
         log.warning(
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -10,6 +10,7 @@ from urllib.parse import quote_plus
 import uuid
 
 from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 from django.db import transaction
 from django.http.response import Http404
 from django.shortcuts import get_object_or_404
@@ -29,6 +30,7 @@ from ecommerce.models import (
     Line,
     Order,
 )
+from financialaid.api import get_course_price_for_learner
 from profiles.api import get_social_username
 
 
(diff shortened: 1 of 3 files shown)


ecommerce/api.py
```diff
@@ -88,7 +93,18 @@ def create_unfulfilled_order(course_id, user):
         Order: A newly created Order for the CourseRun with the given course_id
     """
     course_run = get_purchasable_course_run(course_id, user)
-    price = course_run.courseprice_set.get(is_valid=True).price
+
+    price_dict = get_course_price_for_learner(user, course_run.course.program)
+    price = price_dict['course_price']
```
giocalitri: Question: can you be sure of the existence of the `'course_price'` key?

ecommerce/api_test.py
```diff
@@ -108,28 +129,61 @@ def test_already_purchased(self):
...
         assert ex.exception.args[0] == 'Course run {} is already purchased'.format(course_run.edx_course_key)
 
+    def test_less_or_equal_to_zero(self):
+        """
+        An order may not have a negative or zero price
+        """
+        course_run, user = create_purchasable_course_run()
+        price_obj = course_run.courseprice_set.get(is_valid=True)
+
+        for invalid_price in (0, -1.23):
```
giocalitri: Nit: being a tuple, can you please write it as `(0, -1.23,)`

giocalitri:
Code looks good. Checking functionality

giocalitri:
Not sure it is a bug related to this PR, but if I do not have a `FinancialAid` object at all for the use in the program, I can pay automatically the full price.

ecommerce/api.py
noisecapella: Yes, the function has only one return statement and the key is defined in the dict there

noisecapella:
We should fix that, I'll change the `get_course_price_for_learner` function to raise a `ValidationError` in that case. Our workflow should create a `FinancialAid` with a skipped status somehow before this API is called

noisecapella:
Now that I think about it the check should be done in the checkout API, not in `get_course_price_for_learner`

noisecapella:
Ok, the user should now not be able to purchase without at least one `FinancialAid` attached

giocalitri:
I think that the same problem now exists if the FinancialAid object has a status that should not allow a purchase.
For example I do not expect that a FinancialAid with status `created` should me allow to make a payment.

noisecapella:
That makes sense, I'll restrict it to the three final states

noisecapella:
I also filtered on the user which I missed earlier. Should be ready for review again

ecommerce/api.py
```diff
@@ -55,12 +64,24 @@ def get_purchasable_course_run(course_key, user):
...
 
+    if not FinancialAid.objects.filter(
+            tier_program__current=True,
+            tier_program__program__course__courserun=course_run,
+            user=user,
+            status__in=FinancialAidStatus.TERMINAL_STATUSES,
+    ).exists():
+        log.warning("Course run %s has no attached financial aid for user %s", course_key, get_social_username(user))
+        raise ValidationError(
+            "Course run {} does not have a current attached financial aid application".format(course_key)
```
giocalitri: nit: I could specify "valid" here, considering that there could be one in a different state, but feel free to leave as is

giocalitri:
👍

giocalitri:
I tested a discount equal to the maximum price and the the backend correctly raises `ImproperlyConfigured`.
This translates to a 500 error. Should we address this translating the `ImproperlyConfigured` to a proper JSON answer instead of an HTML?
screen shot 2016-09-30 at 11 54 52 am

giocalitri:
In the meantime I confirm the 👍

Issue #1098, opened 2016-09-22
Financial Aid UI: Documents required for verification


pdpinch:
Some users will need to send documents to verify income. They will receive a message in their dashboard and a short form to record when they sent their documents. 

![image](https://cloud.githubusercontent.com/assets/430126/18765914/b732589c-80e5-11e6-9d10-4d2d1fb4d5eb.png)

## Acceptance criteria

- [x] use API in #1053 to determine whether docs are required
- [x] form for submitting date docs were sent
- [ ] after date is submitted, update message:

![image](https://cloud.githubusercontent.com/assets/430126/18765997/12c1ec72-80e6-11e6-851c-1e473f0c6e0b.png)

alvinsiu:
I don't think we have a specific API endpoint for determining if docs are required yet, but we can build an endpoint for that.

We also don't have a way of explicitly tracking the date documents were mailed (did you mean received?).  We can retrieve it via the auditing trail, or naively by using the "updated_on" timestamp when the status is changed from "pending-docs" to "pending-manual-approval", but neither of those are the most robust.  Should we add a field explicitly for tracking when documents are mailed?

pdpinch:
For docs required, I'm not sure how you wanted to message that to the front-end. It could just be the return from the calculator submission, but then there's the question of repeat visits from the user. Can you discuss with @aliceriot ?

The date is a new feature. I think adding a new field makes sense. You'll see that it appears in the financial aid review view as well.

alvinsiu:
Docs required should probably be a separate endpoint, that way we can still know when to display the message if they log back in another time separate from the calculator submission.  @aliceriot If we return the `FinancialAid` object so that you can see if `object.status == "pending-docs"`, would that work?

aliceriot:
@alvinsiu yea I think that sounds straightforward enough, I've been working on something else and haven't dug into this yet so I can't say for sure though.

alvinsiu:
Great, we'll implement that API for you, and you can let us know if you need any additional information or any changes to what's returned!

aliceriot:
OK reading over this it would make sense to me if the docs endpoint returned something like:

```js
{
  application_status: 'docs-required-pending' || 'docs-required-received', // etc
  docs_mailing_date: 'YYYY-MM-DD',
}
```

@pdpinch I haven't been so up on the financial aid stuff, does this require that we have the financial aid calculator done?

aliceriot:
just wondering about which one to take on first

[...]

PR #1162, opened 2016-09-28
Implemented API endpoint for submitting date documents were sent

alvinsiu:
#1098 

#### What's this PR do?
Adds API endpoint for learner to submit date documents were sent.


gsidebo:
@alvinsiu im gonna have a PR up for the course price front end work soon. i'll get to this right after that's up

alvinsiu:
No problem, thanks!

financialaid/serializers.py
```diff
@@ -109,7 +109,7 @@ def validate(self, data):
                 self.instance.status != FinancialAidStatus.PENDING_MANUAL_APPROVAL):
             raise ValidationError("Cannot approve application that is not pending manual approval.")
         if (data['action'] == FinancialAidStatus.PENDING_MANUAL_APPROVAL and
-                self.instance.status != FinancialAidStatus.PENDING_DOCS):
+                self.instance.status not in [FinancialAidStatus.PENDING_DOCS, FinancialAidStatus.DOCS_SENT]):
             raise ValidationError("Cannot mark documents as received for application not pending docs.")
```
gsidebo: it looks like this validation message needs to change, and tbh the language is kind of obtuse as it is right now

micromasters/urls.py
```diff
@@ -52,6 +53,8 @@
         name='financial_aid_action'),
     url(r'^api/v0/financial_aid_course_price/(?P<program_id>[\d]+)/$',
         GetLearnerPriceForCourseView.as_view(), name='financial_aid_course_price'),
+    url(r'^api/v0/financial_aid_documents_sent/(?P<financial_aid_id>[\d]+)/$',
+        DocumentsSentView.as_view(), name='financial_aid_documents_sent'),
```
gsidebo: general question: this 'documents sent' request is basically being used to update a value for a FinancialAid record, right? if so, i feel like it makes more sense to add a detail view for the `FinancialAid` and send these requests to update the doc sent date as a `PUT` request

financialaid/serializers.py
alvinsiu: Since you're much more familiar with the front end and what users might see, if you have better messages, we're happy to put them in!

micromasters/urls.py
alvinsiu: Yep, that's correct.  The `UpdateAPIView` implements these requests as a `PUT` request, so it should behave the way you're describing already, unless I'm misunderstanding?

financialaid/serializers.py
alvinsiu: I see - what about adding "or docs sent" at the end of the error message?  Or a more generic "Cannot mark documents as received for an application not awaiting documents"?

micromasters/urls.py
gsidebo: in short, i'm asking why the url isn't `api/v0/financial_aid/(?P<financial_aid_id>[\d]+)/$`, and the view isn't `FinancialAidDetailView` or something like that. adding a brand new url path and view for every action we want to perform is not a good pattern. if we're at least considering REST principles, we should be thinking about this in terms of the resource, which is `FinancialAid`. in essence, we're adding this functionality to be able to update (ie: `PUT`) a `FinancialAid` record, identified by id. do we need to follow REST principles exactly? no, but this seems like a pretty simple case to me. with that said, if there's not (a) a clear and easy path forward to implement what i'm describing, nor (b) a future need to update `FinancialAid` records via API, i'm ok with leaving this as-is due to time constraints.

alvinsiu: Ah, I see what your concern is.  The reason this is a separate endpoint is because of permissions.  We need to have an endpoint that is restricted to staff who can see and edit financial aid applications (`FinancialAidActionView`), an endpoint for users to create an application (`FinancialAidRequestView`), an endpoint that's only for users to edit their own applications (`DocumentsSentView`), and so on.

So we can certainly change the name of `DocumentsSentView` to `FinancialAidDetailView` and make it a `RetrieveUpdateAPIView`, but to my knowledge, there isn't a need for the `.get()` view right now, since that's all handled by the dashboard API.

We could also group it with the `FinancialAidRequestView`, but the `DocumentsSentView` will require a similar, but separate, url anyway since it requires `financial_aid_id` in the URL.

We're happy to restructure it differently if you think they should be, what do you think?

financialaid/serializers.py
gsidebo: the second one 😄

alvinsiu: Great!  Updated :]

micromasters/urls.py
gsidebo: do you see anything wrong with just changing these names to be more general? `DocumentsSentView` -> `FinancialAidDetailView`, `DocumentsSentSerializer` -> `FinancialAidSerializer`, `api/v0/financial_aid_documents_sent/(?P<financial_aid_id>[\d]+)/` -> `api/v0/financial_aid/(?P<financial_aid_id>[\d]+)/`. one idea is that you could expect the desired status to be passed to the endpoint in the request body in addition to `date_documents_sent`. honestly, if this sounds time-consuming, just say so and we can drop it. it's just a little irksome that the pattern right now is to add a brand new, specifically-named view for every type of action we want to perform via API, and that pattern is going to continue to be followed

alvinsiu: > one idea is that you could expect the desired status to be passed to the endpoint in the request body in addition to 'date_documents_sent'
Is "desired status" referring to the actions in currently in `FinancialAidActionView` (like approving an application)?  If so, that won't work because the permissions are different between those and `date_documents_sent`

For changing the names, the functionality is currently restricted to only updating `date_documents_sent`, it feels slightly misleading for the name to not indicate that.  But I don't have a strong opinion on that, so I'll go ahead and change those names as you requested.

financialaid/permissions.py
```diff
@@ -23,3 +23,20 @@ def has_object_permission(self, request, view, obj):
             boolean
         """
         return has_object_permission(Permissions.CAN_EDIT_FINANCIAL_AID, request.user, obj.tier_program.program)
+
+
+class UserCanIndicateDocumentsSent(BasePermission):
```
gsidebo: please rename to `RequestUserMatchesLoggedInUser` or something like that since we're 'generalizing' this functionality, and update docstrings

financialaid/views_test.py
```diff
@@ -511,3 +539,67 @@ def test_get_learner_price_for_enrolled_but_no_financial_aid_availability(self):
...
+            kwargs={"financial_aid_id": cls.financialaid_pending_docs.id}
+        )
+        cls.data = {
+            "financial_aid_id": cls.financialaid_pending_docs.id,
+            "date_documents_sent": datetime.datetime(2016, 9, 25).strftime("%Y-%m-%d")
+        }
+
+    def test_learner_can_indicate_documents_sent(self):
+        """
+        Tests DocumentSentView for user editing their own financial aid document status
```
gsidebo: please update this docstring and the 2 below this one

financialaid/views.py
```diff
@@ -224,6 +228,18 @@ class FinancialAidActionView(UpdateAPIView):
     queryset = FinancialAid.objects.all()
 
 
+class FinancialAidDetailView(UpdateAPIView):
+    """
+    View for learners to indicate that they have sent financial aid documents
```
gsidebo: please update this docstring, eg: "View for updating a FinancialAid record"

gsidebo:
@alvinsiu needs a rebase. after that's done and tests are passing, looks good 👍

alvinsiu:
Great!  Thanks for your help!

giocalitri:
How close are we to merge this PR? Need it for other work.

alvinsiu:
It's merged!

Issue #1153, opened 2016-09-28
Home: Style mailchimp form


pdpinch:
Should look like this:

![image](https://cloud.githubusercontent.com/assets/430126/18921899/9914ef62-8574-11e6-9ab3-fb1010daac34.png)

Not like this:

![image](https://box.everhelper.me/attachment/592280/294cebcd-2b1c-407e-886d-de8090c6ecb0/623857-EVN9fugAdiUZu3kL/screen.png)

pdpinch:
Check with @pdpdinch before working on this, to see if there have been any changes to the form fields.

PR #1166, opened 2016-09-29
Fixed Style of mailchimp form

amir-qayyum-khan:
- Fixes https://github.com/mitodl/micromasters/issues/1153

#### What's this PR do?
- Fixed Style of mailchimp form in dialog

- home_page.html/ mailchimp dialog

- see home_page.html/ mailchimp dialog

@pdpinch @aliceriot 

screen shot 2016-09-29 at 5 43 50 pm

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -235,72 +235,58 @@
     <!-- Modal content-->
     <div class="modal-content">
       <div class="modal-header">
-        <button type="button" class="close" data-dismiss="modal">&times;</button>
-        <h4 class="modal-title">Sign up to be notified when new programs are announced.</h4>
+        <h6 class="modal-title text-center">Sign up to be notified when new programs are announced.</h6>
       </div>
       <div class="modal-body">
         <div class="container">
             <form class="form-horizontal subscription-form mailchimp">
                 <!-- SUBSCRIPTION SUCCESSFUL OR ERROR MESSAGES -->
-                <h4 class="col-sm-offset-4 subscription-result success"></h4>
-                <h4 class="col-sm-offset-4 subscription-result error"></h4>
+                <h4 class="subscription-result success"></h4>
+                <h4 class="subscription-result error"></h4>
 
-                <div class="form-group">
-                    <label class="col-sm-4 control-label" for="inputname">Name</label>
-                    <div class="col-sm-8">
-                        <input type="text" class="form-control" id="inputname" name="FNAME">
-                    </div>
+                <div class="form-group col-sm-12">
+                  <input type="text" class="form-control flat-field" id="inputname" name="FNAME" placeholder="Name" />
                 </div>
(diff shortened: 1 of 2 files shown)


pdpinch:
@roberthouse54 can you take a quick look at this?

I don't think it's worth sweating over too much, as I'm still waiting on a new form. But if the styling looks good, I'd like to merge it.

roberthouse54:
Seems you are using different styles/classes for buttons, form fields, and also the modal header. 

Buttons: Please use the same button styles that we have elsewhere (Example: look on the settings page). 

Form Fields: Use the form fields styles that we have elsewhere (example: in the Sign Up). 

Modal Header: Use the header style for the modal that we use on the profile page... clicking "Add a degree, etc).

amir-qayyum-khan:
@roberthouse54 these  ``Sign Up`` fields and model are part if react-mdl lib.
They dont have custom scss. Though found classes for buttons
@pdpinch updated text.

screen shot 2016-10-03 at 4 05 34 pm

roberthouse54:
Can you check with another engineer on this... These are different form fields than we are using in other places, and I just want consistency.

aliceriot:
The form fields are different than what we're using elsewhere. Unfortunately, though, the form fields we're using elsewhere are implemented in JavaScript (they're party of [material-ui](http://www.material-ui.com/)) so we can't use them here directly. Are we ok with what they look like? If not, we could try to get the standard material-design-light text fields working in plain HTML: https://getmdl.io/components/index.html#textfields-section

cms/templates/cms/home_page.html
```diff
@@ -228,72 +228,58 @@ <h2 class="quote">"Micromasters is an important project for me. I believe in the
...
+                    <label><input type="radio" name="MMERGE2" value="I am an interested learner" id="learner">
+                        I am an interested learner
+                    </label>
+                  </div>
+                  <div class="radio">
+                    <label><input type="radio" name="MMERGE2" value="I represent an interested corporation" id="corporation">
+                        I represent an interested corporation
+                    </label>
+                  </div>
+
```
aliceriot: could you delete this blank line?

amir-qayyum-khan: fixed this

amir-qayyum-khan:
@aliceriot do i need to import css from the link that you sent?
I think this style is implement from link as well as javascript code

Ferdi:
I suggest we leave the markup that Amir has currently in the PR, even if it's different from what we use elsewhere.

pdpinch:
@aliceriot if you accept Amir's latest change, I think we can merge this.

aliceriot:
My comment above is more of a question, how much do we care that the mailchimp form doesn't really look like the rest of the site that much? If we think it's fine then this PR is ready to merge, if not it needs to be changed. @roberthouse54 ?

roberthouse54:
Let's just merge it

PR #1169, opened 2016-09-29
Rh/responsive style tweaks

roberthouse54:
#### What's this PR do?
This PR fixes most of the css layout issues for responsive devices... mainly mobile phones, and large monitors. I have ignored tablet views mostly.

Makes the top image on the home page expand to the full browser height. Fixes all layout issues for phones. Makes the faculty carousel just a vertical layout, etc.

I am using "vh" and "vw" units on the home page. These are supported in most modern browsers, but let me know if you see any issues, bugs etc.


roberthouse54:
I don't think the "The" is necessary.

giocalitri:
I think the buttons are not of the same size.
screen shot 2016-09-30 at 3 28 34 pm

giocalitri:
this program on my machine seems to have an active link in the background
screen shot 2016-09-30 at 3 32 14 pm

giocalitri:
Looks mostly fine to me

roberthouse54:
For your second comment, I do not think that is an active link in the background. I think it is just how this is displaying when there is a program defined, but there is no thumbnail image.

roberthouse54:
OK, so this second issue was that the thumbnail image had been deleted, but the CMS was still set to look for that image. There is a default image to show when no image has been set. So I think we should just ignore this as an edge case. It could happen, but it shouldn't happen.

roberthouse54:
This should be all set for review. The only thing that is not working is that the tabs should be scrollable tabs (see Material design scrollable tabs). I could not figure this out, but I think we can address this in a separate issue, or someone else can do it.

giocalitri:
Please open a new issue for the tabs. For the rest 👍

roberthouse54:
Scrollable tabs: https://github.com/mitodl/micromasters/issues/1203

PR #1171, opened 2016-09-29
Added course price API call to front end

### LINKED ISSUE #1151: Fetch course prices in front end
We need to call the course price endpoint (merged in #1099) in the MM front end to make those prices available in our UI. This work will support several issues where course prices need to be displayed (eg: #1102 #1097)

gsidebo:

Closes #1151 

#### What's this PR do?

- Adds a new API view to get course prices for _all_ of a user's enrolled programs
- Adds a call to that API at the app level on the front end; this puts course price info in the application state (a la dashboard data)

Probably static/js/containers/App.js and financialaid/views.py

1. Remove staff permissions for 1 or more programs for your user
1. Make sure you have relevant course price information in your data (I did this simply by adding a relevant `CoursePrice` record)
1. Run the app and check your network tab to make sure the `course_prices` endpoint is being called and the response contains relevant data

Note that this data isn't being used anywhere on the front end yet. That's for later issues

### DIFF
--- a/financialaid/views.py
+++ b/financialaid/views.py
@@ -19,7 +19,8 @@ from rest_framework.views import APIView
 from rolepermissions.verifications import has_object_permission
 
 from courses.models import Program
-from financialaid.api import get_course_price_for_learner
+from dashboard.models import ProgramEnrollment
+from financialaid.api import get_formatted_course_price
 from financialaid.models import (
     FinancialAid,
     FinancialAidStatus,
@@ -224,23 +225,48 @@ class FinancialAidActionView(UpdateAPIView):
     queryset = FinancialAid.objects.all()
 
 
-class GetLearnerPriceForCourseView(APIView):
+class CoursePriceListView(APIView):
     """
-    View for retrieving a leaner's price for a course run
+    View for retrieving a learner's price for course runs in all enrolled programs
+    """
+    authentication_classes = (SessionAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, *args, **kwargs):
+        """
+        GET handler
+        """
+        user = request.user
+        program_enrollments = (
+            ProgramEnrollment.objects
+            .select_related('user', 'program')
+            .filter(user=user, program__live=True).all()
+        )
+        formatted_course_prices = []
+        for program_enrollment in program_enrollments:
+            response_dict = get_formatted_course_price(program_enrollment)
+            formatted_course_prices.append(response_dict)
+        return Response(data=formatted_course_prices)
+
+
+class CoursePriceDetailView(APIView):
+    """
+    View for retrieving a learner's price for a course run
     """
     authentication_classes = (SessionAuthentication, )
     permission_classes = (IsAuthenticated, )
 
     def get(self, request, *args, **kwargs):
         """
-        Get request for GetLearnerPriceForCourseView
+        GET handler
         """
-        learner = request.user
-        program = get_object_or_404(
-            Program,
-            id=self.kwargs["program_id"],
-            live=True
+        user = request.user
+        program_enrollment = get_object_or_404(
+            ProgramEnrollment,
+            user=user,
+            program__id=self.kwargs["program_id"],
+            program__live=True
         )
         return Response(
-            data=get_course_price_for_learner(learner, program)
+            data=get_formatted_course_price(program_enrollment)
         )
--- a/micromasters/urls.py
+++ b/micromasters/urls.py
@@ -21,7 +21,8 @@ from ecommerce.views import (
 from financialaid.views import (
     FinancialAidRequestView,
     FinancialAidActionView,
-    GetLearnerPriceForCourseView
+    CoursePriceListView,
+    CoursePriceDetailView
 )
 from profiles.views import ProfileViewSet
 from search.views import ElasticProxyView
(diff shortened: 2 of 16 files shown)


static/js/constants.js
```diff
@@ -5,6 +5,49 @@ export const BACHELORS = 'b';
 export const MASTERS = 'm';
 export const DOCTORATE = 'p';
```
gsidebo: i moved these values to separate them from the test data constants

financialaid/views.py
```diff
@@ -224,23 +227,56 @@ class FinancialAidActionView(UpdateAPIView):
...
+    """
+    authentication_classes = (SessionAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, *args, **kwargs):
+        """
+        GET handler
+        """
+        user = request.user
+        program_enrollments = ProgramEnrollment.objects\
```
noisecapella: Can you put parentheses around the expression here instead of using `\` to allow multiple lines?

financialaid/views.py
```diff
@@ -224,23 +227,56 @@ class FinancialAidActionView(UpdateAPIView):
...
+            .select_related('user', 'program')\
+            .filter(user=user, program__live=True).all()
+        formatted_course_prices = []
+        for program_enrollment in program_enrollments:
+            try:
+                response_dict = get_formatted_course_price(program_enrollment)
+                response_dict['program_id'] = program_enrollment.program.id
+                formatted_course_prices.append(response_dict)
+            except ImproperlyConfigured:
```
noisecapella: I'm not sure we want to silence this exception here. If `get_formatted_course_price` doesn't have a valid course price to choose from then something's wrong with our configuration. I think it would be better to have the exception cause a 500 error here

financialaid/views.py
```diff
@@ -224,23 +227,56 @@ class FinancialAidActionView(UpdateAPIView):
...
+        """
+        user = request.user
+        program_enrollments = ProgramEnrollment.objects\
+            .select_related('user', 'program')\
+            .filter(user=user, program__live=True).all()
+        formatted_course_prices = []
+        for program_enrollment in program_enrollments:
+            try:
+                response_dict = get_formatted_course_price(program_enrollment)
+                response_dict['program_id'] = program_enrollment.program.id
```
noisecapella: Can we add `program_id` to `get_formatted_course_price`? It would simplify implementation and testing for this view

financialaid/views.py
```diff
@@ -224,23 +227,56 @@ class FinancialAidActionView(UpdateAPIView):
...
+            ProgramEnrollment,
+            user=user,
+            program__id=self.kwargs["program_id"],
+            program__live=True
         )
+        try:
+            return Response(
+                data=get_formatted_course_price(program_enrollment)
+            )
+        except ImproperlyConfigured:
```
noisecapella: Same comment here about silencing the exception. All live programs should have a valid price

micromasters/urls.py
```diff
@@ -50,8 +51,9 @@
     url(r'^api/v0/financial_aid_request/$', FinancialAidRequestView.as_view(), name='financial_aid_request'),
     url(r'^api/v0/financial_aid_action/(?P<financial_aid_id>[\d]+)/$', FinancialAidActionView.as_view(),
         name='financial_aid_action'),
-    url(r'^api/v0/financial_aid_course_price/(?P<program_id>[\d]+)/$',
-        GetLearnerPriceForCourseView.as_view(), name='financial_aid_course_price'),
+    url(r'^api/v0/course_prices/$', CoursePriceListView.as_view(), name='course_price_api'),
+    url(r'^api/v0/course_prices/(?P<program_id>[\d]+)/$',
+        CoursePriceDetailView.as_view(), name='course_price_api'),
```
noisecapella: The convention for DRF views is to append a `-detail` to the detail URL name and a `-list` to the list name. Can you do that here as well?

static/js/util/integration_test_helper.js
```diff
@@ -68,12 +72,15 @@ class IntegrationTestHelper {
   renderComponent(url = "/", extraTypesToAssert = [], isSuccessExpected = true) {
     let expectedTypes = [
       REQUEST_DASHBOARD,
+      REQUEST_COURSE_PRICES,
       REQUEST_GET_USER_PROFILE,
       REQUEST_GET_PROGRAM_ENROLLMENTS,
     ];
+      // RECEIVE_COURSE_PRICES_SUCCESS,
```
noisecapella: Remove this comment please

noisecapella:
Overall everything looks very straightforward. Testing functionality now...

noisecapella:
Functionality looks great

financialaid/views.py
gsidebo: here's my problem with that: if we dont handle the exception and we let the server throw a 500, the response that makes it back to the user will be a stack trace. do we have anything sensitive in there? don't know, really, but i'd rather handle the exception and return our own error code

noisecapella: No, in the production servers the user will just get a `500 server error` message saying that 'an engineer will be blamed for this.` They will not see a stack trace. See https://github.com/mitodl/micromasters/blob/master/ui/templates/500.html

financialaid/views.py
gsidebo: i'll need to be convinced of that. with the current code, a user calls the /course_prices endpoint to get a list of prices for all of their enrolled programs. if their enrolled programs don't have a course price set up, then i think it makes sense for that call to return an empty list. also worth noting is the fact that this would cause a ton of headaches for anyone running the app locally without the right data.

noisecapella: Returning an empty list isn't really more helpful for the user. If we do get that exception it is something we need to deal with immediately, else the financial aid and ecommerce parts of the app will fail when the user makes use of them. By raising the `ImproperlyConfigured` exception we are saying that the configuration is broken and needs to be repaired by the administrator, the application is not able to be useful to the user (especially not the course list API part of it)

gsidebo:
@noisecapella ?

financialaid/views.py
gsidebo: ok fair enough. the only advantages i see would be very edgy edge cases, so i'll remove this

noisecapella:
Looks good after fixing lint error :+1:

Issue #1111, opened 2016-09-24
MM home page: Improvements


Ferdi:
Finalize the home page

Todos:
- [x] #1140 copy changes 
- [x] Finalize design/styling
- [x] #1200 fixup metatags per https://github.com/joshbuchea/HEAD
  - [x] Favicons
  - [x] split overloaded `<meta>` tag into two (charset first). See https://github.com/mitodl/micromasters/blob/master/ui/templates/base.html#L4
  - [x] Google+ / Schema.org tags  -- see https://github.com/joshbuchea/HEAD#google--schemaorg
- [x] #1173 Introduce social icons to share
- [x] Finalize imagery  
- [x] Make it responsive
- [x] #1153 Style the mailing list sign up popup
- [x] ~~#1040 add a short description field on the ProgramPage and display it in the home page~~
- [X] #1270 Remove zendesk help button from home page
- [x] #1290 Programs dont always show in the same order

pdpinch:
@annagav @singingwolfboy I added a few more details that need to be addressed in the "SEO" section.

pdpinch:
Regarding favicons, we've already implemented the only size that is approved by MIT. See http://mit.edu/graphicidentity/download-logos.html

pdpinch:
Everything on this this list is done. If we need more home page improvements, new issues should be opened.

PR #1173, opened 2016-09-29
Add Social Media Icons

annagav:
Related to #1111
#### What's this PR do?
Adding Social Media Icons.

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -9,6 +9,7 @@
     <link rel="shortcut icon" href="{% static 'images/favicon.ico' %}" />
     <script type="text/javascript">
     var SETTINGS = {{ js_settings_json|safe }};
+    var CURRENT_PAGE_URL = '{{ request.build_absolute_uri }}';
     </script>
     {% if style_public_src %}
     <script type="text/javascript" src="{{ style_public_src }}"></script>
--- a/ui/templates/footer.html
+++ b/ui/templates/footer.html
@@ -1,25 +1,72 @@
 {% load static %}
 <footer>
-     <div class="container">
-        <div class="row">
-           <div class="col-md-9">
-              <img src="{% static 'images/mit-logo-ltgray-white@72x38.svg' %}" alt="MIT" width="72" height="38">
-              <div class="footer-links">
-                  <a href="https://www.edx.org/" target="_blank">edX</a>
-                  <a href="https://odl.mit.edu/" target="_blank">Office of Digital Learning</a>
-              </div>
-              <div class="footer-address">
-                 Massachusetts Institute of Technology<br/> Cambridge, MA 02139
-              </div>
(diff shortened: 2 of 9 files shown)


pdpinch:
You've got some linting errors in the scss. See https://travis-ci.org/mitodl/micromasters/builds/163855027#L3357-L3366

Can I help with the missing URLs?

Should I be able to test this by actually doing a fb/twitter/etc. post?

annagav:
screen shot 2016-09-30 at 3 15 22 pm

annagav:
screen shot 2016-09-30 at 3 14 53 pm
screen shot 2016-09-30 at 3 15 11 pm

annagav:
screen shot 2016-09-30 at 3 46 35 pm

pdpinch:
That's more like what I was expecting. I can take do a review later this afternoon.

annagav:
Having some trouble with facebook. George thinks it is a caching problem.

screen shot 2016-09-30 at 5 32 28 pm

Ferdi:
I wouldn't worry much about it. It looks like you have the right fb share
URL. If it doesn't look right,  it's probably out of the scope of this
issue

- Sent from my phone

pdpinch:
@singingwolfboy can you review this?

ui/templates/footer.html
```diff
@@ -1,25 +1,75 @@
...
+            </li>
+
+          </ul>
+        </div>
+        <div class="footer-copy">&copy; 2016, Massachusetts Institute of Technology</div>
+      </div>
+    </div>
+  </div>
+  <script type="text/javascript">
+    var url='{{ request.build_absolute_uri }}';
```
singingwolfboy: Are you setting this `url` variable so that `public.js` can reference it? If so, I recommend two things:

* Move this to `templates/base.html`, where the `SETTINGS` variable is defined. We should have as few places as possible where global variables are set; ideally, that should be the *only* place.
* Pick a more descriptive name than `url`. That variable name is just begging to be overwritten by some other temporary `url` variable in some other script. I would recommend `currentPageUrl` or `CURRENT_PAGE_URL`.

annagav: That's right. I moved it.

ui/templates/base.html
```diff
@@ -9,6 +9,7 @@
     <link rel="shortcut icon" href="{% static 'images/favicon.ico' %}" />
     <script type="text/javascript">
     var SETTINGS = {{ js_settings_json|safe }};
+    var CURRENT_PAGE_URL='{{ request.build_absolute_uri }}';
```
singingwolfboy: Can you put spaces around the `=`?
```js
var CURRENT_PAGE_URL = '{{ request.build_absolute_uri }}';
```

annagav: oops, did not notice :)

PR #1175, opened 2016-09-29
Implemented API for learners to skip financial aid and pay full price

### LINKED ISSUE #1076: Financial Aid: Allow learners to bypass financial aid & pay full price
Some learners will want to skip the financial aid process. At any stage, we'll allow them to shortcut the process and be approved to pay full price. 

This function will link from 3 locations:

1. Initial Financial Aid description widget: 

  image

2. Cost Calculator:

  image

3. The Verify Documents widget

image

Since there's a lot of money at stake, choosing this opens a confirmation dialog:

image

## Acceptance Criteria

- [x] API and modifications to financial aid status to track that the user has chosen to pay full price
- [x] UI 1) above
- [x] UI 2) above
- [x] UI 3) above

alvinsiu:
#1076 

#### What's this PR do?
Implements the API endpoint for when a learner clicks on "Skip this and Pay Full Price".  It marks their `FinancialAid.status` as `FinancialAidStatus.SKIPPED` (creating a `FinancialAid` object if it hasn't already been created) and logs this action for audit.


giocalitri:
Can you please resolve the conflicts first?

financialaid/models.py
```diff
@@ -105,8 +105,17 @@ class FinancialAidStatus:
...
+    ALL_STATUSES = [
+        CREATED,
+        APPROVED,
+        AUTO_APPROVED,
+        REJECTED,
+        PENDING_DOCS,
+        PENDING_MANUAL_APPROVAL,
+        SKIPPED
+    ]
     TERMINAL_STATUSES = [APPROVED, AUTO_APPROVED, REJECTED]
```
giocalitri: Please update this list (it has been introduced by a recent PR

financialaid/serializers.py
```diff
@@ -90,6 +90,33 @@ def save(self):
...
+    """
+    def validate(self, data):
+        """
+        Validators for this serializer
+        """
+        unpermitted_statuses = [
+            FinancialAidStatus.APPROVED,
+            FinancialAidStatus.AUTO_APPROVED,
+            FinancialAidStatus.REJECTED
+        ]
```
giocalitri: I think you can used directly `TERMINAL_STATUSES`

financialaid/views.py
```diff
@@ -51,6 +56,35 @@ def get_queryset(self):  # pragma: no cover
...
+    """
+    View for financial aid skip API. Takes user and program, then determines whether a financial
+    aid object exists, and then either creates or updates a financial aid object to reflect
+    the user skipping financial aid.
+    """
+    serializer_class = FinancialAidSkipSerializer
+    authentication_classes = (SessionAuthentication, )
+    permission_classes = (IsAuthenticated, )
+
+    def get_object(self):  # pragma: no cover
```
giocalitri: why the ` # pragma: no cover`?

financialaid/views.py
```diff
@@ -51,6 +56,35 @@ def get_queryset(self):  # pragma: no cover
...
+        """
+        Overrides get_object in case financialaid object does not exist, as the learner may skip
+        financial aid either after starting the process or in lieu of applying
+        """
+        program = get_object_or_404(Program, id=self.kwargs["program_id"])
+        if not program.financial_aid_availability:
+            raise ValidationError("Financial aid not available for this program.")
+        if not ProgramEnrollment.objects.filter(program=program.id, user=self.request.user).exists():
+            raise ValidationError("User not in program.")
+        tier_program = get_no_discount_tier_program(program.id)
```
giocalitri: `get_no_discount_tier_program` can raise `DoesNotExist`.  We should wrap that call inside `get_no_discount_tier_program` to return a `ImproperlyConfigured` exception in this case.

financialaid/views.py
```diff
@@ -51,6 +56,35 @@ def get_queryset(self):  # pragma: no cover
...
+        financialaid, _ = FinancialAid.objects.get_or_create(
+            user=self.request.user,
+            tier_program__program=program,
+            defaults={"tier_program": tier_program}
+        )
```
giocalitri: I do not think it is a good idea to put this in the `get_object` method.

Why don't you just move this to the `get` method?

giocalitri: sorry, this should support `PUT/POST`.... not sure how to do that, but still not convinced about modifying or creating the object inside the `get_object `

financialaid/views_test.py
```diff
@@ -548,6 +548,86 @@ def test_get_learner_price_for_enrolled_but_no_financial_aid_availability(self):
...
+        self.assert_http_status(self.client.put, self.skip_url, status.HTTP_400_BAD_REQUEST)
+
+    def test_financialaid_object_cannot_be_skipped_if_not_enrolled_in_program(self):
+        """
+        Tests that a FinancialAid object cannot be skipped if the user is not enrolled in program
+        """
+        self.client.force_login(self.enrolled_profile3.user)
+        url = reverse("financial_aid_skip", kwargs={"program_id": self.program2.id})
+        self.assert_http_status(self.client.put, url, status.HTTP_400_BAD_REQUEST)
+
```
giocalitri: Can you test that the other HTTP method are not supported? Does POST work with this view?

giocalitri:
The logic is in place: there are few changes I asked

financialaid/models.py
alvinsiu: Ah, I didn't know what that was for, but now I understand.

financialaid/serializers.py
alvinsiu: Makes sense

financialaid/views.py
alvinsiu: Oh, I think that probably got in there by accident.  Good catch!

financialaid/views.py
alvinsiu: Good idea!

financialaid/views_test.py
giocalitri: Can you also test requests for non existing programs and user not enrolled in the program

financialaid/views.py
alvinsiu: This was implemented this way because there isn't a guarantee that the learner has a `FinancialAid` object, otherwise we would have just left get_object() in place.  But since there is a chance we need to create a new object, we implemented it this way.  The `UpdateAPIView` implements automatically `PUT`, which calls `self.update()`, which calls `self.get_object()`, which is why we overrode the `get_object()` method.

If you can think of a better way, I'd love to hear it, otherwise I think it makes sense to put it here because it is still doing what it's meant to do (which is retrieve the `FinancialAid` object).

financialaid/views_test.py
alvinsiu: Sure!  We'll add those tests - and `POST` doesn't work by default with `UpdateAPIView`.

alvinsiu:
Still need to add tests - will add later and change label when that happens.

financialaid/views.py
giocalitri: what about overriding the `.update()` method of the view?

alvinsiu: If there's no `FinancialAid` object, we can't pass it into `.update()`, and if we pass in the `Program` to `.update()` to fetch/create the `FinancialAid` object, then `.update()` is misleading, since we're not updating the `Program` object.

giocalitri: The problem I have with this implementation is that `UpdateAPIView` implements PUT and PATCH.
PUT is fine because it is supposed to create objects other than just update them in full, but PATCH can be a partial update.
Now in this case `UpdateAPIView` is implemented as 
```
class UpdateAPIView(mixins.UpdateModelMixin,
                    GenericAPIView):

    """
    Concrete view for updating a model instance.
    """
    def put(self, request, *args, **kwargs):
        return self.update(request, *args, **kwargs)

    def patch(self, request, *args, **kwargs):
        return self.partial_update(request, *args, **kwargs)
```

and `mixins.UpdateModelMixin` is
```
class UpdateModelMixin(object):
    """
    Update a model instance.
    """
    def update(self, request, *args, **kwargs):
        partial = kwargs.pop('partial', False)
        instance = self.get_object()
        serializer = self.get_serializer(instance, data=request.data, partial=partial)
        serializer.is_valid(raise_exception=True)
        self.perform_update(serializer)
        return Response(serializer.data)

    def perform_update(self, serializer):
        serializer.save()

    def partial_update(self, request, *args, **kwargs):
        kwargs['partial'] = True
        return self.update(request, *args, **kwargs)
```
meaning that both call `UpdateModelMixin.update` for both full update/creation and partial update.

The problem is if I call for a partial update for an object that does not exist. Your implementation will create the object anyway and maybe it is the right thing to do, but it looks weird to me.

If you are positive this is the right way to go, I am fine with it.

financialaid/api.py
```diff
@@ -74,7 +75,10 @@ def get_no_discount_tier_program(program_id):
     Returns:
         TierProgram: the no discount TierProgram program associated with the Program
     """
-    return TierProgram.objects.get(program_id=program_id, current=True, discount_amount=0)
+    try:
+        return TierProgram.objects.get(program_id=program_id, current=True, discount_amount=0)
+    except TierProgram.DoesNotExist:
+        raise ImproperlyConfigured("No discount TierProgram has not yet been configured for this Program.")
```
giocalitri: Please add tests for this change.

financialaid/views.py
alvinsiu: Hmmm, I think this is one of those cases where we're implementing something sufficiently complex and non-standard that it won't really fit perfectly into REST.  I still think this is the right way to do it, given the intended functionality, but it's also likely to change in the future if there is going to be more functionality added to financial aid.

Why don't we leave it this way for now, and if either of us thinks of a better way to do this, we can bring it back up?  If you're still concerned about the partial update, we could also just override `patch()` and raise a 405/403/400?

financialaid/api.py
alvinsiu: Added tests!

financialaid/views.py
giocalitri: Let's leave it in this way.

financialaid/views_test.py
```diff
@@ -639,6 +641,109 @@ def test_get_learner_price_for_enrolled_but_no_financial_aid_availability(self):
...
+        """
+        Tests that an existing FinancialAid object that has already been approved or rejected
+        cannot be skipped.
+        """
+        self.client.force_login(self.enrolled_profile2.user)
+        unpermitted_statuses = [
+            FinancialAidStatus.APPROVED,
+            FinancialAidStatus.AUTO_APPROVED,
+            FinancialAidStatus.REJECTED
+        ]
```
giocalitri: I think this should include the skipped status, right?

giocalitri: and if so, please rename the test

giocalitri:
Another small comment

financialaid/views_test.py
alvinsiu: Hmmm, good point!

giocalitri:
👍 after tests pass

PR #1182, opened 2016-09-30
Add back terms of service link in dialog, and change Log in link to act the same as signup

### LINKED ISSUE #1113: Sign in / up workflow
Wrap up sign in/up workflow
- [x]  Remove MM dropdown and related code from the sign up popup (see #1105)
- [x]  #1105 Display program drop-down for users with no program 
- [x] #1182 : Add TOS to sign in/up popup
- [x]  #1231: After creating new profile, user gets a "No Program Enrollment is available" exception

noisecapella:
Fixes #1113

#### What's this PR do?
Changes the login link to act the same as sign up, and the dialog box will contain a link to the terms of service again

### DIFF
--- a/static/js/containers/SignupDialog.js
+++ b/static/js/containers/SignupDialog.js
@@ -36,7 +36,7 @@ const SignupDialog = ({
         <img src="/static/images/mit_logo_grey_red.png" />
       </div>
       <span>
-        The MIT MicroMasters program is powered by edX.
+        All MITx MicroMasters courses are delivered on edX.
         To sign up for a MIT MicroMasters program you need an edX account.
       </span>
 
--- a/static/js/containers/SignupDialog.js
+++ b/static/js/containers/SignupDialog.js
@@ -43,6 +43,14 @@ const SignupDialog = ({
       <a className="mm-button signup" href="/login/edxorg">
         Continue with edX
       </a>
+      <div className="terms-of-service-text">
+        By clicking "Continue with edX" I certify that I agree with <a href="/terms_of_service" target="_blank">
+          MIT MicroMasters Terms of Service.
+        </a> Read our <a
+          href="http://web.mit.edu/referencepubs/nondiscrimination/index.html"
+          target="_blank"
+        >Nondiscrimination Policy.</a>
+      </div>
     </div>
   </Dialog>;
 };
--- a/ui/templates/header.html
+++ b/ui/templates/header.html
@@ -12,7 +12,7 @@
             Sign Out
           </a>
           {% else %}
-          <a class="members-login btn ghost upper-case" href="{% url 'social:begin' 'edxorg' %}">Log In</a>
+          <a class="members-login btn ghost open-signup-dialog upper-case">Log In</a>
           <a class="btn btn-login white-button open-signup-dialog upper-case">
           Sign Up
           </a>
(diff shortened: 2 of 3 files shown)


pdpinch:
While you're in there, can you change the copy in this line: https://github.com/mitodl/micromasters/blob/950a82440ea9dde81ab5d51c9d047dcd7c8c904e/static/js/containers/SignupDialog.js#L39

```
        All MITx MicroMasters courses are delivered on edX.
```

"powered by edx" was misleading.

noisecapella:
Anything else?

pdpinch:
Do you know if we are still setting the "agreed to TOS" boolean? I was trying to test that in the PR deployment, but the server isn't configured properly.

I'll give it another go, but I thought it might be easier to ask.

noisecapella:
We currently set `agreed_to_terms_of_service` to `True` when the user creates the profile via edX

ui/templates/header.html
```diff
@@ -12,7 +12,7 @@
             Sign Out
           </a>
           {% else %}
-          <a class="members-login btn ghost upper-case" href="{% url 'social:begin' 'edxorg' %}">Log In</a>
+          <a class="members-login btn ghost open-signup-dialog upper-case">Log In</a>
           <a class="btn btn-login white-button open-signup-dialog upper-case">
```
singingwolfboy: This is probably outside of the scope of this button request, but for accessibility purposes, both the "Log In" and "Sign Up" buttons should use the `<button>` tag, not the `<a>` tag. The `<button>` tag is used to indicate some action that the user can take on the page, which does not immediately bring them to another page. The `<a>` tag is used to indicate a link to a different page on the internet. Since both "Log In" and "Sign Up" open a dialog rather than taking the user to another page, they should be `<button>`s.

Ferdi: Thanks @singingwolfboy do you mind to open a new issue ?

static/js/containers/SignupDialog.js
```diff
@@ -36,13 +36,21 @@ const SignupDialog = ({
...
-        The MIT MicroMasters program is powered by edX.
+        All MITx MicroMasters courses are delivered on edX.
         To sign up for a MIT MicroMasters program you need an edX account.
       </span>
 
       <a className="mm-button signup" href="/login/edxorg">
         Continue with edX
       </a>
+      <div className="terms-of-service-text">
+        By clicking "Continue with edX" I certify that I agree with <a href="/terms_of_service" target="_blank">
```
singingwolfboy: Wouldn't it be better to use `{% url 'terms_of_service' %}` instead of hard-coding `/terms_of_service`?

ui/templates/header.html
singingwolfboy: Created #1222

static/js/containers/SignupDialog.js
singingwolfboy: Oh wait, this is a Javascript file, not a Django template. I guess it's not worth doing the URL resolution into a Javascript variable and then referring to that variable here, so never mind.

noisecapella:
Any other comments @pdpinch?

pdpinch:
Has anyone tested the functionality? (sorry, I haven't yet been able to)

pdpinch:
It work! (something was wrong with my oauth settings, but they're fixed now). 

👍 🚢

Issue #1115, opened 2016-09-25
Financial aid workflow : UI


Ferdi:
- [x] #1097 : Cost Calculator Form
- [x] #1044 : Auto-approved requests
- [x] #1098 / PR #1291 : Documents required for verification
- [x] #1302 Change address and phone number for financial aid applications
- [x] #1076 : Allow learners to bypass financial aid & pay full price
- [x] #1100 : Docs reviewed and approved / rejected
- [x] #1308 : Seed currency conversion table on -rc and prod instances 
- [x] #1265 / PR #1310 : Dollar sign for personalized course price in financial aid email message
- [x] #1311 : Use calculated income not original income for determining tier
- [x]  Review page:  Country codes are not expanded

pdpinch:
I updated the description with all the UI stories. I marked #1045 as related, since @alvinsiu has already done it as part of the back-end.

pdpinch:
Updated description on 10/10 by adding a few additional items.

PR #1186, opened 2016-09-30
Extracted DateField from boundDateField

noisecapella:
None
#### What's this PR do?
Creates `DateField` which is not specific to the profile code. I will use this in the UI work regarding #1115

Edit your birthdate and a graduation date in the profile.

### DIFF
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -140,7 +140,7 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
   return <DateField
     data={profile}
     errors={errors}
-    update={updateProfile}
+    updateHandler={updateProfile}
     validator={validator}
     keySet={keySet}
     label={label}
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -1,22 +1,12 @@
 // @flow
 import React from 'react';
-import moment from 'moment';
 import _ from 'lodash';
 import TextField from 'material-ui/TextField';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 import Checkbox from 'material-ui/Checkbox';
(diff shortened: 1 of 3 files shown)


static/js/util/profile_edit.js
```diff
@@ -147,146 +137,15 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
...
-      }}
-      fullWidth={true}
-      value={edit.year !== undefined ? edit.year : ""}
-      onChange={e => setNewDate(undefined, undefined, e.target.value)}
-    />
-  </div>;
+  return <DateField
+    data={profile}
+    errors={errors}
+    update={updateProfile}
```
gsidebo: would `updateFunc` or `updateHandler` be a better name here? `update` is ambiguous and kind of confusing

static/js/util/profile_edit.js
```diff
@@ -147,146 +137,15 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
...
-      style={{
-        maxWidth: "4em"
-      }}
-      fullWidth={true}
-      value={edit.year !== undefined ? edit.year : ""}
-      onChange={e => setNewDate(undefined, undefined, e.target.value)}
-    />
-  </div>;
+  return <DateField
+    data={profile}
```
gsidebo: not wild about a prop as ambiguous as `data`, but truth be told, i don't think it was any more sensible before this refactor. we can look into this post-deadline

gsidebo:
just one small request

static/js/util/profile_edit.js
noisecapella: Sure, I'll switch to `updateHandler`

noisecapella:
Made the change

gsidebo:
looks good 👍

PR #1188, opened 2016-09-30
Implemented email templates for financial aid status change emails

### LINKED ISSUE #1118: Financial Aid: personalize email sent to learners
PR #1096 sets some constants for the e-mail body when notifying users of changes to their financial aid status. See [this comment](https://github.com/mitodl/micromasters/pull/1096/files/208bbaaa39fec22640e5b4f88e73608ef483858b#r80265876)

Ideally, these e-mails would be customized with the user's name and a few other details. Draft templates at https://goo.gl/moqYQu

alvinsiu:
Closes #1118 

#### What's this PR do?
Implements the templates from https://goo.gl/moqYQu for sending financial aid emails.  The only thing it currently omits is modifying the "from" email address, which can be updated at a later time if the default email address from `settings.MAILGUN_FROM_EMAIL` isn't the correct email to use.


mail/api.py
```diff
@@ -158,3 +169,37 @@ def send_financial_aid_email(cls, acting_user, financial_aid, subject, body):
...
+    if financial_aid.status in [FinancialAidStatus.APPROVED, FinancialAidStatus.REJECTED]:
+        program_enrollment = ProgramEnrollment.objects.get(
+            user=financial_aid.user,
+            program=financial_aid.tier_program.program
+        )
+        message = FINANCIAL_AID_APPROVAL_MESSAGE.format(
+            program_name=financial_aid.tier_program.program.title,
+            price=get_formatted_course_price(program_enrollment)["course_price"]
+        )
+        subject = FINANCIAL_AID_APPROVAL_SUBJECT
```
noisecapella: This is a format string but we aren't substituting any parameters here

noisecapella: Nevermind, it looks like this is done in `subject`

noisecapella: Can we change that to do the formatting here and in `FINANCIAL_AID_DOCUMENT_RECEIVED_SUBJECT`? It would be best to handle both cases explicitly, in case we change the template in the future and the parameters don't match up exactly anymore

financialaid/serializers.py
```diff
@@ -133,31 +128,22 @@ def save(self):
...
-            email_data.update({
-                "subject": FINANCIAL_AID_REJECTION_SUBJECT_TEXT,
-                "body": FINANCIAL_AID_REJECTION_MESSAGE_BODY
-            })
         elif self.instance.status == FinancialAidStatus.PENDING_MANUAL_APPROVAL:
-            email_data.update({
-                "subject": FINANCIAL_AID_DOCUMENTS_SUBJECT_TEXT,
-                "body": FINANCIAL_AID_DOCUMENTS_MESSAGE_BODY
-            })
+            pass
```
noisecapella: Can we remove this `elif` block since it's not doing anything? If you want to say that `PENDING_MANUAL_APPROVAL` is intentionally left blank, leave a comment instead

mail/api.py
alvinsiu: Good idea, I'll make that change!

mail/api_test.py
```diff
@@ -216,3 +240,38 @@ def test_financial_aid_email_with_blank_subject_and_body(self, mock_post):
         assert audit.from_email == settings.MAILGUN_FROM_EMAIL
         assert audit.email_subject == ''
         assert audit.email_body == ''
```
noisecapella: Can you test that `generate_financial_aid_email` raises a `ValidationError` for all other statuses?

alvinsiu: Good idea!

noisecapella:
:+1:

PR #1192, opened 2016-10-03
Added pre-enroll dashboard course states

### LINKED ISSUE #1102: Dashboard: display courses with enrollment open
There are three course states the learner can see before they have enrolled in a course:

### Enrollment Open - Basic View

image

This is the normal view when Enrollment is open. Users can click the “Pay” button to pay, or click on Enroll and Pay later to audit. 

For Variable pricing courses, like DEDP, this will start the ecommerce flow. 

For verified track courses, this will link to the edX course info page , https://courses.edx.org/course_modes/choose/{course_key}/

For all courses, Enroll and Pay can enroll the user via edX API. 

### Enrollment Open - Before Calculating Cost

image

This view is for DEDP only. The user has to Calculate their cost before paying. If they click “Enroll and Pay Later” they are immediately enrolled in the audit track. Clicking “Calculate” opens the Calculator dialog.

### Enrollment Open - Pending Income Verification

image

This view is for DEDP only. This is the status after the user has submitted documents but before they have been confirmed. The user can still enroll in the course (as audit) but they cannot pay, the button is disabled.

gsidebo:

Closes #1102 

#### What's this PR do?

- handles financial aid and course price information coming from our APIs
- covers the pre-enroll states outlined in #1102 (full invision design [here](https://projects.invisionapp.com/share/NB7VM1GSG#/screens/185611170))

static/js/components/dashboard/CourseAction.js

FYI: I have a few code snippets that create the states needed for this. If you'd like to use them, let me know and I'll tidy them up and send them over

First you'll need to set up some data:

1. CoursePrice records for any programs you intend to test
1. No previous enrollments in the CourseRuns you want to test for your user
1. For any CourseRun you want to test, set it to have a future start_date and enrollment_start_date

Second, you'll need to satisfy the following conditions to test the 3 states in #1102:

1. Program with no financial aid
1. Program with financial aid, FinancialAid record with null status
1. Program with financial aid, FinancialAid record with a pending status (`FinancialAidStatus.PENDING_DOCS` or `FinancialAidStatus.PENDING_MANUAL_APPROVAL`)

### DIFF
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -4,7 +4,8 @@ import React from 'react';
 import moment from 'moment';
 import Button from 'react-mdl/lib/Button';
 
-import type { Course, CourseRun } from '../../flow/programTypes';
+import type { Course, CourseRun, FinancialAidUserInfo } from '../../flow/programTypes';
+import type { CoursePrice } from '../../flow/dashboardTypes';
 import {
   STATUS_NOT_PASSED,
   STATUS_PASSED,
@@ -12,39 +13,88 @@ import {
   STATUS_CURRENTLY_ENROLLED,
   STATUS_OFFERED,
   DASHBOARD_FORMAT,
+  FA_PENDING_STATUSES,
+  FA_STATUS_SKIPPED
 } from '../../constants';
+import { formatPrice } from '../../util/util';
 
 export default class CourseAction extends React.Component {
   props: {
     checkout: Function,
     course: Course,
+    coursePrice: CoursePrice,
+    financialAid: FinancialAidUserInfo,
+    hasFinancialAid: boolean,
+    openFinancialAidCalculator: () => void,
     now: moment$Moment,
   };
 
-  makeEnrollButton = (text: string, run: CourseRun, disabled: boolean) => {
-    const { checkout } = this.props;
+  getCoursePrice(): string {
+    const { coursePrice } = this.props;
+    return formatPrice(coursePrice.price);
+  }
+
+  isCurrentlyEnrollable(enrollmentStartDate: ?Object): boolean {
+    const { now } = this.props;
+    return enrollmentStartDate !== null &&
+      enrollmentStartDate !== undefined &&
+      enrollmentStartDate.isSameOrBefore(now, 'day');
+  }
+
+  needsPriceCalculation(): boolean {
+    const { financialAid, hasFinancialAid } = this.props;
+    return hasFinancialAid &&
+      !financialAid.has_user_applied &&
+      financialAid.application_status !== FA_STATUS_SKIPPED;
+  }
+
+  hasPendingFinancialAid(): boolean {
+    const { financialAid, hasFinancialAid } = this.props;
+    return hasFinancialAid && FA_PENDING_STATUSES.includes(financialAid.application_status);
+  }
+
+  renderEnrollButton(run: CourseRun): React$Element<*> {
+    const {
+      checkout,
+      openFinancialAidCalculator
+    } = this.props;
+    let text = '';
+    let needsPriceCalculation = this.needsPriceCalculation();
     let buttonProps = {};
 
-    if (!disabled) {
-      buttonProps.onClick = () => {
-        checkout(run.course_id);
-      };
+    if (needsPriceCalculation) {
+      text = 'Calculate Cost';
     } else {
+      text = `Pay Now - ${this.getCoursePrice()}`;
+    }
+
+    if (this.hasPendingFinancialAid()) {
       buttonProps.disabled = true;
+    } else {
+      if (needsPriceCalculation) {
+        buttonProps.onClick = openFinancialAidCalculator;
+      } else {
+        buttonProps.onClick = () => {
+          checkout(run.course_id);
+        };
+      }
     }
 
     return (
-      <span>
-        <Button className="dashboard-button" {...buttonProps}>
-          {text}
-        </Button>
-        <span className="sr-only"> in {run.title}</span>
-      </span>
+      <Button className="dashboard-button" {...buttonProps}>
+        {text}
+      </Button>
     );
-  };
+  }
+
+  renderPayLaterLink(): React$Element<*> {
+    return (
+      <a href="#">Enroll and pay later</a>
+    );
+  }
 
   render() {
-    const { course, now } = this.props;
+    const { course } = this.props;
     let firstRun: CourseRun = {};
     if (course.runs.length > 0) {
       firstRun = course.runs[0];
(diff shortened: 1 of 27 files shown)


gsidebo:
FYI: i'm still cleaning up existing unit tests that were broken by this work

static/js/components/dashboard/CourseAction.js
```diff
@@ -34,17 +67,20 @@ export default class CourseAction extends React.Component {
...
-  };
+      <Button className="dashboard-button" {...buttonProps}>
+        {text}
+      </Button>
+    )
+  }
+
+  renderPayLaterLink(run: CourseRun): React$Element<*> {
+    return (
+      <a href="#">Enroll and pay later</a>
```
gsidebo: i believe the work to enabled direct enrollment is still in PR

static/js/components/dashboard/CourseStatus.js
```diff
@@ -1,50 +1,28 @@
 // @flow
```
gsidebo: FYI: all of this code is essentially deprecated. we're no longer showing the price in the center column of the course row, and it didn't make sense to refactor this to update the handling of the course price. there is another story to show the current grade instead in this column

noisecapella:
Can you fix the lint error?

static/js/components/dashboard/CourseAction.js
noisecapella: Yes, we need both front and back end work for this

static/js/constants.js
```diff
@@ -540,11 +550,24 @@ export const PROGRAM_ENROLLMENTS = [
   },
 ];
 
+export const FINANCIAL_AID_PARTIAL_RESPONSE = {
+  application_status: null,
```
noisecapella: Is there a sensible default we can put here?

static/js/components/dashboard/CourseRow_test.js
```diff
@@ -8,16 +8,35 @@ import CourseRow from './CourseRow';
...
+    const financialAid = FINANCIAL_AID_PARTIAL_RESPONSE;
+    const coursePrice = COURSE_PRICES_RESPONSE[0];
     const now = moment();
     const checkout = () => null;
-    const wrapper = shallow(<CourseRow course={course} now={now} checkout={checkout} />);
+    const wrapper = shallow(
+      <CourseRow
+        course={course}
+        hasFinancialAid={hasFinancialAid}
+        financialAid={financialAid}g
```
noisecapella: Typo after the `}`

static/js/components/dashboard/CourseAction_test.js
```diff
@@ -59,134 +90,159 @@ describe('CourseAction', () => {
...
+
+      FA_PENDING_STATUSES.forEach(pendingStatus => {
+        financialAidObject.application_status = pendingStatus;
+
+        const wrapper = shallow(
+          <CourseAction course={course} {...aidParams(financialAidObject)} {...defaultParamsNow} />
+        );
+        let elements = getElements(wrapper);
+
+        assert.isDefined(elements.button.props().disabled);
```
noisecapella: I think we also want to assert that it's true, not just defined

static/js/components/dashboard/CourseAction_test.js
```diff
@@ -5,34 +5,65 @@ import { shallow } from 'enzyme';
...
+    return {
+      buttonContainer: buttonContainer,
+      button: button,
+      buttonText: buttonText,
+      description: description
+    };
+  };
+
+  let alterFirstRun = (course, overrideObject) => {
+    course.runs[0] = Object.assign({}, course.runs[0], overrideObject);
```
noisecapella: It looks like this is cloning the run but not the course. It doesn't look like we clone the course before this function is called, `findCourse` just returns a reference to some course in `DASHBOARD_RESPONSE`. Can we change this to return a copy?

noisecapella:
Can you update `FinancialAidCard` to change `course_price` to `price`?

noisecapella:
It looks like `Calculate Cost` currently attempts a checkout like `Pay Now` does. Can we change it to open the calculator like the other calculate button does?

gsidebo:
re: linting and typo, fix is incoming. i was in the middle of fixing flow errors but i think i broke something in the process.

re: Calculate Cost, 👍 . wasn't 100% sure that had been merged. do you know if the 'enroll now and pay later' functionality is available yet?

noisecapella:
Right, 'enroll now' needs an API on the server first which we haven't gotten to yet AFAIK

noisecapella:
It looks like the `Payment due:` message is under the pay button in one mockup but it's on the left in other mockups... @roberthouse54 where should it go?

static/js/components/dashboard/CourseAction_test.js
gsidebo: i'm actually trying to modify the object in place.  `CourseAction` takes a `course` object, and `course` is a complex object. the goal here is just to alter some data on the first run. I realize this is a side effect and not the best pattern. would you rather i change this to return a copy of `course` with the modified first run? if so, this function would be more accurately called `courseWithAlteredFirstRun` or something like that

gsidebo:
fyi i'm going off of this mockup [here](https://projects.invisionapp.com/share/NB7VM1GSG#/screens/185611170)

static/js/components/dashboard/CourseStatus.js
gsidebo: more work? we need _some_ component in the middle there. its name is likely to change, but i'd much rather leave the placeholder than delete it and worry about cleaning up unit tests

static/js/constants.js
gsidebo: if the user hasn't started the application process, `has_user_applied: false` already indicates that and i don't think we need a specific status value to reinforce that

noisecapella: Sounds good

static/js/components/dashboard/CourseAction_test.js
noisecapella: It's fine to modify the object in place here but we need to clone the object before the function is called so we don't persist changes in `DASHBOARD_RESPONSE` between tests

static/js/components/dashboard/CourseAction.js
```diff
@@ -57,34 +96,30 @@ export default class CourseAction extends React.Component {
       action = <i className="material-icons">done</i>;
       break;
     case STATUS_CAN_UPGRADE: {
-      action = this.makeEnrollButton("Upgrade", firstRun, false);
+      let formattedUpgradeDate = moment(firstRun.upgrade_deadline).format(DASHBOARD_FORMAT);
```
gsidebo: changed... but do you have any idea why that is?? seems so arbitrary. we're literally just changing the key from `upgrade_deadline` to `course_upgrade_deadline` before returning the results.

noisecapella:
Either place for the payment due message is fine with me, if @roberthouse54 had a different idea we can change it in another PR

static/js/components/dashboard/CourseAction.js
noisecapella: I agree that it's arbitrary. I'd feel fine changing it after MVP

gsidebo:
@noisecapella should be ready

noisecapella:
Looks good :+1:

PR #1194, opened 2016-10-03
Removed filter on program enrollments in dashboard API

### LINKED ISSUE #1193: Remove program enrollments filtering on dashboard
There's the assumption that the dashboard contains all live programs, so we need to remove the filter which was added recently restricting this to program enrollments. After MVP we should add back this filter and make use (and secure) the `/api/v0/programs/` API in the front end for this purpose instead

noisecapella:
Fixes #1193

#### What's this PR do?
Removes the filter on program enrollments. The dashboard API will again show programs the user is not currently enrolled in.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -15,14 +15,13 @@ from edx_api.client import EdxApi
 
 from backends import utils
 from backends.edxorg import EdxOrgOAuth2
-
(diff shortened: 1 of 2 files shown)


giocalitri:
👍 when tests pass

Issue #1191, opened 2016-10-03
Program pages: Unique url for each tab


Ferdi:
Navigation menu items currently don't get a unique url. This create a problem when someone wants to e-mail/share a particular page (e.g. FAQ page)

pdpinch:
@Ferdi are hashes OK for this, or do we need more?

I'm wondering in particular, once we add categories to the FAQ (#1190) if we might need to target particular parts of the FAQ page.

Ferdi:
I'm not really sure why we are implementing these tabs in JS.   For example, in terms of SEO, are metatags going to change based on where you are homepage or FAQ?

pdpinch:
I don't think there was a particularly strong reason for implementing this in JavaScript. It does make the UI more responsive, but other than that we move to one page per tab. 

How would the metatags be different for a program page and a program FAQ? We can give them different title, keyword and description, but I don't think that would have much of an effect on SEO (other than potentially getting a hit for each page, instead of just one).

roberthouse54:
I suggest that we hold off on any further changes to this page. The SEO is not a huge issue in my opinion, and we can revisit after the MVP launch if needed.

pdpinch:
@Ferdi and I spoke this morning and came to the same conclusion. We'll probably revisit when we get to work on #597

pdpinch:
Since we've discovered a bug in the JavaScript implementation, it seems wisest to change direction and go with a full URL for each tab.

PR #1204, opened 2016-10-03
Set unique URL for tabs in program page

annagav:
Fix ##1191
#### What's this PR do?
Set a unique hash in the url for tabs in the Program page.

The user can directly go to desired tab having the correct hash in the url. If hash is invalid then the hash defaults to the panel that is active.

screen shot 2016-10-03 at 3 35 39 pm

### DIFF
--- a/static/js/public.js
+++ b/static/js/public.js
@@ -121,3 +121,38 @@ jQuery(document).ready(function ($) {
     emailBody: description + CURRENT_PAGE_URL
   });
 });
+
+/**
+ * Set url hash if hash provided in the url,
+ * or set hash based on the active panel
+ */
+$(function(){
+  $('.mdl-tabs__tab').click(function(){
+    document.location.hash = $(this).attr('href');
+  });
+  if (document.location.hash){
+    setPanelActive(document.location.hash);
+  } else {
+    location.hash = $('.mdl-tabs__tab.is-active').attr('href');
+  }
+});
+
+$(window).on('hashchange', function () {
+  if (location.hash) {
+    setPanelActive(location.hash);
+  }
+});
+
+/**
+ *  Given a valid hash, set the corresponding panel active.
+ */
+function setPanelActive(hash){
+  var $panel = $(hash);
+  if ($panel.length > 0) {
+    $(".mdl-tabs__panel, .mdl-tabs__tab").removeClass('is-active');
+    $panel.addClass('is-active');
+    $(`a.mdl-tabs__tab[href="${hash}"]`).addClass('is-active');
+  } else {
+    location.hash = $('.mdl-tabs__tab.is-active').attr('href');
+  }
+}


pdpinch:
Let's hold off on reviewing this until the discussion in #1191 is resolved.

pdpinch:
@singingwolfboy can you review this?

static/js/public.js
```diff
@@ -121,3 +121,37 @@ jQuery(document).ready(function ($) {
     emailBody: description + CURRENT_PAGE_URL
   });
 });
+
+/**
+ * Set url hash if hash provided in the url,
+ * or set hash based on the active panel
+ */
+$(function(){
+  $('.mdl-tabs__tab').click(function(){
```
singingwolfboy: Should we call `event.preventDefault()` as part of this function? If not, I would expect that clicking the `<a>` element will cause the page to scroll up to the top. (I haven't tested that yet, though.) Is that the behavior we want?

static/js/public.js
```diff
@@ -121,3 +121,37 @@ jQuery(document).ready(function ($) {
...
+});
+
+/**
+ *  Given a valid hash, set the corresponding panel active.
+ */
+function setPanelActive(hash){
+  if ($(hash).length > 0) {
+    $(".mdl-tabs__panel, .mdl-tabs__tab").removeClass('is-active');
+    $(hash).addClass('is-active');
+    $(`a[href="${hash}"]`).addClass('is-active');
```
singingwolfboy: Should this have more context on the selector? Could there be other `<a>` tags on the page that would be affected?

static/js/public.js
```diff
@@ -121,3 +121,37 @@ jQuery(document).ready(function ($) {
...
+  }
+});
+
+/**
+ *  Given a valid hash, set the corresponding panel active.
+ */
+function setPanelActive(hash){
+  if ($(hash).length > 0) {
+    $(".mdl-tabs__panel, .mdl-tabs__tab").removeClass('is-active');
+    $(hash).addClass('is-active');
```
singingwolfboy: Nitpick: you could create a variable at the start of this function and reference that here, instead of calling `$(hash)` multiple times.
```js
var $hash = $(hash);
if ($hash.length > 0) {
  $hash.addClass('is-active');
  // ...
```

static/js/public.js
annagav: Right now it doesn't scroll up to the top. Does it make sense to add it  anyway?

pdpinch:
#1191 is irrelevant. @annagav can you respond to @singingwolfboy's comments and then this should be good to merge.

annagav:
I did.

annagav:
@singingwolfboy, Could you please take a look at the changes?

singingwolfboy:
@annagav looks good to me 👍

PR #1208, opened 2016-10-04
Improve social links with Google+

### LINKED ISSUE #1200: fixup meta tag and add schema.org metatags
A couple more things recommended by  https://github.com/joshbuchea/HEAD

- [ ] split the meta tag at https://github.com/mitodl/micromasters/blob/master/ui/templates/base.html#L4 into two. 
  First one with `charset`, as close as possible to the start of the page. 
  2nd one with the viewport settings.
- [ ] Add Google+ / Schema.org tags
  see https://github.com/joshbuchea/HEAD#google--schemaorg

singingwolfboy:
Fixes #1200

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -34,7 +34,6 @@
   <meta property="og:image:height" content="630">
   <meta name="twitter:card" content="summary">
   <meta name="twitter:site" content="@MITxonedX">
-  <link href="https://plus.google.com/+MITX" rel="publisher">
   <meta itemprop="name" content="MIT MicroMasters">
   <meta itemprop="description" content="MicroMasters is a
     new digital credential for online learners. The MicroMasters
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -34,6 +34,15 @@
   <meta property="og:image:height" content="630">
   <meta name="twitter:card" content="summary">
   <meta name="twitter:site" content="@MITxonedX">
+  <link href="https://plus.google.com/+MITX" rel="publisher">
+  <meta itemprop="name" content="MIT MicroMasters">
+  <meta itemprop="description" content="MicroMasters is a
+    new digital credential for online learners. The MicroMasters
+    credential will be granted to learners who complete an
+    integrated set of graduate-level online courses. With the MicroMasters
+    credentials, learners can apply for an accelerated master's degree
+    program on campus, at MIT or other top universities.">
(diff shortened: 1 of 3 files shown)


singingwolfboy:
As an aside, Wagtail already provides fields for [`seo_title`](http://docs.wagtail.io/en/latest/reference/pages/model_reference.html#wagtail.wagtailcore.models.Page.seo_title) and [`search_description`](http://docs.wagtail.io/en/latest/reference/pages/model_reference.html#wagtail.wagtailcore.models.Page.search_description) -- we might want to actually use them.

cms/templates/cms/home_page.html
```diff
@@ -34,6 +34,15 @@
   <meta property="og:image:height" content="630">
   <meta name="twitter:card" content="summary">
   <meta name="twitter:site" content="@MITxonedX">
+  <link href="https://plus.google.com/+MITX" rel="publisher">
```
pdpinch: Do we need a Google Plus account? Because this one isn't ours!

cms/templates/cms/program_page.html
```diff
@@ -21,6 +21,10 @@
   <meta property="og:image:height" content="{{ page.background_image.height }}">
   <meta name="twitter:card" content="summary">
   <meta name="twitter:site" content="@MITxonedX">
+  <link href="https://plus.google.com/+MITX" rel="publisher">
```
pdpinch: Same comment as home_page.html.

pdpinch:
How are `seo_title` and `search_description` used in a default install of Wagtail? For the moment, I'm happy to have them be static, but if we need to start customizing them, I can see changing the template.

cms/templates/cms/home_page.html
pdpinch: I've confirmed that MITx doesn't have a Google+ account. I hope we can just drop this line.

singingwolfboy: Line dropped, on both templates.

singingwolfboy:
@pdpinch In a template, you can refer to `{{ page.seo_title }}` or `{{ page.search_description }}`, which are set in the "Promote" tab of the Wagtail CMS. If we're allowing non-technical users to set information about programs in Wagtail, it makes sense to also allow them to set these values, and use the values they set in the template. However, that seems outside the scope of this pull request. Maybe that could be another issue for the future.

pdpinch:
Thanks @singingwolfboy this is good to merge.

pdpinch:
@singingwolfboy don't forget to delete your branch

Issue #1106, opened 2016-09-23
add ZenDesk help button


pdpinch:
We need a prominent location on the site for users to contact us -- through ZenDesk -- with questions. The button needs to be on all pages, but it probably makes sense to have separate PRs for (in priority order):

- [x] Homepage, Program Page, TOS
- [x] Dashboard and search
- [ ] Financial Aid Review

We'll embed a [ZenDesk widget](https://support.zendesk.com/hc/en-us/articles/203908456)

```
<!-- Start of odl Zendesk Widget script -->
<script>/*<![CDATA[*/window.zEmbed||function(e,t){var n,o,d,i,s,a=[],r=document.createElement("iframe");window.zEmbed=function(){a.push(arguments)},window.zE=window.zE||window.zEmbed,r.src="javascript:false",r.title="",r.role="presentation",(r.frameElement||r).style.cssText="display: none",d=document.getElementsByTagName("script"),d=d[d.length-1],d.parentNode.insertBefore(r,d),i=r.contentWindow,s=i.document;try{o=s}catch(e){n=document.domain,r.src='javascript:var d=document.open();d.domain="'+n+'";void(0);',o=s}o.open()._l=function(){var o=this.createElement("script");n&&(this.domain=n),o.id="js-iframe-async",o.src=e,this.t=+new Date,this.zendeskHost=t,this.zEQueue=a,this.body.appendChild(o)},o.write('<body onload="document._l();">'),o.close()}("https://assets.zendesk.com/embeddable_framework/main.js","odl.zendesk.com");
/*]]>*/</script>
<!-- End of odl Zendesk Widget script -->
```

pdpinch:
@roberthouse54 can you give this some thought? I think a mailto link is adequate at this point, but the ZenDesk widget is easy to implement too, and they would both go to the same inbox.

pdpinch:
I spoke with @roberthouse54 and we're going to go with the default anchored "Help" button from ZenDesk. He is interested in styling the button though, if that's possible. 

I'll generate the necessary javascript from zendesk and then you can add it @amir-qayyum-khan

amir-qayyum-khan:
@pdpinch  sure, please tell me where (on which page ) I should add this button? And what will it do?

pdpinch:
Updated the description to answer @amir-qayyum-khan's questions in https://github.com/mitodl/micromasters/issues/1106#issuecomment-250456645

amir-qayyum-khan:
[...]

PR #1211, opened 2016-10-04
Added ZenDesk help button on MM

amir-qayyum-khan:
- Fixes https://github.com/mitodl/micromasters/issues/1106

#### What's this PR do?
- Adds ZenDesk help button

- see home_page,

See 
- [x] Homepage
- [x] Program Page
- [x] TOS
- [x] Dashboard and search

@pdpinch @noisecapella @roberthouse54 
screen shot 2016-10-04 at 5 12 47 pm
screen shot 2016-10-04 at 5 19 05 pm

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -39,10 +39,7 @@
 
 <div id="home-page">
 {% include "header.html" %}
-  <!-- Start of odl Zendesk Widget script -->
-  <script>/*<![CDATA[*/window.zEmbed||function(e,t){var n,o,d,i,s,a=[],r=document.createElement("iframe");window.zEmbed=function(){a.push(arguments)},window.zE=window.zE||window.zEmbed,r.src="javascript:false",r.title="",r.role="presentation",(r.frameElement||r).style.cssText="display: none",d=document.getElementsByTagName("script"),d=d[d.length-1],d.parentNode.insertBefore(r,d),i=r.contentWindow,s=i.document;try{o=s}catch(e){n=document.domain,r.src='javascript:var d=document.open();d.domain="'+n+'";void(0);',o=s}o.open()._l=function(){var o=this.createElement("script");n&&(this.domain=n),o.id="js-iframe-async",o.src=e,this.t=+new Date,this.zendeskHost=t,this.zEQueue=a,this.body.appendChild(o)},o.write('<body onload="document._l();">'),o.close()}("https://assets.zendesk.com/embeddable_framework/main.js","odl.zendesk.com");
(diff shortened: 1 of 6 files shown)


pdpinch:
Is there a reason why you didn't put the zendesk code in the header template?

amir-qayyum-khan:
i was just trying different things, the button is by default floating. I think putting in header is good idea

pdpinch:
@aliceriot do you have any comment on where the the widget code is put in the dashboard?

I still need to test functionality and @roberthouse54 needs to look at design.

aliceriot:
do we want to have the zendesk thingy on every page? if so it would make sense to me to put it in the base html template instead of duplicating it on two templates.

pdpinch:
Yes, every page. base.html sounds even better than header.html.

aliceriot:
yea, that's included on the dashboard as well.

amir-qayyum-khan:
Done @aliceriot @pdpinch

roberthouse54:
Is there any way we can customize the color of that button? If so, it would be good to make it $mm-brand-bg

aliceriot:
^^ I agree, the style of the button is pretty different than anything else we have on the site. It looks like the button is in an iframe though, so our ability to style it may be limited. Can we set a stylesheet to be used on Zendesk's website?

amir-qayyum-khan:
@aliceriot @roberthouse54 this widget appears in and iframe. I dont thinks so we can change its color.

screen shot 2016-10-06 at 1 09 56 pm

pdpinch:
Looking at the Zendesk docs (linked off the issue) it looks like I can set the color in Zendesk. I'll try that and give you updated code later this morning.

amir-qayyum-khan:
sure, thanks @pdpinch

pdpinch:
I don't see an easy way to override the stylesheet entirely, without rewriting the widget entirely (which is an option). 

I can change the color, but I have to give it a fixed value. I just changed it to the current value of `$mm-brand-bg` #30bb5c

pdpinch:
I just checked the functionality, and it works as expected. We'll want to do some customizations later (like recognizing logged-in users, and detecting which program the question is about) but that can come later. 

Thumbs up from me if @aliceriot and @roberthouse54 can live with this for now.

pdpinch:
I found this discussion forum about hacking the CSS, but it's not supported:

https://support.zendesk.com/hc/en-us/community/posts/207616307-Customize-the-Web-Widget-with-CSS

I'm pretty surprised, frankly, that they don't have a solution to this.

aliceriot:
yeah it's not optimal but I think it's fine.

PR #1216, opened 2016-10-04
Added batch refresh of current grades

### LINKED ISSUE #763: Update current grades when batch updating learner data from edx.org
When we batch update learner data from edx.org we need to also update current grades. 

Related to #671 and #762.

Depends on mitocw/edx-platform#259

giocalitri:
closes #763

#### What's this PR do?
Adds bach refresh for current grades. It also fixes a missing refresh for users who never accessed the dashboard.

### DIFF
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -9,7 +9,7 @@ from celery import group
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.cache import cache
-
+from django.db.models import Q
 from edx_api.client import EdxApi
 from social.apps.django_app.default.models import UserSocialAuth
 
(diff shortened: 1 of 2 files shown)


dashboard/tasks.py
```diff
@@ -68,18 +69,34 @@ def release_lock():
...
+        ).values_list('user', flat=True).distinct()
+
+        users_expired_grades = CachedCurrentGrade.objects.filter(
+            last_request__lt=refresh_time_limit
+        ).values_list('user', flat=True).distinct()
 
-        certificated_students = CachedCertificate.objects.filter(
-            last_request__lt=datetime.utcnow() - timedelta(hours=MAX_HRS_MAIN_TASK)
-        ).values_list('user', flat=True)
+        users_expired_cache = list(
```
noisecapella: We don't need to convert to a `list` here, Django will accept a set for the `id__in` parameter

giocalitri: you are right

noisecapella:
functionality and code look good :+1:

Issue #1190, opened 2016-10-03
Program pages:  FAQ categories


Ferdi:
For some MM programs, FAQs needs to be grouped in categories. 

## Acceptance Criteria

- [ ] Add a field to the FAQ model in the CMS to categorize
  - [ ] If possible, the CMS should display categories as grouped to aid authoring
- [ ] Group Questions / Answers based on category in the template

roberthouse54:
For the FAQ tab UI, please just add an H2 header element, and I can tweak the styles later.

singingwolfboy:
Is this more like a tagging scheme, where each question can be tagged with multiple categories, and each category can be used for multiple questions? Or is it a many-to-one relationship, where each category can contain multiple questions, but each question has only one category? Also, is it mandatory for each question to have at least one category, or should we support uncategorized questions? (Which is basically an "uncategorized" category, I suppose...)

singingwolfboy:
Also, are the categories hierarchical? That is, can you have a categories and sub-categories? Or is there only one level of categorization?

pdpinch:
I believe we were thinking one level, with a many to one relationship -- each category has multiple questions. 

It would be useful to allow an "Uncategorized" option, but I'm not sure what the design would be.

@ferdi? @roberthouse?

Ferdi:
- no hierarchy - flat
- tags is a superset of many-to-one in this context but harder to implement
in cms? Yes?
- a program either use categories (then tagging /categorization becomes
required) or not at all

- Sent from my phone

singingwolfboy:
Wagtail doesn't have built-in support for tagging, but it has a [tagging recipe right in the official documentation](http://docs.wagtail.io/en/v1.6.2/reference/pages/model_recipes.html#tagging), demonstrating how such a feature could be built using other modules in the Django ecosystem. There's no equivalent recipe for many-to-one categories, although I did find [a StackOverflow question with some code that may work](http://stackoverflow.com/questions/33651722/nested-categories-inlinepanels-in-wagtail).

Tags are certainly more flexible than categories, but that flexibility may not be what we want. If a question has more than one tag, how would it be displayed in the template? Would the question be rendered multiple times, one for each tag? Should we say that each question has one "primary" tag -- and if so, how is that different from a many-to-one category? If a question has multiple tags, is the ordering of those tags important, or not?

Ferdi:
@singingwolfboy and I just chatted a bit. 

[...]

PR #1232, opened 2016-10-04
Categorized FAQs

singingwolfboy:
#1190

#### What's this PR do?
Allow FAQ questions to be categorized

Go into the CMS, click on the "Snippets" section, and make a FAQ category or two. Then add a few FAQs to a course, and assign them to categories

View the FAQ tab on the program page. Be sure to test with several different variations of questions and categories. Note that @roberthouse54 has asked me to just put the category in a `<h2>` tag and leave the styling to him, so the styling isn't very nice right now -- he will take care of that later (or perhaps push a commit to this pull request).

snippets section
snippet FAQ categories
added a General FAQ category
selecting a category for a FAQ
setting up FAQs in a program
viewing FAQs in a program


Ferdi:
@singingwolfboy thanks for the fast response on this.  This looks pretty good as a first pass.  
Is there anyway to (custom) order categories for program owners ?

Ferdi:
Can we make "Category" a required field when present ?   They have the option to category="other"/"uncategorized".

singingwolfboy:
@Ferdi: At the moment, there's no way to implement custom category orderings. If we support that, FAQ categories will have to be scoped to programs, so that different programs can specify different orderings for their FAQ categories. I'm not sure the best way to do that in Wagtail -- make it another page, instead of a snippet?

I don't really like the idea of making "Category" a required field. The way I've coded the pull request, if the program owner makes a question without specifying a category, it will end up in an "Uncategorized" section at the bottom. (See the screenshots I've posted.) It means that "Uncategorized" will always be the last category, even if the program owner makes categories with names that begin with "V", "X", "Y", or "Z". Does that do what you want?

singingwolfboy:
Making "Category" an optional field also allows program owners to make uncategorized FAQs, which might be handy if the FAQ is very short. If there's only one or two questions, why bother to organize those questions into categories at all? If none of the questions have categories assigned to them, the template won't render any category `<h2>` tags, not even an "Uncategorized" one.

pdpinch:
Should this be reviewed, or are we waiting for the changes described in https://github.com/mitodl/micromasters/issues/1190#issuecomment-251549015?

Ferdi:
No need for review at the moment. DB will open a new PR

singingwolfboy:
@Ferdi: Unfortunately, the changes we discussed may not be possible after all. I can easily set up the relationships between models so that each Program has many Categories and many FrequentlyAskedQuestions, and each FrequentlyAskedQuestion is associated with one Category, but the tricky part is making sure that questions are only associated with categories in the same program. Wagtail renders the Category dropdown menu with a list of _all_ categories across _all_ programs, which is very confusing: for every category in that list, there's no way to know which program it's associated with.

I pored through the documentation for both Django and Wagtail, and there doesn't seem to be any way to limit the association from the Django Models side. I asked several people about this online, and everyone I talked to said that the only way to do this is by customizing the Form class used to render the dropdown menu. However, Wagtail creates the Form class automatically, based on the Model! There is [a short section in Wagtail's documentation about customizing the generated forms](http://docs.wagtail.io/en/v1.6.3/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms), but when I tried the code that the documentation suggested, it doesn't seem to actually *work*. I suspect there's some edge-case around customizing an _inline_ form, rather than a form for a full-page model. I went as far as reading form generation code within Wagtail, and it seems to account for inline forms, but the code is very complex, and I'm not familiar with it. There may well be a bug in how generated inline forms are customized.

I'm going to keep trying to get this to work, but I wanted to provide a status update, since I've spent several hours on this already. It may not be worth the time and effort.

pdpinch:
Can we just make this really stupid and add a third field to the FAQ object: i.e., question, answer and category? We could leave it up to the view/template to handle the logic for grouping FAQs with the same category text.

singingwolfboy:
@pdpinch: That's what's in this pull request right now. In this pull request, categories are global across all programs, and are always ordered alphabetically in the template. @Ferdi wanted me to make a new pull request where each program can define its own categories, and its own ordering for those categories. I thought that was possible, but it seems like I'm running up against the limitations of the Django ORM.

singingwolfboy:
@pdpinch: oh, you're suggesting that we make "Category" just another text field on the FAQ object, rather than its own model?

pdpinch:
Indeed, that's what I was suggesting. Maybe "denormalized" would have been clearer. 

I think @giocalitri has some thoughts. He said he'd get in touch after lunch -- in an hour or so?

Ferdi:
Thanks @singingwolfboy 

Is it possible to override the form element generated by wagtail to only show categories that belong in the current program ? 

http://docs.wagtail.io/en/v1.6.3/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms

singingwolfboy:
@Ferdi see my comment above:

> There is [a short section in Wagtail's documentation about customizing the generated forms](http://docs.wagtail.io/en/v1.6.3/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms), but when I tried the code that the documentation suggested, it doesn't seem to actually *work*. I suspect there's some edge-case around customizing an _inline_ form, rather than a form for a full-page model. I went as far as reading form generation code within Wagtail, and it seems to account for inline forms, but the code is very complex, and I'm not familiar with it. There may well be a bug in how generated inline forms are customized.

It may be possible, but it wasn't working when I followed the instructions in the documentation.

singingwolfboy:
@pdpinch I made a new pull request with "category" as a simple text field: #1243

singingwolfboy:
@Ferdi: if you want to see the code that I tried, I pushed the commit to GitHub: https://github.com/mitodl/micromasters/commit/14cce0725534046cbea9e690b25002b68693a5b7 As I said, it doesn't seem to work. It's also not clear how to customize the form to make the queryset dynamic, depending on which program is being rendered.

pdpinch:
@singingwolfboy can you rebase this? (not sure how that happened...)

We'll move forward with this approach, but please open an issue for limiting the scope of the categories per program. Maybe you can try Wagtail support channels to see why your filter code wasn't working.

pdpinch:
@amir-qayyum-khan can you review?

singingwolfboy:
I've rebased this branch, and created #1266 as a follow-up issue.

amir-qayyum-khan:
please rebase PR

amir-qayyum-khan:
Functionality works as mentioned

cms/migrations/0019_add_faqcategory.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import modelcluster.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cms', '0018_update_programfaculty'),
```
amir-qayyum-khan: Need to update this after rebase

amir-qayyum-khan: These is 0019 migration that i recently merged

amir-qayyum-khan:
Code looks good

pdpinch:
Now that #1259 has merged, we can approach this a little differently. If we add a category field to the FAQPage, we can write a template that combines all the FAQ pages into one, using the category from each one.

pdpinch:
If #1332 works as expected, we'll replace this PR with that one.

PR #1241, opened 2016-10-05
Added logging for IsSignedByCyberSource

### LINKED ISSUE #1239: Add logging for IsSignedByCybersource
This is a permission class which verifies the HMAC signature for posts to `/api/v0/order_fulfillment/`. We should add logging for cases when this fails, otherwise we won't know if we have a Cybersource configuration problem here

noisecapella:
Fixes #1239 

#### What's this PR do?
Adds logging for signature verification failures due to misconfigurations. Additionally I need this PR build to diagnose this on micromasters-rc

### DIFF
--- a/ecommerce/permissions.py
+++ b/ecommerce/permissions.py
@@ -1,9 +1,14 @@
 """
 Permission classes for ecommerce
 """
+import logging
+
 from rest_framework.permissions import BasePermission
 
 from ecommerce.api import generate_cybersource_sa_signature
+
+
+log = logging.getLogger(__name__)
 
 
 class IsSignedByCyberSource(BasePermission):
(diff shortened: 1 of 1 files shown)


aliceriot:
this LGTM but I'm not sure if I should try to test it - what do you think @noisecapella ?

noisecapella:
You can try something like this to provide an invalid message

    curl http://micromasters.local:8079/api/v0/order_fulfillment/ -X POST --header "Content-type: application/json" --data '{"signed_field_names": "x", "signature": "", "x":"y"}'

Then check that you got the right logging message

aliceriot:
when I try that I got

```
{"detail":"Authentication credentials were not provided."}
```

noisecapella:
That's expected, do you get a logging message in the console? The docker-compose stdout is finicky sometimes, we can also use a PR build if that's easier

aliceriot:
this is what's in the docker-compose output:

output

PR #1243, opened 2016-10-05
Add category to FAQ as a text field

singingwolfboy:
A different fix for #1190, competing with #1232. This pull request adds an optional "category" text field to the `FrequentlyAskedQuestion` model. The template is responsible for grouping together questions of the same category. There is one notable change in the behavior of this pull request, compared to #1232: In this pull request, FAQs are always displayed in the order that they are defined in the Wagtail interface. In #1232, FAQs are first sorted by category alphabetically.

As a result of the above change, if FAQs have the same category but are not consecutive, the category will be repeated in the template. For example, a list of questions like this:

* question="question one" category="foo"
* question="question two" category="foo"
* question="question three" category="bar"
* question="question four" category="foo"

will be displayed like this:

    <h2>foo</h2>
    <p>question one</p>
    <p>question two</p>
    <h2>bar</h2>
    <p>question three</p>
    <h2>foo</h2>
    <p>question four</p>

Notice how questions one and two are combined under the same heading, but question four is not. If this is not the desired behavior, and questions should be ordered by the `category` attribute rather than the `sort_order` attribute, I can change this.


pdpinch:
Closing this in favor of #1232

PR #1244, opened 2016-10-05
fixes FinancialAidAudit JSONfields

### LINKED ISSUE #1229: FinancialAidAudit data_before and data_after store JSON as text representation

mirfark:

Fixes #1229

#### What's this PR do?

Adds to_dict function to FinancialAid objects so that FinancialAidAudit objects are stored as searchable dicts. Note that this is a slight content change; here is the data now stored in FinancialAidAudit.data_before and .data_after:

{'country_of_income': 'USA',
 'date_exchange_rate': None,
 'id': 34,
 'income_usd': 39000,
 'original_currency': 'USD',
 'original_income': 39000,
 'status': 'created',
 'tier_program': 2,
 'user': 117}

### DIFF
--- a/financialaid/models.py
+++ b/financialaid/models.py
@@ -4,13 +4,13 @@ Models for the Financial Aid App
 import datetime
 
 from django.contrib.auth.models import User
-from django.core import serializers
 from django.contrib.postgres.fields import JSONField
 from django.core.exceptions import ValidationError
 from django.db import (
     models,
     transaction,
 )
+from django.forms.models import model_to_dict
 
 from courses.models import Program
 
@@ -149,6 +149,14 @@ class FinancialAid(TimestampedModel):
     date_exchange_rate = models.DateTimeField(null=True)
     date_documents_sent = models.DateField(null=True)
 
+    def to_dict(self):
+        """
+        Get the model_to_dict of self
+        """
+        ret = model_to_dict(self)
+        ret["date_exchange_rate"] = str(ret["date_exchange_rate"])
+        return ret
+
     def save(self, *args, **kwargs):
         """
         Override save to make sure only one FinancialAid object exists for a User and the associated Program
@@ -171,8 +179,8 @@ class FinancialAid(TimestampedModel):
         FinancialAidAudit.objects.create(
             acting_user=acting_user,
             financial_aid=self,
-            data_before=serializers.serialize("json", [financialaid_before, ]),
-            data_after=serializers.serialize("json", [self, ])
+            data_before=financialaid_before.to_dict(),
+            data_after=self.to_dict()
         )
 
 
--- a/financialaid/models.py
+++ b/financialaid/models.py
@@ -155,9 +155,9 @@ class FinancialAid(TimestampedModel):
         """
         ret = model_to_dict(self)
         if self.date_exchange_rate is not None:
-            ret["date_exchange_rate"] = str(ret["date_exchange_rate"])
+            ret["date_exchange_rate"] = ret["date_exchange_rate"].isoformat()
         if self.date_documents_sent is not None:
-            ret["date_documents_sent"] = str(ret["date_documents_sent"])
+            ret["date_documents_sent"] = ret["date_documents_sent"].isoformat()
         return ret
 
     def save(self, *args, **kwargs):
--- a/financialaid/models.py
+++ b/financialaid/models.py
@@ -154,7 +154,10 @@ class FinancialAid(TimestampedModel):
         Get the model_to_dict of self
         """
         ret = model_to_dict(self)
-        ret["date_exchange_rate"] = str(ret["date_exchange_rate"])
+        if self.date_exchange_rate is not None:
+            ret["date_exchange_rate"] = str(ret["date_exchange_rate"])
+        if self.date_documents_sent is not None:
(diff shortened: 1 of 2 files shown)


financialaid/models_test.py
```diff
@@ -90,3 +87,25 @@ def test_save_and_log(self):  # pylint: disable=no-self-use
...
+        self.assertAlmostEqual(financial_aid_dict["income_usd"], financial_aid.income_usd)
+        self.assertAlmostEqual(financial_aid_dict["original_income"], financial_aid.original_income)
+        if financial_aid.date_exchange_rate is None:
+            assert financial_aid_dict["date_exchange_rate"] == financial_aid.date_exchange_rate
+        else:
+            assert financial_aid_dict["date_exchange_rate"] == str(financial_aid.date_exchange_rate)
+        if financial_aid.date_documents_sent is None:
+            assert financial_aid_dict["date_documents_sent"] == financial_aid.date_documents_sent
+        else:
+            assert financial_aid_dict["date_documents_sent"] == str(financial_aid.date_documents_sent)
```
giocalitri: I think you need multiple tests to test each case and not have an `if` statement in the test.

financialaid/models_test.py
```diff
@@ -90,3 +87,25 @@ def test_save_and_log(self):  # pylint: disable=no-self-use
                 self.assertAlmostEqual(value, original_after_json[field])
             else:
                 assert value == original_after_json[field]
+
+    def test_to_dict(self):
+        """
+        Tests that to_dict() function on FinancialAid
+        """
+        financial_aid = FinancialAidFactory.create()
+        financial_aid_dict = financial_aid.to_dict()
```
giocalitri: you may want to assert that all the keys you want are in the dict before the following asserts

financialaid/models_test.py
```diff
@@ -90,3 +88,40 @@ def test_save_and_log(self):  # pylint: disable=no-self-use
...
+            "user",
+            "tier_program",
+            "status",
+            "original_currency",
+            "country_of_income",
+            "date_exchange_rate",
+            "date_documents_sent",
+            "income_usd",
+            "original_income"
+        ]
```
giocalitri: nit: please sort them

financialaid/models.py
```diff
@@ -171,8 +182,8 @@ def save_and_log(self, acting_user, *args, **kwargs):
         FinancialAidAudit.objects.create(
             acting_user=acting_user,
             financial_aid=self,
-            data_before=serializers.serialize("json", [financialaid_before, ]),
-            data_after=serializers.serialize("json", [self, ])
+            data_before=financialaid_before.to_dict(),
```
giocalitri: Question: can `financialaid_before = FinancialAid.objects.get(id=self.id)` raise `FinancialAid.DoesNotExist`?

financialaid/models_test.py
```diff
@@ -90,3 +88,40 @@ def test_save_and_log(self):  # pylint: disable=no-self-use
...
+        assert set(financial_aid_dict.keys()).issuperset(set(audit_key_list))
```
giocalitri: why `issuperset`? If someone adds more fields to `FinancialAid`, this test should probably fail and force the engineer to add the extra field here.

financialaid/models_test.py
```diff
@@ -90,3 +88,40 @@ def test_save_and_log(self):  # pylint: disable=no-self-use
...
+        self.assertAlmostEqual(financial_aid_dict["original_income"], financial_aid.original_income)
+
+    def test_to_dict_with_null_values(self):  # pylint: disable=no-self-use
+        """
+        Tests the to_dict() function on FinancialAid with none-values in date fields
+        """
+        financial_aid = FinancialAidFactory.create(date_exchange_rate=None, date_documents_sent=None)
+        financial_aid_dict = financial_aid.to_dict()
+        assert financial_aid_dict["date_exchange_rate"] == financial_aid.date_exchange_rate
+        assert financial_aid_dict["date_documents_sent"] == financial_aid.date_documents_sent
```
giocalitri: Maybe it is better to be explicit and asset 
`assert financial_aid_dict["date_exchange_rate"] is None`

financialaid/models.py
```diff
@@ -149,6 +149,17 @@ class FinancialAid(TimestampedModel):
...
 
+    def to_dict(self):
+        """
+        Get the model_to_dict of self
+        """
+        ret = model_to_dict(self)
+        if self.date_exchange_rate is not None:
+            ret["date_exchange_rate"] = str(ret["date_exchange_rate"])
+        if self.date_documents_sent is not None:
+            ret["date_documents_sent"] = str(ret["date_documents_sent"])
```
noisecapella: Please use `isoformat()` here and above. `str` is rarely the right way to convert an object to a string

financialaid/models.py
zags: A DoesNotExist error would only come from a bug.  There are three ways one could happen:

* Someone calls `save_and_log` on a FinancialAid object that has not yet been saved to the database
* Someone calls `save_and_log` on a FinancialAid object that has been deleted
* Someone modifies the `id` field and then calls `save_and_log`

None of these are correct uses of the FinancialAid object

giocalitri:
👍

PR #1249, opened 2016-10-05
Added log.error for every ImproperlyConfigured exception raised

giocalitri:
N/A
#### What's this PR do?
Given that we have a custom exception handler that intercepts all the `ImproperlyConfigured` exceptions, we need a way to track when the error happens, so we should `log.error` before any `ImproperlyConfigured` is raised.

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -1,11 +1,15 @@
 """
 Models for course structure
 """
+import logging
 from datetime import datetime
 
 import pytz
 from django.core.exceptions import ImproperlyConfigured
 from django.db import models
+
+
+log = logging.getLogger(__name__)
 
 
 class Program(models.Model):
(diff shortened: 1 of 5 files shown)


aliceriot:
well when I do `ag --ignore='*test*.py' -l ImproperlyConfigured` I get:

```sh
courses/models.py
dashboard/api.py
dashboard/utils.py
ecommerce/api.py
financialaid/api.py
micromasters/settings.py
micromasters/utils.py
```

but when I do `git diff --name-only origin/master` on your branch I get:

```
courses/models.py
dashboard/api.py
dashboard/utils.py
ecommerce/api.py
financialaid/api.py
```

do we not care about the usage of `ImproperlyConfigured` in `micromasters/settings.py`? It looks like in `micromasters/utils.py` is where we set up the exception handler, so obviously we don't need to change it there.

aliceriot:
Giovanni explained things!

PR #1257, opened 2016-10-05
Financial Aid Exchange Rate management command

### LINKED ISSUE #1235: Currency exchange objects are not created until celery tasks are run
Need to create a management command or similar to run on deployment to generate the currency exchange objects

mirfark:

Fixes #1235 

#### What's this PR do?

Adds a management command to fetch exchange rate data so that the financial aid currency conversion functionality is available on the first day of initial deployment (as opposed to waiting for the celery task).

### DIFF
--- a/financialaid/management/commands/generate_exchange_rates.py
+++ b/financialaid/management/commands/generate_exchange_rates.py
@@ -1,0 +1,16 @@
+"""
+Generates CurrencyExchangeRate objects
+"""
+from django.core.management import BaseCommand
+
+from financialaid.tasks import sync_currency_exchange_rates
+
+
+class Command(BaseCommand):
+    """
+    Generates CurrencyExchangeRate objects by calling sync_currency_exchange_rates
+    """
+    help = "Generates CurrencyExchangeRate objects by calling sync_currency_exchange_rates"
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        sync_currency_exchange_rates()
--- a/financialaid/management/commands/generate_exchange_rates_test.py
+++ b/financialaid/management/commands/generate_exchange_rates_test.py
@@ -1,0 +1,47 @@
+"""
+Test for management command generating exchange rates
+"""
+from mock import patch
+
+from financialaid.constants import CURRENCY_EXCHANGE_RATE_API_REQUEST_URL
+from financialaid.management.commands import generate_exchange_rates
+from financialaid.models import CurrencyExchangeRate
+from search.base import ESTestCase
+
+
+@patch('financialaid.tasks.requests.get')
+class GenerateExchangeRatesTest(ESTestCase):
+    """
+    Tests for generate_exchange_rates management command
+    """
+    @classmethod
+    def setUpTestData(cls):
+        cls.command = generate_exchange_rates.Command()
+
+    def setUp(self):
+        super(GenerateExchangeRatesTest, self).setUp()
+        self.data = {
+            "extraneous information": "blah blah blah",
+            "rates": {
+                "CBA": "3.5",
+                "FED": "1.9",
+                "RQP": "0.5"
+            }
+        }
+
+    def test_currency_exchange_rate_command(self, mocked_request):
+        """
+        Assert currency exchange rates are created using management command
+        """
+        mocked_request.return_value.json.return_value = self.data
+        assert CurrencyExchangeRate.objects.count() == 0
+        self.command.handle("generate_exchange_rates")
+        called_args, _ = mocked_request.call_args
+        assert called_args[0] == CURRENCY_EXCHANGE_RATE_API_REQUEST_URL
+        assert CurrencyExchangeRate.objects.count() == 3
+        currency = CurrencyExchangeRate.objects.get(currency_code="CBA")
+        assert currency.exchange_rate == 3.5
+        currency = CurrencyExchangeRate.objects.get(currency_code="FED")
+        assert currency.exchange_rate == 1.9
+        currency = CurrencyExchangeRate.objects.get(currency_code="RQP")
+        assert currency.exchange_rate == 0.5
--- a/financialaid/management/commands/update_exchange_rates.py
+++ b/financialaid/management/commands/update_exchange_rates.py
@@ -8,9 +8,10 @@ from financialaid.tasks import sync_currency_exchange_rates
 
 class Command(BaseCommand):
     """
-    Generates CurrencyExchangeRate objects by calling sync_currency_exchange_rates
+    Update the local database with the latest exchange rate information from the
+    Open Exchange Rates API (openexchangerates.org)
     """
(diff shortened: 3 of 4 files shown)


financialaid/management/commands/generate_exchange_rates.py
```diff
@@ -0,0 +1,16 @@
...
+from django.core.management import BaseCommand
+
+from financialaid.tasks import sync_currency_exchange_rates
+
+
+class Command(BaseCommand):
+    """
+    Generates CurrencyExchangeRate objects by calling sync_currency_exchange_rates
+    """
+    help = "Generates CurrencyExchangeRate objects by calling sync_currency_exchange_rates"
```
singingwolfboy: To an operations person running this management command, who probably does _not_ have intimate knowledge of this codebase, this description is not helpful. Instead, I would suggest: "Update the local database with the latest exchange rate information from the Open Exchange Rates API (openexchangerates.org)"

financialaid/management/commands/generate_exchange_rates.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        sync_currency_exchange_rates()
```
singingwolfboy: I just took a look at the implementation of the `sync_currency_exchange_rates()` function, and it doesn't handle API failures at all. This is a big problem -- again, consider that the person running this command might be an operations person who does not have intimate knowledge of the codebase. API failures happen, and they need to be accounted for.

Can you add a bit of failure handling logic to the `sync_currency_exchange_rates()` function, and make sure that it handles failures gracefully? At a minimum, the function should call [`raise_for_status()`](http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status) on the HTTP response _before_ calling `.json()` on it. This management command should use a `try`/`except` block when calling the function, and provide a helpful error message that includes the exact URL that failed, as well as the HTTP status code of the response.

financialaid/management/commands/generate_exchange_rates_test.py
```diff
@@ -0,0 +1,47 @@
...
+        self.data = {
+            "extraneous information": "blah blah blah",
+            "rates": {
+                "CBA": "3.5",
+                "FED": "1.9",
+                "RQP": "0.5"
+            }
+        }
+
+    def test_currency_exchange_rate_command(self, mocked_request):
```
singingwolfboy: Please add tests that simulate API failures. At a minimum, we should be able to handle a "500 Server Error" failure and a "429 Rate Limit Exceeded" failure. Also check out [the list of documented error responses](https://docs.openexchangerates.org/docs/errors).

financialaid/management/commands/generate_exchange_rates_test.py
```diff
@@ -0,0 +1,47 @@
...
+        self.command.handle("generate_exchange_rates")
+        called_args, _ = mocked_request.call_args
+        assert called_args[0] == CURRENCY_EXCHANGE_RATE_API_REQUEST_URL
+        assert CurrencyExchangeRate.objects.count() == 3
+        currency = CurrencyExchangeRate.objects.get(currency_code="CBA")
+        assert currency.exchange_rate == 3.5
+        currency = CurrencyExchangeRate.objects.get(currency_code="FED")
+        assert currency.exchange_rate == 1.9
+        currency = CurrencyExchangeRate.objects.get(currency_code="RQP")
+        assert currency.exchange_rate == 0.5
```
singingwolfboy: Nitpick: could you use more descriptive names for the variables you're asserting on here?
```python
cba_curr = CurrencyExchangeRate.objects.get(currency_code="CBA")
assert cba_curr.exchange_rate == 3.5
fed_curr = CurrencyExchangeRate.objects.get(currency_code="FED")
assert fed_curr.exchange_rate == 1.9
rqp_curr = CurrencyExchangeRate.objects.get(currency_code="RQP")
assert fqp_curr.exchange_rate == 0.5
```

singingwolfboy: Also, I'm not familiar with the details of the test framework here. Is there something that will automatically clean up or rollback the objects that this test creates in the database? If not, it's a good idea to write a `cleanUp()` method that removes the objects that the test created, so that the tests are independent of each other.

financialaid/management/commands/generate_exchange_rates.py
singingwolfboy: I'd also suggest naming this file `update_exchange_rates.py`, so that the command to use it is `./manage.py update_exchange_rates`.

financialaid/management/commands/generate_exchange_rates_test.py
mirfark: Anything in `setUpTestData()` will rollback after the class of tests; anything in `setUp()` will rollback after each test method in the class. I'll make the changes to the variable names.

financialaid/management/commands/generate_exchange_rates.py
mirfark: I'll create a separate issue for this.

mirfark: #1267

financialaid/management/commands/generate_exchange_rates_test.py
mirfark: Those should be handled in the other issue.

financialaid/management/commands/generate_exchange_rates_test.py
```diff
@@ -0,0 +1,47 @@
...
+from mock import patch
+
+from financialaid.constants import CURRENCY_EXCHANGE_RATE_API_REQUEST_URL
+from financialaid.management.commands import generate_exchange_rates
+from financialaid.models import CurrencyExchangeRate
+from search.base import ESTestCase
+
+
+@patch('financialaid.tasks.requests.get')
+class GenerateExchangeRatesTest(ESTestCase):
```
singingwolfboy: Ah, OK. In that case, we can still inherit from `ESTestCase`. (Although perhaps that class should be renamed to something more generic, if we want *all* tests to inherit from it...)

financialaid/management/commands/generate_exchange_rates.py
mirfark: I fixed the variable names and edited the management command description (as well as name).

financialaid/management/commands/generate_exchange_rates_test.py
mirfark: I just changed this to TestCase instead of ESTestCase -- should I change it back?

noisecapella: Either is fine with me

noisecapella:
Looks good to me :+1:

alvinsiu:
@singingwolfboy If this looks good to you too, could you approve your change request?  Thanks!

noisecapella:
I don't think he's working today, so feel free to merge since all comments were addressed

Issue #1198, opened 2016-10-03
Make Django Admin for audit logs read-only


alvinsiu:
We should make audit logs read-only in Django Admin.

PR #1258, opened 2016-10-05
makes fields read_only in django admin for financialaidaudit objects

mirfark:

Addresses #1198.

#### What's this PR do?

Makes FinancialAidAudit objects not editable in the Django Admin interface. Note that there is not currently a way to disable editing on objects in the Django admin while still having them appear in the interface (https://code.djangoproject.com/ticket/8936). As a result we just set all of the fields to be read_only.

### DIFF
--- a/financialaid/admin.py
+++ b/financialaid/admin.py
@@ -20,7 +20,9 @@ class FinancialAidAdmin(admin.ModelAdmin):
 class FinancialAidAuditAdmin(admin.ModelAdmin):
     """Admin for FinancialAidAudit"""
     model = FinancialAidAudit
-    readonly_fields = ["acting_user", "financial_aid", "data_before", "data_after"]
+    readonly_fields = [
+        f.name for f in FinancialAidAudit._meta.get_fields() if not f.auto_created  # pylint: disable=protected-access
+    ]
 
     def has_add_permission(self, *args, **kwargs):  # pylint: disable=unused-argument
         return False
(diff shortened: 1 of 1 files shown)


financialaid/admin.py
```diff
@@ -20,6 +20,7 @@ class FinancialAidAdmin(admin.ModelAdmin):
 class FinancialAidAuditAdmin(admin.ModelAdmin):
     """Admin for FinancialAidAudit"""
     model = FinancialAidAudit
+    readonly_fields = ["acting_user", "financial_aid", "data_before", "data_after"]
```
singingwolfboy: Since the goal is to make _all_ fields read-only, it's better to dynamically get the list of fields from the model, instead of hard-coding them here. You can get the fields from a model by calling `Model._meta.get_fields()`. So in this case, the code would be:

```python
readonly_fields = [f.name for f in FinancialAidAudit._meta.get_fields()]
```

When I ran that locally, I got the following list of fields: `['id', 'created_on', 'updated_on', 'acting_user', 'financial_aid', 'data_before', 'data_after']` That's more than you have listed here -- is that an intended difference?

singingwolfboy: @zags In that case, maybe do this instead?
```python
readonly_fields = [f.name for f in FinancialAidAudit._meta.get_fields()
                   if f.editable and not f.auto_created]
```

PR #1259, opened 2016-10-05
Created FaqsPage

### LINKED ISSUE #1191: Program pages: Unique url for each tab

annagav:
Fix #1191
#### What's this PR do?
Make FAQs a separate page, with a unique url.
program_page.html, cms/models.py

For an existing program page (or newly created), in cms click 'add a child page'. Enter title (which will be the title in the tab), and define sample questions. Publish the page, and view it.

screen shot 2016-10-07 at 12 05 23 am

screen shot 2016-10-07 at 12 06 47 am


cms/models.py
```diff
@@ -62,6 +62,45 @@ def get_context(self, request):
...
+        context["public_src"] = get_bundle_url(request, "public.js")
+        context["style_public_src"] = get_bundle_url(request, "style_public.js")
+        context["authenticated"] = not request.user.is_anonymous()
+        context["signup_dialog_src"] = get_bundle_url(request, "signup_dialog.js")
+        context["faculty_carousel_src"] = get_bundle_url(request, "faculty_carousel.js")
+        context["username"] = username
+        context["js_settings_json"] = json.dumps(js_settings)
+        context["title"] = self.get_parent().title
+        context["courses_info"] = courses_info
+        context['programpage'] = self.get_parent().first()
```
pdpinch: There's a lot of copy and paste here from the ProgramPage. Is it all necessary? If it is, can we refactor to keep it DRY?

cms/models.py
```diff
@@ -62,6 +62,45 @@ def get_context(self, request):
...
+    content_panels = Page.content_panels + [
+        InlinePanel('faqs', label='Frequently Asked Questions'),
+    ]
+
+    def get_context(self, request):
+        js_settings = {
+            "gaTrackingID": settings.GA_TRACKING_ID,
+            "host": webpack_dev_server_host(request),
+            "programId": self.get_parent().program.id,
+            "faculty": faculty_for_carousel(self.faculty_members.all()),
```
pdpinch: I don't think we have faculty on the FAQ page?

annagav: I asked Rob. We do.

cms/migrations/0019_auto_20161005_2008.py
```diff
@@ -0,0 +1,37 @@
...
+        migrations.RemoveField(
+            model_name='frequentlyaskedquestion',
+            name='program_page',
+        ),
+        migrations.AddField(
+            model_name='frequentlyaskedquestion',
+            name='faqs_page',
+            field=modelcluster.fields.ParentalKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='faqs', to='cms.FaqsPage'),
+        ),
+    ]
```
roberthouse54: This is @giocalitri :
this migration cannot be reverted and should not be merged as is.

This file should be renamed to something meaningful

cms/models.py
```diff
@@ -131,37 +150,42 @@ class ProgramPage(Page):
...
-        }
-        username = get_social_username(request.user)
-        context = super(ProgramPage, self).get_context(request)
 
-        courses_info = []
-        for course in self.program.course_set.all().order_by(
-                'position_in_program'
-        ):
-            courses_info.append((course, get_course_enrollment_text(course), get_course_url(course)))
+        return get_program_page_context(self, request)
```
noisecapella: Why move this code into a separate function?

cms/templates/cms/faqs_page.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% block panel %}
+  <div class="faqs">
+    {% for faq in faqpage.faqs.all %}
+    <div class="faq">
+      <div class="question">
+        {{ faq.question }}
+      </div>
+      <div class="answer">
+        {{ faq.answer|safe }}
```
noisecapella: This isn't a good idea. If you're intending to allow rich text use the `richtext` filter instead: http://docs.wagtail.io/en/v1.6.3/topics/writing_templates.html#rich-text-filter

cms/models.py
noisecapella: Nevermind I see now that it's used in two places

noisecapella:
The general direction of this looks good!

pdpinch:
@annagav showed me the functionality and this is what we want. 

Currently it allows for multiple FAQ pages (using multiple tabs). We can do a follow up PR to merge the multiple pages into one and thereby address #1266 as well.

Ferdi:
Sounds good to me. Is there any way to make FAQ pages orderable ?

annagav:
I think so.

noisecapella:
Can you fix the lint errors?

cms/models.py
```diff
@@ -210,7 +240,7 @@ class FrequentlyAskedQuestion(Orderable):
     """
     FAQs for the program
     """
-    program_page = ParentalKey(ProgramPage, related_name='faqs')
+    faqs_page = ParentalKey(FaqsPage, related_name='faqs', null=True)
```
noisecapella: I think the migration reverse problem comes from this change. `program_page` doesn't allow null values so when it's added back with a null default, it causes problems. I don't think there's anything we can do about this, we just have to live with the migration being non-reversable. Any thoughts @giocalitri?

giocalitri: fine with me

pdpinch:
@annagav showed me the functionality and this is what we want. 

Currently it allows for multiple FAQ pages (using multiple tabs). We can do a follow up PR to merge the multiple pages into one and thereby address #1266 as well.

noisecapella:
:+1:

Issue #1197, opened 2016-10-03
Audit financial aid object changes in Django Admin


alvinsiu:
We should add logging to changes made to `FinancialAid` objects in Django admin.

PR #1263, opened 2016-10-06
Adds audit logging to FinancialAid admin

mirfark:

Addresses #1197.

#### What's this PR do?

Adds logging to FinancialAid object changes made in the Django admin interface. FinancialAidAudit objects are created when a FinancialAid object is edited through the model admin. In order to edit a FinancialAid object in the admin interface without being forced to fill out the date_documents_sent field, that particular field in the model now has the option to be blank.


noisecapella:
Looks good :+1:

PR #1269, opened 2016-10-06
Limited the course grade cache refresh to the enrolled runs

giocalitri:
N/A

#### What's this PR do?
Limited the course grade cache refresh to the enrolled runs and not for all the runs in MM, given that an user cannot have a grade if she was not enrolled in the first place

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -356,13 +356,14 @@ def update_cached_enrollment(user, enrollment, course_id, now):
     )
 
 
-def _check_if_refresh(user, cached_model, refresh_delta):
+def _check_if_refresh(user, cached_model, refresh_delta, limit_to_courses=None):
     """
     Helper function to check if cached data in a model need to be refreshed.
     Args:
         user (django.contrib.auth.models.User): A user
         cached_model (dashboard.models.CachedEdxInfoModel): a model containing cached data
         refresh_delta (datetime.datetime): time limit for refresh the data
+        limit_to_courses (list): a list of course ids to limit the check
 
     Returns:
         tuple: a tuple containing:
@@ -373,6 +374,9 @@ def _check_if_refresh(user, cached_model, refresh_delta):
     course_ids = CourseRun.objects.filter(course__program__live=True).exclude(
         Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
     ).values_list("edx_course_key", flat=True)
+
+    if limit_to_courses is not None:
+        course_ids = course_ids.filter(edx_course_key__in=limit_to_courses)
 
     model_queryset = cached_model.objects.filter(
         user=user,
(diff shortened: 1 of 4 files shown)


dashboard/api_test.py
```diff
@@ -1352,9 +1362,11 @@ def test_new(self):
         with patch('dashboard.api.get_social_username', autospec=True, return_value=username):
             ret = api.get_student_current_grades(self.user, edx_client)
         # Test that the current grades object returned by the edx_api function is exactly the same
-        # object as what our function tests
+        # object as what our function tests]
```
noisecapella: Typo

noisecapella:
If the user unenrolls from edX at some point, do we clear the grades we have cached from them after a while?

dashboard/api.py
```diff
@@ -498,8 +502,10 @@ def get_student_current_grades(user, edx_client):
     refresh_delta = now - datetime.timedelta(hours=REFRESH_GRADES_CACHE_HOURS)
 
     with transaction.atomic():
+        enrolled_curses = models.CachedEnrollment.objects.filter(user=user).exclude(data=None).values_list(
+            "course_run__edx_course_key", flat=True)
         is_data_fresh, grades_queryset, course_ids = _check_if_refresh(
-            user, models.CachedCurrentGrade, refresh_delta)
+            user, models.CachedCurrentGrade, refresh_delta, enrolled_curses)
```
gsidebo: typo. `enrolled_curses` -> `enrolled_courses`

dashboard/api.py
```diff
@@ -498,8 +502,10 @@ def get_student_current_grades(user, edx_client):
     refresh_delta = now - datetime.timedelta(hours=REFRESH_GRADES_CACHE_HOURS)
 
     with transaction.atomic():
+        enrolled_curses = models.CachedEnrollment.objects.filter(user=user).exclude(data=None).values_list(
```
noisecapella: It seems like this would create a dependency where `get_student_enrollments` must get run first or else some grades may not get fetched. Can you add a comment saying this in `dashboard/tasks.py` and `dashboard/views.py` where we call the function?

dashboard/api.py
```diff
@@ -498,8 +502,10 @@ def get_student_current_grades(user, edx_client):
...
+            "course_run__edx_course_key", flat=True)
```
noisecapella: I forget, what's the significance of `exclude(data=None)` again?

noisecapella:
Looks good overall, just a couple comments

dashboard/api.py
giocalitri: it removes from the queryset the `data` fields in the table with `null` values

noisecapella: I mean, in what cases should `CachedEnrollment`s have `data=null`?

giocalitri: we create an entry in the cache for each user and each course run no matter if there is data or not.
This because it is easier to understand when the cache needs to be refreshed or not.
But this also means that the majority of the entries have `data=null`

noisecapella: Ah makes sense

giocalitri:
@noisecapella 
> If the user unenrolls from edX at some point, do we clear the grades we have cached from them after a while?

If the user refreshes the dashboard, we refresh after only few minutes, otherwise the bach refresh will take care of it every 6 hours.

noisecapella:
:+1:

PR #1271, opened 2016-10-06
Fetch course prices and dashboard after every relevant change on the server

### LINKED ISSUE #1231: After creating new profile, user gets a "No Program Enrollment is available" exception
Steps to reproduce:

1. Sign up for the portal using an edX account with no profile. 
2. Complete the profile. Note in particular the program chosen on the first form.
3. Finish the profile and load /dashboard

Result:

image

Expected result: 

A complete dashboard with the correct program data displayed. On micromasters-rc, this might look like:

screen shot 2016-10-04 at 6 54 09 pm

Note that if you reload the page after getting the error, it will clear and render as above. 

This seems to only happen with new users.

### LINKED ISSUE #1238: When user enrolls in a new program they get a 'No program enrollment is available' message
To produce this bug I was enrolled in three `ProgramEnrollment`s on micromasters-rc. I deleted the one I was on currently then refreshed the page. I added it back using the 'Enroll in a new program' dialog. After the refresh happened I got the 'No program enrollment is available.'

I think this is due to course prices not being refreshed at the same time as the dashboard API. If I refresh the page manually it works fine afterwards

noisecapella:
Fixes #1231 
Fixes #1238 

#### What's this PR do?
Calls `fetchDashboard` and `fetchCoursePrices` after each successful API call that modifies the server in a relevant way. This also moves some existing function calls into the reducer to make them easier to test.

Have at least one program enrollment which needs to be added. Add it in the program selector in the dashboard. The dashboard should refresh and you should see the new program.

Remove the program enrollment via Django shell and go to `/profile`. Select the program and continue through the signup flow. The dashboard should be refreshed with the course selected and you should see no error.

You should also test that these things still work:
 - adding a new financial aid request
 - submitting the document date for the financial aid request

### DIFF
--- a/static/js/actions/financial_aid.js
+++ b/static/js/actions/financial_aid.js
@@ -36,6 +36,8 @@ export const addFinancialAid = (income: number, currency: string, programId: num
     return api.addFinancialAid(income, currency, programId).then(
       () => {
         dispatch(receiveAddFinancialAidSuccess());
+        dispatch(fetchCoursePrices());
+        dispatch(fetchDashboard());
         return Promise.resolve();
       },
       () => {
--- a/static/js/actions/documents.js
+++ b/static/js/actions/documents.js
@@ -2,6 +2,10 @@
 import { createAction } from 'redux-actions';
 import type { Dispatch } from 'redux';
 
+import {
+  fetchDashboard,
+  fetchCoursePrices,
+} from './';
 import * as api from '../util/api';
 import type { Dispatcher } from '../flow/reduxTypes';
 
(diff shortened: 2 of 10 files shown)


aliceriot:
functionality looks great! looking at code now.

static/js/actions/financial_aid.js
```diff
@@ -36,6 +36,8 @@ export const addFinancialAid = (income: number, currency: string, programId: num
     return api.addFinancialAid(income, currency, programId).then(
       () => {
         dispatch(receiveAddFinancialAidSuccess());
+        dispatch(fetchCoursePrices());
+        dispatch(fetchDashboard());
```
aliceriot: this is a good idea overall, this way you can't forget to put the right stuff in `.then`!

noisecapella: Yeah, it's also easier to test

PR #1274, opened 2016-10-06
Switched to hosted jquery and bootstrap

### LINKED ISSUE #1264: Host jquery, bootstrap, google fonts locally
This would prevent problems when one of the CDNs goes down like jquery's did this morning

noisecapella:
Fixes #1264 

#### What's this PR do?
Includes bootstrap css and jquery in the static directory and updates links on base.html. The bootstrap JS is already included using public.js

Make sure the mailing list button still works, make sure everything looks ok in general

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3360,9 +3360,9 @@
       "resolved": "https://registry.npmjs.org/jodid25519/-/jodid25519-1.0.2.tgz"
     },
     "jquery": {
-      "version": "3.1.1",
-      "from": "jquery@>=1.7.2",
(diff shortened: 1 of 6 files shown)


pdpinch:
Is there a way that we can pull this in at build time, instead of adding jquery to the repo? I know it leaves in vulnerable to network problems -- but only at build time.

noisecapella:
I'll see about adding it to NPM

noisecapella:
Ready for review again

giocalitri:
Everything seems to work

giocalitri:
👍

Issue #1170, opened 2016-09-29
Currency codes don't match up perfectly between JS client and the server


noisecapella:
Related to #1146

`json_codes` comes from `JSON.stringify(cc.codes())` on the client. `py_codes` comes from `CurrencyExchangeRate.objects.values_list('currency_code' , flat=True)`

    In [32]: py_codes - json_codes
    Out[32]: {'BTC', 'BYN', 'EEK', 'GGP', 'IMP', 'JEP', 'MTL', 'SVC', 'ZMK', 'ZWL'}
    
    In [33]: json_codes - py_codes
    Out[33]: 
    {'BOV',
     'CHE',
     'CHW',
     'COU',
     'MXV',
     'SSP',
     'USN',
     'USS',
     'UYI',
     'XBA',
     'XBB',
     'XBC',
     'XBD',
     'XBT',
     'XFU',
     'XTS',
     'XXX'}

aliceriot:
@noisecapella so just to confirm everything on the list is a currency code we want to exclude on the client? I can easily take care of this as part of #1275

noisecapella:
Yeah I would say we should support only the currencies on both lists for right now, then later on use the server side list. I'm not sure how many of those currencies are rare or unused or otherwise not worth supporting, but since we only use the currencies to convert the users income it's not a big deal to not support a currency IMO

PR #1275, opened 2016-10-06
set default currency in UI

### LINKED ISSUE #1048: Financial Aid: Currency Conversion

aliceriot:

closes #1048 

#### What's this PR do?

This guesses what a user's currency should be when they start the financial aid calculator process, based on the `country` value in their profile.

read over the code!

1. Delete any financial aid objects on your user
2. go to `/dashboard`
3. open the financial aid calculator

when you do the appropriate currency for the country on your profile should already be selected.

### DIFF
--- a/static/js/util/currency.js
+++ b/static/js/util/currency.js
@@ -1,6 +1,27 @@
 // @flow
 import cc from 'currency-codes';
 import R from 'ramda';
+import iso3166 from 'iso-3166-2';
+
+export const excludedCurrencyCodes = [
+  'BOV',
+  'CHE',
+  'CHW',
+  'COU',
+  'MXV',
+  'SSP',
+  'USN',
+  'USS',
+  'UYI',
+  'XBA',
+  'XBB',
+  'XBC',
+  'XBD',
+  'XBT',
+  'XFU',
+  'XTS',
+  'XXX'
+];
 
 const codeToOption = code => (
   { value: code, label: cc.code(code).currency }
@@ -8,10 +29,22 @@ const codeToOption = code => (
 
 const labelSort = R.sortBy(R.compose(R.toLower, R.prop('label')));
 
-const excludeUnusedCodes = R.reject(R.flip(R.contains)(['USS', 'USN']));
+const invalidCurrency = R.flip(R.contains)(excludedCurrencyCodes);
 
 const codesToOptions = R.compose(
-  labelSort, R.map(codeToOption), excludeUnusedCodes
+  labelSort, R.map(codeToOption), R.reject(invalidCurrency)
 );
 
 export const currencyOptions = codesToOptions(cc.codes());
+
+const codeToCountryName = code => iso3166.country(code).name || '';
+
+const currencyToCode = currency => (
+  currency.length === 0 ? '' : currency[0].code
+);
+
+const excludeSingleCode = code => invalidCurrency(code) ? '' : code;
+
+export const currencyForCountry = R.compose(
+  excludeSingleCode, currencyToCode, cc.country, R.toLower, codeToCountryName
+);
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -30,7 +30,10 @@ import {
   setDocumentSentDate,
   updateDocumentSentDate,
 } from '../actions/documents';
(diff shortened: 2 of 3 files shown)


noisecapella:
Looks good overall but I found this bug: I tried setting my current location to Palestine and I got this error in the console:

screenshot from 2016-10-06 18-12-46

static/js/util/currency.js
```diff
@@ -15,3 +16,13 @@ const codesToOptions = R.compose(
 );
 
 export const currencyOptions = codesToOptions(cc.codes());
+
+const codeToCountryName = code => iso3166.country(code).name || '';
+
+const countryNameToCurrency = name => (
+  name === '' ? '' : cc.country(name)[0].code
```
zags: Should the default be USD?  I'm not sure what the desired behavior is here

static/js/util/currency.js
```diff
@@ -15,3 +16,13 @@ const codesToOptions = R.compose(
 );
 
 export const currencyOptions = codesToOptions(cc.codes());
```
zags: I know this isn't part of the current PR, but it looks like there is the possibility of a mismatch between the currencies in the library an the currencies supported by the conversion API.  Is that addressed elsewhere or should we open an issue for it?

noisecapella: I made #1170 a few days ago for this

static/js/util/currency.js
aliceriot: no, we're not defaulting to USD. The default is whatever the default currency for your country it.

static/js/util/currency.js
```diff
@@ -1,17 +1,50 @@
 // @flow
 import cc from 'currency-codes';
 import R from 'ramda';
+import iso3166 from 'iso-3166-2';
+
+export const excludedCurrencyCodes = [
```
alvinsiu: This list may change over the course of time, depending on whether the frontend or backend APIs decide they want to add or remove supported currencies.  Our backend automatically updates our list of supported currencies based on what the backend API returns to us, which means `CurrencyExchangeRate.objects.values_list('currency_code' , flat=True)` will always return the list of supported and current exchange rates.

I don't think it's that likely that the list of supported/not supported codes will change too quickly, but fetching these dynamically instead of hardcoding them would protect against those cases.

aliceriot: is there an endpoint for getting the list of excluded codes?

alvinsiu: Not currently, but we can build one really quickly (or it may be easier to pass it along with the dashboard API).  @noisecapella what do you think?

aliceriot:
@noisecapella I fixed the issue with Palestine (well, I made the code safe for any valid country code for which there is not a currency issue) and also made sure that currencies we don't support are filtered out.

alvinsiu:
#1282 is relevant to this. @pdpinch - this PR doesn't implement the UI for 400 in cost calculation, but does prevent the 400 from not supported currency from occurring 99% of the time.

We've made a separate issue #1285 for the UX change for after MVP.  If that sounds good to you, @aliceriot can add "Closes #1282" to this PR.

static/js/util/currency.js
aliceriot: my inclination would be to file an issue for that and go with the hardcoded list for now

pdpinch: I'm OK with that. I don't think this list will change all that much.

aliceriot: yea that's my thinking too, and probably the API will only ever expand to include more currencies, so I doubt we'd run into a situation where we're allowing the user to selected a currency that's invalid on the backend.

noisecapella: I want to avoid APIs built really quickly unless we have an immediate need for them

noisecapella:
:+1:

static/js/util/currency.js
zags: I meant default to USD if there is no currency listed for the country

Issue #1248, opened 2016-10-05
page_404 error


noisecapella:
I ran into this when I was directed back to the dashboard after fulfilling the order

    2016-10-05T19:11:21.107661+00:00 app[web.1]: [2016-10-05 19:11:21] WARNING 18 [django.request] base.py:182 - [b88fcf9b-672f-4997-9725-26581357feca] - Not Found: /favicon.ico
    2016-10-05T19:11:21.128502+00:00 app[web.1]: [2016-10-05 19:11:21] DEBUG 18 [django.template] base.py:927 - [b88fcf9b-672f-4997-9725-26581357feca] - Exception while resolving variable 'style_public_src' in template '404.html'.
    2016-10-05T19:11:21.128504+00:00 app[web.1]: Traceback (most recent call last):
    2016-10-05T19:11:21.128505+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.5/site-packages/django/core/handlers/base.py", line 134, in get_response
    2016-10-05T19:11:21.128506+00:00 app[web.1]:     resolver_match = resolver.resolve(request.path_info)
    2016-10-05T19:11:21.128507+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.5/site-packages/newrelic-2.54.0.41/newrelic/hooks/framework_django.py", line 553, in wrapper
    2016-10-05T19:11:21.128507+00:00 app[web.1]:     return _wrapped(*args, **kwargs)
    2016-10-05T19:11:21.128508+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.5/site-packages/newrelic-2.54.0.41/newrelic/hooks/framework_django.py", line 541, in _wrapped
    2016-10-05T19:11:21.128509+00:00 app[web.1]:     result = wrapped(path)
    2016-10-05T19:11:21.128509+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.5/site-packages/django/core/urlresolvers.py", line 404, in resolve
    2016-10-05T19:11:21.128510+00:00 app[web.1]:     raise Resolver404({'tried': tried, 'path': new_path})
[...]

PR #1277, opened 2016-10-06
Fix for 404 page when passed exception kwarg

alvinsiu:
#1248


noisecapella:
Can you remove the code changes regarding the Cybersource signature? The `KeyError` you're guarding against shouldn't happen since Cybersource guarantees the field to exist

alvinsiu:
Would it make more sense to leave it in so that it returns `403` instead of `500` if that endpoint is hit from not-Cybersource?

alvinsiu:
I'll remove it for now, let me know if you'd like me to put it back in for the reason above! :]

noisecapella:
:+1:

PR #1289, opened 2016-10-07
Implemented enroll links

### LINKED ISSUE #1223: Enable 'enroll and pay later' functionality in UI
design: https://projects.invisionapp.com/share/NB7VM1GSG#/screens/185611170

In several different cases, we need to provide users with the ability to enroll in a course immediately and pay later. Currently this link exists but doesn't do anything. The API endpoint to perform this enrollment is being done in a separate PR (#1226)

noisecapella:
Fixes #1223

#### What's this PR do?
Implements the enroll link on the course row

Unenroll from a course on edX. Then click the enroll link for that same course on micromasters. Go back to edX and verify that the course is enrolled again.

I noticed that we show the enroll link even if the user is currently enrolled when they need to upgrade, but I figured that's outside the scope of this PR

### DIFF
--- a/static/js/flow/enrollmentTypes.js
+++ b/static/js/flow/enrollmentTypes.js
@@ -13,4 +13,5 @@ export type ProgramEnrollmentsState = {
   getStatus?: string,
   getErrorInfo?: APIErrorInfo,
   postStatus?: string,
+  coursePostStatus?: string,
 };
--- a/static/js/flow/enrollmentTypes.js
+++ b/static/js/flow/enrollmentTypes.js
@@ -13,5 +13,5 @@ export type ProgramEnrollmentsState = {
   getStatus?: string,
   getErrorInfo?: APIErrorInfo,
   postStatus?: string,
-  coursePostStatus?: string,
+  courseEnrollAddStatus?: string,
 };
--- a/static/js/reducers/enrollments_test.js
+++ b/static/js/reducers/enrollments_test.js
@@ -181,7 +181,7 @@ describe('enrollments', () => {
         assert.ok(addCourseEnrollmentStub.calledWith(courseKey));
         assert.ok(fetchCoursePricesStub.calledWith());
         assert.ok(fetchDashboardStub.calledWith());
-      })
+      });
     });
 
     it('should fail to add a course enrollment', () => {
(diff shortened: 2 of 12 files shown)


gsidebo:
functionality looks great. looking at code now

static/js/reducers/enrollments_test.js
```diff
@@ -163,6 +168,36 @@ describe('enrollments', () => {
...
+
+      let courseKey = 'course_key';
+      return dispatchThen(addCourseEnrollment(courseKey), [
+        REQUEST_ADD_COURSE_ENROLLMENT,
+        RECEIVE_ADD_COURSE_ENROLLMENT_FAILURE,
+      ]).then(state => {
+        assert.equal(state.coursePostStatus, FETCH_FAILURE);
+        assert.ok(addCourseEnrollmentStub.calledWith(courseKey));
+        assert.notOk(fetchCoursePricesStub.calledWith());
+        assert.notOk(fetchDashboardStub.calledWith());
```
gsidebo: `assert.isTrue` and `assert.isFalse` would be a lot more clear for the 3 assertions above and the ones at lines 181-183

static/js/flow/enrollmentTypes.js
```diff
@@ -13,4 +13,5 @@ export type ProgramEnrollmentsState = {
   getStatus?: string,
   getErrorInfo?: APIErrorInfo,
   postStatus?: string,
+  coursePostStatus?: string,
```
gsidebo: on first look it seems appropriate to add the state information for course enrollment results to the program enrollments state (they're both enrollments after all), but this is kinda awkward. course enrollments are, after all, very different from program enrollments, and `ProgramEnrollmentsState` was created to reflect information about API calls to program enrollment endpoints. we're just sort of gluing this extra piece of information to this unrelated state when we need to. i guess it's not crucial to create some new state type to store this, and i'm not sure that storing it in DashboardState is any cleaner - if you think of something easy and more sensible to do with this, that'd be great, but i won't hold up the PR for it. if you decide to keep this as part of `ProgramEnrollmentsState`, please change the key to `courseEnrollAddStatus` or something like that. `coursePostStatus` is misleading

static/js/reducers/enrollments_test.js
noisecapella: Sure

static/js/flow/enrollmentTypes.js
noisecapella: I agree that it's a bad place for this information, I'll go with the rename of `coursePostStatus` in the interest of moving things along since I can't think of anything better right now

noisecapella:
Comments addressed

PR #1297, opened 2016-10-07
Expand country code to country name on review page

### LINKED ISSUE #1287: Financial Aid: expand country names in review form
We store learner's countries as ISO codes currently and those are displayed in the review form. Unfortunately, that's not too helpful for financial aid reviewers. 

Inside of "AL" I'd like to see "Albania"

alvinsiu:
Closes #1287 

#### What's this PR do?
Expands the country code on the financial review page to the full country name

Fire up the review page at `/financial_aid/review/<program_id>`.

This uses the handy `codeToCountryName` function implemented in #1275


aliceriot:
looks like the working commits from the branch I was using for #1275 got included, that PR is merged into master now so it might be a good idea to clean that up.

static/js/util/currency.js
```diff
@@ -37,7 +37,7 @@ const codesToOptions = R.compose(
 
 export const currencyOptions = codesToOptions(cc.codes());
 
-const codeToCountryName = code => iso3166.country(code).name || '';
+export const codeToCountryName = code => iso3166.country(code).name || '';
```
aliceriot: this function needs to have the type of it's argument annotated, now that it's exported. if you're not familiar with Flow, it will end up looking like this:

```js
export const codeToCountryName = (code: string) => iso3166.country(code).name || '';
```

aliceriot: functions that are not exported have their types inferred, but for efficient type-checking across modules flow enforces that all exported functions have to have their types annotated.

alvinsiu:
Maybe your browser had a cached version of this PR?  It looks merged to me!

aliceriot:
do you see something different than this under the 'commits' tab?

commits

alvinsiu:
Ah, I just merged master in so the diff is the same.  When merging, I just squash and merge so these commits won't show up anyway - it'll just be a single commit with the changes.

aliceriot:
yep squash and merge is the standard approach. generally I think rebasing is cleaner than merging in master, but it's not a huge issue.

aliceriot:
functionality looks good! looking at code now.

static/js/financial_aid/functions.js
```diff
@@ -1,9 +1,27 @@
 /* eslint-disable no-undef, max-len, no-var, prefer-template, no-unused-vars */
-var financialAidReview = function() {
+global.jQuery = require('jquery');
+
+require("bootstrap");
+
+import { codeToCountryName } from "../util/currency";
+
+window.codeToCountryName = codeToCountryName;
```
aliceriot: it doesn't look like you are using this anywhere, am I missing something?

static/js/financial_aid/functions.js
```diff
@@ -1,9 +1,27 @@
...
+
+/* Wrapper for all of the functions used by the financial aid review page */
+window.financialAidReview = (function(window, document, $) {
```
aliceriot: what's the reason this needs to be on `window`?

static/js/financial_aid/functions.js
alvinsiu: Ah, you're right - that was left over from testing

static/js/financial_aid/functions.js
alvinsiu: This is so that the review page (outside of React) can access it.

static/js/util/currency.js
alvinsiu: Oh, I see - thanks!

static/js/financial_aid/functions.js
aliceriot: Ok I see - there's no React here, but I assume you mean outside of the webpack minified dealio that ends up being executed.

alvinsiu: Yeah, that's what I mean :]

alvinsiu:
Changes requested are implemented!

PR #1300, opened 2016-10-07
Fixed image size for faculty carousel

### LINKED ISSUE #1292: Faculty thumbnail images on the program page are cropping to the wrong size.
The faculty images on the program page are supposed to crop to an aspect ration of 500px wide and 385px tall. Currently they seem to be cropping to a square that is 385px by 385px.

Basically what we are shooting for is that a user uploads a hi-res image that is equal to or larger than 500x385 in both directions, and then they are cropped to the correct aspect ratio.

noisecapella:
Fixes #1292 

#### What's this PR do?
Updates the size to 500x385 to match the help text on the CMS, and also fixes some CSS to allow the user to click the buttons on the left and right sides

screenshot from 2016-10-07 15-56-49

### DIFF
--- a/cms/serializers.py
+++ b/cms/serializers.py
@@ -18,7 +18,7 @@ class FacultyImageSerializer(serializers.ModelSerializer):
     alt = serializers.CharField(source="default_alt_text")
     rendition = serializers.SerializerMethodField()
 
-    def get_rendition(self, image):
+    def get_rendition(self, image):  # pylint: disable=no-self-use
         """Serialize a rendition for the faculty image"""
         return RenditionSerializer().to_representation(image.get_rendition('fill-500x385'))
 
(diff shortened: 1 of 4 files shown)


gsidebo:
just a question: is it wagtail that specifies the Image's title attribute as the default alt text?

noisecapella:
The alt text comes from `default_alt_text` on the wagtail model `Image`, so yes but we can change it if we want in the future. I'll look into the description

noisecapella:
The description tweaking is nontrivial (I'm not sure what the end result should look like) so I'll open another issue for it

roberthouse54:
I have already fixed the text in another PR

noisecapella:
Ok, all comments should be addressed

gsidebo:
👍  ship it!

PR #1303, opened 2016-10-07
Country income threshold database model

### LINKED ISSUE #1301: Move auto-approval income thresholds into database
I think @noisecapella mentioned this last week, but we should move these constants into the database since leaving it in the codebase if it's open sourced could allow people to maliciously use this information.

alvinsiu:
Closes #1301 

#### What's this PR do?
Implements `CountryIncomeThreshold` database model and its associated factory and django admin registration.  Removes the constant and data files.


pdpinch:
Can we use a .csv as a fixture? If not, can you make it into one -- json or yaml, whichever is easier. 

You could commit the fixture with all zero threshholds and then we can update them before we import the fixture.

alvinsiu:
Sure, I'll do that next week!

alvinsiu:
I've pushed a fixture using json with a default threshold of 75,000 so that we don't get any inadvertent auto-approvals from countries that weren't changed by accident.  One caveat to using fixtures here is that `created_on` and `updated_on` timestamps will not reflect the actual time they were created (since fixtures assume you have complete data and will not automatically populate `auto_now` fields).

We can also make this be able to import .csv via a management command instead if that will be more useful than this fixture.  One other benefit of using a management command is that `created_on` and `updated_on` timestamps _will_ reflect the true time they were created.

pdpinch:
It sounds like the effort of adding csv import is worth it.

pdpinch:
@alvinsiu I'm assuming this is ready for review, so I added the label.

giocalitri:
A first view of the code seems fine to me.
I have 2 questions:
1- do we need a migration to load `country_income_threshold_data.json` in the database?
2- given that I have no clue if the number make sense, @pdpinch do you sign on the numbers in `country_income_threshold_data.json`?

financialaid/api_test.py
```diff
@@ -41,7 +40,9 @@ class FinancialAidBaseTestCase(TestCase):
     @classmethod
     def setUpTestData(cls):
         with mute_signals(post_save):
-            cls.profile = ProfileFactory.create()
+            cls.profile = ProfileFactory.create(
+                country=FuzzyText(length=2)
```
giocalitri: if `country` can contain only the country codes, maybe it is better to change the `ProfileFactory` directly.

alvinsiu: I have an issue written for this since it seems like it might be a slightly bigger refactor to ensure country codes are 2 letters long in more than just one model. #1309

alvinsiu:
> 1- do we need a migration to load country_income_threshold_data.json in the database?
This is a fixture, so you can run `python manage.py loaddata country_income_threshold_data`. Then it can be edited via the django admin.

>2- given that I have no clue if the number make sense, @pdpinch do you sign on the numbers in country_income_threshold_data.json?
We defaulted these to 75000 which is the current max for a threshold. The plan was to change the values in django admin. In the future, we can do this via a csv import management command, but that is for after MVP.

financialaid/factories.py
```diff
@@ -80,3 +81,14 @@ def create(cls, **kwargs):
...
     class Meta:  # pylint: disable=missing-docstring
         model = FinancialAid
+
+
+class CountryIncomeThresholdFactory(DjangoModelFactory):
+    """
+    Factory for CountryIncomeThreshold
+    """
+    country_code = FuzzyText(length=2)
+    income_threshold = FuzzyInteger(low=0, high=12345)
```
alvinsiu: Sure, I'll change this.

giocalitri:
>> 1- do we need a migration to load country_income_threshold_data.json in the database?

>This is a fixture, so you can run `python manage.py loaddata country_income_threshold_data`. Then it can be edited via the django admin.

I understand, but given that this fixture is needed at the beginning, probably it is better to load it anyway

alvinsiu:
That I think should be part of DevOps - @pdpinch does it make sense to add loading this fixture as part of #1308?

giocalitri:
I think that given that the app does not work without this data, the initial loading of the fixture should be done in a migration here.
The idea is that if we had data in files used directly by some functions before I do not see why now we should not have the same data in the DB now.

noisecapella:
I agree, it seems like the fixture here is used for testing purposes. We shouldn't have a migration to load the production data, it should be done separate from this code base

alvinsiu:
@giocalitri That makes sense to load the data into the DB via migrations since we're loading dummy data anyway.

@noisecapella Just wanted to double check, did you mean "we _should_ have a migration" or it should be part of DevOps and "done separate from this code base"?  I like @giocalitri's solution of loading the data with migrations since this is all dummy data anyway and will need to be changed via django admin in config anyway.

financialaid/api.py
```diff
@@ -46,13 +44,26 @@ def determine_tier_program(program, income):
...
+        income_threshold = country_income_threshold.income_threshold
+    except CountryIncomeThreshold.DoesNotExist:
+        message = (
+            "Country code {country_code} does not exist in CountryIncomeThreshold for financial "
+            "aid id {financial_aid_id}"
+        ).format(
+            country_code=financial_aid.country_of_income,
+            financial_aid_id=financial_aid.id
+        )
+        log.error(message)
```
giocalitri: log messages should be formatted directly in the log method:
something like
```
log.error((
             "Country code %s does not exist in CountryIncomeThreshold for financial "
             "aid id %s"
         ), financial_aid.country_of_income, financial_aid.id
)

File: financialaid/api_test.py (hunk 1 above)
Comment (giocalitri) [inline financialaid/api_test.py]: ok

File: financialaid/migrations/0016_load_country_income_thresholds.py [hunk 4]
```diff
@@ -0,0 +1,42 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('financialaid', '0015_countryincomethreshold'),
+    ]
+
+    operations = [
+        migrations.RunPython(load_country_income_thresholds),
+    ]
```
giocalitri: can you try to revert this migration? if it does work we do not need to add a "south" migration because this is the first data we load, otherwise we need one.

financialaid/models.py
```diff
@@ -172,3 +172,11 @@ class CurrencyExchangeRate(TimestampedModel):
     """
     currency_code = models.CharField(null=False, max_length=3)
     exchange_rate = models.FloatField(null=False)  # how much foreign currency is per 1 USD
+
+
+class CountryIncomeThreshold(TimestampedModel):
+    """
+    Table of country income thresholds for financial aid auto approval
+    """
+    country_code = models.CharField(null=False, max_length=2)
```
giocalitri: ok

financialaid/migrations/0016_load_country_income_thresholds.py
alvinsiu: Could you explain a bit more for me?  I've never used South with Django so I don't have very much context.

giocalitri:
Small comment from testing: the rest looks good

giocalitri:
👍 after tests pass

PR #1306, opened 2016-10-08
Added sentry to app

blarghmatey:
**Added sentry and coveralls to app**
Began adding Sentry to API and JS for error monitoring.

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -19,7 +19,6 @@ import yaml
 from celery.schedules import crontab
 from django.core.exceptions import ImproperlyConfigured
 
-import raven
 
 VERSION = "0.16.0"
 
@@ -391,7 +390,8 @@ LOGGING = {
 }
 
 # Sentry
-ENVIRONMENT = os.environ.get('MICROMASTERS_ENVIRONMENT', 'dev')
+ENVIRONMENT = get_var('MICROMASTERS_ENVIRONMENT', 'dev')
+SENTRY_CLIENT = 'raven.contrib.django.raven_compat.DjangoClient'
 RAVEN_CONFIG = {
     'dsn': get_var('SENTRY_DSN', ''),
     'environment': ENVIRONMENT,
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -14,10 +14,12 @@ import ast
 import os
 import platform
 
-from celery.schedules import crontab
 import dj_database_url
-from django.core.exceptions import ImproperlyConfigured
 import yaml
+from celery.schedules import crontab
+from django.core.exceptions import ImproperlyConfigured
+
+import raven
 
 VERSION = "0.16.0"
 
@@ -104,6 +106,7 @@ INSTALLED_APPS = (
 
     # other third party APPS
     'rolepermissions',
+    'raven.contrib.django.raven_compat',
 
     # Our INSTALLED_APPS
     'backends',
@@ -120,6 +123,7 @@ INSTALLED_APPS = (
 )
 
 MIDDLEWARE_CLASSES = (
+    'raven.contrib.django.raven_compat.middleware.SentryResponseErrorIdMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
@@ -351,6 +355,11 @@ LOGGING = {
             'filters': ['require_debug_false'],
             'class': 'django.utils.log.AdminEmailHandler'
         },
+        'sentry': {
+            'level': 'ERROR',
+            'class': 'raven.contrib.django.raven_compat.handlers.SentryHandler',
+            'formatter': 'verbose'
+        }
     },
     'loggers': {
         'root': DEFAULT_LOG_STANZA,
@@ -374,7 +383,19 @@ LOGGING = {
         'elasticsearch': {
             'level': ES_LOG_LEVEL,
         },
+        'raven': {
+            'level': 'DEBUG',
+            'handlers': []
+        }
     },
+}
+
+# Sentry
+ENVIRONMENT = os.environ.get('MICROMASTERS_ENVIRONMENT', 'dev')
+RAVEN_CONFIG = {
+    'dsn': get_var('SENTRY_DSN', ''),
+    'environment': ENVIRONMENT,
+    'release': VERSION
 }
 
 # to run the app locally on mac you need to bypass syslog
--- a/package.json
+++ b/package.json
@@ -21,12 +21,6 @@
     "css-loader": "0.23.1",
     "currency-codes": "^1.1.2",
     "enzyme": "^2.4.1",
-    "eslint": "^2.13.1",
-    "eslint-config-defaults": "9.0.0",
-    "eslint-plugin-babel": "^3.3.0",
-    "eslint-plugin-flow-vars": "^0.4.0",
-    "eslint-plugin-flowtype": "^2.3.0",
-    "eslint-plugin-react": "5.0.1",
     "fetch-mock": "4.4.0",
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -100,7 +92,20 @@
...
+    "eslint-config-defaults": "9.0.0",
+    "eslint-plugin-babel": "^3.3.0",
+    "eslint-plugin-flow-vars": "^0.4.0",
+    "eslint-plugin-flowtype": "^2.3.0",
+    "eslint-plugin-react": "5.0.1",
+    "istanbul": "1.0.0-alpha.2",
+    "coveralls": "^2.11.14",
+    "mocha": "2.4.5",
+    "mocha-lcov-reporter": "^1.2.0",
+    "webpack-dev-server": "1.14.1"
```
aliceriot: we don't use `devDependencies` because the version of packages in `devDependencies` cannot be locked down with `npm shrinkwrap`. please put things back as they were :)

package.json
```diff
@@ -36,18 +30,17 @@
...
     "jsdom-global": "1.7.0",
     "keycode": "2.1.1",
     "lodash": "4.11.2",
     "material-ui": "0.15.0",
-    "mocha": "2.4.5",
     "moment": "2.13.0",
     "node-sass": "3.7.0",
     "object.entries": "1.0.3",
     "ramda": "^0.22.1",
+    "raven-js": "^3.7.0",
```
aliceriot: additionally, it looks like your installation of `raven-js` isn't reflected in `shrinkwrap.json`. here's what you'll probably need to do to get that to work:

```
git checkout master -- package.json
# restart docker-compose
docker-compose run watch npm prune
docker-compose run watch npm install --save raven-js
```

then check in the changes to `package.json` and `shrinkwrap.json`. That should do it!

micromasters/settings.py
```diff
@@ -374,9 +383,21 @@ def get_var(name, default):
...
         },
+        'raven': {
+            'level': 'DEBUG',
+            'handlers': []
+        }
     },
 }
 
+# Sentry
+ENVIRONMENT = os.environ.get('MICROMASTERS_ENVIRONMENT', 'dev')
```
noisecapella: Why not use `DEBUG` here?

package.json
aliceriot: it looks like you added `coveralls` as well, that should be handled the same way.

package.json
```diff
@@ -100,7 +92,20 @@
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",
     "scss_lint": "node ./node_modules/sass-lint/bin/sass-lint.js --verbose --no-exit",
     "test": "./js_test.sh",
-    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --compilers js:babel-register static/js/global_init.js 'static/**/*/*_test.js'",
+    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha --report lcovonly -- --compilers js:babel-register static/js/global_init.js 'static/**/*/*_test.js' && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
     "flow": "flow check"
+  },
+  "devDependencies": {
+    "eslint": "^2.13.1",
```
noisecapella: Can you keep these in `dependencies`? We've decided to keep everything there for simplicity's sake, and since webpack will already exclude the packages we don't use

package.json
```diff
@@ -100,7 +92,20 @@
     "lint": "node ./node_modules/eslint/bin/eslint.js ./static/js",
     "scss_lint": "node ./node_modules/sass-lint/bin/sass-lint.js --verbose --no-exit",
     "test": "./js_test.sh",
-    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --compilers js:babel-register static/js/global_init.js 'static/**/*/*_test.js'",
+    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha --report lcovonly -- --compilers js:babel-register static/js/global_init.js 'static/**/*/*_test.js' && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
```
noisecapella: Would the user still be able to run coverage locally on their dev machine if they didn't have coveralls set up?

static/js/sentry_client.js
```diff
@@ -0,0 +1,6 @@
+import Raven from 'raven-js';
+
+Raven.config(SETTINGS.sentry_dsn, {
+    release: SETTINGS.release_version,
+    environment: SETTINGS.environment
+}).install();
```
aliceriot: does this pass the linter? we indent JS two spaces.

aliceriot: you can run the linter by doing:

```
docker-compose run watch npm run lint
```

I have that aliased to `dlint`.

package.json
aliceriot: this is a nice change! changes like this though (to testing commands and so on) we generally propose through an issue first.

@noisecapella @gsidebo what do y'all think?

ui/templates/base.html
```diff
@@ -6,11 +6,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,400i,500,700" />
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
+    {% load raven %}
```
noisecapella: What does this do?

package.json
aliceriot: to your first question, no, it gave me an error when I tried it.

noisecapella:
Just had a few comments, leaving this to @aliceriot

aliceriot:
I was thinking I would just comment on the JS! So your input is most welcome @noisecapella  :)

package.json
aliceriot: maybe we could move this to a new command, perhaps `coveralls`?

noisecapella:
I think we need the context updated in `terms_of_service`

micromasters/settings.py
blarghmatey: The environment flag is a way of tracking the stage of the pipeline that an error occurs in for Sentry. I'm not sure what you mean about using `DEBUG`.

package.json
blarghmatey: Done

package.json
blarghmatey: I updated this to be a separate command for use in the travis.yml

.travis.yml
```diff
@@ -7,7 +7,7 @@ install:
 
 script:
   - docker-compose -f travis-docker-compose.yml run web tox
-  - docker-compose -f travis-docker-compose.yml run watch npm run-script coverage
+  - docker-compose -f travis-docker-compose.yml run watch npm run-script coveralls
```
aliceriot: will this still show some form of coverage report on travis?

micromasters/settings.py
noisecapella: Ah, I thought it was a production/development flag

aliceriot:
All the JS stuff looks good to me! @noisecapella do you wanna take a look at the python stuff or should I?

test_requirements.txt
```diff
@@ -1,14 +1,15 @@
-ipython
 bpython
+coveralls
```
singingwolfboy: If you're removing coveralls from this pull request, you should remove it from this file as well.

noisecapella:
Looks good to me aside from @singingwolfboy's comment :+1:

test_requirements.txt
blarghmatey: Done

PR #1313, opened 2016-10-11
Add ga tracking to program pages

### LINKED ISSUE #1154: Per-program Google Analytics
MicroMasters teams are going to want some custom analytics -- at least for their program page -- but we don't want everyone to see each other's traffic. 

We can solve this by creating a GA property for each program and then adding it to appropriate pages. For now, that means the program page. 

- [x] @pdpinch will create GA properties for SCM and DEDP (and maybe a little research on whether this can be automated)
- [ ] Add a field to the program model to store the GA Tracking Id
- [ ] Add the program's Tracking Id to the program template (in addition to the existing MM tracking ID)

annagav:
Fix #1154
#### What's this PR do?
Conditionally create ga tracking object for program pages.

When creating a program set the `ga_tracking_id` to a certain value:
SCM: UA-85136501-1
DEDP: UA-85136501-2
Need to ask Peter for help with testing this.


ui/templates/base.html
```diff
@@ -29,6 +29,9 @@
     })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
     ga('create', '{{ GA_TRACKING_ID }}', 'auto');
+    if('{{ tracking_id }}'){
```
noisecapella: Can you use a Django template `if` block instead?

ui/templates/base.html
```diff
@@ -29,6 +29,9 @@
...
+      ga('create', '{{ tracking_id }}', 'auto');
```
noisecapella: It looks like this takes a fourth parameter which is required when creating multiple trackers. Can you create some name including the program id for that parameter? https://developers.google.com/analytics/devguides/collection/analyticsjs/creating-trackers#working_with_multiple_trackers

courses/models.py
```diff
@@ -20,6 +20,7 @@ class Program(models.Model):
     live = models.BooleanField(default=False)
     description = models.TextField(blank=True, null=True)
     financial_aid_availability = models.BooleanField(default=False, null=False)
+    ga_tracking_id = models.CharField(max_length=255, blank=True, null=True, default="")
```
noisecapella: It seems like we wouldn't want to allow `null` here, is it possible to remove `null=True`, or is it necessary to allow empty strings?

ui/templates/base.html
```diff
@@ -29,6 +29,9 @@
...
+    }
     ga('send', 'pageview');
```
noisecapella: It seems like we should also do this for the other tracker after we create it, in the if block above

annagav:
@noisecapella, I think I addressed all the comments.

noisecapella:
Yes, functionality works great :+1:

Issue #1311, opened 2016-10-11
Use calculated income not original income for determining tier


pdpinch:
## Steps to reproduce

1. Click on Calculate Your Cost
2. Enter 100000 for income and Albanie Lek for currency. (equivalent to about $813/yr in USD)
3. Click Submit

## Expected Result

Lowest-price tier: $100 per course

## Actual Result

Highest-price tier: $1000 per course

I suspect the issue is that we're using `original_income` instead of `income_usd` for calculating the tier. 

See https://github.com/mitodl/micromasters/blob/master/financialaid/serializers.py#L70

pdpinch:
closed via  6af03a6

PR #1314, opened 2016-10-11
Fixed bug in financial aid request for determining tier

alvinsiu:
#1311 

#### What's this PR do?
Fixes a bug where the incorrect tier is assigned if submitted currency is not USD.  Also adds new and fixes existing tests to test for this case.

### DIFF
--- a/financialaid/api_test.py
+++ b/financialaid/api_test.py
@@ -41,7 +41,9 @@ class FinancialAidBaseTestCase(TestCase):
     @classmethod
     def setUpTestData(cls):
         with mute_signals(post_save):
-            cls.profile = ProfileFactory.create()
+            cls.profile = ProfileFactory.create(
+                country="US"
+            )
             cls.profile2 = ProfileFactory.create()
             cls.staff_user_profile = ProfileFactory.create()
             cls.staff_user_profile2 = ProfileFactory.create()
--- a/financialaid/api_test.py
+++ b/financialaid/api_test.py
@@ -41,9 +41,7 @@ class FinancialAidBaseTestCase(TestCase):
     @classmethod
     def setUpTestData(cls):
         with mute_signals(post_save):
-            cls.profile = ProfileFactory.create(
-                country="US"
-            )
+            cls.profile = ProfileFactory.create()
             cls.profile2 = ProfileFactory.create()
             cls.staff_user_profile = ProfileFactory.create()
             cls.staff_user_profile2 = ProfileFactory.create()
--- a/financialaid/views_test.py
+++ b/financialaid/views_test.py
@@ -18,7 +18,6 @@ from ecommerce.factories import CoursePriceFactory
 from financialaid.api import determine_tier_program, determine_income_usd
 from financialaid.api_test import FinancialAidBaseTestCase
 from financialaid.constants import (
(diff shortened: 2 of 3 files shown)


financialaid/api_test.py
```diff
@@ -41,7 +41,9 @@ class FinancialAidBaseTestCase(TestCase):
     @classmethod
     def setUpTestData(cls):
         with mute_signals(post_save):
-            cls.profile = ProfileFactory.create()
+            cls.profile = ProfileFactory.create(
+                country="US"
```
noisecapella: What does this do? Do we need to set the country to `US` for any of the other profiles created in this function?

alvinsiu: This isn't strictly needed, but I wanted to ensure that this profile's auto-approval income threshold was the highest possible (75k).  Currently, this will always be the case since the fuzzy generator for profiles creates the country code with a suffix of "land", which will never match the two character county code.  I'm anticipating a future refactor where two letter country codes will be better enforced, and I'd rather not leave the potential for the generator to match an existing auto-approval country and fail tests.

Now given that, these tests likely need a refactor anyway so I'm happy to leave out this change in favor of doing it in a later refactor.  Thoughts?

financialaid/views_test.py
```diff
@@ -128,24 +131,43 @@ def test_income_validation_user_not_enrolled(self):
         self.program_enrollment.save()
         self.assert_http_status(self.client.post, self.request_url, status.HTTP_400_BAD_REQUEST, data=self.data)
 
-    def test_income_validation_currency_not_usd(self):
+    def test_income_validation_currency_not_usd_gto(self):
         """
-        Tests IncomeValidationView post with a currency that is not USD
+        Tests FinancialAidRequestView post with a currency that is not USD with exchange rate greater than 1
```
alvinsiu: This bug was never caught in tests because the currency exchange always resulted in an income that led to the same tier assignment.  Having both of these explicitly mitigates tests not covering this case in the future.

financialaid/api_test.py
noisecapella: Leaving it for a future refactor sounds good to me

financialaid/views_test.py
noisecapella: I'm not sure of the logic here, it's the income which ultimately determines the tier. The exchange rate being less or greater than 1 doesn't matter when you can set the input income to any value. But feel free to leave it since it's just an extra test

financialaid/api_test.py
alvinsiu: Sure, will remove this line.

noisecapella:
Functionality works great :+1:

financialaid/views_test.py
alvinsiu: The issue was that we should be using `income_usd` to calculate the tier, but instead we used `original_income`, and since the original test for currency conversion didn't convert to a value that would change which tier they would have been assigned with `original_income`, we never caught this bug.

I'll leave it in for now!

alvinsiu:
Will merge once it passes tests, thanks!

pdpinch:
@alvinsiu you forgot the magic word(s): fixes/closes/etc

alvinsiu:
Oh whoops!  Good catch!

PR #1332, opened 2016-10-12
Add FAQs Categories

### LINKED ISSUE #1190: Program pages:  FAQ categories

annagav:
Fix #1190
#### What's this PR do?

Created CategorizedFaqsPage to be displayed within the FaqsPage.
cms/models.py, facs_page.html

For a program page add a child `FacsPage`, where the title will be the the tab name. Use this `FaqsPage` to add few children of type `CategorizedFaqsPage` where the title is the category name. For each of those pages you can add any number of `FrequentlyAskedQuestions` that are relevant to this category.

screen shot 2016-10-12 at 11 07 14 am


giocalitri:
it works as expected.
the only doubt I have is about the sorting: as far as I can see there is no explicit sorting of the FAQ categories and I might have seen (but cannot reproduce) a re-sorting after changing something.

pdpinch:
Good point. Can we open a separate issue for that? Maybe we can get it done by tomorrow, but I want to make sure this gets done.

giocalitri:
ok, I created https://github.com/mitodl/micromasters/issues/1371, so 👍

PR #1346, opened 2016-10-12
Fixed bug with current grade refresh and no enrollments

giocalitri:
N/A
#### What's this PR do?
If the user did not have any enrollment, the current grades in the cache were not created leading to possible problems (the code expects entries with `null` data anyway).

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -377,15 +377,20 @@ def _check_if_refresh(user, cached_model, refresh_delta, limit_to_courses=None):
         Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
     ).values_list("edx_course_key", flat=True)
 
+    all_course_count = course_ids.count()
+
     if limit_to_courses is not None:
         course_ids = course_ids.filter(edx_course_key__in=limit_to_courses)
 
     model_queryset = cached_model.objects.filter(
         user=user,
         last_request__gt=refresh_delta,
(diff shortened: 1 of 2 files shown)


dashboard/api.py
```diff
@@ -377,15 +377,20 @@ def _check_if_refresh(user, cached_model, refresh_delta, limit_to_courses=None):
...
     model_queryset = cached_model.objects.filter(
         user=user,
         last_request__gt=refresh_delta,
-        course_run__edx_course_key__in=course_ids,
     )
-    return model_queryset.count() == len(course_ids), model_queryset, course_ids
+    all_cached_elements_count = model_queryset.count()
+    model_queryset = model_queryset.filter(course_run__edx_course_key__in=course_ids)
+
+    is_data_fresh = model_queryset.count() == len(course_ids) and all_cached_elements_count == all_course_count
```
aliceriot: I'm a bit confused here, is there a situation where `all_course_count` would not equal `len(course_ids)`?

giocalitri: There are many cases I found, but the simplest to explain is when we add a new course_run in Micromasters and the user is not enrolled. Given that she is not enrolled before we never refreshed the current grade for that run and we never created an "empty" entry in the cache either.

aliceriot: The cached vs. actual course runs comparison makes sense. I was more asking about `len(course_ids)` vs `course_ids.count()`, which seems like they should be capturing the same quantity.

PR #1352, opened 2016-10-12
Handle invalid dates

### LINKED ISSUE #1294: Handle invalid dates:
screen shot 2016-10-07 at 1 31 27 pm

aliceriot:

closes #1294 

#### What's this PR do?

This deals with invalid dates in the `CourseDescription` and `CourseAction` components, basically by testing to see if the date is valid before we try to format and display it.

What I did was alter the `getDashboard`, `getProgramEnrollments` and `getCoursePrices` functions in `api.js` to return their respective constants (from `constants.js`). Then you can alter relevant courseRuns in order to make invalid dates, to test that every case is covered. A good string to use is `'1999-13-92'`.

### DIFF
--- a/static/js/components/dashboard/CourseAction_test.js
+++ b/static/js/components/dashboard/CourseAction_test.js
@@ -22,7 +22,11 @@ import {
   FA_PENDING_STATUSES,
   FA_STATUS_SKIPPED
 } from '../../constants';
-import { findCourse, findAndCloneCourse } from '../../util/test_utils';
+import {
+  findCourse,
+  alterFirstRun,
+  findAndCloneCourse
+} from '../../util/test_utils';
 
 describe('CourseAction', () => {
   const now = moment();
@@ -62,10 +66,6 @@ describe('CourseAction', () => {
       descriptionText: descriptionText,
       linkText: linkText
     };
-  };
-
-  let alterFirstRun = (course, overrideObject) => {
-    course.runs[0] = Object.assign({}, course.runs[0], overrideObject);
   };
 
   let assertCheckoutButton = (button, courseId) => {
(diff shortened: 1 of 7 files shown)


static/js/components/dashboard/CourseAction_test.js
```diff
@@ -117,6 +117,17 @@ describe('CourseAction', () => {
     assertCheckoutButton(elements.button, firstRun.course_id);
   });
 
+  it('hides an invalid date with STATUS_OFFERED', () => {
+    let course = findCourse(course => (
+      course.runs.length > 0 &&
+      course.runs[0].status === STATUS_OFFERED
+    ));
+    alterFirstRun(course, {enrollment_start_date: '1999-13-92'});
```
noisecapella: `alterFirstRun` will modify `course` so you'll need to clone it first. Can you change `findCourse` to `findAndCloneCourse` above?

static/js/util/date.js
```diff
@@ -0,0 +1,5 @@
+import R from 'ramda';
+
+export const ifValidDate = R.curry((def, fn, date) => (
```
noisecapella: Can you write a test for this in `date_test.js`?

static/js/components/dashboard/CourseAction_test.js
noisecapella: This also needs to be done in the other two tests added below

static/js/util/date.js
aliceriot: oh yeah I meant to, just slipped my mind :)

aliceriot:
@noisecapella comments addressed!

noisecapella:
Looks good :+1:

PR #1355, opened 2016-10-12
Limit HomePage to have only ProgramPage as a child page

### LINKED ISSUE #1279: CMS: Limit options for child pages
Currently in the CMS you can add a page of any type as a child of any other type. This doesn't make sense. 

screen shot 2016-10-06 at 2 03 32 pm

- [ ] The home page should allow children of type program. 
- [x] Currently, the program page shouldn't allow children of any type. However, #1204 and #597 would presumably add new page types.

annagav:
Fix #1279
#### What's this PR do?
Allow only ProgramPage type to be added as a child to HomePage in cms.

To test this try to add a child page to the homepage.
screen shot 2016-10-12 at 4 23 27 pm

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -33,6 +33,7 @@ class HomePage(Page):
     CMS page representing the homepage.
     """
     content_panels = []
+    subpage_types = ['ProgramPage']
 
     def get_context(self, request):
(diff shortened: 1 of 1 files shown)


pdpinch:
👍

PR #1356, opened 2016-10-12
Program enrollment does not return error if already exists

giocalitri:
N/A

#### What's this PR do?
Modified the program enrollment rest API to not return an error in case an enrollment for the same program already exists.

### DIFF
--- a/courses/views.py
+++ b/courses/views.py
@@ -85,20 +85,19 @@ class ProgramEnrollmentListView(ListCreateAPIView):
         if not isinstance(program_id, int):
             raise ValidationError('A `program_id` parameter must be specified')
 
-        if ProgramEnrollment.objects.filter(user=request.user, program__pk=program_id).exists():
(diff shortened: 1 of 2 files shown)


aliceriot:
Functionality looks good! I can go through the profile flow and add a program I'm already enrolled in and there is no error.

courses/views.py
```diff
@@ -85,8 +85,14 @@ def create(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
 
+        serializer = self.get_serializer_class()
+
         if ProgramEnrollment.objects.filter(user=request.user, program__pk=program_id).exists():
-            raise ResourceConflict('The enrollment for the specified program already exists')
+            program = Program.objects.get(pk=program_id)
+            return Response(
+                status=status.HTTP_200_OK,
+                data=serializer(program).data
+            )
```
aliceriot: could you instead use `Program.objects.get_or_create` below?

PR #1373, opened 2016-10-13
Fix sharing buttons on error pages

### LINKED ISSUE #1245: base_error.html needs header to be updated
The popup isn't showing for the 'sign up with edx' link on `base_error.html`. I don't think we care too much about how this page looks but if we're relying on the user having a link to the terms of service we should add the popup, or alternatively remove the sign up link completely

We decided that we should remove the pop-up from the error template.

### LINKED ISSUE #1293: 404 page UI is broken
screen shot 2016-10-07 at 1 06 55 pm

## Related Issue

#1299

annagav:
Fix #1293 
Fix #1245
#### What's this PR do?
Updated base_error.html

screen shot 2016-10-13 at 5 17 25 pm


noisecapella:
There's no meta tag information on the error pages, though I'm not sure how often people will want to share 404 or 500 pages. Is that a concern @pdpinch?

noisecapella:
Looks great :+1:

annagav:
@noisecapella, I changed the template to include the social buttons only `public_src` is provided. So now the error pages and "Terms of Service" page should not include the sharing buttons.

annagav:
screen shot 2016-10-14 at 12 10 01 pm

noisecapella:
Looks good :+1:

PR #1376, opened 2016-10-13
Modified mail.views responses to catch 401 status codes from mailgun …

### LINKED ISSUE #1368: Failed mailgun calls result in exception raised instead of returning a response with proper formatting
The calls made to mailgun in `mail.api.MailgunClient` don't actually return json if there is an error, so the calls to `mail.views` when there is an error will actually raise an exception due to `response.json()`.

alvinsiu:
Closes #1368 

#### What's this PR do?
This PR modifies `mail.views` to raise `ImproperlyConfigured` if the response from mailgun returns 401. This fixes the bug where `response.json()` normally returns a dictionary, except for a mailgun 401 response which would raise an exception that the 401 response is not JSON-serializable.

3 tests were also added to cover this case.

You can test this out in `/financial_aid/review/<program_id>` on your local machine _without_ a valid mailgun key.  Instead of returning a blob of html dump, it should return a 500 response from `micromasters.utils.custom_exception_handler` with the proper message.

Since this adds an extra function, I created a new file `mail.utils` for the new function and also moved `mail.api.generate_financial_aid_email` over there.

…response to raise ImproperlyConfigured instead

### DIFF
--- a/mail/utils.py
+++ b/mail/utils.py
@@ -75,8 +75,15 @@ def generate_mailgun_response_json(response):
     Returns:
         dict
     """
-    if response.status_code == status.HTTP_401_UNAUTHORIZED:
-        message = "Mailgun API keys not properly configured."
-        log.error(message)
-        raise ImproperlyConfigured(message)
-    return response.json()
+    try:
+        response_json = response.json()
+    except ValueError:  # Includes JSONDecodeError since it inherits from ValueError
+        if response.status_code == status.HTTP_401_UNAUTHORIZED:
+            message = "Mailgun API keys not properly configured."
+            log.error(message)
+            raise ImproperlyConfigured(message)
+        else:
+            response_json = {
+                "message": response.reason
+            }
+    return response_json
--- a/mail/utils.py
+++ b/mail/utils.py
@@ -75,15 +75,14 @@ def generate_mailgun_response_json(response):
     Returns:
         dict
     """
+    if response.status_code == status.HTTP_401_UNAUTHORIZED:
+        message = "Mailgun API keys not properly configured."
+        log.error(message)
+        raise ImproperlyConfigured(message)
     try:
         response_json = response.json()
     except ValueError:  # Includes JSONDecodeError since it inherits from ValueError
-        if response.status_code == status.HTTP_401_UNAUTHORIZED:
-            message = "Mailgun API keys not properly configured."
-            log.error(message)
-            raise ImproperlyConfigured(message)
-        else:
-            response_json = {
-                "message": response.reason
-            }
+        response_json = {
+            "message": response.reason
+        }
     return response_json
--- a/mail/utils.py
+++ b/mail/utils.py
@@ -1,0 +1,82 @@
+"""
+Utils for mail
+"""
+import logging
+
+from django.core.exceptions import (
+    ImproperlyConfigured,
+    ValidationError
+)
+from rest_framework import status
+
+from dashboard.models import ProgramEnrollment
+from financialaid.api import get_formatted_course_price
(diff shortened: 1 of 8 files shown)


mail/utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Args:
+        response (requests.Response): response object
+    Returns:
+        dict
+    """
+    if response.status_code == status.HTTP_401_UNAUTHORIZED:
+        message = "Mailgun API keys not properly configured."
+        log.error(message)
+        raise ImproperlyConfigured(message)
```
giocalitri: I would avoid to match on the status code. In my opinion it would be better to try to parse as JSON and if it fails try to understand the status code and the type of error and build a JSON response anyway. 
In this way we could be safe against any problem caused by mailgun.

mail/views.py
```diff
@@ -88,4 +89,4 @@ def post(self, request, *args, **kwargs):
             subject=serializer.data['email_subject'],
             financial_aid=financial_aid
         )
-        return Response(data=mailgun_response.json(), status=mailgun_response.status_code)
+        return Response(data=generate_mailgun_response_json(mailgun_response), status=mailgun_response.status_code)
```
giocalitri: Nit: please move this to multiple rows.

mail/utils.py
alvinsiu: That's a fair point, but a 401 error should be something fixed by an admin/devops, not something that's really useful for a user trying to send an email.  For that reason, I felt that `ImproperlyConfigured` was a better option to go with since it's more reflective of the actual error.  Otherwise, if we simply catch `JSONDecodeError` and return a 401 message, I feel that it isn't very helpful to the user.

Thoughts?

alvinsiu: Alright, that sounds good to me!

alvinsiu: Hmm just realized a problem with this - we should be raising `ImproperlyConfigured` if there's a 401 response regardless, since that's not something the user can fix.  If we do it this way, if there's a valid `response.json()` call, even if it's a 401, it won't raise `ImproperlyConfigured`.

So either, we should instead do something more like
```
if response.status_code = status.HTTP_401_UNAUTHORIZED:
    ...
try:
    response_json = response.json()
except ValueError:
    response_json = ...
```

giocalitri: Maybe we are going in too deep details, but given that what I was proposing included dealing with the status code and propagating back the error, I think the 401 is a just a special case where you raise instead of returning.
Anyway, I can accept this solution if you return a proper error code with the json and not just return a 200 in `views.py`

alvinsiu:
OK to merge?

mail/utils.py
giocalitri: ok

giocalitri:
can you add some tests for errors that are not 401?

alvinsiu:
Added a test - can I merge when tests pass?

mail/utils_test.py
```diff
@@ -99,15 +99,28 @@ def test_generate_mailgun_response_json(self):
...
+        response_value_error = Mock(
+            spec=Response,
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            json=lambda: (_ for _ in []).throw(ValueError),  # To get .json() to throw ValueError
+            reason="reason"
+        )
+        self.assertDictEqual(
+            generate_mailgun_response_json(response_value_error),
+            {"message": response_value_error.reason}
+        )
```
giocalitri: The new test is good, but please put it in another test method

alvinsiu:
Is it okay to merge?

giocalitri:
👍

Issue #1237, opened 2016-10-05
Truncate the Course Descriptions on the Program page, with a link to Read More on Edx


roberthouse54:
The course descriptions are a bit too long on the Program Page, and we also want to let users know that the link will go to edX. 

The solution is to show the description in a popup when the user clicks the description link. This popup will include the course name, the description and a link to "View on edX". Clicking outside of the popup should make it close. For mobile, the popup should take up the full width and height of the screen (minus a margin all around), and be scrollable, with an "X" to close the popup for mobile view.

annagav:
What is the limit on the length of the paragraph?

pdpinch:
Looks like 150 characters to me.

roberthouse54:
Let's test it and see which character length works best.

roberthouse54:
@ferdi Do we want to switch to a popup for this instead? Like this:

![course_desc_popup](https://cloud.githubusercontent.com/assets/20047260/19278030/d902b912-8fa9-11e6-9bbc-5bec10c7cad7.jpg)

Ferdi:
Considering the current overall design and everything, I personally think this is a better option.

pdpinch:
@roberthouse54 can you update the description to match your new image?

@annagav make sure you're using the latest designs when you work on this

roberthouse54:
See updated description above.

singingwolfboy:
@roberthouse54: the UI element that you're using in the mockup above resembles a tooltip, rather than a popup. [Material Design Lite has a tooltip component](https://material.google.com/components/tooltips.html), but the official usage recommendations don't match with how you've mocked this up. In particular, MDL recommends _not_ using a directional arrow, and not filling tooltips with rich text and a lot of information. They seem to be designed as short labels, rather than as fully-featured components of a UI. [Bootstrap also has a tooltip component](http://getbootstrap.com/javascript/#tooltips), but it also recommends using them for short labels.

I believe that the standard UI element for displaying rich content like this is a modal window. [MDL has a "dialog" component that does this](https://material.google.com/components/dialogs.html), [and Bootstrap calls it a "modal"](http://getbootstrap.com/javascript/#modals). Modals are more conducive to rich content, and it's easier to read this content long-form, since there's an explicit "dismiss" action to close the modal. (Tooltips usually disappear when the mouse moves away from the element that is being described, so it's easy to close them by accident.)

Would you like me to implement this as a tooltip, the way you have in the mockup above? Or should I use a modal instead?

[...]

PR #1378, opened 2016-10-13
Serialize program courses to SETTINGS object

singingwolfboy:
This pull request causes the courses in a program to be serialized into the Javascript `SETTINGS` object on the program page. This pull request is laying the groundwork for a second pull request to address #1237, with the intent of using the [Material-UI Popover component](http://www.material-ui.com/#/components/popover). That future pull request will render the course listing on the page using React, which requires having this information available in Javascript.

Note that it would be better to serialize the course run dates to Javascript as well, and construct the enrollment text on the client side. However, in the interest of not making this pull request any larger than it already is, I kept the enrollment test logic on the server side, and it can be modified later.

Also note that the `get_course_enrollment_text()` and `get_course_url()` functions have been folded into the `Course` model as properties, which makes it easier to retrieve this data for serialization purposes, and also makes more sense. The existing tests for these functions have been moved, but are mostly still intact -- I only deleted one or two that no longer made sense.

This pull request should not cause any user-visible changes to how the site works.


dashboard/utils.py
```diff
@@ -55,12 +55,12 @@ def __init__(self, user, program, enrollments, current_grades, certificates):
         self.financial_aid_available = program.financial_aid_availability
 
         with transaction.atomic():
-            self.course_ids = list(CourseRun.objects.filter(course__program=self.program).exclude(
+            self.course_ids = set(CourseRun.objects.filter(course__program=self.program).exclude(
```
singingwolfboy: The changes in this file are unrelated to the rest of the pull request, but were necessary to make the tests pass. This query does not have an `ORDER BY` clause, so the order of the results is undefined. Some of our tests were comparing this `course_ids` property to a list, where the ordering is important: these tests were only passing coincidentally. My other changes apparently made the database decide to return query results in a different order, which is totally legal, but caused those tests to fail.

I checked with @giocalitri, who confirmed that this variable is only used to check whether a course ID is present in this list or not, so a set is a better data type to use, anyway. Sets can be compared without regard to ordering, so it doesn't matter what order the database decides to return the results.

courses/factories.py
```diff
@@ -45,27 +44,38 @@ def _setup_next_sequence(cls):
...
-    end_date = fuzzy.FuzzyDateTime(datetime(1980, 1, 1, tzinfo=pytz.utc))
-    fuzzy_start_date = fuzzy.FuzzyText(prefix="Starting ")
-    fuzzy_enrollment_start_date = fuzzy.FuzzyText(prefix="Enrollment starting ")
-    enrollment_url = fuzzy.FuzzyText(prefix="http://")
-    prerequisites = fuzzy.FuzzyText(prefix="CourseRun requires ")
+    edx_course_key = factory.LazyAttribute(
+        lambda x: "course:/()+&/" + FAKE.sentence()
+    )
+    enrollment_start = factory.LazyAttribute(
+        lambda x: FAKE.date_time_this_month(before_now=True, after_now=False, tzinfo=pytz.utc)
```
noisecapella: It seems like these changes make sure all course runs are currently running, but I think we want to generate course runs in the past or future too. Though we don't use these values generated by the factory at the moment

noisecapella:
Looks good overall, courses are appearing in SETTINGS

courses/factories.py
```diff
@@ -45,27 +44,38 @@ def _setup_next_sequence(cls):
...
+        lambda x: FAKE.date_time_this_month(before_now=False, after_now=True, tzinfo=pytz.utc)
+    )
+    end_date = factory.LazyAttribute(
+        lambda x: FAKE.date_time_this_year(before_now=False, after_now=True, tzinfo=pytz.utc)
+    )
+    fuzzy_start_date = factory.LazyAttribute(
+        lambda x: "Starting " + FAKE.sentence()
+    )
+    fuzzy_enrollment_start_date = factory.LazyAttribute(
+        lambda x: "Enrollment starting" + FAKE.sentence()
```
singingwolfboy: Fixed.

courses/factories.py
singingwolfboy: Making course runs that are currently running seems like the right default to me. It's easy to ask the factory to create course runs that are not running -- just specify the values you want to set in the `.create()` function.

courses/serializers_test.py
```diff
@@ -2,11 +2,40 @@
...
+    Tests for CourseSerializer
+    """
+
+    def test_course(self):  # pylint: disable=no-self-use
+        """
+        Make sure course serializes correctly
+        """
+        course = CourseFactory.create()
+        result = CourseSerializer().to_representation(course)
+        assert result['title'] == course.title
```
singingwolfboy: Done.

requirements.txt
```diff
@@ -10,8 +10,8 @@ djangorestframework==3.3.2
 edx-api-client==0.3.0
 elasticsearch==2.3.0
 elasticsearch-dsl==2.1.0
-factory_boy==2.6.0
-fake-factory==0.5.3
+factory_boy==2.7.0
```
singingwolfboy: I needed to update it in order to use the `tzinfo` parameter of `FAKE.date_time_this_month()` function.

ui/views_test.py
```diff
@@ -374,26 +374,6 @@ def test_program_thumbnail(self):
         resp = self.client.get('/')
         self.assertContains(resp, image.get_rendition('fill-690x530').url)
 
-    def test_courses_enrollment_text(self):
```
singingwolfboy: No, I dropped this test. It looks like it's asserting that the `get_course_enrollment_text()` function is called once for every course object. However, now that this function has been rolled into the `Course` model, it doesn't really make sense to assert on that. Maybe we could accomplish the same thing by writing a test to ensure that every Course model's enrollment text is present on the page -- but that seems like a very specific, rather brittle thing to test on. Do you think that's a good idea? Anything else I should test instead?

noisecapella: I mostly just care that we are rendering the courses in the correct order (using `position_in_program`). Can you write a test for that if one doesn't exist?

courses/factories.py
noisecapella: Sounds good

ui/views_test.py
singingwolfboy: I added a new test to verify this. Since we'll be switching away from rendering the course listing in HTML at all, I wrote a test that verifies that the courses are ordered correctly in the `SETTINGS` JS variable. Can you take a look, and tell me if it makes sense?

ui/views_test.py
```diff
@@ -374,6 +396,30 @@ def test_program_thumbnail(self):
...
+        courses = [
+            CourseFactory.create(
+                program=self.program_page.program,
+                position_in_program=i,
+            )
+            for i in range(5)
+        ]
+        # render the page
+        response = self.client.get(self.program_page.url)
+        js_settings = get_js_settings(response.content)
```
noisecapella: Can you just get this from the response context instead?

singingwolfboy: Yep, you're right. Fixed.

noisecapella:
Looks great :+1:

PR #1380, opened 2016-10-13
Adds email audits to Django admin

### LINKED ISSUE #1377: Add email audits to django admin

mirfark:

Closes #1377 

#### What's this PR do?

Adds a Django admin model for the financial aid email audit model.

### DIFF
--- a/mail/admin.py
+++ b/mail/admin.py
@@ -1,0 +1,25 @@
+"""
+Admin views for Mail app
+"""
+
+from django.contrib import admin
+
+from mail.models import FinancialAidEmailAudit
+
+
+class FinancialAidEmailAuditAdmin(admin.ModelAdmin):
+    """Admin for FinancialAidEmailAudit"""
+    model = FinancialAidEmailAudit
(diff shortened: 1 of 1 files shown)


mail/admin.py
```diff
@@ -0,0 +1,25 @@
+"""
+Admin views for Mail app
+"""
+
+from django.contrib import admin
+
+from mail.models import FinancialAidEmailAudit
+
+
+class FinancialAidEmailAuditAdmin(admin.ModelAdmin):
```
noisecapella: I'm seeing an option to save the audit, which changes the modified time. If `has_change_permission` was implemented and returned `False`, would that disable the ability to save in the admin?

zags: Setting `has_change_permission` to false causes the model to not show up in the django admin.  As noted in #1258, there is an open django issue for this (https://code.djangoproject.com/ticket/8936), so, consistent with the financialaidaudit, we just set the fields to be readonly

noisecapella:
Looks good :+1:

PR #1382, opened 2016-10-13
Fixes failing test on master

alvinsiu:
N/A

#### What's this PR do?
The tests were failing on master because when #1303 was merged, `CountryIncomeThreshold` objects are automatically created but the tests don't account for that.  This PR adds a uniqueness constraint on `CountryIncomeThreshold` to have `country_code` be unique and it modifies tests to account for migrations already creating `CountryIncomeThreshold` data.

If tests pass :]

This failure would only occur sometimes because `ProfileFactory.create()` would only sometimes intersect with an existing `CountryIncomeThreshold`.

…sts to account for migration data loading


financialaid/api_test.py
```diff
@@ -131,15 +133,22 @@ def setUpTestData(cls):
             user=cls.enrolled_profile2.user,
             tier_program=cls.tier_programs["25k"],
             status=FinancialAidStatus.PENDING_MANUAL_APPROVAL
-        )
-        # Country income thresholds
-        cls.country_income_threshold_0 = CountryIncomeThresholdFactory.create(
-            income_threshold=0
-        )
-        cls.country_income_threshold_50000 = CountryIncomeThresholdFactory.create(
+        )\
```
alvinsiu: Oh, that's probably an accident - will remove

financialaid/api_test.py
```diff
@@ -131,15 +133,22 @@ def setUpTestData(cls):
...
+        # Country income thresholds (these are already created via migrations, so just need to modify them)
+        cls.country_income_threshold_50000, _ = CountryIncomeThreshold.objects.get_or_create(
```
noisecapella: If the object already exists, why not just use `get`?

financialaid/api_test.py
alvinsiu: Consequently, that was also what caused another linting error xD

financialaid/api_test.py
alvinsiu: It may not exist, since we're using `country=FuzzyText(length=2)` in the profile creation above.  I was thinking about using a hard-coded country code, but I felt this solution was more robust.

noisecapella: If it may not exist, can you change the comment above accordingly?

alvinsiu: Sure thing!

noisecapella:
I'm a little confused with the `CountryIncomeThreshold` logic but the change seems reasonable and local to the test case :+1:

PR #1387, opened 2016-10-13
Sorted programs in id order

### LINKED ISSUE #1290: Home: predictable order for program tiles
I'm not sure what I did, but for the first time I've seen the order of program tiles change on the home page, and I realize that this is something that we'll want to have control over.

noisecapella:
Fixes #1290 

#### What's this PR do?
Sorts programs in id order

Create a couple live programs and view the thumbnails on the home page. They should be in the order they were created. Change the first program's id number to come after the second program, then refresh. The programs should swap in the home page.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -36,7 +36,7 @@ class HomePage(Page):
     subpage_types = ['ProgramPage']
 
     def get_context(self, request):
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
💯 👍

PR #1388, opened 2016-10-13
Make program selector use full dialog width

### LINKED ISSUE #1281: Dashboard: when enrolling in DEDP, the program title doesn't fit.
screen shot 2016-10-06 at 9 17 04 am

screen shot 2016-10-06 at 9 17 08 am

noisecapella:
Fixes #1281

#### What's this PR do?
Makes the select take up the full width of the dialog
screenshot from 2016-10-13 17-18-59

### DIFF
--- a/static/js/components/NewEnrollmentDialog.js
+++ b/static/js/components/NewEnrollmentDialog.js
@@ -94,6 +94,14 @@ export default class NewEnrollmentDialog extends React.Component {
         onChange={this.handleSelectedProgramChange}
         floatingLabelText="Select Program"
         errorText={enrollDialogError}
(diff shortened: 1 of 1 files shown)


noisecapella:
@roberthouse54 @pdpinch Is this a solution you had in mind, or should it have some other width?

pdpinch:
this works for me

roberthouse54:
It doesn't need to be full width, but it should be about 2x what it was previously. And if it extends past that width, it should truncate, or overflow hidden (not scroll).

noisecapella:
Should be ready for review

gsidebo:
i'll take over from here

PR #1389, opened 2016-10-14
Added past course run UI to dashboard

### LINKED ISSUE #939: Dashboard: Display multiple runs of the same course
This is a lower priority issue because it may be some time before we have multiple runs of courses available in the MicroMasters programs. 

The relevant cases are (see [Google Doc](https://docs.google.com/document/d/1C8JwVxLW0YZBAFwUR06754uz-uyCCEbnvt2TIMJtKk0/) for screenshots): 

### Missed Deadline to Pay

image

If the user has missed the payment deadline, they will not be able to get MicroMasters credit for the course. We show them the “re-enroll” UI, the same as if they have failed.

### Failed Course

image

If the user has failed, the course status in the right column changes to “Failed”. We then display the “Re-enroll” UI, same as if the user was auditing a course but missed the Payment Due date.

### Failed Course - Enrollment Not Open

image

If the user has failed, but the enrollment is not open, we display this status.

### Failed Course - No Future Course Runs are defined

image

This course status is a combination of “Failed” and “No Future Runs defined”

### Multiple Past Course runs

image

This mock shows the UI when a user has multiple completed past course runs.

### Re-Enrolled

image

After the student has re-enrolled, either after failing a course, or for any other reason, the current run is displayed at the top, like normal, and any “Completed Runs” are displayed below, in gray boxes.

### LINKED ISSUE #1121: (Possible bug) Unverified, non-upgradeable course runs assigned failed status
from dashboard/api.py:265

```python
    if course_enrollment.is_verified:
        if course_run.is_current:
            status = CourseRunStatus.CURRENTLY_ENROLLED
        elif course_run.is_past:
            status = CourseRunStatus.READ_CERT
        elif course_run.is_future:
            status = CourseRunStatus.WILL_ATTEND
    else:
        if (course_run.is_current or course_run.is_future) and course_run.is_upgradable:
            status = CourseRunStatus.CAN_UPGRADE
        else:
            status = CourseRunStatus.NOT_PASSED
```

If an unverified course run is current/future, but not upgradable, we are assigning a status that indicates that it is a failed course run

### LINKED ISSUE #1122: (Possible bug) Logic for fetching enroll-able CourseRuns may be incorrect
from courses/models.py:146

```python
    def get_active_enrollment_queryset():
        """
        Returns a Q object that encapsulates the logic for filtering for CourseRun objects
        that are active for enrollment.
        """
        now = datetime.now(pytz.utc)
        return (
            # there is only the start date
            (models.Q(enrollment_end=None) &
             models.Q(end_date=None) & models.Q(start_date__lte=now)) |
            # there is no enrollment date but the course has not ended yet
            (models.Q(enrollment_end=None) & models.Q(end_date__gte=now)) |
            # the enrollment end date is simply greater than now
            models.Q(enrollment_end__gte=now)
        )
```

If a CourseRun has a start_date in the future (lets say 5 days from now) and no end_date or enrollment_end, that should probably be an enroll-able course, but this logic will exclude it

gsidebo:

Closes #939 
Fixes #1121 
Fixes #1122 

#### What's this PR do?

Handles courses with multiple runs per state. This is relevant in cases where the user has failed or missed an upgrade deadline for a course and we want to show them a future run, and also when the user has taken the course before multiple times.

Probably CourseRow.js as well as dashboard/api.py where a few important changes were made

Here's a [gist](https://gist.github.com/gsidebo/478920b3ead11b4de742c076c2e8d64c) with some functions to help set up test data. Let me know if anything is confusing or doesn't work. Check #939 for the designs that these states should match

ss 2016-10-13 at 21 23 02

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -69,17 +69,27 @@ class Course(models.Model):
     class Meta:
         unique_together = ('program', 'position_in_program',)
 
-    def get_next_run(self):
+    def get_first_unexpired_run(self, course_run_to_exclude=None):
         """
-        Gets the next run associated with this course.
+        Gets the soonest unexpired run associated with this course.
         Note that it could be current as long as the enrollment window
         has not closed.
+
+        Args:
+            course_run_to_exclude (CourseRun): A CourseRun to exclude
+                from the query
+
+        Returns: CourseRun or None: An unexpired course run
+
         """
-        next_run_set = self.courserun_set.filter(CourseRun.get_active_enrollment_queryset())
-        if not next_run_set.count():
-            return
-        next_run_set = next_run_set.order_by('start_date')
-        return next_run_set[0]
+        future_run_queryset = (
+            self.courserun_set
+            .filter(CourseRun.unexpired_courserun_queryset())
+            .order_by('start_date')
+        )
+        if course_run_to_exclude:
+            future_run_queryset = future_run_queryset.exclude(pk=course_run_to_exclude.pk)
+        return future_run_queryset.first()
 
     def get_promised_run(self):
         """
@@ -95,7 +105,7 @@ class Course(models.Model):
         """
         Construct the course page url
         """
-        course_run = self.get_next_run()
+        course_run = self.get_first_unexpired_run()
         if not course_run:
             return ""
         if not course_run.edx_course_key:
@@ -111,8 +121,8 @@ class Course(models.Model):
         Return text that contains start and enrollment
         information about the course.
         """
-        course_run = self.get_next_run()
-        if not course_run:
+        course_run = self.get_first_unexpired_run()
+        if not course_run or not course_run.start_date:
             promised_run = self.get_promised_run()
             if promised_run:
                 return "Coming " + promised_run.fuzzy_start_date
@@ -210,7 +220,6 @@ class CourseRun(models.Model):
                     return self.enrollment_start <= now <= self.enrollment_end
                 else:
                     return self.enrollment_start <= now
-
         return False
 
     @property
@@ -223,18 +232,16 @@ class CourseRun(models.Model):
                 (self.upgrade_deadline > datetime.now(pytz.utc)))
 
     @staticmethod
-    def get_active_enrollment_queryset():
+    def unexpired_courserun_queryset():
         """
-        Returns a Q object that encapsulates the logic for filtering for CourseRun objects
-        that are active for enrollment.
+        Returns a Q object for CourseRuns that have not expired (ie: the enrollment
+        period has not passed, or the enrollment period is unspecified and the end
+        date occurs in the future)
         """
         now = datetime.now(pytz.utc)
         return (
-            # there is only the start date
-            (models.Q(enrollment_end=None) &
-             models.Q(end_date=None) & models.Q(start_date__lte=now)) |
-            # there is no enrollment date but the course has not ended yet
-            (models.Q(enrollment_end=None) & models.Q(end_date__gte=now)) |
-            # the enrollment end date is simply greater than now
+            (models.Q(enrollment_end=None) & (
+                models.Q(end_date__gte=now) | models.Q(end_date=None)
+            )) |
             models.Q(enrollment_end__gte=now)
         )
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -40,13 +40,14 @@ class CourseStatus:
     CURRENTLY_ENROLLED = 'currently-enrolled'
     WILL_ATTEND = 'will-attend'
     CAN_UPGRADE = 'can-upgrade'
+    MISSED_DEADLINE = 'missed-deadline'
     OFFERED = 'offered'
 
     @classmethod
     def all_statuses(cls):
         """Helper to get all the statuses"""
         return [cls.PASSED, cls.NOT_PASSED, cls.CURRENTLY_ENROLLED,
-                cls.CAN_UPGRADE, cls.OFFERED, cls.WILL_ATTEND]
+                cls.CAN_UPGRADE, cls.OFFERED, cls.WILL_ATTEND, cls.MISSED_DEADLINE]
 
 
 class CourseRunStatus:
@@ -58,6 +59,7 @@ class CourseRunStatus:
     CHECK_IF_PASSED = 'check-if-passed'
     WILL_ATTEND = 'will-attend'
     CAN_UPGRADE = 'can-upgrade'
+    MISSED_DEADLINE = 'missed-deadline'
     NOT_PASSED = 'not-passed'
 
 
@@ -207,27 +209,29 @@ def get_info_for_course(course, mmtrack):
     run_statuses.remove(run_status)
 
     if run_status.status == CourseRunStatus.NOT_ENROLLED:
-        next_run = course.get_next_run()
+        next_run = course.get_first_unexpired_run()
         if next_run is not None:
             _add_run(next_run, mmtrack, CourseStatus.OFFERED)
     elif run_status.status == CourseRunStatus.NOT_PASSED:
-        next_run = course.get_next_run()
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.NOT_PASSED)
+        next_run = course.get_first_unexpired_run()
         if next_run is not None:
             _add_run(next_run, mmtrack, CourseStatus.OFFERED)
-        if next_run is None or run_status.course_run.pk != next_run.pk:
-            _add_run(run_status.course_run, mmtrack, CourseStatus.NOT_PASSED)
+    elif run_status.status == CourseRunStatus.MISSED_DEADLINE:
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.MISSED_DEADLINE)
+        next_run = course.get_first_unexpired_run(course_run_to_exclude=run_status.course_run)
+        if next_run is not None:
+            _add_run(next_run, mmtrack, CourseStatus.OFFERED)
     elif run_status.status == CourseRunStatus.CURRENTLY_ENROLLED:
         _add_run(run_status.course_run, mmtrack, CourseStatus.CURRENTLY_ENROLLED)
     # check if we need to check the certificate
     elif run_status.status == CourseRunStatus.CHECK_IF_PASSED:
         # if the user never passed the course she needs to enroll in the next one
         if not mmtrack.has_passed_course(run_status.course_run.edx_course_key):
-            next_run = course.get_next_run()
+            _add_run(run_status.course_run, mmtrack, CourseRunStatus.NOT_PASSED)
+            next_run = course.get_first_unexpired_run()
             if next_run is not None:
                 _add_run(next_run, mmtrack, CourseStatus.OFFERED)
-            # add the run of the status anyway if the next run is different from the one just added
-            if next_run is None or run_status.course_run.pk != next_run.pk:
-                _add_run(run_status.course_run, mmtrack, CourseStatus.NOT_PASSED)
         else:
             _add_run(run_status.course_run, mmtrack, CourseStatus.PASSED)
     elif run_status.status == CourseRunStatus.WILL_ATTEND:
@@ -238,9 +242,8 @@ def get_info_for_course(course, mmtrack):
     # add all the other runs with status != NOT_ENROLLED
     # the first one (or two in some cases) has been added with the logic before
     for run_status in run_statuses:
-        if run_status.status != CourseRunStatus.NOT_ENROLLED:
-            if (run_status.status == CourseRunStatus.CHECK_IF_PASSED and
-                    mmtrack.has_passed_course(run_status.course_run.edx_course_key)):
+        if run_status.status == CourseRunStatus.CHECK_IF_PASSED:
+            if mmtrack.has_passed_course(run_status.course_run.edx_course_key):
                 # in this case the user might have passed the course also in the past
                 _add_run(run_status.course_run, mmtrack, CourseStatus.PASSED)
             else:
@@ -272,10 +275,13 @@ def get_status_for_courserun(course_run, mmtrack):
         elif course_run.is_future:
             status = CourseRunStatus.WILL_ATTEND
     else:
-        if (course_run.is_current or course_run.is_future) and course_run.is_upgradable:
-            status = CourseRunStatus.CAN_UPGRADE
-        else:
+        if course_run.is_past:
             status = CourseRunStatus.NOT_PASSED
+        else:
+            if course_run.is_upgradable:
+                status = CourseRunStatus.CAN_UPGRADE
+            else:
+                status = CourseRunStatus.MISSED_DEADLINE
     return CourseRunUserStatus(
         status=status,
         course_run=course_run
--- a/dashboard/api_test.py
+++ b/dashboard/api_test.py
@@ -348,7 +348,7 @@ class CourseRunTest(CourseTests):
         current_run.upgrade_deadline = self.now-timedelta(weeks=1)
         current_run.save()
         run_status = api.get_status_for_courserun(current_run, self.mmtrack)
-        assert run_status.status == api.CourseRunStatus.NOT_PASSED
+        assert run_status.status == api.CourseRunStatus.MISSED_DEADLINE
 
     def test_not_paid_not_passed(self):
         """test for get_status_for_courserun for course not paid but that is past"""
@@ -490,42 +490,44 @@ class InfoCourseTest(CourseTests):
                 api.get_info_for_course(self.course, None)
             )
         # the mock object has been called 2 times
-        # one for the course that is current run
-        mock_format.assert_any_call(self.course_run, api.CourseStatus.OFFERED, None, position=1)
         # one for the one that is past
-        mock_format.assert_any_call(self.course_run_ver, api.CourseStatus.NOT_PASSED, None, position=2)
+        mock_format.assert_any_call(self.course_run_ver, api.CourseStatus.NOT_PASSED, None, position=1)
+        # one for the course that is current run
+        mock_format.assert_any_call(self.course_run, api.CourseStatus.OFFERED, None, position=2)
 
     @patch('dashboard.api.format_courserun_for_dashboard', autospec=True)
     def test_info_not_enrolled_not_passed_not_offered(self, mock_format):
         """test for get_info_for_course for course with run not passed and nothing offered"""
+        self.mmtrack.configure_mock(**{'has_passed_course.return_value': False})
         with patch(
             'dashboard.api.get_status_for_courserun',
             autospec=True,
             side_effect=self.get_mock_run_status_func(
-                api.CourseRunStatus.NOT_PASSED, self.course_run, api.CourseRunStatus.NOT_PASSED),
-        ), patch('courses.models.Course.get_next_run', autospec=True, return_value=None):
+                api.CourseRunStatus.CHECK_IF_PASSED, self.course_run, api.CourseRunStatus.CHECK_IF_PASSED),
+        ), patch('courses.models.Course.get_first_unexpired_run', autospec=True, return_value=None):
             self.assert_course_equal(
                 self.course,
-                api.get_info_for_course(self.course, None)
+                api.get_info_for_course(self.course, self.mmtrack)
             )
-        mock_format.assert_any_call(self.course_run, api.CourseStatus.NOT_PASSED, None, position=1)
-        mock_format.assert_any_call(self.course_run_ver, api.CourseStatus.NOT_PASSED, None, position=2)
+        mock_format.assert_any_call(self.course_run, api.CourseStatus.NOT_PASSED, self.mmtrack, position=1)
+        mock_format.assert_any_call(self.course_run_ver, api.CourseStatus.NOT_PASSED, self.mmtrack, position=2)
 
     @patch('dashboard.api.format_courserun_for_dashboard', autospec=True)
     def test_info_grade(self, mock_format):
         """test for get_info_for_course for course with a course current and another not passed"""
+        self.mmtrack.configure_mock(**{'has_passed_course.return_value': False})
         with patch(
             'dashboard.api.get_status_for_courserun',
             autospec=True,
             side_effect=self.get_mock_run_status_func(
-                api.CourseRunStatus.CURRENTLY_ENROLLED, self.course_run, api.CourseRunStatus.NOT_PASSED),
+                api.CourseRunStatus.CURRENTLY_ENROLLED, self.course_run, api.CourseRunStatus.CHECK_IF_PASSED),
         ):
             self.assert_course_equal(
                 self.course,
-                api.get_info_for_course(self.course, None)
+                api.get_info_for_course(self.course, self.mmtrack)
             )
-        mock_format.assert_any_call(self.course_run, api.CourseStatus.CURRENTLY_ENROLLED, None, position=1)
-        mock_format.assert_any_call(self.course_run_ver, api.CourseStatus.NOT_PASSED, None, position=2)
+        mock_format.assert_any_call(self.course_run, api.CourseStatus.CURRENTLY_ENROLLED, self.mmtrack, position=1)
+        mock_format.assert_any_call(self.course_run_ver, api.CourseStatus.NOT_PASSED, self.mmtrack, position=2)
 
     @patch('dashboard.api.format_courserun_for_dashboard', autospec=True)
     def test_info_check_but_not_passed(self, mock_format):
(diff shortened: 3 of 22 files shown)


gsidebo:
FYI: I'm still working on unit tests for this. It's pretty sizable so I needed to get this up to get some eyes on it first.

gsidebo:
unit tests fixed now after some painful merges

gsidebo:
all unit tests fixed now. should pass

noisecapella:
Can you look at the lint errors?

noisecapella:
It looks like the `View on edX` link appears even if there are no course run available, can you take a look at that?

static/js/components/dashboard/util.js
```diff
@@ -16,3 +17,9 @@ export const courseListToolTip = (text: string, id: string) => (
     </ReactTooltip>
   </div>
 );
+
+export const isCurrentlyEnrollable = (enrollmentStartDate: ?Object, now: ?moment$Moment): boolean => (
```
noisecapella: Can you change `enrollmentStartDate` to also use `moment$Moment` as a type?

noisecapella:
In the future please separate the backend and frontend work into two separate PRs, and break it up further if possible. It's difficult to do a functional review here because so many things are changing

gsidebo:
@noisecapella the way i had this coded originally, splitting front and backend PRs would have broken the app. in hindsight i think i see an intelligent way to break them up without breaking. sorry i didnt see it sooner

gsidebo:
@roberthouse54 is there a clear reasoning behind courses that show the 'View on edX' link vs. courses that don't? going off of this issue (#939) and these [mockups](https://projects.invisionapp.com/share/NB7VM1GSG#/screens/185611170/comments). showing/hiding the link seems pretty arbitrary. eg: the design indicates that we show it for a failed course run, and hide it for a course run that a user has missed the upgrade deadline.

noisecapella:
If a user gets a `missed-deadline` status, wouldn't we want to show the pay button to register for the next course run if there is one? The upgrade deadline applies only to a particular course run I think

noisecapella:
Nevermind, looks like I had the enrollment start date in the future for that run

noisecapella:
In the mockups the 'Failed' for the previous courses is italic, can you make that change in this PR?

noisecapella:
Actually it looks like the whole row is italic

noisecapella:
I'm seeing some rounding weirdness:
screenshot from 2016-10-17 15-21-55

We should make a function to format grade percents consistently, here or in another PR if you prefer

gsidebo:
@roberthouse54 see george's comment about italics above. the designs seem to have arbitrary italic font styling for some types of past course runs. can you clarify what you want here?

noisecapella:
After a lot of testing functionality looks great! Looking at code now

gsidebo:
@noisecapella re: missed deadline, that does bring up an interesting point. the MISSED_DEADLINE state is only assigned if your course run is current, and the code assumes that the next run to show must have a start date in the future. is it possible for these course runs to overlap? if so, it would probably make more sense to offer up a course run that is enrollable now or in the future, and (obviously) is not the same as the missed deadline run

gsidebo:
that grade issue is strange. this is the code to render the final grade in the sub row component:

```javascript
        <Cell col={2} key="2">
          <div className="course-grade">
            { (courseRun.status === STATUS_NOT_PASSED) ? `${courseRun.final_grade}%` : ' ' }
          </div>
        </Cell>
```

it's just spitting out the value with a % at the end - no calculation involved (unless that's happening upstream). i'll look into a formatting function anyway. that's a good idea

dashboard/api.py
```diff
@@ -207,27 +209,29 @@ def _add_run(run, mmtrack_, status):
...
-        next_run = course.get_next_run()
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.NOT_PASSED)
+        next_run = course.get_first_unexpired_run()
         if next_run is not None:
             _add_run(next_run, mmtrack, CourseStatus.OFFERED)
-        if next_run is None or run_status.course_run.pk != next_run.pk:
-            _add_run(run_status.course_run, mmtrack, CourseStatus.NOT_PASSED)
+    elif run_status.status == CourseRunStatus.MISSED_DEADLINE:
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.MISSED_DEADLINE)
+        future_run = course.get_first_future_run()
```
noisecapella: Why are we using `get_first_future_run` instead of `get_first_unexpired_run` here?

gsidebo: yeah, this touches on my [previous comment](https://github.com/mitodl/micromasters/pull/1389#issuecomment-254308767). the `MISSED_DEADLINE` status implies that the course run is current. `get_first_unexpired_run` would return the current course run, which we obviously don't want in this case. still, only getting future runs (ie: runs with start_date > now) is probably not correct here. one solution would be to add an optional parameter to `get_first_unexpired_run` that would exclude a specific course run/course run id. what do you think?

roberthouse54:
For all past course runs, use italic for all text.
All courses should have the link to View on edX, as long as a link is available. The omissions on the mock were just a mistake.

dashboard/api.py
gsidebo: ok i just set this to call `get_first_unexpired_run` and tweaked that function to accept a course run to exclude

noisecapella: Sorry for the delay, that sounds good with me

courses/models.py
```diff
@@ -69,17 +69,20 @@ def __str__(self):
     class Meta:
         unique_together = ('program', 'position_in_program',)
 
-    def get_next_run(self):
+    def get_first_unexpired_run(self, course_run_to_exclude=None):
         """
-        Gets the next run associated with this course.
+        Gets the soonest unexpired run associated with this course.
         Note that it could be current as long as the enrollment window
         has not closed.
```
noisecapella: Can you add the argument and return value to the docstring here?

dashboard/api.py
```diff
@@ -207,27 +209,29 @@ def _add_run(run, mmtrack_, status):
...
             _add_run(next_run, mmtrack, CourseStatus.OFFERED)
     elif run_status.status == CourseRunStatus.NOT_PASSED:
-        next_run = course.get_next_run()
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.NOT_PASSED)
+        next_run = course.get_first_unexpired_run()
+        if next_run is not None:
+            _add_run(next_run, mmtrack, CourseStatus.OFFERED)
+    elif run_status.status == CourseRunStatus.MISSED_DEADLINE:
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.MISSED_DEADLINE)
+        next_run = course.get_first_unexpired_run(course_run_to_exclude=run_status.course_run)
```
noisecapella: It seems like the exclude logic was originally applied to the `NOT_PASSED` state, should we have it there too?

dashboard/api.py
```diff
@@ -207,27 +209,29 @@ def _add_run(run, mmtrack_, status):
...
-            _add_run(run_status.course_run, mmtrack, CourseStatus.NOT_PASSED)
     elif run_status.status == CourseRunStatus.CURRENTLY_ENROLLED:
         _add_run(run_status.course_run, mmtrack, CourseStatus.CURRENTLY_ENROLLED)
     # check if we need to check the certificate
     elif run_status.status == CourseRunStatus.CHECK_IF_PASSED:
         # if the user never passed the course she needs to enroll in the next one
         if not mmtrack.has_passed_course(run_status.course_run.edx_course_key):
-            next_run = course.get_next_run()
+            _add_run(run_status.course_run, mmtrack, CourseRunStatus.NOT_PASSED)
+            next_run = course.get_first_unexpired_run()
```
noisecapella: Same question here about excluding `run_status.course_run`

static/js/components/dashboard/util.js
```diff
@@ -16,3 +17,26 @@ export const courseListToolTip = (text: string, id: string) => (
     </ReactTooltip>
   </div>
 );
+
+export const isCurrentlyEnrollable = (enrollmentStartDate, now) => (
```
noisecapella: Can you add flow types here?

static/js/components/dashboard/util.js
```diff
@@ -16,3 +17,26 @@ export const courseListToolTip = (text: string, id: string) => (
...
+  enrollmentStartDate !== null &&
+    enrollmentStartDate !== undefined &&
+    enrollmentStartDate.isSameOrBefore(now || moment(), 'day')
+);
+
+export const formatGrade = (grade) => {
```
noisecapella: Can you add flow types here?

static/js/components/dashboard/util.js
```diff
@@ -16,3 +17,26 @@ export const courseListToolTip = (text: string, id: string) => (
...
+  if (_.isNil(grade) || grade === '') {
+    return '';
+  } else {
+    let gradeString;
+    if (_.isString(grade)) {
+      gradeString = _.round(Number(grade));
+    } else if (_.isInteger(grade)) {
+      gradeString = grade;
```
noisecapella: I think this is unnecessary because there aren't integers in javascript, at least not in the standard types. Also, the `gradeString = _.round(Number(grade))` case should cover numbers and strings, and you've already tested against nil and an empty string, so I don't think you need the if block here at all

static/js/util/test_utils.js
```diff
@@ -21,14 +22,31 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
+export function setCourseRunCount(course: Course, desiredRunCount: number, courseRunToCopy: ?CourseRun) {
+  let currentRunCount = course.runs.length;
+  if (currentRunCount < desiredRunCount) {
+    let courseRun = currentRunCount === 0 ? courseRunToCopy : course.runs[0];
+    if (!courseRun) {
+      throw new Error('Need a course run to copy.');
+    }
+    let runsNeeded = desiredRunCount - currentRunCount;
+    for (let i = 0; i < runsNeeded; i++) {
+      course.runs.push(_.cloneDeep(courseRun));
```
noisecapella: When we're creating course runs can you modify the course run so that it's unique? We should make sure the `id` and `course_id` are unique and `position` is unique for that particular course

static/js/components/dashboard/util.js
```diff
@@ -16,3 +17,26 @@ export const courseListToolTip = (text: string, id: string) => (
...
+    } else if (_.isNumber(grade)) {
+      // If 'grade' is a number but not an integer, round it.
+      gradeString = _.round(grade);
+    }
+    return gradeString ? `${gradeString}%` : '';
```
noisecapella: This would fail if `gradeString` is `0`. It seems like `${gradeString}%` should cover all cases here since the code checks for nil above

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -105,15 +106,14 @@ describe('CourseDescription', () => {
...
+    let firstRun = alterFirstRun(course, {course_end_date: '1999-13-92'});
+    const wrapper = shallow(<CourseDescription courseRun={firstRun} courseTitle={course.title} />);
     let elements = getElements(wrapper);
     assert.equal(elements.detailsText, '');
   });
 
   it('does not show anything when there are no runs for a course', () => {
-    let course = findCourse(course => course.runs.length === 0);
-    const wrapper = shallow(<CourseDescription course={course} />);
+    const wrapper = shallow(<CourseDescription courseRun={null} courseTitle={null} />);
```
noisecapella: It looks like `CourseRow` passes an empty object to `CourseDescription` when there are no rows, not `null`, at least that's what I'm getting from `getFirstRun`. Can we make it consistent one way or the other?

dashboard/api.py
gsidebo: you can't have a `NOT_PASSED` course unless it's in the past, or 'expired' in other words. adding that kwarg would be a bit more explicit, but i'm not sure we need it. what do you think?

static/js/components/dashboard/util.js
gsidebo: flow was choking on some issue yesterday, and alice nor i could figure it out, but i read something last night that gave me a hint. i'll look into this again

static/js/components/dashboard/util.js
gsidebo: same response as above

static/js/components/dashboard/util.js
gsidebo: fair enough. i was aiming to avoid unnecessary operations, but the performance is quite unlikely to take a hit

static/js/util/test_utils.js
gsidebo: will do

static/js/components/dashboard/util.js
gsidebo: good catch

static/js/components/dashboard/CourseDescription.js
```diff
@@ -21,12 +22,13 @@ const edxLinkBase = `${SETTINGS.edx_base_url}/courses/`;
...
-    course: Course
+    courseRun: CourseRun,
+    courseTitle: ?string
   };
 
   needsEdxLink(run: CourseRun): boolean {
-    // Any status besides 'offered' implies that the user has at some point enrolled 
-    return run.status && run.status !== STATUS_OFFERED;
+    // Any status besides 'offered' implies that the user has at some point enrolled
+    return run && run.status !== STATUS_OFFERED;
```
gsidebo: actually i just need to get rid of this function entirely. it's not needed anymore per rob's comment

static/js/components/dashboard/CourseRow.js
```diff
@@ -1,56 +1,163 @@
...
+    let firstRun = this.getFirstRun();
+
+    let subRows = [];
+    let subRowRuns = [];
+
+    if (this.needsToEnrollAgain(firstRun)) {
+      subRowRuns.push(this.futureEnrollableRun(course));
+    }
+    let completedRuns = this.pastCourseRuns(course);
+    if (completedRuns.length > 0) {
```
noisecapella: This check is unnecessary since the concat won't do anything with an empty list

static/js/components/dashboard/CourseRow.js
```diff
@@ -1,56 +1,163 @@
...
+      subRowRuns = subRowRuns.concat(completedRuns);
+    }
+
+    if (subRowRuns.length > 0) {
```
noisecapella: This check is also unnecessary since the for loop below won't do anything on an empty list

static/js/components/dashboard/CourseRow.js
```diff
@@ -1,56 +1,163 @@
...
+
+    if (this.shouldDisplayGradeColumn(run)) {
+      columns.push(
+        <Cell col={2} key="2">
+          <CourseGrade courseRun={run} />
+        </Cell>
+      );
+      lastColumnSize = 4;
+    } else {
+      lastColumnSize = 6;
```
noisecapella: Can you add a test to cover this case?

noisecapella:
Finally got through it. There's a lot of good stuff in this PR, just some minor comments here and there!

dashboard/api.py
noisecapella: Makes sense, feel free to leave it as is. Is that also true for `CHECK_IF_PASSED`?

static/js/components/dashboard/util.js
noisecapella: I see, feel free to leave it alone if it's being difficult

dashboard/api_test.py
```diff
@@ -490,42 +490,44 @@ def test_info_not_passed_offered(self, mock_format):
...
     @patch('dashboard.api.format_courserun_for_dashboard', autospec=True)
     def test_info_grade(self, mock_format):
         """test for get_info_for_course for course with a course current and another not passed"""
+        self.mmtrack.configure_mock(**{'has_passed_course.return_value': False})
         with patch(
             'dashboard.api.get_status_for_courserun',
             autospec=True,
             side_effect=self.get_mock_run_status_func(
-                api.CourseRunStatus.CURRENTLY_ENROLLED, self.course_run, api.CourseRunStatus.NOT_PASSED),
+                api.CourseRunStatus.CURRENTLY_ENROLLED, self.course_run, api.CourseRunStatus.CHECK_IF_PASSED),
```
gsidebo: i changed this b/c the test was failing - can't remember exactly why this fixed it. i'll look into it

gsidebo:
@noisecapella @aliceriot i cant get flow to shut up about `formatGrade`. here it is in its current form:

```javascript
export const formatGrade = (grade: number|string|null): string => {
  if (_.isNil(grade) || grade === '') {
    return '';
  } else {
    if (_.isString(grade)) {
      grade = Number(grade);
    }
    // isFinite will return true for numbers, false for strings and NaN
    return _.isFinite(grade) ? `${_.round(grade)}%` : '';
  }
};
```

and here is flow's complaint:

```
static/js/components/dashboard/util.js:36
 36:     return _.isFinite(grade) ? `${_.round(grade)}%` : '';
                                       ^^^^^^^^^^^^^^ call of method `round`
 36:     return _.isFinite(grade) ? `${_.round(grade)}%` : '';
                                               ^^^^^ null. This type is incompatible with
325:     round(number: number, precision?: number): number;
                       ^^^^^^ number. See lib: node_modules/iflow-lodash/index.js.flow:325

static/js/components/dashboard/util.js:36
 36:     return _.isFinite(grade) ? `${_.round(grade)}%` : '';
                                       ^^^^^^^^^^^^^^ call of method `round`
 36:     return _.isFinite(grade) ? `${_.round(grade)}%` : '';
                                               ^^^^^ string. This type is incompatible with
325:     round(number: number, precision?: number): number;
                       ^^^^^^ number. See lib: node_modules/iflow-lodash/index.js.flow:325
```

i don't know if there is a way to (a) specify a type that flow will accept here, or (b) tell flow to ignore this method because it doesn't know what it's talking about. if we can't do either of those, i guess i should leave flow typing out of this file entirely? this is getting stressful

gsidebo:
(ignore previous comment about flow typing issues 😄 )

dashboard/api_test.py
gsidebo: i remember now. for past course runs, we want to ignore any course runs that the user took as an audit (ie: they missed the deadline to upgrade). this if/else logic was being used to handle past course runs: `if (run_status.status == CourseRunStatus.CHECK_IF_PASSED and mmtrack.has_passed_course(run_status.course_run.edx_course_key)):`. that was made irrelevant with the choice to ignore `MISSED_DEADLINE` course runs. now essentially all we care about is non-audit course runs that have a grade, so that logic became an if/else with this condition: `run_status.status == CourseRunStatus.CHECK_IF_PASSED`. hence the change to this unit test

static/js/components/dashboard/CourseAction.js
```diff
@@ -177,8 +175,10 @@ export default class CourseAction extends React.Component {
       }
       break;
     }
-    case STATUS_NOT_PASSED:
-      // do nothing;
+    case STATUS_MISSED_DEADLINE:
+      description = this.renderTextDescription(
+        'You missed the payment deadline and will not receive Micromasters credit for this course.'
```
noisecapella: It seems like you missed this one

courses/models.py
```diff
@@ -69,17 +69,27 @@ def __str__(self):
...
+
+        Args:
+            course_run_to_exclude (CourseRun): A CourseRun to exclude
+                from the query
+
+        Returns (CourseRun or None): An unexpired course run
```
noisecapella: The syntax here should be `Returns: CourseRun or None: An unexpired course run`, broken into separate lines if you prefer: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#docstrings

noisecapella:
Can you update the MicroMasters text I mentioned in [this comment](https://github.com/mitodl/micromasters/pull/1389#discussion_r83679585)?

static/js/components/dashboard/CourseRow.js
```diff
@@ -1,56 +1,163 @@
...
+      subRowRuns.push(this.futureEnrollableRun(course));
+    }
+    let completedRuns = this.pastCourseRuns(course);
+    if (completedRuns.length > 0) {
+      subRowRuns = subRowRuns.concat(completedRuns);
+    }
+
+    if (subRowRuns.length > 0) {
+      let i = 1;
+      for (let subRowRun of subRowRuns) {
```
noisecapella: It looks like you missed this one

gsidebo:
fixed

static/js/util/test_utils.js
```diff
@@ -21,14 +22,38 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
 
+export function setCourseRunCount(course: Course, desiredRunCount: number, courseRunToCopy: ?CourseRun) {
+  let currentRunCount = course.runs.length;
+  if (currentRunCount < desiredRunCount) {
+    let courseRun = currentRunCount === 0 ? courseRunToCopy : course.runs[0];
+    if (!courseRun) {
+      throw new Error('Need a course run to copy.');
+    }
+    let runsNeeded = desiredRunCount - currentRunCount;
+    let idMax = _.max(_.map(course.runs, run => run.id)) || 0;
```
noisecapella: `id` needs to be unique across all programs, not just within this course

static/js/util/test_utils.js
```diff
@@ -21,14 +22,38 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
+    let positionMax = _.max(_.map(course.runs, run => run.position)) || 0;
+    for (let i = 0; i < runsNeeded; i++) {
+      let newCourseRun = _.cloneDeep(courseRun);
+      positionMax++;
+      idMax++;
+      newCourseRun.position = positionMax;
+      newCourseRun.id = idMax;
+      course.runs.push(newCourseRun);
```
noisecapella: This still needs to create a `course_id`. See the test in `constants_test.js` for the things we need to check for

noisecapella:
Can you look at this comment? https://github.com/mitodl/micromasters/pull/1389#discussion_r83884187

dashboard/api.py
```diff
@@ -207,27 +209,29 @@ def _add_run(run, mmtrack_, status):
...
         if next_run is not None:
             _add_run(next_run, mmtrack, CourseStatus.OFFERED)
     elif run_status.status == CourseRunStatus.NOT_PASSED:
-        next_run = course.get_next_run()
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.NOT_PASSED)
+        next_run = course.get_first_unexpired_run()
+        if next_run is not None:
+            _add_run(next_run, mmtrack, CourseStatus.OFFERED)
+    elif run_status.status == CourseRunStatus.MISSED_DEADLINE:
+        _add_run(run_status.course_run, mmtrack, CourseRunStatus.MISSED_DEADLINE)
```
noisecapella: It looks like this case is uncovered, can you add a test for it?

screenshot from 2016-10-19 14-41-52

static/js/util/test_utils.js
gsidebo: this is creating an entirely new course object. it's not changing/altering the existing values in that giant`DASHBOARD_RESPONSE` variable, nor does it need to be concerned about the other values in there. i wrote this for the convenience of writing these tests, but it's also an intermediate step toward handling test data like we should, ie: generating the test data we need on the fly instead of depending on a giant static object that satisfies every possible condition that we want to test for. the need to check for unique id's/etc in that variable is just one symptom of the anti-pattern

noisecapella: `course_id` and `id` need to be unique within the response data, not just within the course or program. If we were not going to check `DASHBOARD_RESPONSE` or other response objects then this function would need to create a new response object, not just a new course

gsidebo: the point of this function is to generate data for a single Course. it's not concerned with the structure of the entire response data. i don't quite understand why this function, or the tests that use it, need to worry about that. let's chat if you want to explain further

static/js/util/test_utils.js
```diff
@@ -21,14 +22,47 @@ export function findCourse(courseSelector: (course: Course, program: Program) =>
...
+        position: positionMax,
+        id: idMax,
+        course_id: `${newCourseRun.course_id}-new-${i}`
+      });
+      course.runs.push(newCourseRun);
+    }
+  } else if (currentRunCount > desiredRuns) {
+    course.runs = _.take(course.runs, desiredRuns);
+  }
+  course = Object.assign(course, {
```
noisecapella: `Object.assign` assigns to the first parameter. If you wanted a shallow clone it should be something like `Object.assign({}, ...`. Can you either remove the return value or add the `{}` parameter?

gsidebo: i don't want a shallow clone. i just want to update the object with specific key-values, like dict.update() in python. if there is a more preferred way to do this, lmk

noisecapella: That's fine, can you remove the return value then since it's unnecessary?

static/js/components/dashboard/FinancialAidCard.js
```diff
@@ -119,7 +119,7 @@ export default class FinancialAidCard extends React.Component {
         { courseListToolTip('filler-text', 'course-price') }
       </div>
       <div className="explanation">
-        The cost of courses in the {title} Micromasters varies
+        The cost of courses in the {title} MicroMasters varies
```
noisecapella: Thanks for being proactive here!

PR #1400, opened 2016-10-14
Changed logic to assign the student ID

### LINKED ISSUE #1399: Race condition during student id assignment
The code for student id assignment is prone to race condition issues leading to exceptions during user creation when load is high.

giocalitri:
fixes #1399
#### What's this PR do?
Changes the code for assigning a student id to the user profile. Now the profile pk is used.

### DIFF
--- a/profiles/factories.py
+++ b/profiles/factories.py
@@ -14,7 +14,6 @@ from factory.fuzzy import (
     FuzzyAttribute,
     FuzzyChoice,
     FuzzyDate,
-    FuzzyInteger,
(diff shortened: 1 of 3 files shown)


noisecapella:
What about existing student ids? Should we have a data migration for this?

giocalitri:
existing highest `student_id` should be <= of highest `profile.pk`, so I do not think a migration is necessary

noisecapella:
Functionality works :+1:

PR #1403, opened 2016-10-14
Added tracking_id to silence warnings

### LINKED ISSUE #1383: trakingid in base.html template causes crashes
found during load tests

noisecapella:
Fixes #1383 

#### What's this PR do?
Adds `tracking_id` so the template warnings are silenced


pdpinch:
Is tracking id related to #1154 ? Will that still work with this change?

noisecapella:
Yes, this just addresses template warnings for pages which don't already set `tracking_id`.

PR #1404, opened 2016-10-14
Added webpack_public_path

### LINKED ISSUE #1398: Add webpack_public_path to new bundles
This would help make dev work a little easier by re-enabling hot reloading

noisecapella:
Fixes #1398

#### What's this PR do?
Sets `__webpack_public_path__` on bundles where it is missing. This helps enable hot reloading, a dev feature

Change the color in some CSS and verify that there is not a full page refresh


aliceriot:
oh webpack you're always so intuitive and easy to use

PR #1406, opened 2016-10-14
Fixed console warning for faculty carousel

### LINKED ISSUE #1322: Console error when faculty carousel contains zero items
I think we just need to check if the div exists before connecting it with React

noisecapella:
Fixes #1322

#### What's this PR do?
Adds a check to make sure the faculty carousel div exists before populating it

### DIFF
--- a/static/js/faculty_carousel.js
+++ b/static/js/faculty_carousel.js
@@ -5,7 +5,9 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 const carouselDiv = document.querySelector('#faculty-carousel');
-ReactDOM.render(
-  <FacultyCarousel faculty={SETTINGS.faculty} />,
-  carouselDiv
-);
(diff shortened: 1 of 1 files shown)


static/js/faculty_carousel.js
```diff
@@ -5,7 +5,9 @@ import React from 'react';
...
-ReactDOM.render(
-  <FacultyCarousel faculty={SETTINGS.faculty} />,
-  carouselDiv
-);
+if (carouselDiv !== null) {
+  ReactDOM.render(
+    <FacultyCarousel faculty={SETTINGS.faculty}/>,
+    carouselDiv
+  );
+}
```
aliceriot: OK so looking at `cms/templates/cms/program_page.html` we'll have `#faculty-carousel` if there are faculty to show, and otherwise we won't.

I was worried that the error was because of a race condition, but I see that's not a concern.

Issue #1362, opened 2016-10-13
Search: hide facets when they have no hits


pdpinch:
This looks bad (no current grades in search results):

![image](https://cloud.githubusercontent.com/assets/430126/19387627/687aa492-91e9-11e6-900d-d7458b6f01db.png)

roberthouse54:
![image](https://cloud.githubusercontent.com/assets/20047260/19354179/da47903c-9133-11e6-8e89-c9c4726a020b.png)

roberthouse54:
So this state will happen as follows:

1. The "Filter box" in the left column will only be empty if there are no users in a program.
2. The "Results box" in the right column will be empty either if there are no users in a program (as shown above), OR if the user has selected some filters and there are no results.

For point 2, we could show a different message depending on the reason the box is empty, or we could just show the same message for both cases, as I am assuming that is easier to do.

pdpinch:
@roberthouse54 I'm going to open separate issues for your design changes.

pdpinch:
#1390 and #1391

PR #1407, opened 2016-10-17
hid facets when they have no hits

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1362

#### What's this PR do?
- hid facets when they have no hits

- /static/js/components/LearnerSearch.js

- see ``/learners/`` as staff with data
- see ``/learners/`` page as staff with no data **OR**
- see ``/learners/`` page as student then you will data have no 

screen shot 2016-10-17 at 3 26 56 pm

screen shot 2016-10-17 at 3 27 04 pm

### DIFF
--- a/static/js/components/search/FilterVisibilityToggle.js
+++ b/static/js/components/search/FilterVisibilityToggle.js
@@ -1,8 +1,10 @@
 // @flow
 import React from 'react';
+import _ from 'lodash';
+import { SearchkitComponent } from 'searchkit';
 import Icon from 'react-mdl/lib/Icon';
 
-export default class FilterVisibilityToggle extends React.Component {
+export default class FilterVisibilityToggle extends SearchkitComponent {
   props: {
     filterName:             string,
     checkFilterVisibility:  (filterName: string) => boolean,
@@ -15,13 +17,29 @@ export default class FilterVisibilityToggle extends React.Component {
     return checkFilterVisibility(filterName) ? "" : "closed";
   };
 
-  openStateIcon: Function = (): React$Element<*> => (
-    <Icon
+  isInResults: Function = (id: string): boolean => {
+    if (this.getResults()) {
+      const elmId = (id === "birth_location") ? "profile.birth_country3" : id;
+      const docCount = _.get(this.getResults(), ['aggregations', elmId, 'doc_count']);
+
+      if (docCount > 0) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  openStateIcon: Function = (children: React$Element<*>): React$Element<*>|null => {
+    if (this.getResults() === null || !this.isInResults(children.props.id)) {
+      return null;
+    }
+
+    return <Icon
       name="arrow_drop_down"
       onClick={this.toggleFilterVisibility}
       className={this.openClass()}
-    />
-  );
+    />;
+  }
 
   toggleFilterVisibility: Function = (): void => {
     const {
@@ -39,7 +57,7 @@ export default class FilterVisibilityToggle extends React.Component {
     const { children } = this.props;
     return (
       <div className={`filter-visibility-toggle ${this.openClass()}`}>
-        { this.openStateIcon() }
+        { this.openStateIcon(children) }
         { children }
       </div>
     );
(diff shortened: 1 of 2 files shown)


noisecapella:
It seems like `FilterVisibilityToggle` is the only class you would need to modify for this, since it wraps the facets, and since the visibility icons are the only things that appear to show up when there is no data. Can you change this to modify `FilterVisibilityToggle` instead? It seems like you can use `this.props.children` to tell whether the data is errored or if there is no data

amir-qayyum-khan:
@noisecapella ``this.props.children`` gives one valid react component which is no rendered at parent level.
Also if their is no data then these components still renders with empty div.
i did not find any way to see from parents that child have data against given fields, 

if anything i am missing?

noisecapella:
How about this instead:
 - change `FilterVisibilityToggle` to extend `SearchkitComponent`
 - This will give you access to `this.getResults()` within `FilterVisibiltyToggle`. You can then use the aggregation `doc_count` information to see whether to show the toggle or not. If `getResults()` returns `null` for some reason (permissions error for example) we should also not show the toggle

amir-qayyum-khan:
Done @noisecapella

static/js/components/search/FilterVisibilityToggle.js
```diff
@@ -15,13 +17,29 @@ export default class FilterVisibilityToggle extends React.Component {
     return checkFilterVisibility(filterName) ? "" : "closed";
   };
 
-  openStateIcon: Function = (): React$Element<*> => (
-    <Icon
+  isInResults: Function = (id: string): boolean => {
+    if (this.getResults()) {
+      const elmId = (id === "birth_location") ? "profile.birth_country3" : id;
+      const docCount = _.get(this.getResults(), ['aggregations', elmId, 'doc_count']) || 0;
```
noisecapella: `_.get`'s third parameter is the value returned when it can't find the value at the path given. Can you use that instead of `|| 0`?

static/js/components/search/FilterVisibilityToggle.js
```diff
@@ -15,13 +17,29 @@ export default class FilterVisibilityToggle extends React.Component {
...
+
+      if (docCount > 0) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  openStateIcon: Function = (children: React$Element<*>): React$Element<*>|null => {
+    if (this.getResults() === null || !this.isInResults(children.props.id)) {
```
noisecapella: The code already does the `this.getResults()` check in `isInResults`, can you remove that part of the expression?

static/js/components/search/FilterVisibilityToggle_test.js
```diff
@@ -33,6 +34,15 @@ describe('FilterVisibilityToggle', () => {
     };
   });
 
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+    sandbox.stub(FilterVisibilityToggle.prototype, 'getResults').returns(null);
```
noisecapella: Can you move this line into the test right before it's needed?

static/js/components/search/FilterVisibilityToggle_test.js
```diff
@@ -45,12 +55,52 @@ describe('FilterVisibilityToggle', () => {
...
+        <div id="test">Test Text</div>
+      </FilterVisibilityToggle>
     );
-    let icon = TestUtils.findRenderedDOMComponentWithTag(
+    let icon = TestUtils.scryRenderedDOMComponentsWithTag(
+      componentTree, 'i'
+    );
+    assert.equal(icon.length, 0);
+
+    // when search results available.
```
noisecapella: Please break these cases into separate tests

amir-qayyum-khan:
Done @noisecapella

static/js/components/search/FilterVisibilityToggle_test.js
```diff
@@ -33,19 +35,66 @@ describe('FilterVisibilityToggle', () => {
     };
   });
 
+  beforeEach(() => {
```
noisecapella: Can you combine this `beforeEach` with the one right above it?

static/js/components/search/FilterVisibilityToggle_test.js
```diff
@@ -33,19 +35,66 @@ describe('FilterVisibilityToggle', () => {
...
+
   it('sets filter visibility when clicked', () => {
+    sandbox.stub(FilterVisibilityToggle.prototype, 'getResults').returns({
+      aggregations: {
+        test: {
+          doc_count: 9
+        }
+      }
+    });
     let componentTree = TestUtils.renderIntoDocument(
```
noisecapella: Can you change this test to use enzyme too?

amir-qayyum-khan:
it is already there,  see the next test.

amir-qayyum-khan:
i mean see the lsst test in file. it has case when doc_count > 0

Issue #1408, opened 2016-10-17
course run without enrollment start date causes a 500 error on the program page


pdpinch:
```
Traceback:  

File "/app/.heroku/python/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
 149.                     response = self.process_exception_by_middleware(e, request)

File "/app/.heroku/python/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
 147.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/newrelic-2.54.0.41/newrelic/hooks/framework_django.py" in wrapper
 499.                 return wrapped(*args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/wagtail/wagtailcore/views.py" in serve
 27.     return page.serve(request, *args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/wagtail/wagtailcore/models.py" in serve
 764.             self.get_context(request, *args, **kwargs)

File "./cms/models.py" in get_context
 172.         context = get_program_page_context(self, request)

File "./cms/models.py" in get_program_page_context
 187.         "courses": courses_for_popover(courses_query),

File "./cms/models.py" in courses_for_popover
 33.     return [CourseSerializer().to_representation(c) for c in courses]

File "./cms/models.py" in <listcomp>
 33.     return [CourseSerializer().to_representation(c) for c in courses]

File "/app/.heroku/python/lib/python3.5/site-packages/rest_framework/serializers.py" in to_representation
 463.                 attribute = field.get_attribute(instance)

File "/app/.heroku/python/lib/python3.5/site-packages/rest_framework/fields.py" in get_attribute
 405.             return get_attribute(instance, self.source_attrs)

File "/app/.heroku/python/lib/python3.5/site-packages/rest_framework/fields.py" in get_attribute
 80.                 instance = getattr(instance, attr)

File "./courses/models.py" in enrollment_text
 139.                 end=end_text,

Exception Type: UnboundLocalError at /scm/
Exception Value: local variable 'end_text' referenced before assignment
```

amir-qayyum-khan:
#### To produce it:

- Create future course i.e start date after today. lets say 2017.
- Remove enrollment start date date
- Open program page. you will see exception.

[...]

PR #1416, opened 2016-10-17
Fix a bug in course enrollment text

singingwolfboy:
Fix for #1408. Restores the correct logic from the `get_course_enrollment_text()` function in 89d844044e578a179de3bce01067a927e618c872.


noisecapella:
Can you add a test covering this case too?

singingwolfboy:
Test added.

PR #1417, opened 2016-10-17
Fixed error in year validation logic

### LINKED ISSUE #1405: Birthdate year validation on the Profile Page is incorrect
A user can enter a value for their date of birth year ranging from 1800-2100.
Steps to reproduce:
Log onto MicroMasters.
Create a new edX profile.
On the first profile page, tab down (or mouse) to the Date of Birth Field.
For the Year value if a user enters "1111" the value changes to "1800"
(or you can manually enter 1800).
Similarly, a user can enter a value greater than 2017 all the way up 2100.
(or you can enter "5555" which reverts to "2100")

aliceriot:

closes #1405 

#### What's this PR do?

this fixes the year validation for the profile. basically, the range of allowed years `n` now looks like `x - 120 <= n <= x` where `x` is the current year. when `x = 2016` the range is `1896 - 2016`.

look over the logic and test changes.

go to `/learner` and open the basic info dialog. edit your birthday. you should only be able to enter years in the range above. years greater than or less than the bounds should be corrected to the closest bound (so that `1700 -> 1896`, and `2250 -> 2016`, for instance).

### DIFF
--- a/static/js/util/profile_edit_test.js
+++ b/static/js/util/profile_edit_test.js
@@ -1,6 +1,7 @@
 import { assert } from 'chai';
 import sinon from 'sinon';
 import { shallow } from 'enzyme';
+import moment from 'moment';
 
 import {
   boundTextField,
@@ -9,6 +10,7 @@ import {
   saveProfileStep,
 } from './profile_edit';
 import * as validation from '../util/validation';
+import { YEAR_VALIDATION_CUTOFF } from '../constants';
 
 describe('Profile Editing utility functions', () => {
   let that, sandbox;
@@ -149,6 +151,11 @@ describe('Profile Editing utility functions', () => {
       }
     };
 
+    let currentYear = moment().year();
+    let cutoff = currentYear - YEAR_VALIDATION_CUTOFF;
+    let tooOld = cutoff - 5;
+    let tooYoung = currentYear + 5;
+
     it("has proper props for an invalid or missing value", () => {
       for (let dateOfBirth of ['', {}, null, undefined]) {
         that.props.profile.date_of_birth = dateOfBirth;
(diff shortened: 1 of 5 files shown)


static/js/util/profile_edit_test.js
```diff
@@ -268,21 +272,21 @@ describe('Profile Editing utility functions', () => {
       that.props.profile.date_of_birth = "1917-11-07";
       rerender(false);
       yearTextField.props.onChange({target: {value: "1233"}});
```
gsidebo: can you set a `futureYear` and `tooFarInPastYear` (et al) variable near `currentYear` and `cutoff` and replace the hardcoded values (eg: "1233", "2077") with those?

static/js/util/validation.js
```diff
@@ -294,14 +294,16 @@ export function validateYear(input: string|number|null): Maybe<number> {
   if (year === undefined) {
     return Nothing();
   }
-  if ( year < 1800 ) {
+  let now = moment().year();
+  let cutoff = now - 120;
```
gsidebo: can you put the `120` value into constants.js and refer to it here and in the unit tests?

static/js/util/validation_test.js
```diff
@@ -385,15 +385,17 @@ describe('Profile validation functions', () => {
...
 
-    it('returns 2100 for years >= 2100', () => {
-      assertMaybeEquality(Just(2100), validateYear("2100"));
-      assertMaybeEquality(Just(2100), validateYear("2300"));
-      assertMaybeEquality(Just(2100), validateYear("52300"));
+    it('returns a maximum of the current year', () => {
+      let now = moment().year();
+      assertMaybeEquality(Just(now), validateYear("2100"));
+      assertMaybeEquality(Just(now), validateYear("2300"));
+      assertMaybeEquality(Just(now), validateYear("52300"));
```
gsidebo: these are redundant. i think this assertion alone would do the trick

`assertMaybeEquality(Just(now), moment().add(3, "years").year());`

aliceriot: oh yea, I was just blindly fixing the existing tests there, haha

aliceriot:
@gsidebo comments addressed

gsidebo:
looks great! 👍

PR #1431, opened 2016-10-18
Added persistent connection settings for DB

blarghmatey:
#### What's this PR do?
Adds persistent DB connections by default and adds the beat scheduler to the Celery worker

Using the locust load testing tool to validate performance improvements

When this config is absent a new connection is made to the DB for every transaction which adds significant overhead. Adding this setting reduces latency and I/O for improved performance and capacity.

- Added var for setting DB persistent connections
- Added Celery beat and extra_worker to Procfile for task scheduling


giocalitri:
👍

PR #1436, opened 2016-10-18
Add/remove custom tabs on program page

### LINKED ISSUE #597: Allow CMS user to add (and remove) tabs
As a CMS user, I may want to

annagav:
Fix #597
#### What's this PR do?

Add GenericPage model to cms, for creating custom tabs on the program page.

To test this, in cms add a child page to an existing ProgramPage.

screen shot 2016-10-18 at 2 06 39 pm

Then define the title (tab name) and content of the page and click publish.

screen shot 2016-10-18 at 2 08 27 pm


cms/models.py
```diff
@@ -97,6 +97,30 @@ class CategorizedFaqsPage(Page):
     parent_page_types = ['FaqsPage']
 
 
+class GenericPage(Page):
+    """
+    CMS Page for custom tabs on the program page
```
gsidebo: if this is an accurate description of the class, i think `ProgramTab` or `ProgramTabPage` would be a better/more accurate name than `GenericPage`

cms/models.py
```diff
@@ -97,6 +97,30 @@ class CategorizedFaqsPage(Page):
     parent_page_types = ['FaqsPage']
 
 
+class GenericPage(Page):
```
gsidebo: we can abstract the idea of `ProgramPage` children like this class and `FaqsPage`. it can be defined like this:

```python
class ProgramChildPage(Page):
    parent_page_types = ['ProgramPage']

    def parent_page(self):
        """ Get the parent ProgramPage"""
        return ProgramPage.objects.ancestor_of(self).first()

    def get_context(self, request):
        context = get_program_page_context(self.parent_page(), request)
        context['page'] = self
        context['active_tab'] = self.title
        return context
```

this class and `FaqPage` could inherit from that. note that `FaqPage` would have a `context['page']` key instead of a `context['faqpage']` key if you made this change. that's fine since it's just the name of a variable in the template.

annagav:
@gsidebo, I refactored the program child pages.

gsidebo:
code looks great! going to take a quick look at functionality

gsidebo:
@annagav just wondering, is this what the FAQ edit page should look like? I can only add a title. Looks like I can add child page to an FAQ page, and then add some questions to that child page. Just want to make sure this is intended

ss 2016-10-21 at 10 41 45

annagav:
Yes, this is how we implemented Categories for faqs.

gsidebo:
looks great! 👍

Issue #1421, opened 2016-10-17
ecommerce: audit table for orders


pdpinch:
We need to create a permanent record of changes to orders, similar to the financial aid audit table.

PR #1447, opened 2016-10-18
Added serialize_model, replaced to_dict

noisecapella:
Part of #1421 

#### What's this PR do?
Creates a general serializer method for models

Checkout master and make a financial aid request using the UI. Verify that `FinancialAidAudit` has information in `data_before` and `data_after` matching the most recent request. Then checkout this branch and in the UI skip financial aid. Verify that the information in `FinancialAidAudit` is substantially the same. `data_before` for the most recent entry and `data_after` for the second most recent entry should match up almost exactly. The formatted datetime output may differ slightly and there will be extra fields, but nothing else should change.

### DIFF
--- a/micromasters/utils.py
+++ b/micromasters/utils.py
@@ -65,7 +65,7 @@ def custom_exception_handler(exc, context):
         )
 
 
-def serialize_model(obj):
+def serialize_model_object(obj):
     """
     Serialize model into a dict representable as JSON
 
@@ -77,6 +77,6 @@ def serialize_model(obj):
     """
     # serialize works on iterables so we need to wrap object in a list, then unwrap it
     data = json.loads(serialize('json', [obj]))[0]
-    fields = data['fields']
-    fields['id'] = data['pk']
-    return fields
+    serialized = data['fields']
+    serialized['id'] = data['pk']
+    return serialized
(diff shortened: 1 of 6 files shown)


micromasters/utils.py
```diff
@@ -62,3 +63,20 @@ def custom_exception_handler(exc, context):
             status=status.HTTP_500_INTERNAL_SERVER_ERROR,
             data=[formatted_exception_string]
         )
+
+
+def serialize_model(obj):
```
gsidebo: i think `serialize_model_object` would be a more helpful name here

micromasters/utils.py
```diff
@@ -62,3 +63,20 @@ def custom_exception_handler(exc, context):
...
+
+    Args:
+        obj (django.db.models.Model): An instantiated Django model
+    Returns:
+        dict:
+            A representation of the model
+    """
+    # serialize works on iterables so we need to wrap object in a list, then unwrap it
+    data = json.loads(serialize('json', [obj]))[0]
+    fields = data['fields']
```
gsidebo: nit: 'fields' is only really relevant as a key name in the return value of `serialize`. `serialized` or `serialized_obj` is probably more on the mark for this variable

noisecapella:
Comments addressed, I'll merge after tests passed since you approved the PR

Issue #1391, opened 2016-10-14
Search: display error message when there are no search results


pdpinch:
The "Results box" in the right column will be empty if the user has selected some filters and there are no results.

We could show a different message depending on the reason the box is empty, or we could just show the same message for both cases, as I am assuming that is easier to do.

![image](https://cloud.githubusercontent.com/assets/20047260/19354203/f2819882-9133-11e6-8d43-22e5585d38c8.png)

Issue #1390, opened 2016-10-14
Search: display error message when there are no users in program 


pdpinch:
The "Filter box" in the left column will only be empty if there are no users in a program.

The "Results box" in the right column will be empty either if there are no users in a program (as shown above), OR if the user has selected some filters and there are no results.

![image](https://cloud.githubusercontent.com/assets/20047260/19354203/f2819882-9133-11e6-8d43-22e5585d38c8.png)

amir-qayyum-khan:
depends on https://github.com/mitodl/micromasters/pull/1407

pdpinch:
Why does this depend on #1407?

Issue #1366, opened 2016-10-13
Search: return error messsage on failed permissions


pdpinch:
The search API appropriately prevents users without the appropriate role from searching. However, the UI is not clear about why the search failed. 

<img width="986" alt="screen shot 2016-10-13 at 12 02 04 pm" src="https://cloud.githubusercontent.com/assets/430126/19356778/eecf6bde-913c-11e6-8a36-1721c937802e.png">

pdpinch:
@roberthouse54 what are your thoughts on this? Should we direct the user to a generic error page (usually a 404 for permissions failures) or return something in the UI?

Ferdi:
This is clearly a 404 or 403

amir-qayyum-khan:
@pdpinch @Ferdi 

- in case of no results on search page there is pr for this https://github.com/mitodl/micromasters/pull/1449
 
- if user is neither staff nor instructor should we take him to 404 in case he hit learners page?

pdpinch:
Yes. See #1366

PR #1449, opened 2016-10-19
Shown message when no search results

amir-qayyum-khan:

- fixes https://github.com/mitodl/micromasters/issues/1391
- fixes https://github.com/mitodl/micromasters/issues/1390
- fixes https://github.com/mitodl/micromasters/issues/1366

#### What's this PR do?

Displayed message when no search results.
- see static/js/components/LearnerSearch.js
- open `/learners/` page when no results or no users to program

screen shot 2016-10-19 at 7 27 17 pm

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -4,7 +4,6 @@ import {
   SearchkitComponent,
   HierarchicalMenuFilter,
   Hits,
-  NoHits,
   SelectedFilters,
   RefinementListFilter,
   HitsStats,
@@ -17,6 +16,7 @@ import Grid, { Cell } from 'react-mdl/lib/Grid';
 import Card from 'react-mdl/lib/Card/Card';
 import iso3166 from 'iso-3166-2';
 import { StickyContainer, Sticky } from 'react-sticky';
+import _ from 'lodash';
 
 import ProgramFilter from './ProgramFilter';
 import LearnerResult from './search/LearnerResult';
@@ -26,6 +26,7 @@ import CustomResetFiltersDisplay from './search/CustomResetFiltersDisplay';
 import CustomSortingSelect from './search/CustomSortingSelect';
 import FilterVisibilityToggle from './search/FilterVisibilityToggle';
 import HitsCount from './search/HitsCount';
+import CustomNoHits from './search/CustomNoHits';
 import EmailCompositionDialog from './EmailCompositionDialog';
 import type { Option } from '../flow/generalTypes';
 import type { Email } from '../flow/emailTypes';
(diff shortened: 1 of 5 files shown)


static/js/components/search/CustomNoHits.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import React from 'react';
+import { NoHits } from 'searchkit';
+
+export default class CustomNoHits extends NoHits {
+  render() {
+    if ((this.hasHits() || this.isInitialLoading() || this.isLoading()) && !this.getError()) {
```
amir-qayyum-khan: no, this widget shows error message

static/js/components/LearnerSearch.js
```diff
@@ -99,6 +99,91 @@ export default class LearnerSearch extends SearchkitComponent {
...
+        <Cell col={12} className="mm-filters">
+          <SelectedFilters />
+          <ResetFilters component={CustomResetFiltersDisplay}/>
+        </Cell>
+      </Grid>
+    );
+  }
+
+  renderFacets: Function = (): React$Element<*> => {
+    if (this.getResults() === null) {
```
amir-qayyum-khan: there was unexpected issue with here, because  of ```if (!this.getResults())``` facet panel was rendering ambiguously. showing error message.

static/js/components/LearnerSearch.js
```diff
@@ -99,6 +99,91 @@ export default class LearnerSearch extends SearchkitComponent {
...
+            {...this.props}
+            filterName="grade-average"
+          >
+            <RangeFilter
+              field="program.grade_average"
+              id="grade-average"
+              min={0}
+              max={100}
+              showHistogram={true}
+              title="Program Avg. Grade"
```
amir-qayyum-khan: this is out of context of this PR, can you create a separate issue for this?

static/scss/search-page.scss
```diff
@@ -361,3 +361,25 @@
...
+  align-items: center;
+  background-color: #efefef;
+  margin: 7px 0 4px 0;
+  height: 300px;
+  color: #848484;
+  font-style: italic;
+  font-weight: 200;
+
+  &.left-nav {
+    margin: 19px 6px 22px 6px !important;
```
amir-qayyum-khan: removed, it was useless

amir-qayyum-khan:
done @singingwolfboy

static/js/components/LearnerSearch.js
```diff
@@ -99,6 +100,91 @@ export default class LearnerSearch extends SearchkitComponent {
...
+      </Grid>
+    );
+  }
+
+  renderFacets: Function = (): React$Element<*> => {
+    if (_.isNull(this.getResults())) {
+      return (
+        <Card className="fullwidth" shadow={1}>
+          <div className="no-hits left-nav">
+            There are no users in the program.
```
singingwolfboy: Is it possible to display the program name, to make this clearer? For example, if the program is named "Analog Learning", this message should say: "There are no users in the Analog Learning program."

If this is too difficult, we can at least make this clearer by changing the word "the" to the word "this", so that the message says: "There are no users in this program."

static/js/components/search/CustomNoHits.js
singingwolfboy: That's a little concerning. What happens if `this.getError()` returns errors? This `render()` function will return `null`, and the user will just see a blank screen instead of the error messages.

Can you add a test where `this.hasHits()` returns `false`, but `this.getError()` includes at least one error? I would expect the screen to show that error message to the user, so the user knows *why* they're having a problem.

static/js/components/LearnerSearch.js
singingwolfboy: Good point. I've created #1537.

amir-qayyum-khan:
@singingwolfboy done

static/js/components/search/CustomNoHits.js
amir-qayyum-khan: dine

static/js/components/search/CustomNoHits.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    if (this.getError()) {
+      const suggestion = this.getSuggestion();
+      const query = this.getQuery().getQueryString();
+      let infoKey = suggestion ? "NoHits.NoResultsFoundDidYouMean" : "NoHits.NoResultsFound";
+
+      return (
+        <div className="no-hits">
+          {this.translate(infoKey, {query:query, suggestion:suggestion})}
```
singingwolfboy: If the query is empty, the string ends up as "No results found for .", which is exactly what I saw when I tested it:
screen shot 2016-10-26 at 3 30 49 pm

Maybe treat an empty query as a special case, and ask the user to search for something? @roberthouse54, do you have any thoughts on how this should be handled?

roberthouse54: Is there any reason we can't just say "There are no results for this search."

amir-qayyum-khan: @singingwolfboy empty query also appears when user tries to access ``/learners`` page. but he is not staff.
 thats why he dont see any results. i am going to show ``There are no results for this search.`` message for this case.

amir-qayyum-khan:
Done @singingwolfboy

amir-qayyum-khan:
Done @singingwolfboy

static/js/components/search/CustomNoHits.js
```diff
@@ -0,0 +1,32 @@
...
+  render() {
+    let message;
+
+    if ((this.hasHits() || this.isInitialLoading() || this.isLoading()) && !this.getError()) {
+      return null;
+    }
+
+    if (this.getError()) {
+      if (this.getError().data && this.getError().data.error) {
+        message = this.getError().data.error;
```
singingwolfboy: This is great! However, while I was manually testing this, I discovered that Django Rest Framework sets the error description in a property called `detail`, not a property called `error`. As a result, if I try the search as a user that is forbidden to access the search API due to lack of permissions, I get this AJAX response:
error-message
But the UI displays the generic error message:
generic-error

Could we modify the error response in `search/views.py` to use the `detail` property, and then make this Javascript check that _instead_ of the `error` property? That way, the detailed error message will show up in either case.

amir-qayyum-khan:
Done @singingwolfboy

PR #1456, opened 2016-10-19
Added audit table for Order, Line

### LINKED ISSUE #1421: ecommerce: audit table for orders

noisecapella:
Fixes #1421 

#### What's this PR do?
Adds an audit table for `Order` and code to log changes to `Order` in the checkout and order fulfillment APIs.

Create an order by clicking the 'Buy now' button. Verify that there is a change in `OrderAudit` which also includes the data for the `Line`.

Then feed the purchase email to the order fulfillment API. (This doesn't have to succeed, failure will work too.) Verify that there is a new change in `OrderAudit` reflecting that change of status.


ecommerce/models.py
```diff
@@ -46,6 +50,43 @@ def __str__(self):
         """Description for Order"""
         return "Order {}, status={} for user={}".format(self.id, self.status, self.user)
 
+    def to_dict(self):
+        """
+        Get a serialized representation of the Order and any attached Lines
+        """
+        data = serialize_model(self)
+        data['lines'] = [serialize_model(line) for line in self.line_set.all()]
+        return data
```
gsidebo: i'm sort of confused by this pattern now. why aren't we just defining a new django rest framework ModelSerializer to perform this serialization? especially looking at `'lines'` - that is explicitly looping through and serializing a related object, which is one of the things that DRF serializers takes care of for you. also, i don't think it's optimal that this class gets its serialization by calling `obj.to_dict()` while the `FinancialAid` class does the same by invoking `serialize_model(obj)`. i think we should have a mixin/decorator for any class that needs a serialized version of its objects for an audit table. that way we could have a consistent API for auditing. 

we should probably have a base model for `*Audit` tables too. `acting_user`, `data_before/after` and `save_and_log` should behave pretty much the same for all audit tables, right?

noisecapella: Sure, I'll look into making the audit tables share more code and be more consistent

gsidebo: fair enough. as long as we have a somewhat consistent API for the serialization, this approach should be fine

noisecapella:
What do you think of the code as it is? I haven't started working on fixing or adding tests yet

noisecapella:
All tests should be there now

gsidebo:
sorry, didn't see that previous comment. looking now

micromasters/models.py
```diff
@@ -0,0 +1,83 @@
...
+    class Meta:
+        abstract = True
+
+    def to_dict(self):
+        """
+        Returns:
+            dict:
+                A serialized representation of the model object
+        """
+        raise NotImplementedError
```
gsidebo: i feel like `serialize_mobel_object(self)` is a reasonable base implementation for this

noisecapella: I'd rather leave it the way it is since this requires the implementing class to be explicit in how they are serializing the object in the subclass.

micromasters/models.py
```diff
@@ -0,0 +1,83 @@
...
+    SET_NULL,
+)
+
+
+class AuditModel(Model):
+    """An abstract base class for audit models"""
+
+    acting_user = ForeignKey(User, null=True, on_delete=SET_NULL)
+    created_on = DateTimeField(auto_now_add=True)
+    updated_on = DateTimeField(auto_now=True)
```
noisecapella: Sure, I'll make this change

noisecapella:
Comments addressed

gsidebo:
code looks great. going to test functionality again

PR #1457, opened 2016-10-19
Change gen_fake_data command to also make CoursePrice objects

singingwolfboy:
Because `CoursePrice` objects are necessary to make the application work. :(

### DIFF
--- a/courses/management/commands/gen_fake_data.py
+++ b/courses/management/commands/gen_fake_data.py
@@ -10,6 +10,7 @@ from courses.factories import (
     CourseRunFactory,
     ProgramFactory,
 )
+from ecommerce.factories import CoursePriceFactory
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


noisecapella:
We should just get rid of `gen_fake_data` IMO. `gen_realistic_search_data` also exists and covers its use cases. Can you remove that management command and also update `gen_realistic_search_data` to add `CoursePrice` as you did here for `gen_fake_data`?

gsidebo:
on top of that, can you change this to only add a CoursePrice if one doesn't already exist? you could use get_or_create

courses/management/commands/gen_fake_data.py
```diff
@@ -53,6 +62,9 @@ def handle(self, *args, **options):
                 course_runs_range = range(randint(1, int(options['course_runs'])))
                 for _ in course_runs_range:
                     record_count += 1
-                    CourseRunFactory.create(course=course)
+                    run = CourseRunFactory.create(course=course)
+                    if options['course_prices']:
+                        CoursePriceFactory.create(course_run=run)
```
noisecapella: We should also set `is_valid=True` here. Also, all course runs need to have the same price

noisecapella: All course runs for a single program I mean

singingwolfboy:
Closing in favor of #1463

Issue #1144, opened 2016-09-27
Move course price from CoursePrice into Program


noisecapella:
This isn't an urgent change but at the same time I feel like we will want this soon for consistency's sake.

gsidebo:
To clarify, I think this issue is suggesting that we move `CoursePrice` into courses/models.py.

Just an example of an unfortunate side effect of this model being defined in ecommerce instead of courses: the need to import `CoursePrice` within `courses.models.Program#get_course_price`. 

This issue will involve creating the necessary migrations that will change the name of the table associated with the model (and, if necessary, migrate existing data from one table to the other).

noisecapella:
@gsidebo This is suggesting removing the `CoursePrice` model entirely and storing only a `price` field on `Program`

PR #1463, opened 2016-10-19
Seed data app

singingwolfboy:
This pull request deletes the `gen_fake_data` management command, and renames the `gen_realistic_search_data` management command to `seed_db` (and renames `delete_realistic_search_data` to `unseed_db`). It also makes this command create `CoursePrice` objects to match the `CourseRun` objects it creates.

I've also updated the `README.md` to reference this management command. Since this command calls `search.indexing_api.recreate_index`, it is no longer necessary to call the `recreate_index` management command, so that has been removed from the README.

Reviewers should be sure to try running these commands in several different configurations to be sure that they work properly:

    docker-compose run web ./manage.py seed_db
    docker-compose run web ./manage.py create_tiers

The `create_tiers` command should be rolled into the `seed_db` command, but that can be done in a later pull request. Since it sounds like we want to refactor the tiers models, I didn't want to touch that in this pull request.

This pull request does not have automated tests, because this functionality is not used in production, and developers will be manually testing this functionality every time they reinstall the application. For this specific feature, writing and running automated tests is a waste of time and effort.

In the process of writing this pull request, I discovered [Django-seed](https://github.com/Brobin/django-seed), which seems designed for this use-case. However, rewriting this code to use Django-seed is unnecessary when it works fine as-is, and wouldn't gain us any real benefit.


README.md
```diff
@@ -176,21 +176,21 @@ Run this command:
...
+#### 4) Create initial data
+First, create the database tables from the Django models:
 
     docker-compose run web ./manage.py migrate
 
-#### 5) Create an Elasticsearch index
-To do this, run this command:
+Then, populate example data into these tables, for development purposes:
 
-    docker-compose run web ./manage.py recreate_index
```
noisecapella: I don't think we should remove this step. `recreate_index` is essential for deploying the app, and production instances will not call `seed_db` beforehand

seed_data/management/programs.json
```diff
@@ -176,7 +232,11 @@
                         "edx_course_key": "course-v1:MITx+Analog+Learning+200+Jan_2016",
                         "title": "Analog Learning 200 - January 2016",
                         "start_date": "2016-01-15T00:00:00",
-                        "enrollment_end": "2016-01-29T00:00:00"
+                        "enrollment_end": "2016-01-29T00:00:00",
+                        "course_price": {
+                            "price": 50,
```
noisecapella: All course runs in a program need to have the same price. Can you make them consistent?

singingwolfboy: Sure, I can do that. But if that's the case, shouldn't `CoursePrice` have a foreign key to `Program`, instead of `CourseRun`?

noisecapella: Yes, I filed #1144 for that

seed_data/management/commands/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+Globals for commands relating to seeding data into the database
+"""
+
+from datetime import datetime, timedelta
+
+USER_DATA_PATH = 'seed_data/management/users.json'
+PROGRAM_DATA_PATH = 'seed_data/management/programs.json'
+FAKE_USER_USERNAME_PREFIX = 'fake.'
+FAKE_PROGRAM_DESC_PREFIX = '[FAKE] '
+CACHED_MODEL_LAST_REQUEST = datetime.now() + timedelta(days=365)
```
noisecapella: Can we calculate this at the time when we use it?

singingwolfboy: Done.

seed_data/management/programs.json
singingwolfboy: Great. I've changed this file to make the prices consistent. (The two programs have two different prices, though.)

noisecapella:
Can you rename `delete_realistic_search_data` to match the rename of the other management command?

singingwolfboy:
@noisecapella I already did. It's called `unseed_db`.

seed_data/management/programs.json
```diff
@@ -16,7 +16,11 @@
                         "edx_course_key": "course-v1:MITx+Digital+Learning+100+Jan_2015",
                         "title": "Digital Learning 100 - January 2015",
                         "start_date": "2015-01-15T00:00:00",
-                        "enrollment_end": "2015-01-29T00:00:00"
+                        "enrollment_end": "2015-01-29T00:00:00",
+                        "course_price": {
+                            "price": 1000,
+                            "is_valid": true
```
gsidebo: i think we should get rid of all of these `is_valid` key-values and just set that value to True when we deserialize. we do this when we deserialize Programs (see `deserialize_program_data_list`). i know there are drawbacks when we set values in two different places, but it's safe to assume we want a specific value for this attribute when we're seeding the db (like we do with `Program.live`). also, we get a shorter json file

PR #1464, opened 2016-10-19
Updated the app.json to include required env vars

blarghmatey:
**Updated the app.json to include required env vars**
Updated `app.json` to include the environment variables that should be
present for running the app. This will fix issues with PR apps not
including the necessary settings from the CI environment.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,106 +1,129 @@
 {
-    "name": "micromasters",
-    "description": "Web Portal for Micromasters",
-    "keywords": [
-        "Django",
-        "Python",
-        "MIT",
-        "Office of Digital Learning"
-    ],
-    "website": "https://github.com/mitodl/micromasters",
-    "repository": "https://github.com/mitodl/micromasters",
-    "success_url": "/",
-    "scripts": {
-        "postdeploy": "./manage.py migrate"
+  "addons": [
+    "heroku-postgresql:hobby-dev",
+    "newrelic:wayne"
+  ],
+  "buildpacks": [
+    {
+      "url": "https://github.com/heroku/heroku-buildpack-nodejs"
     },
-    "addons": [
-        "heroku-postgresql:hobby-dev",
-        "newrelic:wayne"
-    ],
-    "buildpacks": [
-        {
-            "url": "https://github.com/heroku/heroku-buildpack-nodejs"
-        },
-        {
-            "url": "https://github.com/heroku/heroku-buildpack-python"
-        }
-    ],
-    "env": {
-        "ALLOWED_HOSTS": {
-            "description": "Array of allowed hostnames",
-            "default":  "['*']"
-        },
-        "MICROMASTERS_ADMIN_EMAIL": {
-            "description": "E-mail to send 500 reports to.",
-            "required": false
-        },
-        "MICROMASTERS_EMAIL_HOST": {
-            "description": "Outgoing e-mail settings",
-            "required": false
-        },
-        "MICROMASTERS_EMAIL_PASSWORD": {
-            "description": "Outgoing e-mail settings",
-            "required": false
-        },
-        "MICROMASTERS_EMAIL_PORT": {
-            "description": "Outgoing e-mail settings",
-            "required": false,
-            "value":  "587"
-        },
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -1,106 +1,129 @@
...
+    },
+    "SENTRY_DSN": {
+      "description": "The connection settings for Sentry"
+    },
+    "STATUS_TOKEN": {
+      "description": "Token to access the status API.",
+      "required": true
     }
+  },
+  "keywords": [
```
pdpinch: what are these for?

aliceriot:
looks good! is there anything you suggest doing to test this?

app.json
```diff
@@ -1,106 +1,129 @@
...
+    }
+  ],
+  "description": "Web Portal for Micromasters",
+  "env": {
+    "ALLOWED_HOSTS": {
+      "default": "['*']",
+      "description": "Array of allowed hostnames"
+    },
+    "AWS_ACCESS_KEY_ID": {
+      "description": "AWS Access Key for S3 storage."
```
pdpinch: what does it mean if there's no value for "required" ?

app.json
```diff
@@ -1,106 +1,129 @@
...
+    "MIT",
+    "Office of Digital Learning"
+  ],
+  "name": "micromasters",
+  "repository": "https://github.com/mitodl/micromasters",
+  "scripts": {
+    "postdeploy": "./manage.py migrate"
+  },
+  "success_url": "/",
+  "website": "https://github.com/mitodl/micromasters"
```
pdpinch: what's this for?

app.json
blarghmatey: It just means that it will inherit from the cloned env (e.g. PR apps) but it won't block deployment if it's not present.

app.json
blarghmatey: Those were already there, I just sorted the whole JSON object so it messed with the diff. I'm honestly not sure what they're for.

blarghmatey:
WRT testing I think the way to check is if all of the configured vars were inherited from the CI environment without having to be explicitly set.

app.json
```diff
@@ -1,106 +1,129 @@
...
+  "keywords": [
+    "Django",
+    "Python",
+    "MIT",
+    "Office of Digital Learning"
+  ],
+  "name": "micromasters",
+  "repository": "https://github.com/mitodl/micromasters",
+  "scripts": {
+    "postdeploy": "./manage.py migrate"
```
noisecapella: Can we add `./manage.py recreate_index` here too?

blarghmatey: This doesn't get run on each deploy. See https://devcenter.heroku.com/articles/app-json-schema#scripts

blarghmatey:
I just checked and the vars didn't get copied, but I'm thinking that's because the updated app.json wasn't present in the app when the review instance got created.

app.json
noisecapella: It should theoretically only need to be run right after a new deploy, otherwise the app's signals should take care of keeping Elasticsearch up to date. But without a generated `ELASTICSEARCH_INDEX` we probably can't do this step automatically without colliding with other indexes

app.json
```diff
@@ -1,106 +1,129 @@
...
+    },
+    "EDXORG_CLIENT_SECRET": {
+      "description": "The OAuth client secret configured in the edX instance.",
+      "required": true
+    },
+    "ELASTICSEARCH_HTTP_AUTH": {
+      "descriptin": "Basic auth settings for connecting to Elasticsearch"
+    },
+    "ELASTICSEARCH_INDEX": {
+      "description": "Index to use on Elasticsearch"
```
noisecapella: Would it be possible to enable random values for this variable and allow all index names on the MicroMasters ES instance?

blarghmatey: That's not possible because of the security plugin that we are using on ES. If a PR app needs to use a different index than CI then it can set the var to anything prefixed by `micromasters-ci`

noisecapella: In that case we should probably remove `./manage.py recreate_index` from the postdeploy so that other deploys don't touch the same instance before we can set this environment variable manually

PR #1467, opened 2016-10-19
Changed discount_amount to have a min of 1 (so we never have multiple…

alvinsiu:
N/A

#### What's this PR do?
Changes the minimum for `TierProgramFactory.discount_amount` so that you have to explicitly create a $0-discount `TierProgram`.

The chance for tests to fail without this fix was already really low (like on the order of `n/12345`), and it was only caught because it ended up happening on one build.

… 0 discount tier programs)

### DIFF
--- a/financialaid/factories.py
+++ b/financialaid/factories.py
@@ -48,7 +48,7 @@ class TierProgramFactory(DjangoModelFactory):
     tier = SubFactory(TierFactory)
     discount_amount = FuzzyInteger(low=1, high=12345)
     current = FuzzyAttribute(FAKE.boolean)
-    income_threshold = FuzzyInteger(low=0, high=10000)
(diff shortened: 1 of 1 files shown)


financialaid/factories.py
```diff
@@ -46,7 +46,7 @@ class TierProgramFactory(DjangoModelFactory):
     """
     program = SubFactory(ProgramFactory)
     tier = SubFactory(TierFactory)
-    discount_amount = FuzzyInteger(low=0, high=12345)
+    discount_amount = FuzzyInteger(low=1, high=12345)
     current = FuzzyAttribute(FAKE.boolean)
     income_threshold = FuzzyInteger(low=0, high=10000)
```
alvinsiu: Sure, that makes sense!

PR #1470, opened 2016-10-19
Added unique constraint to CourseRun edx_course_key

### LINKED ISSUE #1409: Multiple course runs with the same edx_course_key causes a 500 error
I can't think of a reason why this should be allowed to happen, so we could put a unique constraint on this CourseRun.edx_course_key

```
Traceback:  

File "/app/.heroku/python/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
 149.                     response = self.process_exception_by_middleware(e, request)

File "/app/.heroku/python/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
 147.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/newrelic-2.54.0.41/newrelic/hooks/framework_django.py" in wrapper
 499.                 return wrapped(*args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/django/views/decorators/csrf.py" in wrapped_view
 58.         return view_func(*args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/django/views/generic/base.py" in view
 68.             return self.dispatch(request, *args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/newrelic-2.54.0.41/newrelic/hooks/component_djangorestframework.py" in _nr_wrapper_APIView_dispatch_
 27.         return wrapped(*args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/rest_framework/views.py" in dispatch
 466.             response = self.handle_exception(exc)

File "/app/.heroku/python/lib/python3.5/site-packages/rest_framework/views.py" in dispatch
 463.             response = handler(request, *args, **kwargs)

File "./dashboard/views.py" in get
 65.         enrollments = get_student_enrollments(request.user, edx_client)

File "./dashboard/api.py" in get_student_enrollments
 427.             update_cached_enrollment(user, enrollment, course_id, now)

File "/app/.heroku/python/lib/python3.5/contextlib.py" in inner
 30.                 return func(*args, **kwds)

File "./dashboard/api.py" in update_cached_enrollment
 347.     course_run = CourseRun.objects.get(edx_course_key=course_id)

File "/app/.heroku/python/lib/python3.5/site-packages/django/db/models/manager.py" in manager_method
 122.                 return getattr(self.get_queryset(), name)(*args, **kwargs)

File "/app/.heroku/python/lib/python3.5/site-packages/django/db/models/query.py" in get
 391.             (self.model._meta.object_name, num)

Exception Type: MultipleObjectsReturned at /api/v0/dashboard/
Exception Value: get() returned more than one CourseRun -- it returned 2!
Request information:
```

gsidebo:

Fixes #1409 

#### What's this PR do?

Adds a unique constraint to CourseRun.edx_course_key

The following should cause an error in a django shell:

```python
from courses.models import Program, Course, CourseRun
edx_key = '12345'
course = Course.objects.first()
new_course_run = CourseRun.objects.create(
  course=course,
  edx_course_key=edx_key
)
CourseRun.objects.create(
  course=course,
  edx_course_key=edx_key
)
new_course_run.delete()
```

The second to last line should cause an integrity error

~~Not sure if we're going to have an issue with this `edx_course_key` being nullable. Only one run would be able to have a null value. Should this field be nullable anyway?~~ <-- this shouldn't be a problem (see comment below), but the reviewer should make sure to test this


pdpinch:
Oh, that will be a problem. We have the very real probability that a course will be created in the portal before it exists in EdX. The thinking was that they would have a null course-key until it can be filled in. 

Are there other options?

gsidebo:
@pdpinch looks like a false alarm. Apparently SQL doesn't consider NULL to equal NULL, and Django doesn't either since about 2008 😄  (https://github.com/django/django/commit/49ef21d9ea)

courses/models.py
```diff
@@ -151,7 +151,7 @@ class CourseRun(models.Model):
       rather a specific instance of that course being taught.
     """
     title = models.CharField(max_length=255)
-    edx_course_key = models.CharField(max_length=255, blank=True, null=True, db_index=True)
+    edx_course_key = models.CharField(max_length=255, blank=True, null=True, unique=True, )
```
gsidebo: `unique=True` automatically implies an indexed field, so `db_index=True` is redundant (https://docs.djangoproject.com/en/1.10/ref/models/fields/#unique)

PR #1473, opened 2016-10-20
Add app config for seed_data app

singingwolfboy:
Fix for a problem we're seeing on Heroku. @blarghmatey, can you test this out?


blarghmatey:
:+1:

PR #1476, opened 2016-10-20
Improve profile factories using Faker library

singingwolfboy:
[Faker](http://faker.readthedocs.io/en/master/) has some nifty generators that we might as well take advantage of, for generating realistic-looking fake data.

### DIFF
--- a/profiles/factories.py
+++ b/profiles/factories.py
@@ -5,6 +5,7 @@ from django.contrib.auth.models import User
 from factory import (
     Sequence,
     SubFactory,
+    LazyAttribute,
 )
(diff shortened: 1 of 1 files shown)


gsidebo:
nice! can you add a generator for first_name/last_name/preferred_name as well? also, not sure if it will work, but [.language_code()](http://faker.readthedocs.io/en/master/locales/en_AU.html#faker-providers-misc) might work for preferred_language

singingwolfboy:
Good thought. Done.

gsidebo:
dope 👍

Issue #1477, opened 2016-10-20
Program Page: fix line height in course card


pdpinch:
When course titles go to two lines, there should be less line height between the lines. 

<img width="368" alt="screen shot 2016-10-20 at 10 36 13 am" src="https://cloud.githubusercontent.com/assets/430126/19564490/be9ef0ae-96b1-11e6-969d-4e858a670a43.png">

PR #1480, opened 2016-10-20
Line-height fix for Course list on program page

roberthouse54:

#1477 

#### What's this PR do?

This is a very quick fix to the line-height and spacing in the courses box on the Program page.

Look at the layout of the Courses text in the gray box on the program page. Make sure the text is long enough so it breaks to 2 lines.

#### Screenshot 

image


annagav:
👍

Issue #1262, opened 2016-10-06
Pre-populate ZenDesk web widget with name, email and program


pdpinch:
We want to pass some data into the ZenDesk web widget to make it easier for users and also to help automate routing in ZenDesk. 
- [x] @pdpinch Set up a custom field in ZenDesk "MicroMasters Program" which can contain the current  user's current program or "None"
- [x] @pdpinch Update the ZenDesk widget code to accept these fields
- [x] Implement the [ZenDesk JavaScript API](https://developer.zendesk.com/embeddables/docs/widget/api) to set user name and email 
- [x] research how to pass the MicroMasters program name
- [ ] hide pre-populated fields so the user can't change them

pdpinch:
I've updated the widget to add the "MicroMasters Program" field, and to use the current background color.

amir-qayyum-khan:
yes, i saw update , i am getting some issue with local MM, just installing new instace

singingwolfboy:
It looks like the ZenDesk API does not use the [asynchronous module definition API](https://github.com/amdjs/amdjs-api/blob/master/AMD.md), which means it can't be imported in our Javascript files. I spent some time trying to figure out if we could shim it, but I don't see a good way to do so.  We'll probably just need to write a separate `<script>` tag on the page and reference the global [`zE.identify()`](https://developer.zendesk.com/embeddables/docs/widget/api#ze.identify) function.

pdpinch:
@singingwolfboy what's left on this?

singingwolfboy:
Currently, the program name will _only_ be pre-populated in the Zendesk widget if the user is on a program page. If the user is on any other page of the application, the program name will not be pre-populated.

It might make sense to pre-populate the program name in the widget on other pages in certain cases -- for example, if the user is only enrolled in one program, it's very likely that they are seeking help about that program. We may want to define exactly what logic should be used for pre-populating the program name on other pages, and then implement that logic.

pdpinch:
Some pages (notably the dashboard) allows the user to select their program context . We should use the value of that select box for the help widget.

pdpinch:
Just spoke to Ferdi and we'd like to keep this simple and not try to get the program from the dashboard. 

However, one change to add -- can we hide the fields that have values in them, so users can't change their name and e-mail (when we have it)?

pdpinch:
[...]

PR #1482, opened 2016-10-20
Zendesk prepopulate name/email

singingwolfboy:
Partial fix for #1262. This pull request sets the name and email address for the logged-in user in the [ZenDesk web widget](https://developer.zendesk.com/embeddables/docs/widget/api). A significant amount of the work in this pull request is in serializing the User object, and testing that serialization. I was running into import errors when I tried to keep the tests in the `profiles` module, so I ended up moving the serialization machinery, as well as the tests for it, into the `micromasters` module.

Note that this results in some duplication on the `SETTINGS` object. We can clean that up in a later pull request -- changing all the existing references would make this pull request even larger.

Open the ZenDesk widget while logged out, and verify that it is not populated with information. Then log in, and open the ZenDesk widget again, and verify that the "name" and "email" fields are pre-populated. You can also view the source of the page, and verify that the user object is getting serialized to the `SETTINGS` object in Javascript.


micromasters/serializers.py
```diff
@@ -0,0 +1,78 @@
...
+    first_name = serializers.SerializerMethodField()
+    last_name = serializers.SerializerMethodField()
+    preferred_name = serializers.SerializerMethodField()
+    is_staff = serializers.SerializerMethodField()
+    social_username = serializers.SerializerMethodField()
+
+    class Meta:
+        model = User
+        fields = (
+            "username", "email", "social_username",
```
noisecapella: Can we leave out `username` here and rename `social_username` to `username`? The only username the JS client should know about is the social one which is the one from edX. Often `User.username` will be the same but it's not guaranteed, especially if there was a backend error when logging in

micromasters/serializers_test.py
```diff
@@ -0,0 +1,50 @@
...
+from django.db.models.signals import post_save
+from factory.django import mute_signals
+from micromasters.serializers import UserSerializer
+from micromasters.factories import UserFactory
+from profiles.factories import ProfileFactory
+from search.base import ESTestCase
+# pylint: disable=no-self-use
+
+
+class UserTests(ESTestCase):
```
noisecapella: Can you add tests for `to_internal_value`?

micromasters/serializers_test.py
```diff
@@ -0,0 +1,50 @@
...
+    "User tests"
```
noisecapella: All docstrings should be surrounded by three double quotes: https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring

noisecapella: Can you also do this for the functions in this file?

micromasters/serializers.py
```diff
@@ -0,0 +1,78 @@
...
+            "first_name", "last_name", "preferred_name",
+            "is_staff",
+        )
+
+    def get_first_name(self, obj):  # pylint: disable=no-self-use
+        "Get first_name from user profile"
+        if obj.profile:
```
noisecapella: If this fails a `DoesNotExist` exception is thrown, so the `if` expression will never be false. Can you change this and the ones below to catch the `DoesNotExist` instead?

micromasters/serializers.py
```diff
@@ -0,0 +1,78 @@
...
+            return obj.profile.last_name
+        return None
+
+    def get_preferred_name(self, obj):  # pylint: disable=no-self-use
+        "Get preferred_name from user profile"
+        if obj.profile:
+            return obj.profile.preferred_name
+        return None
+
+    def get_is_staff(self, obj):  # pylint: disable=no-self-use
```
noisecapella: We shouldn't conflate staff and instructor here. Can you change this to return the roles assigned to the user instead, or remove it since it's not being used? We currently output roles information to `SETTINGS` here, thought this isn't done on the program pages yet: https://github.com/mitodl/micromasters/blob/master/ui/views.py#L53

noisecapella:
Functionality looks good

micromasters/serializers.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        if obj.is_anonymous():
+            return None
+        return super().to_representation(obj)
+
+    def to_internal_value(self, data):
+        """
+        Recognize that `None` means an anonymous user.
+        """
+        if not data:
```
noisecapella: Just to be explicit, can you make this `data is None` instead?

micromasters/serializers_test.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Tests for serializing anonymous users.
+    """
+    def test_serialize_anonymous_user(self):
+        anon_user = AnonymousUser()
+        result = UserSerializer().to_representation(anon_user)
+        assert result is None
+
+    def test_deserialize_anonymous_user(self):
+        result = UserSerializer().to_internal_value(None)
```
singingwolfboy: I spent a fair amount of time researching and trying to figure out the best way to do this, and I'm not sure if I can. Turns out we're not really using DRF's API correctly: the canonical way of serializing an instance to data is:
```python
instance = MyModel()
serializer = MySerializer(instance)
# access `.data` attribute to get the serialized data from the instance
serializer.data
```
And the canonical way of deserializing data is:
```python
data = {"foo": "bar"}
serializer = MySerializer(data=data)
# call `.save()` to turn the data into a model instance and insert/update it in the DB
instance = serializer.save()
```

The `to_representation()` and `to_internal_value()` methods are meant to be overridden when necessary, but never called directly. In addition, when deserializing data, DRF always expects a dictionary -- if I do `UserSerializer(data=None)`, all the built-in validation functions go haywire.

So, I'm not sure what the best course of action is here. I don't think we have any real need to deserialize user objects to Python right now, so we could just take out the `to_internal_value()` method and merge this as-is, perhaps with a comment indicating that this class doesn't support deserializing `None` into an anonymous user. We may want to do a clean-up in the future to use DRF more the way it's meant to be used, but that *definitely* doesn't belong in this pull request.

Thoughts?

noisecapella: Ah, I didn't know that. I feel like taking out `to_internal_value()` is the best course here, then a refactor later on

singingwolfboy:
@noisecapella can you do one final review of this before I merge it?

PR #1483, opened 2016-10-20
Added ability to mark order as refunded

### LINKED ISSUE #1420: ecommerce: Add `refunded` state to order table
We need a state for when the learner has had their purchase refunded. 

When the state of the order is `refunded` the dashboard should behave as if the user had never purchased the course.

noisecapella:
Fixes #1420 

#### What's this PR do?
Adds a refunded state to the Order and provides a way to let the user change the order status in the Django Admin.

Change an `Order` status via the Django admin, and verify that the change is reflected in the `OrderAudit` table


giocalitri:
👍

PR #1488, opened 2016-10-20
Protected audit tables

### LINKED ISSUE #1450: Protect audit tables in postgres
We should make a migration to protect audit tables by disabling update and delete. We can add a rule like this:

    CREATE RULE shoe_ins_protect AS ON INSERT TO shoe
        DO INSTEAD NOTHING;
    CREATE RULE shoe_upd_protect AS ON UPDATE TO shoe
        DO INSTEAD NOTHING;

From https://www.postgresql.org/docs/9.4/static/rules-update.html

noisecapella:
Fixes #1450 

#### What's this PR do?
Adds postgres rules for `OrderAudit`, `FinancialAidAudit`, and `FinancialAidEmailAudit` to prevent `UPDATE` or `DELETE` of the table.

You should be able to run the migrations and reverse the migrations successfully. Any attempt to edit and save an existing instance of one of the audit models will cause an exception. For example, try saving an `OrderAudit` instance in the Django Admin. If you reverse the migration, you should be able to save the `OrderAudit` instance again without error.

### DIFF
--- a/ecommerce/migrations/0008_protect_audit.py
+++ b/ecommerce/migrations/0008_protect_audit.py
@@ -9,11 +9,11 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunSQL(
-            "CREATE RULE delete_protect AS ON DELETE TO ecommerce_orderaudit DO INSTEAD NOTHING",
-            "DROP RULE delete_protect ON ecommerce_orderaudit",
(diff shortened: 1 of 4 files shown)


giocalitri:
can you please add kwargs for each `RunSQL`? It is hard to understand that the first is the SQL for the forward migration and the second is for the backwards one.

noisecapella:
Done

giocalitri:
👍

PR #1491, opened 2016-10-20
Added check for missing migrations

### LINKED ISSUE #1461: Add management command to check configuration
This would check the cases where we currently raise `ImproperlyConfigured`

noisecapella:
Fixes #1461

#### What's this PR do?
Runs migrations with `--fake` enabled to see if there's a warning message about migrations not matching `models.py`. The build should fail if there is a migration missing

Make a minor change to some field in a model, then run `./detect_missing_migrations.sh` in Docker. Run `echo $?` to see the return value of the last command. You should get `1`. Revert your change to the field, then run `./detect_missing_migrations.sh` again, then run `echo $?`. You should get `0` now.


detect_missing_migrations.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e -o pipefail
+
+if ./manage.py migrate --fake | grep "Your models have changes that are not yet reflected" > /dev/null
+then
+    echo "Error: one or more migrations are missing:"
+    echo
+    ./manage.py migrate --fake
+    exit 1
+fi
```
giocalitri: nit: missing empty line at end of file

giocalitri:
given that I only run `tox` locally, I wonder if there is aa way to call this script directly in the `tox.ini` and not in the `.travis.yml`

if there is not an easy solution, just leave as is

noisecapella:
Comments addressed

detect_missing_migrations.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e -o pipefail
+
+if ./manage.py migrate --fake | grep "Your models have changes that are not yet reflected" > /dev/null
+then
+    echo "Error: one or more migrations are missing:"
+    echo
+    ./manage.py migrate --fake
```
giocalitri: I do not want to make this a big deal, but it would be nice to run this command only once before, save the output to a variable and than only print here.
I also understand that the case when there are migration to create is the minority, so if you think it's not worth the time, feel free to merge as is.

giocalitri:
One more (optional) comment, otherwise 👍

detect_missing_migrations.sh
noisecapella: I'll leave it as is just to keep the script simple

PR #1492, opened 2016-10-20
Refactored course tests

noisecapella:
None

#### What's this PR do?
Refactors the tests in `courses/models_test.py` to use `ddt` for parameterized testing

Look at the code and make sure the same cases are being covered as before.

### DIFF
--- a/courses/models_test.py
+++ b/courses/models_test.py
@@ -4,6 +4,7 @@ Model tests
 # pylint: disable=no-self-use
 from datetime import datetime, timedelta
 
+from ddt import ddt, data, unpack
 import pytz
 
 from courses.factories import (
@@ -22,6 +23,15 @@ class ProgramTests(ESTestCase):
         """Test for __str__ method"""
         prog = ProgramFactory.build(title="Title")
         assert "{}".format(prog) == "Title"
+
+
+def from_weeks(weeks, now=None):
+    """Helper function to get a date adjusted by a number of weeks"""
+    if weeks is None:
+        return None
+    if now is None:
+        now = datetime.now(tz=pytz.UTC)
+    return now + timedelta(weeks=weeks)
 
 
 class CourseModelTests(ESTestCase):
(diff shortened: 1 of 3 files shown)


giocalitri:
Question before even starting the review: I recall a conversation that we had about pytest having some built in functionality to do what `ddt` does. Am I wrong?

noisecapella:
It doesn't work with our test cases which extend `TestCase`, it only works with function tests: http://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests

giocalitri:
>It doesn't work with our test cases which extend TestCase, it only works with function tests: http://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests

OK: will proceed with review

courses/models_test.py
```diff
@@ -24,6 +25,15 @@ def test_to_string(self):
...
 
+def from_weeks(weeks, now=None):
+    """Helper function to get a date adjusted by a number of weeks"""
+    if weeks is None:
+        return None
+    if now is None:
+        now = datetime.now(tz=pytz.UTC)
+    return now + timedelta(weeks=weeks)
+
+
```
giocalitri: why did you define this function here and not in the `CourseModelTests` base class?

courses/models_test.py
```diff
@@ -60,120 +77,51 @@ def test_no_run(self):
...
+        [-52, -45, -62, -53, False],
+        [-1, 2, -10, -3, False],
+        [1, 3, -10, -3, False],
+        [-10, -3, None, None, False],
+        [-52, None, None, None, True],
+        [1, 4, -10, 1, True],
+        [-1, 2, -10, 1, True],
+        [-1, 2, -10, None, True],
+        [52, 62, 40, 50, True],
+        [1, 4, None, None, True],
```
giocalitri: I definitively need this otherwise it is way too complicated to match the old tests with these ddt lists

courses/models_test.py
noisecapella: I wanted it to be usable to make data for the parameterized tests, and I figured it was general enough to not be tied to a particular test

noisecapella:
I added comments before each set of parameters

giocalitri:
there are some conflicts: once resolved 👍

Issue #1468, opened 2016-10-19
Break up financial aid tests


noisecapella:
There are some tests which should be broken into smaller tests, for example `test_determine_auto_approval` in `financialaid/api_test.py`

pdpinch:
Is this done, now that #1495 has merged?

noisecapella:
Yes, I'll close this

PR #1495, opened 2016-10-21
Refactored financial aid tests

noisecapella:
Part of #1468 

#### What's this PR do?
Uses ddt to split financial aid tests


singingwolfboy:
Rather than defining all of these pull requests using relative time (X number of weeks in the future from when this test run is executed), wouldn't it be better to use a library like [freezegun](https://github.com/spulec/freezegun) to mock the datetime module and define tests with absolute timestamps?

noisecapella:
I'm not sure it's useful in this case, and the relative time makes it a little more intuitive what is being tested here

annagav:
Looks good 👍

Issue #1367, opened 2016-10-13
Search: omit staff from learner profile search results


pdpinch:
Currently, search results can return staff as well as learners in the learner profile search. I don't think we want to do this. 

<img width="985" alt="screen shot 2016-10-13 at 12 07 51 pm" src="https://cloud.githubusercontent.com/assets/430126/19357004/b65064ce-913d-11e6-8ac8-bf294f1d82df.png">

PR #1502, opened 2016-10-23
Omitted staff from search results

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/1367

#### What's this PR do?
- Omits staff from search results on learners page

- see /search/api.py

- go to /learners/ page and see if staff of program is in list or not

@pdpinch @giocalitri @noisecapella 
screen shot 2016-10-23 at 11 48 49 pm

### DIFF
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -4,10 +4,7 @@ Provides functionality for serializing a ProgramEnrollment for the ES index
 from decimal import Decimal
 
 from rolepermissions.verifications import has_role
-from roles.models import (
-    Instructor,
-    Staff,
-)
+from roles.models import NON_LEARNERS
 
 
 class UserProgramSerializer:
@@ -54,5 +51,5 @@ class UserProgramSerializer:
             'enrollments': program_cached_enrollments,
             'certificates': program_cached_certificates,
             'grade_average': cls.calculate_certificate_grade_average(program_cached_certificates),
-            'is_staff': has_role(user, [Staff.ROLE_ID, Instructor.ROLE_ID])
+            'non_learner': has_role(user, NON_LEARNERS)
         }
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -3,8 +3,10 @@ Provides functionality for serializing a ProgramEnrollment for the ES index
 """
 from decimal import Decimal
 
-from rolepermissions.verifications import has_role
-from roles.models import NON_LEARNERS
+from roles.models import (
+    NON_LEARNERS,
+    Role
+)
 
 
 class UserProgramSerializer:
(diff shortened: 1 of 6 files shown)


dashboard/serializers.py
```diff
@@ -47,5 +53,6 @@ def serialize(cls, program_enrollment):
             'id': program.id,
             'enrollments': program_cached_enrollments,
             'certificates': program_cached_certificates,
-            'grade_average': cls.calculate_certificate_grade_average(program_cached_certificates)
+            'grade_average': cls.calculate_certificate_grade_average(program_cached_certificates),
+            'is_staff': has_role(user, [Staff.ROLE_ID, Instructor.ROLE_ID])
```
giocalitri: I think that you should find a way to define the roles that are not `learners` directly in the `role.models` and just list them here.
The reason is because we do not want forget to add here a new role if we create one.

search/api_test.py
```diff
@@ -82,6 +82,9 @@ def test_create_search_obj_program_limit(self):
             'bool',
             should=[
                 Q('term', **{'program.id': self.program.id})
+            ],
+            must_not=[
+                Q('term', **{'program.is_staff': True})
```
giocalitri: this is not enough.
you need to add another user that is not staff and make sure that that user is in the search result and the user that is `staff is not`

amir-qayyum-khan:
@noisecapella done

dashboard/serializers.py
```diff
@@ -47,5 +50,6 @@ def serialize(cls, program_enrollment):
             'id': program.id,
             'enrollments': program_cached_enrollments,
             'certificates': program_cached_certificates,
-            'grade_average': cls.calculate_certificate_grade_average(program_cached_certificates)
+            'grade_average': cls.calculate_certificate_grade_average(program_cached_certificates),
+            'non_learner': has_role(user, NON_LEARNERS)
```
giocalitri: this is not good:
our roles are per Programs while this does not verify that the user might not be `Staff` in this program while she is in another one.
I think that the right way to verify this is to create a utility function that checks something like
```
has_role(user, NON_LEARNERS) and Role.objects.filter(user=user, role__in= NON_LEARNERS, program=program).exists()
```

giocalitri: You will need also to write another test to cover the case that the user is not Staff of the program and in this case it appears in the search result

roles/models.py
```diff
@@ -10,6 +10,12 @@
 from roles.roles import Staff, Instructor
 
 
+NON_LEARNERS = [
```
giocalitri: please add a comment before this to say that this list needs to be updated in case we add new roles

search/indexing_api.py
```diff
@@ -248,6 +248,7 @@ def program_enrolled_user_mapping():
     mapping.field("program", "object", properties={
         'id': LONG_TYPE,
         'grade_average': LONG_TYPE,
+        'non_learner': BOOL_TYPE,
```
giocalitri: why don't you call this `is_learner` and reverse the logic so that you include in the result only learners instead of excluding the other roles?

dashboard/serializers.py
amir-qayyum-khan: i think we should only omit staff of selected program from search,
not the staff of other programs. 
They can be learner on selected program? what are your thoughts.

cc @pdpinch

giocalitri:
another additional test then we are ready to merge

amir-qayyum-khan:
Done @giocalitri

giocalitri:
👍 when tests pass

Issue #1394, opened 2016-10-14
In Profile Flow, MicroMasters link in header is confusing


j2tausig:
If a user hasn't completed their profile and the click the MicroMasters link in the header, that triggers validation alerts on the page.
Per Peter, this is the intended behavior.  In this case, we should remove the MicroMasters link.

Steps to Reproduce:
Create an account in edX.
Don't complete the profile info.
Select a program.
Click on Login.
After logging in, click on the MicroMasters link in the header.
**note** it's not the regular link -- only text is displayed, we've lost the MicroMasters Logo.

First Screen shot shows the incomplete profile:
![profile path - initial](https://cloud.githubusercontent.com/assets/13886722/19392291/331a4072-91fe-11e6-80de-6de0837cb4ac.PNG)

Second Screen shot shows the display when a user clicks on the MicroMasters link in the header:
![profile path - post header link](https://cloud.githubusercontent.com/assets/13886722/19392334/609ea538-91fe-11e6-83be-91ef420a73a2.PNG)

Ferdi:
So, leave it as text (remove the link) from profile pages.

PR #1505, opened 2016-10-24
Removed dashboard links from profile page header

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1394

- open profile page ``/profile`` click MM logo or test, you will not navigate
- open settings page, user page and then click on same on header, you will be navigated dashboard or learners page respectively.
 
@pdpinch @aliceriot 

screen shot 2016-10-24 at 5 15 48 pm

### DIFF
--- a/static/js/components/Navbar.js
+++ b/static/js/components/Navbar.js
@@ -20,6 +20,7 @@ import { getPreferredName } from '../util/util';
 import type { Profile } from '../flow/profileTypes';
 
 const PROFILE_SETTINGS_REGEX = /^\/profile\/?|settings\/?|learner\/[a-z]?/;
+const PROFILE_REGEX = /^\/profile\/?/;
 
 export default class Navbar extends React.Component {
   props: {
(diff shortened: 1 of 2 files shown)


aliceriot:
functionality looks good!

one thing though, the appearance is different on `/profile`:

profile_logo

than it is on `/dashboard` and elsewhere on the site:

dashboard_logo

aliceriot:
on `master` it looks the same in both places.

aliceriot:
other than that functionality is good! taking a look at the code now.

static/js/components/Navbar.js
```diff
@@ -13,6 +13,7 @@ import ProgramSelector from './ProgramSelector';
 import UserMenu from '../containers/UserMenu';
 
 const PROFILE_SETTINGS_REGEX = /^\/profile\/?|settings\/?|learner\/[a-z]?/;
+const PROFILE_PAGE_REGEX = /^\/profile\/[a-z]?/;
```
aliceriot: what is `[a-z]` for in this regex? I think the only url we care about is `/profile`, right?

aliceriot: I think `PROFILE_REGEX` might be a slightly better name for this as well

static/js/components/Navbar.js
```diff
@@ -37,6 +38,36 @@ export default class Navbar extends React.Component {
...
+        </span>
+      );
+    }
+
+    return (
+      <span className="mdl-layout-title">
+        MicroMasters
+      </span>
+    );
+  }
```
aliceriot: Why not have the header text and the image each be rendered by the same method, with a different method for each case?

Then you'd have something like:

```js
renderProfileHeader = () => ([
  <span />,
  <img />
]);

renderNormalHeader = () => ([
  <span />,
  <img />
]);
```

then in the JSX for the `render` method you'd just have something like:

```js
{ PROFILE_REGEX.test(pathname) ? renderProfileHeader() : renderNormalHeader() }
``` 

What do you think?
 
I like this kind of approach because the methods are simpler and therefore easier to refactor.

amir-qayyum-khan: sure, let me do this

amir-qayyum-khan:
Done @aliceriot

PR #1522, opened 2016-10-24
Add Smartlook tracking

### LINKED ISSUE #1474: Add smartlook usability tracking app
We'd like to try out this service, smartlook.com, which can be used to record user interaction with the web application. Apparently, we just need to add the following javascript to activate it:

```
    window.smartlook||(function(d) {
    var o=smartlook=function(){ o.api.push(arguments)},h=d.getElementsByTagName('head')[0];
    var c=d.createElement('script');o.api=new Array();c.async=true;c.type='text/javascript';
    c.charset='utf-8';c.src='//rec.smartlook.com/recorder.js';h.appendChild(c);
    })(document);
    smartlook('init', '78d6283937e3e0ae259e686a4ffff5d2c7f76a00');
</script>
```

Presumably the `78d6283937e3e0ae259e686a4ffff5d2c7f76a00` is an account key of some sort. It would be nice for this to be managed via an environment setting. If it isn't set, the javascript shouldn't be included at all.

annagav:
Fix #1474
#### What's this PR do?
Add SmartLook tracking.
For testing,  set an environment variable `SL_TRACKING_ID`. Then check in the header that the code for smartlook is set with the correct value.

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -215,6 +215,7 @@ TEMPLATES = [
                 'social.apps.django_app.context_processors.backends',
                 'social.apps.django_app.context_processors.login_redirect',
                 'ui.context_processors.google_analytics',
+                'ui.context_processors.smartlook',
             ],
         },
     },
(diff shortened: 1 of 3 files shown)


pdpinch:
When this merges, don't forget to set the environment variable for SL_TRACKING_ID on micromasters-ci

PR #1532, opened 2016-10-24
Logged exception being handled in custom_exception_handler

noisecapella:
None
#### What's this PR do?
Logs the `ImproperlyConfigured` exception which is raised sometimes

In `courses/serializers.py` add a field name to the `fields` list which obviously doesn't exist. Go to `/api/v0/programs/`. You should see the `ImproperlyConfigured` exception response. Look in your console output and you should see an exception about a field name along with an accurate stack trace.

### DIFF
--- a/micromasters/utils.py
+++ b/micromasters/utils.py
@@ -2,6 +2,7 @@
 General micromasters utility functions
 """
 import json
+import logging
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
@@ -9,6 +10,9 @@ from django.core.serializers import serialize
 from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.views import exception_handler
+
+
+log = logging.getLogger(__name__)
 
 
 def webpack_dev_server_host(request):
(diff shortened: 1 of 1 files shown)


singingwolfboy:
I tested this pull request as described in the description, but I'm not seeing any exception logging or tracebacks in my console. The only logs I see in the console appear to be SQL statements being executed. @noisecapella, is there something special I need to do to display more logging information?

noisecapella:
docker-compose is finicky with stdout and the logging output. @giocalitri do you have any experience working around that?

giocalitri:
yes, I usually run (on Mac) the django app separately with a different command to have better control over logging.

So I run:
```
docker-compose up redis elastic db
```

and in another shell 
```
docker-compose run --service-ports web ./manage.py runserver 0.0.0.0:8079
```

in my `.env` I have this entries:

```
MICROMASTERS_LOG_LEVEL=DEBUG
DJANGO_LOG_LEVEL=WARNING
```

I hope this helps

singingwolfboy:
I tried @giocalitri's suggestions, and I'm still not seeing any exception logging or tracebacks, either when I've modified `courses/serializers.py` or when I leave it as it is in the repo. Any other ideas?

singingwolfboy:
@noisecapella and I verified this on the Heroku pull-request deployment. It would be nice to figure out how to verify this locally, but that's another problem for another day.

PR #1544, opened 2016-10-25
Use bulk indexing for seed_db

noisecapella:
None

#### What's this PR do?
Mutes the signals which index the profiles as they are created. Instead they are bulk indexed via `recreate_index` after all profiles are created.

Run `./manage.py seed_db --staff-user staff` and verify that you can see the profiles which are created in the learner search at `/learners`.

The `mute_signals` will also prevent profiles from being created automatically with the `User`, so there's a new `Profile.objects.create(...)` added to fix this.

### DIFF
--- a/seed_data/management/commands/seed_db.py
+++ b/seed_data/management/commands/seed_db.py
@@ -202,6 +202,7 @@ def deserialize_user_data_list(user_data_list, course_runs):
     for user_data in user_data_list:
         new_user = deserialize_user_data(user_data, course_runs)
         new_user_count += 1
+        # This function is run with mute_signals(post_save) so we need to create the profile explicitly.
         profile = Profile.objects.create(user=new_user)
         deserialize_model_data_on_object(profile, user_data)
         deserialize_profile_detail_data(profile, Employment, user_data['work_history'])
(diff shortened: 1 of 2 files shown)


noisecapella:
@gsidebo What do you think of this solution?

seed_data/management/commands/seed_db.py
```diff
@@ -200,6 +202,7 @@ def deserialize_user_data_list(user_data_list, course_runs):
     for user_data in user_data_list:
         new_user = deserialize_user_data(user_data, course_runs)
         new_user_count += 1
+        Profile.objects.create(user=new_user)
         deserialize_model_data_on_object(new_user.profile, user_data)
```
singingwolfboy: Nitpick: it would be clearer to save the result of `Profile.objects.create(user=new_user)` into a variable called `new_profile`, and refer to that on the subsequent lines, rather than referring to `new_user.profile` on those lines.

noisecapella: Sure

noisecapella:
Done, I also updated `unseed_db` to do the same bulk indexing on delete to save time there

noisecapella:
Any other comments?

gsidebo:
looks great 👍

PR #1545, opened 2016-10-25
Added Django Debug Toolbar in Debug mode

giocalitri:

N/A
#### What's this PR do?

Adds Django Debug Toolbar

rebuild the web container then verify that the toolbar works

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -498,3 +498,37 @@ CYBERSOURCE_REFERENCE_PREFIX = get_var("CYBERSOURCE_REFERENCE_PREFIX", None)
 # Open Exchange Rates
 OPEN_EXCHANGE_RATES_URL = get_var("OPEN_EXCHANGE_RATES_URL", None)
 OPEN_EXCHANGE_RATES_APP_ID = get_var("OPEN_EXCHANGE_RATES_APP_ID", "")
+
+# django debug toolbar only in debug mode
+if DEBUG:
+    INSTALLED_APPS += ('debug_toolbar', )
+    # it needs to be enabled before other middlewares
+    MIDDLEWARE_CLASSES = (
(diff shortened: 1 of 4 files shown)


noisecapella:
When I open the tab, go to SQL and click 'sel' for example I get an error, just a 403 Forbidden. In the logs I see 

    web_1      | [2016-10-25 18:27:43] WARNING 20 [django.request] csrf.py:103 - [d8846d3d8903] - Forbidden (CSRF token missing or incorrect.): /__debug__/sql_explain/

giocalitri:
I am struggling to understand why it is not working:

screen shot 2016-10-25 at 3 35 23 pm

any suggestion?

giocalitri:
actually it seems that that token is incorrect

noisecapella:
Could just be a bug on their end. Anyway I'm fine merging if you want, it's not essential functionality and everything else appears to work

giocalitri:
I actually do not like it because in edX on master it works just fine...
I would like to make more investigations

giocalitri:
@noisecapella you were right: the problem was with the URL import sequence.
Now it is fixed

micromasters/settings.py
```diff
@@ -498,3 +498,26 @@ def get_var(name, default):
...
+    INSTALLED_APPS += ('debug_toolbar', )
+    # it needs to be enabled before other middlewares
+    MIDDLEWARE_CLASSES = (
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+    ) + MIDDLEWARE_CLASSES
+
+    def show_toolbar(request):
+        """
+        Custom function needed because of bug in wagtail.
+        Theoretically this bug has been fixed in django 1.10 and wagtail 1.6.3
```
giocalitri: I think we need to check if this is still an issue after we upgrade to django 1.10

PR #1546, opened 2016-10-25
Show only published children pages on the ProgramPage

### LINKED ISSUE #1543: Program Page: Tabs should honor published setting of child pages
Currently, when you unpublish a child of the Program (like a FAQ page) in the CMS, its tab remains in the Program page. Clicking on the tab results in a 404.

annagav:
Fix  #1543

On the program page show only published child pages.
To test this, unpublish an existing faqs page, and check if it has disappeared from the the program page.

### DIFF
--- a/cms/templates/cms/faqs_page.html
+++ b/cms/templates/cms/faqs_page.html
@@ -6,7 +6,7 @@
 
 {% block panel %}
   <div class="faqs">
-    {% for categorized in child_page.get_children.specific %}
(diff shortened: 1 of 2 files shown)


noisecapella:
I think we need to do something similar for each FAQ group since it looks like you can unpublish those individually

annagav:
I did the same for `CategorizedFaqsPage`.

PR #1548, opened 2016-10-25
Fixed course description JS bug

gsidebo:

No associated issue

#### What's this PR do?

Fixes the CourseAction bug that was breaking the dashboard

Load the dashboard, preferably with a course that has course run with a `null` value for all date fields

### DIFF
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -103,12 +103,12 @@ export default class CourseAction extends React.Component {
   }
 
   renderDescription = R.curry(
-    (className: string, runStatus: ?string, text: string): React$Element<*>|null => {
+    (className: string, runStatus: ?string, text: ?string): React$Element<*>|null => {
       let classDefinition = className;
       if (runStatus && this.statusDescriptionClasses[runStatus]) {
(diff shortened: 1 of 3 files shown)


gsidebo:
@roberthouse54 this shows a course description 'Enrollment information unavailable' when an offered course run has null values for start date _and_ fuzzy start date. There was no design for that case, and it's very unlikely to happen. Let me know if you're ok with that change

aliceriot:
functionality looks good!

not_available

static/js/components/dashboard/CourseAction.js
```diff
@@ -103,12 +103,12 @@ export default class CourseAction extends React.Component {
   }
 
   renderDescription = R.curry(
-    (className: string, runStatus: ?string, text: string): React$Element<*>|null => {
+    (className: string, runStatus: ?string, text: ?string): React$Element<*>|null => {
```
aliceriot: It looks like Flow doesn't know how to properly check the types of functions that are wrapped with `R.curry`, I was able to change `className: string` to `className: number` without any errors. So, I think it's good to document types for readability, but keep in mind that Flow won't catch errors here :(

Maybe they're working on stuff like that! I hope so.

Issue #1550, opened 2016-10-25
Purchases are failing on micromasters-rc due to enrollment failure


noisecapella:
Found this when trying to check off my boxes for the release. Here's the response when the enrollment:

```
{
    "enrollment_detail": {
        "course_details": {
            "course_end": null,
            "course_id": "ccx-v1:MITx+Giovanni101+2015_T3+ccx@38",
            "course_modes": [
                {
                    "bulk_sku": null,
                    "currency": "usd",
                    "description": null,
                    "expiration_datetime": null,
                    "min_price": 0,
                    "name": "Audit",
                    "sku": null,
                    "slug": "audit",
                    "suggested_prices": ""
                }
            ],
            "course_start": "2016-03-04T12:02:25Z",
            "enrollment_end": null,
            "enrollment_start": "2016-01-01T00:00:00Z",
            "invite_only": false
        },
        "created": "2016-04-04T14:22:48Z",
        "is_active": true,
        "mode": "audit",
        "user": "gschneel"
    },
    "message": "An enrollment already exists for user [gschneel] in course run [ccx-v1:MITx+Giovanni101+2015_T3+ccx@38]."
}
```

On my local machine:

```
{
    "course_details": {
        "course_end": null,
        "course_id": "course-v1:MIT+GEO102+GEO_T1",
        "course_modes": [
            {
                "bulk_sku": null,
                "currency": "usd",
                "description": null,
                "expiration_datetime": null,
                "min_price": 0,
                "name": "Audit",
                "sku": null,
                "slug": "audit",
                "suggested_prices": ""
            }
        ],
        "course_start": "2030-01-01T00:00:00Z",
        "enrollment_end": null,
        "enrollment_start": null,
        "invite_only": false
    },
    "created": "2016-06-30T21:50:20.225984Z",
    "is_active": true,
    "mode": "audit",
    "user": "staff"
}
```

pdpinch:
This is an upstream issue at edX that was caused by a change to the enrollment API. The commit at https://github.com/edx/edx-platform/commit/21746f1e93f24244d7dad799aef5c7265cf59e65 changed the response from the enrollment API and broke our edx-api client. 

In the current code, users will not be able to "enroll and pay later" and if they do pay, they can complete the transaction, but the portal will not recognize it. Possible errors include a [DoesNotExist execption](https://sentry.io/mit-office-of-digital-learning/micromasters/issues/174176463/)

[...]

PR #1552, opened 2016-10-25
Prevented an enrollment failure from failing the order

noisecapella:
Related to #1550 

#### What's this PR do?
If the enrollment failed, the order would fail. This is changed to not fail the order but to just log the enrollment failure instead.

Purchase a course without having a course key which exists on the edX instance. It should go through successfully

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -290,7 +290,7 @@ def enroll_user_on_success(order):
         try:
             enrollments.append(enrollments_client.create_audit_student_enrollment(course_key))
         except Exception as ex:  # pylint: disable=broad-except
-            log.error(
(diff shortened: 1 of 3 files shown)


singingwolfboy:
Perhaps it would make more sense to create a new state to record this situation. If the state is "Fulfilled", that indicates that there's nothing more we need to do on our end, which is not true. Perhaps call the new state "Paid"?

noisecapella:
I'm not sure a new state is necessary since this information is stored in edX. We have it available locally in the `CachedEnrollment` model and we currently use both source of information to determine course states in `MMTrack`

giocalitri:
I think this can be merged 👍

PR #1554, opened 2016-10-25
fixed layout bug with footer Give to MIT buttons

roberthouse54:
#### What's this PR do?

This is a quick fix of a layout (margin) problem with the footers.


noisecapella:
The 'give' button looks a bit odd to me, can't put a finger on it though

PR #1565, opened 2016-10-26
Make modals more consistent

### LINKED ISSUE #1452: Refactor all modal windows front-end code
We should refactor the front-end code for all the modals in the app so that:

- All modals have a consistent header/title element and class. This could be an H3 tag with a class "modal_title" for example.
- All modals have consistent buttons.
- All modals have classes added to the modal. There should be one shared class like "mm-modal", and then a class that is specific to the modal (ex. Sign-up-modal"). In part this is so that we can style the modals... for example, we cannot currently style the width of modals for mobile layouts.

singingwolfboy:
Fixes #1452. This pull request does the following things to make the modals in our application more consistent:

* Applies a "dialog" class to every modal window
* Ensures that every modal has a title
* Applies a "dialog-title" class to every modal title
* Unified button styles for all buttons in dashboard modals

However, not all modals are created equally. The "Sign Up" modal on the landing page does _not_ have a title. This is intentional, and was specifically requested by @roberthouse54.

Note that the `className` property on the `<Dialog>` element applies the class name to the wrapper div around the modal window, which is used as an overlay for the entire page to catch people clicking outside of the modal to clear it. The `contentClassName` property applies the class name to the div containing the actual modal window itself, which is what we want to be able to style.

### DIFF
--- a/static/js/components/ConfirmDeletion.js
+++ b/static/js/components/ConfirmDeletion.js
@@ -23,14 +23,14 @@ export default class ConfirmDeletion extends React.Component {
       <Button
         type='button'
         key='close'
-        className="cancel-button"
+        className="secondary-button cancel-button"
         onClick={close}>
         No
       </Button>,
       <Button
         key='delete'
         type='button'
-        className="delete-button"
+        className="primary-button delete-button"
         onClick={this.deleteAndClose}>
         Yes
       </Button>
--- a/static/js/components/ConfirmDeletion.js
+++ b/static/js/components/ConfirmDeletion.js
@@ -37,7 +37,9 @@ export default class ConfirmDeletion extends React.Component {
     ];
     return (
       <Dialog
-        className={"deletion-confirmation"}
+        title="Confirm Delete"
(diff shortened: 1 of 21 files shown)


roberthouse54:
Looks good visually. Maybe someone can do a code review?

static/js/components/ConfirmDeletion.js
```diff
@@ -18,32 +20,34 @@ export default class ConfirmDeletion extends React.Component {
...
+        Delete
       </Button>
     ];
     return (
       <Dialog
-        className={"deletion-confirmation"}
+        title={dialogTitle(itemText)}
+        className="deletion-confirmation-dialog-wrapper"
+        titleClassName="dialog-title"
+        contentClassName="dialog deletion-confirmation-dialog"
```
noisecapella: I can't find `deletion-confirmation-dialog` anywhere. Is it left here for consistency with the other `Dialog` components?

singingwolfboy: Yes, just for consistency.

PR #1569, opened 2016-10-26
Refactored dashboard API

### LINKED ISSUE #1195: Filter dashboard by program enrollments
Related to #1193. After MVP is released we should refactor our code to use `/api/v0/programs/` for the list of available programs and only show programs the user has enrolled in in `/api/v0/dashboard/`

noisecapella:

Fixes #1195

#### What's this PR do?
- Makes `/api/v0/dashboard/` only show programs the user is enrolled in
- Adds programpage url to `/api/v0/programs/`
- Changes JS code to read `/api/v0/programs/` and to use those programs to list program enrollments

Everything should look the same on the front end. In particular check that you can enroll in the profile page and in the dashboard.

### DIFF
--- a/courses/views_test.py
+++ b/courses/views_test.py
@@ -33,8 +33,9 @@ class ProgramTests(ESTestCase):
         resp = self.client.get(reverse('program-list'))
 
         assert len(resp.json()) == 1
-        expected = ProgramSerializer(context={"request": Mock(user=self.user)}).to_representation(prog)
-        assert [expected] == resp.json()
+        context = {"request": Mock(user=self.user)}
+        data = ProgramSerializer(prog, context=context).data
+        assert [data] == resp.json()
 
     def test_doesnt_list_unlive_programs(self):
         """Not-live programs should NOT show up"""
--- a/courses/views_test.py
+++ b/courses/views_test.py
@@ -1,11 +1,14 @@
 """Tests for the API"""
 # pylint: disable=no-self-use
 
+from mock import Mock
+
 from django.core.urlresolvers import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
 
 from courses.factories import ProgramFactory
+from courses.serializers import ProgramSerializer
 from dashboard.factories import ProgramEnrollmentFactory
 from dashboard.models import ProgramEnrollment
 from micromasters.factories import UserFactory
@@ -30,7 +33,8 @@ class ProgramTests(ESTestCase):
         resp = self.client.get(reverse('program-list'))
 
         assert len(resp.json()) == 1
-        assert prog.title == resp.json()[0]['title']
+        expected = ProgramSerializer(context={"request": Mock(user=self.user)}).to_representation(prog)
+        assert [expected] == resp.json()
 
     def test_doesnt_list_unlive_programs(self):
         """Not-live programs should NOT show up"""
@@ -81,7 +85,7 @@ class ProgramEnrollmentTests(ESTestCase, APITestCase):
     def test_anonymous(self):
         """Anonymous user cannot access the endpoint"""
         self.client.logout()
-        resp = self.client.get(self.url)
+        resp = self.client.post(self.url)
         assert resp.status_code == status.HTTP_403_FORBIDDEN
 
     def test_no_enrollments(self):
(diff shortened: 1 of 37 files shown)


courses/views_test.py
```diff
@@ -30,7 +33,8 @@ def test_lists_live_programs(self):
         resp = self.client.get(reverse('program-list'))
 
         assert len(resp.json()) == 1
-        assert prog.title == resp.json()[0]['title']
+        expected = ProgramSerializer(context={"request": Mock(user=self.user)}).to_representation(prog)
```
singingwolfboy: We should avoid calling `.to_representation()` on a Serializer. The standard API is to pass the instance to the constructor, and access the `data` property. Like this:
```python
context = {"request": Mock(user=self.user)}
data = ProgramSerializer(prog, context=context).data
assert [data] == resp.json()
```

noisecapella: Will do. Also if you're reviewing this PR please assign yourself to it

gsidebo:
nitpick: if the dashboard API is giving different results now, that's a functionality change, so i wouldn't call it a 'refactor'.

static/js/components/ProgramSelector.js
```diff
@@ -51,24 +49,24 @@ export default class ProgramSelector extends React.Component {
...
 
     let currentId;
     if (!_.isNil(currentProgramEnrollment)) {
       currentId = currentProgramEnrollment.id;
     }
 
-    const sortedProgramEnrollments = _.sortBy(programEnrollments, 'title');
+    const sortedProgramEnrollments = _.sortBy(programEnrollments, 'title').filter(
+      enrollment => enrollment.enrolled
+    );
```
gsidebo: i think it would make a lot more sense to first sort the program enrollments, then create the lists for `sortedProgramEnrollments` and `unenrolledPrograms` (instead of doing the same operations on the same list in different ways). eg:

```javascript
const sortedPrograms = _.sortBy(programEnrollments, 'title');
let enrolledPrograms = programEnrollments.filter(program => program.enrolled);
let unenrolledPrograms = programEnrollments.filter(program => !program.enrolled);
```

static/js/components/PersonalTab.js
```diff
@@ -35,32 +34,32 @@ class PersonalTab extends React.Component {
...
     const menuItems = R.compose(makeMenuItems, sortPrograms);
-    return menuItems(programs) ;
-  }
+    return menuItems(programs);
+  };
 
   setProgramHelper = (event: Event, key: string, value: string) => {
     const {
-      dashboard: { programs },
+      programs: { programEnrollments },
```
gsidebo: i think there are a couple problems here.

1. `programs` is a `*State` type. correct me if i'm wrong, but the general pattern elsewhere in our codebase is to handle these states with their associated fetch status/error text/etc in a container before passing the raw information down to the components (at least for `*State` types that represent a request for a complex object or array and not just a scalar value). maybe i'm missing something, but we don't seem to be doing anything with the extra state information. it seems like only the list of `ProgramEnrollments` should be passed down to this and other components.
1. speaking of which, the naming of `programEnrollments` is pretty confusing. this line from ProgramSelector.js kinda sums it up: `enrollment => enrollment.enrolled`. in english: "is this enrollment enrolled?". i think just calling this collection `programs` is more on the mark.

static/js/components/ProgramSelector_test.js
```diff
@@ -78,7 +75,9 @@ describe('ProgramSelector', () => {
   });
 
   it("does not render the 'Enroll in a new program' option if there is not at least one available program", () => {
-    let allEnrollments = enrollments.concat(unenrolled);
+    let allEnrollments = enrollments.map(program => Object.assign({}, program, {
```
gsidebo: here's a line (along with line 58) that kinda illustrates my point about the naming. the collection is called `enrollments`, but the individual members are called `program`? i think the instinct is right here; the collection should just be called `programs`

static/js/components/NewEnrollmentDialog.js
```diff
@@ -68,15 +66,13 @@ export default class NewEnrollmentDialog extends React.Component {
...
-      dashboard: { programs },
       enrollDialogError,
       enrollDialogVisibility,
       enrollSelectedProgram,
       programs: { programEnrollments },
     } = this.props;
 
-    let enrollmentLookup = new Map(programEnrollments.map(enrollment => [enrollment.id, null]));
-    let unenrolledPrograms = programs.filter(program => !enrollmentLookup.has(program.id));
+    let unenrolledPrograms = programEnrollments.filter(program => !program.enrolled);
```
gsidebo: looks like you can make lines 75 & 76 into a one-liner

noisecapella:
Ok I think that addresses all comments

courses/views_test.py
```diff
@@ -81,40 +86,20 @@ def assert_program_enrollments_count(self, expected_count=None):
     def test_anonymous(self):
         """Anonymous user cannot access the endpoint"""
         self.client.logout()
-        resp = self.client.get(self.url)
+        resp = self.client.post(self.url)
         assert resp.status_code == status.HTTP_403_FORBIDDEN
 
     def test_no_enrollments(self):
         """Requests for a user with no program enrollments result in an empty list"""
```
gsidebo: looks like this comment needs to be changed as well

static/js/components/Navbar.js
```diff
@@ -27,12 +26,11 @@ export default class Navbar extends React.Component {
     addProgramEnrollment:        (programId: number) => void,
     children?:                   React$Element<*>[],
     currentProgramEnrollment:    ProgramEnrollment,
-    dashboard:                   DashboardState,
     empty:                       boolean,
     enrollDialogError:           ?string,
     enrollDialogVisibility:      boolean,
     enrollSelectedProgram:       ?number,
-    programs:                    ProgramEnrollmentsState,
+    programs:                    ProgramEnrollments,
```
gsidebo: what do you think of changing the variable names and type names from `ProgramEnrollments` (et al) to `AvailablePrograms`? we still have this awkwardness with 'Enrollments'. `programs: ProgramEnrollments`; `programs: PROGRAM_ENROLLMENTS`; etc. maybe `AvailablePrograms` isnt perfect, but it's accurate in that the list is supposed to represent every program available in Micromasters. 

this would change the `ProgramEnrollmentsState` like so:

```javascript
export type AvailableProgramsState = {
  availablePrograms: AvailablePrograms,
  getStatus?: string,
  getErrorInfo?: APIErrorInfo,
  postStatus?: string,
};
```

static/js/components/PersonalTab_test.js
```diff
@@ -0,0 +1,43 @@
...
+import { shallow } from 'enzyme';
+import { assert } from 'chai';
+import R from 'ramda';
+
+import PersonalTab from './PersonalTab';
+import { PROGRAM_ENROLLMENTS } from '../constants';
+
+describe("PersonalTab", () => {
+
+  let renderPersonalTab = (props = {}) => {
```
gsidebo: seems like you should just have this function accept a `selectedProgram` and set that value appropriately in the `PersonalTab` props. with the current implementation you need to repeat the structure of the props in the last test case. doesn't seem necessary

noisecapella:
Comments addressed

gsidebo:
looks good! 👍

PR #1570, opened 2016-10-26
Removed course run view

noisecapella:
Part of #1195

#### What's this PR do?
Removes `/api/v0/courses/` and associated views

Nothing uses this API so there should be no change of behavior


giocalitri:
👍  after tests pass

PR #1572, opened 2016-10-26
Refactor test code

noisecapella:
Part of #1195

#### What's this PR do?
Some flow type and other test refactoring

No functionality should change

### DIFF
--- a/static/js/flow/reduxTypes.js
+++ b/static/js/flow/reduxTypes.js
@@ -1,5 +1,5 @@
 // @flow
-import type { Dispatch } from 'redux';
+import type { Dispatch, State } from 'redux';
 
 export type Action = {
   type: string;
@@ -14,7 +14,6 @@ export type ActionCreator = (...a: any) => Action;
 
 export type AsyncActionCreator<T> = (...a: any) => Dispatcher<T>;
 
-export type Store = {
-  dispatch: Dispatch,
-  getState: () => any,
-};
+export type AssertReducerResultState<T> = (
+  actionFunc: () => Action, stateFunc: ((reducerState: State) => T), defaultValue: any
+) => void;
--- a/static/js/flow/reduxTypes.js
+++ b/static/js/flow/reduxTypes.js
@@ -13,3 +13,8 @@ export type AsyncActionHelper = (...a: any) => Promise<*>;
 export type ActionCreator = (...a: any) => Action;
 
 export type AsyncActionCreator<T> = (...a: any) => Dispatcher<T>;
+
+export type Store = {
+  dispatch: Dispatch,
+  getState: () => any,
+};
--- a/static/js/reducers/order_receipt_test.js
+++ b/static/js/reducers/order_receipt_test.js
@@ -1,7 +1,11 @@
+// @flow
 import configureTestStore from 'redux-asserts';
 import { assert } from 'chai';
 import sinon from 'sinon';
 import moment from 'moment';
+import type { Store } from 'redux';
+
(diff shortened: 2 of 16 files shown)


static/js/reducers/order_receipt_test.js
```diff
@@ -9,9 +9,12 @@ import {
   setTimeoutActive,
 } from '../actions/order_receipt';
 import { createAssertReducerResultState } from '../util/test_utils';
+import type { OrderReceiptState } from './order_receipt';
 
 describe('order receipt reducer', () => {
-  let sandbox, store, assertReducerResultState;
+  let sandbox, store, assertReducerResultState: (
+    a: () => Action, f: ((s: any) => OrderReceiptState), d: any
```
aliceriot: could you declare this as a parameterized type, say `AssertReducerResultState`, which would take a record type as an argument? (like `AssertReducerResultState<OrderReceiptState>`)

that would let you re-use the type declaration down below.

static/js/flow/reduxTypes.js
```diff
@@ -13,3 +13,8 @@ export type AsyncActionHelper = (...a: any) => Promise<*>;
 export type ActionCreator = (...a: any) => Action;
 
 export type AsyncActionCreator<T> = (...a: any) => Dispatcher<T>;
+
+export type Store = {
+  dispatch: Dispatch,
+  getState: () => any,
+};
```
aliceriot: We have an annotation lib file for redux (in `flow-typed`) which might have something that covers this already.

static/js/util/test_utils.js
```diff
@@ -63,27 +65,29 @@ export function generateCourseFromExisting(courseToClone: Course, desiredRuns: n
...
 
-let findActiveDialog = (dialogClassName) => (
-  [...document.getElementsByClassName(dialogClassName)].find(dialog => (
+let findActiveDialog = (dialogClassName: string): HTMLDivElement => {
+  let elements: any = document.getElementsByClassName(dialogClassName);
+  return [...elements].find(dialog => (
     isActiveDialog(dialog)
-  ))
-);
+  ));
```
aliceriot: Looking at this the second line could be simplified to just

```js
return [...elements].find(isActiveDialog);
```

noisecapella:
Yeah, I'm not sure this is a great separation of concerns between the dashboard and programs APIs but I'm not sure of a better way. It seems like `/api/v0/dashboard/` should list information about course states and leave the course description stuff to `/api/v0/programs/`, but I'm not sure where that line is drawn. The dashboard API determines the runs shown for a course so they are dependent on course state, which makes it seem like the runs should belong in `/api/v0/dashboard/`. I guess we could put all relevant information in `/api/v0/programs/` and make `/api/v0/dashboard/` point to it, something like this for the dashboard API:

    [
        {"id": 1, "courses": [{
            "runs": [{
                "id": 2,
                "status": "not-passed",
                "final_grade": 56
            }],
        ]}}
    ]

And have `/api/v0/programs/` provide all the rest of the information. It would be a big refactor though and I'm not sure how much cleaner it would be afterward

aliceriot:
Yeah, it's a bit of a tricky thing - I think we should do whatever is relatively simple and gets us the information we need.

noisecapella:
I'll hold off for now, we should have a longer conversion after standup on this with @pdpinch

noisecapella:
I'm going to remove the rename from this PR but leave the flow stuff. We can do the rename in a future PR after we have the discussion

noisecapella:
Ok, should be ready again

PR #1573, opened 2016-10-26
Protected detect_missing_migrations.sh against hanging for console input

noisecapella:
None

#### What's this PR do?
Feeds `no` to `./manage.py migrate --fake` so that it won't hang on the content type question


giocalitri:
👍

Issue #1517, opened 2016-10-24
Sign up profile form doesn't load existing program enrollments


pdpinch:
This is an edge case, because it requires an enrolled user returning to the initial sign up profile forms. 
## Steps to Reproduce:
1. Sign up. Fill in the profile. View the dashboard
2. Return to /profile/ 
## Expected behavior

The program I enrolled in at first is displayed in the first 
## Actual behavior

No  program is set and I have to select one for the form to validate.

pdpinch:
@amir-qayyum-khan I'm wondering if @noisecapella's work on #1586 affects this?

amir-qayyum-khan:
@pdpinch no, i dont think so.

PR #1577, opened 2016-10-27
Profile: Load existing program enrollments for returning user

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1517
#### What's this PR do?
- Loads enrolled program for returning user on personal tab profile page.

##### previous state

screen shot 2016-10-27 at 2 39 32 pm
##### now state

screen shot 2016-10-27 at 2 38 11 pm

**Note this PR is only for user who already filled his profile and returned back to profile page for next time**

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -24,7 +24,6 @@ import type { UIState } from '../reducers/ui';
 import type {
   AvailableProgram,
   AvailablePrograms,
-  AvailableProgramsState
 } from '../flow/enrollmentTypes';
 import type { Event } from '../flow/eventType';
 import {  validationErrorSelector } from '../util/util';
@@ -38,7 +37,7 @@ export default class PersonalTab extends React.Component {
     ui:             UIState,
     nextStep:       () => void,
     prevStep:       () => void,
-    programs:       AvailableProgramsState,
+    programs:       AvailablePrograms,
     setProgram:     Function,
     addProgramEnrollment: Function,
     currentProgramEnrollment: AvailableProgram,
@@ -55,10 +54,10 @@ export default class PersonalTab extends React.Component {
 
   setProgramHelper = (event: Event, key: string, value: string) => {
     const {
-      programs: { availablePrograms },
+      programs,
       setProgram,
     } = this.props;
-    let selected = availablePrograms.find(program => program.id === parseInt(value));
+    let selected = programs.find(program => program.id === parseInt(value));
     setProgram(selected);
   };
 
@@ -80,7 +79,7 @@ export default class PersonalTab extends React.Component {
 
   selectProgram = () => {
     const {
-      programs: { availablePrograms },
+      programs,
       errors
     } = this.props;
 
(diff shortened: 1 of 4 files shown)


gsidebo:
sorry amir, i assigned this to myself then completely forgot about it. looking into this now!

static/js/components/PersonalTab.js
```diff
@@ -58,16 +60,29 @@ class PersonalTab extends React.Component {
     setProgram(selected);
   }
 
+  getProgramId = (selectedProgram: Program, currentProgramEnrollment: ProgramEnrollment) => {
```
gsidebo: to get flow typing to recognize the return type, you could define the function like this:

`getProgramId = (selectedProgram: Program, currentProgramEnrollment: ProgramEnrollment): number|null => {`

gsidebo: also, i think it makes sense to get the values from `this.props` instead of passing them in as arguments here. 

```javascript
getSelectedProgramId = () : number|null => {
    const {
      ui: { selectedProgram },
      currentProgramEnrollment
    } = this.props;
    
    let programId = null;
    if (selectedProgram) {
      programId = selectedProgram.id;
    } else if(currentProgramEnrollment) {
      programId = currentProgramEnrollment.id;
    }
    return programId;
}
```

static/js/components/PersonalTab.js
```diff
@@ -54,20 +56,35 @@ class PersonalTab extends React.Component {
       dashboard: { programs },
       setProgram,
     } = this.props;
-    let selected = programs.find(program => program.id === value);
+    let selected = programs.find(program => program.id === parseInt(value));
     setProgram(selected);
   }
 
+  getSelectedProgramId = () : number|null => {
+    let programId = null;
```
gsidebo: can you move this definition above the `if (selectedProgram) {` line?

gsidebo:
@amir-qayyum-khan any update on this? i just had one small comment left

amir-qayyum-khan:
no,  @Gavin all logic is giving issue,  somehow it is not working at all

when we land to personal tab, enrollment select will be populated with
existing enrollment (if any)

but if user press next i give select validation error, although field is
populated.

expectation: it should take user to next page as enrollment is selected.

Are you agree with me?

amir-qayyum-khan:
Done @gsidebo

static/js/components/PersonalTab.js
```diff
@@ -34,9 +38,10 @@ export default class PersonalTab extends React.Component {
     ui:             UIState,
     nextStep:       () => void,
     prevStep:       () => void,
-    programs:       AvailablePrograms,
+    programs:       AvailableProgramsState,
```
noisecapella: Why does this need to change? I'm not seeing where we use `programs.getStatus` or anything besides the programs themselves

noisecapella:
We fixed a bug regarding the type mismatch which may be why you changed the type in PersonalTab, you should rebase from master to get the fix

amir-qayyum-khan:
@noisecapella done

static/js/components/PersonalTab.js
```diff
@@ -78,6 +97,19 @@ export default class PersonalTab extends React.Component {
     );
   }
 
+  componentWillMount() {
+    const {
```
noisecapella: Instead of resetting the `selectedProgram` state here, can you set it to null in `componentWillUnmount` to clean up once the component goes away? Then I think `getSelectedProgramId` will take care of populating the correct value

PR #1578, opened 2016-10-27
Raised 404 error when user have no access on program

amir-qayyum-khan:

fixes https://github.com/mitodl/micromasters/issues/1366
#### What's this PR do?
- When leaner tries search, now he will see 404 error if he dont have access.
- Login as learner and go to url `/learners/`.
- you will be redirect to 404

### DIFF
--- a/search/views.py
+++ b/search/views.py
@@ -3,6 +3,7 @@ Views for the Search app
 """
 
 import logging
+from django.http import Http404
 
 from rest_framework import status
 from rest_framework.views import APIView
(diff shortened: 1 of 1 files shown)


noisecapella:
Is there a test which needs to change for this case? If not, can you add one?

amir-qayyum-khan:
work is in progress @noisecapella  :)

amir-qayyum-khan:
will fixed via https://github.com/mitodl/micromasters/pull/1449

PR #1582, opened 2016-10-27
Remove x scroll on program page

### LINKED ISSUE #1561: Odd scrollbar on program page
From micromasters-rc, it's a horizontal scrollbar right below the hero image

screenshot from 2016-10-25 16-55-24

annagav:
Fix #1561
#### What's this PR do?

Fix x scroll that appears under the tabs on the program page.

### DIFF
--- a/static/scss/program-page.scss
+++ b/static/scss/program-page.scss
@@ -101,6 +101,7 @@
     .mdl-tabs__tab-bar {
       margin-bottom: 30px;
       height: 57px;
+      overflow: hidden;
     }
 
     .panel-content {


roberthouse54:
There is a bug here, but I think this may be related to the fact that we wanted the tabs to be scrollable, so that they do not wrap on mobile. This is per material design convention for "scrollable tabs" see: https://material.google.com/components/tabs.html#tabs-usage

Ideally the tabs could scroll, but not display a scroll bar.

@aliceriot

noisecapella:
Can you rebase on master to fix the tests?

annagav:
I updated the code to keep the scroll and hide it.

pdpinch:
So, how can we test this? Look at it on a phone and swipe?

annagav:
If the bug was present on your laptop you can test it there. Otherwise, I could see it on windows machine.

PR #1583, opened 2016-10-27
Fixed detect_missing_migrations.sh

noisecapella:
None

#### What's this PR do?
Adds back `detect_missing_migrations.sh` and fixes its behavior to run migrations without `--fake`

Tests should pass

### DIFF
--- a/detect_missing_migrations.sh
+++ b/detect_missing_migrations.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
-set -e -o pipefail
 
-if ./manage.py migrate --fake --no-input | grep "Your models have changes that are not yet reflected" > /dev/null
+TMPFILE=$(mktemp)
+./manage.py migrate --no-input >& "$TMPFILE"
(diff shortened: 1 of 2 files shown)


gsidebo:
if i add a new field to a model and don't run `makemigrations`, should this script let me know about that change not being reflected in the migration?

noisecapella:
Yes, if you run it locally it should error with some appropriate output

gsidebo:
This script is just running `./manage.py migrate --no-input`. we OSX users need to run `docker-compose run web ./manage.py migrate --no-input` to perform a migration. `./manage.py` on its own understandably throws an error. this should be runnable on its own, right?

noisecapella:
It should be run within docker, ie `docker-compose run web ./detect_missing_migrations.sh`, or you can do it with tox like `docker-compose run web tox`

gsidebo:
sorry, complete failure of brain with previous comment

PR #1584, opened 2016-10-27
Added enrolled field to ProgramSerializer

noisecapella:
Part of #1195

#### What's this PR do?
Adds enrolled field to ProgramSerializer

Look at `/api/v0/programs/` and verify that the `enrolled` field matches the programs you are enrolled in

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -31,13 +31,10 @@ class ProgramSerializer(serializers.ModelSerializer):
 
     def get_enrolled(self, program):
         """
-        Returns true if the user
+        Returns true if the user is enrolled in the program
         """
-        if 'request' in self.context:
-            user = self.context['request'].user
-            return ProgramEnrollment.objects.filter(user=user, program=program).exists()
-        else:
-            return False
+        user = self.context['request'].user
+        return ProgramEnrollment.objects.filter(user=user, program=program).exists()
 
     class Meta:  # pylint: disable=missing-docstring
         model = Program
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -4,12 +4,14 @@ Serializers for courses
 from rest_framework import serializers
 
 from courses.models import Course, Program
+from dashboard.models import ProgramEnrollment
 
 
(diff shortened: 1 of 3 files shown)


courses/serializers.py
```diff
@@ -27,12 +29,23 @@ def get_programpage_url(self, program):
         except ProgramPage.DoesNotExist:
             return
 
+    def get_enrolled(self, program):
+        """
+        Returns true if the user
```
gsidebo: if the user..... is enrolled? 😄

courses/serializers.py
```diff
@@ -27,12 +29,23 @@ def get_programpage_url(self, program):
...
+        """
+        if 'request' in self.context:
+            user = self.context['request'].user
```
gsidebo: is it possible for this serializer to be called when there is a `request`, but no `user` attached to that request? like from a CMS page before the user logs in?

courses/serializers.py
```diff
@@ -27,12 +29,23 @@ def get_programpage_url(self, program):
         except ProgramPage.DoesNotExist:
             return
 
+    def get_enrolled(self, program):
+        """
+        Returns true if the user
+        """
+        if 'request' in self.context:
```
noisecapella: Going to remove this check based on @giocalitri's suggestion, since we never use this serializer without a user logged in

courses/serializers.py
noisecapella: No, `user` should be `AnonymousUser` if the user is not logged in. In any case we require the user to be logged in for the API using it

courses/serializers.py
noisecapella: Ah, not enough coffee today

noisecapella:
Comments addressed

courses/serializers.py
gsidebo: understood. i think it's just coincidental that we only use it from views that require a logged-in user, but this is pretty low-risk

gsidebo:
👍  once tests are fixed

PR #1587, opened 2016-10-27
Made detect_missing_migrations.sh use makemigrations --dry-run

noisecapella:
None
#### What's this PR do?
Switches to `./manage.py makemigrations --dry-run` so the detect script doesn't actually run migrations, which may be disruptive to dev environments.

Rename a field and run the script. It should error and show a migration which could be created. No migration should actually be generated.

### DIFF
--- a/detect_missing_migrations.sh
+++ b/detect_missing_migrations.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 TMPFILE=$(mktemp)
-./manage.py migrate --no-input >& "$TMPFILE"
-if cat "$TMPFILE" | grep "Your models have changes that are not yet reflected" > /dev/null
+./manage.py makemigrations --no-input --dry-run >& "$TMPFILE"
+if cat "$TMPFILE" | grep -v "No changes detected" > /dev/null
 then
     echo "Error: one or more migrations are missing:"
     echo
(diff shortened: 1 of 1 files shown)


gsidebo:
i don't think this is working for me unless im missing something

```
(micromasters) ~/d/micromasters ❯❯❯ docker-compose run web ./manage.py makemigrations --no-input --dry-run
(micromasters) ~/d/micromasters ❯❯❯ docker-compose run web ./manage.py migrate --no-input                                            
Operations to perform:
  Apply all migrations: wagtailcore, wagtailembeds, wagtailsearch, wagtaildocs, profiles, courses, wagtailredirects, roles, default, wagtailforms, authtoken, financialaid, wagtailadmin, admin, taggit, cms, dashboard, sessions, wagtailimages, ecommerce, mail, contenttypes, wagtailusers, auth
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

`docker-compose run web ./manage.py makemigrations --no-input --dry-run` is not outputting anything

noisecapella:
Can you pull and run the script again? I think you got a case where it wanted to ask for a default value from standard input but it couldn't because `--no-input` was set

noisecapella:
Ok, think it's actually fixed now. You should test editing the `FinancialAid` model but also check `Course` as well

PR #1601, opened 2016-10-28
added cybersource settings to app.json

### LINKED ISSUE #1396: Add elasticsearch and cybersource to default PR app env variables

pdpinch:

closes #1396 
#### What's this PR do?

Adds cybserource setting names to app.json. If all works as expected, PR builds will pull the values for these settings from the CI server. 

Did the PR build get the appropriate settings?

### DIFF
--- a/app.json
+++ b/app.json
@@ -26,6 +26,24 @@
     "AWS_STORAGE_BUCKET_NAME": {
       "description": "S3 Bucket name."
     },
+    "CYBERSOURCE_ACCESS_KEY": {
+      "description": "CyberSource Access Key"
+    },
+    "CYBERSOURCE_PROFILE_ID": {
+      "description": "CyberSource Profile ID"
+    },
+    "CYBERSOURCE_REFERENCE_PREFIX": {
+      "description": "a string prefix to identify the application in CyberSource transactions"
+    },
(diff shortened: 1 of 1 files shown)


pdpinch:
@blarghmatey it works!

but... is there anyway to customize the setting for `CYBERSOURCE_REFERENCE_PREFIX` ? It's not a deal-breaker, but I can imagine some potential confusion if all these PR builds have the same prefix of `ci`

noisecapella:
If they all had the same reference prefix it's possible that an order with the same number on CI might be fulfilled instead of the PR build. Also, we will need to adjust the postback URL for each PR build we want to test

noisecapella:
Environment variables look good, I was able to make it to the cybersource checkout

app.json
```diff
@@ -26,6 +26,24 @@
     "AWS_STORAGE_BUCKET_NAME": {
       "description": "S3 Bucket name."
     },
+	"CYBERSOURCE_ACCESS_KEY": {
+      "description": "Cybersource Access Key"
```
singingwolfboy: Nitpick: Isn't it "CyberSource", with a capital S? (for all of these descriptions)

pdpinch:
I fixed the indenting (oops!) and the spelling. 

I can't think of any good way to deal with the CyberSource test center having the wrong post back URL

Issue #1498, opened 2016-10-21
Add Degree Modal on Profile has incorrect Title


j2tausig:
Steps to Reproduce:
Log into MicroMasters.
Complete a Profile.
Navigate to Dashboard.
Select "View Profile" in Header.
Click on "Add degree" link in Education section.
Note: the title of the Add Degree Modal changes to reflect whichever degree type you select

Please change the title of the Modal to "Add Degree"
![add degree modal](https://cloud.githubusercontent.com/assets/13886722/19606966/862e32aa-9794-11e6-94e6-d10ca04afdce.PNG)

pdpinch:
If the user is editing a degree, the title should be "Edit Degree"

PR #1606, opened 2016-10-28
Refactored Education frontend components

noisecapella:

Part of #1498
#### What's this PR do?

Removes `EducationDialog` and `EducationDisplay` and integrates that functionality into `EducationForm`. It's meant to make `EducationForm` more similar to `EmploymentForm` to make future refactoring easier.

Go through the profile editing steps, including `/profile` and `/learner/username`, including editing, adding and deleting education and work history. Nothing should have changed compared to master

However I did change the month/year formatting for course states, which is now  MM/YYYY to make it consistent with the profile. Any comments here?


aliceriot:
this looks a bit awkward to me:

hrrrmmm

Basically just that we effectively duplicate the information in the degree type selector. Other than that, though, functionality looks good!

noisecapella:
It should be dealt with in the next PR for #1498, this PR's just for refactoring. Any other comments?

static/js/components/EducationForm.js
```diff
@@ -151,22 +183,31 @@ class EducationForm extends ProfileFormFields {
         return <IconButton name="error" onClick={editEntry} />;
       }
     };
-    let dateFormat = date => moment(date).format("MM[/]YYYY");
+    let dateFormat = date => moment(date).format(DASHBOARD_MONTH_FORMAT);
```
aliceriot: can we move this function declaration out to be either a class method or a module level function? there's no runtime dependency so we should be OK to do that.

static/js/components/EducationForm.js
```diff
@@ -180,27 +221,174 @@ class EducationForm extends ProfileFormFields {
...
+    setEducationDegreeLevel('');
+    setEducationDialogIndex(null);
+    clearProfileEdit(username);
+  };
+
+  saveEducationForm: Function = (): void => {
+    const { saveProfile, profile, ui } = this.props;
+    saveProfile(educationValidation, profile, ui).then(() => {
+      this.clearEducationEdit();
+    });
```
aliceriot: this could just be

```js
saveProfile(educationValidation, profile, ui).then(this.clearEducationEdit);
```

static/js/components/EducationForm.js
```diff
@@ -180,27 +221,174 @@ class EducationForm extends ProfileFormFields {
...
 
-    if (profile['education'] === undefined) {
-      return null;
+  renderEducationEntries: Function = (): React$Element<*>[] => {
+    const { profile, profile: { education }} = this.props;
+    let rows = [];
+    if (education !== undefined) {
+      let sorted = educationEntriesByDate(education);
+      rows = sorted.map( ([index, entry]) => this.educationRow(entry, index));
     }
```
aliceriot: what do you think about pulling this stuff out to a separate method? If we refactored `educationRow` to take an `[index, entry]` tuple we could declare a method:

```js
const sortedEducationRows = R.compose(R.map(this.educationRow), educationEntriesByDate);
```

or without ramda:


```js
const sortedEducationRows = entries => (
  educationEntriesByDate(entries).map(this.educationRow);
);
```

static/js/components/EducationForm.js
```diff
@@ -25,23 +38,42 @@ import type {
...
+import ValidationAlert from './ValidationAlert';
 
 class EducationForm extends ProfileFormFields {
+  constructor(props: Object) {
+    super(props);
+
+    this.educationLevelLabels = {};
+    this.educationLevelOptions.forEach(level => {
+      this.educationLevelLabels[level.value] = level.label;
+    });
```
aliceriot: what do you think about moving this out of the constructor and to either the `constants` file or to a module-level declaration in this file? there isn't any dependency on runtime data for this calculation.

static/js/components/EmploymentForm.js
```diff
@@ -174,7 +189,7 @@ class EmploymentForm extends ProfileFormFields {
       profile,
     } = this.props;
 
-    let dateFormat = date => moment(date).format("MM[/]YYYY");
+    let dateFormat = date => moment(date).format(DASHBOARD_MONTH_FORMAT);
```
aliceriot: same comment about `dateFormat` here as above, maybe this could be pulled out to a util file?

noisecapella:
Comments addressed, I also changed the CSS to make the education stuff more consistent with employment, which slightly changes how things look.

static/js/util/date_test.js
```diff
@@ -2,19 +2,31 @@
 import { assert } from 'chai';
 import moment from 'moment';
 
-import { ifValidDate } from './date';
+import { ifValidDate, formatMonthDate } from './date';
 
 describe('ifValidDate', () => {
```
aliceriot: we should change this describe name to `'date utility functions'` or something equivalent

Issue #1512, opened 2016-10-24
Make facets numbers vertically  aligned


amir-qayyum-khan:
Before I forget, can you make this numbers vertically aligned ? 

![valign](https://cloud.githubusercontent.com/assets/125561/19647093/853aab70-99ca-11e6-9f74-45caf8bd37ae.png)

PR #1607, opened 2016-10-31
Fixed alignment of Current Residence on search page

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1512

#### What's this PR do?
- Fixed alignment of **Current Residence** facets like **Country of Birth**.

@pdpinch @aliceriot @roberthouse54 
screen shot 2016-10-31 at 5 45 07 pm

### DIFF
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -65,7 +65,7 @@
 
         .sk-item-list-option__text {
           margin-left: 10px;
-          color: m#666;
+          color: #666;
         }
 
         .sk-item-list-option__count {
@@ -75,14 +75,16 @@
     }
 
     .sk-hierarchical-menu-list {
+       width: 100%;
+
       .sk-hierarchical-menu-list__header {
         margin: 15px 5px;
         font-size: 15px;
       }
 
       .sk-hierarchical-menu-list__root {
-        margin-left: 6px;
-        margin-right: 0;
+        margin-left: 10px;
+        margin-right: 10px;
 
         .sk-hierarchical-menu-option {
           >div {
@@ -91,6 +93,14 @@
 
           .sk-hierarchical-menu-option__text {
             margin-right: 15px;
+          }
+        }
+
+        .sk-hierarchical-menu-list__hierarchical-options > div {
+          align-items: stretch;
+
+          .sk-hierarchical-menu-option__count {
+            color: $font-gray;
           }
         }
       }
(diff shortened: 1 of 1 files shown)


static/scss/search-page.scss
```diff
@@ -365,3 +365,20 @@
     color: rgba(0,0,0,0.8);
   }
 }
+
+.sk-hierarchical-menu-list {
+  width: 100%;
+
+  .sk-hierarchical-menu-list__root {
+    margin-left: 10px;
+    margin-right: 10px !important;
```
pdpinch: Why `!important` ?

static/scss/search-page.scss
```diff
@@ -365,3 +365,20 @@
...
+
+    .sk-hierarchical-menu-list__hierarchical-options > div {
+      align-items: stretch;
+
+      .sk-hierarchical-menu-option__count {
+        color: #6f6f6f;
```
pdpinch: Can this be the same $font-grey we use at https://github.com/mitodl/micromasters/blob/d33f2113b35fe1b2100328e92a49c615b1074f9b/static/scss/search-page.scss#L72 ?

pdpinch:
@roberthouse54 this looks good to me, but I was wondering if you had any comments on the CSS approach. 

If you want to see it in action, let me know and I'll give you the staff role on the PR build.

static/scss/search-page.scss
amir-qayyum-khan: these are new classes, we are overriding builtin classes of searchkit. this class already has margin right, here we want suppress that margin

amir-qayyum-khan:
Done @pdpinch

pdpinch:
@roberthouse54 can you take a quick look at this and give it a thumbs up?

roberthouse54:
There is a typo at line 68... other than that it looks good:

        .sk-item-list-option__text {
          margin-left: 10px;
          color: m#666;
        }

static/scss/search-page.scss
pdpinch: Can you add a comment explaining the `!important` ? 

@singingwolfboy is encouraging us to add a comment whenever we use !important so that the next developer who looks at this will know what we're overriding.

roberthouse54: @singingwolfboy I am going to venture to guess that a lot of the times we are using !important are not really necessary. I am not sure about this particular case. But in general this should be avoided if possible. In the previous case, it was overriding inline styles, which I think, makes using !important correct. But in may cases, it is just overriding a more specific css declaration, and in these cases using !important is bad practice. My concern with the idea of "adding a comment whenever we use !important" is that the implication is not just add a comment, but give a good reason for using it. What I am getting at is that the implication is to reassess the use of !important, which is a larger issue.

singingwolfboy: Agreed. Using `!important` is a "code smell": it indicates that something is probably incorrect. A comment helps to justify why `!important` is actually necessary. If you can't find a good reason for using it, don't use it!

amir-qayyum-khan:
@pdpinch there was some issues in style, i fixed those, plus figured out way to remove ``!important``

pdpinch:
👍

PR #1608, opened 2016-10-31
Speed up CI builds

### LINKED ISSUE #1593: CI build is really slow
CI builds are pretty slow, ranging from 13 minutes to 16 minutes or so on average. Since the whole test suite (for python and JS, with linters and all that) only takes about 5 minutes to finish I think we should really try to work on getting the build time down to something closer to the test execution time.

When watching builds in progress it seems to be that the majority of the time is spent assembling the docker images. One piece of low-hanging fruit we might be able to grab is to pre-build docker images (from `master`, say) and upload them to Docker hub. That way the majority of builds would not need to rebuild the python docker container at least. We could explore how to deal with `npm` as well, although I think there are reasons we don't add `npm` dependencies in the build for the `watch` container.

travis has some info about how to accomplish this [here](https://docs.travis-ci.com/user/docker/) and [here](https://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/).

thoughts? I think given the number of tests and the size of our codebase a 16 minute PR build is waayyyy too long.

aliceriot:

closes #1593 

#### What's this PR do?

A bunch of stuff! The goal is to reduce the CI build time. Before this PR our builds range between 13 and 20 minutes in length. A good chunk of that is taken up by building the docker containers and performing other setup steps, and all the tests are run sequentially.

This PR changes things around, without making any code changes, to get the build time down to about 5-6 minutes. I did that by:

- pushing pre-built containers to docker hub (https://hub.docker.com/r/mitodl/mm_watch_travis/ and https://hub.docker.com/r/mitodl/mm_web_travis/) which we pull on travis
- separating JS and Python tests into separate jobs (which run simultaneously)
- removing the `watch` container from `travis-docker-compose.yml`
- building the web, elastic, etc containers only for the python tests
- building the watch container only for the JS tests

That's about it! Hooray speed!

Read over the changes and ensure they make sense!

Take a look at the builds to see the difference in build time. Make sure that using the development environment isn't different.

fast_build_yey

### DIFF
--- a/travis/Dockerfile-travis-watch-build
+++ b/travis/Dockerfile-travis-watch-build
@@ -1,0 +1,25 @@
+FROM node:6.2
+MAINTAINER ODL DevOps <mitx-devops@mit.edu>
+
+# this dockerfile builds the hub image for the watch container
+# we don't use this directly, instead we push to docker-hub, and
+# then Dockerfile-travis-watch uses that pushed image to bootstrap
+# itself
+
+RUN apt-get update && apt-get install libelf1
+
+RUN mkdir /src
+
+WORKDIR /src
+
+RUN adduser --disabled-password --gecos "" mitodl
+
+RUN chown -R mitodl:mitodl /src
+
+USER mitodl
+
+COPY package.json npm-shrinkwrap.json ./webpack_if_prod.sh /src/
+
+RUN npm install
+
+COPY . /src
--- a/travis/js_tests.sh
+++ b/travis/js_tests.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+status=0
+
+function run_test {
+    "$@"
+    local test_status=$?
+    if [ $test_status -ne 0 ]; then
+        status=$test_status
+    fi
+    return $status
+}
+
+run_test docker run -t travis-watch npm run coverage
+run_test docker run -t travis-watch npm run lint
+run_test docker run -t travis-watch npm run scss_lint
+run_test docker run -t travis-watch npm run flow
+run_test docker run -e "NODE_ENV=production" -t travis-watch ./webpack_if_prod.sh
+
+exit $status
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,22 @@
 # Set Ruby as the language so it doesn't download the pip things. Instead, let docker do that.
 language: ruby
-cache: bundler
-install:
-  - docker-compose -f travis-docker-compose.yml build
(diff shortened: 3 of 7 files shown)


pdpinch:
I think this is failing a javascript test?

aliceriot:
@pdpinch the failing test is a flakey test which often hits the timeout. the test isn't failing, it just takes too long to complete.

pdpinch:
Ok. I restarted Travis.

aliceriot:
that test was still failing, so I upped the timeout on that particular file.

travis/js_tests.sh
```diff
@@ -0,0 +1,18 @@
+status=0
```
rhysyngsun: Should this have a shebang line at the top?

travis/Dockerfile-travis-watch-build
```diff
@@ -0,0 +1,29 @@
...
+
+WORKDIR /src
+
+RUN adduser --disabled-password --gecos "" mitodl
+
+RUN chown -R mitodl:mitodl /src
+
+USER mitodl
+
+COPY package.json /src
```
rhysyngsun: What do you think of combining these multiple `COPY` statements into one (e.g. `COPY package.json npm-shrinkwrap.json ./webpack_if_prod.sh /src/`)? It would be a very minor reduction to the layers Travis would have to pull. Tradeoff is probably that's a usage that I haven't seen in other ODL Dockerfiles so it might be unfamiliar to some.

aliceriot: Yeah, we add the package manager stuff separately from the source code to get faster builds for development (in our other dockerfiles, like `Dockerfile-node`, for instance).

Here it won't make much of a difference, since we're only using this dockerfile to build the image we push to docker hub, so it will be rebuilt only occasionally. 

I'm fine either way, I think the deciding factor is readability.

aliceriot: Actually, I think this is a good idea, I'm going to go for it.

aliceriot: Just made that change, then built and pushed a new image for `mitodl/mm_watch_travis`. The next build on this branch should pull that image.

rhysyngsun:
Looks good!

aliceriot:
Cool! What we typically do to approve a pull request is either leave a comment with a :+1: emoji, or you can use the new github review feature to add an 'approved' review.

rhysyngsun:
Ah ok, I had added the :+1: to the PR description.

PR #1609, opened 2016-10-31
Small change to width of modals on mobile

roberthouse54:
#### What's this PR do?

This adjusts the width of modal dialogs for mobile devices, and removes some of the margin in the signup dialog.

### DIFF
--- a/static/scss/responsive-styles.scss
+++ b/static/scss/responsive-styles.scss
@@ -74,6 +74,10 @@
     }
   }
 
+  div.dialog {
+    width: 87% !important;
+  }
+
   // Profile Sign Up Pages
 
   .profile-pagination {
--- a/static/scss/responsive-styles.scss
+++ b/static/scss/responsive-styles.scss
@@ -75,7 +75,13 @@
   }
 
   div.dialog {
-    width: 87% !important; // this is important to override inline styles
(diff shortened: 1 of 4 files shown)


static/scss/responsive-styles.scss
```diff
@@ -74,6 +74,10 @@
     }
   }
 
+  div.dialog {
+    width: 87% !important;
```
singingwolfboy: Can you add a comment explaining why we need to use `!important`?

static/scss/responsive-styles.scss
```diff
@@ -74,6 +74,10 @@
     }
   }
 
+  div.dialog {
+    width: 87% !important; // this is important to override inline styles
```
singingwolfboy: This comment is unclear. I would recommend something like:
> The `dialog` class is applied to Dialog components from Material UI (http://www.material-ui.com/#/components/dialog). Material UI styles its components by setting inline styles on the DOM nodes it creates. The CSS cascade defines inline styles to have a higher priority than styles defined on an external stylesheet, so normal CSS rules won't override Material UI's styles. The exception is `!important` rules, which override everything else. `!important` should be used very sparingly, but in this case, it is necessary in order to override Material UI's inline styles.

PR #1611, opened 2016-10-31
Add inline-validation

### LINKED ISSUE #1510: Sign up: validate when user hits tab (or on blur)
In usability testing, we saw a number of users attempt to fill out the profile form quickly. They skipped over some fields, assuming they weren't required only to find out when they clicked submit. 

We can address this by validating when the field loses focus.

aliceriot:

this closes #1510 

#### What's this PR do?

This implements inline-validation, so that all profile fields are validated in-line as the user fills them out, instead of waiting until the user presses save the first time to start doing validation.

OK it's a bit tricky. But basically, on `/profile` and `/learner` all profile fields (including on education and work history) should validate as-you-type, so that, for instance:

- if you tab to the first input (first name) and delete the input and tab away you should see a validation error
- if you click 'save' on `/profile` with a blank profile you should see all validation messages appear, which will disappear as you deal with them.
- in general, errors should disappear as you fix them
- fancier date valiation (like start date cannot be before end date) should happen in-line as well.

There are probably edge cases! Let's find them!


static/js/components/Navbar.js
```diff
@@ -74,9 +74,14 @@ export default class Navbar extends React.Component {
       profile,
       setNavDrawerOpen,
       setPhotoDialogVisibility,
+      navDrawerOpen,
     } = this.props;
 
-    const closeDrawer = () => setNavDrawerOpen(false);
+    const closeDrawer = () => {
```
noisecapella: Is this just to ignore the extra unnecessary dispatches?

static/js/components/inputs/inputs_test.js
```diff
@@ -156,6 +156,7 @@ describe('Profile inputs', () => {
     });
 
     it('should clear edit text when onBlur is called', () => {
+      inputProps.updateValidationVisibility = sandbox.stub();
```
noisecapella: Can we set this in the `beforeEach` instead?

static/js/reducers/index.js
```diff
@@ -141,9 +146,40 @@ export const profiles = (state: Profiles = INITIAL_PROFILES_STATE, action: Actio
...
+      return patchProfile({
+        edit: Object.assign({}, profile.edit, { errors })
+      });
+    }
+  case UPDATE_VALIDATION_VISIBILITY:
+    profile = getProfile();
+    if (profile === undefined || profile.edit === undefined) {
+      return state;
+    } else {
+      // let { edit: { visibility } } = profile;
```
noisecapella: Stray comment :)

static/js/components/inputs/DateField.js
```diff
@@ -184,6 +186,7 @@ export default class DateField extends React.Component {
         fullWidth={true}
         value={edit.year !== undefined ? edit.year : ""}
         onChange={e => setNewDate(undefined, undefined, e.target.value)}
+        onBlur={onBlur}
```
noisecapella: Should we also do this on the day and month fields? Though I'm not sure if it's possible to do it without causing weird behavior when the user moves from the day to month field for example

static/js/components/Navbar.js
aliceriot: yeah, I found them a bit annoying :)

static/js/components/inputs/DateField.js
aliceriot: yeah, that's exactly why I ended up deciding not to, I couldn't figure out a good way to keep the validation errors from appearing until after the user had finished entering a date or had clearly abandoned it.

static/js/actions/profile.js
```diff
@@ -64,6 +64,16 @@ export const updateProfileValidation = createAction(UPDATE_PROFILE_VALIDATION,
   (username, errors) => ({ errors, username })
 );
 
+export const UPDATE_VALIDATION_VISIBILITY = 'UPDATE_VALIDATION_VISIBILITY';
```
noisecapella: Can you add `profile_test.js` and add tests for these two actions? Feel free to leave out the other ones in this file for this PR

static/js/actions/profile.js
```diff
@@ -64,6 +64,16 @@ export const updateProfileValidation = createAction(UPDATE_PROFILE_VALIDATION,
...
+export const updateValidationVisibility = createAction(UPDATE_VALIDATION_VISIBILITY,
+  (username, keySet) => ({ username, keySet })
+);
+
+export const CLEAR_VALIDATION_VISIBILITY = 'CLEAR_VALIDATION_VISIBILITY';
+export const clearValidationVisibility = createAction(CLEAR_VALIDATION_VISIBILITY,
```
noisecapella: My IDE says this isn't being used anywhere, and the reducer for this doesn't have any coverage when running tests. Should we delete this?

aliceriot: oh yeah, I wrote this but ended up not using it. `clearProfileEdit` blows the whole `edit` object away, so I'm just leaning on that.

static/js/containers/ProfileFormContainer.js
```diff
@@ -68,24 +70,41 @@ class ProfileFormContainer extends React.Component {
...
     dispatch(updateProfile(username, profile));
-    this.updateProfileValidation(this.props, profile, validator);
+    if ( !skipValidation ) {
+      this.updateProfileValidation(profile, validator);
+    }
+  };
+
+  updateValidationVisibility: Function = keySet => {
+    const { dispatch, profiles } = this.props;
+    const username = SETTINGS.username;
```
noisecapella: Can you update this to `SETTINGS.user.username` here and below, to avoid collisions with #1615?

static/js/containers/ProfilePage.js
```diff
@@ -20,13 +20,19 @@ import {
...
 
 class ProfilePage extends ProfileFormContainer {
   currentStep: Function = (): string => {
     const { ui: { profileStep } } = this.props;
     return profileStep;
   };
 
+  componentDidMount() {
+    const { dispatch } = this.props;
+    dispatch(startProfileEdit(SETTINGS.username));
```
noisecapella: Can you also change `SETTINGS.username` to `SETTINGS.user.username` here and in the test?

static/js/reducers/index.js
```diff
@@ -141,9 +145,28 @@ export const profiles = (state: Profiles = INITIAL_PROFILES_STATE, action: Actio
       // caller must have dispatched START_PROFILE_EDIT successfully first
       return state;
     } else {
+      let errors = {};
+      let visibility = profile.edit.visibility;
+      if ( R.isEmpty(visibility) || R.contains(ALL_ERRORS_VISIBLE, visibility) ) {
```
noisecapella: I think `R.isEmpty` is unnecessary here, since `visibility` is always set to an empty list initially and `edit` already exists

noisecapella: Nevermind, misread this

aliceriot: Oh yeah you're right - I think I wrote this line earlier when it was possible that `visibility === undefined` and this point, and `R.containers('foo', undefined)` throws an error.

aliceriot: If you can't tell this code took a lot of tweaking to get working! hahaha

noisecapella:
It seems like the enroll functionality isn't working on the `/profile` page in this branch

static/js/containers/ProfileFormContainer.js
```diff
@@ -68,35 +70,52 @@ class ProfileFormContainer extends React.Component {
...
+      this.updateProfileValidation(profile, validator);
+    }
+  };
+
+  updateValidationVisibility: Function = keySet => {
+    const { dispatch, profiles } = this.props;
+    const username = SETTINGS.user.username;
+    if ( !profiles[username].edit ) {
+      dispatch(updateValidationVisibility(username, keySet));
+    } else if ( !R.contains(keySet, profiles[username].edit.visibility) ) {
```
noisecapella: Ah I see

static/js/containers/ProfilePage.js
```diff
@@ -20,13 +20,19 @@ import {
...
 
 class ProfilePage extends ProfileFormContainer {
   currentStep: Function = (): string => {
     const { ui: { profileStep } } = this.props;
     return profileStep;
   };
 
+  componentDidMount() {
+    const { dispatch } = this.props;
+    dispatch(startProfileEdit(SETTINGS.user.username));
```
noisecapella: Why is this needed here, and why isn't it in the other classes that extend `ProfileFormContainer`?

aliceriot: It's needed here to get around some UI edge cases, I can probably explain better outside of a comment.

It's not necessary on the other components because we're opening dialogs, and we dispatch `startProfileEdit` when we open the dialog.

aliceriot:
@noisecapella program issue is fixed, genuinely not sure what I did to break it...anyway, moving on!

aliceriot:
@noisecapella I had to do make some changes for merge conflicts with https://github.com/mitodl/micromasters/pull/1615 and https://github.com/mitodl/micromasters/pull/1606, would you mind taking a quick look again? functionality shouldn't be different.

noisecapella:
Yep

aliceriot:
thanks!

noisecapella:
I think everything's still working fine :+1:

PR #1617, opened 2016-11-01
Fixed course date issue in alter_data command

### LINKED ISSUE #1612: Course Status Start Date mismatch
Program:  Supply Chain Management
Course: Supply Chain Analytics
Micromasters-RC
Site is displaying a Start Date of 11/1/2016
(see snapshot)

But the following is showing a Start Date in the database of 11/2/2016
Peter noticed that the start time in the Data is 00:00 -- so wondering if the issue might involve timezones?

 python manage.py alter_data --action=course_info --email='jtausig@mit.edu' --program-title='Chain' --course-title='Analytics'
{
  "course_runs": [
    {
      "upgrade_deadline": "12/07/2016",
      "start_date": "11/02/2016",
      "enrollment_end": "12/07/2016",
      "end_date": "12/28/2016",
      "edx_course_key": "course-v1:MITx+CTL.SC0x+3T2016",
      "enrollment_start": "05/01/2016",
      "id": 59
    }
  ],
  "course_title": "Supply Chain Analytics"

start date

### LINKED ISSUE #1613: Course Status Enrollment Date doesn't match data
Not sure if this is the same issue as #1612  -- (that was Start date and this involves Enrollment begins date).

Ran the following query:
python manage.py alter_data --action=course_info --email='jtausig@mit.edu' --program-title='Chain' --course-title='Fundamental'

Which returned:
"course_runs": [
    {
      "edx_course_key": "course-v1:MITx+CTL.SC1x_3+1T2017",
      "start_date": null,
      "id": 62,
      "enrollment_start": "01/01/2030",
      "end_date": null
    },
    {
      "enrollment_end": "05/10/2016",
      "edx_course_key": "course-v1:MITx+CTL.SC1x_2+1T2016",
      "start_date": "02/10/2016",
      "id": 61,
      "enrollment_start": "02/10/2016",
      "end_date": "05/10/2016"
    },
    {
      "enrollment_end": "08/28/2015",
      "edx_course_key": "course-v1:MITx+CTL.SC1x_1+2T2015",
      "start_date": "05/20/2015",
      "id": 14,
      "enrollment_start": "05/20/2015",
      "end_date": "08/28/2015"
    },
    {
      "enrollment_end": "12/24/2014",
      "edx_course_key": "course-v1:MITx+ESD.SCM1x+3T2014",
      "start_date": "09/24/2014",
      "id": 60,
      "enrollment_start": "09/24/2014",
      "end_date": "12/24/2014"
    },
    {
      "edx_course_key": "MITx/ESD.SCM1x/3T2014",
      "start_date": "09/24/2014",
      "id": 17,
      "enrollment_start": "09/24/2014",
      "end_date": "12/24/2014"
    }
  ],
  "course_title": "Supply Chain Fundamentals"
}

That suggests an enrollment begins date of 1/1/2030, but on rc, we are displaying an enrollment begins date of 12/31/2029.
(screenshot below)

Additionally, there is no start date, yet we are showing a fuzzy date of January 2017 -- where is that coming from?

enroll future fuzzy start

gsidebo:

Fixes #1612 
Fixes #1613 

#### What's this PR do?

For the course_info convenience method, prints out a localized datetime (since that's what moment.js does)

Just run the `alter_data` command with the `course_info` action and some relevant parameters to find a program, course, and user, eg: `manage.py alter_data --action=course_info --username=staff --program-title='Digital' --course-title='100'`

### DIFF
--- a/seed_data/lib.py
+++ b/seed_data/lib.py
@@ -5,9 +5,10 @@ import re
 from datetime import timedelta
 from functools import wraps
 from django.db import IntegrityError
+from django.db.models import Q
 
 from django.contrib.auth.models import User
-from courses.models import Program, Course
+from courses.models import Program, Course, CourseRun
 from dashboard.models import CachedCertificate, CachedEnrollment, CachedCurrentGrade
 from financialaid.factories import TierProgramFactory
 from financialaid.models import TierProgram
@@ -111,7 +112,7 @@ def course_state_editor(func):
         course_run = kwargs.get('course_run')
         if not course_run:
             course = kwargs['course']
-            clear_edx_data_for_course(user, course)
+            clear_edx_data(user, course=course)
             course_run = course.courserun_set.order_by('fuzzy_start_date', '-start_date').first()
             kwargs['course_run'] = course_run
             del kwargs['course']
@@ -147,9 +148,18 @@ class CachedHandler(object):
     def __init__(self, user):
         self.user = user
 
-    def clear_all(self, course_run=None):
-        """Sets data to None for all cached edX objects associated with a User (and CourseRun, if provided)"""
+    def get_missing_course_runs(self, course_runs):
+        """Given a list of CourseRuns, return the ones that do not have a cached edX object for the User"""
+        course_run_ids = set(course_run.id for course_run in course_runs)
+        cached_course_run_ids = self.model_cls.objects.filter(user=self.user).values_list("course_run__id", flat=True)
+        missing_course_run_ids = course_run_ids.difference(set(cached_course_run_ids))
+        return [cr for cr in course_runs if cr.id in missing_course_run_ids]
+
+    def clear_all(self, course=None, course_run=None):
+        """Sets data to None for all cached edX objects associated with a User (and Course/CourseRun, if provided)"""
         params = {'user': self.user}
+        if course:
+            params['course_run__course'] = course
         if course_run:
             params['course_run'] = course_run
         return self.model_cls.objects.filter(**params).update(data=None)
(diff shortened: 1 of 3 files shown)


seed_data/management/commands/alter_data.py
```diff
@@ -197,28 +202,34 @@ def add_past_failed_run(user=None, course=None, now=None, grade=DEFAULT_FAILED_G
     return chosen_past_run
 
 
+def _formatted_datetime(dt, dt_format='%m/%d/%Y'):
+    return localized_datetime(dt).strftime(dt_format) if dt else None
```
singingwolfboy: Nitpick: there's a _lot_ of logic on this line. Can we break it apart a bit?
```python
if not dt:
    return None
return localized_datetime(dt).strftime(dt_format)
```

seed_data/utils.py
```diff
@@ -64,3 +65,11 @@ def create_past_date_range(ended_days_ago=5, day_spread=10, now=None):
 def future_date(now=None, days_in_future=30):
     """Creates a date in the future"""
     return now + timedelta(days=days_in_future)
+
+
+def localized_datetime(dt, tz='US/Eastern'):
```
singingwolfboy: Why is the "US/Eastern" timezone the default? Why not UTC?

gsidebo: moment.js on the front end is parsing these dates in the local timezone (US/Eastern). this function is only being used by a convenience method that spits out course information. it's being used by jared right now to aid in testing our various course states. this allows him to match up the dates that are being shown to him on the front end with the actual data, and that matching can only happen when the dates are localized consistently

seed_data/management/commands/alter_data.py
gsidebo: done

gsidebo:
any other changes @singingwolfboy ?

gsidebo:
@singingwolfboy sorry, can you review my most recent commit too? we found another bug and i want to smuggle the fix into this PR

seed_data/lib.py
```diff
@@ -250,25 +253,36 @@ def set_data(self, obj, course_run, grade=DEFAULT_GRADE, **kwargs):
...
-def clear_edx_data_for_course(user, course, models=None):
-    """Clears all of a User's cached edX data for a Course, or just the data for specific cached models"""
+def ensure_cached_edx_data_for_user(user):
+    """
+    Ensures that all possible cached edX records have been initialized for a User
+    """
+    all_course_runs = CourseRun.objects.filter(course__program__live=True).exclude(
+        Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
+    ).all()
+    all_course_run_ids = set([course_run.id for course_run in all_course_runs])
```
singingwolfboy: Nitpick: you don't need those braces at all. You can use a generator comprehension, like this:
```python
set(course_run.id for course_run in all_course_runs)
```
It's faster too, because it will turn the query results directly into a set, instead of turning them into a list first.

seed_data/lib.py
```diff
@@ -250,25 +253,36 @@ def set_data(self, obj, course_run, grade=DEFAULT_GRADE, **kwargs):
...
+    results = []
+    for cached_model in [CachedEnrollment, CachedCertificate, CachedCurrentGrade]:
+        all_cached_enrollments = cached_model.objects.filter(user=user).values_list("course_run__id", flat=True)
+        uninit_course_run_ids = set(all_course_run_ids).difference(set(all_cached_enrollments))
```
singingwolfboy: `all_course_run_ids` is already a set. `all_cached_enrollments` isn't, but you could easily make it one when it is defined. Then you can just do operations on these sets, without calling `set()` on them again.

seed_data/lib.py
```diff
@@ -250,25 +253,36 @@ def set_data(self, obj, course_run, grade=DEFAULT_GRADE, **kwargs):
...
+        if len(uninit_course_run_ids) > 0:
+            uninit_course_runs = [cr for cr in all_course_runs if cr.id in uninit_course_run_ids]
+            create_result = cached_model.objects.bulk_create([
```
singingwolfboy: Can you add a comment here, explaining what this code does? There's a lot going on here, and a lot of abstraction. A comment would help make this much clearer.

gsidebo:
@singingwolfboy just added some details and cleared up some names. i hope this doesn't sound like a cop-out, but this code is unrelated to the normal function of the app. it's just supporting tools that help us tinker with test data. as such, the coding standards probably don't need to be as stringent. also, this `ensure_cached_edx_data_for_user` method is 100% a one-off - it isn't conceptually similar to anything else in the module, so i'm more comfortable with an 18 line function with semi-complicated logic than i otherwise would be

PR #1618, opened 2016-11-01
Made small optimization to user serialization for search results

### LINKED ISSUE #1600: Optimize (and validate) UserProgramSerializer queries
UserProgramSerializer in dashboard/serializers.py queries cached enrollments and certificates for a user, but does not take advantage of prefetching. This is an easy fix that aligns with Django best practices and will probably result in performance gains

Also, this class is used to serialize data for search results in ES, and it hasn't been updated since the introduction of financial aid programs. We should make sure that this serialization still works for both FA and non-FA programs

gsidebo:

Closes #1600 

#### What's this PR do?

I initially thought we could optimize the queries to retrieve cached data for the User on a ProgramEnrollment in dashboard/serializers.py. It turns out that we pretty much optimal. I poked around this area of the code and found that (a) a User's Profile could be fetched in a single query using select_related, eliminating 1 query per user serialized, and (b) cached model serialization only needed the data property. Both of these changes can probably be extended to other areas of the code

Make sure you can still recreate the index and the dashboard API still works. You can also count the queries executed by search.indexing_api.index_users on master, then with the same users on this branch

### DIFF
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -13,7 +13,7 @@ from elasticsearch_dsl.connections import connections
 from profiles.models import Profile
 from profiles.serializers import ProfileSerializer
 from dashboard.models import ProgramEnrollment
-from dashboard.serializers import UserProgramSerializer
+from dashboard.serializers import UserProgramSearchSerializer
 from search.exceptions import ReindexException
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


giocalitri:
there is a conflict

gsidebo:
updated the manual testing info

search/indexing_api.py
```diff
@@ -134,7 +134,7 @@ def index_users(users, chunk_size=100):
     """
     Indexes a list of users via their ProgramEnrollments
     """
-    program_enrollments = ProgramEnrollment.objects.filter(user__in=users).select_related('user', 'program').all()
+    program_enrollments = ProgramEnrollment.prefetched_qset().filter(user__in=users).all()
```
giocalitri: Nit: I think the `.all()` is unnecessary here

giocalitri:
one very small comment.

giocalitri:
functionality looks good.
👍 after conflict is addressed and tests pass

PR #1619, opened 2016-11-01
Staff users can see only students they are allowed to

### LINKED ISSUE #1588: Staff user of program should only see that program's learners in the learners display
Currently it looks like they can see learners in other programs too

giocalitri:
fixes #1588 

#### What's this PR do?
Fixes a bug that allowed staff users to see students of program where the staff user was not staff.

This PR also removes the GET request handling support from the ES proxy

the bug has been fixed in
`search/api.py`

all the rest is work to fix the reason why this bug did not come out from the tests before.

Make sure the staff user can see only the students in the programs where she is staff

### DIFF
--- a/roles/api.py
+++ b/roles/api.py
@@ -15,7 +15,7 @@ def get_advance_searchable_programs(user):
     Returns:
         list: list of courses.models.Program instances
     """
-    user_role_program = Role.objects.filter(user=user)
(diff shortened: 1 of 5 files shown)


gsidebo:
functionality looks good. looking at code now

giocalitri:
@gsidebo done

PR #1623, opened 2016-11-02

blarghmatey:
None

#### What's this PR do?
Adds the UJSON renderer to the Django Rest Framework

Validate that the API is still able to send and receive JSON data.

The UJSON renderer is up to 2x or 3x faster than the default one that ships with Python.

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -21,7 +21,7 @@ from celery.schedules import crontab
 from django.core.exceptions import ImproperlyConfigured
 
 
-VERSION = "0.20.0"
+VERSION = "0.18.0"
 
 CONFIG_PATHS = [
     os.environ.get('MICROMASTERS_CONFIG', ''),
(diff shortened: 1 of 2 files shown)


noisecapella:
Speaking of parsing JSON I saw this recently, about the corner cases of JSON parsing that are treated differently by different parsers: http://seriot.ch/parsing_json.html

Though I feel like if our tests are passing we should generally be good for that

micromasters/settings.py
```diff
@@ -21,7 +21,7 @@
 from django.core.exceptions import ImproperlyConfigured
 
 
-VERSION = "0.20.0"
+VERSION = "0.18.0"
```
noisecapella: Can you revert this?

noisecapella:
Looks like everything still works, merge if tests pass :+1:

blarghmatey:
This is actually showing a negative performance impact and increased error rate during load tests. Abandoning this PR.

PR #1627, opened 2016-11-02
Added indices for all dates in the CourseRun model

giocalitri:
N/A

#### What's this PR do?
Introduces indices for all the dates in the CourseRun

Recent load tests revealed a slow query performed on these dates.


noisecapella:
Did you get any improvement in the performance tests? Code looks fine

giocalitri:
not in a significant way: I am open to not merge this PR, but I would prefer to leave it in, given the very law impact

what do you think?

noisecapella:
Yeah, I agree with you that inserts and updates will be very rare for `CourseRun` :+1:

PR #1615, opened 2016-11-01
Replace `SETTINGS.username` with `SETTINGS.user.username`

singingwolfboy:
Follow-up to #1482: that pull request introduced the `SETTINGS.user` object, this pull request makes the rest of the existing Javascript use this object. It also _removes_ the `SETTINGS.username` string, so that no new code will reference it.

This pull request should have no user-visible changes.

PR #1628, opened 2016-11-02
Zendesk prepopulate program on program page

singingwolfboy:
Partial fix for #1262. This pull request builds on #1482 to also prepopulate the program field in the Zendesk ticket submission form. This pull request does several things:

* ~~Adds a `slug` field to the Program model, which is used to note which Zendesk field value this program is associated with~~
* Adds a `ProgramPageSerializer` class, which is used to serialize information about the program into the `SETTINGS` object in the template
* Refactors the callbacks in `zendesk_widget.js` to reflect the fact that Zendesk inserts _four_ iframes into the page, not just one
* Adds a `ticketSubmissionFormLoaded` callback so that, if we are on a program page, the program field dropdown will be set to whatever value matches the program on the page

Note that this pull request _only_ pre-populates the program field on a program page. No other page will cause the program field in the Zendesk form to pre-populate. A future pull-request can do that.

Also note that by including the faculty list and course list in the `ProgramPageSerializer`, we get some duplicated information in the `SETTINGS` object. I will clean up this duplication in a future pull request, as I did with  #1482 and #1615.

**How should this be manually tested?**

Go to the Programs model in the Django Admin (`/admin/courses/program/`). Create or edit a program so that the `slug` value is "scm". Next, make sure there is a ProgramPage for this program in Wagtail (`/cms`). If there isn't, go to the "Explorer" section in Wagtail, add a child page, and set the "Program" dropdown to the program you just edited in the Django admin. Visit the live page in your browser -- you'll need to publish the page in Wagtail before you can visit it. Once the page is loaded, click on the "Help" button at the bottom right of the page. Verify that the "MicroMasters Program" field in the ticket submission form is already set to "Supply Chain Management".

Next, go back to the Django Admin, and change the `slug` value to "dedp". Go back to the program page for this program, refresh the page, and click the "Help" button again. This time, the "MicroMasters Program" field in the ticket submission form should already be set to "Data, Economics, and Development Policy".

Finally, go back to the Django Admin, and change the `slug` value to any other string. Go back to the program page for this program, refresh the page, and click the "Help" button again. The "MicroMasters Program" field in the form should be set to the default value of "-".

### DIFF
--- a/cms/serializers.py
+++ b/cms/serializers.py
@@ -3,6 +3,7 @@ Serializers for Wagtail-related models
 """
 from rest_framework import serializers
 from wagtail.wagtailimages.models import Image, Rendition
+from django.utils.text import slugify
 
 from cms.models import ProgramPage, ProgramFaculty
 from courses.serializers import CourseSerializer
@@ -55,10 +56,10 @@ class ProgramPageSerializer(serializers.ModelSerializer):
         return programpage.program.id
 
     def get_slug(self, programpage):  # pylint: disable=no-self-use
-        """Get the slug from the program"""
+        """Slugify the program's title for Zendesk"""
         if not programpage.program:
             return None
-        return programpage.program.slug
+        return slugify(programpage.program.title)
 
     class Meta:  # pylint: disable=missing-docstring
         model = ProgramPage
--- a/cms/serializers.py
+++ b/cms/serializers.py
@@ -4,7 +4,8 @@ Serializers for Wagtail-related models
 from rest_framework import serializers
 from wagtail.wagtailimages.models import Image, Rendition
 
-from cms.models import ProgramFaculty
+from cms.models import ProgramPage, ProgramFaculty
+from courses.serializers import CourseSerializer
 
 
 class RenditionSerializer(serializers.ModelSerializer):
@@ -36,3 +37,29 @@ class FacultySerializer(serializers.ModelSerializer):
     class Meta:  # pylint: disable=missing-docstring
         model = ProgramFaculty
         fields = ('name', 'title', 'short_bio', 'image')
+
+
+class ProgramPageSerializer(serializers.ModelSerializer):
+    """
+    Used to output info into the SETTINGS object on a program page.
+    """
+    id = serializers.SerializerMethodField()
+    slug = serializers.SerializerMethodField()
+    faculty = FacultySerializer(source='faculty_members', many=True)
+    courses = CourseSerializer(source='program.course_set', many=True)
+
+    def get_id(self, programpage):  # pylint: disable=no-self-use
+        """Get the ID of the program"""
+        if not programpage.program:
+            return None
+        return programpage.program.id
+
+    def get_slug(self, programpage):  # pylint: disable=no-self-use
+        """Get the slug from the program"""
+        if not programpage.program:
+            return None
+        return programpage.program.slug
+
+    class Meta:  # pylint: disable=missing-docstring
+        model = ProgramPage
+        fields = ('id', 'title', 'slug', 'faculty', 'courses')
--- a/courses/models.py
+++ b/courses/models.py
@@ -19,6 +19,7 @@ class Program(models.Model):
     A degree someone can pursue, e.g. "Supply Chain Management"
     """
     title = models.CharField(max_length=255)
+    slug = models.CharField(max_length=255, help_text="Used for Zendesk")
     live = models.BooleanField(default=False)
     description = models.TextField(blank=True, null=True)
     financial_aid_availability = models.BooleanField(default=False, null=False)
--- a/courses/models.py
+++ b/courses/models.py
@@ -19,7 +19,6 @@ class Program(models.Model):
     A degree someone can pursue, e.g. "Supply Chain Management"
     """
     title = models.CharField(max_length=255)
-    slug = models.CharField(max_length=255, help_text="Used for Zendesk")
     live = models.BooleanField(default=False)
     description = models.TextField(blank=True, null=True)
     financial_aid_availability = models.BooleanField(default=False, null=False)
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -1,4 +1,4 @@
-/* global SETTINGS:false zE:false */
+/* global SETTINGS:false zE:false _:false */
 __webpack_public_path__ = `http://${SETTINGS.host}:8078/`;  // eslint-disable-line no-undef, camelcase
 
 // Start of odl Zendesk Widget script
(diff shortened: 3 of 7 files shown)


noisecapella:
Instead of adding a slug field, could we construct one by doing something like `program-{id}` or by slugifying the title?

singingwolfboy:
@noisecapella only if whoever is maintaining Zendesk follows the same algorithm, and does so consistently. @pdpinch suggested just adding a slug field.

pdpinch:
My rationale for the slug is that we use one in the CMS ("scm" and "dedp", and I configured ZenDesk to use the same slug. 

But I can change ZenDesk. And I value isn't visible in ZenDesk, just the label.

noisecapella:
I'd prefer to avoid an explicit slug field if possible, just bad experiences with LORE. Would `program-{id}` be sufficient for the value in ZenDesk or would something like `supply-chain-management` be more useful?

noisecapella:
Also, does Zendesk know if the message is coming from micromasters production or micromasters-rc, for example?

noisecapella:
cc @pdpinch

pdpinch:
@noisecapella ZenDesk can't currently differentiate between the -ci, -rc and prod servers and I'm not sure it's worth the effort. So `program-{id}` isn't ideal, but it's still workable since I expect we'll be copying data from production back to -ci and -rc more often going forward.

pdpinch:
@noisecapella ZenDesk can't currently differentiate between the -ci, -rc and prod servers and I'm not sure it's worth the effort. So `program-{id}` isn't ideal, but it's still workable since I expect we'll be copying data from production back to -ci and -rc more often going forward.

noisecapella:
Ok, can you switch to use `program-{id}` @singingwolfboy?

singingwolfboy:
@noisecapella @pdpinch: just to be clear, you would like me to _remove_ the `slug` field entirely, and use the program's database ID instead? This will require whoever is setting up the Zendesk side to be aware of the program's database ID when creating/editing these custom tags. Is that correct?

noisecapella:
@pdpinch Do you know how/if the user would see these id's in Zendesk?

pdpinch:
George and I just talked about this. Let's go with the slugified title (using django's `slugify()`).

I would like to test what happens if ZenDesk receives a ticket with a value for program it doesn't recognize. If we test this PR with slugify we'll get an answer to that question.

cms/serializers.py
```diff
@@ -36,3 +38,29 @@ class FacultySerializer(serializers.ModelSerializer):
...
+    courses = CourseSerializer(source='program.course_set', many=True)
+
+    def get_id(self, programpage):  # pylint: disable=no-self-use
+        """Get the ID of the program"""
+        if not programpage.program:
+            return None
+        return programpage.program.id
+
+    def get_slug(self, programpage):  # pylint: disable=no-self-use
+        """Slugify the program's title"""
```
noisecapella: Can you add that this is for Zendesk?

courses/models.py
```diff
@@ -68,6 +68,7 @@ def __str__(self):
 
     class Meta:
         unique_together = ('program', 'position_in_program',)
+        ordering = ('position_in_program',)
```
singingwolfboy: I pulled this out into a new pull request: #1692. Maybe we can get that one merged first, and then rebase this PR on top of it.

noisecapella: Sounds good

noisecapella:
I'm getting an error when I view a program page on Firefox. Chrome works however:

screenshot from 2016-11-07 12-13-32

singingwolfboy:
@noisecapella I believe I fixed the problem. Can you double-check?

singingwolfboy:
@pdpinch Also, this pull request will not cause Zendesk to submit tickets for programs it doesn't recognize. If the program on the page has a slug that doesn't match any of the values in the ticket submission form, then the Javascript in this pull request will leave the form alone, and the "Program" dropdown will remain at it's default value of "-".

noisecapella:
I'm getting this 400 error which only happens occasionally. Here are some screenshots, do you have any idea what's going on here?

screenshot from 2016-11-08 10-40-07
screenshot from 2016-11-08 10-40-36

singingwolfboy:
Looks like it's a problem on Zendesk's end. I don't think there's anything we can do about it.

noisecapella:
It looks like the `identify` 400 error was because we weren't passing an email address, which we didn't have because I was logged in as root. So I think we can ignore it

static/js/entry/zendesk_widget.js
```diff
@@ -47,65 +47,104 @@ zE(function() {
...
+    };
+    btn.onmouseenter = setHover;
+    btn.onmouseleave = unsetHover;
+    unsetHover();
+
+    // prepopulate ticket submission form
+    if (SETTINGS.program) {
+      const programSlug = SETTINGS.program.slug;
+      btn.onclick = () => {
+        setTimeout(() => {
```
noisecapella: If this is a workaround for something, can you add a comment describing what it is?

static/js/entry/zendesk_widget.js
```diff
@@ -47,65 +47,104 @@ zE(function() {
...
+            select.value = programSlug;
+          }
+        }, 100);
+      };
+    }
+  }
+};
+
+const setupZendeskCallbacks = () => {
+  const zendeskIFrames = ["launcher", "ticketSubmissionForm", "nps", "ipm"];
```
noisecapella: Were you expecting to use this with any of the other iframes besides `launcher`? It's fine to leave it, just curious why this change was made

singingwolfboy: I had thought that I would need to use the `ticketSubmissionFormLoaded` function, to modify the submission form as soon as it loaded. However, it appears that Zendesk's Javascript re-renders that form when the launcher button is clicked, so my modifications got clobbered by the re-render. Once I restructured the code so that I could define these arbitrary lifecycle functions, though, I figured it was an improvement over what was there before, so I left it as-is for the pull request.

static/js/entry/zendesk_widget.js
```diff
@@ -47,65 +47,104 @@ zE(function() {
     zE.identify(identity);
   }
 
-  // trigger onZendeskIFrameExists at the appropriate time
+  setupZendeskCallbacks();
+});
+
+const zendeskCallbacks = {
```
singingwolfboy: Done.

static/js/entry/zendesk_widget.js
singingwolfboy: Added a comment.

pdpinch:
@singingwolfboy I just updated the settings in ZenDesk

screenshot 2016-11-16 16 51 43

singingwolfboy:
@pdpinch the values in the Zendesk web widget still haven't updated, oddly. I took this screenshot just now.
zendesk-web-widget

pdpinch:
Sorry, it appears that my changes weren't saved (maybe it wasn't my fault, but probably was). 

Can you try again?

singingwolfboy:
Now it works!

PR #1633, opened 2016-11-03
Save leading zeros in month and date field

### LINKED ISSUE #1529: Sign up: display zeroes in date fields
The form label asks for dates in various places with "mm / yyyy" and "mm / dd / yyyy" but if you type "08" only the 8 displays. 

It would be less surprising for users if the zero displayed as well as the other digits -- even if it was later omitted.

annagav:
Fix #1529
#### What's this PR do?
Keep a leading zero in a month and date field if the user entered it.

### DIFF
--- a/static/js/components/inputs/DateField.js
+++ b/static/js/components/inputs/DateField.js
@@ -92,14 +92,15 @@ export default class DateField extends React.Component {
         day: day !== undefined ? day : edit.day
       });
 
+      const firstIfNumEqual = R.curry((x, y) => Number(x) === y ? x : y);
       let validatedDay = Maybe.of(1);
       if (!omitDay) {
         validatedDay = validateDay(newEdit.day);
-        newEdit.day = mstr(validatedDay);
+        newEdit.day = mstr(validatedDay.map(firstIfNumEqual(newEdit.day)));
       }
 
       let validatedMonth = validateMonth(newEdit.month);
-      newEdit.month = mstr(validatedMonth);
+      newEdit.month = mstr(validatedMonth.map(firstIfNumEqual(newEdit.month)));
 
       let validatedYear;
       if ( allowFutureYear ) {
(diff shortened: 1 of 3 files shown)


static/js/components/inputs/DateField.js
```diff
@@ -95,11 +95,11 @@ export default class DateField extends React.Component {
       let validatedDay = Maybe.of(1);
       if (!omitDay) {
         validatedDay = validateDay(newEdit.day);
-        newEdit.day = mstr(validatedDay);
+        newEdit.day = dateForm(mstr(validatedDay), newEdit.day);
```
aliceriot: I don't think you need this `dateForm` function here. 

When `newEdit.day` is a valid day number you'll get `Just(day)` back from `validateDay`, and otherwise you'll get `Nothing`. I'd suggest you use the `.map` method on `Maybe` to sort this out! Something like:

```js
newEdit.day = mstr(validatedDay.map(day => day === Number(newEdit.day) ? newEdit.day : day));
```

If `validatedDay` is `Nothing`, then calling `.map` on it will just return another `Nothing`, so we're safe to call `mstr` with it as an argument (`mstr(Nothing())` returns `''`).

If `validatedDay` is a day, but it's different than what the user typed in, then we want to only show them the validated value. If the user types in a bigger number than 31, for instance, we change it to be 31 (take a look at `lib/validation/date.js` for details).

If the value wrapped in `validatedDay` is equal to `Number(newEdit.day)` then we can safely let the user's input (with or without a leading zero) through.

does that make sense?

aliceriot: The little comparator function could be made more general too, something like:

```js
const firstIfNumEqual = R.curry((x, y) => Number(x) === y ? x : y);
```

then in `.map` you'd just have:

```js
mstr(validatedDay.map(firstIfNumEqual(newEdit.day)))
```

which would make it a little easier to use the same pattern for months, below.

annagav:
@aliceriot Thanks for the suggestion!

PR #1636, opened 2016-11-03
Upgrade Django to 1.10.3

### LINKED ISSUE #781: upgrade to django 1.10
The latest release contains veri nice features:
https://docs.djangoproject.com/en/1.10/releases/1.10/

But we cannot upgrade without fixing something.
1- TEMPLATE_\* warning 

```
    WARNINGS:
    ?: (1_8.W001) The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_CONTEXT_PROCESSORS.
```

(and anything else we might find)

singingwolfboy:
Fixes #781.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ dj-static==0.0.6
 django-role-permissions==1.2
 django-server-status==0.3.2
 django-storages-redux==1.3.2
-Django==1.9.10
+Django==1.10.3
 djangorestframework==3.5.2
 edx-api-client==0.3.0
 elasticsearch==2.3.0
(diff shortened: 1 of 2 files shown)


giocalitri:
only about the last commit: why isn't this failing with the current libraries versions?

singingwolfboy:
@giocalitri I don't know. I'm not sure what's causing Python-Social-Auth to behave differently on Django 1.9 vs Django 1.10.

giocalitri:
👍 if tests pass

PR #1635, opened 2016-11-03
Upgrade Wagtail to 1.7

singingwolfboy:
Fixes #1525

PR #1637, opened 2016-11-03
Upgrade Pillow to 3.4.2

singingwolfboy:
Wagtail 1.7 (#1635) updates the version of [Willow](https://pypi.python.org/pypi/Willow) that it depends on. I figured it would make sense to also make sure that the underlying Pillow library was up to date. [ChangeLog is here.](https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst)


giocalitri:
👍 after you fix conflicts

PR #1638, opened 2016-11-03
Upgrade Django REST Framework to 3.5.2

singingwolfboy:
Django REST Framework is one of the core frameworks that this project uses. It's better to do smaller, more frequent upgrades than to wait until bugs or deprecations force us to do larger, infrequent upgrades. [ChangeLog is here.](http://www.django-rest-framework.org/topics/release-notes/)


giocalitri:
I think this can be merged 👍

PR #1641, opened 2016-11-03
Docker Compose version 2

singingwolfboy:
This upgrades our [Docker Compose](https://www.docker.com/products/docker-compose) file to the [version 2 syntax](https://docs.docker.com/compose/compose-file/#/versioning). This new syntax allows you to specify named volumes and mount them onto containers. This pull request creates a named volume for holding Django media and mounts it onto the `web` container. Previously, every time the `web` container was rebuilt, Wagtail lost all of its uploaded images; with this change, uploaded images live in this named volume, and will persist even if the developer rebuilds the `web` container.

There should be no user-visible changes from this pull request.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN curl --silent --location https://bootstrap.pypa.io/get-pip.py | python3 -
 # Add, and run as, non-root user.
 RUN mkdir /src
 RUN adduser --disabled-password --gecos "" mitodl
+RUN mkdir /var/media && chown -R mitodl:mitodl /var/media
 
 # Install project packages
 COPY requirements.txt /tmp/requirements.txt
@@ -30,8 +31,6 @@ RUN chown -R mitodl:mitodl /src
 RUN ./manage.py collectstatic --noinput
 RUN apt-get clean && apt-get purge
 USER mitodl
-
-RUN mkdir /var/media && chown -R mitodl:mitodl /var/media
 
 # Set pip cache folder, as it is breaking pip when it is on a shared volume
 ENV XDG_CACHE_HOME /tmp/.cache
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,9 @@ RUN ./manage.py collectstatic --noinput
 RUN apt-get clean && apt-get purge
 USER mitodl
 
+RUN mkdir /var/media
+RUN chown -R mitodl:mitodl /var/media
+
 # Set pip cache folder, as it is breaking pip when it is on a shared volume
 ENV XDG_CACHE_HOME /tmp/.cache
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,7 @@ RUN ./manage.py collectstatic --noinput
 RUN apt-get clean && apt-get purge
 USER mitodl
 
-RUN mkdir /var/media
-RUN chown -R mitodl:mitodl /var/media
+RUN mkdir /var/media && chown -R mitodl:mitodl /var/media
 
 # Set pip cache folder, as it is breaking pip when it is on a shared volume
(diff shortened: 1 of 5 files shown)


travis/Dockerfile-travis-web
```diff
@@ -13,5 +13,6 @@ WORKDIR /src
 COPY . /src
 
 RUN chown -R mitodl:mitodl /src
+RUN mkdir /var/media && chown -R mitodl:mitodl /var/media
```
aliceriot: Do we need to do this here? the `mm_web_travis` container that this inherits from is built using the root-level `Dockerfile`.

Dockerfile
```diff
@@ -31,6 +31,8 @@ RUN ./manage.py collectstatic --noinput
 RUN apt-get clean && apt-get purge
 USER mitodl
 
+RUN mkdir /var/media && chown -R mitodl:mitodl /var/media
```
aliceriot: Just running `docker-compose build` on my machine locally it seems like the `mitodl` user doesn't have permission to create this directory.

travis/Dockerfile-travis-web
aliceriot: Also the container built with this dockerfile only persists for the duration of the tests, do we need this directory at all?

Dockerfile
singingwolfboy: I just ran `docker-compose build` on my machine, and didn't have any permission problems. Can you post a gist of your stack trace?

singingwolfboy: Weird! [Here's my successful traceback.](https://gist.github.com/singingwolfboy/c24a0785db48e68911f8c5260efa509d). Can you make sure you're running the latest version of Docker and Docker Compose? I don't know if that would have anything to do with it, but I suppose it's worth checking.

noisecapella: I'm not sure why there would be a difference (not familiar with the state of the art with OS X Docker) but it seems like moving `USER mitodl` after this line would fix the problem.

singingwolfboy: We can definitely try that, but I want to spend some time and see if we can get to the bottom of this issue. It would help to know what the difference is, here.

aliceriot: docker v 1.12.1, docker-compose 1.8.0

aliceriot: Yep moving `USER mitodl` after that line makes the build pass for me. It shouldn't build correctly as-is, the `mitodl` shouldn't have permission to modify `/var` right?

travis/Dockerfile-travis-web
aliceriot: yea this line is unnecessary. the line above was necessary to avoid a permissions error.

aliceriot: This dockerfile inherits from an image which is build from the root level dockerfile, so I don't think we really need to strive for this being the same. The reason for this dockerfile is just to handle the case where a branch introduces a dependency change, we want to make sure we run a reinstall of the dependencies.

Dockerfile
noisecapella: I don't know, I'm not really sure what docker-machine's linux environment looks like. Maybe the `1000` user on the machine's environment has permission to alter `/var/media`? ¯\\\_(ツ)\_/¯

aliceriot:
things are building now!

singingwolfboy:
@aliceriot: but the tests are failing on Travis, due to permission errors. Are you certain that line in `Dockerfile-travis-web` was unnecessary? I think that removing it caused the tests to fail.

aliceriot:
It looks like the tests aren't passing, probably because the changes to the root level Dockerfile are not reflected in the images pushed to docker hun.

aliceriot:
I think we should try pushing an image built with the changed Dockerfile and see if that fixes it.

aliceriot:
looks like things are passing now! weee

PR #1643, opened 2016-11-03
Upgrade python-social-auth to 0.2.21

singingwolfboy:
[ChangeLog is here.](https://github.com/omab/python-social-auth/blob/master/CHANGELOG.md) Based on a quick skim, there seem to be some security-related updates -- not sure if we're vulnerable, but probably a good idea to upgrade regardless.

In addition, Django 1.10 (#1636) depends on this upgrade.


giocalitri:
👍

PR #1649, opened 2016-11-03
Updated requirements to use pip-tool

blarghmatey:
#### What's this PR do?
Introduces `pip-tools` for managing Python dependency pinning. Instead of using pip freeze to create a `requirements.txt` file a `requirements.in` file is created with loose version pinning and the frozen dependencies are generated to a `requirements.txt` file using the `pip-compile` command.

By using pip freeze for generating requirements.txt there is a lot of confusion created as to which packages we actually care about. Using pip-compile lets us specify the top-level dependencies and then freeze the specific versions of all dependent packages periodically.

**Updated requirements to use pip-tool**
Added `pip-tool` to test_requirements.txt and updated handling of
requirements files to use `pip-compile` command. This lets us pin the
exact versions of all requirements used in production while loosely
specifying the top-level requirements in the `.in` files.

To regenerate e.g. `requirements.txt` one would run `pip-compile requirements.in`.


amir-qayyum-khan:
please rebase this PR @blarghmatey

amir-qayyum-khan:
@blarghmatey so do we need to add ``pip-compile requirements.in`` command to generate requirements.txt in any script for docker, trivas, hiroku ?

blarghmatey:
@amir-qayyum-khan we will run it manually when we want to update packages and the resulting requirements file(s) will be tracked in git.

amir-qayyum-khan:
@blarghmatey please rebase it

amir-qayyum-khan:
👍

PR #1653, opened 2016-11-03
Removed react-loader, use react-mdl react

### LINKED ISSUE #1350: Use react-mdl spinner in place of react-loader for more consistent design

### LINKED ISSUE #1563: When dashboard refreshes only the footer is visible for a period of time
screenshot from 2016-10-25 17-37-01

noisecapella:
Fixes #1350

#### What's this PR do?
Replaces the existing spinner with a material design spinner

Load a page and look at the spinner while it's loading.

screenshot from 2016-11-03 17-15-59

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "react-dropzone": "^3.6.0",
     "react-ga": "1.4.1",
     "react-hot-loader": "1.3.0",
-    "react-loader": "2.4.0",
     "react-mdl": "1.5.4",
     "react-page-click": "^3.0.0",
     "react-redux": "4.4.5",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4723,11 +4723,6 @@
       "from": "react-input-autosize@>=1.1.0 <2.0.0",
       "resolved": "https://registry.npmjs.org/react-input-autosize/-/react-input-autosize-1.1.0.tgz"
     },
-    "react-loader": {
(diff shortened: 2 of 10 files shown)


noisecapella:
@roberthouse54 Any comments on how this looks?

roberthouse54:
Looks a bit big, but I am having a hard time seeing it. The footer taking up the screen seems to block it. I do see it but only for a fraction of a second. I think if you can make it about 2/3 the size you have it now, that would be better.

package.json
```diff
@@ -59,7 +59,6 @@
     "react-dropzone": "^3.6.0",
     "react-ga": "1.4.1",
     "react-hot-loader": "1.3.0",
-    "react-loader": "2.4.0",
```
aliceriot: can you make sure `npm-shrinkwrap.json` is updated to account for this change? if you do a fresh install with the `package.json` and `npm-shrinkwrap.json` from `master` and then do `npm uninstall --save react-loader` that should do the trick.

noisecapella: I did that and it wasn't working, there was some unrelated dependency of searchkit that caused an error. So to fix it I deleted `node_modules`, ran `npm install`, then ran `npm install --save`, then did the `npm uninstall --save react-loader`. Not sure I'll ever understand npm

noisecapella:
Any other comments?

package.json
aliceriot: haha yea npm is a cryptic thing

Issue #1427, opened 2016-10-17
Please grey out "save" button on Upload a Profile Photo Pop-up until image uploaded


j2tausig:
On the "Upload a Profile Photo Pop-up" the "save" button is active and can be clicked even though no image has been uploaded.
Per a conversation with Rob, please grey out the button until an image has been uploaded.

Steps to reproduce:
Log into MicroMasters
Create a profile in edX.
Click on the camera icon on the Profile Tab  -- this spawns the Upload a Profile Photo Pop-up.
When a user clicks on the "save" button without uploading an image, nothing happens and currently, we don't provide any messaging to the user.

![save button](https://cloud.githubusercontent.com/assets/13886722/19452394/deb074d8-947e-11e6-82ba-b5b5df5b5aa7.PNG)

PR #1654, opened 2016-11-04
Allow to use save button only if photo is selected on 'photo upload dialog'

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1427

#### What's this PR do?
- when user selects photo then updates ``save button`` classes. so that user can save photo. 

@pdpinch @roberthouse54 @aliceriot 
screen shot 2016-11-04 at 4 18 24 pm

screen shot 2016-11-04 at 4 18 36 pm

### DIFF
--- a/static/js/components/ProfileImageUploader.js
+++ b/static/js/components/ProfileImageUploader.js
@@ -74,7 +74,7 @@ const ProfileImageUploader = ({
       </Button>,
       <Button
         type='button'
-        className={photo ? 'primary-button save-button' : 'secondary-button save-button disabled'}
+        className={`save-button ${photo ? 'primary-button' : 'secondary-button disabled'}`}
         key="save"
         onClick={updateUserPhoto}>
         Save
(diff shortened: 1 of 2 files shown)


static/js/components/ProfileImageUploader.js
```diff
@@ -74,7 +74,7 @@ const ProfileImageUploader = ({
       </Button>,
       <Button
         type='button'
-        className='primary-button save-button'
+        className={photo ? 'primary-button save-button' : 'secondary-button save-button disabled'}
```
aliceriot: this is a bit redundant. since we always want `.save-button` why not do something like:

```js
className={`save-button ${photo ? 'secondary-button disabled' : 'primary-button'}`}
```

aliceriot: oh that ternary should be the other way around lol

```js
className={`save-button ${photo ? 'primary-button' : 'secondary-button disabled'}`}
```

amir-qayyum-khan:
Done @aliceriot

Issue #1374, opened 2016-10-13
The backend code to fetch the user profile should handle errors


giocalitri:
if the request to `/api/mobile/v0.5/my_user_info` fails, the app crashes

this might be dependent on #1375

amir-qayyum-khan:
@giocalitri i was thinking to look into https://github.com/omab/python-social-auth/blob/f13a5cc3c8670a1a3bbc16410bb30ac0423d9934/social/backends/base.py#L228,

we cant catch exception here by overriding this method, 
2nd if response has error we can log it and exit from here

why do you think?

giocalitri:
the problem with this approach is: what happens to the user in Micromaster? this API is called during the login and if it is the very first login, we might leave the user locally in an inconsistent state.

So the approach might be good depending on what happens to MM.

amir-qayyum-khan:
so if that is the scenario then we should display error messages on MM or
may be we can log him out.  (first time login)

else if there is already data cached on MM then we can use that, ( we can
cache user api),  if no cache available then need to go with logout/error
message

giocalitri:
It can work if nothing is created for that user in the MM database.
But it is still unclear to me how you make the Pipeline of the Python Social Auth stop without an exception.

pdpinch:
We decided that there wasn't any user benefit to making this change.

PR #1657, opened 2016-11-04
Exception logging on user api calls

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1374

#### What's this PR do?
- handle exception on user api(s) to avoid crash on pipeline

@pdpinch @giocalitri

### DIFF
--- a/backends/edxorg.py
+++ b/backends/edxorg.py
@@ -1,11 +1,16 @@
 """
 EdX.org backend for Python Social Auth
 """
+import logging
 from datetime import datetime
 from urllib.parse import urljoin
+import requests
 
 from django.conf import settings
 from social.backends.oauth import BaseOAuth2
+from social.exceptions import AuthFailed
+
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
@giocalitri any thoughts?

noisecapella:
It seems like if we are unable to request the username from edX we can't find which user to log in as. I'm not sure if this solution would help anything.

amir-qayyum-khan:
should we logout user then? and show toast or error message?

pdpinch:
@giocalitri @noisecapella what should we do with this PR? 

Was the goal here just to make sure that we know there is a serious issue with the edx APIs?

noisecapella:
I feel like it should be closed, I'm not sure it gets us anything. If the API call fails with an uncaught exception we should see it in sentry I hope. If we fail to get the username for the user from edX then there's nothing we can do to help the user

PR #1673, opened 2016-11-04
Removed cheaper setting for uWSGI

blarghmatey:
**Removed cheaper setting for uWSGI**
When trying to set the value of `processes` to match the value of
`cheaper` uWSGI crashes. Removed the `cheaper` setting and lowered the
default process count.


giocalitri:
👍

PR #1679, opened 2016-11-04
Implemented sending emails on order errors

### LINKED ISSUE #1336: Send an email when an Order is marked as failed
We should add code to monitor that failed orders are dealt with promptly. I think there are two main cases where it may fail: when we enroll the user in edX, and when the payment processor fails for some internal reason. 

We may also want to set up some trigger in the cybersource receipt email account to look for failed decisions, in case we don't get them on the server

noisecapella:
Fixes #1336

#### What's this PR do?
Sends emails if the order fulfillment failed, or if the enrollment failed afterwards

Set up mailgun's environment variables, in particular `MAILGUN_RECIPIENT_OVERRIDE` should point to an email address you control. Start a purchase of a course and in Cybersource click 'cancel'. You should get an email saying that there was an order fulfillment failure. Make sure the recipient email matches `EMAIL_SUPPORT` in `settings.py`.

Next, change the course key for the course run you purchased to something invalid. Go through a purchase workflow and you should see the order fulfillment fail in the UI due to the enrollment in the course not working. You should get an email saying that the enrollment failed and a stack trace.

### DIFF
--- a/ecommerce/views.py
+++ b/ecommerce/views.py
@@ -1,5 +1,6 @@
 """Views from ecommerce"""
 import logging
+import traceback
 from urllib.parse import urljoin
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


noisecapella:
Thanks for taking the effort for a thorough review!

Issue #1614, opened 2016-11-01
If user switches to their profile before the profile API is loaded, they go to /learner/undefined


noisecapella:
We should disable the user menu while the APIs are loading

pdpinch:
Which page is this? The dashboard?

Why is the profile necessary to render the user settings?

noisecapella:
It looks like we use `profile.username` to get the username for this link, so before the profile is actually loaded this doesn't have any information. We can switch to `SETTINGS.user.username` which should be always available

PR #1690, opened 2016-11-07
Loaded username param of profile page url from SETTINGS

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1614

#### What's this PR do?
- Now getting username for profile page url from SETTINGS object

@noisecapella @pdpinch

### DIFF
--- a/static/js/components/Navbar.js
+++ b/static/js/components/Navbar.js
@@ -101,7 +101,8 @@ export default class Navbar extends React.Component {
               </div>
               <div className="link">
                 <Icon name="person" />
-                <Link to={`/learner/${profile.username}`} onClick={closeDrawer} >
+                <Link to={`/learner/${SETTINGS.user ? SETTINGS.user.username : profile.username}`}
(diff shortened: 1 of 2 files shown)


static/js/components/Navbar.js
```diff
@@ -101,7 +101,8 @@ export default class Navbar extends React.Component {
               </div>
               <div className="link">
                 <Icon name="person" />
-                <Link to={`/learner/${profile.username}`} onClick={closeDrawer} >
+                <Link to={`/learner/${SETTINGS.user ? SETTINGS.user.username : profile.username}`}
```
noisecapella: Can we change this to just `SETTINGS.user.username`? I don't think the user will see this code unless they are logged in

amir-qayyum-khan:
actually this part is executed for mobile,  and there in no check for
setting as it is in usermanu.  thats why some unit test were failing

amir-qayyum-khan:
done @noisecapella

PR #1691, opened 2016-11-07
Script to update docker hub images

aliceriot:
This is just a little script to automate building new docker images and pushing them to the [mitodl organization](https://hub.docker.com/u/mitodl/dashboard/) on docker hub. In order to use it you'll need to be logged in to a docker hub account which has permissions to push to the mitodl organization.

I just used it to push new images, so I guess if the builds are still passing it works! hahaha.

### DIFF
--- a/travis/update-docker-hub.sh
+++ b/travis/update-docker-hub.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -eo pipefail
+
+docker build -t mitodl/mm_web_travis -f Dockerfile .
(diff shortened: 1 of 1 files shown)


travis/update-docker-hub.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
```
noisecapella: Can you add `set -eo pipefail` here to stop execution if a non-zero value is returned?

aliceriot: oh yah good idea

PR #1693, opened 2016-11-07
Use dicts for Python-Social-Auth instead of strings

singingwolfboy:
This is necessary to upgrade Django to 1.10 (#1636).


singingwolfboy:
@giocalitri It appears that this upgrade can be done independently of the Django one. Perhaps Django 1.10 drops backwards-compatibility for JSON-as-a-string?

giocalitri:
👍

PR #1694, opened 2016-11-07
Add repository and license fields to package.json

singingwolfboy:
Because I'm tired of `npm install` complaining every single time.


rhysyngsun:
:+1: lgtm

PR #1695, opened 2016-11-07
Allow all hosts while in DEBUG mode

singingwolfboy:
I discovered a problem when running Django 1.10.3 via Docker: it forbids access based on the IP address, even while `DEBUG=True`. [This change to Django was introduced in 1.10.3.](https://docs.djangoproject.com/en/1.10/releases/1.10.3/#dns-rebinding-vulnerability-when-debug-true)

Because we never debug against production data, we can safely disable this protection. Doing so will allow us to run within Docker and use [Docker-Machine](https://docs.docker.com/machine/overview/).


giocalitri:
👍

PR #1701, opened 2016-11-07
Make camera icon accessible

### LINKED ISSUE #1684: Accessibility: change alt tag for camera icon
> The camera icon adjacent to the Profile image reads as "camera_alt" to the screen reader user, which is not understandable.

The HTML needs to be updated to include a `<button>` tag to signal to the screenreader that this icon is functional

singingwolfboy:
Fixes #1684. Sets an accessible description for the camera icon on the user profile page, and uses a `<button>` element to indicate that clicking on the camera will open a dialog.


rhysyngsun:
:+1:

PR #1703, opened 2016-11-08
Switch to `manage.py showmigrations`

singingwolfboy:
The `manage.py migrate --list` command is removed in Django 1.10. It has been replaced with the `manage.py showmigrations` command. https://docs.djangoproject.com/en/1.10/ref/django-admin/#django-admin-showmigrations

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -11,7 +11,7 @@ MANAGE_FILE=${MANAGE_FILE:2}
 # Run migrations
 
 echo "-----> Running django migrations"
-python $MANAGE_FILE migrate --list 2>&1 | indent
+python $MANAGE_FILE showmigrations --list 2>&1 | indent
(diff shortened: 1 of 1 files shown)


bin/post_compile
```diff
@@ -11,7 +11,7 @@ MANAGE_FILE=${MANAGE_FILE:2}
 # Run migrations
 
 echo "-----> Running django migrations"
-python $MANAGE_FILE migrate --list 2>&1 | indent
+python $MANAGE_FILE showmigrations --list 2>&1 | indent
```
giocalitri: is this exactly the same as before?

singingwolfboy: Seems that way, yes. https://gist.github.com/singingwolfboy/7a12076520eccf19851ae00f461f5ab2

giocalitri:
I think this can be merged 👍

Issue #1605, opened 2016-10-28
On learner page, employment or education cards should disappear if there is no information to display


noisecapella:
When a user views another learner's profile and that learner doesn't have an employment or education, the empty cards shouldn't be displayed.

pdpinch:
If the cards are hidden, how will the learner enter additional employment or education?

@roberthouse54

pdpinch:
@noisecapella I updated the description -- is that what you have in mind?

noisecapella:
Yes that's correct

amir-qayyum-khan:
then how to display add work history or education link? should we display only cards with ``Add`` links?

currently ui is like 

<img width="984" alt="screen shot 2016-11-07 at 6 29 01 pm" src="https://cloud.githubusercontent.com/assets/10431250/20059441/1bfe118e-a518-11e6-8fee-784efd5c89ed.png">

pdpinch:
@amir-qayyum-khan take a look at the updated description. This only applies for users looking at another learner's profile, not their own. A user isn't allowed to edit other users' profiles.

amir-qayyum-khan:
got it thanks

PR #1704, opened 2016-11-08
Removed empty education and work history cards on learners page

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1605

#### How to test:
- open your profile ``/learner``.
- open other user profile with no work history and education``/learner``.
- open other user profile with work history and education ``/learner``.
- open your ``/profile`` and see education and work tabs

@pdpinch @noisecapella 
#### Logged in user profile
screen shot 2016-11-08 at 7 54 20 pm

#### Other user profile
screen shot 2016-11-08 at 7 54 17 pm

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -129,12 +129,8 @@ class EducationForm extends ProfileFormFields {
     </Cell>;
   }
 
-  renderEducationLevelEntries(level: any): Array<React$Element<*>|void>|void|null {
+  renderEducationLevelEntries(level: any): Array<React$Element<*>|void>|void {
     const { profile } = this.props;
-    if (!isProfileOfLoggedinUser(profile) && (!profile.education || profile.education.length === 0)) {
-      return null;
-    }
-
     let levelValue = HIGH_SCHOOL;
     let filterDegreeName = () => true;
     let title;
@@ -169,7 +165,7 @@ class EducationForm extends ProfileFormFields {
     ];
   }
 
-  renderEducationLevel(level: Option): Array<React$Element<*>|void>|React$Element<*>|void|null {
+  renderEducationLevel(level: Option): Array<React$Element<*>|void>|React$Element<*>|void {
     if (this.hasEducationAtLevel(level.value)) {
       return this.renderEducationLevelEntries(level);
     } else {
@@ -341,21 +337,20 @@ class EducationForm extends ProfileFormFields {
         </Card>;
       });
     } else if (profile !== undefined) {
-      let educationLevelEntries = this.renderEducationLevelEntries(null);
-      if (educationLevelEntries) {
-        return (
-          <Card shadow={1} className="profile-form" id="education-card">
-            <Grid className="profile-form-grid">
-              <Cell col={12} className="profile-form-row profile-card-header">
-                <span className="title">
-                  Education
-                </span>
-              </Cell>
-              { this.renderEducationLevelEntries(null) }
-            </Grid>
-          </Card>
-        );
+      if (!isProfileOfLoggedinUser(profile) && (!profile.education || profile.education.length === 0)) {
+        return null;
       }
+
+      return <Card shadow={1} className="profile-form" id="education-card">
+        <Grid className="profile-form-grid">
+          <Cell col={12} className="profile-form-row profile-card-header">
+            <span className="title">
+              Education
+            </span>
+          </Cell>
+        { this.renderEducationLevelEntries(null) }
+        </Grid>
+      </Card>;
     } else {
       return null;
     }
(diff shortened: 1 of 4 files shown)


static/js/components/EducationForm.js
```diff
@@ -334,16 +341,21 @@ class EducationForm extends ProfileFormFields {
         </Card>;
       });
     } else if (profile !== undefined) {
```
aliceriot: why don't you just check here that

```js
 } else if (profile !== undefined && profileOfLoggedinUser(profile) ) {
```

then the rest of the code in this function can stay the same, and you don't need to call `this.renderEducationLevelEntries`. Calling that function as control flow is pretty confusing IMO, because you're using a side-effect of it's implementation. This means, for someone reading the code, that they can no longer assume that `renderEducationLevelEntries` does what it's name implies, and they have to go read the implementation to see what's going on.

static/js/components/EducationForm.js
```diff
@@ -126,8 +129,12 @@ class EducationForm extends ProfileFormFields {
     </Cell>;
   }
 
-  renderEducationLevelEntries(level: any): Array<React$Element<*>|void>|void {
+  renderEducationLevelEntries(level: any): Array<React$Element<*>|void>|void|null {
     const { profile } = this.props;
+    if (!profileOfLoggedinUser(profile) && (!profile.education || profile.education.length === 0)) {
+      return null;
+    }
```
aliceriot: This function shouldn't have this control flow in it, it should be a more pure function which just takes a level as an argument and pulls the profile out of `this.props`. See my other comment, but I think this change is confusing. This function should stay the same, and below in the `renderCard` method use the `profileOfLoggedInUser` to check whether it should be called.

static/js/components/EmploymentForm.js
```diff
@@ -287,6 +294,12 @@ class EmploymentForm extends ProfileFormFields {
       ui: { workHistoryEdit }
     } = this.props;
     let cardClass = workHistoryEdit ? '' : 'profile-tab-card-greyed';
+    let cardBody = this.renderCardBody();
+
+    if (!cardBody) {
```
aliceriot: in what circumstance will this be true?

static/js/util/util.js
```diff
@@ -40,6 +40,10 @@ export function sendGoogleAnalyticsEvent(category: any, action: any, label: any,
   ga.event(event);
 }
 
+export function profileOfLoggedinUser(profile: Profile): boolean {
+  return (SETTINGS.user && profile.username === SETTINGS.user.username);
+}
+
```
aliceriot: can you rewrite this with parentheses? just lets us skip the `return` keyword.

static/js/components/EmploymentForm.js
amir-qayyum-khan: it will be true in case:

- when user is viewing employment on profile
- when user is viewing his own profile
- when user is viewing other user profile with data

It will be false when.
- user is viewing other learner profile and their is no data

static/js/components/EducationForm.js
amir-qayyum-khan: - we want to stop rendering here for other users if they do not have education or work history.
- But for own profile, if you do not have data i.e education or work history then we will show you empty card with ``Add degree`` or ``Add employment`` button

static/js/components/EducationForm.js
aliceriot: My point more broadly is the the check that you have up in `renderEducationLevelEntries` should be done in this function. `renderEducationLevelEntries` should do what it says and not have control flow in it, and the logic for when to render things and when not to should be in one place - in this function (`renderCard`).

Basically we want to do this, if I'm understanding you correctly:

```js
} else if (profile !== undefined)
  if (!isProfileOfLoggedinUser(profile) && (!profile.education || profile.education.length === 0)) {
    return null;
  }
  return <Card ... />;
```

If I'm understanding you correctly there are 4 cases:

1. profile not logged in, education present
2. profile not logged in, education not present
3. profile logged in, education present
4. profile logged in, education not present
  
With this setup in case 1, 3, and 4 we would render education entries, and we would omit them in case 2.

Isn't that what we're shooting for? If not let me know what I'm missing.
 So in this function we check for all the conditions where we would want to hide the education entries (if the profile is not logged in and there aren't education entries on it). In all other cases (education entries present on non-logged in user

static/js/components/EmploymentForm.js
aliceriot: In the last case won't we have already executed line 166?

amir-qayyum-khan:
done @aliceriot

static/js/util/util_test.js
```diff
@@ -258,7 +287,7 @@ describe('utility functions', () => {
       assert.equal(userPrivilegeCheck(profile, privilegedCallback, unprivilegedString), "emacs");
     });
 
-    it('should return the value of the second function if user is not login', () => {
+    it('should return the value of the second function if user is not logged in', () => {
```
aliceriot: haha good catch :)

PR #1706, opened 2016-11-08
Upgrade pylint to 1.6.4

singingwolfboy:
Also fixing the new errors that the new version of Pylint discovered. Pulling this out of #1649, to make that pull request smaller.


giocalitri:
👍

PR #1708, opened 2016-11-08
Added reset status to financial aid

### LINKED ISSUE #1201: Reset learner's financial aid request
Learners are not allowed to make multiple financial aid submissions, but accidents do happen. If a learner contacts us, we need a method to reset their state and let them submit again. 

We don't need a course-team UI for this. It can be via django admin, if that is convenient. 

As discussed, we don't want to delete the original submission, but rather change the state and allow a new submission.

giocalitri:
closes #1201

#### What's this PR do?
Adds a `reset` state to the Financial Aid applications so that an user can have another application if all the others are in `reset` state

probably from `financialaid/models.py`

Everything should work as before, but from the ADMIN you can change the financial aid to `reset` and in the dashboard the user will be able to apply again.

This PR needs extensive manual testing especially. 
@pdpinch I think you should take care of validate that all the previous functionality and constraints are intact.


pdpinch:
Functionality looks perfect for me!

Out of curiosity, how did you populate the database? You're missing a tier for incomes below $10k. If that was generated from a script, I'll open an issue to fix that.

noisecapella:
:+1: Feel free to merge after addressing comment

financialaid/constants.py
```diff
@@ -47,7 +49,7 @@ class FinancialAidStatus:
         DOCS_SENT: "Documents Sent by User",
         PENDING_DOCS: "Started Applications",
         PENDING_MANUAL_APPROVAL: "Pending Approval (Documents Received)",
-        SKIPPED: "Skipped"
+        SKIPPED: "Skipped",
```
noisecapella: The user can still see these applications by adding 'reset' to the URL, but feel free to leave it as is since this isn't exposing any vulnerability

PR #1715, opened 2016-11-08
Clean up unnecessary JS references in program page

singingwolfboy:
Follow-up from #1628. Removes unnecessary, duplicated content from the `SETTINGS` Javascript object on the programs page.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -15,23 +15,11 @@ from wagtail.wagtailcore.models import Orderable, Page
 from wagtail.wagtailimages.models import Image
 
 from courses.models import Program
-from courses.serializers import CourseSerializer
 from micromasters.serializers import serialize_maybe_user
 from micromasters.utils import webpack_dev_server_host
 from profiles.api import get_social_username
 from roles.models import Instructor, Staff
 from ui.views import get_bundle_url
-
-
-def faculty_for_carousel(faculty):
-    """formats faculty info for the carousel"""
-    from cms.serializers import FacultySerializer
-    return FacultySerializer(faculty, many=True).data
-
-
-def courses_for_popover(courses):
-    """formats course info for the popover"""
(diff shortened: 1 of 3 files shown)


static/js/entry/course_list.js
```diff
@@ -9,10 +9,16 @@ import getMuiTheme from 'material-ui/styles/getMuiTheme';
...
+  courseList = SETTINGS.program.courses;
+}
 
-ReactDOM.render(
-  <MuiThemeProvider muiTheme={getMuiTheme()}>
-    <CourseListWithPopover courses={SETTINGS.courses} />
-  </MuiThemeProvider>,
-  courseListEl
-);
+if (courseList && courseListEl !== null) {
```
noisecapella: If `SETTINGS.program` exists, `courseList` should always be an array of some length. Can you remove that part of the expression?

static/js/entry/faculty_carousel.js
```diff
@@ -7,9 +7,14 @@ import ReactDOM from 'react-dom';
 import FacultyCarousel from '../components/FacultyCarousel';
 
 const carouselDiv = document.querySelector('#faculty-carousel');
-if (carouselDiv !== null) {
+let facultyList = null;
```
noisecapella: It looks like `FacultyCarousel` is expecting this to not be null. Can you default it to `[]` instead?

noisecapella: Nevermind, it looks like this is covered by the `if` check below

static/js/entry/course_list.js
singingwolfboy: I changed this to `!_.isEmpty(courseList)`. Does that address your comment?

noisecapella: Yes that's fine

PR #1716, opened 2016-11-08
Hide photo upload on public profile (#1603)

### LINKED ISSUE #1603: Remove the user photo upload from profile when viewed by other users

rhysyngsun:

Fixes #1603 

#### What's this PR do?

This PR hides the image upload button on non-private profile pages when viewed by anyone other than the logged in user whose profile is being viewed.

Easiest way to test this is to set your profile to public and then view it anonymously. You should not see the camera icon unless you're logged in as that user.

Logged in and viewing own profile:
selection_008

Anonymous or viewing someone else's profile:
selection_009

### DIFF
--- a/static/js/containers/ProfileImage.js
+++ b/static/js/containers/ProfileImage.js
@@ -6,7 +6,11 @@ import Icon from 'react-mdl/lib/Icon';
 import moment from 'moment';
 import type { Dispatch } from 'redux';
 
-import { makeProfileImageUrl, getPreferredName } from '../util/util';
+import {
+  makeProfileImageUrl,
+  getPreferredName,
+  userPrivilegeCheck,
+} from '../util/util';
 import type { Profile } from '../flow/profileTypes';
 import ProfileImageUploader from '../components/ProfileImageUploader';
 import { createActionHelper } from '../lib/redux';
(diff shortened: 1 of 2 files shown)


static/js/containers/ProfileImage.js
```diff
@@ -91,7 +95,7 @@ class ProfileImage extends React.Component {
           alt={`Profile image for ${getPreferredName(profile, false)}`}
           className="card-image"
         />
-        { this.cameraIcon(editable) }
+        { userPrivilegeCheck(profile, () => this.cameraIcon(editable) ) }
```
aliceriot: what do you think about quickly refactoring `this.cameraIcon` so that it pulls `editable` out of `this.props` by itself? Then we could avoid declaring the extra function literal here, and just do

```js
{ userPrivilegeCheck(profile, this.cameraIcon) }
```

static/js/containers/ProfileImage_test.js
```diff
@@ -0,0 +1,78 @@
...
+  const thatProfile = {
+    username: 'rfeather',
+    email: 'rf@example.com',
+    first_name: 'Reginald',
+    last_name: 'Feathersworth',
+    preferred_name: 'Reggie'
+  };
+
+  beforeEach(() => {
+    helper = new IntegrationTestHelper();
```
aliceriot: is this used anywhere?

rhysyngsun: Yes, on line 21. I found the `<ProfileImage/>` component throw an error because a `store` wasn't being provided. If there's a better way to accomplish that automatically, I'm all ears because it's not the most elegant to pass something the component should get implicitly like that.

aliceriot: Oh gotcha. Yeah, I've just created an independent store in similar situations, this looks like an easier solution, haha.

rhysyngsun:
@aliceriot I've refactored the `cameraIcon` as you suggested, good catch!

PR #1720, opened 2016-11-08
Fixed footer display while JS loads

### LINKED ISSUE #1563: When dashboard refreshes only the footer is visible for a period of time

noisecapella:
Fixes #1563

#### What's this PR do?
Makes the footer disappear while the javascript loads, then adds it back afterwards by toggling the display of the footer element.

Refresh the browser and see if you notice the footer during the refresh

I'm not sure if there's a better way to toggle visibility but it seems simple enough


rhysyngsun:
:+1:

PR #1723, opened 2016-11-09
Implemented auto approve for TierProgram where discount is $0

### LINKED ISSUE #1341: If user is assigned the zero discount tier, we should proceed as if they were auto-approved from a UX point of view

noisecapella:
Fixes #1341

#### What's this PR do?
Auto-approves the user if they are below the income threshold for their country but they are matched with a $0 discount anyway.

Configure the tier program to have a $0 discount level that's lower than the income threshold for your country of residence (this is $75k for US). Then apply for financial aid with an income slightly above the $0 discount tier program level

### DIFF
--- a/financialaid/views_test.py
+++ b/financialaid/views_test.py
@@ -8,6 +8,7 @@ import ddt
 from django.core.exceptions import ImproperlyConfigured
 from django.core.urlresolvers import reverse
 from django.db.models import Q
+import pytz
 from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.test import APIClient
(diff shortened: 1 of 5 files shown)


financialaid/views_test.py
```diff
@@ -67,39 +74,61 @@ def setUp(self):
...
+            assert financial_aid.status == FinancialAidStatus.AUTO_APPROVED
+        self.assertAlmostEqual(financial_aid.income_usd, income_usd)
+        assert financial_aid.user == self.profile.user
+        self.assertAlmostEqual(financial_aid.original_income, original_income)
+        assert financial_aid.original_currency == original_currency
+        assert financial_aid.country_of_income == self.profile.country
+        assert financial_aid.country_of_residence == self.profile.country
+        now = datetime.datetime.now(tz=pytz.UTC)
+        five_seconds = datetime.timedelta(0, 5)
+        assert now - five_seconds < financial_aid.date_exchange_rate < now + five_seconds
```
gsidebo: i would implement the 3 lines above like so:

```python
date_diff = financial_aid.date_exchange_rate - now
assert abs(date_diff.total_seconds()) < 5
```

that's a stylistic suggestion. you can take or leave it. i would also prefer creating a utility function for this, but it looks like this is the only time we this kind of functionality

noisecapella:
Comments addressed

PR #1732, opened 2016-11-09
Fixed race condition bug with FinancialAidCalculator

### LINKED ISSUE #1717: Error when enrolling in new program
I got this error:

screenshot from 2016-11-08 11-28-20

noisecapella:
Fixes #1717 

#### What's this PR do?
Fixes a rare race condition bug which results in an error message (see screenshot in #1717)

I'm not sure, it only happened to me once and I've been having trouble reproducing it. I just enrolled in a new program using the enroll link in the program selector.

### DIFF
--- a/static/js/containers/FinancialAidCalculator.js
+++ b/static/js/containers/FinancialAidCalculator.js
@@ -134,6 +134,10 @@ const FinancialAidCalculator = ({
   programs,
 }: CalculatorProps) => {
   let program = programs.find(prog => prog.id === id);
+  if (!program) {
+    return null;
+  }
+
   let minPossibleCost, maxPossibleCost;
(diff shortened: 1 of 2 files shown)


aliceriot:
Could you explain what leads to the race condition?

noisecapella:
I'm not sure exactly. There's a `RECEIVE_ADD_PROGRAM_ENROLLMENT_SUCCESS` action which both affects the `programs` reducer and the `currentProgramEnrollment` reducer, so it seems like they should change at the same time. It seems like the program id in `currentProgramEnrollment` doesn't exist in `programs` at the time that the component is rendered.

I'm not sure how to reproduce it, it happens intermittently for me. I just delete an enrollment, add it back again and occasionally I get the error message

PR #1733, opened 2016-11-10
Various style changes

roberthouse54:
#### What's this PR do?
- Changes the button color to be more inline with the MIT red.
- Tightens up the spacing in the learner search, the user chip hover.
- Adds a hover state for the learner search rows
- changes the color of the grade histogram
- changes the color of the "Clear Filters" background
- Changes the color of the "Email Learners" button and add a hover state
- Changes the background color of the sort by dropdown
- Changes the color of the financial aid box

### How to Test

Especially see if there are any buttons in the old color, or hover state color problems.

### DIFF
--- a/static/scss/navbar.scss
+++ b/static/scss/navbar.scss
@@ -62,7 +62,7 @@
       }
 
       .name {
-        font-weight: 500;
+        font-weight: 400;
       }
 
       .program-selector:last-child, .name:last-child {
@@ -152,7 +152,7 @@
 
       @include breakpoint(desktop) {
         img {
-          height: 30px;
+          height: 27px;
           margin-right: 25px;
         }
       }
@@ -172,7 +172,7 @@
 
     .micromasters-header {
       justify-content: space-between;
-      height: 68px;
+      height: 60px;
 
       @include breakpoint(desktop) {
         padding: 0 40px 0 40px;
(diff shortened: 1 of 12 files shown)


pdpinch:
Rob, can you break this down into smaller PRs? 8 changes spread across 12 files are going to slow the review down. 

If you need some help with the git part, I can do it or you can grab someone else.

aliceriot:
Looking this over I think it's manageable as a single PR

static/scss/user-chip.scss
```diff
@@ -70,3 +71,8 @@
     display: block;
   }
 }
+
+.learner-result:hover {
+  background: $bg-light-gray;
+}
```
aliceriot: could you move this into `search-page.scss`, where the rest of the styling for `.learner-result` is?

static/scss/navbar.scss
```diff
@@ -285,3 +282,7 @@ $header_font_color: white;
   text-decoration: none;
   font-weight: 300;
 }
+
+.mdl-layout-title > a:hover, .mdl-layout-title > a:focus {
+  color: white;
+}
```
aliceriot: what's this change for? just curious.

roberthouse54: This keeps the "MITx Micromasters" link in the navbar from changing to the default hover color.

PR #1740, opened 2016-11-10
Fixed dashboard API course status regression

### LINKED ISSUE #1697: Dashboard API ignores past course runs (regression)
A regression was introduced in our performance optimization efforts. The dashboard API no longer returns information about past course runs. Any course runs that a user has passed/failed, for example, are no longer retrieved.

gsidebo:

Fixes #1697

#### What's this PR do?

Fixes the dashboard API regression that prevents past course runs from being evaluated

dashboard/views.py and dashboard/api.py

Make sure you have the standard seed data loaded via the management command. Run the following management commands to get a course with past runs:

```
manage.py alter_data --action=set_course_to_failed --username=YOUR_USERNAME --program-title='Analog' --course-title='100'
manage.py alter_data --action=add_past_failed_run --username=YOUR_USERNAME --program-title='Analog' --course-title='100' --add-if-exists
```

You can also use `--email=` instead of `--username=` to match your desired user. You can run `alter_data --help` and `alter_data --usage` for more details.

Check the dashboard for the 'Analog Learning' program and make sure it's showing you that the 100-level course is failed with another past failed run before it.


dashboard/views_test.py
```diff
@@ -93,13 +67,10 @@ def test_anonym_access(self):
...
         """Test for GET"""
-        res = self.get_with_mocked_refresh_backends()
-
-        assert res.status_code == status.HTTP_200_OK
-        program_ids = [program['id'] for program in res.data]
-        # not live and not enrolled programs are missing
-        assert program_ids == [self.program_1.pk, self.program_2.pk]
-        assert Program.objects.count() == 4
+        with patch('backends.utils.refresh_user_token', autospec=True), \
+                patch('dashboard.api.get_user_program_info', autospec=True):
```
noisecapella: Please:
 - assert the arguments passed into each of these mocked functions
 - provide appropriate return values
 - assert the value of res.data

It is really important that we test the dashboard API well to keep regressions like this one from happening again.

dashboard/api.py
```diff
@@ -121,6 +122,35 @@ def __repr__(self):
         )
 
 
+def get_user_program_info(user, edx_client):
+    """
+    Provides a detailed serialization all of a User's enrolled Programs with enrollment/grade info
```
noisecapella: Can you add args and a return value to the docstring?

noisecapella:
Can you add a unit test covering the steps in the manual testing section of this PR (or some subset that also covers the issue). We should have a unit test that would fail if this PR was not merged, so that we can make sure this regression does not happen again.

micromasters/utils.py
```diff
@@ -100,3 +100,10 @@ def get_field_names(model):
     return [
         field.name for field in model._meta.get_fields() if not field.auto_created  # pylint: disable=protected-access
     ]
+
+
+def first_matching_item(iterable, predicate):
```
noisecapella: Can you add tests for this function?

dashboard/api.py
```diff
@@ -121,6 +122,42 @@ def __repr__(self):
...
+
+    Args:
+        user (User): A User
+        edx_client (EdxClient): An EdxClient
```
noisecapella: This type should be `EdxApi`

dashboard/api_test.py
```diff
@@ -735,6 +735,68 @@ def mocked_get_status_for_courserun(run, enrollments):  # pylint: disable=unused
...
 
+class UserProgramInfoIntegrationTest(ESTestCase):
+    """Integration tests for get_user_program_info"""
+    @classmethod
+    def setUpTestData(cls):
+        super(UserProgramInfoIntegrationTest, cls).setUpTestData()
+        cls.user = UserFactory()
+        # create the programs
+        cls.program1 = ProgramFactory.create(full=True, live=True)
+        cls.program2 = ProgramFactory.create(full=True, live=True, financial_aid_availability=True)
```
noisecapella: Can you rename the variable to indicate that it's a financial aid program?

courses/factories.py
```diff
@@ -21,6 +20,25 @@ class ProgramFactory(DjangoModelFactory):
...
+        program = model_class(*args, **kwargs)
+        program.save()
+        if full_create:
+            course = CourseFactory.create(program=program)
+            course_run = CourseRunFactory.create(course=course)
+            from ecommerce.factories import CoursePriceFactory
+            course_price = CoursePriceFactory.create(course_run=course_run, is_valid=True)
+            if program.financial_aid_availability:
+                from financialaid.factories import TierProgramFactory
+                TierProgramFactory.create(
```
gsidebo: 👍

dashboard/api.py
gsidebo: whoopsiedaisy

dashboard/api_test.py
gsidebo: 👍

gsidebo:
@noisecapella should be all set

dashboard/api_test.py
```diff
@@ -735,6 +735,68 @@ def mocked_get_status_for_courserun(run, enrollments):  # pylint: disable=unused
...
+            }
+            assert is_subset_dict(expected, result[i])
+
+    def test_past_course_runs(self):
+        """Test that past course runs are returned in the API results"""
+        # Set a course run to be failed
+        program = self.program_non_fin_aid
+        course = program.course_set.first()
+        failed_course_run = \
+            set_course_to_failed(user=self.user, course=course)  # pylint: disable=unexpected-keyword-arg
```
noisecapella: Why is this pylint warning being disabled?

noisecapella:
Can you remove `first_matching_item` in `seed_data/utils.py` since it's now duplicated by the one in `micromasters/utils.py`?

dashboard/api_test.py
gsidebo: that method (and several others in the module) has decorator that takes a course kwarg, fetches a 'default' course run, scraps the course kwarg, and inserts the course_run kwarg. the kwarg setup is a bit of an anti-pattern - the alter_data commands were written with a little less attention to standards. anyway, it works for now. i'm going to open up and issue to move to some more formal integration tests for the dashboard API. part of that issue will entail cleaning this up this code and making sure it's used properly in test cases

dashboard/api_test.py
```diff
@@ -22,8 +22,10 @@
 )
 from dashboard.utils import MMTrack
 from micromasters.factories import UserFactory
-from micromasters.utils import load_json_from_file
+from micromasters.utils import load_json_from_file, is_subset_dict
 from search.base import ESTestCase
+from seed_data.lib import ensure_cached_edx_data_for_user, set_course_run_past
+from seed_data.management.commands.alter_data import set_course_to_failed, add_past_failed_run
```
gsidebo: see [this comment](https://github.com/mitodl/micromasters/pull/1740#discussion_r88065190)

gsidebo:
done

micromasters/utils.py
```diff
@@ -100,3 +100,38 @@ def get_field_names(model):
...
+        dict_to_test (dict): The subset dictionary
+        master_dict (dict): The dictionary to test against
+    Returns:
+        bool: Whether or not the first dictionary is a subset of the second
+    """
+    result = False
+    try:
+        for pkey, pvalue in dict_to_test.items():
+            if isinstance(pvalue, dict):
+                result = is_subset_dict(pvalue, master_dict[pkey])
```
noisecapella: We don't have a test for the case of nested dicts, and why would we want to do this anyway? I feel like a simple for loop comparing values using `==` would be the way to go.

micromasters/utils.py
```diff
@@ -100,3 +100,38 @@ def get_field_names(model):
...
+                if not result:
+                    break
+            else:
+                assert master_dict[pkey] == pvalue
```
noisecapella: Why are we using `assert` here?

dashboard/api_test.py
```diff
@@ -735,6 +735,68 @@ def mocked_get_status_for_courserun(run, enrollments):  # pylint: disable=unused
...
+
+        assert len(result) == 2
+        for i in range(2):
+            expected = {
+                "id": self.expected_programs[i].id,
+                "description": self.expected_programs[i].description,
+                "title": self.expected_programs[i].title,
+                "financial_aid_availability": self.expected_programs[i].financial_aid_availability,
+            }
+            assert is_subset_dict(expected, result[i])
```
gsidebo: yes, that would be a nice thing to have, and i'm going to consider it out of scope for this PR. i will link the issue for dashboard integration testing shortly

micromasters/utils.py
gsidebo: i have this snippet hanging around from a previous job. it works, so i added it to the codebase. i don't see the drawback of having this additional functionality

gsidebo:
@noisecapella fyi https://github.com/mitodl/micromasters/issues/1767

micromasters/utils.py
noisecapella: Can we add some more unit tests for this? We don't have any coverage of these parts of the function:

screenshot from 2016-11-15 13-49-46

noisecapella: IMO the drawback to applying this to nested dicts is that it complicates the function and it's not described in the documentation. I'm fine to leave it if the documentation is updated and the other conditions are tested.

micromasters/utils.py
noisecapella: `assert` means something different when run via `pytest` and when run with regular python. It seems like if this comparison failed, wouldn't you just want to return false?

gsidebo:
@noisecapella should be all set

micromasters/utils_test.py
```diff
@@ -146,3 +148,40 @@ def test_get_field_names(self):
...
+        """
+        iterable = [1, 2, 3, 4, 5]
+        first_matching = first_matching_item(iterable, lambda item: item == 1)
+        second_matching = first_matching_item(iterable, lambda item: item == 5)
+        third_matching = first_matching_item(iterable, lambda item: item == 10)
+        assert first_matching == 1
+        assert second_matching == 5
+        assert third_matching is None
+
+    def is_subset_dict(self):
```
noisecapella: This needs to be prefixed with `test_` to run

Issue #1596, opened 2016-10-28
Add environment variable to affect Elasticsearch pagination size


noisecapella:
The environment variable should default to `10` which is Elasticsearch's default for `size`, and we should always set a `size` when we construct the query on the server

amir-qayyum-khan:
@noisecapella are you taking abt this size?

![screen shot 2016-11-09 at 5 49 56 pm](https://cloud.githubusercontent.com/assets/10431250/20138908/fe5f8d72-a6a4-11e6-8ace-9feea87d684d.png)

pdpinch:
That looks right to me -- we want to make the number of results per page configurable so it's easier to test the page navigation.

amir-qayyum-khan:
cool, it means from now on word this size will be maintain on server side instead on client side?

pdpinch:
That's right.

PR #1743, opened 2016-11-11
Added environment variable to affect Elasticsearch pagination size

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1596

#### config:
set ``ELASTICSEARCH_DEFAULT_PAGE_SIZE=10`` in .env file.

- put elastic search page size in env variable. So that we can user same data across app and tests.
- fixed minor UI issue on search header. 

screen shot 2016-11-11 at 3 23 08 pm


noisecapella:
What part of this PR applies the page size setting to Elasticsearch?

noisecapella:
Sorry, to be more specific, can you change the back end to add the size from the environment variable, so that unit tests can easily change this value?

amir-qayyum-khan:
ok

amir-qayyum-khan:
Done @noisecapella

search/api.py
```diff
@@ -72,6 +72,7 @@ def create_search_obj(user, search_param_dict=None):
     # the following filter should come first because the sequence matters in applying them
     search_obj = search_obj.filter(create_program_limit_query(user))
     if search_param_dict is not None:
+        search_param_dict['size'] = settings.DEFAULT_ES_PAGE_SIZE
```
noisecapella: What if `search_param_dict` is `None`? We still need to insert the size in that case

micromasters/settings.py
```diff
@@ -487,6 +487,7 @@ def get_var(name, default):
 CELERY_TIMEZONE = 'UTC'
 
 # Elasticsearch
+DEFAULT_ES_PAGE_SIZE = get_var('DEFAULT_ES_PAGE_SIZE', 10)
```
noisecapella: Can you rename this to `ELASTICSEARCH_DEFAULT_PAGE_SIZE` to be consistent with the other `ELASTICSEARCH` environment variables?

noisecapella:
Can you use `override_settings` on the classes `SearchTests` in `search/views_test.py` and `SearchAPITests` in `search/api_test.py`? We should set a high value for that environment variable so that all search results comes through in the first page

PR #1746, opened 2016-11-14
Style the react select field to look like material ui field

roberthouse54:
#### What's this PR do?

This PR styles the react select field to look more-or-less like the material UI fields. So I changed the react select label, the spacing and margins of the field, the color of the bottom-border. I also removed the bold style for the program selector on page 1 of the sign up, and removed the "x" icon to clear the selection. This last one is pending discussion with Ferdi who said he thought having the x was useful.

image

### DIFF
--- a/static/scss/select-field.scss
+++ b/static/scss/select-field.scss
@@ -23,3 +23,36 @@
     border-top-left-radius: 4px!important;
   }
 }
+
+/* React Select styling */
+
+.Select--single > .Select-control {
+
+  border-radius: 0;
+  border: none;
+  border-bottom: 1px solid #dcdcdc;
+
+  .Select-value {
+    padding-left: 2px;
+    top: 4px;
+  }
+  .Select-input {
+    padding-left: 3px;
+  }
+}
+
+.Select-placeholder {
+  top: 4px;
+  padding-left: 2px;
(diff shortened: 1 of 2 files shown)


aliceriot:
looks good! looking at code

static/scss/select-field.scss
```diff
@@ -23,3 +23,41 @@
...
+}
+
+.select-field {
+  .select-label {
+    font-size: 12px;
+    font-weight: 400;
+    color: rgba(0,0,0,.5);
+    position: relative;
+    top: 4px;
+    left: 2px;
```
aliceriot: can you move these rules up to the `.select-label` block above?

static/scss/select-field.scss
```diff
@@ -23,3 +23,41 @@
     border-top-left-radius: 4px!important;
   }
 }
+
+/* React Select styling */
```
aliceriot: this comment is a bit redundant, this whole file is for styling react-select

roberthouse54: Removed.

static/scss/select-field.scss
roberthouse54: Done.

PR #1747, opened 2016-11-14
add error display to financial aid calculator

### LINKED ISSUE #1285: Submitting cost calculation with error response has no UX change
Spawned from #1282 for handling after MVP

The most likely scenario is submitting a not-supported currency.
#1275 handles general syncing of frontend-supported and backend-supported currency codes, but there may be a rare case where a learner may end up submitting a code that isn't supported.

The learner should get some sort of message when errors come up when they click calculate cost, perhaps directing them to the helpdesk or telling them to refresh the page and try again.

aliceriot:

closes #1285 

#### What's this PR do?

This adds a little error message display to the financial aid calculator!

errrorororor

The easiest way to generate error messages is to not have have the currency conversion set-up on your computer. Then submitting the form with anything other than USD will result in a 400 error.

### DIFF
--- a/static/js/containers/FinancialAidCalculator.js
+++ b/static/js/containers/FinancialAidCalculator.js
@@ -1,3 +1,4 @@
+/* global SETTINGS:false */
 import React from 'react';
 import Dialog from 'material-ui/Dialog';
 import { connect } from 'react-redux';
@@ -5,6 +6,8 @@ import R from 'ramda';
 import Button from 'react-mdl/lib/Button';
 import TextField from 'material-ui/TextField';
 import Checkbox from 'react-mdl/lib/Checkbox';
+import Select from 'react-select';
+import _ from 'lodash';
 
 import {
   updateCalculatorEdit,
@@ -18,7 +21,6 @@ import {
   setConfirmSkipDialogVisibility,
 } from '../actions/ui';
 import { createSimpleActionHelpers } from '../lib/redux';
-import SelectField from '../components/inputs/SelectField';
 import { currencyOptions } from '../lib/currency';
 import { validateFinancialAid } from '../lib/validation/profile';
 import { sanitizeNumberString } from '../lib/validation/date';
@@ -26,18 +28,28 @@ import type { AvailableProgram } from '../flow/enrollmentTypes';
 import type {
   FinancialAidState,
   FinancialAidValidation,
+  FetchError,
 } from '../reducers/financial_aid';
 import type { Program } from '../flow/programTypes';
 import { formatPrice } from '../util/util';
 
+const updateCurrency = R.curry((update, financialAid, selection) => {
+  let _financialAid = R.clone(financialAid);
+  _financialAid.currency = selection ? selection.value : null;
+  update(_financialAid);
+});
+
 const currencySelect = (update, current) => (
-  <SelectField
+  <Select
     options={currencyOptions}
-    keySet={['currency']}
-    profile={current}
-    updateProfile={update}
+    value={current.currency}
+    onChange={updateCurrency(update, current)}
     name="currency"
     id="currency-select"
+    inputProps={{
+      'aria-required': "true",
+      'aria-invalid': _.has(current, ['validation', 'currency'])
+    }}
   />
 );
 
@@ -50,6 +62,9 @@ const salaryUpdate = R.curry((update, current, e) => {
 const salaryField = (update, current) => (
   <TextField
     name="salary"
+    required="true"
+    aria-invalid={_.has(current, ['validation', 'income'])}
+    label="income (yearly)"
     id="user-salary-input"
     className="salary-field"
     value={current.income}
@@ -70,6 +85,8 @@ const checkboxUpdate = (update, current, bool) => {
 const checkBox = (update, current) => (
   <Checkbox
     checked={current.checkBox}
+    required="true"
+    aria-invalid={_.has(current, ['validation', 'checkBox'])}
     label={checkboxText}
     onChange={() => checkboxUpdate(update, current, !current.checkBox)}
   />
@@ -84,6 +101,7 @@ const actionButtons = R.map(({ name, onClick, label}) => (
     { label }
   </Button>
 ));
+
 const calculatorActions = (openSkipDialog, cancel, save) => {
   const buttonManifest = [
     { name: 'cancel', onClick: cancel, label: 'Cancel' },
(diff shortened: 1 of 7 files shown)


Ferdi:
Not sure this is ideal UX.  If it's not supported , maybe we remove from the selection list ?

aliceriot:
This is a general error display, that is simply a particular example of something that could generate an error.

aliceriot:
i.e. this pr doesn't have anything to do with unsupported currencies in particular.

pdpinch:
@roberthouse54 can you weigh in here on the error message UX?

In an error situation like this, I'd like to encourage users to contact us, by e-mailing micromasters-support@mit.edu or using the help button at the bottom of the page.

noisecapella:
Code and functionality look good to me. I'll wait on approving until the copy text is sorted out

aliceriot:
@noisecapella I'm also going to (on @singingwolfboy 's suggestion) work on making the calculator accessible here, if that's alright with you.

noisecapella:
Sounds good

roberthouse54:
How about something like this: 

image

aliceriot:
looks good to me!

aliceriot:
@noisecapella ready for review again!

static/js/containers/FinancialAidCalculator.js
```diff
@@ -18,26 +20,35 @@ import {
...
     options={currencyOptions}
-    keySet={['currency']}
-    profile={current}
-    updateProfile={update}
+    value={current.currency}
+    onChange={updateCurrency(update, current)}
     name="currency"
     id="currency-select"
+    inputProps={{
+      'aria-required': "true",
```
singingwolfboy: I believe the [`required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-required) attribute should be used rather than `aria-required`. Most assistive tech should treat it the same, but with `required` you also get to use [the browser's built-in JavaScript form validation APIs](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/Data_form_validation#Validating_forms_using_JavaScript), if you want.

static/js/containers/FinancialAidCalculator.js
```diff
@@ -109,6 +124,16 @@ const validationMessage = (key, errors) => {
   </div>;
 };
 
+const apiError = ({ message, code }: FetchError) => (
+  <div className="api-error">
```
singingwolfboy: Can we use the [`aria-errormessage`](https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage) attribute on the text field to point to this element, when there's an error? That way, screen readers (and other assistive tech) can inform the user _why_ the text field is in an invalid state.

static/js/containers/FinancialAidCalculator_test.js
```diff
@@ -182,6 +195,46 @@ describe('FinancialAidCalculator', () => {
...
+      }).then(() => {
+        assert(
+          addFinancialAidStub.calledWith('1000', 'USD', program.id),
+          'should be called with the right arguments'
+        );
+        assert.equal(
+          document.querySelector('.api-error').textContent,
+          `There was an error (Error 500: an error message). Please contact micromasters-support@mit.edu \
+if you continue to have problems.`
+        );
```
singingwolfboy: Can you also assert on the `aria-invalid` attribute on the input?

static/js/containers/FinancialAidCalculator.js
```diff
@@ -109,6 +124,16 @@ const validationMessage = (key, errors) => {
...
+    {`There was an error (Error ${code}: ${message}). Please contact `}
+    <a href="mailto:micromasters-support@mit.edu">
```
noisecapella: This is actually defined in the django settings as `EMAIL_SUPPORT`. Can you add it to `SETTINGS` and swap it in here?

pdpinch:
Rob's giving some more thought to the UX with these kind of error messages, so it's not worth sweating the details too much right now.

static/js/containers/FinancialAidCalculator.js
aliceriot: This isn't for the text field, but for when an error comes back from the API. Should we still do what you suggested?

singingwolfboy: Hmm. This isn't something that the user can take action on to resolve, but it's still something that they should be made aware of. This API error will prevent them from successfully submitting the form, right? If that's the case, maybe we should set `aria-invalid` and `aria-errormessage` on the `<form>` element, to indicate that the form can't be submitted but it's not due to any particular field being invalid.

I'll be honest, I don't actually know the best way to handle this situation. Maybe it's not worth handling right now. But the basic idea is, if this error message is important enough to show it to a sighted user, then it should also be important enough to be presented to non-sighted users as well.

aliceriot: We're not actually using a `<form>` tag here either, not sure if that changes things.

singingwolfboy: Hmm, that seems like a problem as well. I believe that HTML form elements are supposed to always be enclosed in a `<form>` tag, otherwise the HTML is invalid. But that's another problem for another day. Sounds like this is way out of scope for this pull request, so we can just leave it as-is for now.

noisecapella:
@pdpinch are you referring to the `EMAIL_SUPPORT` thing?

static/js/containers/FinancialAidCalculator.js
aliceriot: haha, definitely out of scope. We don't use `<form>` tags anywhere on the site, except perhaps on the homepage for the mailchimp signup dialog

static/js/containers/FinancialAidCalculator.js
aliceriot: It's already available in settings, looks like (git blame says you added it :D )

aliceriot:
ready for review again!

noisecapella:
If the browser height is short the error message overlaps the buttons:

screenshot from 2016-11-15 15-54-05

Not sure if this is an edge case or not. In any case doesn't the content just scroll by default?

noisecapella:
Functionality looks good, feel free to merge if you consider the above issue something for another PR :+1:

aliceriot:
I just fixed that, there's a prop (`autoScrollBodyContent`) which sets `overflow: scroll` on the body of the dialog.

aliceriot:
bah we're in a pickle on that one. We need `overflow: visible` on the body for the currency drop-down to work, but we need `overflow: auto` or `overflow: scroll` to have a decent display on a small screen.

@noisecapella what do you think we should do?

singingwolfboy:
@aliceriot: you could use `@media` queries to change from one to the other depending on the screen size.

noisecapella:
Good question, I guess we should just have `overflow: visible`. I'm not sure a media query would be appropriate since we may run into the currency dropdown being hidden there too

aliceriot:
yea, those things are mutually exclusive, we can have the dropdown be visible or we can have the thing scroll. I'm going to merge as-is and we can figure out a solution / how much we care about it later.

PR #1748, opened 2016-11-14
Final grade serialization

### LINKED ISSUE #1632: Program average grade in search results will not work for FA programs
The calculation of grades in FA programs is different from non-FA programs, and the serialization of average grades for search results was never updated to account for this. Code can be found in dashboard/serializers.py

giocalitri:
fixes #1632

#### What's this PR do?
Adds support for the financial grade final grade serialization for the search.
It also refactors many factories because of specific needs for this PR

`dashboard/serializers.py`

You need a course run in a financial aid program where the user is enrolled mmtrack (meaning is enrolled and paid for the course run).
Once this is done, the easiest way is to modify the course run CachedCurrentGrade object for the user to something like 
```
{
		"course_key": <course run edx key>,
		"passed": True,
		"percent": 0.77,
		"letter_grade": "Pass",
		"username": <your user's username>
}
```
Then recreate the index and you should be able to see the final grade in the search result.

the `seed_db` command does not help much for the test: there is another issue open to add support for financial aid programs

### DIFF
--- a/dashboard/factories.py
+++ b/dashboard/factories.py
@@ -5,7 +5,7 @@ from datetime import datetime, timedelta
 from random import randint
 import pytz
 
-from factory import SubFactory
+from factory import SubFactory, LazyAttribute
 from factory.django import DjangoModelFactory
 from factory.fuzzy import (
     FuzzyAttribute,
@@ -15,8 +15,9 @@ import faker
 
 from dashboard.models import (
     CachedCertificate,
+    CachedCurrentGrade,
     CachedEnrollment,
-    ProgramEnrollment
+    ProgramEnrollment,
 )
 from courses.factories import (
     CourseRunFactory,
@@ -32,10 +33,12 @@ FAKE = faker.Factory.create()
 class CachedCertificateFactory(DjangoModelFactory):
     """Factory for Certificate"""
     user = SubFactory(UserFactory)
-    data = FuzzyAttribute(lambda: dict(
-        grade=randint(60, 100), **{'key{}'.format(i): FAKE.text() for i in range(3)}
-    ))
     course_run = SubFactory(CourseRunFactory)
+    data = LazyAttribute(lambda x: {
+        "certificate_type": "verified",
+        "grade": randint(60, 100)/100.0,
+        "course_id": x.course_run.edx_course_key,
+    })
     # certificates expire after 6 hours, this generates a last request between 6:15 hours ago and now
     last_request = FuzzyDateTime(datetime.now(tz=pytz.utc) - timedelta(hours=6, minutes=15))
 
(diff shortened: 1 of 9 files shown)


dashboard/serializers.py
```diff
@@ -15,13 +20,15 @@ class UserProgramSearchSerializer:
...
+    def calculate_final_grade_average(mmtrack):
         """
-        Calculates an average grade (integer) from a list of certificates with <1 decimal grades
+        Calculates an average grade (integer) from the program final grades
         """
-        return None if len(certificates) == 0 else round(
-            (sum((Decimal(certificate['grade']) for certificate in certificates)) / len(certificates)) * 100
-        )
+        if not mmtrack.course_ids:
+            return
```
gsidebo: looks like this check is unnecessary. if `course_ids` is blank in this object, `get_all_final_grades` will just return an empty dict. you should just add an `else` clause below and return None there

dashboard/serializers_test.py
```diff
@@ -28,13 +45,23 @@ class UserProgramSearchSerializerTests(TestCase):
...
+        fake = faker.Factory.create()
         course = CourseFactory.create(program=program)
-        course_runs = [CourseRunFactory.create(course=course) for _ in range(num_course_runs)]
+        course_runs = [
+            CourseRunFactory.create(
+                course=course,
+                enrollment_start=fake.date_time_this_month(before_now=True, after_now=False, tzinfo=pytz.utc),
+                start_date=fake.date_time_this_month(before_now=True, after_now=False, tzinfo=pytz.utc),
+                enrollment_end=fake.date_time_this_month(before_now=True, after_now=False, tzinfo=pytz.utc),
+                end_date=fake.date_time_this_year(before_now=True, after_now=False, tzinfo=pytz.utc),
```
gsidebo: can you set these params (`before_now=True, after_now=False, tzinfo=pytz.utc`) to a dict, and invoke `fake.date_time_this_year` by passing in that dict? something like `fake.date_time_this_month(**current_date_params)`

financialaid/api_test.py
```diff
@@ -57,6 +57,29 @@ def create_program():
     return program
 
 
+def create_tier_programs(program):
```
gsidebo: this function is always called immediately after `financialaid.api_test.create_program`. that makes perfect sense bc it's another necessary step for creating a valid program. i think you should either get rid of this method and move the code inside `create_program`, or just invoke this method at the end of `create_program`

dashboard/serializers_test.py
```diff
@@ -44,47 +71,107 @@ def _generate_cached_enrollments(user, program, num_course_runs=1, data=None):
...
+                CachedCurrentGradeFactory.create(
+                    user=self.user,
+                    course_run=enrollment.course_run,
+                    data={
+                        "passed": True,
+                        "percent": self.current_grades_vals[i],
+                        "course_key": enrollment.course_run.edx_course_key,
+                    }
+                )
+            )
```
gsidebo: you have `setUpTestData` creating data only for `test_full_program_user_serialization`, then for this method you perform similar setup steps in the body of the test case function. a cleaner approach might be to create everything up to the `Program`s in the `setUpTestData` method (and anything else that all test cases need), then have separate helper methods for creating the enrollments/grades/etc for the separate test cases. i suggest creating the `Program`s in  `setUpTestData` b/c it will ensure that they will be deleted after every test case, and that will cascade down to delete all the associated edx data

dashboard/factories.py
```diff
@@ -32,22 +33,47 @@
...
     """Factory for Certificate"""
     user = SubFactory(UserFactory)
-    data = FuzzyAttribute(lambda: dict(
-        grade=randint(60, 100), **{'key{}'.format(i): FAKE.text() for i in range(3)}
-    ))
     course_run = SubFactory(CourseRunFactory)
+    data = LazyAttribute(lambda x: {
+        "certificate_type": "verified",
+        "grade": randint(60, 100)/100.0,
+        "course_key": x.course_run.edx_course_key,
```
gsidebo: in `edx_api.certificates.models.Certificate`, this key is `course_id` not `course_key`

gsidebo:
just wondering: did you check out the impact on the number of DB queries executed?

giocalitri:
@gsidebo comments addressed

financialaid/api_test.py
```diff
@@ -35,7 +35,7 @@
 from roles.roles import Staff, Instructor
 
 
-def create_program():
+def create_program(create_tiers=True):
```
gsidebo: since not all tests need these tier programs, i think it makes more sense to split this into `create_program` and `create_program_with_tiers`. tests that only need the program can use `program = create_program()`, and tests that need both can use `program, tier_programs = create_program_with_tiers()`. `create_program_with_tiers` would just call `create_program` to start

gsidebo:
👍

PR #1750, opened 2016-11-14
Moved dashboard API call to DashboardPage

### LINKED ISSUE #1655: Long delay in loading programs in profile form
Steps to reproduce:

1. On an account without a profile, load the profile form

## Expected Behavior

When the form loads and becomes interactive, the programs select list has two (or more) options. 

## Actual Behavior

When the form loads and the programs select is interactive, there are no options available. 

It takes a few more seconds after the form loads for the programs drop down to be populated. Since it's the first field in the form, users are confused why there are no choices from the drop-down.

noisecapella:
Fixes #1655

#### What's this PR do?
Delays loading the dashboard API until the user looks at the dashboard page. The user will load the dashboard API when they click next and enroll in a program, but this shouldn't block anything on their end.

Load `/profile` and verify that no request to `/api/v0/dashboard/` is loaded and that the user still sees programs they can enroll in.

Also, go to `/settings` and refresh the browser. Verify that `/api/v0/dashboard/` is not loaded. Then click the logo on the upper left to direct you to the dashboard. Verify that the dashboard loaded properly.

### DIFF
--- a/static/js/components/ErrorMessage_test.js
+++ b/static/js/components/ErrorMessage_test.js
@@ -5,6 +5,8 @@ import { assert } from 'chai';
 import _ from 'lodash';
 
 import {
+  REQUEST_DASHBOARD,
+  REQUEST_COURSE_PRICES,
   RECEIVE_DASHBOARD_FAILURE,
   RECEIVE_DASHBOARD_SUCCESS,
   RECEIVE_COURSE_PRICES_SUCCESS,
@@ -20,7 +22,10 @@ import {
   RECEIVE_PATCH_USER_PROFILE_FAILURE,
   CLEAR_PROFILE_EDIT,
 } from '../actions/profile';
-import { RECEIVE_GET_PROGRAM_ENROLLMENTS_SUCCESS } from '../actions/programs';
+import {
+  REQUEST_GET_PROGRAM_ENROLLMENTS,
+  RECEIVE_GET_PROGRAM_ENROLLMENTS_SUCCESS,
+} from '../actions/programs';
 import {
   DASHBOARD_RESPONSE,
   ERROR_RESPONSE,
@@ -74,7 +79,6 @@ describe("ErrorMessage", () => {
 
   describe('showing errors on pages', () => {
     let renderComponent, helper, patchUserProfileStub, listenForActions;
-    let dashboardErrorActions;
 
     const confirmErrorMessage = (div, codeMessage, extraInfo = '') => {
       let alert = div.querySelector('.alert-message');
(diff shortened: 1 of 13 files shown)


static/js/containers/DashboardPage.js
```diff
@@ -126,6 +135,26 @@ class DashboardPage extends React.Component {
...
 
+  updateRequirements = () => {
+    this.fetchDashboard();
+    this.fetchCoursePrices();
+    this.handleOrderStatus();
+  };
+
+  fetchDashboard() {
+    const { dashboard, dispatch } = this.props;
+    if (dashboard.fetchStatus === undefined) {
```
noisecapella: It's not possible for globals to be replaced in that matter, only locals: http://www.2ality.com/2013/04/check-undefined.html

noisecapella: TIL too, I was going to argue for it for stylistic reasons and then I found evidence to support my preference :)

aliceriot: As far as I can tell this is exactly the check that `_.isUndefined` makes anyhow: https://github.com/lodash/lodash/blob/master/lodash.js#L12197

static/js/containers/App_test.js
```diff
@@ -38,29 +41,37 @@ import IntegrationTestHelper from '../util/integration_test_helper';
...
+    editProfileActions = successActions.concat([
       START_PROFILE_EDIT,
       START_PROFILE_EDIT,
       UPDATE_PROFILE_VALIDATION,
       SET_PROFILE_STEP,
-    ];
+    ]);
+    redirectActions = successActions.concat([
+      START_PROFILE_EDIT,
+    ]);
```
gsidebo: looks like you're assigning expected actions in the `describe` handler in some test files, and inside the `beforeEach` handler in others. it doesn't look like you're actually changing these lists anywhere - when you need to add to the list, you are setting a different variable, which i think is the right approach. can we make this consistent?

static/js/components/ErrorMessage_test.js
```diff
@@ -111,10 +108,32 @@ describe("ErrorMessage", () => {
...
+      const dashboardSuccessActions = [
+        REQUEST_DASHBOARD,
+        RECEIVE_DASHBOARD_SUCCESS,
+        REQUEST_COURSE_PRICES,
+        RECEIVE_COURSE_PRICES_SUCCESS,
+        REQUEST_GET_USER_PROFILE,
+        RECEIVE_GET_USER_PROFILE_SUCCESS,
+        REQUEST_GET_PROGRAM_ENROLLMENTS,
+        RECEIVE_GET_PROGRAM_ENROLLMENTS_SUCCESS,
+      ];
```
gsidebo: this is defined in 3 different places. can you just defined this and export it in `DashboardPage_test.js` or something like that?

gsidebo:
functionality looks great!

noisecapella:
Comments addressed

PR #1754, opened 2016-11-14
Turn profile links into buttons

### LINKED ISSUE #1705: "Add employment" and "Add degree" links on profile page can't be tabbed to
The two links don't have href tags so browsers/screen readers won't recognize them as actionable.

Per #1222 we should either add an `href` attribute or switch to `<button>`.

singingwolfboy:
Fixes #1705. Right now, this is the simplest possible fix: changing the `<a>` tags into `<button>` tags. It's now much clearer that these are actions that the user can take, but it's not particularly pretty. @roberthouse54, do you want to style this up nicely?

profile-with-buttons


roberthouse54:
Yes, I will do that as a separate PR. Can you create an issue for styling the buttons and assign it to me?

singingwolfboy:
@roberthouse54: you got it! Issue #1757

Issue #1762, opened 2016-11-15
Sign up: make avatar required


pdpinch:
## Acceptance criteria

- [ ] Add redesigned field and drop-zone to first page of sign up form
- [ ] validate that an image has been uploaded
- [ ] remove the form and button from 2nd & 3rd pages of sign up form

pdpinch:
@roberthouse54 can you add a mock-up to this when its ready?

pdpinch:
@aliceriot can the validation on an image field like this work like other fields, or do we need a different approach? different UX?

aliceriot:
I think this requires a pretty different UX - if a photo is required we should make that obvious and clear (like 'Step One: Add a photo' or something). So I think we need a clearer idea of what the UX is before we talk about implementation details.

roberthouse54:
![upload_photo_in_signup 2x](https://cloud.githubusercontent.com/assets/20047260/20367059/df18388e-ac1b-11e6-9160-f8e97814d5dc.png)

aliceriot:
@roberthouse54 what's the UX look like for that drop area? When the user adds a photo should a dialog open with the cropper, or should the cropper live in the card?

roberthouse54:
![image](https://cloud.githubusercontent.com/assets/20047260/20499252/aee2ce22-affe-11e6-830e-9c4b1c72c8ad.png)

roberthouse54:
So, as shown above, I'd like to just have a simple link here, which opens the current "Edit Profile Photo" modal and from there the UX is identical to what we have currently. After a profile photo has been uploaded, we should show it here, and switch the link to read "Edit my profile photo".

PR #1766, opened 2016-11-15
It's OK to have a blank profile image

singingwolfboy:
Without `blank=True`, the Django admin will not allow you to alter a user profile without uploading an image. Uploading an image should not be necessary.


singingwolfboy:
Does this conflict with #1762?

pdpinch:
This probably will conflict with #1762. Did you have a reason for needing it?

singingwolfboy:
No, I discovered that the profile image was required while doing a review of another pull request. I assumed that it should be optional, but that was simply an assumption. Sounds like that assumption was wrong, and this _should_ be required, so I'll close this pull request.

PR #1772, opened 2016-11-15
Turn on more sass-lint rules, add autoprefixer

### LINKED ISSUE #1764: update sass-lint config
I'd like to turn on a few more rules that will enforce better CSS practices.

aliceriot:

closes #1764 

#### What's this PR do?

This turns on a bunch more sass-lint rules. I also updated sass-lint and installed the `postcss` loader for webpack, with the [autoprefixer](https://github.com/postcss/autoprefixer) plugin (which automatically adds vendor prefixes).

Basically, make sure adding the sass-lint rules makes sense, and also that I didn't clobber anything on the site in the process.


static/scss/footer.scss
```diff
@@ -22,9 +22,9 @@
     margin: 24px 0;
 
     a {
-      display:inline-block;
-      margin:0 24px 0 0;
-      color:#4c8399;
+      display: inline-block;
+      margin: 0 24px 0 0;
+      color: #5c8399;
```
noisecapella: Is the color change intentional here?

postcss.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  plugins: [
+    require('autoprefixer')({
+      browsers: ['> 5%']
```
noisecapella: When I run the command line tool to list these browsers I only get these four:

    and_chr 53
    and_uc 11
    chrome 53
    ios_saf 10

If we change to 1 % we get this list instead:

    and_chr 53
    and_uc 11
    android 4.4.3-4.4.4
    android 4.4
    chrome 54
    chrome 53
    chrome 49
    edge 14
    firefox 49
    ie 11
    ios_saf 10
    ios_saf 9.3
    op_mini all
    safari 10
    samsung 4

noisecapella:
There are a few places in `homepage.scss` which still use the prefixes, can you remove the extraneous ones there as well?

postcss.config.js
aliceriot: cool, I meant to leave a comment asking what value would be good for this setting. > 1% it is!

static/scss/footer.scss
aliceriot: oh, no, must have been too quick when editing.

aliceriot:
Oh yeah, I'll see how difficult it is to remove `homepage.scss` from the ignore list too.

aliceriot:
I removed `homepage.scss` from the ignore list, deleted a bunch of unused CSS from it, and then fixed all the errors in it.

Also noticed I had goofed up on the financial aid review page, so I fixed that real quick as well.

PR #1779, opened 2016-11-15
Filtered out reset FinancialAid statuses

### LINKED ISSUE #1777: E-Comm: After submitting Fin Aid, "Skip this and Pay full Price" does not work
Steps to recreate:
RC
Enroll in DEDP Program
Fin Aid Calculator
1,000 Indonesian Rupiah
Submit
Status changes to pending-docs w/ with Tier 1 discount applied.
Click on "Skip this & Pay Full Price"
Click on "Pay Full Price" on the "are you sure pop-up".
Expecting to see the status change from Pending-docs and the full price displayed.
Instead, nothing happens.

Status remains as "pending docs"

noisecapella:
Fixes #1777

#### What's this PR do?
Adds a filter to prevent reset `FinancialAid`s from being seen in `FinancialAidSkipView`

Create a financial aid, set it to `RESET` in the Django admin. Then attempt to skip financial aid in the UI.

### DIFF
--- a/financialaid/views.py
+++ b/financialaid/views.py
@@ -1,6 +1,7 @@
 """
 Views for financialaid
 """
+import datetime
 import json
 from functools import reduce
 
@@ -17,16 +18,20 @@ from rest_framework.generics import (
 )
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
+from rest_framework.status import HTTP_200_OK
 from rest_framework.views import APIView
 from rolepermissions.verifications import has_object_permission
 
 from courses.models import Program
 from dashboard.models import ProgramEnrollment
+from financialaid.api import (
+    get_formatted_course_price,
+    get_no_discount_tier_program,
+)
 from financialaid.constants import (
     FinancialAidJustification,
     FinancialAidStatus
 )
-from financialaid.api import get_formatted_course_price
 from financialaid.models import (
     FinancialAid,
     TierProgram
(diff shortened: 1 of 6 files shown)


giocalitri:
minor comment

noisecapella:
Ready for review again

noisecapella:
I removed the FinancialAidSkipSerializer and moved the functionality into the `update` method of the view. Hopefully that will make it a little clearer to read

financialaid/views.py
```diff
@@ -74,30 +75,43 @@ class FinancialAidSkipView(UpdateAPIView):
...
+                country_of_residence=user.profile.country,
+            )
+
+        if financialaid.status in FinancialAidStatus.TERMINAL_STATUSES:
+            raise ValidationError("Financial aid application cannot be skipped once it's been approved or skipped.")
+
+        financialaid.tier_program = get_no_discount_tier_program(program.id)
+        financialaid.status = FinancialAidStatus.SKIPPED
+        financialaid.save_and_log(user)
+        return Response(status=HTTP_200_OK)
```
giocalitri: theoretically we should return the serialized object here for the patch/put request.

giocalitri: but I leave this to you

micromasters/utils_test.py
```diff
@@ -10,6 +11,7 @@
     RequestFactory,
     TestCase,
 )
+import pytz
```
giocalitri: nit: `import` statement should go before `from ... import` statements

giocalitri:
two small comments

financialaid/views.py
noisecapella: I'd prefer not to since this isn't actually a `FinancialAid` endpoint but instead something specific to skip requests. If we're going to make it more RESTful we should just combine all the APIs into one first

micromasters/utils_test.py
noisecapella: Done

giocalitri:
Functionality looks good
👍 after tests pass

PR #1780, opened 2016-11-16
pass validator into PrivacyForm

aliceriot:
The privacy form has to have a `validator` on `this.props` in order for the `boundRadioGroupField` to pull it out.


static/js/containers/SettingsPage.js
```diff
@@ -39,7 +39,7 @@ class SettingsPage extends ProfileFormContainer {
       <Loader loaded={loaded}>
         <div className="single-column privacy-form">
           <h4 className="privacy-form-heading">Settings</h4>
-          <PrivacyForm {...props} />
+          <PrivacyForm {...props} validator={privacyValidation} />
```
aliceriot: oh ya will do

PR #1784, opened 2016-11-16
fix gradient syntax error warning

aliceriot:

Should have caught this in #1772 

#### What's this PR do?

Fixes a spot where we were using the old syntax for `linear-gradient`.

Just check there's no warning from `postcss` showing up in the JS console.


aliceriot:
@roberthouse54 I'm not super familiar with gradient stuff, does this change make sense to you? It seems to look the same but ¯\\\_(ツ)_/¯

PR #1786, opened 2016-11-16
CSS fallbacks for home page

### LINKED ISSUE #1559: Programs are laid out oddly on the homepage
On micromasters-rc
screenshot from 2016-10-25 16-51-15

singingwolfboy:
Fixes #1559. Changes the CSS so that course cards tile correctly, even without course images and with variable course description lengths.
home-page-tiles

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -111,10 +111,10 @@
               {% endif %}
             </div>
             <div class="program-description">
-              <p id="program-{{ program.id }}-description">{{ program.description|default:"No description available for this program." }}</p>
+              <p class="program-description-text" id="program-{{ program.id }}-description">{{ program.description|default:"No description available for this program." }}</p>
 
               {% if program.programpage %}
-                <p>
+                <p class="program-description-link">
                 {% if program.programpage.external_program_page_url %}
                   <a href="{{ program.programpage.external_program_page_url }}" class="program-link">
                 {% else %}
--- a/static/scss/homepage.scss
+++ b/static/scss/homepage.scss
@@ -416,6 +416,10 @@ body.app-media {
     max-width: 1400px;
     margin: 0 auto;
 
+    li {
+      padding-bottom: 1em;
+    }
+
     .program-thumbnail {
       position: relative;
       background-color: black;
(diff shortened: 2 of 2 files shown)


amir-qayyum-khan:
there are issues with UI when program description is large, can you look into this.

screen shot 2016-11-17 at 5 09 37 pm

amir-qayyum-khan:
issue with image title

imageedit_2_9197766907

singingwolfboy:
When the program description is very long, it becomes scrollable -- if you hover your mouse over the description, you can scroll down. I set a defined height for the description text, to ensure that all the course cards are the same height, which is necessary for the course cards to tile correctly.

The only other way I can think of to make this work properly is to use the [CSS Flexbox layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes), which might make this work without requiring scrollable description text. However, I'm not very familiar with that. @roberthouse54, maybe you can help out?

Also, I'm not sure what the problem is with the image title. When the title is too wide, it will wrap to the next line. How should it behave instead?

roberthouse54:
I think the solution may be to set the height of the div (as done here) and then set a character limit on the program description in the CMS. About the course title wrapping... let me think a bit more on that, but please just add an issue for that and assign it to me.

amir-qayyum-khan:
@singingwolfboy there is one issue , with big height the space between 2 program rows gets 0,

please see 
screen shot 2016-11-17 at 5 09 37 pm

singingwolfboy:
I've changed the CSS so that the "More about..." link is always shown, even when the description is long enough that it will scroll. I also added some space between rows. Does that look better?
screen shot 2016-11-17 at 12 06 26 pm

roberthouse54:
Looks good, but can you please increase the space more? Like this: 
image

singingwolfboy:
Rebased and updated. How does this look?

bacon-ipsum

amir-qayyum-khan:
@singingwolfboy please reposition 

screen shot 2016-11-28 at 7 57 03 pm

**according to this.**
https://cloud.githubusercontent.com/assets/20047260/19897923/66ae314c-a030-11e6-9201-d5f0bf54cc71.png

singingwolfboy:
@amir-qayyum-khan That's outside the scope of this pull request. This pull request is _only_ about adjusting the course cards CSS so that they tile correctly. The mailing list button can be repositioned in a separate pull request.

amir-qayyum-khan:
then good to go. 👍

noisecapella:
@singingwolfboy please delete your branch after you merge the PR

PR #1788, opened 2016-11-16
Upgrade Raven

singingwolfboy:
Raven is the error-reporting client for [Sentry](https://sentry.io).

* [Python Changelog](https://github.com/getsentry/raven-python/blob/master/CHANGES)
* [JS Changelog](https://github.com/getsentry/raven-js/blob/master/CHANGELOG.md)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,7 +9,7 @@
     },
     "abbrev": {
       "version": "1.0.9",
-      "from": "abbrev@>=1.0.0 <1.1.0",
+      "from": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz"
     },
     "accepts": {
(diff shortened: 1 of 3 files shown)


aliceriot:
same comment about `npm-shrinkwrap.json` :)

singingwolfboy:
Fixed.

noisecapella:
:+1:

aliceriot:
The `npm-shrinkwrap` change seems to cover almost every dependency we use - is it possible you're using a different version of `npm` than the one we use in the docker container (3.9.5)?

singingwolfboy:
@aliceriot Because I'm on a Mac, I'm using a `node`/`npm` installation local to my Mac, managed via Homebrew. I like to keep things up to date, so I'm currently running `npm` version 3.10.9. I didn't realize that would be a problem.

Homebrew makes it hard to downgrade -- not impossible, but hard. Would you like me to try to do so, to get this pull request smaller? I'm not clear on how the `npm-shrinkwrap` file is affected by different versions of `node` and `npm`.

aliceriot:
I think generally you should try to use the version of node / npm we're standardized on. I'm not sure about the differences b/tw npm 3.9.5 and 3.10.9, but regardless we should be using the same versions of language runtimes and package managers everywhere.

The [node version manager](https://github.com/creationix/nvm) might be helpful in doing this.

singingwolfboy:
@aliceriot I just installed Node 6.2.2 and npm 3.9.5, and still got the same changes to the npm-shrinkwrap file when I did `npm install --save raven-js`. Maybe it's a mac-vs-linux thing? I'm confused.

aliceriot:
huh that's weird. idk! It's weird that it is changing lines for unrelated packages, that doesn't happen on linux.

what steps did you do exactly?

aliceriot:
bah `npm === pain`, we're just going to check in the big thing.

PR #1789, opened 2016-11-16
A few small changes and reorder sections on the home page

roberthouse54:
#### What's this PR do?

This PR fixes a few additional linear-gradient problems, and reorders the sections on the home page.


amir-qayyum-khan:
is this ok behaviour?  see coming soon, no background  image, and it is placed on right.
For example in given image i have only 3 programs. then coming soon is on second row right side

screen shot 2016-11-17 at 5 42 08 pm

amir-qayyum-khan:
seems the image you added not working due to cache or something
**images/more-coming-soon.png**

amir-qayyum-khan:
code looks good 👍

roberthouse54:
@amir-qayyum-khan There is a fix that @singingwolfboy did that (I believe) addresses the layout issue above (see: https://github.com/mitodl/micromasters/pull/1786). So the layout should be a 3-column grid that wraps and is left aligned if there is some number of items in the grid that is not divisible by 3. There is also another issue in which I address this with a new design for the page, when (in the future) we have more than 2 programs, and the "More Programs coming soon" is moved below the grid. That is: https://github.com/mitodl/micromasters/issues/1559

amir-qayyum-khan:
cool 👍

cms/templates/cms/home_page.html
```diff
@@ -187,6 +142,52 @@ <h4 class="program-title" aria-describedby="program-{{ program.id }}-description
...
+
+      <div class="centered-content">
+
+        <h3>Why should I enroll?</h3>
+        <div class="enroll-reason col-md-4">
+          <div class="reasons-images-wrap">
+             <img src="{% static 'images/online.png' %}" alt="">
+          </div>
+
+          <h5 class="upper-case">Convenient &amp; Flexible</h5>
```
singingwolfboy: `Why should I enroll?` is in an `<h3>` tag. Therefore, these reasons should all be in `<h4>` tags, not `<h5>` tags.

Issue #1222, opened 2016-10-04
Use <button> tag for accessibility


singingwolfboy:
For accessibility purposes, we should use the `<button>` tag for any action the user can take within the scope of a single page, and the `<a>` tag for a link from the current page to a different page. As a rule of thumb, any time see `<a href="#">`, or an `<a>` tag without an `href` attribute, you should convert it to a `<button>` instead.

pdpinch:
When we do this, it's important to review the UX

PR #1790, opened 2016-11-16
Correct buttons on cost calculator

singingwolfboy:
Adds `primary-button` and `secondary-button` CSS classes, and uses `<button>` for accessibility (#1222)


pdpinch:
Is this an accessibility fix?

singingwolfboy:
@roberthouse54 pointed out that the buttons on the financial calculator had the wrong background color on hover. In investigating that, I also found that the "Skip this and Pay Full Price" actions on the financial calculators were `<a>` tags, when they should be `<button>` tags for accessibility purposes. This pull request fixes both issues.

PR #1804, opened 2016-11-17
Increased default page size

noisecapella:
None

#### What's this PR do?
Increases default page size to 50, to match what it was set to before in the JS


rhysyngsun:
:+1: lgtm

PR #1805, opened 2016-11-17
Fixes the sort by dropdown layout

### LINKED ISSUE #1802: Fix layout of the Sort dropdown on the learner search
Somehow, the "Sort by" label has broken, and it now wraps.

image

roberthouse54:
#### What's this PR do?

This sets a min-width on the Sort By dropdown. It also makes the pagination buttons a bit less wide, and makes the sort less tall, so it looks less like a button.

Closes #1802

image

### DIFF
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -198,8 +198,13 @@
 
         .sk-toggle-option {
           background: #f5f5f5;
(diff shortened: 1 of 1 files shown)


rhysyngsun:
:+1: lgtm

PR #1806, opened 2016-11-17
edX user cached objects refactoring

### LINKED ISSUE #1664: Remove entries with `data=null` from the cache
Simplify what we store in the edX cache model by removing the entries with `data=null`.
This also means that such entries should not be created during the cache refresh.

### LINKED ISSUE #1667: Implement cache update logic simply based on last refresh call and expiration time
dependent on #1665

giocalitri:
closes #1664
closes #1667 

#### What's this PR do?
Refactors the edX cached elements. Now only elements (enrollments, certificates, current grades) that are available for the user are cached. There are no more cached elements with `data=None`.
The calls for cache updates are still done synchronously during a request to the dashboard REST API

`dashboard.api_edx_cache`

Run the migrations and verify that the dashboard rest api gives the same results as before.
Make sure that a payment for a course still results in a new cached enrollment. 
Make sure that the refresh via celery beat still works.

This PR does not implement the asynchronous refresh of the edx user data.

NOTE: the tests in `api_edx_cache_test` are conceptually different from the tests removed from `api_test`, so do not try to match them.


dashboard/models.py
```diff
@@ -19,13 +30,19 @@ class CachedEdxInfoModel(Model):
     user = ForeignKey(User)
     course_run = ForeignKey(CourseRun)
     data = JSONField(null=True)
```
noisecapella: If we are getting rid of `null` values should we set this to `null=False`?

dashboard/api_edx_cache.py
```diff
@@ -0,0 +1,264 @@
...
+        Checks if the specified cache type is expired and in case takes care to update it.
+
+        Args:
+            user (django.contrib.auth.models.User): A user
+            edx_client (EdxApi): EdX client to retrieve enrollments
+            cache_type (str): a string representing one of the cached data types
+        Returns:
+            None
+        """
+        cache_update_mothods = {
```
noisecapella: Should be `cache_update_methods`

dashboard/api_edx_cache.py
```diff
@@ -0,0 +1,264 @@
...
+        Checks if the specified cache type is expired and in case takes care to update it.
+
+        Args:
+            user (django.contrib.auth.models.User): A user
+            edx_client (EdxApi): EdX client to retrieve enrollments
+            cache_type (str): a string representing one of the cached data types
+        Returns:
+            None
+        """
+        cache_update_methods = {
```
noisecapella: Can we move this to a field on the class? I'm not sure if there's a circular reference problem here

dashboard/tasks_test.py
```diff
@@ -49,17 +49,14 @@ def test_celery_task_works(self):
...
-            self, mocked_refresh, mocked_get_enrollments, mocked_get_certificates, mocked_get_current_grades):
+            self, mocked_refresh, mocked_refresh_cache):
         """
         Assert get_student_enrollments is actually called
         """
         batch_update_user_data_subtasks.s(self.students).apply(args=()).get()
-        assert mocked_get_enrollments.called
-        assert mocked_get_certificates.called
-        assert mocked_get_current_grades.called
+        assert mocked_refresh_cache.called
```
giocalitri: the problem with asserting this here is that the function is called with an instance of `EdxApi` that is created inside `batch_update_user_data_subtasks`.
Would it be reasonable to make assertions like
`mocked_refresh_cache.assert_any_call(self.user, mock.ANY, 'enrollment')`
basically bypassing the `EdxApi` instance?

dashboard/models.py
```diff
@@ -18,14 +29,20 @@ class CachedEdxInfoModel(Model):
...
-    data = JSONField(null=True)
-    last_request = DateTimeField()
+    data = JSONField()
 
     class Meta:
         unique_together = (('user', 'course_run'), )
         abstract = True
 
     @classmethod
+    def active_all_qset(cls, user):
```
noisecapella: This name and `active_qset` are a bit confusing. Can we rename `active_qset` to `active_all_qset_for_program`? Or something else that distinguishes the two of them

dashboard/models.py
```diff
@@ -18,14 +29,20 @@ class CachedEdxInfoModel(Model):
...
+        """
+        Returns a queryset for the active records associated with a User
+        """
+        return cls.objects.filter(user=user)
+
+    @classmethod
     def active_qset(cls, user, program):
         """
         Returns a queryset for the active records associated with a User/Program pair
```
noisecapella: Can you remove the `exclude(data__isnull=True)` below?

noisecapella:
Can you remove the two `.exclude(data=None)` filters in `dashboard/views_test.py`?

noisecapella:
It looks like there's some code in `seed_data/lib.py` `CachedHandler` which may need to be updated

dashboard/migrations/0006_remove_all_cache.py
```diff
@@ -0,0 +1,38 @@
...
+
+def forwards_func(apps, schema_editor):
+    """
+    It deletes the cache because it will not be valid after the 0006 migration is applied
+    """
+    CachedEnrollment = apps.get_model("dashboard", "CachedEnrollment")
+    CachedCertificate = apps.get_model("dashboard", "CachedCertificate")
+    CachedCurrentGrade = apps.get_model("dashboard", "CachedCurrentGrade")
+    db_alias = schema_editor.connection.alias
+    CachedEnrollment.objects.using(db_alias).delete()
```
noisecapella: This is fine but just wondering what's going on here. Is it necessary to specify the `db_alias` here?

dashboard/api_edx_cache.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @classmethod
+    def update_cache_last_access(cls, user, cache_type, timestamp=None):
+        """
+        Updates the UserCacheRefreshTime model timestamp for the provided cache type
+
+        Args:
+            user (django.contrib.auth.models.User): A user
+            cache_type (str): a string representing one of the cached data types
+            timestamp (datetime): a timestamp
```
noisecapella: Just being pedantic but I think you want `datetime.datetime` here for the arg type

dashboard/models.py
```diff
@@ -50,6 +67,28 @@ def active_data(cls, user, program):
         """
         return cls.active_qset(user, program).values_list('data', flat=True).all()
 
+    @classmethod
+    def active_course_ids(cls, user):
+        """
+        Returns a list of all the Course IDs for the cached data
```
noisecapella: Can you add args and return values in the docstrings in this class?

dashboard/factories.py
```diff
@@ -38,9 +38,8 @@ class CachedCertificateFactory(DjangoModelFactory):
         "certificate_type": "verified",
         "grade": randint(60, 100)/100.0,
         "course_id": x.course_run.edx_course_key,
+        "username": x.user.username,
```
noisecapella: We will need to be careful about this. Some tests use `UserFactory` but set the edX username to a different name to make sure that we can handle the two being different. Where do we use this username in our code?

dashboard/models.py
```diff
@@ -66,28 +105,59 @@ class CachedEnrollment(CachedEdxInfoModel):
...
+    user = OneToOneField(User)
     enrollment = DateTimeField(null=True)
     certificate = DateTimeField(null=True)
     current_grade = DateTimeField(null=True)
 
+    def __str__(self):
+        """
+        String representation of the model object
+        """
+        return 'user "{0}"'.format(self.user.username)
```
noisecapella: Can you use `get_social_username` here?

dashboard/tasks.py
```diff
@@ -71,20 +67,12 @@ def release_lock():
     if acquire_lock():
         refresh_time_limit = datetime.utcnow() - timedelta(hours=MAX_HRS_MAIN_TASK)
```
noisecapella: Can we make constants of the time values in `CachedEdxUserData.CACHE_EXPIRATION_DELTAS`, then make `MAX_HRS_MAIN_TASK` calculated based on the max value?

dashboard/views.py
```diff
@@ -105,7 +101,7 @@ def post(self, request):  # pylint: disable=no-self-use
                 status=status.HTTP_500_INTERNAL_SERVER_ERROR,
                 data={'error': str(exc)}
             )
-        update_cached_enrollment(request.user, enrollment, enrollment.course_id, datetime.now(pytz.UTC))
+        CachedEdxUserData.update_cached_enrollment(request.user, enrollment, enrollment.course_id)
```
noisecapella: Should `index_user=True` here?

dashboard/views_test.py
```diff
@@ -38,6 +39,12 @@ def setUpTestData(cls):
             uid="{}_edx".format(cls.user.username),
             extra_data={"access_token": "fooooootoken"}
         )
+        cls.user_access_cache = UserCacheRefreshTimeFactory(
```
noisecapella: It doesn't look like we use this anywhere. Does it have some other effect on the tests?

search/indexing_api_test.py
```diff
@@ -215,38 +216,43 @@ def test_add_edx_record(self):
         Test that cached edX records are indexed after being added
         """
         program_enrollment = ProgramEnrollmentFactory.create()
-        for edx_cached_model_factory in [CachedCertificateFactory, CachedEnrollmentFactory]:
+        for edx_cached_model_factory in [CachedCertificateFactory, CachedEnrollmentFactory, CachedCurrentGradeFactory]:
             assert es.search()['total'] == 1
             course = CourseFactory.create(program=program_enrollment.program)
             course_run = CourseRunFactory.create(course=course)
             edx_cached_model_factory.create(user=program_enrollment.user, course_run=course_run)
+            index_program_enrolled_users([program_enrollment])
```
noisecapella: Do we want to create the timestamp too? I'm not sure if it's relevant for these tests

seed_data/lib.py
```diff
@@ -2,14 +2,13 @@
 Library functions for interacting with test data
```
noisecapella: There are some references to `last_request` here which was deleted. I mentioned `CachedHandler` before but there's also `deserialize_user_data` and `CachedModelDeserializer`

noisecapella:
It looks like there are some migration errors in the tests caused by the not null constraint

dashboard/api_test.py
```diff
@@ -754,16 +751,17 @@ def setUp(self):
...
+    @patch('dashboard.api_edx_cache.CachedEdxUserData.update_cache_if_expired', new_callable=MagicMock)
+    @patch('dashboard.api_edx_cache.CachedEdxUserData.get_cached_edx_data', new_callable=MagicMock)
+    def test_format(self, mock_cache_get, mock_cache_refresh):
         """Test that get_user_program_info fetches edx data and returns a list of Program data"""
         result = api.get_user_program_info(self.user, self.edx_client)
 
-        for mocked_edx_get_function in [mock_grades, mock_cert, mock_enr]:
-            assert mocked_edx_get_function.call_count == 1
-            assert mocked_edx_get_function.call_args[0] == (self.user, self.edx_client)
+        assert mock_cache_refresh.call_count == 3
```
noisecapella: Can you use `len(CachedEdxUserData.SUPPORTED_CACHES)` in place of `3` here, just to make the connection more explicit?

dashboard/api_edx_cache.py
giocalitri: the reason because I put it here is because I cannot create a class variable referencing the class methods before defining the class methods.
The only way I know would be putting all the other methods in another class and then subclassing it to add this variable and this method.
what do you think?

dashboard/models.py
giocalitri: sure

dashboard/models.py
giocalitri: sure

dashboard/models.py
```diff
@@ -18,14 +29,20 @@ class CachedEdxInfoModel(Model):
...
 
     @classmethod
+    def active_all_qset(cls, user):
+        """
+        Returns a queryset for the active records associated with a User
+        """
+        return cls.objects.filter(user=user)
+
+    @classmethod
     def active_qset(cls, user, program):
```
giocalitri: sure

giocalitri:
> Can you remove the two `.exclude(data=None)` filters in `dashboard/views_test.py`?

Sure

dashboard/migrations/0006_remove_all_cache.py
giocalitri: From what I understood [here](https://docs.djangoproject.com/en/1.10/ref/migration-operations/#runpython) it is needed to be sure to use the right version of the model.
In this specific case it is probably superfluous.

dashboard/api_edx_cache.py
giocalitri: ok

dashboard/factories.py
giocalitri: the problem I have is that I need that the username in the cached data is consistent, because some code (edx_api_client.CurrentGrades) checks that for obvious reasons.
Now I really do not care what we put here, but I guess that in the cases you are talking about the solution is to explicitly pass a custom defined `data` parameter for the factory.

seed_data/lib.py
gsidebo: i'm happy to clean up this seed data code in a later PR. if some parts of it break as a result of this PR, there isn't a significant impact on us

dashboard/api_edx_cache.py
noisecapella: I'm fine leaving it here

dashboard/models.py
giocalitri: we decided to keep it as is

dashboard/views.py
giocalitri: yes

dashboard/views_test.py
giocalitri: we do not user `cls.user_access_cache`, but the entry in `UserCacheRefreshTime` is necessary (note that is set in the future).
Removing the class variable anyway

search/indexing_api_test.py
giocalitri: Actually should not be relevant because the indexing process does not care if the cache is fresh, it simply indexes what is in the DB.

dashboard/api_test.py

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+            grade.update({'username': cls.user.username})
+            current_grades_json.append(grade)
+        cls.current_grades = CurrentGrades([CurrentGrade(grade_json) for grade_json in current_grades_json])
+
+        cls.certificates_ids = set(cls.certificates.all_courses_certs)
+        cls.verified_certificates_ids = set(cls.certificates.all_courses_verified_certs)
+        cls.enrollment_ids = set(cls.enrollments.get_enrolled_course_ids())
+        cls.grades_ids = set(cls.current_grades.all_course_ids)
+        cls.all_course_run_ids = list(
+            set(cls.certificates_ids) | set(cls.enrollment_ids) | set(cls.grades_ids))
```
noisecapella: You don't need the `set()` here since these three are already sets

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+            CachedCurrentGradeFactory.create(user=self.user, course_run=run)
+        self.assert_cache_in_db(self.all_course_run_ids, self.all_course_run_ids, self.all_course_run_ids)
+
+    def test_update_cache_last_access(self):
+        """Test for update_cache_last_access"""
+        with self.assertRaises(ValueError):
+            CachedEdxUserData.update_cache_last_access(self.user, 'footype')
+        assert UserCacheRefreshTime.objects.filter(user=self.user).exists() is False
+
+        CachedEdxUserData.update_cache_last_access(self.user, CachedEdxUserData.ENROLLMENT)
```
noisecapella: Can you use `ddt` to do this test for each of the three cache types?

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+            "is_active": True,
+            "mode": "verified",
+            "user": self.user.username
+        }
+        enrollment_new = Enrollment(enr_json)
+        CachedEdxUserData.update_cached_enrollment(self.user, enrollment_new, course_id, True)
+        self.assert_cache_in_db(enrollment_keys=[course_id])
+        cached_enr.refresh_from_db()
+        assert cached_enr.data == enr_json
+        assert mocked_index.delay.called is True
```
noisecapella: Can we assert the called argument here?

noisecapella: And in the other two tests below please

dashboard/api_edx_cache_test.py
giocalitri: this might be hard-ish
the tests use data accessible via `self` that is not accessible in the decorator.
Plus the certificates test is slightly different.

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+        cls.certificates_ids = set(cls.certificates.all_courses_certs)
+        cls.verified_certificates_ids = set(cls.certificates.all_courses_verified_certs)
+        cls.enrollment_ids = set(cls.enrollments.get_enrolled_course_ids())
+        cls.grades_ids = set(cls.current_grades.all_course_ids)
+        cls.all_course_run_ids = list(
+            set(cls.certificates_ids) | set(cls.enrollment_ids) | set(cls.grades_ids))
+        cls.all_runs = []
+
+        for course_id in cls.all_course_run_ids:
+            run = CourseRunFactory.create(edx_course_key=course_id)
```
giocalitri: sure

dashboard/api_edx_cache_test.py
giocalitri: sure

dashboard/api_edx_cache_test.py
noisecapella: That's fine, let's leave it as is

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+        CachedEdxUserData.update_cached_enrollments(self.user, self.edx_client)
+        self.assert_cache_in_db(enrollment_keys=self.enrollment_ids)
+        cache_time.refresh_from_db()
+        assert cache_time.enrollment >= now
+        assert mocked_index.delay.called is True
+
+    @patch('search.tasks.index_users', autospec=True)
+    def test_update_cached_certificates(self, mocked_index):
+        """Test for update_cached_certificates."""
+        assert len(self.certificates_ids) > len(self.verified_certificates_ids)
```
noisecapella: Instead of asserting length here can you assert that one is a subset of the other?

giocalitri: sure

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+        self.assert_cache_in_db()
+        assert UserCacheRefreshTime.objects.filter(user=self.user).exists() is False
+        CachedEdxUserData.update_cached_certificates(self.user, self.edx_client)
+        self.assert_cache_in_db(certificate_keys=self.verified_certificates_ids)
```
noisecapella: Is there some way to assert that there are some course keys in `self.certificates_ids` which are not in the cache?

giocalitri: `self.assert_cache_in_db(certificate_keys=self.verified_certificates_ids)`
verifies that ONLY the passed ids are in the cache. This together with the first call (and the mocked responses) should answer your question.
If you want I can make an explicit query for this.

dashboard/tasks_test.py
```diff
@@ -49,17 +58,16 @@ def test_celery_task_works(self):
...
         Assert get_student_enrollments is actually called
         """
         batch_update_user_data_subtasks.s(self.students).apply(args=()).get()
-        assert mocked_get_enrollments.called
-        assert mocked_get_certificates.called
-        assert mocked_get_current_grades.called
-        assert mocked_refresh.called
+        assert mocked_refresh_cache.call_count == len(self.all_users) * len(CachedEdxUserData.SUPPORTED_CACHES)
+        assert mocked_refresh.call_count == len(self.all_users)
+        for user, cache_type in product(self.all_users, CachedEdxUserData.SUPPORTED_CACHES):
```
noisecapella: That's great, I've never seen a good use of `product` before

dashboard/api_edx_cache_test.py
noisecapella: No, leave it as is

noisecapella:
Functionality looks good

dashboard/signals.py
```diff
@@ -4,7 +4,7 @@
 from django.db.models.signals import post_save, post_delete
 from django.dispatch import receiver
 
-from dashboard.models import CachedEnrollment, CachedCertificate, ProgramEnrollment
+from dashboard.models import ProgramEnrollment
 from search.tasks import index_program_enrolled_users, remove_program_enrolled_user
```
noisecapella: Can you remove the now unused function `_index_enrollment` below?

dashboard/api_edx_cache_test.py
```diff
@@ -0,0 +1,306 @@
...
+
+        old_timestamp = datetime.now(tz=pytz.UTC) - timedelta(days=1)
+        CachedEdxUserData.update_cache_last_access(self.user, CachedEdxUserData.ENROLLMENT, old_timestamp)
+        cache_time.refresh_from_db()
+        assert cache_time.enrollment == old_timestamp
+
+    def test_is_cache_fresh(self):
+        """Test for is_cache_fresh"""
+        with self.assertRaises(ValueError):
+            CachedEdxUserData.update_cache_last_access(self.user, 'footype')
```
giocalitri: right

dashboard/signals.py
giocalitri: sure

dashboard/models.py
```diff
@@ -18,46 +29,99 @@ class CachedEdxInfoModel(Model):
...
+    def active_course_ids(cls, user):
         """
-        Returns the number of active records associated with a User/Program pair
+        Returns a list of all the Course IDs for the cached data
+
+        Args:
+            user (User): an User object
+
+        Returns:
+            QuerySet: a queryset of all the course key fields for the provided user
```
noisecapella: Can you change the type to `list`?

giocalitri: sorry

Issue #1712, opened 2016-11-08
Search: results should be displaying First Name and Last Name.


giocalitri:
The default search sorting is by `Last Name` but then we only show the preferred name (my guess).

![screen shot 2016-11-08 at 1 35 18 pm](https://cloud.githubusercontent.com/assets/1005204/20111989/7063dcf4-a5b8-11e6-86fb-e8223d0afe30.png)

pdpinch:
@roberthouse54

roberthouse54:
The search results should be displaying First Name and Last Name.

pdpinch:
@roberthouse54 <First Name> <Last Name> or <Last Name>, <First Name> ?

Ferdi:
Sorting should not have any effect on what's shown in the search result

giocalitri:
Sorting works.
It is only the display that is confusing.

roberthouse54:
First Name Last Name, as in the mock.

pdpinch:
@amir-qayyum-khan this doesn't need a "Work in progress" label does it?

aliceriot:
I said this in standup this morning, but I think the preferred name should be included in the search result too. Something like `First Last (Preferred)`. I think we should ensure that people who are e.g. sending an email to a user would know what name to use when addressing them.

roberthouse54:
OK, I think First Last (Preferred) is an acceptable solution for the time being, but I think at some point I'd like to add column headers here as the table will be getting a bit more complex. If we did that the column headers could be First Name/Last Name and Preferred Name

PR #1808, opened 2016-11-18
Set learners name on search to first and last name pair

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1712

#### What's this PR do?
- On search page, instead of showing preferred name of learners, it shows first and last name pair.

@pdpinch @giocalitri

### DIFF
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -451,7 +451,9 @@ export const labelSort = R.sortBy(R.compose(R.toLower, R.prop('label')));
  * Return first and last names if available else username
  */
 export function getUserDisplayName(profile: Profile): string {
-  return (
-    profile.first_name && profile.last_name ? `${profile.first_name} ${profile.last_name}` : profile.username
-  );
+  let first = profile.first_name || profile.username;
+  let last =  profile.last_name || '';
+  let preferred_name = profile.preferred_name ? `(${profile.preferred_name})` : '';
+
+  return `${first} ${last} ${preferred_name}`;
 }
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -446,3 +446,12 @@ export const classify: (s: string) => string = (
 );
 
 export const labelSort = R.sortBy(R.compose(R.toLower, R.prop('label')));
+
+/**
+ * Return first and last names if available else username
+ */
+export function getUserDisplayName(profile: Profile): string {
+  return (
+    profile.first_name && profile.last_name ? `${profile.first_name} ${profile.last_name}` : profile.username
+  );
+}
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -448,12 +448,12 @@ export const classify: (s: string) => string = (
 export const labelSort = R.sortBy(R.compose(R.toLower, R.prop('label')));
 
 /**
- * Return first and last names if available else username
+ * Return first, last, preferred_name names if available else username
  */
 export function getUserDisplayName(profile: Profile): string {
   let first = profile.first_name || profile.username;
   let last =  profile.last_name || '';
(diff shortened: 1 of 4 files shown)


rhysyngsun:
@amir-qayyum-khan realized I took this but it doesn't have a `Needs Review` tag. Let me know if that's the case and I'll give it a review!

pdpinch:
I'd like to get @roberthouse54 to weigh in on the preferred name question before we merge this.

roberthouse54:
I am ok with showing the name in the format FirstName LastName (preferredName).

amir-qayyum-khan:
@aliceriot @pdpinch @roberthouse54 

screen shot 2016-11-22 at 2 42 08 pm

gsidebo:
@roberthouse54 @pdpinch design question: should we hide this preferred name if it's the exact same as the user's first name? it seems unnecessary for us to show, for example, "Gavin Sidebottom (Gavin)"

amir-qayyum-khan:
this  means that his name is Gavin Sidebottom (Gavin) but he prefer Gavin only.
i think this is ok.

static/js/util/util.js
```diff
@@ -428,3 +428,14 @@ export function findCourseRun(
...
+
+/**
+ * Return first and last names if available else username
+ */
+export function getUserDisplayName(profile: Profile): string {
+  let first = profile.first_name || profile.username;
+  let last =  profile.last_name || '';
+  let preferred_name = profile.preferred_name ? `(${profile.preferred_name})` : '';
+
+  return `${first} ${last} ${preferred_name}`;
```
gsidebo: nitpick: can you make this `${first} ${last}${preferred_name}` and move the space into the `preferred_name` value so we don't have an extra space?

static/js/util/util_test.js
```diff
@@ -538,4 +539,30 @@ describe('utility functions', () => {
...
+    beforeEach(() => {
+      profile = {
+        username: 'jane_username',
+        first_name: 'jane',
+        last_name: 'doe',
+        preferred_name: 'test'
+      };
+    });
+
+    it('assert first and last name', () => {
```
gsidebo: please change this `it` text to: `shows first, last, and preferred names`

static/js/util/util_test.js
```diff
@@ -538,4 +539,30 @@ describe('utility functions', () => {
...
+      assert.equal('jane doe (test)', getUserDisplayName(profile));
+    });
+
+    it('returns username when first not available', () => {
```
gsidebo: please change this `it` text to: `shows username when first name is blank`

static/js/util/util_test.js
```diff
@@ -538,4 +539,30 @@ describe('utility functions', () => {
...
+      profile.first_name = null;
+      assert.equal('jane_username doe (test)', getUserDisplayName(profile));
+    });
+
+    it('not return preferred_name when preferred_name not available', () => {
```
gsidebo: please change this `it` text to: `does not show preferred_name when that value is blank`

amir-qayyum-khan:
Done @gsidebo

roberthouse54:
We have spent probably too much time on this in my opinion, and I think part of the problem is that this is a required field, when it probably shouldn't be. The point of this field is really to allow users to specify a Preferred name if it is different from their first name. Probably in the majority of cases, the first name is going to be the preferred name, but we added this field in the cases where it is not. I do agree that this looks odd. But I think we are going to have problems like this as long as fields which should be optional are required. I think I agree with Gavin's suggestion as a solution for the time being.

amir-qayyum-khan:
Done @roberthouse54 @gsidebo

PR #1831, opened 2016-11-21
Allow the user to create new entries in 'industry' and 'field of study'

### LINKED ISSUE #1519: Sign up: add support for "other" in industry and field of study select fields
This will need some UX design first. 
- [ ] Add "other" to the list of values. 
- [ ] When user has chosen "other," display an additional field below the select field where the user can add a free text value
- [ ] In search, these users will appear with a facet value of "other"
- [ ] later, we can analyze the values that are entered and determine if we need to add a new value to our list
- [ ] If the user types something that matches nothing in the list, then, instead of displaying "Other", display "No results. Enter another Industry/Field of Study[link]" Clicking the link is the same as clicking "Other" from the list of options.

See also #1513

aliceriot:

closes #1519 

#### What's this PR do?

This adds support to adding a custom entry to our `SelectField` component. In order to turn this on, you just pass a boolean `allowCreate` prop.

Read over the changes to `components/inpust/SelectField.js`.

On the profile everything should have all of it's normal functionality.

The new thing is if you're adding a work history or education history element. 

- If you type in something which is not in the list, you should see an option which says `'Create option "your input so far"'`.
- If you press tab or click on this your input should be the new selected value of the field.
- If you clear the field, the entry you entered should appear as an option you can re-select.
- If you reload the app and go to the same entry, the custom entry you saved should appear.

I think that's it?

potato_surgery

### DIFF
--- a/static/js/components/inputs/SelectField.js
+++ b/static/js/components/inputs/SelectField.js
@@ -2,6 +2,7 @@
 import React from 'react';
 import _ from 'lodash';
 import VirtualizedSelect from 'react-virtualized-select';
+import { Creatable } from 'react-select';
 
 import { validationErrorSelector, classify } from '../../util/util';
 import type { Option } from '../../flow/generalTypes';
@@ -12,36 +13,37 @@ import type {
   UpdateProfileFunc,
 } from '../../flow/profileTypes';
 
+export const CREATE_OPTION_REGEX = /^Create option "|"$/g;
+
+type SelectFieldProps = {
+  allowCreate?:               boolean,
+  className:                  string,
+  errors:                     ValidationErrors,
+  id?:                        string,
+  keySet:                     Array<string>,
+  label:                      string,
+  onChange:                   Function,
+  options:                    Array<Option>,
+  profile:                    Profile,
+  topMenu:                    boolean,
+  updateProfile:              UpdateProfileFunc,
+  updateValidationVisibility: (xs: Array<string>) => void,
+  validator:                  Validator|UIValidator,
+};
+
 class SelectField extends React.Component {
-  props: {
-    className:                  string,
-    errors:                     ValidationErrors,
-    id?:                        string,
-    keySet:                     Array<string>,
-    label:                      string,
-    onChange:                   Function,
-    options:                    Array<Option>,
-    profile:                    Profile,
-    topMenu:                    boolean,
-    updateProfile:              UpdateProfileFunc,
-    updateValidationVisibility: (xs: Array<string>) => void,
-    validator:                  Validator|UIValidator,
-  };
-
-  state: {
-    id: string,
-  };
-
-  componentWillMount: Function = (): void => {
-    const { label } = this.props;
-    let id = this.props.id;
-    if (!id) {
-      id = _.uniqueId(classify(label));
-    }
-    this.setState({id: id});
+  constructor (props: SelectFieldProps) {
+    super(props);
+    this.state = {
+      customOptions: []
+    };
   }
 
-  onChange: Function = (selection: Option): void => {
+  state: {
+    customOptions: Array<Option>
+  };
+
+  onChange = (selection: Option): void => {
     const {
       profile,
       updateProfile,
@@ -49,11 +51,15 @@ class SelectField extends React.Component {
       validator,
     } = this.props;
     let clone = _.cloneDeep(profile);
-    _.set(clone, keySet, selection ? selection.value : "");
+    _.set(
+      clone,
+      keySet,
+      selection && selection.value ? selection.value.replace(CREATE_OPTION_REGEX, '') : ""
+    );
     updateProfile(clone, validator);
   };
 
-  onBlur: Function = (): void => {
+  onBlur = (): void => {
     const {
       updateValidationVisibility,
       validator,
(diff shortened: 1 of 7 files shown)


noisecapella:
If I press `<` when the create option item is selected, it selects the item. I've only seen this with the field of study select so far

noisecapella:
Functionality looks good other than the quirk I mentioned, looking at code...

static/js/components/inputs/SelectField.js
```diff
@@ -72,34 +63,46 @@ class SelectField extends React.Component {
...
+      { element }
+    </label>
+  );
+
+  renderSelect = (): React$Element<*> => {
+    const { keySet, profile, allowCreate, options } = this.props;
+
+    let value = _.get(profile, keySet);
+    if ( allowCreate && value && !options.find(option => option.value === value) ) {
+      options.push({ value: value, label: value });
```
noisecapella: We should use `options = options.concat(...)` here since we can't alter props directly

static/js/components/inputs/inputs_test.js
```diff
@@ -39,13 +40,15 @@ describe('Profile inputs', () => {
...
-    let renderGenderSelectField = () => {
-      Object.assign(inputProps, {
-        keySet: ['gender'],
-        label: "Gender",
-        options: genderOptions
-      });
-      return renderTestSelectField();
+    let renderGenderSelectField = (props = {}) => {
+      return renderTestSelectField(
+        Object.assign(inputProps, {
```
noisecapella: Can you change this line to `Object.assign({}, inputProps, {` so `inputProps` isn't altered?

static/js/components/inputs/SelectField.js
aliceriot: It seems to work fine when I use `.push`, do you mean we shouldn't alter props? My thought in using `.push` was that I want the option to persist between re-renders, so that if you enter a 'created' option and then change the value, you can change your selection and your created value will still be there. If I use `concat` we don't get that behavior, so I would need to implement something more complicated, I think, to keep track of new `Option`s which had been added.

noisecapella: Yes, if props are mutated it breaks the assumption that `render` is a pure function: https://facebook.github.io/react/docs/components-and-props.html#props-are-read-only

If we wanted to have some state persist we should put it in redux or use `setState` if that state is confined to this particular component

static/js/components/inputs/SelectField.js
```diff
@@ -12,8 +13,11 @@ import type {
   UpdateProfileFunc,
 } from '../../flow/profileTypes';
 
+const CREATE_OPTION_REGEX = /^Create option \"|\"$/g;
```
aliceriot: hrm if we do a capture group we lose the ability to call `.replace` regardless of what the input is, since `/NO_MATCHES/.exec('a different string') === null`.

noisecapella: I'm fine using `replace` as long as we have tests for various inputs that match against the regex. By the way my IDE says that the `\` in the `\"` in the regex are unnecessary

static/js/components/inputs/SelectField.js
aliceriot: Ok, I guess I'll use `this.state` for it then.

static/js/components/inputs/SelectField.js
aliceriot: ok cool, I'll add some tests.

noisecapella:
Can you rebase this too? I want to see how the newly merged code coverage works :)

static/js/components/inputs/inputs_test.js
```diff
@@ -78,19 +81,46 @@ describe('Profile inputs', () => {
...
+        date_of_birth: '',
+        gender: 'genderqueer',
+        date_field: ''
+      }), 'should be called with the right thing');
+    });
+
+    it('should enter the new option in the options list', () => {
+      selectField = renderGenderSelectField({allowCreate: true}).find(VirtualizedSelect);
+      modifyWrapperSelectField(selectField, 'genderqueer');
+      let newOption = selectField.props().options.find(option => (
```
aliceriot: which div are you referring to exactly?

noisecapella: On this line, the id prop using `formattedId()`: https://github.com/mitodl/micromasters/pull/1831/files#diff-f57bc642ed5426be1405cd5b2aaca97dR125

aliceriot:

noisecapella:
It looks like a part of the `componentDidUpdate` function aren't covered by tests: https://codecov.io/gh/mitodl/micromasters/compare/203f027dc92ca8d9b91add7219f8e14e3b864f31...77db88ba2374066674384b1e480a6ca4992e6933?src=pr#7374617469632F6A732F636F6D706F6E656E74732F696E707574732F53656C6563744669656C642E6A73-99

Can you write a test which will get the `setState` code executed?

static/js/components/inputs/SelectField_test.js
```diff
@@ -0,0 +1,30 @@
+import { assert } from 'chai';
+
+import { CREATE_OPTION_REGEX } from './SelectField';
+
+// export const CREATE_OPTION_REGEX = /^Create option "|"$/g;
```
noisecapella: Can you remove this line?

static/js/components/inputs/inputs_test.js
noisecapella: Was this comment addressed?

aliceriot: sorry didn't see your comment

pdpinch:
Do you know what happens in the search index?

noisecapella:
It just stores the field of study as a text field in the database, so there shouldn't be any change there

aliceriot:
@noisecapella I think the coverage issue should be fixed, although I'm not sure how to check!

noisecapella:
Yes, the most recent coverage is updated in place in the comment (see the `edited` above)

aliceriot:
oh nifty!

aliceriot:
@noisecapella added new tests for the `id` and `class` on the rendered component.

PR #1847, opened 2016-11-22
Remove border between row and sub rows

### LINKED ISSUE #1599: (Bug) Course with multiple past runs has incorrect border placement
Style rules need to be changed to get rid of the border between the past runs and the currently displayed run 

ss 2016-10-28 at 11 58 28

noisecapella:
Fixes #1599 

#### What's this PR do?
Removes border between row and sub rows

I set up the PR build to have past runs, so you should be able to see the fix there, under Digital Learning. There should be no border between the past run message and the current run.

### DIFF
--- a/static/scss/dashboard.scss
+++ b/static/scss/dashboard.scss
@@ -143,7 +143,6 @@
       font-size: 11pt;
       align-items: center;
       padding: 14px 20px;
-      border-bottom: 1px solid $lightgrey;
 
       &:last-child {
         border-bottom: none;
--- a/static/scss/dashboard.scss
+++ b/static/scss/dashboard.scss
@@ -144,10 +144,6 @@
       align-items: center;
       padding: 14px 20px;
 
-      &:last-child {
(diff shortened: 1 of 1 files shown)


rhysyngsun:
@noisecapella do I need to login as a certain user? I don't see past runs on that PR build under a new user.

aliceriot:
Did you check the digital learning program?

static/scss/dashboard.scss
```diff
@@ -143,7 +143,6 @@
       font-size: 11pt;
       align-items: center;
       padding: 14px 20px;
-      border-bottom: 1px solid $lightgrey;
 
       &:last-child {
```
rhysyngsun: @noisecapella actually, since we're no longer putting a border on `.course-row`, this `&:last-child` block can probably go too.

noisecapella: Sounds good

rhysyngsun:
:+1:

PR #1848, opened 2016-11-22
Removes the x in the react select component

roberthouse54:
#### What's this PR do?

This is a simple PR that removes the "x" on the react select component in the Sign Up forms.


noisecapella:
Can you rebase?

noisecapella:
Can you also add `clearable={false}` here? https://github.com/mitodl/micromasters/blob/master/static/js/containers/FinancialAidCalculator.js#L43

PR #1855, opened 2016-11-23
Fix bug with custom entry

### LINKED ISSUE #1854: custom entries aren't being display when editing existing saved entries
I thought I had this working when I merged #1831, but it seems to be broken today.

aliceriot:

closes #1854 

#### What's this PR do?

This ensures that when opening an existing education or work history entry (which has a custom entry) the user will see the value in the field.

1. edit work history -> industry or educatoin -> field of study
2. save a custom option (e.g. 'Salmon psychology' or something)
3. open the entry again (w/ the edit button)
4. confirm that the custom entry is in the field still

### DIFF
--- a/static/js/components/inputs/SelectField.js
+++ b/static/js/components/inputs/SelectField.js
@@ -95,7 +95,7 @@ class SelectField extends React.Component {
     </label>
   );
 
-  componentDidUpdate () {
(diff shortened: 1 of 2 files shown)


aliceriot:
@noisecapella would you mind taking a quick look at this? I'm going to add a test as well.

PR #1858, opened 2016-11-23
make profile photo required during signup

### LINKED ISSUE #1762: Sign up: make avatar required

aliceriot:

closes #1762 

#### What's this PR do?

This PR adds a new requirement to the signup process, which is that the user add a profile image.

First,check that the validation is working correctly:

1. use the django shell or similar to delete your user's profile image (if present)
2. visit `/profile/personal`
3. click 'next'

there should be a validation error, which will disappear when you add a profile image. Then you should be able to proceed with the rest of the signup flow.

The other thing to check is that the edit state for the personal form is persisted when updating the profile image. This was a bit tricky to get working, because the currently edited `profile` is what we check via validation, and the user may have started editing the form before adding an image.

1. visit `/profile/personal`
2. make some edit to the form which results in a validation error (for instance removing your last name) and also an unsaved change to the form (like changing your first name to 'Salmon' or something).
3. Go through the profile image update process.
4. The form should still be the same.

In particular, if you're doing the above _without_ a profile image the validation error for the missing image should go away, while the rest of the edit state is persisted.

I think that's it, besides generally checking I didn't clobber any other profile functionality.

add_a_photo_error

add_a_photo

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -3,6 +3,7 @@
 import React from 'react';
 import _ from 'lodash';
 import Card from 'react-mdl/lib/Card/Card';
+import CardTitle from 'react-mdl/lib/Card/CardTitle';
 import R from 'ramda';
 import Select from 'react-select';
 
@@ -12,9 +13,11 @@ import {
   combineValidators,
   personalValidation,
   programValidation,
+  profileImageValidation,
 } from '../lib/validation/profile';
 import type {
   Profile,
+  ProfileFetchResponse,
   SaveProfileFunc,
   ValidationErrors,
   UpdateProfileFunc,
@@ -28,7 +31,13 @@ import {  validationErrorSelector } from '../util/util';
 import type { Option } from '../flow/generalTypes';
 import { setProfileStep } from '../actions/ui';
 import { PERSONAL_STEP } from '../constants';
+import ProfileImage from '../containers/ProfileImage';
 
+const personalTabValidator = combineValidators(
+  personalValidation,
+  programValidation,
+  profileImageValidation,
+);
 
 export default class PersonalTab extends React.Component {
   props: {
@@ -38,6 +47,7 @@ export default class PersonalTab extends React.Component {
     nextStep:                 () => void,
     prevStep:                 () => void,
     profile:                  Profile,
+    uneditedProfile:          Profile,
     programs:                 AvailablePrograms,
     saveProfile:              SaveProfileFunc,
     setProgram:               Function,
@@ -89,8 +99,16 @@ export default class PersonalTab extends React.Component {
     );
   };
 
+  afterImageUpload = (fetchResponse: ProfileFetchResponse): void => {
+    const { profile, updateProfile } = this.props;
+    let editState = _.cloneDeep(profile);
+    let { payload: { profile: { image } }} = fetchResponse;
+    let newEditState = Object.assign({}, editState, { image });
+    updateProfile(newEditState, personalTabValidator);
+  };
+
   render() {
-    const { ui: { selectedProgram }, errors } = this.props;
+    const { ui: { selectedProgram }, errors, uneditedProfile } = this.props;
 
     return (
       <div>
@@ -104,6 +122,21 @@ export default class PersonalTab extends React.Component {
             {_.get(errors, ['program'])}
           </span>
         </Card>
+        <Card shadow={1} className="profile-image">
+          <CardTitle>
+            Upload a Profile Photo
+          </CardTitle>
+          <ProfileImage
+            profile={uneditedProfile}
+            editable={true}
+            showLink={true}
+            linkText="Click here to add a profile photo"
+            afterImageUpload={this.afterImageUpload}
+          />
+          <span className="validation-error-text">
+            {_.get(errors, ['image'])}
+          </span>
+        </Card>
         <Card shadow={1} className="profile-form">
           <PersonalForm {...this.props} validator={personalValidation} />
         </Card>
@@ -113,12 +146,7 @@ export default class PersonalTab extends React.Component {
           nextBtnLabel="Next"
           programIdForEnrollment={selectedProgram ? selectedProgram.id : null}
           isLastTab={false}
-          validator={
-            combineValidators(
-              personalValidation,
-              programValidation
-            )
-          }
+          validator={personalTabValidator}
         />
       </div>
     );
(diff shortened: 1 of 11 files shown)


static/js/components/PersonalTab.js
```diff
@@ -38,6 +47,7 @@ export default class PersonalTab extends React.Component {
     nextStep:                 () => void,
     prevStep:                 () => void,
     profile:                  Profile,
+    _profile:                 Profile,
```
noisecapella: Is there another name we can use? It's not obvious what the difference is here

static/js/containers/ProfileImage_test.js
```diff
@@ -43,29 +49,27 @@ describe('ProfileImage', () => {
 
   describe('upload button', () => {
     it('should be hidden if not editable', () => {
-      let image = renderProfileImage(SETTINGS.user, false);
+      let image = renderProfileImage({profile: SETTINGS.user}, false);
```
noisecapella: `SETTINGS.user` isn't a profile, though this code matches the previous code so ¯\\\_(ツ)\_/¯

aliceriot: oh haha, that's goofy, I'll change it to `USER_PROFILE_RESPONSE`.

static/js/components/PersonalTab.js
aliceriot: sure, what do you think of `uneditedProfile`?

static/js/containers/ProfileImage_test.js
```diff
@@ -1,26 +1,29 @@
...
-          profile={profile} 
-          store={helper.store}
-          editable={editable}
-        />
+        <Provider store={helper.store}>
+          <ProfileImage
+            editable={editable}
+            {...props}
+          />
+      </Provider>
```
aliceriot: oops lol

static/js/components/PersonalTab.js
noisecapella: Sounds good

static/js/containers/ProfileImage_test.js
aliceriot: the JSX indentation for https://github.com/pangloss/vim-javascript is a bit screwy sometimes...

singingwolfboy:
Any chance that we could check if the user has a [Gravatar](http://gravatar.com/) for their email address, and default to that image if it exists? That way, the user can potentially skip a step in the already-too-long signup flow.

Gravatar is a pretty simple service, and the documentation for Gravatar is pretty good. We would need to first [calculate the hash of the email address](https://secure.gravatar.com/site/implement/hash/), and then use that hash to [request the user's image](https://secure.gravatar.com/site/implement/images/). For our use case, we would use the `d=404` option to indicate that Gravatar should return a 404 Not Found response if the user doesn't have an image, to make it easier to detect that case.

If this feature is too complicated to roll into this pull request, I could make a GitHub issue instead, and we can tackle it in a future pull request instead.

aliceriot:
I don't remember exactly why, but we didn't finish integrating gravatar (we have done some work on it in the past). I think that we decided that for this project it doesn't really make sense to use gravatar, but I do not remember exactly why.

Maybe @giocalitri does?

aliceriot:
@noisecapella all comments should be addressed

static/js/components/PersonalTab_test.js
```diff
@@ -9,10 +9,11 @@ import getMuiTheme from 'material-ui/styles/getMuiTheme';
 import PersonalTab from './PersonalTab';
 import { PROGRAMS } from '../constants';
 import IntegrationTestHelper from '../util/integration_test_helper';
+import { USER_PROFILE_RESPONSE } from '../constants';
 
 describe("PersonalTab", () => {
   let helper;
-  let renderPersonalTab = (selectedProgram = null, props = {}) => {
+  let renderPersonalTab = (selectedProgram = null, props = {uneditedProfile: USER_PROFILE_RESPONSE}) => {
```
noisecapella: Can you define this as an explicit prop below instead? Otherwise this would get overwritten when the user passes an argument for `props`

static/js/containers/ProfileImage_test.js
```diff
@@ -1,26 +1,31 @@
...
+import * as api from '../lib/api';
+import { startPhotoEdit, updatePhotoEdit } from '../actions/image_upload';
+import { USER_PROFILE_RESPONSE } from '../constants';
 
 describe('ProfileImage', () => {
-  let helper;
+  let helper, sandbox, updateProfileImageStub, defaultProps;
 
-  const renderProfileImage = (profile, editable=true) => (
+  const renderProfileImage = (editable=true, props = defaultProps) => (
```
noisecapella: Can we set a default `profile` below instead and use `{}` in place of `defaultProps`?

static/js/containers/ProfileFormContainer.js
```diff
@@ -183,7 +183,7 @@ class ProfileFormContainer extends React.Component {
       dispatch,
       currentProgramEnrollment
     } = this.props;
-    let errors, isEdit, profile;
+    let errors, isEdit, profile, uneditedProfile;
 
     if ( profileFromStore === undefined ) {
       profile = {};
```
noisecapella: Should we set `uneditedProfile = {}` here?

static/js/containers/ProfileImage.js
```diff
@@ -30,43 +30,52 @@ const formatPhotoName = photo => (
...
       clearPhotoEdit,
       setDialogVisibility,
+      afterImageUpload,
     } = this.props;
+
     return dispatch(updateUserPhoto(username, edit, formatPhotoName(photo))).then(() => {
       clearPhotoEdit();
       setDialogVisibility(false);
-      this.fetchUserProfile();
+      return this.fetchUserProfile().then(arg => {
```
noisecapella: If we don't care what the arguments are, can we accept `...args` instead and pass it through?

static/js/containers/ProfileFormContainer.js
aliceriot: yeah I think that makes sense

static/js/containers/ProfileImage.js
aliceriot: well, we only care about the first one. I meant to rename this to `resp` or similar.

noisecapella:
Code looks good, just waiting for PR build to test functionality

pdpinch:
We actually started out with gravatar support, but at the time Giovanni and I were the only ones who had gravatars, and we thought the UX for updating your avatar in that scenario was too complex.

noisecapella:
Actually hold up on merging, I got a sentry error. One sec...

noisecapella:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/187536028/

noisecapella:
I'm having trouble reproducing this, I'll just file another issue to deal with it later :+1:

noisecapella:
#1875

Issue #1800, opened 2016-11-17
Exams: add additional profile data to profile model


pdpinch:
part of #1796 

Pearson VUE requires the following fields which don't exist in our current profile model:

- Address1
- PostalCode (required for US and Canada; optional for other countries)
- Phone
- Phone CountryCode

Other fields we don't have that aren't required:

- MiddleName
- Suffix
- Salutation
- Address2
- Address3

All these fields can only contain characters that can be encoded in CP-1252. However, they will be transferred (and stored) as UTF8. 

For more details, see https://drive.google.com/a/pearson.com/file/d/0B5lnP24agPPYN2M5bV9sWHc2VXM

pdpinch:
If we don't force users to use CP-1252, we may also want "shadow" fields for first name, last name and city.

rhysyngsun:
@pdpinch @roberthouse54 to Rob's question in planning, each address field is limited to 40 characters on the Pearson side, it might be best to support all bust only require the first one. I think at this point I'm very much in favor of storing the Pearson profile data separately, but populating it initially from our existing profile.

rhysyngsun:
I'll be adding factories and associated tests as part of this work.

PR #1859, opened 2016-11-23
Nl/1800 add pearson profile

rhysyngsun:

Implements #1800 

#### What's this PR do?

Adds a model for Pearson profiles. This is a separate model from our existing profiles because the requirements on the Pearson side are disjoint with our / edX's requirements. Specifically, Pearson requires all profile data (names, addresses, etc) to be in the character range of [CP-1252](https://en.wikipedia.org/wiki/Windows-1252), albeit UTF-8 encoded. We will initialize these profile with existing profile data we have on file.

I decided to implement this in a new django app as I can foresee exam / Pearson related code being substantial enough that I didn't want to pollute an existing app with it.


rhysyngsun:
Closing this because I'm going to open on a different branch.

PR #1861, opened 2016-11-23
Updated session to use cookie session instead of DB

blarghmatey:
N/A

#### What's this PR do?
Sets the session middleware to store the session in a browser cookie instead of the database.

Make sure that it is still possible to log in and browse the site and that the session is maintained

Browse the site

The get request for the session data was run on every request using a date range query which was causing significant slowdown in the overall performance of the application. This is unnecessary because we don't store anything sensitive in the session data.

The DB session queries are our most expensive query due to date ranges


giocalitri:
functionality looks fine 👍

Issue #1792, opened 2016-11-16
Profile: User Bio


pdpinch:
To assist with social features, users should have a short bio included in their profile. 

Currently, we're thinking that this would be a card only on the profile page.

## Acceptance Criteria

- [x] Add card to user profile
- [x] Users can only edit their own bio
- [x] Users can view learner's bios if they can view learner's profile (same as learner profile page)
- [x] plain text only, except preserve line breaks

pdpinch:
@Ferdi edX includes a bio in their profile. We discard this when we create user accounts. Should we keep it, just in case there's something in it?

pdpinch:
@roberthouse54 can you update this with a mock-up?

roberthouse54:
![about_me 2x](https://cloud.githubusercontent.com/assets/20047260/20366587/cfc0ace2-ac19-11e6-9317-6fba84b2b5f4.png)

PR #1862, opened 2016-11-24
Added 'about me' field to profile

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1792

screen shot 2016-11-24 at 4 20 57 pm

screen shot 2016-11-24 at 9 24 23 pm

screen shot 2016-11-28 at 2 51 03 pm

screen shot 2016-11-28 at 2 51 22 pm

screen shot 2016-11-28 at 2 51 28 pm


static/js/components/UserInfoCard_test.js
```diff
@@ -0,0 +1,72 @@
...
+    defaultRowProps['profile'] = Object.assign(_.cloneDeep(USER_PROFILE_RESPONSE), {
+      about_me: "Hello world"
+    });
+    let wrapper = shallow(<UserInfoCard {...defaultRowProps} />);
+    assert.equal(wrapper.find(".heading").text(), 'About Me');
+    assert.equal(
+      wrapper.find(".bio").text(),
+      "Hello world"
+    );
+  });
```
singingwolfboy: Can you add another test with a multi-line "about me" section? Looks like it's getting rendered properly due to the `white-space: pre-wrap` in the CSS, but we should still have a test for it, since it can be a somewhat finicky case.

static/js/components/UserInfoCard.js
```diff
@@ -26,8 +27,45 @@ export default class UserInfoCard extends React.Component {
     </div>
   );
 
+  renderAboutMeWithData = (profile: Profile): Array<React$Element<*>> => ([
+    <h3 className="heading" key="heading">About Me</h3>,
+    <div className="bio" key="bio">{profile.about_me}</div>
```
aliceriot: when react elements are declared in an array they must have unique `key` props passed in.

static/js/components/UserPageAboutMeDialog.js
```diff
@@ -0,0 +1,77 @@
...
+        contentClassName="dialog about-me-dialog"
+        className="about-me-dialog-wrapper"
+        open={userPageAboutMeDialogVisibility}
+        onRequestClose={this.closeAboutMeDialog}
+        actions={actions}
+        autoScrollBodyContent={true}>
+       {
+         this.boundTextField(
+           ["about_me"],
+           "",
```
singingwolfboy: You are creating a text field without a `label`, which means that a screenreader can't determine what this text field is for. I had thought/hoped that the `hintText` field would be accessible, but it isn't -- I just verified with VoiceOver. Might be easiest to move the `hintText` to the `label`, or maybe duplicate them.

static/js/containers/ProfileFormContainer.js
```diff
@@ -112,15 +113,19 @@ class ProfileFormContainer extends React.Component {
     dispatch(startProfileEdit(username));
   };
 
-  saveProfile(isEdit: boolean, validator: Validator|UIValidator, profile: Profile, ui: UIState) {
+  saveProfile(isEdit: boolean, validator: Validator|UIValidator|null, profile: Profile, ui: UIState) {
```
singingwolfboy: I don't understand: the function is called `saveProfile`, implying that the validator should be for the profile as a whole, not for an individual field. Surely there would be a validator for the profile as a whole, even if that validator skips over an individual field in the profile. Is this function poorly named? Or am I misunderstanding how this works?

static/js/components/UserInfoCard.js
```diff
@@ -26,8 +27,47 @@ export default class UserInfoCard extends React.Component {
...
+    }
+
+    return (
+      <div className="profile-form-row">
+        <div className="about-me">{aboutMeContent}</div>
+        <div className="edit-about-me-holder">
+          {
+            userPrivilegeCheck(
+              profile,
+              () => <IconButton name="edit about me section" onClick={toggleShowAboutMeDialog}/>
```
singingwolfboy: I guess this works -- it renders fine, and the screenreader picks up the full text correctly. It's not the solution I expected, but if it works, then great!

static/js/components/UserPageAboutMeDialog_test.js
```diff
@@ -0,0 +1,90 @@
...
+    assert.equal(document.querySelector("textarea").textContent, "Hello world");
+  });
+
+  it('render dialog without data', () => {
+    renderDialog();
+    assert.equal(document.querySelector('h3.dialog-title').textContent, "About Me");
+    assert.equal(document.querySelector("textarea").textContent, "");
+  });
+
+  it('render dialog when visibility det to false', () => {
```
singingwolfboy: Can we add a test that, when viewing a profile that is _not_ your own, when the "about me" value is unset, the section doesn't show up at all? Or is that what this test is doing? I'm not clear on what this `userPageAboutMeDialogVisibility` property is actually for.

static/js/containers/ProfileFormContainer.js
aliceriot: The `validator` is a function which validates part or all of the profile. `saveProfile` runs the validator and, if there are no errors, it issues the `PATCH` request to save the updated profile.

The reason for the `validator` is that we do not have any restriction on the backend for requiring fields. We need to do this because we fill the profile out in steps, so we want to be able to save some fields, take the user to the next page, and save some more. So we don't always want to validate the whole profile, sometimes we want to validate just the section of interest.

I do think here there should be a `validator` which does not let the user save until they've entered something for 'about me'. That would be in line with the pattern we use everywhere else (that passing a `validator` to `saveProfile` is required).

aliceriot: An example of such a simple validator is the `profileImageValidation` function I implemented recently (in `lib/validation/profile.js`)

static/js/components/UserPageAboutMeDialog_test.js
amir-qayyum-khan: userPageAboutMeDialogVisibility is for show and hide about me dialog

singingwolfboy: OK. Looks like this line has a typo as well: I think you mean "set" instead of "det".

amir-qayyum-khan:
done @singingwolfboy

PR #1870, opened 2016-11-28
fix Virgin Islands labels

### LINKED ISSUE #541: Virgin Islands appears twice in country data set

aliceriot:

closes #541 

#### What's this PR do?

Makes sure that the label is different for the US and the British virgin islands.

Go to `/profile` or similar and make sure that two distinct options come up if you type in 'Virgin Islands' into a country select field

Check that the values updated in the profile are the correct country code for the US and British islands.

helpful:

https://en.wikipedia.org/wiki/United_States_Virgin_Islands

https://en.wikipedia.org/wiki/British_Virgin_Islands

country_stuff_vi


rhysyngsun:
:+1:

PR #1874, opened 2016-11-28
Change a few colors

roberthouse54:
#### What's this PR do?

This is a small PR that changes the color of a few elements on the site:

- the progress card to white, and the bar itself to a lighter shade of the MIT brand color.
- the link color slightly lighter blue
- the button color to blue

Also, a slight tweak to the material card box-shadow

### DIFF
--- a/static/scss/_button.scss
+++ b/static/scss/_button.scss
@@ -1,7 +1,9 @@
 .mdl-button {
 
   font-weight: 400;
-  border-radius: 3px;
+  border-radius: 2px;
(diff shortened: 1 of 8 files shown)


noisecapella:
> the progress card to dark slate gray, and the bar itself to a lighter shade of the MIT brand color

It looks like the box background is white now:
screenshot from 2016-11-29 16-04-31

Is that correct?

roberthouse54:
@noisecapella Yes, that is correct, but the description in the git PR is wrong. I will update it.

Issue #1875, opened 2016-11-28
Exception when user clicks 'Save' before selecting a profile image


noisecapella:
from https://sentry.io/mit-office-of-digital-learning/micromasters/issues/187536028/
and https://sentry.io/mit-office-of-digital-learning/micromasters/issues/187541556/

I think one happens with Firefox, the other happens with Chrome, but it's basically the same problem

PR #1876, opened 2016-11-28
Fix 'save photo' exception

aliceriot:

#1875 

#### What's this PR do?

Basically just ensures that if there isn't a photo selected we don't try to call the function to issue the request.

Open the photo upload dialog, and click 'save' without selecting a photo. You shouldn't see any errors in the JS console.

### DIFF
--- a/static/js/components/ProfileImageUploader.js
+++ b/static/js/components/ProfileImageUploader.js
@@ -76,7 +76,7 @@ const ProfileImageUploader = ({
         type='button'
         className={`save-button ${photo ? 'primary-button' : 'secondary-button disabled'}`}
         key="save"
-        onClick={updateUserPhoto}>
+        onClick={photo ? updateUserPhoto : undefined}>
         Save
       </Button>
     ]}


amir-qayyum-khan:
@aliceriot is this open for review?

aliceriot:
yes, I put the 'needs review' label on it.

amir-qayyum-khan:
@aliceriot save is already unclickable 

screen shot 2016-11-29 at 1 41 34 am

aliceriot:
yeah it is disabled (visually) but the `onclick` handler still fires when it is clicked. click the button with the JS console open, you'll see an exception (on `master`).

amir-qayyum-khan:
good to go 👍

static/js/components/ProfileImageUploader.js
```diff
@@ -76,7 +76,7 @@ const ProfileImageUploader = ({
         type='button'
         className={`save-button ${photo ? 'primary-button' : 'secondary-button disabled'}`}
         key="save"
-        onClick={updateUserPhoto}>
+        onClick={photo ? updateUserPhoto : undefined}>
```
noisecapella: Wouldn't the `undefined` get executed as a function in this case?

noisecapella: I guess not actually, in HTML if the `onclick` is missing it's treated as if the function weren't there

aliceriot: I think that under the hood `props` are just objects, so `undefined` looks like the button didn't receive an `onClick` handler at all.

aliceriot: It doesn't cause an error or anything

PR #1877, opened 2016-11-28
Update a couple of JavaScript packages (eslint, jsdom, moment)

aliceriot:
This updates eslint and the related plugin packages, turns on a few eslint rules relating to flow, and updates jsdom and moment.

Read over the changes, and assuming that you agree with the rule changes make sure that there are no regressions.


aliceriot:
JSDom update was causing problems, so I'm leaving that off for now.

static/js/financial_aid/functions.js
```diff
@@ -126,7 +126,7 @@ window.financialAidReview = (function(window, document, $) {
    */
   function initiateSearch() {
     var searchQuery = $("#search-query").val();
-    location = BASE_PATH + searchQuery;
+    location = BASE_PATH + searchQuery; // eslint-disable-line no-global-assign
```
singingwolfboy: Might be clearer to do `window.location = BASE_PATH + searchQuery;`, which would avoid the need for the eslint disable.

aliceriot: good idea!

aliceriot:
ok so I hate npm but npm-shrinkwrap is sorted (T_T)

static/js/util/util_test.js
```diff
@@ -1,4 +1,5 @@
 /* global SETTINGS: false */
+/* eslint-disable no-global-assign */
```
noisecapella: You should be able to remove this since another PR today removed the global assign of `SETTINGS`

noisecapella:
:+1: after addressing comment

PR #1880, opened 2016-11-29
Remove SETTINGS.username, update tests

### LINKED ISSUE #1869: Add tests for value of START_PROFILE_EDIT
We should test for this https://github.com/mitodl/micromasters/pull/1858/files#diff-92a6563e47b379f3b8fb212041af925dR111

And also fix a similar bug and add a similar test in `fetchUserProfile` in App.js

noisecapella:
Fixes #1869 

#### What's this PR do?
Fixes `SETTINGS.username` reference that we forgot to remove earlier, adds tests to cover that case

### DIFF
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -96,7 +96,7 @@ class App extends React.Component {
     if (userProfile.getStatus === undefined) {
       dispatch(fetchUserProfile(username)).then(() => {
         if ( PROFILE_REGEX.test(pathname) ) {
-          dispatch(startProfileEdit(SETTINGS.username));
+          dispatch(startProfileEdit(SETTINGS.user.username));
         }
       });
     }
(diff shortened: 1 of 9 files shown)


gsidebo:
code looks good. do you mind explaining a little more about what this is doing? the linked issue doesn't reveal much

noisecapella:
We removed `SETTINGS.username` a little while back in favor of `SETTINGS.user.username` but we missed updating the reference in `App.js`. This fixes that bug and also updates the tests so that using the incorrect username like `/api/v0/profiles/undefined/` will cause an error in the tests

noisecapella:
To verify that the test changes work, you can revert the change to `App.js` or a similar one mentioned in #1869 and some tests should break.

gsidebo:
got it. looks great! 👍

PR #1882, opened 2016-11-29
Hide name, email, program fields in Zendesk

singingwolfboy:
Refs #1262. Hides the "name", "email", and "Micromasters program" fields in the Zendesk widget. Note that this will hide the fields ~~every time, regardless of whether they are pre-filled or not~~ that have values pre-set, and display the others.

zendesk-hidden-fields

### DIFF
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -135,10 +135,8 @@ const zendeskCallbacks = {
       'name':  !(SETTINGS.user && SETTINGS.user.first_name
                  && SETTINGS.user.last_name),
       'email': !(SETTINGS.user && SETTINGS.user.email),
+      [programFieldName]: !(SETTINGS.program && SETTINGS.program.slug),
     };
-    fieldVisibility[programFieldName] = !(
-      SETTINGS.program && SETTINGS.program.slug
-    );
 
     const adjustFieldsVisibility = R.map(name => {
       if ( !fieldVisibility[name] ) {
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -125,11 +125,8 @@ const zendeskCallbacks = {
     const fieldSelector = name => (
       `input[name="${name}"], select[name="${name}"]`
     );
-    const iframeQuerySelector = query => (
-      iframe.contentDocument.querySelector(query)
-    );
-    const fieldElement = R.compose(
-      iframeQuerySelector, fieldSelector
+    const fieldElement = name => (
+      iframe.contentDocument.querySelector(fieldSelector(name))
     );
 
     // Zendesk uses the ID 24690866 to refer to the MicroMasters program selector
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -1,5 +1,6 @@
 /* global SETTINGS:false zE:false _:false */
 __webpack_public_path__ = `http://${SETTINGS.host}:8078/`;  // eslint-disable-line no-undef, camelcase
+import "element-closest";
 
 // Start of odl Zendesk Widget script
 /*<![CDATA[*/
@@ -116,6 +117,35 @@ const zendeskCallbacks = {
         }, 100);
       };
     }
+  },
+  ticketSubmissionFormLoaded: () => {
+    const iframe = document.querySelector("iframe.zEWidget-ticketSubmissionForm");
+    // field 24690866 is the MicroMasters program selector
+    const fieldNames = ["name", "email", "24690866"];
+    const fieldHidden = [
+      SETTINGS.user && SETTINGS.user.first_name && SETTINGS.user.last_name,
+      SETTINGS.user && SETTINGS.user.email,
+      SETTINGS.program && SETTINGS.program.slug,
+    ];
+    const fields = _.map(
+      fieldNames,
+      (name) => {
+        return iframe.contentDocument.querySelector(
+          `input[name="${name}"], select[name="${name}"]`
+        );
+      }
+    );
+    _.zip(fields, fieldHidden).forEach((values) => {
+      const [ field, isHidden ] = values;
+      if (isHidden) {
+        const label = field.closest('label');
+        label.style.setProperty("display", "none", "important");
+      }
+    });
+    // adjust the iframe to the correct (smaller) height
+    const height = iframe.contentDocument.body.childNodes[0].offsetHeight;
+    // zendesk adds a 15px margin
+    iframe.style.height = `${height + 15}px`;
   }
 };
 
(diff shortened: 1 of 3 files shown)


singingwolfboy:
@aliceriot suggested using the [element-closest polyfill](https://www.npmjs.com/package/element-closest), but I'm not sure the best way to set it up. [According to caniuse](http://caniuse.com/#feat=element-closest), the `closest()` method isn't supported in Internet Explorer or Edge.

Ferdi:
> Note that this will hide the fields every time, regardless of whether they are pre-filled or not.

That's not good.

singingwolfboy:
@Ferdi Would you prefer that the field shows up only when it is empty? I thought that you were trying to prevent the user from needing to know that their information was being collected at all.

Having the fields sometimes show up and sometimes not seems like a confusing user experience to me. In particular, if the user knows that the information is there but invisible, they have no way to confirm that their information is correct.

Would you like me to make these fields show up only when empty? What should happen if some fields are prefilled (like name and email) while others are not (like program)?

Ferdi:
>  I thought that you were trying to prevent the user from needing to know that their information was 
> being collected at all.

I think the primary goals are a) Know who is contacting us b)  Simplify UI/X for users 

I suggest show up fields only when they are empty

static/js/entry/zendesk_widget.js
```diff
@@ -116,6 +117,27 @@ const zendeskCallbacks = {
...
+      hiddenFieldNames,
+      (name) => {
+        return iframe.contentDocument.querySelector(
+          `input[name="${name}"], select[name="${name}"]`
+        )
+      }
+    );
+    for (const field of hiddenFields) {
+      // hide each field
+      field.closest('label').style.setProperty("display", "none", "important");
```
singingwolfboy: The [`Element.closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) method returns the closest ancestor of the current element (or the current element itself) which matches the selectors given in parameter. If there isn't such an ancestor, it returns `null`.

It's the native equivalent of jQuery's [`.parents()`](http://api.jquery.com/parents/) method.

aliceriot:
@singingwolfboy going by https://github.com/jonathantneal/closest/blob/master/element-closest.js I think you just need to require the library. It should take care of the rest.

Do this in whichever webpack entry point script is relevant.

aliceriot:
Also the same issue is happening here with `npm-shrinkwrap.json`

singingwolfboy:
I don't know why npm keeps changing its mind about the format of the `npm-shrinkwrap.json` file -- it's very confusing. At any rate, @Ferdi I've changed this pull request so that only pre-filled fields are hidden.

static/js/entry/zendesk_widget.js
```diff
@@ -116,6 +117,35 @@ const zendeskCallbacks = {
...
+        );
+      }
+    );
+    _.zip(fields, fieldHidden).forEach((values) => {
+      const [ field, isHidden ] = values;
+      if (isHidden) {
+        const label = field.closest('label');
+        label.style.setProperty("display", "none", "important");
+      }
+    });
```
aliceriot: the implementation of this is a bit confusing to me. It seems like what we're doing is that we have three different fields ('name', 'email', and '24690866'), which have differing conditions for when they are hidden.

What do you think about breaking this up into a series of functions? Something like:

```js
const fieldSelector = name => (
  `input[name="${name}"], select[name="${name}"]`
);

const fieldElement = R.compose(
  iframe.contentDocument.querySelector, fieldSelector
);

const fieldVisibility = {
  'key': visibility // true | false
  ...
};


const adjustFieldsVisibility = R.map(name => {
  if ( !fieldVisibility(name) ) {
    const label = fieldElement(name).closest('label');
    label.style.setProperty("display", "none", "important");
  };
});

adjustFieldsVisibility([
  'name', 'email', '24690866'
]);
```

static/js/entry/zendesk_widget.js
```diff
@@ -116,6 +118,46 @@ const zendeskCallbacks = {
...
       };
     }
+  },
+  ticketSubmissionFormLoaded: () => {
+    const iframe = document.querySelector("iframe.zEWidget-ticketSubmissionForm");
+
+    const fieldSelector = name => (
+      `input[name="${name}"], select[name="${name}"]`
+    );
+    const iframeQuerySelector = query => (
```
aliceriot: oh! yeah I've run into that before. also not sure, haha.

static/js/entry/zendesk_widget.js
```diff
@@ -116,6 +118,46 @@ const zendeskCallbacks = {
...
+    // Zendesk uses the ID 24690866 to refer to the MicroMasters program selector
+    const programFieldName = '24690866';
+    const fieldVisibility = {
+      'name':  !(SETTINGS.user && SETTINGS.user.first_name
+                 && SETTINGS.user.last_name),
+      'email': !(SETTINGS.user && SETTINGS.user.email),
+    };
+    fieldVisibility[programFieldName] = !(
+      SETTINGS.program && SETTINGS.program.slug
+    );
```
aliceriot: is there a reason this one can't be included in the object literal above?

static/js/entry/zendesk_widget.js
```diff
@@ -116,6 +118,46 @@ const zendeskCallbacks = {
...
+
+    const fieldSelector = name => (
+      `input[name="${name}"], select[name="${name}"]`
+    );
+    const iframeQuerySelector = query => (
+      iframe.contentDocument.querySelector(query)
+    );
+    const fieldElement = R.compose(
+      iframeQuerySelector, fieldSelector
+    );
```
aliceriot: given the workaround here it might be a little more elegant to just do:

```js
const fieldElement = name => (
  iframe.contentDocument.querySelector(fieldSelector(name))
);
```

up to you though!

static/js/entry/zendesk_widget.js
singingwolfboy: Yes, this is a bug that bit me while I was developing this feature. Let's say you have this code:
```python
variable = "value"
obj = {
   "literal": "foo",
   variable: "bar",
}
```
If you run that in Python, you'll end up with an object that looks like this:
```python
{"literal": "foo", "value": "bar"}
```
But if you run that in JavaScript, you'll get this:
```js
{"literal": "foo", "variable": "bar"}
```

I had to do this as a workaround instead. I wish there was a way to do this properly in Javascript, specifying that you want it to evaluate the variable for the key.

aliceriot: There is!

```js
const myKey = 'a key';

const obj = {
  otherKey: 'other value',
  [myKey]: 'my value',
};
```

the syntax feature is called a computed property name, there's some info here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

aliceriot: Basically you just wrap it in square brackets.

singingwolfboy: Amazing

noisecapella:
Please delete the branch after you've merged the PR @singingwolfboy

PR #1886, opened 2016-11-29
Updated constants to fix grade and to add a currently enrolled course

noisecapella:
None

#### What's this PR do?
Updates `grade` to `current_grade` or `final_grade`, makes the price four digits, and adds a course run which is currently enrolled

Update api.js to return the constants instead of making API requests, then view the dashboard.


aliceriot:
Can't seem to get the API changes to work right, what am I missing? This is what I've done:

```
diff --git a/static/js/lib/api.js b/static/js/lib/api.js
index 1abbdbf..5035ab3 100644
--- a/static/js/lib/api.js
+++ b/static/js/lib/api.js
@@ -10,6 +10,13 @@ import type { Dashboard } from '../flow/dashboardTypes';
 import type { AvailableProgram, AvailablePrograms } from '../flow/enrollmentTypes';
 import type { EmailSendResponse } from '../flow/emailTypes';
 
+import {
+  USER_PROFILE_RESPONSE,
+  DASHBOARD_RESPONSE,
+  COURSE_PRICES_RESPONSE,
+  USER_PROGRAM_RESPONSE
+} from '../constants';
+
 export function getCookie(name: string): string|null {
   let cookieValue = null;
 
@@ -149,7 +156,9 @@ export function patchUserProfile(username: string, profile: Profile): Promise<Pr
 }
 
 export function getDashboard(): Promise<Dashboard> {
-  return mockableFetchJSONWithCSRF('/api/v0/dashboard/', {}, true);
+  // return mockableFetchJSONWithCSRF('/api/v0/dashboard/', {}, true);
+  return Promise.resolve(DASHBOARD_RESPONSE);
+  
 }
 
 export function checkout(courseId: string): Promise<CheckoutResponse> {
@@ -173,7 +182,8 @@ export function sendSearchResultMail(subject: string, body: string, searchReques
 }
 
 export function getPrograms(): Promise<AvailablePrograms> {
-  return mockableFetchJSONWithCSRF('/api/v0/programs/', {}, true);
+  // return mockableFetchJSONWithCSRF('/api/v0/programs/', {}, true);
+  return Promise.resolve(USER_PROGRAM_RESPONSE);
 }
 
 export function addProgramEnrollment(programId: number): Promise<AvailableProgram> {
@@ -209,7 +219,8 @@ export function addFinancialAid(income: number, currency: string, programId: num
 }
 
 export function getCoursePrices(): Promise<*> {
-  return mockableFetchJSONWithCSRF('/api/v0/course_prices/', {});
+  // return mockableFetchJSONWithCSRF('/api/v0/course_prices/', {});
+  return Promise.resolve(COURSE_PRICES_RESPONSE);
 }
 
 export function skipFinancialAid(programId: number): Promise<*> {
```

noisecapella:
`USER_PROGRAM_RESPONSE` is a different thing, I think you want `PROGRAMS` instead

aliceriot:
ah! haha. trying that now.

aliceriot:
ok got it.

It looks like you edited 'not passed program'. when I select it this is what I see:

not_passed

is that right?

noisecapella:
I edited one run for not passed program but the others should be in master program. I'll double check what the output should be for not passed program, I feel like some grade should be displayed there

noisecapella:
So for passed past courses we don't actually show the grade at all. I updated `constants.js` to also put a grade on the failed course, which does show up in the UI

PR #1893, opened 2016-11-30
Improve accessibility text for edit profile button

singingwolfboy:
Improve accessibility text for edit profile button

Activate VoiceOver or ChromeVox, and tab to the edit buttons on your user profile page. The screen reader should clearly explain what these buttons do.


rhysyngsun:
:+1:

PR #1894, opened 2016-11-30
Update selector for Zendesk launcher button

singingwolfboy:
Zendesk changed the HTML that they use for their launcher button. As a result, the JS that I wrote to change the background color of the button broke. This pull request restores the background color that we want for the launcher button.


aliceriot:
is this what it is supposed to look like?

zendesk_button

it's invisible against the footer

singingwolfboy:
@aliceriot: yes, that's correct. @roberthouse54 is in charge of the design, so if there's a problem there, you should probably talk to him.

aliceriot:
hum ok, well I'm ok with merging this, I'll file an issue for the invisibility.

Issue #1662, opened 2016-11-04
Performance: add expires headers


pdpinch:
Another issue identified by yslow. Without expires headers, browsers may not cache our static assets for as long as they could. 

```
(no expires) https://micromasters.herokuapp.com/static/bundles/sentry_client.js
(no expires) https://micromasters.herokuapp.com/static/images/mit-logo-transparent.svg
(no expires) https://micromasters.herokuapp.com/static/images/edx_logo.png
(no expires) https://micromasters.herokuapp.com/static/images/micromasters_logo.png
(no expires) https://micromasters.herokuapp.com/static/bundles/public.js
(no expires) https://micromasters.herokuapp.com/static/bundles/style_public.js
(no expires) https://micromasters.herokuapp.com/static/images/lp_hero.jpg
(no expires) https://bam.nr-data.net/1/123785e709?...
(no expires) https://micromasters.herokuapp.com/static/images/online.png
(no expires) https://micromasters.herokuapp.com/static/images/blended.png
(no expires) https://micromasters.herokuapp.com/static/images/cost.png
(no expires) https://micromasters.herokuapp.com/static/images/reif1.png
(no expires) https://micromasters.herokuapp.com/static/images/more-coming-soon.png
(no expires) https://micromasters.herokuapp.com/static/images/sanjay1.png
(no expires) https://odl-micromasters-production.s3.amazonaws.com/images/scm_thumb.2e16d0ba.fill-690x530.jpg?...
(no expires) https://odl-micromasters-production.s3.amazonaws.com/images/econ_thumb.2e16d0ba.fill-690x530.jpg?...
(no expires) https://micromasters.herokuapp.com/static/images/mit-logo-ltgray-white@72x38.svg
(no expires) https://micromasters.herokuapp.com/static/bundles/style.js
(no expires) https://micromasters.herokuapp.com/static/bundles/signup_dialog.js
(no expires) https://micromasters.herokuapp.com/static/images/favicon.ico
```

PR #1897, opened 2016-11-30
Turned on webpack stats generation, use stats for filenames, use templatetags

noisecapella:
Part of #1662

#### What's this PR do?
Bundles now have a hash concatenated on the filename for production builds. This will allow us to set a far future expiration header for these assets which will make it easier to deploy on a CDN.

This also removes `get_bundle_url` and uses a template tag in its place

ui/templatetags/render_bundle.py

Make sure that the JS bundles are still properly rendered for each page, for both development and production builds. Nothing should break. Hot reloading should still work.

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -12,14 +12,15 @@
     var SETTINGS = {{ js_settings_json|safe }};
     var CURRENT_PAGE_URL = '{{ request.build_absolute_uri }}';
     </script>
-    <script type="text/javascript" src="{{ common_src }}"></script>
-    <script type="text/javascript" src="{{ sentry_client }}"></script>
-    {% if style_public_src %}
-    <script type="text/javascript" src="{{ style_public_src }}"></script>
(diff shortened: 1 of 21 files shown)


aliceriot:
Things are working locally and on the PR build, HMR is still working. Looking at code!

ui/templates/base.html
```diff
@@ -12,14 +12,15 @@
...
+    {% render_bundle "sentry_client" %}
+    {% if is_public %}
+      {% render_bundle "style_public" %}
     {% endif %}
-    <script type="text/javascript" src="{{ style_src }}"></script>
-    {% if zendesk_widget %}
-      <script type="text/javascript" src="{{ zendesk_widget }}"></script>
+    {% render_bundle "style" %}
+    {% if has_zendesk_widget %}
+      {% render_bundle "zendesk_widget" %}
```
aliceriot: this is a much cleaner pattern overall!

PR #1899, opened 2016-11-30
Send ImproperlyConfigured exceptions to Sentry

### LINKED ISSUE #1817: Better error handling of enrollment errors
When an enrollment fails (for example, if a course key is incorrect, of if the enrollment end date has passed), edX is returning a 400 error, which leads our enrollment API to raise a 500.

The error is logged:

`Error creating audit enrollment for course key course-v1:MITx+CTL.SC0x+3T2016 for user Jared`

-- but it doesn't get raised to sentry. Since these errors are caused by configuration problems, I'd like to have them show up in Sentry. 

See also #1815 Notify users on enrollment errors.

giocalitri:
closes #1817 

#### What's this PR do?
Sends `ImproperlyConfigured` exceptions (and subclasses of it) to Sentry even if they are handled.

`dashboard/views.py`

configure the `SENTRY_DSN` environment variable, try to make an enrollment for a course with closed enrollments on edX but open on micromasters and verify that the exception appears on Sentry.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -5,6 +5,7 @@ import logging
 
 from django.conf import settings
 from edx_api.client import EdxApi
+from requests.exceptions import HTTPError
 from rest_framework import (
     authentication,
     permissions,
@@ -18,6 +19,7 @@ from backends import utils
 from backends.edxorg import EdxOrgOAuth2
 from dashboard.api import get_user_program_info
 from dashboard.api_edx_cache import CachedEdxUserData
+from micromasters.exceptions import PossiblyImproperlyConfigured
 from profiles.api import get_social_username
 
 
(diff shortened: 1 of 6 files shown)


dashboard/views.py
```diff
@@ -92,8 +94,18 @@ def post(self, request):  # pylint: disable=no-self-use
...
+                raise PossiblyImproperlyConfigured(
+                    'Got a 400 status code from edX server while trying to create '
+                    'audit enrollment. This might happen if the course is improperly '
+                    'configured on MicroMasters. Course key '
+                    '{course_key}, edX user "{edX_user}"'.format(
+                        edX_user=get_social_username(request.user),
+                        course_key=course_id,
+                    )
+                )
             log.error(
```
noisecapella: Can we use `log.exception` here instead?

dashboard/views_test.py
```diff
@@ -238,13 +239,33 @@ def test_enrollment_fails(self, mock_refresh, mock_edx_enr):  # pylint: disable=
...
+        mock_edx_enr.side_effect = error
         resp = self.client.post(self.url, {'course_id': self.course_id}, format='json')
         assert resp.status_code == status.HTTP_500_INTERNAL_SERVER_ERROR
         assert mock_edx_enr.call_count == 1
         # assert just the second argument, since the first is `self`
         assert mock_edx_enr.call_args[0][1] == self.course_id
 
+    @patch('edx_api.enrollments.CourseEnrollments.create_audit_student_enrollment', autospec=True)
+    @patch('backends.utils.refresh_user_token', autospec=True)
+    def test_enrollment_fails_configuration(self, mock_refresh, mock_edx_enr):  # pylint: disable=unused-argument
```
noisecapella: It looks like this is basically the same test as above. Can we combine the two tests?

micromasters/exceptions.py
```diff
@@ -0,0 +1,11 @@
+"""
+Custom exceptions for the micromasters app
```
noisecapella: Nit: Capitalize as MicroMasters please

giocalitri:
Comments addressed

PR #1900, opened 2016-11-30
Added new section to the home page for "how it works"

roberthouse54:
#### What's this PR do?

This PR adds a new section to the home page called "How It Works".

Please verify that the layout and css in this section look correct... especially the ::after selectors and absolute positioning of the arrow images.


cms/templates/cms/home_page.html
```diff
@@ -75,7 +75,46 @@
         </div>
       </div>
 
-            <section class="row current-programs-section">
+      <div class="row how-it-works-section">
```
singingwolfboy: Is there a particular reason why you're using `<div>` rather than `<section>`? The latter is more semantically meaningful.

cms/templates/cms/home_page.html
```diff
@@ -75,7 +75,46 @@
...
+
+      <div class="centered-content">
+
+        <h3>How It Works</h3>
+        <div class="hiw col-md-3">
+          <div class="reasons-images-wrap">
+             <img src="{% static 'images/hiw1.png' %}" alt="">
+          </div>
+
+          <h5>Online Learning</h5>
```
singingwolfboy: "How It Works" is in a `<h3>`, so the next heading level should be `<h4>`. Please adjust all the headings in this section.

cms/templates/cms/home_page.html
```diff
@@ -75,7 +75,46 @@
...
+
+          <h5>Proctored Tests</h5>
+          <p>Pass one (or more) proctored exams—either a single final exam, or one exam for each course, depending on the program.</p>
+
+        </div>
+        <div class="hiw col-md-3">
+          <div class="reasons-images-wrap">
+            <img src="{% static 'images/hiw3.png' %}" alt="">
+          </div>
+           <h5>Earn a Valuable Credential</h5>
```
singingwolfboy: Spacing is off.

noisecapella:
screenshot from 2016-11-30 15-30-06

The wording around "Master's degree" seems clumsy

noisecapella:
Other than that everything looks fine to me

cms/templates/cms/home_page.html
roberthouse54: OK, fixed.

cms/templates/cms/home_page.html
roberthouse54: Fixed it.

cms/templates/cms/home_page.html
roberthouse54: This is what we had before... can you create a new issue for this, as it is a semantic/accessibility issue, and I want to get this change out as soon as possible.

singingwolfboy: It looks to me like this is new content you're adding, and that we were using `<section>` before for this... section. Nonetheless, I've created #1917 to track this.

roberthouse54: We have not been consistent unfortunately. I basically copied the benefits section that was not using <section>. I just didn't want to break anything before the release.

PR #1903, opened 2016-11-30
Replace mock with unittest.mock

singingwolfboy:
The [`mock`](http://www.voidspace.org.uk/python/mock/) library is bundled in Python 3 as [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html), so we don't need to install the library from PyPI.


giocalitri:
👍 if tests pass

PR #1905, opened 2016-11-30
Switch to yarn package manager instead of npm

### LINKED ISSUE #1884: switch to yarnpkg instead of using straight npm
I know, I know, JavaScript hipster status and all, but yarn is actually a much nicer package manager than npm.

It:

- has 100% deterministic reproducible installs
- is really, really fast
- its dependency resolution algorithm is more aggressive, and results in a flatter dependency graph and a smaller `node_modules` dir
- has a nice lockfile format which has less churn (more readable and check-able) than `npm-shrinkwrap.json`.

I think we see a lot of pain with npm, and upgrading a single package sometimes takes a bit of time to get the shrinkwrap sorted correctly. Then the huge shrinkwrap diff (which does an ok job of getting us repeatable installs) isn't really reviewable by hand (because it's so big).

Anyway, thoughts? @noisecapella @gsidebo @singingwolfboy

aliceriot:
closes #1884 

this pr:

- installs and configures `yarn` in the watch docker container
- checks in a `yarn.lock` file and deletes `npm-shrinkwrap.json`
- sets up the travis docker containers to use yarn

Merging this will require a little bit of follow-up. I changed the `Dockerfile-travis-watch-build` file to install yarn, and then built a new image which is up on docker hub tagged `mm_watch_travis_yarn`. I changed the `Dockerfile-travis-watch` to inherit from this image, so that there's no issue with building this PR before it's been merged. Once the PR is merged and all open PRs are rebased over master we can introduce a commit to undo this change, by pushing a new yarn image to `mm_watch_travis` and then changing `Dockerfile-travis-watch` back to inherit from that.

#### testing

mac:
- install yarn 0.17.9 from npm (`npm install -g yarn@0.17.9`)
- make sure you can install everything and that the app works and so on

linux:
- delete `node_modules`
- pull down the branch and rebuild your `watch` container (`docker-compose build watch`)
- ensure you can boot the app
- ensure you can run some yarn commands (like `yarn install --pure-lockfile`)

More info on yarn CLI usage here: https://yarnpkg.com/en/docs/yarn-workflow


package.json
```diff
@@ -93,7 +93,7 @@
     "sanctuary": "^0.11.1",
     "sass-lint": "^1.10.2",
     "sass-loader": "3.2.0",
-    "searchkit": "^0.10.0",
+    "searchkit": "^0.10.1",
```
noisecapella: Can we do this change in a separate PR, just in case?

noisecapella:
Works fine for me. I think we should wait until after the release to merge this, just to reduce the risk of interfering with last minute fixes

aliceriot:
I agree. There's also a problem with the `codecov` command failing even through the tests aren't, I suspect it has to do with the dependency resolution algorithm pulling in a different version of something but I haven't tracked it down yet. Looks like it's related to searchkit...

pdpinch:
As mentioned, we don't want to merge this before the release on Monday

aliceriot:
yep, we're on the same page about that (see above).

aliceriot:
@noisecapella would you mind taking a look at this again? I had to pin the searchkit version explicitly to get it to pass, but it's working now.

noisecapella:
Is the `.npm` directory used anymore? I see a reference to it on `.travis.yml`, I'm wondering if we should remove it

noisecapella:
There's a reference to `npm rebuild` in the README, does that need to be updated or is it compatible with yarn?

noisecapella:
There is a reference to `npm install` in the README, we should update that at least

aliceriot:
Oh I think that cache directory was erroneous anyway, I think since I set up the docker stuff it wouldn't have been used (since the cache for npm is inside the docker container). I'll take that out.

I'll change the references to `npm install`. Yarn is very fast when doing reinstalls, so I imagine if there was a build error (such as what we point out in the README with node-sass) the thing to do it just another `yarn install --pure-lockfile`.

aliceriot:
I updated the README with some info about using Yarn and removed that cache thing.

noisecapella:
:+1: Can you notify everyone on what they need to do to update their systems in slack?

aliceriot:
Yes, will do. Hooray yarn! So excited to never mess with `npm-shrinkwrap` again.

aliceriot:
Checking with Tobias about what needs to be changed on the heroku end of things. Once we get that sorted I'll merge this. I updated the commit message with a quick intro to yarn and the differences from npm, I'll point people to that when I merge it.

pdpinch:
What's going to happen after this is merged -- for developers, for Travis and for heroku. 

FYI @blarghmatey

aliceriot:
I wrote a bunch of information about the implications in the commit message. This is not going to be merged until we have the heroku end of things sorted out. You can read the details in the commit message.

aliceriot:
@noisecapella I had to upgrade node-sass in order to fix a build issue, but things seem to be working and heroku now supports yarn (so the builds are working ok). Would you mind taking a quick look before I merge?

noisecapella:
LGTM :+1:

PR #1906, opened 2016-11-30
Set CELERY_ALWAYS_EAGER to default to False

### LINKED ISSUE #1904: Set CELERY_ALWAYS_EAGER to False by default

noisecapella:
Fixes #1904


rhysyngsun:
@noisecapella is the issue here that this should really only be on for tests (that's about as much as I could glean from what little is documented about this setting)?

noisecapella:
Yes, exactly. When it's set to true it causes the celery tasks to execute in the web container which is useful for testing but it makes the docker container less of a real world environment

rhysyngsun:
:+1:

PR #1907, opened 2016-11-30
Added fields to Profile for address and roman name

### LINKED ISSUE #1800: Exams: add additional profile data to profile model

rhysyngsun:

Fixes #1800 

#### What's this PR do?

Adds the new fields to our existing `Profile` model. Most of these are generally useful to us (location fields). The structuring of this data is based off the constraints that Pearson has on the data, which we are assuming are reasonable.

General fields that were added:

 - `address1`
 - `address2`
 - `address3`
 - `phone_number`
 - `phone_country_code`
 - `date_updated`

Two fields were added to cover romanized names:

 - `romanized_first_name`
 - `romanized_last_name`

`profiles/models.py`

Pearson requires profile data to fit in the [CP-1252](https://en.wikipedia.org/wiki/Windows-1252) which basically means we can't send them any non-romanized profile data. In another ticket we'll be handling this validation at the UI level before we ship the data over to Pearson.


pdpinch:
Are the field lengths all derived from Pearson requirements?

rhysyngsun:
@pdpinch for the new fields yes, existing fields are longer than Pearson maximums, so we'll need to validate those outside of the db.

rhysyngsun:
Fields changed and rebased off reasonably recent master.

profiles/factories.py
```diff
@@ -43,10 +44,21 @@ class ProfileFactory(DjangoModelFactory):
...
+    address2 = LazyAttribute(lambda x: FAKE.secondary_address())
+    address3 = None
+
+    postal_code = LazyAttribute(lambda x: FAKE.postcode())
     city = LazyAttribute(lambda x: FAKE.city())
     country = LazyAttribute(lambda x: FAKE.country_code())
     state_or_territory = LazyAttribute(lambda x: FAKE.state())
 
+    phone_number = LazyAttribute(lambda x: FAKE.numerify('###-###-####'))
+    phone_country_code = FuzzyInteger(0, 999)
```
gsidebo: i'm guessing that it's intentional that we aren't using `FAKE.country_code()` for this. what's the difference between our idea of country codes and pearsons?

gsidebo:
is there anything i should be testing aside from code style and successful data migration here? should i be confirming these field specs with the CP-1252 spec in the link above?

profiles/models.py
```diff
@@ -142,6 +180,8 @@ class Profile(models.Model):
     linkedin = JSONField(blank=True, null=True)
     student_id = models.IntegerField(blank=True, null=True, unique=True)
 
+    date_updated = models.DateTimeField(blank=True, null=True, auto_now=True)
```
rhysyngsun: I wasn't aware of that, but if we have a standard way of doing this I'm all in favor of it. I'll update with that change.

rhysyngsun:
@gsidebo we're not bothering with CP-1252 in the data model - it'd cause encoding headaches and Pearson requires data to be in the CP-1252 charset _range_ but _encoded_ in UTF-8. From That a bit whacky, but from what we gather, it's because there's some newer systems on their end that handle UTF-8 data , but probably some legacy system that causes CP-1252 to be the least common denominator.

profiles/models.py
rhysyngsun: @gsidebo Hm, on further inspection, `TimestampedModel` doesn't declare it's fields as nullable which is problematic for a table that already has records. There also is already a `date_joined_micromasters` which is `auto_now_add` so adding `created_on` that you'd get from `TimestampedModel` would be repetitive. The good news there is that we would have values for `created_on` for existing records should we wish to switch to that.

I'd suggest as an intermediate step that I switch the definition of this new field to match `created_on` in `TimestampedModel` and then file a new issue that we should normalize `Profile` to `TimestampedModel` since that's nontrivial and should be tackled separately from the work I'm doing here.

Thoughts?

profiles/factories.py
gsidebo: got it. i'm guessing that's a value that we fill in on our end so we wouldn't need to do any special validation on it

profiles/models.py
gsidebo: sounds great. let's do that

rhysyngsun: @gsidebo I'm proposing changing only `date_updated` -> `updated_on` and to deal with `date_joined_micromasters` -> `created_on` and `Model` -> `TimestampedModel` in a different issue since that will require some migration of existing data and backpopulating `updated_on` so that we can make the switch.

profiles/factories.py
rhysyngsun: Actually, your comment here made me scrutinize this field a little further. I've changed the type on the model to a `CharField` because it can have leading zeroes so the previous type was inappropriate.

gsidebo:
looks good 👍

PR #1890, opened 2016-11-29
Fixed alter data commands

gsidebo:
#### What are the relevant tickets?

More fixes related to #1833 
Rebased on #1833 

#### What's this PR do?

The refactored edX data (#1806) broke our seed data functionality. #1833 fixes up the seed data functionality, and this PR fixes up the alter_data management commands specifically

#### How should this be manually tested?

With seed data loaded, run `./manage.py alter_data --usage` and/or `./manage.py alter_data --help` for usage details. Run through a few of the commands and check the /dashboard page for the Analog Program to make sure the results are what you expect

Issue #933, opened 2016-08-29
Search: Courses facet 


pdpinch:
![image](https://cloud.githubusercontent.com/assets/20047260/20535693/dfc1c45c-b0b3-11e6-864d-decd45cfe97e.png)

Clicking on a course filters by that course, and shows all users who are currently taking the course, or who have taken the course in the past.

## Acceptance Criteria:

- [ ] Add courses to elastic search index
- [ ] The courses facet is the first facet
- [ ] This will be an "and" facet, so users can only choose one of the items from the list. 
- [ ] It will include all the courses (not course runs) in the current program. 
- [ ] If possible, sort by course order

In later issues we want to expand an area under the course name, with 2 additional facets: the "Grade" facet (#1711), which filters by the grade in that course, and the status facet (#1730), which allows users to filter by whether the student has paid for the course or is auditing.

## See Also

#1710 semester facet
#1711 grade facet
#1730 status facet

roberthouse54:
<img width="700" alt="upload" src="https://cloud.githubusercontent.com/assets/20047260/20190174/5e2c9444-a74e-11e6-8b74-3c947a89cbef.png">

Clicking on a course filters by that course, and also expands an area under the course name, with 2 additional facets: the "Grade" facet (#1711), which filters by the grade in that course, and the status facet (#1730), which allows users to filter by whether the student has paid for the course or is auditing.

pdpinch:
Integrated Rob's comment into the description. 

I'll delete these comments next.

pdpinch:
Add a course to the index for each learner. Include the most recent grade (current or final) the learner earned.

gsidebo:
@pdpinch @roberthouse54 are there any invision links for this?

roberthouse54:
There is this: https://invis.io/E49IZSEWM#/202316455_Learner_Search

gsidebo:
@roberthouse54 @pdpinch a couple questions: 

1. Do we only want to show *currently* enrolled students? Do we want to count students who were enrolled in the course previously and passed?
1. Is it a requirement to show 'empty' courses (ie: courses that have no students enrolled)? It may reduce complexity if we don't have that requirement

roberthouse54:
1. No, we want to show both currently enrolled students, and students who were enrolled in previous runs of the course... including students who enrolled and failed or passed.
[...]

PR #1916, opened 2016-11-30
Refactor user edx data fetching

gsidebo:

No issue - done as a precursor to #933 
**Rebased on #1890 (to help with testing)**

#### What's this PR do?

Streamlines user edX data fetching and avoids potential bugs 

dashboard/api_edx_cache.py and dashboard/models.py

Most of these changes are naming changes and repetitive changes to parameters, etc. The most significant functional change is the addition of [CachedEdxUserData](https://github.com/mitodl/micromasters/compare/master...refactor_user_edx_data_fetching#diff-54f823943ee528710d6f9d178443a994L18) and 

To test that the dashboard API still works correctly:

1. Unseed and seed the database, specifying `--staff-user=YOUR_USERNAME` when you run `seed_db`
1. Check the /dashboard page and make sure 'Digital Learning' and 'Analog Learning' look right
1. Enroll in a non-FA course using the `alter_data` commands. Example: `manage.py alter_data --action=set_course_to_enrolled --username=YOUR_USERNAME --program-title='Analog' --course-title='100'`
1. Reload the /dashboard page and check that the enrolled course shows up correctly

To test that serialization for search results still works:

1. Look at the /learners page and pick one of the fake 'Analog Learning' program users
1. Set one or more courses to passed for that user. Example: `alter_data --action=set_course_to_passed --email=roberto.cortes --program-title='Analog' --course-title='100' --grade=100`
1. Run recreate_index: `manage.py recreate_index`
1. Reload the page and confirm that the user's average grade has been changed

### DIFF
--- a/dashboard/api_edx_cache.py
+++ b/dashboard/api_edx_cache.py
@@ -16,6 +16,45 @@ log = logging.getLogger(__name__)
 
 
 class CachedEdxUserData:
+    """Represents all edx data related to a User"""
+    # pylint: disable=too-many-instance-attributes
+
+    enrollments = None
+    certificates = None
+    current_grades = None
+    raw_enrollments = None
+    raw_certificates = None
+    raw_current_grades = None
+
+    def __init__(self, user, program=None, include_raw_data=False):
+        """
+        Fetches the given User's edx data and sets object properties
+
+        Args:
+            user (User): a User object
+            program (Program): an optional Program to filter on
+            include_raw_data (bool): Set to True if raw edx data is needed (in addition to actual edX objects)
+        """
+        self.user = user
+        self.program = program
+        self.fetch_and_set_edx_data()
+        if include_raw_data:
+            self.fetch_and_set_raw_data()
+
+    def fetch_and_set_edx_data(self):
+        """Fetches edx data and sets object properties"""
+        self.enrollments = models.CachedEnrollment.get_edx_data(self.user, program=self.program)
+        self.certificates = models.CachedCertificate.get_edx_data(self.user, program=self.program)
+        self.current_grades = models.CachedCurrentGrade.get_edx_data(self.user, program=self.program)
+
+    def fetch_and_set_raw_data(self):
+        """Fetches raw cached data and sets object properties"""
+        self.raw_enrollments = list(models.CachedEnrollment.data_qset(self.user, program=self.program))
+        self.raw_certificates = list(models.CachedCertificate.data_qset(self.user, program=self.program))
+        self.raw_current_grades = list(models.CachedCurrentGrade.data_qset(self.user, program=self.program))
+
+
+class CachedEdxDataApi:
     """
     Class to handle the retrieval and update of the users' cached edX information
     """
--- a/dashboard/api_edx_cache_test.py
+++ b/dashboard/api_edx_cache_test.py
@@ -10,9 +10,9 @@ from edx_api.certificates.models import Certificate, Certificates
 from edx_api.enrollments.models import Enrollment, Enrollments
 from edx_api.grades.models import CurrentGrade, CurrentGrades
 
-from courses.factories import CourseRunFactory
+from courses.factories import ProgramFactory, CourseFactory, CourseRunFactory
 from dashboard import models
-from dashboard.api_edx_cache import CachedEdxUserData
+from dashboard.api_edx_cache import CachedEdxUserData, CachedEdxDataApi
 from dashboard.factories import (
     CachedEnrollmentFactory,
     CachedCertificateFactory,
@@ -34,6 +34,66 @@ from search.base import ESTestCase
 class CachedEdxUserDataTests(ESTestCase):
     """
     Tests for the CachedEdxUserData class
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = UserFactory.create()
+        # Create Programs, Courses, CourseRuns...
+        cls.p1_course_run_keys = ['p1_course_run']
+        cls.p2_course_run_keys = ['p2_course_run_1', 'p2_course_run_2']
+        cls.p1_course_run = CourseRunFactory.create(edx_course_key=cls.p1_course_run_keys[0])
+        p2 = ProgramFactory.create(full=True)
+        first_course = p2.course_set.first()
+        extra_course = CourseFactory.create(program=p2)
+        cls.p2_course_run_1 = CourseRunFactory.create(course=first_course, edx_course_key=cls.p2_course_run_keys[0])
+        cls.p2_course_run_2 = CourseRunFactory.create(course=extra_course, edx_course_key=cls.p2_course_run_keys[1])
+        all_course_runs = [cls.p1_course_run, cls.p2_course_run_1, cls.p2_course_run_2]
+        # Create cached edX data
+        cls.enrollments = [
+            CachedEnrollmentFactory.create(user=cls.user, course_run=course_run) for course_run in all_course_runs
+        ]
+        cls.certificates = [
+            CachedCertificateFactory.create(user=cls.user, course_run=course_run) for course_run in all_course_runs
+        ]
+        cls.current_grades = [
+            CachedCurrentGradeFactory.create(user=cls.user, course_run=course_run) for course_run in all_course_runs
+        ]
+
+    def assert_edx_data_has_given_ids(self, edx_user_data, ids):
+        """Asserts that all edX object course id sets match the given list of ids"""
+        assert sorted(edx_user_data.enrollments.get_enrolled_course_ids()) == ids
+        assert sorted(edx_user_data.certificates.all_courses_verified_certs) == ids
+        assert sorted(edx_user_data.current_grades.all_course_ids) == ids
+
+    def test_edx_data_fetch_and_set(self):
+        """Test that a user's edX data is properly fetched and set onto object properties"""
+        edx_user_data = CachedEdxUserData(self.user)
+        assert isinstance(edx_user_data.enrollments, Enrollments)
+        assert isinstance(edx_user_data.certificates, Certificates)
+        assert isinstance(edx_user_data.current_grades, CurrentGrades)
+        self.assert_edx_data_has_given_ids(edx_user_data, self.p1_course_run_keys + self.p2_course_run_keys)
+
+    def test_edx_data_with_program(self):
+        """Test that a user's edX data is filtered by program when specified"""
+        p1_course_run_program = self.p1_course_run.course.program
+        edx_user_data = CachedEdxUserData(self.user, program=p1_course_run_program)
+        self.assert_edx_data_has_given_ids(edx_user_data, self.p1_course_run_keys)
+        p2_course_run_program = self.p2_course_run_1.course.program
+        edx_user_data = CachedEdxUserData(self.user, program=p2_course_run_program)
+        self.assert_edx_data_has_given_ids(edx_user_data, self.p2_course_run_keys)
+
+    def test_raw_data_fetch_and_set(self):
+        """Test that a user's raw edX data is properly fetched and set onto object properties"""
+        edx_user_data = CachedEdxUserData(self.user, include_raw_data=True)
+        assert edx_user_data.raw_enrollments == [obj.data for obj in self.enrollments]
+        assert edx_user_data.raw_certificates == [obj.data for obj in self.certificates]
+        assert edx_user_data.raw_current_grades == [obj.data for obj in self.current_grades]
+
+
+class CachedEdxDataApiTests(ESTestCase):
+    """
+    Tests for the CachedEdxDataApi class
     """
 
     @classmethod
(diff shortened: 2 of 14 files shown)


amir-qayyum-khan:
please rebase it

gsidebo:
apparently there are a bunch of tests failing now after the rebase. looking into it

giocalitri:
I have a preliminary question:
did you change the logic of MMTrack to include only the data of a single program at the time?

gsidebo:
MMTrack was basically untouched (https://github.com/mitodl/micromasters/pull/1916/files#diff-bf545a2544a83f1acac6cd6d14942b19). when we fetch data from the dashboard API, it includes all programs. when we serialize a user for search results, it is limited to a single program. this was true before this PR. we should probably change the way we serialize to include all of a users enrollments/certs/etc but that would be a decently large effort

gsidebo:
@giocalitri the branch that i rebased on is now merged into master and i rebased. the diff should be more friendly now!

pdpinch:
The size and code that this touches is making me nervous. I definitely don't want this included in the 0.25.0 release -- so don't merge before that gets out. 

When we are ready to merge this, what additional testing can/should we do? Does this need another round of performance testing?

dashboard/models.py
```diff
@@ -105,10 +91,15 @@ def delete_all_but(cls, user, course_ids_list):
...
             None
         """
-        cls.active_qset_all(user).exclude(course_run__edx_course_key__in=course_ids_list).delete()
+        cls.active_qset(user).exclude(course_run__edx_course_key__in=course_ids_list).delete()
 
     @classmethod
-    def get_edx_data(cls, user):
+    def get_raw_data(cls, user, program=None):
+        """Returns a list of 'data' values for all matching objects"""
+        return list(cls.data_qset(user, program=program).all())
```
giocalitri: what's the point of making this additional method? I think either `data_qset` or this one are more than enough.

gsidebo:
not putting this in the 0.25.0 release is fine. as far as the diff size, its an overall line change of +243 −226. the bulk of those changes have to do with a naming change and a change in the arguments that a particular class accepts. those are very rote, low-risk changes. not all line changes are created equal, after all :)

dashboard/utils_test.py
```diff
@@ -26,21 +25,21 @@ class MMTrackTest(TestCase):
...
-        cls.enrollments = Enrollments(enrollments_json)
-
-        certificates_json = load_json_from_file('dashboard/fixtures/certificates.json')
-        cls.certificates = Certificates([Certificate(cert_json) for cert_json in certificates_json])
-
-        current_grades_json = load_json_from_file('dashboard/fixtures/current_grades.json')
-        cls.current_grades = CurrentGrades([CurrentGrade(grade_json) for grade_json in current_grades_json])
+        with patch('dashboard.models.CachedEnrollment.data_qset', return_value=cls.enrollments_json), \
+                patch('dashboard.models.CachedCertificate.data_qset', return_value=cls.certificates_json), \
+                patch('dashboard.models.CachedCurrentGrade.data_qset', return_value=cls.current_grades_json):
```
giocalitri: this is inaccurate, given that the `data_qset` returns a query-set and not a list

giocalitri:
>MMTrack was basically untouched (https://github.com/mitodl/micromasters/pull/1916/files#diff-bf545a2544a83f1acac6cd6d14942b19). when we fetch data from the dashboard API, it includes all programs. when we serialize a user for search results, it is limited to a single program. this was true before this PR. we should probably change the way we serialize to include all of a users enrollments/certs/etc but that would be a decently large effort

Yes, after reading the code it was clear also to me.

dashboard/api_edx_cache.py
```diff
@@ -16,6 +16,45 @@
...
+        self.program = program
+        self.fetch_and_set_edx_data()
+        if include_raw_data:
+            self.fetch_and_set_raw_data()
+
+    def fetch_and_set_edx_data(self):
+        """Fetches edx data and sets object properties"""
+        self.enrollments = models.CachedEnrollment.get_edx_data(self.user, program=self.program)
+        self.certificates = models.CachedCertificate.get_edx_data(self.user, program=self.program)
+        self.current_grades = models.CachedCurrentGrade.get_edx_data(self.user, program=self.program)
```
giocalitri: What is the purpose of having the filter by program here?

dashboard/api_edx_cache.py
```diff
@@ -16,6 +16,45 @@
...
+
+    def fetch_and_set_raw_data(self):
+        """Fetches raw cached data and sets object properties"""
+        self.raw_enrollments = models.CachedEnrollment.get_raw_data(self.user, program=self.program)
+        self.raw_certificates = models.CachedCertificate.get_raw_data(self.user, program=self.program)
+        self.raw_current_grades = models.CachedCurrentGrade.get_raw_data(self.user, program=self.program)
```
gsidebo: the query (`data_qset`) is already cached so i figured there's no penalty, but i realize that could potentially change without us noticing it. the objects above are edX objects (`edx_api.enrollments.models.Enrollments`, etc). given those objects, is there a reliable/consistent way to get the raw data as it is stored in our own models?

dashboard/api_edx_cache.py
gsidebo: `program` defaults to none, and the cached model takes care of the logic for applying the program filter if it's specified. this is a way to reuse the logic for fetching and edx-serializing our cached data. if the program isn't passed into this class, the filter won't be applied

dashboard/models.py
```diff
@@ -36,49 +36,35 @@ class Meta:
         abstract = True
 
     @classmethod
-    def active_qset_all(cls, user):
+    def active_qset(cls, user, program=None):
```
gsidebo: yeah, the 'active' notion no longer exists. i'll ponder this one

dashboard/api_edx_cache_test.py
```diff
@@ -38,6 +38,52 @@ class CachedEdxUserDataTests(ESTestCase):
...
+
+    def test_edx_data_fetch_and_set(self):
+        """Test that a user's edX data is properly fetched and set onto object properties"""
+        edx_user_data = CachedEdxUserData(self.user)
+        assert isinstance(edx_user_data.enrollments, Enrollments)
+        assert isinstance(edx_user_data.certificates, Certificates)
+        assert isinstance(edx_user_data.current_grades, CurrentGrades)
+        assert set(edx_user_data.enrollments.get_enrolled_course_ids()) == {'course_run_1', 'course_run_2'}
+        assert set(edx_user_data.certificates.all_courses_verified_certs) == {'course_run_1', 'course_run_2'}
+        assert set(edx_user_data.current_grades.all_course_ids) == {'course_run_1', 'course_run_2'}
```
gsidebo: fair point. i'm happy to change this to compare sorted lists

dashboard/models.py
gsidebo: i'm in the middle on this one. it's obviously not key functionality, but we use do this 3 times in `CachedEdxUserData` and it's nicer to read `CachedEnrollment.get_edx_data(self.user, program=self.program)` than `list(CachedEnrollment.data_qset(self.user, program=self.program).all())`. i think i'm leaning towards getting rid of it

dashboard/api_edx_cache_test.py
```diff
@@ -38,6 +38,52 @@ class CachedEdxUserDataTests(ESTestCase):
...
+        cls.course_run_2 = CourseRunFactory.create(edx_course_key='course_run_2')
+        enrollment1 = CachedEnrollmentFactory.create(user=cls.user, course_run=cls.course_run_1)
+        enrollment2 = CachedEnrollmentFactory.create(user=cls.user, course_run=cls.course_run_2)
+        certificate1 = CachedCertificateFactory.create(user=cls.user, course_run=cls.course_run_1)
+        certificate2 = CachedCertificateFactory.create(user=cls.user, course_run=cls.course_run_2)
+        curgrade1 = CachedCurrentGradeFactory.create(user=cls.user, course_run=cls.course_run_1)
+        curgrade2 = CachedCurrentGradeFactory.create(user=cls.user, course_run=cls.course_run_2)
+        cls.enrollments = [enrollment1, enrollment2]
+        cls.certificates = [certificate1, certificate2]
+        cls.current_grades = [curgrade1, curgrade2]
```
gsidebo: 👍

dashboard/utils_test.py
gsidebo: this is a somewhat tricky one. to start, here's a little snippet that actually mocks a queryset and gets us what we want for enrollments:

```python
from unittest.mock import MagicMock, patch
mocked_enrollment_qset = MagicMock(spec=django.db.models.query.QuerySet, **{'__iter__.return_value': enrollments_json})
with patch('dashboard.models.CachedEnrollment.data_qset', return_value=mocked_enrollment_qset):
    qset = CachedEnrollment.data_qset(user=fake_user)
    print([x for x in qset])
```

it works but obviously a bit wonky. the purpose of this patch was to keep using the same fixtures that we have and avoid repeating the logic for creating edX objects for our raw data. i could also create actual cached records from the raw data, which would require that we pull the `edx_course_key` value for each object type, which would mean more repeated logic. i'm in between on this. how do you feel about the patch example?

dashboard/models.py
giocalitri: do we ever use `cls.data_qset` as a QSET and not only to loop through the entries like in a list?

dashboard/api_edx_cache.py
giocalitri: do we ever call `CachedCurrentGrade.get_edx_data` with a program specified?

dashboard/utils_test.py
giocalitri: Actually I was thinking that if you we get rid of `data_qset`, then you can actually use the same source.
Another approach would be to actually create cached entries for the elements and not mock them.

dashboard/api_edx_cache.py
gsidebo: yes. when this class is instantiated from user search serialization: https://github.com/mitodl/micromasters/pull/1916/files#diff-5e24ceac8b57391a6805615b314c3da8R34

`program` is passed in, and that gets passed along to `CachedCurrentGrade.get_edx_data` in this method

giocalitri: ok

dashboard/api_edx_cache.py
giocalitri: each `edx_api_client` object has all the JSON in a dictionary where the keys are the `course_key`.
for each object there is a method or property to access the values of the dictionary (like `Enrollments.enrolled_courses`.
I think that then it might get a bit too complicated. It is up to you if it is worth it.

dashboard/api_edx_cache_test.py
```diff
@@ -38,6 +38,66 @@ class CachedEdxUserDataTests(ESTestCase):
...
+        assert sorted(edx_user_data.certificates.all_courses_verified_certs) == ids
+        assert sorted(edx_user_data.current_grades.all_course_ids) == ids
+
+    def test_edx_data_fetch_and_set(self):
+        """Test that a user's edX data is properly fetched and set onto object properties"""
+        edx_user_data = CachedEdxUserData(self.user)
+        assert isinstance(edx_user_data.enrollments, Enrollments)
+        assert isinstance(edx_user_data.certificates, Certificates)
+        assert isinstance(edx_user_data.current_grades, CurrentGrades)
+        self.assert_edx_data_has_given_ids(edx_user_data, ['p1_course_run', 'p2_course_run_1', 'p2_course_run_2'])
```
giocalitri: please move the ids to `cls` variables and reference them both here and during the creation of the runs

dashboard/api_edx_cache_test.py
```diff
@@ -38,6 +38,66 @@ class CachedEdxUserDataTests(ESTestCase):
...
+
+    def test_edx_data_with_program(self):
+        """Test that a user's edX data is filtered by program when specified"""
+        p1_course_run_program = self.p1_course_run.course.program
+        edx_user_data = CachedEdxUserData(self.user, program=p1_course_run_program)
+        self.assert_edx_data_has_given_ids(edx_user_data, ['p1_course_run'])
+        p2_course_run_program = self.p2_course_run_1.course.program
+        edx_user_data = CachedEdxUserData(self.user, program=p2_course_run_program)
+        self.assert_edx_data_has_given_ids(edx_user_data, ['p2_course_run_1', 'p2_course_run_2'])
```
giocalitri: same comment as before about the ids

gsidebo: fixed

dashboard/api_edx_cache_test.py
```diff
@@ -38,6 +38,70 @@ class CachedEdxUserDataTests(ESTestCase):
...
+        # Create cached edX data
+        enrollment1 = CachedEnrollmentFactory.create(user=cls.user, course_run=cls.p1_course_run)
+        enrollment2 = CachedEnrollmentFactory.create(user=cls.user, course_run=cls.p2_course_run_1)
+        enrollment3 = CachedEnrollmentFactory.create(user=cls.user, course_run=cls.p2_course_run_2)
+        certificate1 = CachedCertificateFactory.create(user=cls.user, course_run=cls.p1_course_run)
+        certificate2 = CachedCertificateFactory.create(user=cls.user, course_run=cls.p2_course_run_1)
+        certificate3 = CachedCertificateFactory.create(user=cls.user, course_run=cls.p2_course_run_2)
+        curgrade1 = CachedCurrentGradeFactory.create(user=cls.user, course_run=cls.p1_course_run)
+        curgrade2 = CachedCurrentGradeFactory.create(user=cls.user, course_run=cls.p2_course_run_1)
+        curgrade3 = CachedCurrentGradeFactory.create(user=cls.user, course_run=cls.p2_course_run_2)
```
giocalitri: can you make these assignments in a loop?

gsidebo: done

giocalitri:
👍

PR #1918, opened 2016-12-01
Add basic styleguide for wording

singingwolfboy:
I've noticed that different parts of the website use different phrasing to refer to a "master's degree". We should be consistent in our phrasing, to avoid confusing students who are trying to figure out what a master's degree is.

I don't know if what I've put into this styleguide is correct, but if it isn't, we should decide on what *is* correct, and standardize on that instead.

### DIFF
--- a/docs/styleguide-wording.md
+++ b/docs/styleguide-wording.md
@@ -1,0 +1,35 @@
+# Styleguide: Wording & Phrasing
+
+We should be consistent in how we refer to key ideas and phrase key terms
+in this project. Inconsistency is confusing, expecially for students who are
+trying to learn! This styleguide will define how to refer to these key ideas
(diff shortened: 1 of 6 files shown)


docs/styleguide-wording.md
```diff
@@ -0,0 +1,35 @@
+# Styleguide: Wording & Phrasing
```
noisecapella: Is there any preference for `’` over `'`? If so we should add it here

roberthouse54:
👍

docs/styleguide-wording.md
singingwolfboy: I don't know. This styleguide can and should be a living document -- if we want to decide on a preference, then we can make a future pull request to modify this document!

PR #1924, opened 2016-12-01
Added setting to configure Cloudfront

### LINKED ISSUE #1661: Performance: host static assets on a CDN
Part of #1530 

In doing some testing with yslow it identified some static assets that could be hosted on a CDN. 

some of these probably already are on a CDN, but yslow doesn't know it:

- [x] fonts.googleapis.com: 2 components
- [x] js-agent.newrelic.com: 1 component
- [ ] bam.nr-data.net: 1 component

some of these would need to be changed in our build/deploy pipeline:

- [ ] micromasters.herokuapp.com: 16 components

some of these come from the CMS, which may only require a configuration change in our S3 bucket?

- [ ] odl-micromasters-production.s3.amazonaws.com: 2 components

### LINKED ISSUE #1681: performance: enable gzip compression
not sure if the best way to do this is with django middleware, or in uwsgi, or someplace else

noisecapella:
Fixes #1661 
Fixes #1681 

#### What's this PR do?
Adds configuration allowing the use of Cloudfront for static assets (this does not currently include S3 hosted assets)

To make sure the URLs are configured properly, set the `CLOUDFRONT_DIST` environment variable to something random in the PR build. Refresh the page and note that the URLs to JS files are coming from cloudfront. (These URLs will not work for the PR build)

On micromasters-rc set the `CLOUDFRONT_DIST` environment variable. Everything should work as it did before, except that the static assets are fetched from cloudfront.net.

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -14,6 +14,7 @@ import ast
 import logging
 import os
 import platform
+from urllib.parse import urljoin
 
 import dj_database_url
 import yaml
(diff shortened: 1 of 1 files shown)


noisecapella:
There isn't a unit test in this PR because I couldn't figure out how to test conditional settings like this. I think that's what the code coverage check is complaining about

rhysyngsun:
:+1:

PR #1926, opened 2016-12-01
ensure a validator is available on this.props in the about me dialog

aliceriot:


amir-qayyum-khan:
Do you need to update https://github.com/mitodl/micromasters/blob/master/static/js/components/UserPageAboutMeDialog.js#L12 ?

amir-qayyum-khan:
you need to remove https://github.com/mitodl/micromasters/blob/master/static/js/components/UserPageAboutMeDialog.js#L6 and update https://github.com/mitodl/micromasters/blob/master/static/js/components/UserPageAboutMeDialog.js#L32

aliceriot:
can you comment ont he pull request directly? it is much easier to read that way.

aliceriot:
do you mean change that to pull the validator out of `this.props` instead?

amir-qayyum-khan:
i cant see these lines in diff

aliceriot:
you can expand the diff to show other lines.

comments should be addressed.

amir-qayyum-khan:
👍

PR #1932, opened 2016-12-01
Added tests for bundles

noisecapella:
None

#### What's this PR do?
Adds tests to assert that the proper JS bundle links are being rendered on the right pages

Tests should pass

### DIFF
--- a/financialaid/views_test.py
+++ b/financialaid/views_test.py
@@ -243,17 +243,28 @@ class ReviewTests(FinancialAidBaseTestCase, APIClient):
             GA_TRACKING_ID=ga_tracking_id,
             REACT_GA_DEBUG=react_ga_debug,
             EDXORG_BASE_URL=base_url,
-        ):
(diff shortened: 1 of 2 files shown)


financialaid/views_test.py
```diff
@@ -243,17 +243,28 @@ def test_context(self):
...
-            assert response.context['is_public'] is True
-            self.assertContains(response, 'Share this page')
-            assert json.loads(response.context['js_settings_json']) == {
-                'gaTrackingID': ga_tracking_id,
-                'reactGaDebug': react_ga_debug,
-                'authenticated': True,
-                'edx_base_url': base_url,
-            }
+
+        bundles = [bundle[0][1] for bundle in get_bundle.call_args_list]
```
rhysyngsun: I noticed this code lost a level of indentation, was that intentional?

noisecapella: Not really, I unindented some code but I reverted it for the other blocks because I wanted to minimize the change. I'll revert it here too

rhysyngsun:
:+1:

PR #1939, opened 2016-12-02
Remove alt text from course images

### LINKED ISSUE #1891: Home: remove alt text from program thumbnails
The home page already contains program text. I don't think describing the thumbnails is terribly helpful. 

from https://github.mit.edu/ODLaccessibility/MicroMasters-Portal/issues/3

> Right now, the alt text as, for example:

`alt="course image for Supply Chain Management">`

> is not useful.

> Depending on whether or not you consider it essential to see the image content, the alt text provided for the current program images should either be more indicative of what the image is showing - or, if the focus of this page is purely to list and describe the programs, it's fine to make it null.

singingwolfboy:
Fixes #1891.

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -143,11 +143,11 @@
                 </div>
                 {% if programpage and programpage.thumbnail_image %}
                   {% image programpage.thumbnail_image fill-690x530 as thumbnail_image %}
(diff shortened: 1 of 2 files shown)


noisecapella:
I guess the test is failing because we're not showing `program.title` if the link isn't present. You should be able to replace `title` with `description` in the test to fix it. By the way I made some changes here so you should rebase before fixing the test

PR #1944, opened 2016-12-02
Tests: make factories default to populating dependent objects

singingwolfboy:
The [Factory Boy](http://factoryboy.readthedocs.io/en/latest/) library, which we're using for making test data, has some nice utilities for automatically populating dependent objects in a data model. In particular, we can use [`SubFactory`](http://factoryboy.readthedocs.io/en/latest/reference.html#factory.SubFactory) to create a dependent object *before* the target object is finished being created, and we can use [`RelatedFactory`](http://factoryboy.readthedocs.io/en/latest/reference.html#factory.RelatedFactory) to create a dependent object *after* the target object is finished being created. In both cases, it's easy to indicate that the dependent object shouldn't be created at all, simply by passing `<dependent>=None` as keyword arguments when creating the target object.

For example, when creating a Program object, you'll probably also want to create a Course object for it, and a CourseRun object for the Course, and perhaps even a CoursePrice object for the CourseRun. Previously, this was handled by passing a `full=True` parameter to the `ProgramFactory` class, which was detected by overriding the `_create()` method. Now, creating these dependent objects is the default usage. If you _don't_ want to create a Course object, you can call `ProgramFactory.create(course=None)`, and you'll just get a Program object. If you want a Program and a Course but not a CourseRun, you can call `ProgramFactory.create(course__course_run=None)`. You'll still get a Program object, and a Course object will be created for you, but a CourseRun object will not be created.

This pull request is still in progress, because there are a few tests in the codebase that need to be changed to accommodate this new assumption. However, it's mostly working already.

### DIFF
--- a/courses/factories.py
+++ b/courses/factories.py
@@ -7,55 +7,51 @@ import factory
 from factory import fuzzy
 from factory.django import DjangoModelFactory
 
-from .models import Program, Course, CourseRun
+from courses.models import Program, Course, CourseRun
+from ecommerce.models import CoursePrice
 
 FAKE = faker.Factory.create()
 
 
 class ProgramFactory(DjangoModelFactory):
     """Factory for Programs"""
-    title = fuzzy.FuzzyText(prefix="Program ")
-    live = fuzzy.FuzzyAttribute(FAKE.boolean)
-    description = fuzzy.FuzzyText()
+    title = factory.LazyAttribute(lambda x: FAKE.company())
+    description = factory.LazyAttribute(lambda x: FAKE.bs())
+    live = True
+    financial_aid_availability = False
+
+    course = factory.RelatedFactory('courses.factories.CourseFactory', "program")
+
+    @factory.post_generation
+    def tiers(self, create, extracted, **kwargs):  # pylint: disable=unused-argument
+        if not extracted:
+            return
+        if self.financial_aid_availability:
+            from financialaid.factories import TierProgramFactory
+            for discount, threshold in [(0, 1000), (50, 0)]:
+                TierProgramFactory.create(
+                    program=self,
+                    current=True,
+                    discount_amount=discount,
+                    income_threshold=threshold,
+                )
 
     class Meta:  # pylint: disable=missing-docstring
         model = Program
-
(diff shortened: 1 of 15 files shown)


gsidebo:
currently using this in an open PR: https://github.com/mitodl/micromasters/pull/1916

surely there's some way that we can keep something around that accomplishes the goal of this code (reduce the large amount of boilerplate to create a program/course/course run with ancillary records like course price/tier program where needed) instead of removing it entirely

gsidebo:
maybe creating a different factory class for the 'full' Program factory is more easily understood and implemented

gsidebo:
@singingwolfboy what's going on with this pull request? did you see my comments above? also, there's a bunch of work around the `ProgramPage` model which seems to indicate that this PR has shifted purpose, so to speak

singingwolfboy:
@gsidebo Yes, I saw your comment -- sorry for not responding. I started exploring ways to make this work more cleanly, and I learned about Factory Boy's [RelatedFactory](http://factoryboy.readthedocs.io/en/latest/reference.html#factory.RelatedFactory) attribute, which solves this case very neatly. I've updated the description of this pull request to make it more accurate. Does that make the situation more clear?

courses/factories.py
```diff
@@ -7,56 +7,52 @@
...
-        program = model_class(*args, **kwargs)
-        program.save()
-        if full_create:
-            course = CourseFactory.create(program=program)
-            course_run = CourseRunFactory.create(course=course)
-            from ecommerce.factories import CoursePriceFactory
-            course_price = CoursePriceFactory.create(course_run=course_run, is_valid=True)
-            if program.financial_aid_availability:
-                from financialaid.factories import TierProgramFactory
+    @factory.post_generation
```
gsidebo: see, this is what i _really_ wanted when i wrote that `full=True` nonsense :) thanks for finding this! i'll try to get to this PR today

singingwolfboy:
This ended up being more work than I anticipated, for comparatively little benefit. I'm going to close this pull request, but leave the branch around at least for now. If we want to pick this up again in the future, we can do so.

PR #1964, opened 2016-12-05
making the gender radio buttons more accessible

### LINKED ISSUE #1849: Sign up: gender radio buttons not accessible
> Use fieldset or role="group" and aria-labelledby to label the group
> use label element to label each individual field (male, female, other)

aliceriot:

closes #1849 

#### What's this PR do?

Hopefully this makes the gender radio buttons more accessible, by giving them proper labels and so on.

@singingwolfboy I would love to get your feedback on this, since I'm not sure this is the proper way to address this. MaterialUI is a bit frustrating to deal with here.

### DIFF
--- a/static/js/util/profile_edit.js
+++ b/static/js/util/profile_edit.js
@@ -4,9 +4,10 @@ import _ from 'lodash';
 import TextField from 'material-ui/TextField';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 import Checkbox from 'material-ui/Checkbox';
+import R from 'ramda';
 
 import DateField from '../components/inputs/DateField';
-import { validationErrorSelector } from './util';
+import { validationErrorSelector, classify } from './util';
 import type { Validator, UIValidator } from '../lib/validation/profile';
 import type { Profile } from '../flow/profileTypes';
 import type { Option } from '../flow/generalTypes';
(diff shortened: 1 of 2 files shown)


static/js/util/profile_edit.js
```diff
@@ -62,13 +63,17 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
...
+        aria-labelledby={labelId}
       />
     );
   });
 
   const value = String(_.get(profile, keySet));
   return (
-    <div className={validationErrorSelector(errors, keySet)}>
+    <div
+      role="group"
```
rhysyngsun: I think `radiogroup` would be better here. Could you also add a `aria-labelledby` attribute here that points to the `profile-radio-group-label` just below? That will make screen readers will annoucne this section as "Gender Radio Button Group".

static/js/util/profile_edit.js
```diff
@@ -48,8 +48,9 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
     if (obj.helper) {
       helper = `${obj.helper}`;
     }
+    let labelId = `gender-option-${obj.label}`;
```
singingwolfboy: This doesn't work properly if `obj.label` has spaces in it, such as "Other/Prefer not to say".

static/js/util/profile_edit.js
```diff
@@ -62,13 +63,17 @@ export function boundRadioGroupField(keySet: string[], label: string, options: O
         labelStyle={styles.labelStyle}
         value={obj.value}
         label={label}
+        aria-labelledby={labelId}
```
singingwolfboy: It's better to use a `<label>` element rather than `aria-labelledby` if possible. You can nest the `<input>` element inside of the `<label>` element, or use the `for` attribute on the label to point to the element ID that is being labelled. Using `<label>` also makes the browser automatically understand that when the user clicks on the label text, the `<input>` field should be selected -- although it looks like Material UI is manually recreating that behavior using Javascript.

singingwolfboy:
In order for this to be properly accessible, these three `<input>` elements need to be wrapped inside of a `<fieldset>` element, with a `<legend>` element that describes the choice that the user is making. As an example, notice how the day/month/year fields are wrapped in a `<fieldset>` with a `<legend>` element, to inform screen readers that these three fields are all closely related to each other.

rhysyngsun:
@singingwolfboy if you'd like to take over review on this feel free to reassign

static/js/util/profile_edit.js
aliceriot: material-ui is a whole ball of wax here...I'll wrap the `RadioButton` in a label tag, since when you pass the `label` prop in to `RadioButton` it doesn't seem to get properly connected to the input.

aliceriot: Unfortunately I don't think we can do something like:

```js
<label>
  A label
  <RadioButton
    prop={val}
  />
</label>
```

because we render those `RadioButton`s as the children of a `RadioButtonGroup`. `RadioButtonGroup` assumes that all of it's children are `RadioButton` elements at the top level...annoying.

Not sure what to do here. When I set the `label` attribute on `RadioButton` it doesn't render the input inside of the label element OR use `aria-labelledby` to point the screenreader to the appropriate label, at least as far as I can tell. Anyway, tryna figure out what to do here.

aliceriot:
@singingwolfboy I updated to use `<fieldset>` and `<legend>` elements, let me know what you think!

aliceriot:
oops, I think I accidentally forgot to add the right class back onto the `label` element.

aliceriot:
just added it back real quick

singingwolfboy:
Now the labels for the radio buttons are no longer bold, but all the dropdown menus still are bold.

aliceriot:
oh that was a silly CSS error. two seconds.

singingwolfboy:
All better!

PR #1965, opened 2016-12-05
Check element and label in case they're undefined

### LINKED ISSUE #1962: Zendesk error when retrieving elements
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/190233171/
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/190244172/

We should add code to do nothing if the expected DOM nodes don't exist

noisecapella:
Fixes #1962 
#### What's this PR do?
Checks that the element and label exist before settings the display

This error doesn't seem to be easily reproducable. Hopefully the changes are simple enough to not cause problems though


rhysyngsun:
:+1:

PR #1967, opened 2016-12-05
Add test for user without staff or instructor role

noisecapella:
None

#### What's this PR do?
Adds a test to cover uncovered code in search/views.py: https://codecov.io/gh/mitodl/micromasters/src/a13e15315f0fb929157d6a82ff4220c6f5c32d61/search/views.py#L37

This only touches test code so no manual testing should be necessary

### DIFF
--- a/search/views_test.py
+++ b/search/views_test.py
@@ -1,6 +1,7 @@
 """
 Tests for the search view
 """
+import ddt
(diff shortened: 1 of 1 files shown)


rhysyngsun:
:+1:

PR #1972, opened 2016-12-05
don't use png, it is too big

### LINKED ISSUE #1971: UnreadablePostError
We think it may be due to overly large profile images. Some of them seem to be over 18MB. 

See https://sentry.io/share/issue/3130343439372e313930333037313735/ 

Also see NewRelic.

If we aren't checking the size of the image, we should. It should be kept to a reasonable limit.

aliceriot:

part of #1971 

#### What's this PR do?

This changes the encoding when we call `.toBlob` in order to save a `jpg` instead of a `png`. In testing this results in a pretty big decrease in image size. 

Saving a `png` through our image thing actually seems to usually result in bigger images by default. Changing the encoding to `image/jpeg` seems to do a pretty good job of cutting this down.

### DIFF
--- a/static/js/components/CropperWrapper.js
+++ b/static/js/components/CropperWrapper.js
@@ -13,7 +13,7 @@ export default class CropperWrapper extends React.Component {
     const { updatePhotoEdit } = this.props;
     let canvas = this.refs.cropper.getCroppedCanvas();
     if (canvas.toBlob !== undefined) {
-      canvas.toBlob(blob => updatePhotoEdit(blob));
+      canvas.toBlob(blob => updatePhotoEdit(blob), 'image/jpeg');
     }
   };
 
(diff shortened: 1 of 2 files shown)


noisecapella:
Code looks fine, just need to test this on different browsers

aliceriot:
yes, that is a good idea :)

giocalitri:
we should not close #1971 until the second part of what we were talking about is done: we still might have problems even with this solution

aliceriot:
fair enough

noisecapella:
Android, Chrome, Firefox and IE Edge all work fine

giocalitri:
Chrome and Safari on iOS work

giocalitri:
Works on safari on Mac Os

PR #1980, opened 2016-12-05
Added post-buffering option to uWSGI to handle large POSTs

blarghmatey:
**Added post-buffering option to uWSGI to handle large POSTs**
Large POST messages are causing errors in the app. Enabling buffering
of those requests should alleviate those errors.


pdpinch:
Same question I asked on Slack -- how do we test this? I can try different settings, but at which end do I expect it to fail? And do I need to slow down my POST somehow?

aliceriot:
Chrome has the ability to simulate a slow internet connection - that feature may be handy here.

blarghmatey:
We expect this to fail at the receiving end when submitting your user details during the signup flow. I think that slowing down your connection is the best way to try and trigger a failure.

blarghmatey:
Closing this PR because the buffering isn't the solution for the targeted problem and the socket timeout actually makes things worse.

PR #1981, opened 2016-12-05
Fix about me width issue

### LINKED ISSUE #1937: "about me" margins need adjustment
In Mac Chrome and Safari, the margins are too small:

screenshot-micromasters-rc herokuapp com-2016-12-01-21-29-52

In Firefox, the margins are too big:

screenshot-micromasters-rc herokuapp com-2016-12-01-21-32-20

roberthouse54:
Closes #1937

#### What's this PR do?

This fixes a layout bug with the About Me copy on the user page. There are still some weird layout bugs with the user avatar and the user name in the same box, but I think that should be fixed in a separate PR. The problem was an incorrect implementation of flexbox, and widths.

Test this on a few browsers.

Before and after screenshots:
image

image

### DIFF
--- a/static/scss/profile.scss
+++ b/static/scss/profile.scss
@@ -91,6 +91,7 @@
     display: flex;
     flex-direction: row;
     align-items: center;
+    justify-content: space-between;
 
     &.row-with-border {
       border-bottom: 1px #e2e2e2 solid;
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
it appears like below.  there is good space between about me text and edit button.

roberthouse54:
@amir-qayyum-khan That is intended. I set the width to 80%. Your screenshot matches the one I posted above.

amir-qayyum-khan:
👍

PR #1991, opened 2016-12-06
Nl/1978 truncate profile image filenames

### LINKED ISSUE #1978: Profile: Image upload failed if the filename is too long
Image URI exceeds the column size and thus fails:

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/190348344/

rhysyngsun:

Fixes #1978 

#### What's this PR do?

This PR truncates the original filename that we use so the entire S3 object URI will fit in the 100 chars we have for that field. It does so by changing the `upload_to` property of the `Profile.image` field to a `callable` that generates a shortened URI.

profiles/util.py and profiles/models.py

Upload a profile image that has a filename > 100 characters (I'd probably give it 150 or so).

Apparently changes to `ImageField.upload_to` constitute a schema change as far as django is concerned, hence the migration in this PR. If anyone has a better approach or know how to not trigger this behavior I'd appreciate it.


rhysyngsun:
For the sake of completeness, here's the proof the migration is a no-op:

```
~/odl/micromasters nl/1978-truncate-profile-image-filenames
> docker-compose run web ./manage.py sqlmigrate profiles 0022_truncate_profile_image_uri
BEGIN;
--
-- Alter field image on profile
--
COMMIT;
```

profiles/util.py
```diff
@@ -59,3 +66,17 @@ def format_gravatar_url(user_email, size):
...
+def profile_image_upload_uri(_, filename):
+    """
+    Helper to format the uri for the profile image upload
+    """
+    name, ext = path.splitext(filename)
+    timestamp = datetime.now(pytz.utc).replace(microsecond=0)
+
+    return "profile/{name}-{timestamp}{ext}".format(
+        name=name[:IMAGE_FILENAME_MAX_LENGTH],
+        timestamp=timestamp.strftime("%Y-%m-%dT%H%M%S%z"),
```
gsidebo: are you just going for iso format here? pretty sure you can use `timestamp.isoformat()`. if that's not what you're going for, you can leave as is

profiles/util.py
```diff
@@ -1,9 +1,16 @@
 """
 util functions for profiles
 """
-import hashlib
+from datetime import datetime
+from os import path
 from urllib import parse
 
+import hashlib
+
```
gsidebo: can you get rid of this blank line and the one at line 7?

profiles/util.py
rhysyngsun: @gsidebo it's close to, but not quite, what `isoformat()` returns. In particular it drops the colons between the time parts. My goal here was to keep the same suffix convention as before.

profiles/util.py
rhysyngsun: Done

rhysyngsun:
Verifying test suite before I hand it back.

rhysyngsun:
@gsidebo based on what Alice mentioned at standup, I removed the timestamp from the frontend so that the backend becomes authoritative for what the final filename will be.

gsidebo:
just tried it again. looks good! 👍

PR #2003, opened 2016-12-06
Use HTML elements that are more semantic

### LINKED ISSUE #1917: Use <section> element for grouping parts of page in Home and Program templates
The HTML5 [`<section>`](http://mdn.beonex.com/en/HTML/Element/section.html) element is designed to semantically represent a generic section of a page. When combined with header elements (`<h1>` through `<h6>`), the `<section>` element makes it easier for a screen reader user to understand the general flow of a page and jump to the particular part of the page they are interested in.

For accessibility purposes, we should try to use semantically meaningful elements where possible, including replacing `<div>` with `<section>` in the templates where it makes sense to do so.

singingwolfboy:
Fixes #1917. Improves our usage of semantic HTML elements, which improves accessibility. There should be no user-visible changes in this pull request.

It may be helpful to add `?w=1` to the GitHub URL when reviewing this pull request, so that GitHub does not display changes that are only whitespace.

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -45,8 +45,7 @@
 {% include "header.html" %}
 
   <div class="main-content-wrapper">
-    <section class="main-content">
-
+    <main>
       <div class="banner-wrapper">
         <div class="banner-wrapper-content">
           <h1>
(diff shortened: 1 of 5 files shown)


cms/templates/cms/home_page.html
```diff
@@ -231,13 +224,12 @@ <h4 class="program-title" aria-describedby="program-{{ program.id }}-description
...
       </section>
 
-    </section>
+    </main>
   </div>
 
   {% include "footer.html" %}
 
-  <div class="modal fade modal-fade-in-scale-up" id="modal-about" aria-hidden="true" aria-labelledby="modal-about"
-                                                                                     role="dialog" tabindex="-1">
```
noisecapella: This is fine, just wondering what this changes, if anything

noisecapella:
One question, feel free to merge :+1:

cms/templates/cms/home_page.html
singingwolfboy: Ah, I removed those properties unintentionally. Thanks for catching that!

PR #2004, opened 2016-12-06
Limit user photo uploads to 1024x1024 pixels

aliceriot:
#### What's this PR do?

This implements a function that uses `<canvas>` to limit the maximum dimensions of user photos to 1024x1024 (right now, we can use whatever number we want).

Probably just ensure that even if you throttle your connection it is possible to upload an image? hopefully...

### DIFF
--- a/static/js/components/CropperWrapper.js
+++ b/static/js/components/CropperWrapper.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
 import Cropper from 'react-cropper';
+import browser from 'detect-browser';
 
 export default class CropperWrapper extends React.Component {
   props: {
@@ -11,10 +12,16 @@ export default class CropperWrapper extends React.Component {
 
   cropperHelper = () => {
     const { updatePhotoEdit } = this.props;
-    let canvas = this.refs.cropper.getCroppedCanvas();
-    if (canvas.toBlob !== undefined) {
-      canvas.toBlob(blob => updatePhotoEdit(blob), 'image/jpeg');
+    let canvas;
+    if ( browser.name === 'safari' || browser.name === 'ios' ) {
+      canvas = this.refs.cropper.getCroppedCanvas();
+    } else {
+      canvas = this.refs.cropper.getCroppedCanvas({
+        width: 512,
+        height: 512,
+      });
     }
+    canvas.toBlob(blob => updatePhotoEdit(blob), 'image/jpeg');
   };
 
   render () {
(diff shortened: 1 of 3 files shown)


aliceriot:
With this change I can upload a photo on both the simulated 2G and GPRC throttling in chromium on the PR environment.

noisecapella:
Code looks fine, looking into testing on different browsers

noisecapella:
:+1: from me. Phone is more responsive

aliceriot:
not working yet on Safari, seems like it's an encoding issue.

static/js/util/util.js
```diff
@@ -458,3 +458,41 @@ export function getUserDisplayName(profile: Profile): string {
...
+    }
+
+    const canvas = document.createElement('canvas');
+    canvas.width = width;
+    canvas.height = height;
+
+    const ctx: any = canvas.getContext("2d");
+    ctx.drawImage(img, 0, 0, width, height);
+
+    return canvas.toBlob(saveCallback, 'image/jpeg');
```
aliceriot: We want to use JPEG because it is a substantial reduction in size over PNG. I merged a change yesterday which switched that over.

giocalitri:
changed labels (I guess it is still in progress)

aliceriot:
@noisecapella this is ready for further review, and things are working on iOS. The PNG option you suggested didn't pan out, so what I ended up doing is just not attempting to do any resizing on safari or iOS.

pdpinch:
Ugh. Doesn't that defeat the purpose -- especially on iOS?

aliceriot:
Yes, but I could not find a way to make it work.

static/js/components/CropperWrapper.js
```diff
@@ -11,10 +12,16 @@ export default class CropperWrapper extends React.Component {
...
+    let canvas;
+    if ( browser.name === 'safari' || browser.name === 'ios' ) {
+      canvas = this.refs.cropper.getCroppedCanvas();
+    } else {
+      canvas = this.refs.cropper.getCroppedCanvas({
+        width: 512,
+        height: 512,
+      });
     }
+    canvas.toBlob(blob => updatePhotoEdit(blob), 'image/jpeg');
```
noisecapella: We used to check for `toBlob` here. Will it always exist?

aliceriot: yes, we have a polyfill for it.

noisecapella:
If the image is small this will increase the size to 512x512. That might be fine but just wanted to bring it up

IE Chrome, Firefox, and Chrome on Android all work. Can someone do a final test for OS X and iOS?

aliceriot:
It works on my iPhone, not sure about Safari. @giocalitri would you mind testing to see if you can upload a profile image on the PR build?

giocalitri:
it works on Safari on Mac OS

PR #2007, opened 2016-12-06
don't require a profile photo

aliceriot:
1. use the django shell to delete your user's profile image
2. go to `/profile/personal`
3. confirm you can navigate forward without being prompted to add a profile image.


noisecapella:
:+1:

PR #2010, opened 2016-12-06
Increase socket-timeout

### LINKED ISSUE #1971: UnreadablePostError

noisecapella:
fixes #1971 

#### What's this PR do?
Increases `socket-timeout` uwsgi setting to help prevent timeout

In Chrome turn on GPRS throttling and upload a 2mb profile image. It should take a while but it should complete without error.


uwsgi.ini
```diff
@@ -33,3 +33,9 @@ endif =
 if-not-env = DEV_ENV
 memory-report = true
 endif =
+if-env = UWSGI_SOCKET_TIMEOUT
+socket-timeout = %(_)
+endif =
+if-not-env = UWSGI_SOCKET_TIMEOUT
+socket-timeout=30
```
rhysyngsun: My only nit here is the inconsistent spacing between this and the `if-env`. Otherwise :+1:

PR #2014, opened 2016-12-07
Removed First and Last Name from the edxorg pipeline

### LINKED ISSUE #2011: Can't create new accounts with edX name longer than 30 characters

giocalitri:
fixes #2011

#### What's this PR do?
First and Last name in the EDXORG pipeline are used only to create the Django User object,
but given that we store these informations in the profile, we can remove them from there
and not risk to break the app because they are too long.

Create a name in edX where the string after the first space is more than 30 characters.
The MM app should not break.

You should also test that an user that previously failed to login, should be able to login after this PR is in place.

You should be able to see that the newly created user has no first and last name in the Django Admin/User.


pdpinch:
👍

PR #2017, opened 2016-12-07
Created course facet to filter learners by course enrollment

### LINKED ISSUE #933: Search: Courses facet

gsidebo:

Closes #933 

#### What's this PR do?

Adds the course facet to the learner search page

Probably dashboard/api_edx_cache.py and dashboard/serializers.py

With seed data loaded and your user given staff permissions, run the `recreate_index` command and check the learners search page. The course facet should be at the top.

Mockup: https://projects.invisionapp.com/share/E49IZSEWM#/screens/202316455_Learner_Search

There's a design quirk with this implementation: the facet shows an "All" option, and that option potentially shows a higher count than there are students in the search results. Basically it's a count of all of the unique course enrollments, so if User A is enrolled in 3 courses, it adds 3 to the "All" count. The "All" option is not part of the mockup, but it is added by searchkit and I haven't yet found a clear way to get rid of it and make it click-on-click-off like the "Current Residence" facet. We could just hide that row via CSS, but that's more than a little hacky. @aliceriot maybe you and I can talk about this.

ss 2016-12-06 at 22 04 46

### DIFF
--- a/dashboard/api_edx_cache.py
+++ b/dashboard/api_edx_cache.py
@@ -3,9 +3,8 @@ APIs that deal with the edx cached data
 """
 import datetime
 import logging
-
-from django.db import transaction
 import pytz
+from django.db import transaction
 
 from courses.models import CourseRun
 from dashboard import models
@@ -16,42 +15,22 @@ log = logging.getLogger(__name__)
 
 
 class CachedEdxUserData:
-    """Represents all edx data related to a User"""
+    """Represents all edX data related to a User"""
     # pylint: disable=too-many-instance-attributes
 
-    enrollments = None
-    certificates = None
-    current_grades = None
-    raw_enrollments = None
-    raw_certificates = None
-    raw_current_grades = None
-
-    def __init__(self, user, program=None, include_raw_data=False):
+    def __init__(self, user, program=None):
         """
         Fetches the given User's edx data and sets object properties
 
         Args:
             user (User): a User object
             program (Program): an optional Program to filter on
-            include_raw_data (bool): Set to True if raw edx data is needed (in addition to actual edX objects)
         """
         self.user = user
         self.program = program
-        self.fetch_and_set_edx_data()
-        if include_raw_data:
-            self.fetch_and_set_raw_data()
-
-    def fetch_and_set_edx_data(self):
-        """Fetches edx data and sets object properties"""
         self.enrollments = models.CachedEnrollment.get_edx_data(self.user, program=self.program)
         self.certificates = models.CachedCertificate.get_edx_data(self.user, program=self.program)
         self.current_grades = models.CachedCurrentGrade.get_edx_data(self.user, program=self.program)
-
-    def fetch_and_set_raw_data(self):
-        """Fetches raw cached data and sets object properties"""
-        self.raw_enrollments = list(models.CachedEnrollment.data_qset(self.user, program=self.program))
-        self.raw_certificates = list(models.CachedCertificate.data_qset(self.user, program=self.program))
-        self.raw_current_grades = list(models.CachedCurrentGrade.data_qset(self.user, program=self.program))
 
 
 class CachedEdxDataApi:
--- a/dashboard/api_edx_cache_test.py
+++ b/dashboard/api_edx_cache_test.py
@@ -12,7 +12,10 @@ from edx_api.grades.models import CurrentGrade, CurrentGrades
 
 from courses.factories import ProgramFactory, CourseFactory, CourseRunFactory
 from dashboard import models
-from dashboard.api_edx_cache import CachedEdxUserData, CachedEdxDataApi
+from dashboard.api_edx_cache import (
+    CachedEdxUserData,
+    CachedEdxDataApi
+)
 from dashboard.factories import (
     CachedEnrollmentFactory,
     CachedCertificateFactory,
@@ -62,9 +65,9 @@ class CachedEdxUserDataTests(ESTestCase):
 
     def assert_edx_data_has_given_ids(self, edx_user_data, ids):
         """Asserts that all edX object course id sets match the given list of ids"""
-        assert sorted(edx_user_data.enrollments.get_enrolled_course_ids()) == ids
-        assert sorted(edx_user_data.certificates.all_courses_verified_certs) == ids
-        assert sorted(edx_user_data.current_grades.all_course_ids) == ids
+        assert sorted(edx_user_data.enrollments.get_enrolled_course_ids()) == sorted(ids)
+        assert sorted(edx_user_data.certificates.all_courses_verified_certs) == sorted(ids)
+        assert sorted(edx_user_data.current_grades.all_course_ids) == sorted(ids)
 
     def test_edx_data_fetch_and_set(self):
         """Test that a user's edX data is properly fetched and set onto object properties"""
@@ -82,13 +85,6 @@ class CachedEdxUserDataTests(ESTestCase):
         p2_course_run_program = self.p2_course_run_1.course.program
         edx_user_data = CachedEdxUserData(self.user, program=p2_course_run_program)
         self.assert_edx_data_has_given_ids(edx_user_data, self.p2_course_run_keys)
-
-    def test_raw_data_fetch_and_set(self):
-        """Test that a user's raw edX data is properly fetched and set onto object properties"""
-        edx_user_data = CachedEdxUserData(self.user, include_raw_data=True)
-        assert edx_user_data.raw_enrollments == [obj.data for obj in self.enrollments]
-        assert edx_user_data.raw_certificates == [obj.data for obj in self.certificates]
-        assert edx_user_data.raw_current_grades == [obj.data for obj in self.current_grades]
 
 
 class CachedEdxDataApiTests(ESTestCase):
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -1,28 +1,28 @@
 """
 Provides functionality for serializing a ProgramEnrollment for the ES index
 """
-from decimal import Decimal
-
 from dashboard.api_edx_cache import CachedEdxUserData
+from dashboard.models import CachedEnrollment
 from dashboard.utils import MMTrack
-from roles.models import (
-    NON_LEARNERS,
-    Role
-)
+from roles.api import is_learner
 
 
 class UserProgramSearchSerializer:
     """
     Provides functions for serializing a ProgramEnrollment for the ES index
     """
-    @staticmethod
-    def calculate_final_grade_average(mmtrack):
+    @classmethod
+    def serialize_enrollments(cls, enrollments):
         """
-        Calculates an average grade (integer) from the program final grades
+        Serializes a user's enrollment data for search results
+
+        Args:
+            enrollments (iterable): An iterable of CachedEnrollments
+        Returns:
+            list: Serialized courses
         """
-        final_grades = mmtrack.get_all_final_grades()
-        if final_grades:
-            return round(sum(Decimal(final_grade) for final_grade in final_grades.values()) / len(final_grades))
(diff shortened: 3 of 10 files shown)


dashboard/serializers.py
```diff
@@ -1,28 +1,27 @@
...
-)
+from roles.api import is_learner
 
 
 class UserProgramSearchSerializer:
     """
     Provides functions for serializing a ProgramEnrollment for the ES index
     """
-    @staticmethod
-    def calculate_final_grade_average(mmtrack):
```
gsidebo: this was moved into `dashboard.utils.MMTrack`

dashboard/serializers.py
```diff
@@ -40,23 +39,8 @@ def serialize(cls, program_enrollment):
...
-            'grade_average': cls.calculate_final_grade_average(mmtrack),
-            'is_learner': cls.is_learner(user, program),
+            'enrollments': cls.serialize_enrollments(edx_user_data.enrollments.queryset),
+            'grade_average': mmtrack.calculate_final_grade_average(),
+            'is_learner': is_learner(user, program),
             'email_optin': user.profile.email_optin,
         }
-
-    @classmethod
-    def is_learner(cls, user, program):
```
gsidebo: this was moved into `roles.api`

dashboard/serializers.py
```diff
@@ -40,23 +39,8 @@ def serialize(cls, program_enrollment):
         )
         return {
             'id': program.id,
-            'enrollments': edx_user_data.raw_enrollments,
-            'certificates': edx_user_data.raw_certificates,
-            'current_grades': edx_user_data.raw_current_grades,
```
gsidebo: we weren't doing anything with 'certificates' or 'grades'

noisecapella:
I can do the review but @giocalitri can you take a quick glance to make sure all this is consistent with your grade work?

static/js/components/search/FilterVisibilityToggle.js
```diff
@@ -4,6 +4,11 @@ import _ from 'lodash';
 import { SearchkitComponent } from 'searchkit';
 import Icon from 'react-mdl/lib/Icon';
 
+const FILTER_ID_ADJUST = {
+  "birth_location": "profile.birth_country4",
+  "courses": "program.enrollments.title3"
```
gsidebo: i'm kind of worried about this. it seems like the element id's we're getting back from ES/searchkit are almost arbitrary. `profile.birth_country3` changed to `profile.birth_country4` as part of this PR for reasons unknown to me, and i don't know why the id for enrollments is `program.enrollments.title3`. @aliceriot any ideas?

noisecapella: I reviewed the PR which added the original and I couldn't find a simple way around this unfortunately. I think the different Searchkit filters pass different ids and fixing this properly would require modifying the Searchkit classes to do this properly

giocalitri:
I am a bit puzzled by the refactoring of the same code you ( @gsidebo ) wrote last week...

gsidebo:
@giocalitri new requirements, that's all. 2 major changes that affect that code:

1. we figured out that we aren't even using the raw edX data that we've been serializing
1. we need to serialize the course title based on cached enrollments. `CachedEdxUserData` only really needed to provide edX objects from our cached data. to serialize users for search results, we need those edX objects as well as course titles. 

for those reasons it made sense to refactor. i didn't have this exact issue in mind when i was working on that original code

dashboard/api_edx_cache.py
```diff
@@ -16,42 +16,31 @@
...
+        return self.edx_data_tuple(queryset, edx_object)
+
+
+class CachedEdxUserSearchData(CachedEdxUserData):
+    """Represents all edX data related to a User for use in search results"""
+    def fetch_and_set_edx_data_tuple(self, model_cls):
+        """Fetches a user's edX data, deserializes it to an edX API object, and returns a named tuple"""
+        queryset = model_cls.user_course_qset(self.user, program=self.program)
+        edx_object = model_cls.deserialize_edx_data([obj.data for obj in queryset])
+        return self.edx_data_tuple(queryset, edx_object)
```
giocalitri: Why do you need this class? What is different from the one it inherits from?

dashboard/api_edx_cache_test.py
```diff
@@ -62,17 +66,18 @@ def setUpTestData(cls):
 
     def assert_edx_data_has_given_ids(self, edx_user_data, ids):
         """Asserts that all edX object course id sets match the given list of ids"""
-        assert sorted(edx_user_data.enrollments.get_enrolled_course_ids()) == ids
-        assert sorted(edx_user_data.certificates.all_courses_verified_certs) == ids
-        assert sorted(edx_user_data.current_grades.all_course_ids) == ids
+        assert sorted(edx_user_data.enrollments.edx_object.get_enrolled_course_ids()) == ids
+        assert sorted(edx_user_data.certificates.edx_object.all_courses_verified_certs) == ids
+        assert sorted(edx_user_data.current_grades.edx_object.all_course_ids) == ids
```
giocalitri: you probably need to sort also the ids to be sure it will always work

dashboard/api_edx_cache.py
```diff
@@ -16,42 +16,31 @@
...
-        self.raw_current_grades = list(models.CachedCurrentGrade.data_qset(self.user, program=self.program))
+        self.enrollments = self.fetch_and_set_edx_data_tuple(models.CachedEnrollment)
+        self.certificates = self.fetch_and_set_edx_data_tuple(models.CachedCertificate)
+        self.current_grades = self.fetch_and_set_edx_data_tuple(models.CachedCurrentGrade)
+
+    def fetch_and_set_edx_data_tuple(self, model_cls):
+        """Fetches a user's edX data, deserializes it to an edX API object, and returns a named tuple"""
+        queryset = model_cls.data_qset(self.user, program=self.program)
+        edx_object = model_cls.deserialize_edx_data(queryset)
+        return self.edx_data_tuple(queryset, edx_object)
```
giocalitri: I think we should not return the queryset here. I thought that we established that in this case (cached data) the querysets are returned only by the model.

dashboard/serializers.py
giocalitri: why?

dashboard/serializers.py
giocalitri: I understand that `roles.api` is a better place for this, but given that it looks like the only place where we are using it is here, why did you have to refactor this?

dashboard/api_edx_cache.py
gsidebo: does 2 things differently:

1. prefetches the `course_run__course` related to the given edX model
1. since the queryset is not just getting a values list of `data` properties, it needs to pass in `[obj.data for obj in queryset]` to `deserialize_edx_data`

dashboard/api_edx_cache.py
gsidebo: we need the queryset when we serialize the user for search results (https://github.com/mitodl/micromasters/pull/2017/files#diff-5e24ceac8b57391a6805615b314c3da8R42). here's the way i thought about it: this class is representing a user's edX data. in one case (MMTrack) the only thing we care about in terms of a user's edX data is the actual edX objects. in another case (serialization for search) we need those edX objects as well as some related data (course titles, in this case)

dashboard/serializers.py
gsidebo: like below, proper separation of concerns. the job of this class is to serialize. one of the jobs of MMTrack is to provide information about final grades. it's just a better place for this method. if the calculation of final grades changes (and, as we know, it will be changing soon), you have one class to fix and you don't need to remember to make the change here

dashboard/api_edx_cache.py
gsidebo: looking at this again, in the search serialization case we only need the prefetched queryset for enrollments. i'll tinker with this a little

gsidebo: this example gets rid of the two separate classes:

```python
from dashboard.api_edx_cache import CachedEdxDataApi
class CachedEdxUserData:
    """Represents all edX data related to a User"""
    edx_data_tuple = namedtuple('EdxDataTuple', ['queryset', 'edx_object'])
    # Reverse CACHED_EDX_MODELS to allow a reference from a model class to a string name
    model_name_map = {v: k for k, v in CachedEdxDataApi.CACHED_EDX_MODELS}

    def __init__(self, user, program=None, prefetch=None):
        """Instantiates object and fetches all of a user's cached edX data"""
        self.user = user
        self.program = program
        prefetch_list = prefetch or []
        self.enrollments = self.get_edx_data_tuple(models.CachedEnrollment, prefetch_list)
        self.certificates = self.get_edx_data_tuple(models.CachedCertificate, prefetch_list)
        self.current_grades = self.get_edx_data_tuple(models.CachedCurrentGrade, prefetch_list)

    def get_edx_data_tuple(self, model_cls, prefetch_list):
        if self.model_name_map[model_cls] not in prefetch_list:
            # We only need the raw data
            queryset = model_cls.data_qset(self.user, program=self.program)
            data_iterable = queryset
        else:
            # We need the raw data as well as some prefetched related data
            queryset = model_cls.user_course_qset(self.user, program=self.program)
            data_iterable = [obj.data for obj in queryset]
        edx_object = model_cls.deserialize_edx_data(data_iterable)
        return self.edx_data_tuple(queryset, edx_object)
```

for MMTrack it would just be normal usage: `CachedEdxUserData(user)`
for search serialization: `CachedEdxUserData(user, program=program, prefetch=['enrollments'])`

we can also change this to use the actual model class to indicate that it needs to be prefetched: `CachedEdxUserData(user, program=program, prefetch=[CachedEnrollments])`. i think the above example is clear enough and takes care of our needs. let me know what you think

gsidebo:
@giocalitri comments have been addressed and i put a question back to you

dashboard/api_edx_cache.py
giocalitri: IMHO this is getting over complicated because too optimized. I think you should have two methods to get the two pieces of data and until we realize that that is a performance problem we should not worry about that.

dashboard/api_edx_cache.py
giocalitri: Again, I can tell at least another way that does not require to pass around the queryset.
I think that either you have a method that pulls the final data that you need or you use directly the queryset in the serialized.
This introduces a complication that is starting to be a bit too much.
I also think you need to comment more the code. What I read is not always self explanatory.

dashboard/serializers.py
giocalitri: Ok, but was it preventing you from implementing what you needed?

gsidebo: it was not preventing me from implementing what i needed. i looked at the method, realized there was a better place for it, checked its usages in the code, determined that the impact on the PR was extremely minimal, and decided to move the method. all of this happened in a matter of seconds. i understand that we want to limit the size of PRs and i took that into account

gsidebo:
@giocalitri i rolled back the `CachedEdxUserData` changes. please have a look

giocalitri:
I am fine with the status of the PR. I need probably to test the functionality.

giocalitri:
👍

PR #2021, opened 2016-12-07
Open external links in new tabs on public pages

### LINKED ISSUE #1823: FAQ's: Please make all links open in a new tab
RC
SCM Program
Links to URL's should open in a new tab. 
(likely an issue in other program's as well -- so it would make sense to change this across this board)

annagav:
Fix  #1823

For testing, try adding some links in program page content through cms, and make sure those links open in a new tab.

### DIFF
--- a/static/js/entry/public.js
+++ b/static/js/entry/public.js
@@ -180,5 +180,5 @@ $(document).ready(function ($) {
     find('.material-icons').removeClass('rotate-reset').addClass('rotate');
   });
   // All external links should open in new tab
-  $('a[href^="https://"]').attr('target', '_blank');
+  $('a[href^="https://"], a[href^="http://"] ').attr('target', '_blank');
 });
--- a/static/js/entry/public.js
+++ b/static/js/entry/public.js
@@ -179,4 +179,6 @@ $(document).ready(function ($) {
     $(".accordion-content").not($(this).next()).slideUp('fast').prev().
     find('.material-icons').removeClass('rotate-reset').addClass('rotate');
   });
+  // All external links should open in new tab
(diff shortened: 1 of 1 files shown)


noisecapella:
Are we intending this only to be for external links?

static/js/entry/public.js
```diff
@@ -179,4 +179,6 @@ $(document).ready(function ($) {
     $(".accordion-content").not($(this).next()).slideUp('fast').prev().
     find('.material-icons').removeClass('rotate-reset').addClass('rotate');
   });
+  // All external links should open in new tab
+  $('a[href^="https://"]').attr('target', '_blank');
```
noisecapella: This only works for `https` links, can you update for `http` as well?

annagav:
I understood the issue as concerning only external links. @pdpinch what do you think?

noisecapella:
Looks good to me, feel free to merge when @pdpinch confirms that we only want to do this for external links :+1:

Ferdi:
Required reading before this is merged  :-)

https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/  

/cc @roberthouse54

roberthouse54:
Might as well also read this, for the counter opinion: http://uxmovement.com/navigation/why-external-links-should-open-in-new-tabs/

Ferdi:
Yes, please also read the comments on the article you linked and come up with some recommendations about what to do in our case

roberthouse54:
I honestly do not have much of an opinion on this. I think is is one of those things that is debated as a UX thing, but I do not feel strongly about it one way or the other. Jared opened this issue for some reason, but I am not sure why. I think we should just do it, but I also do not think it is a very big issue.

PR #2025, opened 2016-12-07
Revert "Changed profile validation to not require a photo"

aliceriot:
This should restore the photo requirement. Delete your photo in the django shell, and ensure you can't navigate past the personal stage of the profile flow.


noisecapella:
:+1:

Issue #1752, opened 2016-11-14
Disable buttons during API activity


noisecapella:
We should disable buttons to prevent double clicking and to provide an indication to the user of activity being done.

Off the top of my head:
 - [x] signup next and previous buttons
 - [x] edit employment
 - [x] edit education
 - [ ] delete employment
  - [ ] delete work history
 - [ ] new financial aid button
 - [ ] skip financial aid button (also disable both financial aid buttons during either API request)
 - [ ] document sent button
 - [x] pay button
 - [x] enroll link
 - [x] user page dialog buttons
 - [x] photo upload dialog
 - [x] settings page button
 - [ ] enroll in program button
 - [ ] email buttons in search results

Ferdi:
Do you suggest we do them one by one or all at once ?

noisecapella:
One by one is probably best, each button does a different thing so it has a different callback

pdpinch:
This looks like it's in progress to me. I'm moving it.

PR #2031, opened 2016-12-07
Disabled buttons on signup and settings pages during profile update

noisecapella:
Part of #1752 

#### What's this PR do?
Disables buttons when updating the profile on the signup page, settings page and user page

This does not handle the user page at all or employment or education dialogs. Those will be a separate PR

Click next or save on the signup and settings page. The button should replace its text with a white spinner for the duration of the API activity.

### DIFF
--- a/static/js/components/EducationTab.js
+++ b/static/js/components/EducationTab.js
@@ -11,12 +11,12 @@ import { EDUCATION_STEP } from '../constants';
 
 class EducationTab extends React.Component {
   props: {
-    profile:      Profile,
-    profilePatchStatus: ?string,
-    ui:           UIState,
-    saveProfile:  SaveProfileFunc,
-    addProgramEnrollment: Function,
-    dispatch:     Function,
+    profile:               Profile,
+    profilePatchStatus:    ?string,
+    ui:                    UIState,
+    saveProfile:           SaveProfileFunc,
+    addProgramEnrollment:  Function,
+    dispatch:              Function,
   };
 
   componentWillMount() {
--- a/static/js/components/EducationTab.js
+++ b/static/js/components/EducationTab.js
@@ -12,6 +12,7 @@ import { EDUCATION_STEP } from '../constants';
 class EducationTab extends React.Component {
   props: {
     profile:      Profile,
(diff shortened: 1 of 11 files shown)


aliceriot:
functionality looks great! taking a look at the code now.

static/js/components/EducationTab.js
```diff
@@ -12,6 +12,7 @@ import { EDUCATION_STEP } from '../constants';
 class EducationTab extends React.Component {
   props: {
     profile:      Profile,
+    profilePatchStatus: ?string,
```
aliceriot: (same thing in the changes below)

static/js/components/ProfileProgressControls.js
```diff
@@ -55,9 +63,9 @@ export default class ProfileProgressControls extends React.Component {
     if (nextUrl) {
       nextButton = <button
         role="button"
-        className="mdl-button next"
-        onClick={this.saveAndContinue}>
-        <span>{nextBtnLabel}</span>
+        className={`mdl-button next ${disabled ? 'disabled-with-spinner' : ''}`}
+        onClick={disabled ? undefined : this.saveAndContinue}>
+        {nextBtnLabel}
```
aliceriot: could this just be

```js
{disabled ? <Spinner singleColor /> : nextBtnLabel }
```

I often find having the control flow in the JSX to be more readable than having something declared above, that's just my preference though I think.

static/js/components/ProfileProgressControls.js
```diff
@@ -42,7 +45,12 @@ export default class ProfileProgressControls extends React.Component {
   };
 
   render() {
-    const { nextUrl, prevUrl, nextBtnLabel } = this.props;
+    let { nextUrl, prevUrl, nextBtnLabel, profilePatchStatus } = this.props;
+
+    let disabled = profilePatchStatus === FETCH_PROCESSING;
```
aliceriot: since it's really capturing the fact that a request is currently being processed, what do you think about renaming `disabled` to be `inFlight` or similar?

noisecapella:
Comments should be addressed

PR #2033, opened 2016-12-07
Disable buttons on employment and education dialogs during profile update

noisecapella:
Part of #1752

#### What's this PR do?
Disables buttons for saving during profile upload for Education and Employment dialogs

Verify that the save buttons are disabled and that a blue spinner is shown in the buttons place. This should happen for employment and education, for both add and edit, and for both the signup page and the user page.

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -4,6 +4,7 @@ import IconButton from 'react-mdl/lib/IconButton';
 import Button from 'react-mdl/lib/Button';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 import { Card } from 'react-mdl/lib/Card';
+import Spinner from 'react-mdl/lib/Spinner';
 import _ from 'lodash';
 import R from 'ramda';
 import Dialog from 'material-ui/Dialog';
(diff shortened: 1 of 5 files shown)


aliceriot:
functionality looks great!

static/js/components/EducationForm.js
```diff
@@ -161,7 +164,7 @@ class EducationForm extends ProfileFormFields {
       userPrivilegeCheck(profile, () =>
         <Cell col={12} className="profile-form-row add" key="add-row">
           <button
-            className="mm-minor-action"
+            className="mm-minor-action add-education-button"
```
aliceriot: what's the reason for this change?

static/js/containers/UserPage_test.js
```diff
@@ -552,6 +552,29 @@ describe("UserPage", function() {
...
+          return listenForActions([
+            REQUEST_PATCH_USER_PROFILE,
+          ], () => {
+            helper.store.dispatch(requestPatchUserProfile(username));
+          }).then(() => {
+            let button = dialog.querySelector(".save-button");
+            assert(button.className.includes("disabled-with-spinner"));
+
+            TestUtils.Simulate.click(button);
+            assert.isFalse(patchUserProfileStub.called);
```
aliceriot: Maybe check that there's an element with class `.mdl-spinner` here too? Not sure how important that is.

static/js/components/EducationForm.js
aliceriot: oh, I see it being used in the test below, is that the reason for it?

noisecapella: yes

static/js/containers/UserPage_test.js
noisecapella: Sure, makes sense

Issue #2019, opened 2016-12-07
Profile upload spinner


noisecapella:
We should add a spinner somewhere while the profile image uploads

roberthouse54:
OK, here is a proposed solution:

1. Display a spinner in the modal while the image is loading in the browser.
2. Display another spinner and the text "Loading..." on a transparent overlay over the image after the user has clicked Save, and the image is being sent to the server.

See mocks below:

![spinner](https://cloud.githubusercontent.com/assets/20047260/20981294/104029fa-bc82-11e6-9725-50bcd0f1632a.png)

Ferdi:
A percentage based progress would be more useful if it's possible.

roberthouse54:
![spinner](https://cloud.githubusercontent.com/assets/20047260/20981748/d535379a-bc83-11e6-9b82-dde666975ca4.png)

noisecapella:
We will need to replace the `fetch()` with `XMLHttpRequest` to give us access to the completion percent. That's not a huge amount of work, just an extra complication

roberthouse54:
I updated the mock in the comment above. If the percentage is too much of a hassle, I think we can live without it.

aliceriot:
Oops, meant to close this in #2039

PR #2039, opened 2016-12-08
prevent multiple concurrent image uploads

### LINKED ISSUE #2020: Remove ability to start multiple profile image uploads by utilizing the two uploaders on /personal/profile
If you switch back and forth between the two profile upload sections currently on /profile/personal you can trigger multiple uploads. You can click on the grayed out area to exit the upload popup.

aliceriot:

closes #2020 

#### What's this PR do?

this persists the photo edit state if the request is currently in-flight, so that the user is prevented from issuing another request (to upload the profile image) if one is already in-flight. I also added an `<Spinner />` to the save button, which we should while the request is processing (should help with longer requests, I think).

this is how, on master, you can currently initiate multiple requests:

1. throttle your upload speed (gives you more wiggle room, 2G is probably sufficient)
2. add a new profile image, and click 'Cancel' after clicking 'Save'
3. Re-open the photo dialog, and add a new profile image.

if you look in the network tab of your browser's dev tools you should see two requests in flight.

If you follow the same steps here there should be only one, and the previous edit state should still be there (so the spinner should still be going and so on) until the request finishes.

### DIFF
--- a/static/js/components/ProfileImageUploader.js
+++ b/static/js/components/ProfileImageUploader.js
@@ -4,6 +4,7 @@ import Dialog from 'material-ui/Dialog';
 import R from 'ramda';
 import Dropzone from 'react-dropzone';
 import Button from 'react-mdl/lib/Button';
+import Spinner from 'react-mdl/lib/Spinner';
 
 import CropperWrapper from './CropperWrapper';
 import { FETCH_PROCESSING } from '../actions';
@@ -15,7 +16,7 @@ const dropZone = (startPhotoEdit, setPhotoError) => (
   <Dropzone
     onDrop={onDrop(startPhotoEdit)}
     style={{height: uploaderBodyHeight()}}
-    className="photo-active-item photo-dropzone"
+    className="photo-active-item photo-dropzone dashed-border"
     activeClassName="photo-active-item photo-dropzone active"
     accept="image/*"
     onDropRejected={() => setPhotoError('Please select a valid photo')}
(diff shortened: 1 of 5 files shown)


noisecapella:
I think we're planning on putting a spinner on the whole image so I think a spinner on the button is redundant: https://github.com/mitodl/micromasters/issues/2019

aliceriot:
Ah I hadn't seen that. TBH I prefer the spinner in the save button, but if we want it to replace the image I can implement that right now.

aliceriot:
@roberthouse54 what are your thoughts?

roberthouse54:
Hi @aliceriot Alice. Can you try doing it the way I designed in #2019 in this case?

aliceriot:
Sure, I was wondering what you thought about having the spinner in the save button as opposed to in the image. If you just want me to do it in the image that's fine, just wanted to get your thoughts on the other option (which I think is a little cleaner).

aliceriot:
ok! the spinner now takes the place of the image.

static/js/components/ProfileImageUploader.js
```diff
@@ -32,6 +33,31 @@ const uploaderBodyHeight = (): number => (
...
+  startPhotoEdit,
+  setPhotoError,
+  inFlight
+) => {
+  if ( inFlight ) {
+    return <div
+      className="photo-active-item dashed-border"
+      style={{height: uploaderBodyHeight()}}
+    >
+      <Spinner singleColor />
```
rhysyngsun: I think this spinner is supposed to be the MIT red, not blue, according the mocks in #2019.

static/js/containers/ProfileImage_test.js
```diff
@@ -149,7 +149,18 @@ describe('ProfileImage', () => {
...
-      it('should disable the save button while uploading the image', () => {
+      it('should show a spinner while uploading the image', () => {
+        renderProfileImage(true, {
+          profile: thatProfile
+        });
+        helper.store.dispatch(startPhotoEdit({name: 'a name'}));
+        helper.store.dispatch(setPhotoDialogVisibility(true));
+        helper.store.dispatch(requestPatchUserPhoto(SETTINGS.user.username));
+        let dialog = document.querySelector(".photo-upload-dialog");
+        assert.isNotNull(dialog.querySelector('.mdl-spinner'));
```
rhysyngsun: Is it worth asserting that the spinner isn't present during earlier stages of the editing/upload process?

aliceriot: that's probably a good idea! i'll add a few more tests.

aliceriot:
@rhysyngsun comments addressed

rhysyngsun:
:+1: assuming travis passes

PR #2042, opened 2016-12-08
Disable buttons during profile upload for about me and personal info dialogs

noisecapella:
Part of #1752

#### What's this PR do?
Disable buttons during profile upload for about me and personal info dialogs.

AFAIK these dialogs are only accessible on the user page. Edit the about me and click save, and edit the personal info and click save. A blue spinner should appear in place of the save button and more clicks should not cause the button to get clicked anymore.

### DIFF
--- a/static/js/components/UserPagePersonalDialog.js
+++ b/static/js/components/UserPagePersonalDialog.js
@@ -2,11 +2,11 @@
 import React from 'react';
 import Dialog from 'material-ui/Dialog';
 import Button from 'react-mdl/lib/Button';
-import Spinner from 'react-mdl/lib/Spinner';
 
 import { FETCH_PROCESSING } from '../actions';
 import { personalValidation } from '../lib/validation/profile';
 import PersonalForm from './PersonalForm';
+import SpinnerButton from './SpinnerButton';
 import type { Profile, SaveProfileFunc } from '../flow/profileTypes';
 import type { UIState } from '../reducers/ui';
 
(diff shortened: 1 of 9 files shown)


aliceriot:
functionality looks great!

static/js/components/UserPagePersonalDialog.js
```diff
@@ -46,10 +50,10 @@ export default class UserPagePersonalDialog extends React.Component {
       </Button>,
       <Button
         type='button'
-        className='primary-button save-button'
+        className={`primary-button save-button ${inFlight ? 'disabled-with-spinner' : ''}`}
         key='save'
-        onClick={this.savePersonalInfo}>
-        Save
+        onClick={inFlight ? undefined : this.savePersonalInfo}>
+        {inFlight ? <Spinner singleColor /> : 'Save'}
```
aliceriot: This is outside of the scope of this PR, but we have a lot of repeated code in these action buttons - it might make sense to pull that out as a higher level abstraction, maybe something like this:

```js
const profileDialogActions = (onCancel, onSave, inFlight) => ([
  <Button ... />,
  <Button ... />,
]);
```

aliceriot: Would certainly make it easier to change all these buttons in the future! haha

noisecapella: Sure, I'll investigate

aliceriot:
that looks like a good start, I was thinking of something more like the `createDialogActions` function in `static/js/components/EmailCompositionDialog.js`, but more abstracted so it could replace the `actions` declarations in all of the profile files with something like:

```js
actions={profileActionButtons(onClose, onSave, inFlight)}
```

just for boilerplate reduction. As I said, we don't have to do this here! Might make more sense to pull it into another pr.

noisecapella:
Sure, I'll make an issue for it

PR #2046, opened 2016-12-08
Remove React-Sticky

### LINKED ISSUE #1948: facets box goes on top of footer
screen shot 2016-12-02 at 2 31 27 pm

singingwolfboy:
Fixes #1948 by removing [`react-sticky`](https://github.com/captivationsoftware/react-sticky). The library doesn't have a way to keep a sticky element bounded within a container in the way that we need. Note that this change causes the facets sidebar to scroll with the rest of the page, so that scrolling down through the results causes the facets to hide from view.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,7 +9,7 @@
     },
     "abbrev": {
       "version": "1.0.9",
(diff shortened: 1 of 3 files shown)


aliceriot:
looks like there's a merge conflict, but the functionality looks good! looking at code now.

Issue #1945, opened 2016-12-02
Token refresh for user ends up in uncaught exception


giocalitri:
We should investigate and possible fix this.

The error is clear when the user is a made up user from seed_db.

see:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/188568358/

pdpinch:
fwiw, the user cdolivera is a real user. The account was created on prod with an edx.org account. When we copy that user down to -rc and -ci there is no corresponding user on their edX backends.

pdpinch:
@giocalitri any thoughts on what might be an appropriate fix?

giocalitri:
Maybe a generic exception handling.
I actually would like to understand why it is not working now.

On Dec 4, 2016, at 2:14 PM, Peter Pinch <notifications@github.com<mailto:notifications@github.com>> wrote:


@giocalitri<https://github.com/giocalitri> any thoughts on what might be an appropriate fix?

pdpinch:
I assume it's happening because we have accounts in MicroMasters-ci that don't exist on microdmasters.d

giocalitri:
That's true, but the code should not fail in that way, at least not in the celery task.

On Dec 4, 2016, at 3:17 PM, Peter Pinch <notifications@github.com<mailto:notifications@github.com>> wrote:

I assume it's happening because we have accounts in MicroMasters-ci that don't exist on microdmasters.d

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/189430044/

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/188568358/

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/182642956/

giocalitri:
> https://sentry.io/mit-office-of-digital-learning/micromasters/issues/182642956/

This is different from the ones above: this might be caused by the user creating a brand new user and not verifying the user on edX.
Need to double check.

giocalitri:
All the ones before might have different root causes as well: I am in the process to put out a PR to have more granular errors.

pdpinch:
This particular error was resolved by #2047. There are some related errors, which have their own issues now.

PR #2047, opened 2016-12-08
Refactored task code to refresh users edX data.

giocalitri:
part of #1945 

#### What's this PR do?
adds a more granular logging to the task handling the refresh of the edX data

`dashboard/tasks.py`

make sure that whatever happens, the refresh of the user cached data will not end up in a killed task.

### DIFF
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -99,31 +99,42 @@ def batch_update_user_data_subtasks(students):
     Args:
         students (List): List of students.
     """
+    # pylint: disable=bare-except
     for user_id in students:
         try:
-            # pylint: disable=broad-except
             user = User.objects.get(pk=user_id)
+        except:
+            log.exception('edX data refresh task: unable to get user "%s"', user_id)
+            continue
 
-            if UserSocialAuth.objects.filter(user=user).exists():
-                # get the credentials for the current user for edX
-                user_social = user.social_auth.get(provider=EdxOrgOAuth2.name)
+        try:
+            UserSocialAuth.objects.get(user=user)
+        except:
+            log.exception('user "%s" does not have python social auth object', user.username)
+            continue
 
-                try:
-                    utils.refresh_user_token(user_social)
-                except utils.InvalidCredentialStored as exc:
-                    log.error(
-                        "Unable to refresh token for student %s, status code is %d and error is %s",
-                        user.username, exc.http_status_code, str(exc)
-                    )
-                    continue
+        # get the credentials for the current user for edX
+        try:
+            user_social = user.social_auth.get(provider=EdxOrgOAuth2.name)
+        except:
+            log.exception('user "%s" does not have edX credentials', user.username)
+            continue
 
-                edx_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL)
+        try:
+            utils.refresh_user_token(user_social)
+        except:
+            log.exception("Unable to refresh token for student %s", user.username)
+            continue
 
-                for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
-                    CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
+        try:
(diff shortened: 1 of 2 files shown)


dashboard/tasks.py
```diff
@@ -99,31 +99,40 @@ def batch_update_user_data_subtasks(students):
     Args:
         students (List): List of students.
     """
+    # pylint: disable=broad-except
     for user_id in students:
         try:
-            # pylint: disable=broad-except
             user = User.objects.get(pk=user_id)
+        except Exception:
+            log.exception("edX data refresh task: unable to get user")
```
noisecapella: Can we add the user id to the message?

noisecapella: Can we add the user id to the message?

dashboard/tasks.py
```diff
@@ -99,31 +99,40 @@ def batch_update_user_data_subtasks(students):
     Args:
         students (List): List of students.
     """
+    # pylint: disable=broad-except
     for user_id in students:
         try:
-            # pylint: disable=broad-except
             user = User.objects.get(pk=user_id)
+        except Exception:
```
noisecapella: I think we can just use `except:` here and below

dashboard/tasks_test.py
```diff
@@ -44,26 +43,96 @@ def setUpTestData(cls):
...
         """
-        Assert get_student_enrollments is actually called
+        Assert get_student_enrollments is actually called in happy path
         """
         batch_update_user_data_subtasks.s(self.students).apply(args=()).get()
         assert mocked_refresh_cache.call_count == len(self.all_users) * len(CachedEdxDataApi.SUPPORTED_CACHES)
         assert mocked_refresh.call_count == len(self.all_users)
         for user, cache_type in product(self.all_users, CachedEdxDataApi.SUPPORTED_CACHES):
             mocked_refresh_cache.assert_any_call(user, mock.ANY, cache_type)
             mocked_refresh.assert_any_call(user.social_auth.get(provider=EdxOrgOAuth2.name))
```
noisecapella: Can you add a test where the first user has a failure somewhere but the second user doesn't, and therefore the mocked methods are called only for the second user?

dashboard/tasks_test.py
```diff
@@ -44,26 +43,96 @@ def setUpTestData(cls):
...
+
+    @mock.patch('dashboard.api_edx_cache.CachedEdxDataApi.update_cache_if_expired', new_callable=mock.MagicMock)
+    @mock.patch('backends.utils.refresh_user_token', autospec=True)
+    def test_subtask_social_auth_refresh_fails(self, mocked_refresh, mocked_refresh_cache):
+        """
+        Test if the refresh for the user's OAUTH token fails the task still completes
+        """
+        mocked_refresh.side_effect = InvalidCredentialStored
+        batch_update_user_data_subtasks.s(self.students).apply(args=()).get()
+        assert mocked_refresh.called is True
```
noisecapella: Can you assert the arguments passed here and in the other cases below? We should make sure it's called once per user, or for `mocked_refresh_cache` once per user per cache type

dashboard/tasks_test.py
```diff
@@ -29,41 +29,114 @@ class TasksTest(TestCase):
...
+    @mock.patch('backends.utils.refresh_user_token', autospec=True)
+    def test_subtask_user_does_not_exist(self, mocked_refresh, mocked_refresh_cache):
+        """
+        Test if the user has been deleted between the select and the run of the task,
+        the task still completes.
+        """
+        with mock.patch('django.contrib.auth.models.User.objects.get', autospec=True, side_effect=User.DoesNotExist):
+            batch_update_user_data_subtasks.s(self.students).apply(args=()).get()
+        assert mocked_refresh.called is False
+        assert mocked_refresh_cache.called is False
```
singingwolfboy: Nitpick: it's clearer to use `not`:

```python
assert not mocked_refresh.called
assert not mocked_refresh_cache.called
```

Issue #1568, opened 2016-10-26
Notify user when enrollment status doesn't match paid status


noisecapella:
I think a simple solution here is as follows:
1. For any situation where a user is paid, but not enrolled, we show a status that  says "Something went wrong. You paid for this course but are not enrolled. Contact us for help."
2. In the case where the user pays, and there is an error and they are not enrolled, we show a failure toast message at the top of the page, in addition to the status message next to the button. 

See the proposed design below:
![image](https://cloud.githubusercontent.com/assets/20047260/20841498/06eef20e-b882-11e6-8974-d0435e3f3984.png)

roberthouse54:
I need to learn more about this situation.

pdpinch:
There are two conditions (George, correct me if I'm wrong):

1. User has paid us for a course, but for some reason, the edX automatic enrollment has failed (this happened once when edX made a breaking change to their enrollment API. That shouldn't happen again, but you don't know). 

2. The user has paid us for a course but then decided to unenroll themselves (There is an interface in edX to do this). 

In both cases, this is a degenerate state and we'd like the user to "try again" or contact us (possibly for a refund of some kind).

roberthouse54:
This raises a few questions. First, in case #2... Are we going to give refunds to users who want to un-enroll from a course? How does that work? Second, in case #1, if a user has paid but is still unenrolled, what should we encourage them to do, if anything? The UI, as it is, allows a user to enroll by paying for the course. If the user has already paid, they should be enrolled. It seems that we should avoid having this happen, and if it does happen, I don't know that the user needs to do anything. Instead, could we do something in the background to fix the problem? For example, just show them a simple error message (eg. "Something went wrong, and you are not enrolled in the course on edX. We are working on fixing it.") We should monitor for when (if) this ever happens, and have some mechanism for fixing it.

pdpinch:
1. I think a "Something went wrong" message is appropriate. Easiest for us is for the user to try again or try enrolling on the edX web site. A more generic solution though would be for them to contact us (help button or e-mail micromasters-support@mit.edu). We could cover other edge cases in that case. 

2. We will offer refunds, at least in some cases. The process I've been modeling is the user contacts micromasters-support@mit.edu and asks for a refund. If the request is accepted, then we have a set of back-office tasks to do which include setting their order in the portal to the "refund" state and unenrolling the user. 

Criteria for accepting a refund request have not been established yet.

roberthouse54:
[...]

PR #2048, opened 2016-12-09
Notified user when enrollment status doesn't match paid status

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/1568

#### What's this PR do?
- show error message when user paid for course but not enroll.

screen shot 2016-12-09 at 5 50 02 pm

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -148,7 +148,7 @@ class MMTrack:
 
     def paid_but_not_enrolled_mmtrack(self, course_id):
         """
-        Returns whether
+        Returns true if user paid for a course but he is not enrolled.
 
         Args:
             course_id (str): an edX course run id
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -146,6 +146,27 @@ class MMTrack:
         else:
             return course_id in self.paid_course_ids
 
+    def paid_but_not_enrolled_mmtrack(self, course_id):
+        """
+        Returns whether
+
+        Args:
+            course_id (str): an edX course run id
+
+        Returns:
+            bool: whether the user is paid but no enrolled mmtrack in the course
+        """
+        if self.is_enrolled(course_id):
+            return False
+
+        # financial aid programs need to have an audit enrollment and a paid entry for the course
+        if self.financial_aid_available:
+            return course_id in self.paid_course_ids
+
+        # need to find a way to check if user paid for course
+        # for normal programs and he is not entolled
+        return False
+
     def has_passed_course(self, course_id):
         """
         Returns whether the user has passed a course run.
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -154,7 +154,7 @@ class MMTrack:
 
         # normal programs need to have a verified enrollment
         enrollment = self.enrollments.get_enrollment_for_course(course_id)
-        return enrollment.is_verified
+        return enrollment and enrollment.is_verified
 
     def has_passed_course(self, course_id):
         """
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -136,36 +136,25 @@ class MMTrack:
         Returns:
             bool: whether the user is enrolled mmtrack in the course
         """
-        if not self.is_enrolled(course_id):
-            return False
-        # normal programs need to have a verified enrollment
-        if not self.financial_aid_available:
-            enrollment = self.enrollments.get_enrollment_for_course(course_id)
-            return enrollment.is_verified
-        # financial aid programs need to have an audit enrollment and a paid entry for the course
-        else:
-            return course_id in self.paid_course_ids
+        return self.is_enrolled(course_id) and self.has_paid(course_id)
 
-    def paid_but_not_enrolled_mmtrack(self, course_id):
+    def has_paid(self, course_id):
         """
-        Returns true if user paid for a course but he is not enrolled.
+        Returns true if user paid for a course.
 
         Args:
             course_id (str): an edX course run id
 
         Returns:
-            bool: whether the user is paid but no enrolled mmtrack in the course
+            bool: whether the user is paid
         """
-        if self.is_enrolled(course_id):
-            return False
-
         # financial aid programs need to have an audit enrollment and a paid entry for the course
         if self.financial_aid_available:
             return course_id in self.paid_course_ids
 
-        # need to find a way to check if user paid for course
-        # for normal programs and he is not entolled
-        return False
+        # normal programs need to have a verified enrollment
+        enrollment = self.enrollments.get_enrollment_for_course(course_id)
+        return enrollment.is_verified
 
     def has_passed_course(self, course_id):
         """
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -19,9 +19,10 @@ import {
   STATUS_WILL_ATTEND,
   STATUS_OFFERED,
   STATUS_PENDING_ENROLLMENT,
+  STATUS_PAID_BUT_NOT_ENROLLED,
   DASHBOARD_FORMAT,
   FA_PENDING_STATUSES,
-  FA_STATUS_SKIPPED
+  FA_STATUS_SKIPPED,
 } from '../../constants';
 import { isCurrentlyEnrollable } from './util';
 import { formatPrice } from '../../util/util';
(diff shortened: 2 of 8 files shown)


dashboard/utils.py
```diff
@@ -142,6 +142,27 @@ def is_enrolled_mmtrack(self, course_id):
...
+            bool: whether the user is paid but no enrolled mmtrack in the course
+        """
+        if self.is_enrolled(course_id):
+            return False
+
+        # financial aid programs need to have an audit enrollment and a paid entry for the course
+        if self.financial_aid_available:
+            return course_id in self.paid_course_ids
+
+        # need to find a way to check if user paid for course
```
amir-qayyum-khan: @giocalitri @pdpinch is their way we can know if user paid for a course (programs without financial aid) and not enrolled?

static/js/components/dashboard/CourseAction.js
```diff
@@ -192,6 +208,12 @@ export default class CourseAction extends React.Component {
       action = this.renderEnrollButton(run);
       description = this.renderTextDescription('Processing...');
       break;
+    case STATUS_PAID_BUT_NOT_ENROLLED:
+      description = this.renderContactUsDescription(
+        'Something went wrong.\n You paid for this course but \nare not enrolled.',
+        'mailto:micromasters-support@mit.edu'
```
noisecapella: Yes, you should use `settings.EMAIL_SUPPORT`

noisecapella: It should already be available in the javascript as `SETTINGS.support_email`

static/js/components/dashboard/CourseAction.js
```diff
@@ -18,9 +18,10 @@ import {
   STATUS_WILL_ATTEND,
   STATUS_OFFERED,
   STATUS_PENDING_ENROLLMENT,
+  STATUS_PAID_BUT_NOT_ENROLLED,
```
noisecapella: Please remember to update `ALL_STATUSES` in `constants.js`

dashboard/utils.py
```diff
@@ -142,6 +142,27 @@ def is_enrolled_mmtrack(self, course_id):
         else:
             return course_id in self.paid_course_ids
 
+    def paid_but_not_enrolled_mmtrack(self, course_id):
```
noisecapella: - Can you make a function `has_paid` which just returns `True` or `False` depending on if they paid? You should extract the logic from `is_enrolled_mmtrack` which checks both FA and non-FA cases.
 - Then change `is_enrolled_mmtrack` to use this function `has_paid`
 - Then you can change this function to `return self.has_paid(course_id) and not self.is_enrolled(course_id)`

static/js/components/dashboard/CourseAction_test.js
```diff
@@ -121,6 +122,23 @@ describe('CourseAction', () => {
     assert.equal(elements.descriptionText, 'In Progress');
   });
 
+  it('shows a message for course user paid but not enrolled', () => {
```
noisecapella: Can you test for the support email address too?

static/js/components/dashboard/CourseAction.js
```diff
@@ -112,12 +113,27 @@ export default class CourseAction extends React.Component {
     }
   );
 
+  renderDescriptionForError = R.curry(
```
noisecapella: Since we're only using this in one place it seems like we should just move this code below to `the `STATUS_PAID_BUT_NOT_ENROLLED` case. Then we can remove the checks for `contact` and `text`

static/js/components/dashboard/CourseAction.js
```diff
@@ -192,6 +208,12 @@ export default class CourseAction extends React.Component {
       action = this.renderEnrollButton(run);
       description = this.renderTextDescription('Processing...');
       break;
+    case STATUS_PAID_BUT_NOT_ENROLLED:
+      description = this.renderContactUsDescription(
+        'Something went wrong.\n You paid for this course but \nare not enrolled.',
```
noisecapella: This copy text is out of date. Can you use the text in the screenshot on the issue? It looks like we don't need the email address either

static/js/components/dashboard/CourseAction_test.js
```diff
@@ -121,6 +122,23 @@ describe('CourseAction', () => {
...
+    ));
+    let firstRun = course.runs[0];
+    const wrapper = renderCourseAction({
+      courseRun: firstRun
+    });
+    let elements = getElements(wrapper);
+
+    assert.equal(
+      elements.descriptionText,
+      'Something went wrong.\n You paid for this course but \nare not enrolled.Contact us for help.'
```
noisecapella: The `\n` characters are not visible in HTML

static/js/components/dashboard/CourseAction.js
amir-qayyum-khan: this is the same text. i dont find any difference

noisecapella:
Regarding the text, it should match this image which comes from the issue: image

Can you confirm @roberthouse54?

amir-qayyum-khan:
@noisecapella i am getting on pay now button 

```python
 
web_1      | During handling of the above exception, another exception occurred:
web_1      | 
web_1      | Traceback (most recent call last):
web_1      |   File "/usr/local/lib/python3.5/site-packages/django/template/base.py", line 885, in _resolve_lookup
web_1      |     current = current[bit]
web_1      | TypeError: 'WSGIRequest' object is not subscriptable
web_1      | 
web_1      | During handling of the above exception, another exception occurred:
web_1      | 
web_1      | Traceback (most recent call last):
web_1      |   File "/usr/local/lib/python3.5/site-packages/django/template/base.py", line 893, in _resolve_lookup
web_1      |     current = getattr(current, bit)
web_1      | AttributeError: 'WSGIRequest' object has no attribute 'user'
web_1      | 
web_1      | During handling of the above exception, another exception occurred:
web_1      | 
web_1      | Traceback (most recent call last):
web_1      |   File "/usr/local/lib/python3.5/site-packages/django/template/base.py", line 900, in _resolve_lookup
web_1      |     current = current[int(bit)]
web_1      | ValueError: invalid literal for int() with base 10: 'user'
web_1      | 
web_1      | During handling of the above exception, another exception occurred:
web_1      | 
web_1      | Traceback (most recent call last):
web_1      |   File "/usr/local/lib/python3.5/site-packages/django/template/base.py", line 907, in _resolve_lookup
web_1      |     (bit, current))  # missing attribute
web_1      | django.template.base.VariableDoesNotExist: Failed lookup for key [user] in "<WSGIRequest: POST '/dashboard/null'>"
```

amir-qayyum-khan:
@noisecapella are u referring to toast in the image. if yes then that toast will only appear once, when  user paid but fail to enroll. see description of https://github.com/mitodl/micromasters/issues/1568#issue-185420068

noisecapella:
@pdpinch @roberthouse54 Do you know which copy is correct? Is it the text in the description or the text in the screenshot?

noisecapella:
@amir-qayyum-khan That exception looks like one which commonly happens due to the template using variables which aren't defined. We are only seeing these exceptions because our log level is `DEBUG`. If you change your log level to `INFO` in the environment variables you should not see those particular exceptions anymore

roberthouse54:
We should remove the text that says "Click the link below to try to enroll again" from the toast message.

noisecapella:
But what should the rest of the text say? Should it start with 'There was an error' as in the screenshot or 'Something went wrong'?

roberthouse54:
OK, let's be consistent and show the same text in both places. Start with "Something went wrong"

amir-qayyum-khan:
@noisecapella i am blocked on one thing 

 http://192.168.99.100:8079/api/v0/checkout/
**request**: {course_id: "course-v1:edX+DemoX+Demo_Course"}
**response**: {detail: "Not found."}

amir-qayyum-khan:
@noisecapella while debugging it worked may be after i set dummy```CYBERSOURCE_SECURE_ACCEPTANCE_URL``` on .env file

amir-qayyum-khan:
@pdpinch @noisecapella regarding this

>> In the case where the user pays, and there is an error and they are not enrolled, we show a failure toast message at the top of the page, in addition to the status message next to the button.

where should we capture this error? each time user refresh page?

pdpinch:
@amir-qayyum-khan I'm not sure which error you mean when you say "this error" 

The toast message should appear only once, when the user clicked the pay button and enrollment failed. 

The status message next to the button should be displayed whenever the page is loaded and the user is in the state of having paid but isn't enrolled.

amir-qayyum-khan:
@noisecapella done

noisecapella:
Functionality looks good, one last look at the code...

PR #2057, opened 2016-12-09
Adds branded micromasters logo in navbar

roberthouse54:
#### What's this PR do?

This was by request of the MIT Office of Communications, that we use their standard layout for branding the site with logo etc. This change is for the marketing site only. Basically, it replaces the current text in the header that says "MITx Micromasters" with an image that says the same thing. I also replaced the MIT logo. Both logos have a hover state. As part of this I also removed target="blank" on the mit.edu link.

See that this works in different browsers. etc.


noisecapella:
Looks fine to me :+1:

PR #2060, opened 2016-12-09
Nl/1795 export pearson profiles

### LINKED ISSUE #1795: Exams: Export user profiles for syncing with Pearson
In order to create user accounts in the Pearson VUE system, we need to regularly export user profiles to CDD file (Candidate Demographic Data), which is a simple tab-separated file with one row per profile, starting with a header row. The files are stored on an sftp server at Pearson.  After the data are ingested, a file of results will be dropped off in the same location (see #). We can think of this as "RPC over SFTP"

## Acceptance criteria

- [ ] on a scheduled basis, create a tab-separated file containing any profiles that have been created or modified since the last export. 
- [ ] If no profiles have been updated, create a tsv file with just headers
- [ ] filename in the form of `cdd-YYYY- MM-DD- nn.dat`
- [ ] Name and address fields in the tsv file are limited to characters in [CP-1252](https://en.wikipedia.org/wiki/Windows-1252). However, all data is still UTF-8 encoded.
- [ ] Upload the exported tsv files to the SFTP server.

## Questions

- [ ] For Pearson's validation of the profiles, will we use the "all or none" validation method or the "as much as possible" validation method? Pearson recommends all or none. 
- [ ] Since Pearson only updates profiles that have more recent LastUpdated fields, should we just dump the entire database, instead of trying to limit export to profiles that have changed?

More details at https://drive.google.com/drive/folders/0B5lnP24agPPYWmtodUFuSTVxVmc

rhysyngsun:

Fixes #1795 

#### What's this PR do?

This PR adds export of the profile CCD files to Pearson. These are tab-delimited files encoded in UTF-8 and we upload via SFTP to their server. This is done by writing out these files locally and uploading them from the filesystem I added a new `exams.ExamProfile` model to track which profiles should be synced and the status. This is reset on any profile change. 

`exams/pearson.py`

There is non-public documentation on this "API" that Pearson provided us. If you need access to that I believe Peter can give it to you.

giphy


rhysyngsun:
Switched this back to a WIP, I want do a quick refactor to support the next issue better.

exams/pearson_test.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        Tests ccd_writer against a set of profiles
+        """
+        file = io.StringIO()
+
+        exam_profiles = []
+        with mute_signals(post_save):
+            exam_profiles.append(ExamProfileFactory.create(profile=ProfileFactory.create()))
+            exam_profiles.append(ExamProfileFactory.create(profile=ProfileFactory.create(address3='Room B345')))
+            exam_profiles.append(ExamProfileFactory.create(profile=ProfileFactory.create(address2=None)))
```
singingwolfboy: When you're dealing with Factory Boy factories that use subfactories and related factories, you can set attributed in the dependent factories easily by using double-underscores to separate the relations. For example, you can use this code to accomplish the same thing:

```python
exam_profiles.append(ExamProfileFactory.create())
exam_profiles.append(ExamProfileFactory.create(profile__address3='Room B345'))
exam_profiles.append(ExamProfileFactory.create(profile__address2=None))
```

Then you can reduce the code repetition using `**kwargs` and a list comprehension:
```python
exam_profile_kwargs = [{}, {profile__address3: 'Room B345'}, {profile__address2: None}]
with mute_signals(post_save):
    exam_profiles = [ExamProfileFactory.create(**kwargs) for kwargs in exam_profile_kwargs]
```

exams/pearson_test.py
```diff
@@ -0,0 +1,191 @@
...
+                profile.address2 or '',
+                profile.address3 or '',
+                profile.city,
+                profile.state_or_territory,
+                profile.postal_code,
+                country.alpha_3,
+                profile.phone_number,
+                profile.phone_country_code,
+                profile.updated_on.strftime(PEARSON_DATETIME_FORMAT),
+            ])))
```
singingwolfboy: This test feels a little too high-level to me. I understand the desire to not repeat yourself, but it's hard to see what the expected inputs and expected outputs actually are. Can we add a really concrete test? Something like:

```python
profile = ExamProfileFactory.create(
    student_id=37,
    romanized_first_name="Carl",
    romanized_last_name="Smith",
    user__email="abc@example.com",
    address1="123 Main Street",
    city="Springfield",
    state_or_territory="MA",
    postal_code="98765",
    phone_number="999-999-9999",
    phone_country_code="1",
    updated_on=datetime(2016, 12, 15, 10, 15, 0),
)
# ...
assert exported_row == (
    "37\tCarl\tSmith\tabc@example.com\t"
    "123 Main Street\t\t\tSpringfield\tMA\t"
    "98765\tUSA\t"
    "999-999-9999\t1\t"
    "2016/12/15 10:15:00\n"
)
```

That way, it becomes much easier to see what you actually expect the input and output formats to look like.

exams/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@async.task
+def export_exam_profiles():
+    """
+    Sync any outstanding profiles
+    """
+    exam_profiles = ExamProfile.objects \
+        .filter(status=ExamProfile.PROFILE_PENDING) \
+        .select_related('profile')
```
singingwolfboy: You don't need the trailing `\` if you put the entire line within parentheses:
```python
exam_profiles = (ExamProfile.objects
    .filter(status=ExamProfile.PROFILE_PENDING)
    .select_related('profile')
)
```

docker-compose.yml
```diff
@@ -81,11 +87,25 @@ services:
...
       - redis
+      - sftp
+
+  sftp:
+    image: atmoz/sftp
+    volumes:
+        - sftp_share:/home/odl/share
+    ports:
+        - "2022:22"
+    command: odl:123:1001:1001:share
```
noisecapella: What's the `1001` here?

rhysyngsun: Unix user/group id for the `odl` user getting created here.  The argument syntax the author chose is `user:pass[:uid[:gid[:dir]...]]`, all I really wanted was `share`, so I had to specify all of the args leading up to it.

exams/exceptions.py
```diff
@@ -0,0 +1,17 @@
+"""
+Exceptions for exams
+"""
+
+
+class InvalidTsvRow(Exception):
+    """
+    A row for a tsv is invalid
+    """
+    pass
```
noisecapella: We don't need the `pass` here and below, the docstring should take care of that

exams/models.py
```diff
@@ -0,0 +1,37 @@
...
+    PROFILE_PENDING = 'pending'
+    PROFILE_IN_PROGRESS = 'in-progress'
+    PROFILE_FAILED = 'failed'
+    PROFILE_SUCCESS = 'success'
+
+    PROFILE_STATUS_CHOICES = (
+        (PROFILE_PENDING, 'Sync Pending'),
+        (PROFILE_IN_PROGRESS, 'Sync in Progress'),
+        (PROFILE_FAILED, 'Sync Failed'),
+        (PROFILE_SUCCESS, 'Sync Suceeded'),
```
noisecapella: Should be `succeeded`

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+    InvalidTsvRow,
+)
+
+# custom csv dialect for Pearson
+csv.register_dialect(
+    'pearsontsv',
+    delimiter='\t',
+)
+
+PEARSON_DATETIME_FORMAT = "%Y/%m/%d %H:%M:%S"
```
noisecapella: Can you link me to any documentation for this, if they have any?

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+import pysftp
+
+from exams.exceptions import (
+    InvalidProfileDataException,
+    InvalidTsvRow,
+)
+
+# custom csv dialect for Pearson
+csv.register_dialect(
+    'pearsontsv',
```
noisecapella: Can we make `pearsontsv` a constant and use it below as well?

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+)
+
+PEARSON_DATETIME_FORMAT = "%Y/%m/%d %H:%M:%S"
+
+log = logging.getLogger(__name__)
+
+
+def format_datetime(dt):
+    """
+    Formats a datetime to Pearson's required format
```
noisecapella: Can you add the argument and return value using the sphinx-napoleon syntax? We generally follow the edX styleguide for that: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#docstrings

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    return dt.strftime(PEARSON_DATETIME_FORMAT)
+
+
+def get_field_mapper(field):
+    """
+    Returns a field mapper, accepts either a property path in str form or a callable
```
noisecapella: Please add args and a return value here

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+        return attrgetter(field)
+    elif callable(field):
+        return field
+    else:
+        raise TypeError("field_mapper must be a str or a callable")
+
+
+def writer(fields, field_prefix=None):
+    """
+    Creates a new writer for the given field mappins
```
noisecapella: Should this be `mapping`?

noisecapella: Also add args and a return value here please

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+                invalid_rows.append(row)
+
+        return (valid_rows, invalid_rows)
+
+    return _writer
+
+
+def optional_field(field, value):
+    """
+    Returns value if it is not None, otherwise an empty string
```
noisecapella: It looks like it checks if `value` is truthy, not that it's not `None`. If this is the behavior we want can you change to `if value is not None` below?

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+            ...
+        ])
+
+        obj = SourceObj(prop1=1234)
+
+        test_writer(file, [obj])
+    """
+    columns = [column for column, _ in fields]
+    field_mappers = [(column, get_field_mapper(field)) for column, field in fields]
+    prefix_mapper = attrgetter(field_prefix) if field_prefix else None
```
noisecapella: Can you change to `if field_prefix is not None` here, to be explicit?

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+    cnopts = pysftp.CnOpts()
+    cnopts.hostkeys = None  # ignore knownhosts
+    with pysftp.Connection(
+        host=str(settings.EXAMS_SFTP_HOST),
+        port=int(settings.EXAMS_SFTP_PORT),
+        username=str(settings.EXAMS_SFTP_USERNAME),
+        password=str(settings.EXAMS_SFTP_PASSWORD),
+        cnopts=cnopts,
+    ) as sftp:
+        with sftp.cd(str(settings.EXAMS_SFTP_UPLOAD_DIR)):
```
noisecapella: Why do we need to use `str(...)` here? If `EXAMS_SFTP_UPLOAD_DIR` is not set we should raise an exception here. If there is a sensible default we should define it in `settings.py`

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    # Pearson requires ISO-3166 alpha3 codes, but we store as alpha2
+    try:
+        country = pycountry.countries.get(alpha_2=profile.country)
+    except KeyError:
+        raise InvalidProfileDataException()
+    return country.alpha_3
+
+
+ccd_writer = writer([
```
noisecapella: What does `ccd` stand for?

exams/pearson.py
```diff
@@ -0,0 +1,147 @@
...
+
+def profile_country_to_alpha3(profile):
+    """
+    Returns the alpha3 code of a profile's country
+    """
+    # Pearson requires ISO-3166 alpha3 codes, but we store as alpha2
+    try:
+        country = pycountry.countries.get(alpha_2=profile.country)
+    except KeyError:
+        raise InvalidProfileDataException()
```
noisecapella: Can you pass the `KeyError` exception to `InvalidProfileDataException` in so we have the stack trace for it?

micromasters/settings.py
```diff
@@ -553,3 +554,10 @@ def show_toolbar(request):
         return not excluded
 
     DEBUG_TOOLBAR_CONFIG = {"SHOW_TOOLBAR_CALLBACK": show_toolbar, }
+
+# Exams SFTP
+EXAMS_SFTP_HOST = get_var('EXAMS_SFTP_HOST', 'localhost')
+EXAMS_SFTP_PORT = get_var('EXAMS_SFTP_PORT', '22')
+EXAMS_SFTP_USERNAME = get_var('EXAMS_SFTP_USERNAME', '')
```
noisecapella: If we don't have a good default for this can we default it to `None` instead and raise an exception where it's used?

exams/pearson.py
noisecapella: Actually, if you only use these functions in this file feel free to just add a `_` prefix

exams/pearson.py
rhysyngsun: "Candidate Demographic Data" (yes, just noticed the typo...we've been saying "CCD" in our discussions and I never realized we were wrong). I can add a comment in here if that's helpful.

rhysyngsun:
@singingwolfboy @noisecapella implemented your suggestions. George, most of the methods you commented lacking arguments and return values were meant to be private so I adjusted them as such.

noisecapella:
Fine with me, I'll take another look soon, focusing on tests

rhysyngsun:
Looks like this needs some merge conflicts from master resolved now.

rhysyngsun:
Rebased off master to get rid of the conflicts.

noisecapella:
How do I do a manual test for this?

exams/pearson.py
```diff
@@ -0,0 +1,162 @@
...
+PEARSON_DATETIME_FORMAT = "%Y/%m/%d %H:%M:%S"
+
+log = logging.getLogger(__name__)
+
+
+def _format_datetime(dt):
+    """
+    Formats a datetime to Pearson's required format
+    """
+    return dt.strftime(PEARSON_DATETIME_FORMAT)
```
noisecapella: Do they expect this in any particular time zone?

exams/pearson_test.py
```diff
@@ -0,0 +1,214 @@
...
+        assert row == (
+            "14879\tJane\tSmith\tjane@example.com\t"
+            "1 Main St\tRoom B345\t\t"  # double tab is for blank address3
+            "Boston\tMassachusetts\t02115\tUSA\t"
+            "999-999-9999\t1\t2016/05/15 15:02:55"
+        )
+
+    def test_write_cdd_file_skips_invalid_state(self):  # pylint: disable=no-self-use
+        """
+        Tests write_cdd_file against a profiel with invalid state
```
noisecapella: Should be `profile1`

exams/pearson_test.py
```diff
@@ -0,0 +1,214 @@
...
+            ftp_mock = connection_mock.return_value.__enter__.return_value
+            ftp_mock.cd.assert_called_once_with(EXAMS_SFTP_UPLOAD_DIR)
+            ftp_mock.put.assert_called_once_with(FILENAME)
+
+    def test_upload_tsv_raise_improperly_configured(self):  # pylint: disable=no-self-use
+        """
+        Tests that upload raises ImproperlyConfigured if setting is missing
+        """
+
+        with self.settings(
```
noisecapella: Can you break these into separate tests, to make it easier to identify if a particular case failed?

exams/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+        valid_profiles, invalid_profiles = write_cdd_file(exam_profiles, tsv)
+
+        # flush data to disk before upload
+        tsv.flush()
+
+        # upload to SFTP server
+        upload_tsv(tsv.name)
+
+    # update records to reflect the successful upload
+    if valid_profiles:
```
noisecapella: Can you wrap these two blocks in a `transaction.atomic`?

exams/tasks_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ExamSignalsTest(TestCase):
+    """
+    Tests for exam tasks
+    """
+
+    @patch('exams.pearson.upload_tsv')
+    @patch('exams.pearson.write_cdd_file')
+    @override_settings(CELERY_ALWAYS_EAGER=True)
```
noisecapella: This shouldn't be necessary, we set this to `True` in `tox.ini`

noisecapella:
Can you change all instances of `ccd` in this PR to `cdd`? If that's how it's supposed to be spelled, I forget

exams/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+@async.task
+def export_exam_profiles():
+    """
+    Sync any outstanding profiles
+    """
+    exam_profiles = (ExamProfile.objects
+                     .filter(status=ExamProfile.PROFILE_PENDING)
+                     .select_related('profile'))
+    file_prefix = datetime.now(pytz.utc).strftime('cdd-%Y%m%d%H_')
+    valid_profiles, invalid_profiles = [], []
```
noisecapella: It looks like these two variables are unused, they're redefined below

noisecapella:
Looks great overall, testing is pretty thorough. Just a few questions and I need to do a manual test

exams/pearson.py
rhysyngsun: They expect GMT so UTC is sufficient.

exams/tasks_test.py
rhysyngsun: Ok, I saw it in a few other tests, so those probably need to be cleaned up too.

rhysyngsun: Done.

exams/pearson_test.py
rhysyngsun: Done.

exams/tasks.py
rhysyngsun: Done

exams/tasks.py
rhysyngsun: Done

rhysyngsun:
@noisecapella 

Right now the only way to test is to execute the task manually and then pull the file from the sftp docker container. I usually do something like this (I'm not quite sure why I have to force the sequence, I suspect it has to do with our seed data but I haven't dived deep on it):

    from exams.factories import ExamProfileFactory
    from exams.tasks import export_exam_profiles
    OFFSET=1000
    with mute_signals(post_save):
        for x in range(50:
            ExamProfileFactory.create(profile__user____sequence=OFFSET+x)
    export_exam_profiles()

You can then sftp to the sftp service and pull the file down:

    > sftp -P 2022 odl@localhost:share
    > ls 
    > get cdd-<suffix>.dat

That will pull the file down to your working directory and you can check it locally.

exams/tasks_test.py
```diff
@@ -0,0 +1,54 @@
...
+        with mute_signals(post_save):
+            exam_profiles = ExamProfileFactory.create_batch(10, status=ExamProfile.PROFILE_PENDING)
+
+        valid, invalid = exam_profiles[:5], exam_profiles[5:]
+
+        ccd_writer_mock.return_value = (valid, invalid)
+
+        export_exam_profiles()
+
+        assert upload_tsv_mock.call_count == 1
```
rhysyngsun: Done

exams/pearson.py
```diff
@@ -0,0 +1,162 @@
...
+    ('Country', _profile_country_to_alpha3),
+    ('Phone', 'phone_number'),
+    ('PhoneCountryCode', 'phone_country_code'),
+    ('LastUpdate', lambda profile: _format_datetime(profile.updated_on)),
+], field_prefix='profile')
+
+
+def upload_tsv(file_path):
+    """
+    Upload the given TSV files to the remote
```
noisecapella: Can you add args and a return value (if any) to this docstring?

exams/pearson_test.py
```diff
@@ -0,0 +1,234 @@
...
+                username=EXAMS_SFTP_USERNAME,
+                password=EXAMS_SFTP_PASSWORD,
+                cnopts=mock.ANY,
+            )
+
+            ftp_mock = connection_mock.return_value.__enter__.return_value
+            ftp_mock.cd.assert_called_once_with(EXAMS_SFTP_UPLOAD_DIR)
+            ftp_mock.put.assert_called_once_with(FILENAME)
+
+    def test_upload_tsv_fails_if_sftp_host_not_set(self):  # pylint: disable=no-self-use
```
noisecapella: Sorry to be nitpicky about this. Can we use `ddt` here to make this a parameterized test?

exams/tasks_test.py
noisecapella: Thanks, I'll make an issue to clean up the ones in `financialaid/tasks_test.py`

exams/pearson_test.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: Done

exams/pearson.py
```diff
@@ -0,0 +1,169 @@
...
+
+    Args:
+        file_path (str): absolute path to the file to be uploaded
+
+    Return
```
noisecapella: Did you mean to remove this?

rhysyngsun: Good catch.

PR #2063, opened 2016-12-09
Updated app to proxy requests through Nginx

blarghmatey:
#1971 

#### What's this PR do?
Puts Nginx in front of uWSGI for both local dev and in Heroku

config/nginx.conf, config/nginx.conf.erb, docker-compose.yml

Run docker-compose and visit the app at http://localhost:8080 and make sure that all app functions are working as expected.

Using Nginx in front of uWSGI will allow us to properly handle chunked-content encoding without having to resort to arbitrarily long socket timeouts to accomodate users with high latency and/or unreliable network connections.

http://giphy.com/gifs/sinking-this-is-fine-everything-l0HlMG1EX2H38cZeE

Proxying apps through Nginx solves the chunked-content issues that we
have been seeing with the profile image uploads. This commit updates
the docker-compose and procfiles to use Nginx in front of uWSGI for
serving all routes.


config/nginx.conf
```diff
@@ -0,0 +1,31 @@
+# This is the version used in development environments
+server {
+    listen 8080;
```
noisecapella: Is it possible to change this to `8079` and then change the port for `uwsgi` to something else? It would be nice to keep the same port number so all the passwords and history in our browsers can stay the same

config/nginx.conf
```diff
@@ -0,0 +1,31 @@
...
+    root /src;
+
+    if ($host = 'micromasters.herokuapp.com') {
```
noisecapella: Do we need to do this in development environments?

noisecapella:
Should we run tests through nginx as well?

noisecapella:
Nevermind, forgot that our tests all execute with a test client

app.json
```diff
@@ -13,6 +13,9 @@
     },
     {
       "url": "https://github.com/heroku/heroku-buildpack-pgbouncer"
+    },
+    {
+      "url": "https://github.com/heroku/heroku-buildpack-nginx"
```
giocalitri: is this the only thing necessary to run NGINX on heroku?
Is the `config/nginx.conf` common to both heroku and docker

blarghmatey:
The current approach being used is to set the socket timeout in uWSGI to be long enough for the majority of network conditions to succeed in uploading data using chunked content encoding. Refer to #1971 for details about the conditions that necessitate this approach.

The timeout value at the time of this message is 60 seconds, which means that if at least one byte of data is uploaded within a 60 second window then the file upload will continue until it completes. If the network conditions are such that it is not possible to send even a single byte then the read operation on the uWSGI socket will time out and the connection will fail, causing Django to register an `UnreadablePostError`.

By putting Nginx as a proxy layer in front of uWSGI, Nginx will buffer the chunked-content transfer to a temporary file until the entirety of the data transfer has completed, at which point Nginx will pass that temporary file to the uWSGI layer, eliminating the need for a long socket timeout. The reason that this is preferable to the current approach is that with long socket timeouts the uWSGI thread that is handling the upload will be blocked from handling any other requests which makes us susceptible to denial of service attacks. Nginx will also allow us to more efficiently serve static assets when they are not cached by Cloudfront.

An alternative option is to move the application to AWS and off of Heroku, but this will take more time to establish enough confidence in the stability and scalability of the deployment. I do think that this is a worthwhile direction for us to move in for reasons of better control over the environment, as well as the option to have more options in terms of the sizing of the individual instances.

Regardless of whether we start using Nginx on Heroku or skip that step and start implementing the application on AWS and EC2 then we will still need to establish an appropriate configuration for the Nginx vhost as well as performing the same testing to determine that all of the routes are being handled properly so I think that it makes more sense to get this running on Heroku first and then move to AWS/EC2 when we are ready. This will allow us to fix the handling of chunked content encoding sooner while not preventing us from migrating to a new deployment option in the future.

pdpinch:
static assets are loading in the django admin

screenshot 2016-12-13 12 40 22

pdpinch:
static assets are working. 

The Cybersource callback appears to be failing. I'll need to talk to @noisecapella to see how to debug that.

config/nginx.conf
```diff
@@ -0,0 +1,27 @@
...
+    root /src;
+
+    location / {
+        include uwsgi_params;
+        uwsgi_pass web:8077;
+        uwsgi_pass_request_headers on;
+        uwsgi_pass_request_body on;
+    }
+
+    location @webpack {
```
rhysyngsun: I haven't used a named location before, but based on a quick read of the docs, it looks to me like this is not needed since it's not ref'd anywhere else (unless I missed something) and it's not used for routing otherwise.

rhysyngsun:
:+1:

PR #2069, opened 2016-12-09
Replace Object.assign with spread syntax

### LINKED ISSUE #2041: Replace Object.assign with {...spread} syntax where it makes sense
Basically

    Object.assign({}, object, {newKey: "newValue"})

should become

    {...object, newKey: "newValue"}

noisecapella:
Fixes #2041 

#### What's this PR do?
Replaces `Object.assign` with ES6 spread syntax

Make sure nothing broke

There are minor changes here and there other than swapping `Object.assign`, for example I removed some clones which seemed redundant. There shouldn't be anything that should cause a change in behavior though.

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -104,7 +104,10 @@ export default class PersonalTab extends React.Component {
     const { profile, updateProfile } = this.props;
     let editState = _.cloneDeep(profile);
     let { payload: { profile: { image } }} = fetchResponse;
-    let newEditState = Object.assign({}, editState, { image });
+    let newEditState = {
+      ...editState,
+      image
+    };
     updateProfile(newEditState, personalTabValidator);
   };
 
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -104,10 +104,7 @@ export default class PersonalTab extends React.Component {
     const { profile, updateProfile } = this.props;
     let editState = _.cloneDeep(profile);
     let { payload: { profile: { image } }} = fetchResponse;
-    let newEditState = {
-      ...editState,
-      image
-    };
+    let newEditState = { ...editState, image };
     updateProfile(newEditState, personalTabValidator);
   };
 
--- a/static/js/components/ProgramSelector_test.js
+++ b/static/js/components/ProgramSelector_test.js
@@ -74,8 +74,9 @@ describe('ProgramSelector', () => {
   });
 
   it("does not render the 'Enroll in a new program' option if there is not at least one available program", () => {
-    let allEnrollments = programs.map(program => Object.assign({}, program, {
-      enrolled: true
+    let allEnrollments = programs.map(program => ({
+      ...program,
+      enrolled: true,
     }));
     let wrapper = renderProgramSelector({
       programs: allEnrollments
--- a/static/js/components/ProgramSelector_test.js
+++ b/static/js/components/ProgramSelector_test.js
@@ -75,8 +75,7 @@ describe('ProgramSelector', () => {
 
   it("does not render the 'Enroll in a new program' option if there is not at least one available program", () => {
     let allEnrollments = programs.map(program => ({
-      ...program,
-      enrolled: true,
+      ...program, enrolled: true
     }));
     let wrapper = renderProgramSelector({
       programs: allEnrollments
--- a/static/js/components/inputs/inputs_test.js
+++ b/static/js/components/inputs/inputs_test.js
@@ -42,12 +42,14 @@ describe('Profile inputs', () => {
 
     let renderGenderSelectField = (props = {}) => {
       return renderTestSelectField(
-        Object.assign({}, inputProps, {
+        {
+          ...inputProps,
           keySet: ['gender'],
           label: "Gender",
           options: genderOptions,
           updateProfile: updateProfileStub,
-        }, props)
+          ...props,
+        },
       );
     };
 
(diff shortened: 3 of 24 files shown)


aliceriot:
I don't think anything is broken! haha. taking a look at the code now.

static/js/components/PersonalTab.js
```diff
@@ -104,7 +104,10 @@ export default class PersonalTab extends React.Component {
     const { profile, updateProfile } = this.props;
     let editState = _.cloneDeep(profile);
     let { payload: { profile: { image } }} = fetchResponse;
-    let newEditState = Object.assign({}, editState, { image });
+    let newEditState = {
+      ...editState,
+      image
+    };
```
aliceriot: why not leave this as one line? I'm fine either way, I have a slight preference for leaving small object literals on one line though.

static/js/components/ProgramSelector_test.js
```diff
@@ -74,8 +74,9 @@ describe('ProgramSelector', () => {
   });
 
   it("does not render the 'Enroll in a new program' option if there is not at least one available program", () => {
-    let allEnrollments = programs.map(program => Object.assign({}, program, {
-      enrolled: true
+    let allEnrollments = programs.map(program => ({
+      ...program,
+      enrolled: true,
```
aliceriot: to my eyes this would be cleaner like this:

```js
let allEnrollments = programs.map(program => ({
  ...program, enrolled: true
}));
```

static/js/components/inputs/inputs_test.js
```diff
@@ -42,12 +42,14 @@ describe('Profile inputs', () => {
 
     let renderGenderSelectField = (props = {}) => {
       return renderTestSelectField(
-        Object.assign({}, inputProps, {
+        {
```
aliceriot: could you move this curly brace up to the line above? so it's:

```js
return renderTestSelectField({
  ...contents
});
```

static/js/containers/ProfilePage_test.js
```diff
@@ -114,16 +114,15 @@ describe("ProfilePage", function() {
...
         work_history: []
-      });
+      };
       helper.profileGetStub.
         withArgs(SETTINGS.user.username).
         returns(
-          Promise.resolve(Object.assign({}, userProfile, {
-            username: SETTINGS.user.username
-          }))
+          Promise.resolve(userProfile)
```
aliceriot: maybe move this argument up to the line above? `returns(Promise.resolve(userProfile))`

static/js/containers/SettingsPage.js
```diff
@@ -23,7 +23,7 @@ class SettingsPage extends ProfileFormContainer {
 
   render() {
     const { profiles } = this.props;
-    let props = Object.assign({}, this.profileProps(profiles[SETTINGS.user.username]));
+    let props = this.profileProps(profiles[SETTINGS.user.username]);
```
aliceriot: was the `Object.assign` call here just superfluous?

static/js/containers/UserPage_test.js
```diff
@@ -114,10 +114,8 @@ describe("UserPage", function() {
       helper.profileGetStub.
         withArgs(SETTINGS.user.username).
         returns(
-        Promise.resolve(Object.assign({}, USER_PROFILE_RESPONSE, {
-          username: SETTINGS.user.username
-        }))
-      );
+          Promise.resolve(USER_PROFILE_RESPONSE)
+        );
```
aliceriot: this would look nice on one line too :)

static/js/reducers/index.js
```diff
@@ -68,7 +68,16 @@ export const profiles = (state: Profiles = INITIAL_PROFILES_STATE, action: Actio
...
+    return {
+      ...state,
+      [username]: {
+        ...{
+          profile: {}
+        },
+        ...state[username],
+        ...newProfile,
+      },
+    };
```
aliceriot: it looks here like the `clone` object declared above isn't used, can we delete those lines?

also what would you think about adjusting the formatting a little to look something like this:

```js
return {
  ...state,
  [username]: {
    ...{ profile: {} },
    ...state[username],
    ...newProfile,
  },
};
```

static/js/reducers/ui.js
```diff
@@ -120,131 +120,153 @@ export const INITIAL_UI_STATE: UIState = {
...
+      enrollSelectedProgram: action.payload,
+    };
   }
   case SET_ENROLL_DIALOG_VISIBILITY: {
-    return Object.assign({}, state, {
+    return {
+      ...state,
       enrollDialogVisibility: action.payload
-    });
+    };
```
aliceriot: I have a maybe silly idea - what if for this file we defined a function like this:

```js
const mergeState = R.curry((state, diff) => (
  { ...state, ...diff }
));
```

and then above our `switch` block we'd do something like 

```js
const mergeCurState = mergeState(state);
```

then for each case our return values would look like:

```js
return mergeCurState({ enrollDialogVisibility: action.payload });
```

just a thought! might not be more readable that way.

static/js/util/profile_history_edit.js
```diff
@@ -32,8 +32,10 @@ export function openNewEducationForm(level: string, index: number) {
     newIndex = profile['education'].length;
   }
   /* add empty education */
-  let clone = Object.assign({}, profile);
-  clone['education'] = clone['education'].concat(generateNewEducation(level));
+  let clone = {
+    ...profile,
+    education: profile.education.concat(generateNewEducation(level))
```
aliceriot: ooh, we could make this:

```js
education: [...profile.education, generateNewEducation(level)]
```

a little nicer? maybe, haha.

static/js/containers/SettingsPage.js
noisecapella: Yes, `props` is only used as `{...props}` so there's already a shallow copy going on

noisecapella:
Ok, I think that's all of them

PR #2075, opened 2016-12-12
Check for cross-domain security for Zendesk widget

### LINKED ISSUE #2074: SecurityError: Failed to read the 'contentDocument' property from 'HTMLIFrameElement': Blocked a frame with orig...
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/190401132/

```
SecurityError: Failed to read the 'contentDocument' property from 'HTMLIFrameElement': Blocked a frame with origin "https://micromasters.mit.edu" from accessing a cross-origin frame.
  at apply (./static/js/entry/zendesk_widget.js:190:23)
```

singingwolfboy:
Fixes #2074. It appears that sometimes, the Zendesk web widget loads with HTTP headers that instruct the browser to increase cross-domain security. When that happens, our Javascript is unable to modify the Zendesk web widget at all, and the browser raises an exception every time it tries. These exceptions go to Sentry, and clog up the exception logs.

This pull request mades the code more defensive, check to see if it can modify the iframe before doing so. If it can't, it doesn't even try. This should prevent these errors from going to Sentry, but it does _not_ solve the underlying problem of being unable to modify the Zendesk web widget.

### DIFF
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -188,10 +188,22 @@ const zendeskPollForExistence = (name) => {
 
 const zendeskPollForLoaded = (name) => {
   let tries = 0;
+  let iframeDocument = null;
   const intervalID = setInterval(() => {
     tries += 1;
     const iframe = document.querySelector(`iframe.zEWidget-${name}`);
-    const div = iframe.contentDocument.querySelector("div");
+    try {
+      iframeDocument = iframe.contentDocument;
+    } catch (err) {
+      // cross-domain exception: can't continue
+    }
+    if (!iframeDocument) {
+      console.error(`Can't access content of Zendesk iframe: ${name}`);  // eslint-disable-line no-console
+      clearInterval(intervalID);
+      return;
+    }
+
+    const div = iframeDocument.querySelector("div");
     if (div) {
       clearInterval(intervalID);
       const callback = zendeskCallbacks[`${name}Loaded`];


aliceriot:
Code looks fine, but I'm curious if it works on Safari - I was looking around for best-practices on this sort of thing and I found a couple of things saying that this approach could still result in an unhandled exception on Safari. Did you try it out in Safari at all?

singingwolfboy:
@aliceriot Thanks for checking that! Yes, it looks like I do still get an exception in Safari:
safari-exception

Not sure the best way to avoid this, though. Can you show me where you found out about the Safari issue? Do they have an recommendations for how to avoid this error message?

aliceriot:
SO: https://stackoverflow.com/questions/7417835/javascript-check-if-iframe-url-is-same-domain-or-something-external

aliceriot:
Is it just a warning, or a full exception?

singingwolfboy:
It's a full exception. I don't see anything in that StackOverflow question that will help us in this situation -- all the answers are about comparing URLs, rather than determining CORS headers. I think that this is the best solution we're going to get right now, especially considering that this is meant only to stop the flood of errors, rather than actually fix the problem.

Are you OK with merging this as an improvement, even if it's an imperfect one?

aliceriot:
Sure, but if we do that we should also file an issue to come back and implement a full fix.

pdpinch:
@singingwolfboy you can merge this and then please open an issue for Safari as Alice suggested. 

Mostly likely we'll solve this in a more general way before we fix the Safari-specific issue, but I don't want to lose track of the issue.

singingwolfboy:
I've created #2087 to make this fix work properly for Safari.

PR #2078, opened 2016-12-12
limit the number of options displayed for country of birth

aliceriot:
## testing

here's a quick way to set a separate birth country for each user.

first, save the following as `codes.js`:

```js
var fs = require('fs')
var iso = require('iso-3166-2')

let codes = Object.values(iso.codes)

console.log(codes.join("\n"))
```

then, save the following as `codes.py`:

```py

with (open("codes.txt")) as f:
    codes = f.read().splitlines()

for user in User.objects.all():
    user.profile.birth_country = codes[user.id % len(codes)]
    user.profile.save()
```

Then do:

```sh
node codes.js > codes.txt
```

open the django shell and type:

```py
exec(open("codes.py").read())
```

that should do it!

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -165,6 +165,7 @@ export default class LearnerSearch extends SearchkitComponent {
             field="profile.birth_country"
             operator="OR"
             itemComponent={CountryRefinementOption}
+            size={15}
           />
         </FilterVisibilityToggle>
         <FilterVisibilityToggle


singingwolfboy:
Quick note: here's a more "Pythonic" way to iterate over both users and country codes:

```python
from itertools import cycle

def read_codes():
    for line in open("codes.txt"):
        yield line.strip()

users = User.objects.all()
codes = read_codes()

for user, code in zip(users, cycle(codes)):
    user.profile.birth_country = code
    user.profile.save()
```

The [`itertools`](https://docs.python.org/3/library/itertools.html) module is pretty fantastic.

aliceriot:
thanks for the hint!

aliceriot:
closing because this ran into an issue with Searchkit, we won't be able to fix it until they fix the bug.

PR #2083, opened 2016-12-12
Made twitter description tag shorter

### LINKED ISSUE #1863: Tweet is too long
It looks to me like we're using the meta tag description for the sample tweet text, and it's way too long for that. 

Maybe ask Maria for a sample tweet. 

screenshot 2016-11-26 09 46 34

annagav:
Fix #1863
#### What's this PR do?

Add twitter description tag with shorter text.

### DIFF
--- a/static/js/entry/public.js
+++ b/static/js/entry/public.js
@@ -166,8 +166,8 @@ jQuery(document).ready(function ($) {
     description: description,
     emailBody: description + CURRENT_PAGE_URL
   });
-  $('.rrssb-buttons').find('.rrssb-twitter a').attr('href',
-    'https://twitter.com/intent/tweet?text=' + twitterDescription + '%20' + CURRENT_PAGE_URL );
+  const tweetUrl = `https://twitter.com/intent/tweet?text=${twitterDescription}%20${CURRENT_PAGE_URL}`;
+  document.querySelector('.rrssb-buttons .rrssb-twitter a').href = tweetUrl;
 });
 
 /**
--- a/static/js/entry/public.js
+++ b/static/js/entry/public.js
@@ -154,6 +154,8 @@ jQuery(document).ready(function ($) {
     'integrated set of graduate-level online courses. With the MicroMasters ' +
     "credentials, learners can apply for an accelerated master's degree " +
     "program on campus, at MIT or other top universities.";
+  const twitterDescription = 'MITx MicroMasters Programs: a new academic credential ' +
+    'and a new path to a master’s degree from MIT. Learn more ';
 
   $('.rrssb-buttons').rrssb({
     // required:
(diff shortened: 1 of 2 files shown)


singingwolfboy:
This is not a bad change, but it doesn't actually fix the problem. The tweet text is generated by [RRSSB](https://github.com/kni-labs/rrssb/), which is configured here: https://github.com/mitodl/micromasters/blob/master/static/js/entry/public.js#L151-L156

annagav:
@singingwolfboy It should be fixed now.

static/js/entry/public.js
```diff
@@ -164,6 +166,8 @@ jQuery(document).ready(function ($) {
     description: description,
     emailBody: description + CURRENT_PAGE_URL
   });
+  $('.rrssb-buttons').find('.rrssb-twitter a').attr('href',
```
singingwolfboy: Why not do:
```js
const tweetUrl = `https://twitter.com/intent/tweet?text=${twitterDescription}%20${CURRENT_PAGE_URL}`;
document.querySelector('.rrssb-buttons .rrssb-twitter a').href = tweetUrl;
```

Seems like we're trying to move away from jQuery where possible, and this is nicer ES6 syntax.

static/js/entry/public.js
```diff
@@ -154,6 +154,8 @@ jQuery(document).ready(function ($) {
     'integrated set of graduate-level online courses. With the MicroMasters ' +
     "credentials, learners can apply for an accelerated master's degree " +
     "program on campus, at MIT or other top universities.";
+  const twitterDescription = 'MITx MicroMasters Programs: a new academic credential ' +
```
singingwolfboy: Can you change the `var description` above to `const description`, so that they both use `const`?

static/js/entry/public.js
```diff
@@ -154,6 +154,8 @@ jQuery(document).ready(function ($) {
...
+    'and a new path to a master’s degree from MIT. Learn more ';
```
singingwolfboy: Should this end with a colon? `Learn more: `

static/js/entry/public.js
aliceriot: That should be changed on master already.

static/js/entry/public.js
annagav: sounds good.

static/js/entry/public.js
annagav: This is a question to @roberthouse54

roberthouse54:
Can we get an accurate character count on that link?

pdpinch:
I don't think we should count characters based on that dev URL. If I use a production URL, it seems short enough:

screenshot-twitter com-2016-12-14-08-15-56

The two possibilities for now are https://micromasters.mit.edu/dedp and https://micromasters.mit.edu/scm

PR #2100, opened 2016-12-13
only run the getCroppedCanvas method if this.refs.cropper is defined

### LINKED ISSUE #2096: TypeError: Cannot read property 'getCroppedCanvas' of undefined
Alice, can you take a quick look at this to triage it? Is it serious enough to fix? Is it obvious enough that we think we'll be successful?

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/192140142/

```
TypeError: Cannot read property 'getCroppedCanvas' of undefined
  at apply (./static/js/components/CropperWrapper.js:19:33)
...
(1 additional frame(s) were not displayed)
```

aliceriot:
closes #2096

This fixes an error that I think is related to a subtle race condition with how we take down the photo upload dialog. We pass a callback (`ready`) to the `<Cropper />` component, which is something that fires when the image finishes loading. We do this so that as soon as the user's picture is loaded they can click 'save' in order to save it (without that the user would need to make at least one edit to the picture in order to save it). I think what is happening is that after the upload request is successfully completed, we dispatch the `CLEAR_PHOTO_EDIT` action, which forces a redraw of the `<Cropper />` element (because the `photo` prop passed in to it changes). Then the `ready` callback will be called, but I think what's happening is that the dialog closes at the same time, so that, in the function, `this.refs.cropper` is no longer present.

All this PR does to fix this is just check if `this.refs.cropper` is present before we try to call any methods on it.

To test this just ensure that you can upload a photo, and that there are no exceptions after the photo upload has finished. The bug this fixes should repro on master right now, where basically after the upload is finished you'll see three errors like this:

errors_canvas

### DIFF
--- a/static/js/components/CropperWrapper.js
+++ b/static/js/components/CropperWrapper.js
@@ -13,22 +13,24 @@ export default class CropperWrapper extends React.Component {
   cropperHelper = () => {
     const { updatePhotoEdit } = this.props;
     let canvas;
-    if ( browser.name === 'safari' || browser.name === 'ios' ) {
-      canvas = this.refs.cropper.getCroppedCanvas();
-    } else {
-      canvas = this.refs.cropper.getCroppedCanvas({
-        width: 512,
-        height: 512,
-      });
+    if ( this.cropper ) {
+      if ( browser.name === 'safari' || browser.name === 'ios' ) {
+        canvas = this.cropper.getCroppedCanvas();
+      } else {
+        canvas = this.cropper.getCroppedCanvas({
+          width: 512,
+          height: 512,
+        });
+      }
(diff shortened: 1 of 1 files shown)


noisecapella:
Not sure if you want to do this now or in another PR, but it looks like the new way of doing refs is to pass it a callback that sets a field on the React object: https://facebook.github.io/react/docs/refs-and-the-dom.html

aliceriot:
@noisecapella just updated it to use the newer style of setting `refs`, lmk what you think.

aliceriot:
(it's pretty straightforward, haha)

noisecapella:
:+1:

Issue #2038, opened 2016-12-08
Search: decrease the default number of facet hits for "Country of Birth"


pdpinch:
The learner search is getting crazy for DEDP. We need to default to seeing fewer facets and give the user an option to "view more" (ask @roberthouse54 how many is a good number)

Not like this:

<img width="177" alt="screenshot 2016-12-08 11 00 42" src="https://cloud.githubusercontent.com/assets/430126/21016978/b6da7b04-bd35-11e6-8da0-65eff59acd61.png">

More like this (see 
http://demo.searchkit.co/imdb):

<img width="265" alt="screenshot 2016-12-08 10 58 23" src="https://cloud.githubusercontent.com/assets/430126/21016910/6f33ccf6-bd35-11e6-9ba0-6a515f676469.png">

roberthouse54:
Let's try 15

aliceriot:
@pdpinch is the 'Country of Birth' field the only one we need to adjust?

pdpinch:
With the current set of facets, both 'Country of Birth' and 'Current Residence' have this problem. 

Since they are different types of facets, I don't know if the same solution will work for both. 

@roberthouse54

aliceriot:
It's very simple to limit the number of entries shown for 'Country of Birth' but unfortunately it doesn't look like 'Current Residence' has a built in mechanism for doing that (supplied by searchkit, I mean).

aliceriot:
An easy option might be using something like this instead: https://github.com/GregoryPotdevin/searchkit-multiselect

aliceriot:
For the `Country of Birth` facet (which is a `<RefinementListFilter />`) it is really trivial to limit the number of options shown - we can just pass a `size` prop which automatically takes care of showing and hiding the excess entries.

The difficulty is the `<HierarchicalMenuFilter />` component, which we use to implement the `Current Residence` filter. This does not have a built-in mechanism for limiting the number of entries initially shown. I have tried a few hacks but haven't gotten anything to work yet.

I've also asked the searchkit developers if they have any insight, but haven't heard anything yet.

pdpinch:
That's frustrating. 

Let's fix the easy case for now, while we decide what to do about the harder case.

aliceriot:
Sure

pdpinch:
Closing this issue and opening a new one for just "Current Residence" facet.

PR #2102, opened 2016-12-13
Limited the birth country facet to 15 options

aliceriot:
this is part of #2038, just taking care of the easier case. we'll come back for the 'current residence' facet later.

expander

expanded

copying what I wrote about test data on #2078:

first, save the following as `codes.js`:

```js
var fs = require('fs')
var iso = require('iso-3166-2')

let codes = Object.values(iso.codes)

console.log(codes.join("\n"))
```

then, save the following as `codes.py`:

```py

with (open("codes.txt")) as f:
    codes = f.read().splitlines()

for user in User.objects.all():
    user.profile.birth_country = codes[user.id % len(codes)]
    user.profile.save()
```

Then do:

```sh
node codes.js > codes.txt
```

open the django shell and type:

```py
exec(open("codes.py").read())
```

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -165,6 +165,7 @@ export default class LearnerSearch extends SearchkitComponent {
             field="profile.birth_country"
             operator="OR"
             itemComponent={CountryRefinementOption}
+            size={15}
           />
         </FilterVisibilityToggle>
         <FilterVisibilityToggle


noisecapella:
FYI `import codes` will do the same thing as the `exec(...)` code above
:+1:

aliceriot:
ok cool, I think I tried that but couldn't get it to work somehow. django still confuses me, haha.

PR #2103, opened 2016-12-13
Added ECOMMERCE_EMAIL setting, added decision to email subject

### LINKED ISSUE #2093: Define new setting for ecommerce emails
Emails currently go to `settings.EMAIL_SUPPORT`. We need another separate setting so that we can send ecommerce emails to a separate email address

### LINKED ISSUE #2094: Put Cybersource decision status in email subject for easier filtering
We currently have 'Order fulfillment failed' as the subject but we should have something like 'Order fulfillment failed, decision=CANCEL' to allow easier filtering

noisecapella:
Fixes #2093 
Fixes #2094 

#### What's this PR do?
- Adds a setting for ecommerce emails. It's populated with the environment variable `MICROMASTERS_ECOMMERCE_EMAIL`
- Changes email sent from 'Order fulfillment failed' to 'Order fulfillment failed, decision=CANCEL', for example

I'll set up the PR build for this. When testing click the 'Pay now' button, then click the cancel link on Cybersource. An email should be sent to the ecommerce email we defined, which is different from the support email. It should have 'Order fulfillment failed, decision=CANCEL' in the subject


rhysyngsun:
:+1:

PR #2107, opened 2016-12-13
Replace utcnow() with now(tz=pytz.UTC)

### LINKED ISSUE #1960: Database models have fields that aren't timezone-specific
We're still getting a load of django timezone warnings, particularly if we run something that heavily involves the database, eg: the `seed_db` command. A lot of these warnings are likely due to a misunderstanding of python `datetime`. `TimestampedModel` is inherited by a lot of our models, and it uses the following as its queryset:

```python
class TimestampedModelQuerySet(QuerySet):
    """
    Subclassed QuerySet for TimestampedModelManager
    """
    def update(self, **kwargs):
        """
        Automatically update updated_on timestamp when .update(). This is because .update()
        does not go through .save(), thus will not auto_now, because it happens on the
        database level without loading objects into memory.
        """
        if "updated_on" not in kwargs:
            kwargs["updated_on"] = datetime.datetime.utcnow()
        return super().update(**kwargs)
```

`datetime.datetime.utcnow()` does NOT yield a timezone-sensitive datetime. This should be changed to `datetime.datetime.now(pytz.utc)`, and all other examples in the codebase should be changed accordingly. This can be considered done when we can run `seed_db` with no timezone warnings

noisecapella:
Fixes #1960

#### What's this PR do?
Fixes datetime handling to use a specific timezone. It doesn't look like this was causing problems for us since `timestamp()` returns the same value for a naive datetime and one set to `UTC`, but it may prevent problems in the future if we're explicit.

There shouldn't be any change of behavior

### DIFF
--- a/micromasters/models.py
+++ b/micromasters/models.py
@@ -14,6 +14,7 @@ from django.db.models import (
     SET_NULL,
 )
 from django.db.models.query import QuerySet
+import pytz
 
 
 class TimestampedModelQuerySet(QuerySet):
@@ -27,7 +28,7 @@ class TimestampedModelQuerySet(QuerySet):
         database level without loading objects into memory.
         """
         if "updated_on" not in kwargs:
-            kwargs["updated_on"] = datetime.datetime.utcnow()
+            kwargs["updated_on"] = datetime.datetime.now(tz=pytz.UTC)
         return super().update(**kwargs)
 
 
--- a/backends/edxorg.py
+++ b/backends/edxorg.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from urllib.parse import urljoin
 
 from django.conf import settings
+import pytz
 from social.backends.oauth import BaseOAuth2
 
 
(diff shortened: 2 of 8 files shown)


micromasters/models.py
```diff
@@ -27,7 +28,7 @@ def update(self, **kwargs):
         database level without loading objects into memory.
         """
         if "updated_on" not in kwargs:
-            kwargs["updated_on"] = datetime.datetime.utcnow()
+            kwargs["updated_on"] = datetime.datetime.now(tz=pytz.UTC)
```
giocalitri: ok for the `updated_on`, but what about the `created_on`?

giocalitri:
just a comment

micromasters/models.py
noisecapella: That's populated when the user creates the object in question. The database will put a UTC time zone on the date because of the `USE_TZ` setting so we don't need to worry about it there

giocalitri: sorry: where is the `USE_TZ` set?

noisecapella: In [settings.py](https://github.com/mitodl/micromasters/blob/master/micromasters/settings.py#L290)

giocalitri:
functionality seems fine 👍

PR #2118, opened 2016-12-14
Made function for dialog actions

### LINKED ISSUE #2061: Add abstraction for dialog buttons which are frequently reused
From https://github.com/mitodl/micromasters/pull/2042 we should make a function to create the dialog buttons to avoid repetition, something like:

    const profileDialogActions = (onCancel, onSave, inFlight) => ([
      <Button ... />,
      <Button ... />,
    ]);

noisecapella:
Fixes #2061

#### What's this PR do?
Adds function for the dialog buttons to reduce code duplication

Make sure the updated dialog buttons still look and act the same as before

### DIFF
--- a/static/js/components/inputs/util_test.js
+++ b/static/js/components/inputs/util_test.js
@@ -1,0 +1,29 @@
+// @flow
+import { assert } from 'chai';
+
+import { dialogActions } from './util';
+
+describe('input functions', () => {
+
+  describe('dialogActions', () => {
+    it('renders properly', () => {
+      let onCancel = () => null;
+      let onSave = () => null;
+      let inFlight = true;
+      let newClass = "new-class";
+      let newText = "New Text";
+      let actions = dialogActions(onCancel, onSave, inFlight, newText, newClass);
+      assert.lengthOf(actions, 2);
+      let [cancelButton, saveButton] = actions;
+      assert.equal(cancelButton.props.className, "secondary-button cancel-button");
+      assert.equal(cancelButton.props.onClick, onCancel);
+      assert.equal(cancelButton.props.children, "Cancel");
+
+      assert.equal(saveButton.props.className, `primary-button save-button ${newClass}`);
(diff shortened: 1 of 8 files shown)


gsidebo:
initial nitpick: can we find a better place for this function than `static/js/util/util.js`? i know we already have non-utility functions in files called `util.js`, but i think we can do better than the top level general util file. maybe a new file - `static/js/components/inputs/util.js` or `static/js/components/inputs/common.js`?

noisecapella:
sounds good

static/js/components/inputs/util_test.js
```diff
@@ -0,0 +1,29 @@
...
+      let [cancelButton, saveButton] = actions;
+      assert.equal(cancelButton.props.className, "secondary-button cancel-button");
+      assert.equal(cancelButton.props.onClick, onCancel);
+      assert.equal(cancelButton.props.children, "Cancel");
+
+      assert.equal(saveButton.props.className, `primary-button save-button ${newClass}`);
+      assert.equal(saveButton.props.onClick, onSave);
+      assert.equal(saveButton.props.children, newText);
+      assert.equal(saveButton.props.spinning, inFlight);
+      assert.equal(saveButton.type.name, "SpinnerButton");
```
gsidebo: do we get any value from asserting the name of the component here? i think the asserting of `spinning` is enough of an indicator that this is supposed to be a button that 'spins', and i don't know if i like the precedent of asserting component class names in general

noisecapella: Asserting the class name is important because it connects this unit test with the one for the `SpinnerButton`. If we swap out `SpinnerButton` with another similar component it may not have exactly the same behavior, or it may not be tested properly. This is kind of a trivial case admittedly but it seems like asserting the `spinning` prop without asserting the component we're passing the prop to would be like asserting arguments being passed to an `add` function without asserting that it is the `add` function we wanted to call

gsidebo:
looks good 👍

PR #2119, opened 2016-12-14
Added add_past_passed_run command in alter_data

### LINKED ISSUE #2044: Enable adding a passed course run in the past in alter_data commands
The 'alter data' commands (`seed_data/management/commands/alter_data.py`) currently allow users to add a past failed run (`--action=add_past_failed_run`). We need to expand this to allow a passed run in the past as well.

annagav:
Fix #2044
#### What's this PR do?
Added a management command that allows the user to make a past passed course run for a user (in a non FA program)

Enroll in the Analog Learning program.
Run this command and check if you can see the passed course in the dashboard:

`alter_data --action=add_past_passed_run --username=staff --program-title='Analog' --course-title='100' --add-if-exists`

### DIFF
--- a/seed_data/management/commands/alter_data.py
+++ b/seed_data/management/commands/alter_data.py
@@ -81,6 +81,13 @@ USAGE_DETAILS = (
     "--add-if-exists"
     """
 
+    # Add a past passed course run with a specific grade
+    """
+    "alter_data --action=add_past_passed_run --username=staff --program-title='Analog' --course-title='100'"
+    "--add-if-exists"
+    """
+
+
     # Add an enrollable future course run for a program called 'Test Program' and a course called 'Test Course'
     alter_data --action=add_future_run --username=staff --program-title='Test Program' --course-title='Test Course'
     """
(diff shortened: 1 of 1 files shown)


seed_data/management/commands/alter_data.py
```diff
@@ -178,26 +182,58 @@ def clear_future_runs(user=None, course=None):  # pylint: disable=unused-argumen
...
+@accepts_or_calculates_now
+def add_past_passed_run(user=None, course=None, now=None, grade=DEFAULT_GRADE, add_if_exists=False):
+    """Adds a past passed course run for a given user and course"""
+    chosen_past_run = get_past_course_run(user, course, now, add_if_exists)
+    price = "150"
+
+    # Add enrollment and passing grade for course run
     CachedEnrollmentHandler(user).set_or_create(course_run=chosen_past_run)
     CachedCurrentGradeHandler(user).set_or_create(course_run=chosen_past_run, grade=grade)
+    CachedCertificateHandler(user).set_or_create(course_run=chosen_past_run, grade=grade)
```
gsidebo: these operations are only made to work with non-financial-aid programs. that's _really_ poorly documented - sorry about that. as a result, line 222 and lines 224-236 don't need to exist. passed courses in non-FA programs only need a CachedCertificate with a passing grade. they don't need a CachedCertificate, Order or Line

gsidebo:
aside from the [one comment](https://github.com/mitodl/micromasters/pull/2119/files#r92442217), this is pretty much right on

seed_data/management/commands/alter_data.py
annagav: Without the Order and the Line the passed course is being displayed in the Dashboard as a failed course.

seed_data/management/commands/alter_data.py
```diff
@@ -178,26 +185,43 @@ def clear_future_runs(user=None, course=None):  # pylint: disable=unused-argumen
...
                 past_run.title
             ))
     if not chosen_past_run:
         raise Exception("Can't find past run w/o CachedCurrentGrade")
-    # Add enrollment and failed grade for course run
+    return chosen_past_run
+
+
+@accepts_or_calculates_now
+def add_past_passed_run(user=None, course=None, now=None, grade=DEFAULT_GRADE, add_if_exists=False):
```
gsidebo: tiny nitpick: both `add_past_failed_run` and `add_past_passed_run` depend on `get_past_course_run`. i think it makes sense to move the definition of `get_past_course_run` above the definition of `add_past_failed_run`

gsidebo:
testing functionality now

seed_data/management/commands/alter_data.py
annagav: sounds good

seed_data/management/commands/alter_data.py
```diff
@@ -178,26 +185,43 @@ def clear_future_runs(user=None, course=None):  # pylint: disable=unused-argumen
...
     """Adds a past failed course run for a given user and course"""
+    chosen_past_run = get_past_course_run(user, course, now, add_if_exists)
+
+    # Add enrollment and failed grade for course run
+    CachedEnrollmentHandler(user).set_or_create(course_run=chosen_past_run)
+    CachedCurrentGradeHandler(user).set_or_create(course_run=chosen_past_run, grade=grade)
+    return chosen_past_run
+
+
+def get_past_course_run(user=None, course=None, now=None, add_if_exists=False):
```
gsidebo: one other thing i just noticed: this function should be checking for both current grade and certificates. here's a snippet that includes that and some minor cleanup:

```python 
def get_past_course_run(user=None, course=None, now=None, add_if_exists=False):
    """Loop through past course runs and find one without grade data"""
    past_runs = CourseRun.objects.filter(
        course=course,
        end_date__lt=now,
    ).exclude(end_date=None).order_by('-end_date').all()
    for past_run in past_runs:
        if not CachedCurrentGradeHandler(user).exists(past_run) and \
                not CachedCertificateHandler(user).exists(past_run):
            return past_run
        elif not add_if_exists:
            raise Exception("Past failed/passed course run already exists (id: {}, title: {})".format(
                past_run.id,
                past_run.title
            ))
    else:
        raise Exception("Can't find past run w/o grade data")
```

annagav:
@gsidebo I addressed all comments.

seed_data/management/commands/alter_data.py
```diff
@@ -175,32 +182,50 @@ def clear_future_runs(user=None, course=None):  # pylint: disable=unused-argumen
...
     chosen_past_run = None
     past_runs = CourseRun.objects.filter(
         course=course,
         end_date__lt=now,
     ).exclude(end_date=None).order_by('-end_date').all()
-    # Loop through past course runs and find one without CachedCurrentGrade data
     for past_run in past_runs:
-        if not CachedCurrentGradeHandler(user).exists(past_run):
+        if not (CachedCurrentGradeHandler(user).exists(past_run)
+                or CachedCertificateHandler(user).exists(past_run)):
```
gsidebo: in the [code sample i posted](https://github.com/mitodl/micromasters/pull/2119#discussion_r92504826), i fixed a minor bug in this loop. this is incorrectly using `continue` at line 196. the effect is that instead of returning the most recent un-graded course run, it returns the oldest. not a big deal, but it can result in unexpected behavior. since this function doesn't need to do anything with the `chosen_past_run` anymore, we can just return it immediately as soon as we find it

gsidebo: on second thought, i can create a new issue for this. will be an easy fix and very easy to review

annagav: I don't think removing continue will do it. I think you would need to either break out of the loop one you found the right course run or order the `CourseRun`s in the positive direction `.order_by('-end_date')`, so that the last course run it checks is the most recent one.

gsidebo:
looks great! 👍

seed_data/management/commands/alter_data.py
annagav: ok, I am removing `continue` anyway because I think it is redundant.

PR #2125, opened 2016-12-14
Correct spacing for course search facet

### LINKED ISSUE #2124: Search: Fix spacing between facet options on search page
Some of the Course names are longer than we suspected, and they look confusing given that there is little difference between the line height of options, and the margin between options.

Please add a line height 1.3 to options in the course facet, and a margin-bottom of 8px under each option.

singingwolfboy:
Fixes #2124. Corrects the spacing around the course search facet to account for course names that wrap to multiple lines.
over-nine-thousand

### DIFF
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -132,6 +132,17 @@
       }
     }
 
+    .filter--courses {
+      .sk-item-list-option {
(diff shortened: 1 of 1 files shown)


roberthouse54:
👍

PR #2126, opened 2016-12-14
Final grade freezing

### LINKED ISSUE #1958: Final grades: compute final grades
Part of #1856 

Currently, code for calculating final grades is part of `mmtrack` (both for financial aid and non-financial aid programs). This should be moved to the new model written in #1957 

Final grade computation will ideally be run once, per user per courserun, on a set date (see ###). We will also need to be able to run it by management command when ad hoc updates are required.  

### Acceptance Criteria

- [ ] run only once per user per course-run (under normal circumstances)
- [ ] notify admin if calculation fails
- [ ] notify user if calculation fails (?)

### LINKED ISSUE #1959: Final Grades: add final_grade_calculation_date on course run model
Part of #1856 

We plan to have a scheduled job that calculates any final grades that can and need to be calculated. However, we need to know when it is safe to calculate the final grades. For financial aid courses, this can be the same as the last day of the course. For non-financial aid courses, certificates are typically created on a scheduled date which can be found in the MITx BMS. 

Since there is some variability in when this date may be, the safest thing to do is to add a field to the course run model for storing the date. We can manage it by hand (similar to other course dates) for now.

giocalitri:
closes #1958 
closes #1959

#### What's this PR do?
Introduces APIs, celery tasks and management commands to freeze final grades for course runs.

`grades.api`, then `grades.tasks`

to freeze a curse run the entry point is:
in the django shell import `grades.tasks.freeze_course_run_final_grades`

or the management commands `freeze_final_grades`

create cached enrollments, current_grades for all the users you want to considered for grade freeze.
create cached certificate for some of the financial aid course run users: this is needed to test the special case in which the certificate should be considered before the current grade also for financial aid courses.

There is a lot to test in this PR:
1. verify that only the courses with a freeze date in the past can be frozen: for this first try to freeze a course without a freeze date (it should raise `ImproperlyConfigured`) then a course with the freeze date in the future (nothing should happen).
2. verify that all the enrolled users have a final grade (with passed or not according to the current_grade and certificates you created)
3. verify that the course run with an entry as COMPLETE in `FinalGradeRunInfo` are not considered for freeze any more
4. any edge case you have in mind.

Be aware that the task `grades.tasks.freeze_course_run_final_grades` needs to run at least twice to complete the freeze for a course.


grades/api_test.py
```diff
@@ -0,0 +1,237 @@
...
+        api.freeze_user_final_grade(self.user, self.course_run1)
+        mock_refr.assert_called_once_with(self.user)
+        mock_get_fg.assert_called_once_with(self.user, self.course_run1)
+        assert FinalGrade.objects.filter(user=self.user, course_run=self.course_run1).exists() is False
+
+    @patch('grades.api._refresh_cache_final_grade', autospec=True)
+    def test_freeze_user_final_grade(self, mock_refr):
+        """
+        Test for freeze_user_final_grade function in case of problems with getting the final grade
+        """
```
gsidebo: kind of misleading docstring. this is just a straightforward test for the functionality of `freeze_user_final_grade`, right? "in case of problems with getting the final grade" doesn't make sense to me and can probably be left out

grades/api_test.py
```diff
@@ -0,0 +1,237 @@
...
+    @patch('grades.api.get_final_grade', autospec=True)
+    @patch('grades.api._refresh_cache_final_grade', autospec=True)
+    def test_freeze_user_final_grade_error3(self, mock_refr, mock_get_fg):
+        """
+        Test for freeze_user_final_grade function in case of problems with getting the final grade
+        """
+        mock_get_fg.side_effect = AttributeError
+        api.freeze_user_final_grade(self.user, self.course_run1)
+        mock_refr.assert_called_once_with(self.user)
+        mock_get_fg.assert_called_once_with(self.user, self.course_run1)
```
gsidebo: in this test case and the two above, it looks like we only care that no `FinalGrade`s will be created in certian error cases. i don't think we get any meaningful functionality testing from asserting that `mock_refr` or `mock_get_fg` are called

grades/models.py
```diff
@@ -79,3 +90,60 @@ def __str__(self):
...
+            course_run_grade_status.save()
+        return course_run
+
+    @classmethod
+    def create_pending_course_run(cls, course_run):
+        """
+        Creates an entry with status pending
+        """
+        course_fg_info, _ = cls.objects.get_or_create(course_run=course_run)
+        return course_fg_info
```
gsidebo: we already know this model deals with course runs, and these methods all accept a `course_run` argument. including `course_run` in the method name is kinda redundant. some renaming suggestions:

`is_course_run_complete` -> `is_complete`
`is_course_run_pending` -> `is_pending`
`complete_course_run` -> `set_to_complete` (complete is an adjective and a verb, so 'set_to' is more helpful)
`create_pending_course_run` -> `create_pending`

grades/models.py
```diff
@@ -79,3 +90,60 @@ def __str__(self):
             user=self.final_grade.user,
             course_id=self.final_grade.course_run.edx_course_key
         )
+
+
+class FinalGradeRunInfo(TimestampedModel):
```
gsidebo: i wonder if `CourseRunGradeStatus` or `CourseRunGradingStatus` would be a better name here. `Info` is really ambiguous and it doesn't help convey the purpose of this model. compare the following: `FinalGradeRunInfo. is_course_run_complete(course_run)` vs. `CourseRunGradeStatus.is_complete(course_run)`. the second one is much more helpful and readable, IMO

grades/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+    Args:
+        course_run (CourseRun): a course run model object
+
+    Returns:
+        None
+    """
+    # no need to do anything if the course run is not ready
+    if not course_run.can_freeze_grades:
+        log.info('the grades course "%s" cannot be frozen yet', course_run.edx_course_key)
+        return
```
gsidebo: same comment as above with respect to this log line. will we get anything useful from this in our logs? also, this check is repeated in `grades.api.freeze_user_final_grade` and here. is it necessary to have them both?

grades/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # if it has already completed, do not do anything
+    if FinalGradeRunInfo.is_course_run_complete(course_run):
+        log.info('Final Grades freezing for course run "%s" has already been completed', course_run.edx_course_key)
+        return
```
gsidebo: same comment. will we get anything useful from these lines in our logs?

courses/models.py
```diff
@@ -250,3 +251,22 @@ def is_unexpired(self):
...
+            raise ImproperlyConfigured('Missing freeze_grade_date')
+        return datetime.now(pytz.utc) > self.freeze_grade_date
+
+    @classmethod
+    def get_runs_to_freeze(cls, exclude_list=None):
+        """
+        Returns a queryset of all the runs that can freeze final grade according to the freeze date.
+        """
+        course_runs = cls.objects.exclude(freeze_grade_date=None).filter(freeze_grade_date__lt=datetime.now(pytz.utc))
+        if course_runs is not None:
```
giocalitri: ok

courses/models_test.py
```diff
@@ -411,3 +414,36 @@ def test_is_upgradable(self, upgrade_deadline, expected):
...
+    def test_can_freeze_grades_raises(self):
+        """Test for the can_freeze_grades property raises if not configured"""
+        course_run = self.create_run()
+        with self.assertRaises(ImproperlyConfigured):
+            course_run.can_freeze_grades  # pylint: disable=pointless-statement
+
+    def test_get_runs_to_freeze(self):
+        """Test for the get_runs_to_freeze classmethod"""
+        course_runs = []
+        for date in (-1, -2, 1, None):
```
giocalitri: ok

courses/models_test.py
```diff
@@ -411,3 +414,36 @@ def test_is_upgradable(self, upgrade_deadline, expected):
...
+            course_runs.append(self.create_run(
+                freeze_grade_date=self.from_weeks(date)
+            ))
+        runs_to_freeze = CourseRun.get_runs_to_freeze(exclude_list=[course_runs[1].pk])
+        freeze_run_ids = [course_run.pk for course_run in runs_to_freeze]
+        assert course_runs[0].pk in freeze_run_ids
+        for run in course_runs[1:]:
+            assert run.pk not in freeze_run_ids
```
giocalitri: leaving as is

dashboard/models.py
```diff
@@ -122,6 +122,13 @@ def get_edx_data(cls, user, program=None):
         """
         return cls.deserialize_edx_data(cls.data_qset(user, program=program))
 
+    @classmethod
+    def get_cached_users(cls, course_run):
+        """
+        Retrieves all the usernames with entries in the cache for a given course run
```
giocalitri: ok

grades/api.py
```diff
@@ -0,0 +1,179 @@
...
+    Args:
+        course_run (CourseRun): a course run model object
+
+    Returns:
+        function: a function to be called to compute the final grade
+    """
+    return _compute_grade_for_fa if course_run.course.program.financial_aid_availability else _compute_grade_for_non_fa
+
+
+def _refresh_cache_final_grade(user):
```
giocalitri: good point

grades/api.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Refreshes the user edX cache for certificates and current grade.
+
+    Args:
+        user (User): a django User
+
+    Returns:
+        None
```
giocalitri: I think we are putting there. I am leaving them

grades/api.py
```diff
@@ -0,0 +1,179 @@
...
+        course_run (CourseRun): a course run model object
+
+    Returns:
+        UserFinalGrade: a namedtuple of (grade, passed,) representing the final grade
+            of the user in the course run and whether she passed it
+    """
+    # pull the cached data for the user
+    user_data = CachedEdxUserData(user, course_run.course.program)
+    run_data = user_data.get_run_data(course_run.edx_course_key)
+
```
giocalitri: ok

grades/api.py
```diff
@@ -0,0 +1,179 @@
...
+    # pick the right function
+    final_grade_func = _get_compute_func(course_run)
+    return final_grade_func(run_data)
+
+
+def get_users_final_grade_freeze(course_run):
```
giocalitri: ok

grades/api.py
```diff
@@ -0,0 +1,179 @@
...
+    Returns:
+        None
+    """
+    # pylint: disable=bare-except
+    # no need to do anything if the course run is not ready
+    if not course_run.can_freeze_grades:
+        log.info(
+            'the grade for user "%s" course "%s" cannot be frozen yet',
+            user.username, course_run.edx_course_key
+        )
```
giocalitri: IMHO we should run the app with a log level at least `WARNING` and go down to `INFO` or `DEBUG` only in special cases. 
I might change it to `debug` in case we want to run with the app set at `INFO`

grades/api_test.py
```diff
@@ -0,0 +1,237 @@
...
+    def setUpTestData(cls):
+        cls.user = UserFactory.create()
+        cls.user.social_auth.create(
+            provider=EdxOrgOAuth2.name,
+            uid="{}_edx".format(cls.user.username),
+            extra_data={"access_token": "fooooootoken"}
+        )
+
+        cls.course_run1 = CourseRunFactory.create(freeze_grade_date=datetime.now(tz=pytz.UTC)-timedelta(days=1))
+        cls.course_run1.course.program.financial_aid_availability = True
```
giocalitri: ok

grades/api_test.py
```diff
@@ -0,0 +1,237 @@
...
+        assert fa.call_count == 1
+        call_arg = fa.call_args_list[0][0][0]
+        assert isinstance(call_arg, UserCachedRunData)
+        assert call_arg.edx_course_key == self.course_run1.edx_course_key
+
+        assert non_fa.called is True
+        assert non_fa.call_count == 1
+        call_arg = non_fa.call_args_list[0][0][0]
+        assert isinstance(call_arg, UserCachedRunData)
+        assert call_arg.edx_course_key == self.course_run3.edx_course_key
```
giocalitri: I think there is misunderstanding on the testing approach: I am testing a specific logic to finalize the grades and I think we need to go in the details here and not just see what is the final result.
I am leaving the tests as they are.

grades/api_test.py
```diff
@@ -0,0 +1,237 @@
...
+    @patch('grades.api._refresh_cache_final_grade', autospec=True)
+    def test_freeze_user_final_grade_error1(self, mock_refr, mock_get_fg):
+        """
+        Test for freeze_user_final_grade function in case of problems with can_freeze_grades
+        """
+        # case not ready to be frozen
+        api.freeze_user_final_grade(self.user, self.course_run3)
+        assert mock_refr.called is False
+        assert mock_get_fg.called is False
+        assert FinalGrade.objects.filter(user=self.user, course_run=self.course_run3).exists() is False
```
giocalitri: the error is because the the original function will terminate in an erroneous state (without raising anything).
I think `test_future_freeze_date` is more misleading because suggests that the function should work even for future courses

grades/api_test.py
giocalitri: The tested function is structured in a way to never fail (or at least it should not to). The only way to see if the function stops the execution at the expected points is to check if the some calls are or aren't executed. I am open to other solutions to accomplish the same.

grades/api_test.py
giocalitri: this was just a docstring copied from the previous test: changed it

grades/models.py
giocalitri: Changed to `CourseRunGradingStatus`

grades/models.py
giocalitri: ok

dashboard/api_edx_cache.py
```diff
@@ -280,3 +306,22 @@ def update_cache_if_expired(cls, user, edx_client, cache_type):
...
+        Updates only certificates and Current grade.
+        Used before a final grade freeze.
+
+        Args:
+            user (django.contrib.auth.models.User): A user
+        Returns:
+            None
+        """
+        # get the credentials for the current user for edX
+        user_social = user.social_auth.get(provider=EdxOrgOAuth2.name)
```
noisecapella: It looks like this code is uncovered. Can you write a test for it?

courses/models.py
```diff
@@ -250,3 +253,25 @@ def is_unexpired(self):
...
+            raise ImproperlyConfigured('Missing freeze_grade_date')
+        return datetime.now(pytz.utc) > self.freeze_grade_date
+
+    @classmethod
+    def get_freezable(cls):
+        """
+        Returns a queryset of all the runs that can freeze final grade according to the freeze date.
+        """
+        course_runs = cls.objects.exclude(
+            freeze_grade_date=None
```
noisecapella: Is there ever a case where `CourseRun` should have `freeze_grade_date` == `None`, except in the case where it's not set yet?

dashboard/api_edx_cache.py
giocalitri: sure

courses/models.py
giocalitri: AFAIK no.

grades/api.py
```diff
@@ -0,0 +1,155 @@
...
+
+def _compute_grade_for_fa(user_edx_run_data):
+    """
+    Gets the final grade for an user enrolled in the course run that is part of a financial aid program.
+
+    Args:
+        user_edx_run_data (dashboard.api_edx_cache.UserCachedRunData): the edx cached data for an user in a course run
+
+    Returns:
+        UserFinalGrade: a namedtuple of (float, bool,) representing the final grade
```
noisecapella: Is `current_grade.percent` is a number from 0 to 1? Is `certificate.grade` also 0 to 1?

grades/api.py
```diff
@@ -0,0 +1,155 @@
...
+            of the user in the course run and whether she passed it
+    """
+    run_passed = None
+    grade = None
+    if user_edx_run_data.certificate is not None:
```
noisecapella: I know you brought this up before but why do we want to prioritize the certificate grade over the current grade? Is it just because we need to choose one and be consistent?

grades/api.py
```diff
@@ -0,0 +1,155 @@
...
+    If the user has a certificate, she has passed otherwise she has not.
+
+    Args:
+        user_edx_run_data (dashboard.api_edx_cache.UserCachedRunData): the edx cached data for an user in a course run
+
+    Returns:
+        UserFinalGrade: a namedtuple of (float, bool,) representing the final grade
+            of the user in the course run and whether she passed it
+    """
+    run_passed = user_edx_run_data.certificate is not None
```
noisecapella: Does it matter whether the certificate `is_verified` or not?

grades/api.py
giocalitri: Yes and yes

grades/api.py
giocalitri: No, it is because we need to use certificates for old DEDP courses (that are not guaranteed to still have a consistent current grade). A complication is that the certificate is in most cases not verified.

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+            freeze_grade_date=datetime.now(tz=pytz.UTC)-timedelta(days=1),
+            course__program__financial_aid_availability=True,
+        )
+        cls.course_run2 = CourseRunFactory.create(
+            freeze_grade_date=None,
+            course__program=cls.course_run1.course.program,
+        )
+
+        cls.course_run3 = CourseRunFactory.create(
+            freeze_grade_date=datetime.now(tz=pytz.UTC)+timedelta(days=1),
```
noisecapella: It seems like this is redundant if the default `freeze_grade_date` is in the future anyway

grades/api.py
giocalitri: It should be verified in this case. 
I am tempted to put a check about it, but the reality is that the existence of a certificate for a non_fa course should mean that it is verified (meaning there should never be a case where there are both `verified` and `honor` certificates for non_fa

grades/api_test.py
giocalitri: I have not understood the comment

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        cls.certificates = {
+            course_run.edx_course_key: CachedCertificateFactory.create(
+                user=cls.user, course_run=course_run) for course_run in (cls.course_run2, cls.course_run4)
+        }
+
+        cls.user_edx_data = CachedEdxUserData(cls.user)
+
+    def test_compute_grade_for_fa(self):
```
noisecapella: I feel like this is a good candidate for a parameterized test. It's not clear from the names of the course runs what is significant about each of them, and some edge cases are not explicitly tested. For example we should test a non FA course run with a certificate whose freeze date is in the future. Can you rewrite as a parameterized test? Feel free to push back against this if you see a lot of effort in that, or feel it's unnecessary

grades/api_test.py
noisecapella: In `CourseRunFactory` it sets `freeze_grade_date` to some datetime in the future, after today but before the end of the year. This code sets the datetime one day in the future. Since all we care about is whether the datetime is in the future or the past, this line should be redundant

giocalitri: ah yes, I wanted to have the assignment to be explicit so it is more readable and we do not need to go and check the default value. No strong opinions, though.

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+    @patch('grades.api._compute_grade_for_non_fa', autospec=True)
+    @patch('grades.api._compute_grade_for_fa', autospec=True)
+    def test_get_final_grade(self, fa, non_fa):
+        """
+        tests for get_final_grade function
+        """
+        cruns = (self.course_run1, self.course_run3,)
+
+        for crun in cruns:
+            api.get_final_grade(self.user, crun)
```
noisecapella: Does this mutate any state? It seems like it's just calculating the grade, not storing it, though I haven't read through it thoroughly yet

grades/api_test.py
giocalitri: well, the tested function does not check whether the freeze data is in the future, so I am not sure it makes sense.
What I care is to test that if there is a certificate, the course is passed, otherwise it is not.

grades/management/commands/check_final_grade_freeze_status.py
```diff
@@ -0,0 +1,62 @@
...
+            else:
+                self.stdout.write(
+                    self.style.ERROR(
+                        'Final grades for course "{0}" are marked as they are being processed'
+                        ', but no task found.'.format(course_id)
+                    )
+                )
+        else:
+            self.stdout.write(
+                self.style.WARNING(
```
noisecapella: That's neat, I didn't know about `self.style.WARNING`

grades/api_test.py
giocalitri: No, it does not mutate the state nor store the result.

grades/models.py
```diff
@@ -79,3 +82,60 @@ def __str__(self):
...
+    def is_complete(cls, course_run):
+        """
+        Returns True if there is an entry with status 'complete'
+        """
+        return cls.objects.filter(course_run=course_run, status=FinalGradeStatus.COMPLETE).exists()
+
+    @classmethod
+    def is_pending(cls, course_run):
+        """
+        Returns True if there is an entry with status 'complete'
```
noisecapella: Should be `with status 'pending'`

grades/models.py
```diff
@@ -79,3 +82,60 @@ def __str__(self):
...
+    @classmethod
+    def create_pending(cls, course_run):
+        """
+        Creates an entry with status pending
+        """
+        course_fg_info, _ = cls.objects.get_or_create(course_run=course_run)
+        return course_fg_info
+
+    @classmethod
+    def get_frozen(cls):
```
noisecapella: It looks like we're not using this function anywhere. Is that intentional?

grades/models.py
giocalitri: yes

grades/models.py
giocalitri: I moved this logic somewhere else: I'll remove this method

micromasters/utils.py
```diff
@@ -157,3 +157,19 @@ def is_near_now(time):
     now = datetime.datetime.now(tz=pytz.UTC)
     five_seconds = datetime.timedelta(0, 5)
     return now - five_seconds < time < now + five_seconds
+
+
+def chunks(list_to_chunk, chunk_size=20):
```
noisecapella: I know you just copied this code but are we testing it anywhere? If not we should be, maybe in a separate PR

giocalitri: No, it was not tested: I will add a couple of tests here

grades/tasks_test.py
```diff
@@ -0,0 +1,173 @@
...
+        assert info_run.course_run == self.course_run2
+        assert info_run.status == FinalGradeStatus.COMPLETE
+
+    @patch('grades.api.freeze_user_final_grade', autospec=True)
+    def test_freeze_course_run_final_grades_4(self, freeze_single_user):
+        """
+        Test for the test_freeze_course_run_final_grades
+        task in case there are users to be processed
+
+        NOTE: In this test it is moched a function in a subtask and not the subtask.
```
noisecapella: `moched` -> `mocked`

grades/api_test.py
noisecapella: Why are we running it then? To assert that no exception is raised?

grades/api_test.py
noisecapella: Ok, I'm fine to leave it as is

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = UserFactory.create()
+        cls.user.social_auth.create(
+            provider=EdxOrgOAuth2.name,
+            uid="{}_edx".format(cls.user.username),
+            extra_data={"access_token": "fooooootoken"}
+        )
+
+        cls.course_run1 = CourseRunFactory.create(
```
noisecapella: Can we rename these four variables to something more descriptive? For example `run_fa` `run_fa_with_cert` `run_no_fa`, `run_no_fa_with_cert`.

grades/api_test.py
noisecapella: I'm fine leaving the test as is as long as all relevant edge cases are tested. Can you briefly describe what information the test is looking at to the docstring?

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+        tests for _get_compute_func function
+        """
+        assert api._get_compute_func(self.course_run1) == api._compute_grade_for_fa
+        assert api._get_compute_func(self.course_run3) == api._compute_grade_for_non_fa
+
+    @patch('grades.api._compute_grade_for_non_fa', autospec=True)
+    @patch('grades.api._compute_grade_for_fa', autospec=True)
+    def test_get_final_grade(self, fa, non_fa):
+        """
+        tests for get_final_grade function
```
noisecapella: Can you describe what is significant about the course runs being tested below? Why are `course_run2` and `course_run4` not in the list below?

grades/api_test.py
noisecapella: Ah, it looks like you're doing this to test that the mocked methods are called. Can you add a comment explaining this?

noisecapella: Actually it seems like we should assert that the return value for this function is the same as the return value of the compute function. Can you add a return value to the mock and assert that it's the same?

grades/api_test.py
giocalitri: ok

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+        fg_status.save()
+        assert [user.pk for user in api.get_users_without_frozen_final_grade(self.course_run1)] == [self.user.pk]
+
+    @patch('grades.api.get_final_grade', autospec=True)
+    @patch('dashboard.api_edx_cache.CachedEdxDataApi.update_all_cached_grade_data', new_callable=MagicMock)
+    def test_freeze_user_final_grade_error1(self, mock_refr, mock_get_fg):
+        """
+        Test for freeze_user_final_grade function in case of problems with can_freeze_grades
+        """
+        # case not ready to be frozen
```
noisecapella: Why is `course_run3` not ready to be frozen? Can you add this to the comment?

grades/models.py
```diff
@@ -79,3 +82,49 @@ def __str__(self):
...
+        Returns True if there is an entry with status 'complete'
+        """
+        return cls.objects.filter(course_run=course_run, status=FinalGradeStatus.COMPLETE).exists()
+
+    @classmethod
+    def is_pending(cls, course_run):
+        """
+        Returns True if there is an entry with status 'pending'
+        """
+        return cls.objects.filter(course_run=course_run, status=FinalGradeStatus.PENDING).exists()
```
noisecapella: Can you add a test for this function?

grades/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+from micromasters.utils import chunks
+
+
+CACHE_ID_BASE_STR = "freeze_grade_{0}"
+
+log = logging.getLogger(__name__)
+
+
+@async.task
+def find_course_runs_for_grade_freeze():
```
noisecapella: Can you rename this task? `find_course_runs_for_grade_freeze` seem like a task that just collects information, but this task also freezes grades. Maybe `find_course_runs_and_freeze_grades`?

grades/api_test.py
giocalitri: ok

grades/api_test.py
giocalitri: ok

grades/api_test.py
giocalitri: ok

grades/models.py
```diff
@@ -79,3 +82,49 @@ def __str__(self):
...
+
+    @classmethod
+    def set_to_complete(cls, course_run):
+        """
+        Sets the status for the course_run to complete
+        """
+        course_run_grade_status, _ = cls.objects.get_or_create(course_run=course_run)
+        if course_run_grade_status.status != FinalGradeStatus.COMPLETE:
```
noisecapella: It looks like this line is only partially covered. Can you add a test that asserts that nothing happens if all grades are already frozen?

grades/api_test.py
```diff
@@ -0,0 +1,232 @@
...
+        Test for freeze_user_final_grade function in case of problems with getting the final grade
+        """
+        mock_get_fg.side_effect = AttributeError
+        api.freeze_user_final_grade(self.user, self.course_run1)
+        mock_refr.assert_called_once_with(self.user)
+        mock_get_fg.assert_called_once_with(self.user, self.course_run1)
+        assert FinalGrade.objects.filter(user=self.user, course_run=self.course_run1).exists() is False
+
+    @patch('dashboard.api_edx_cache.CachedEdxDataApi.update_all_cached_grade_data', new_callable=MagicMock)
+    def test_freeze_user_final_grade_gio(self, mock_refr):
```
giocalitri: sure

grades/models.py
giocalitri: the reason why this is uncovered is because the only call is done in a management command. 
I think I will add tests for the entire class

travis-docker-compose.yml
```diff
@@ -23,13 +28,38 @@ services:
...
+      BROKER_URL: redis://redis:6379/4
+      CELERY_RESULT_BACKEND: redis://redis:6379/4
     env_file: .env
     ports:
       - "8079:8079"
     links:
       - db
       - elastic
 
+  celery:
```
noisecapella: Which tests do we run using an instance of Celery? I remember talking to you about this earlier but it seems like  `CELERY_ALWAYS_EAGER` is always true, so I'm not sure we're actually using Celery in any of the tests

grades/api_test.py
giocalitri: ok

grades/models.py
giocalitri: ok

grades/tasks.py
giocalitri: ok

grades/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+    # pylint: disable=bare-except
+    for user in users:
+        try:
+            api.freeze_user_final_grade(user, course_run)
+        except:
+            log.exception(
+                'Impossible to freeze final grade for user "%s" in course %s',
+                user.username, course_run.edx_course_key
+            )
+            continue
```
giocalitri: ok

grades/tasks_test.py
giocalitri: ok

travis-docker-compose.yml
giocalitri: The reason why we need Celery is not for running the async task, but for these commands:
https://github.com/mitodl/micromasters/pull/2126/files#diff-bf4aef0ab77eb78aa7345a16af8ad33cR99
and 
https://github.com/mitodl/micromasters/pull/2126/files#diff-bf4aef0ab77eb78aa7345a16af8ad33cR72

grades/tasks_test.py
```diff
@@ -0,0 +1,173 @@
...
+        tasks.freeze_course_run_final_grades.delay(self.course_run1)
+        info_run.refresh_from_db()
+        assert info_run.status == FinalGradeStatus.PENDING
+        cached_celery_task_id_2 = cache.get(cache_id)
+        assert cached_celery_task_id_2 is not None
+        assert cached_celery_task_id_2 != cached_celery_task_id_1
+        remaining_users = self.users[:5]
+        assert freeze_single_user.call_count == len(remaining_users)
+        for user in remaining_users:
+            freeze_single_user.assert_any_call(user, self.course_run1)
```
noisecapella: Can you also assert that `freeze_single_user` was not called again for `successful_users`?

travis-docker-compose.yml
noisecapella: Ah, I see

grades/tasks_test.py
noisecapella: Yes, makes sense

PR #2129, opened 2016-12-14
Disable employment and education delete buttons during API activity

noisecapella:
Part of #1752

#### What's this PR do?
Disable employment and education delete buttons during API activity

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -87,7 +87,7 @@ class EducationForm extends ProfileFormFields {
   };
 
   deleteEducationEntry: Function = (): void => {
-    deleteEducationEntry.call(this);
+    return deleteEducationEntry.call(this);
   };
 
   educationLevelRadioSwitch: Function = (level: Object): React$Element<*> => {
@@ -389,6 +389,7 @@ class EducationForm extends ProfileFormFields {
           open={showEducationDeleteDialog}
           close={this.closeConfirmDeleteDialog}
           itemText="degree"
+          inFlight={inFlight}
         />
         <Dialog
           title="Education"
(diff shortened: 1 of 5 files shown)


aliceriot:
When the deletion request is in-flight the 'Next' button at the bottom spins as well:

weird_spinner_thing

that was a difficult screenshot to grab! haha

noisecapella:
Yeah, that happens with the pay buttons too. When you click one pay button they all have the spinner. I'm not sure what we can do here without increasing the complexity of the fetchStatus code. What do you think?

aliceriot:
I think it could be fine for now, in the future each button could have a `UUID` which would be dispatched to the store on click, which would allow us to discern which button was clicked. Maybe we should file an issue and come back to that? That would be a bit of work to implement so I don't think it quite makes sense to try to solve that problem here.

static/js/components/EducationForm.js
```diff
@@ -87,7 +87,7 @@ class EducationForm extends ProfileFormFields {
   };
 
   deleteEducationEntry: Function = (): void => {
-    deleteEducationEntry.call(this);
+    return deleteEducationEntry.call(this);
```
aliceriot: is this because `deleteEducationEntry` returns a `Promise`?

aliceriot: (below for similar work history change too)

static/js/util/profile_history_edit.js
```diff
@@ -79,5 +79,5 @@ export function deleteWorkHistoryEntry () {
   const { saveProfile, profile, ui } = this.props;
   let clone = _.cloneDeep(profile);
   clone['work_history'].splice(ui.deletionIndex, 1);
-  saveProfile(employmentValidation, clone, ui);
+  return saveProfile(employmentValidation, clone, ui);
```
aliceriot: ah ok I see, these functions were changed to return the `saveProfile` promise. looks good!

noisecapella:
Sounds good, I created https://github.com/mitodl/micromasters/issues/2140

Issue #810, opened 2016-08-04
Search: Facet on number of courses completed


pdpinch:
## Acceptance Criteria

- [x] precompute number of completed courses in a program and index in elastic search
- [x] simple slider that filters on simple count of completed courses in a program
- [x] appears below "Average Grade in Program" facet and above "Country of Residence"

![image](https://cloud.githubusercontent.com/assets/430126/17684912/5f099dae-632d-11e6-884f-51a4e5f1c66a.png)

roberthouse54:
![image](https://cloud.githubusercontent.com/assets/20047260/20192930/e9e0e156-a759-11e6-86d3-b66494242421.png)

Since Course and Semester are now going to be the main facets we show at the top of the filter list, we will push down some of the other facets into a bottom section. All of the facets in this bottom section will be collapsed by default. These will include Averaget Grade in Program, # of Completed Courses, Current Residence, Country of Birth, etc. We will probably add other facets later in this section.

Number of Courses filters by the number of courses a user has completed. It appears below Average Grade in Program.

pdpinch:
@roberthouse54 , does this display a histogram like grade, or is it just a slider?

roberthouse54:
I wasn't thinking that we needed a histogram here. It was more my intuition here. My thinking was that the histogram makes sense in showing a grade distribution. I don't know if it is needed. Did you think it would be helpful here?

noisecapella:
Just to be clear, does completed mean passed?

roberthouse54:
Just a slider. Completed does mean passed. Maybe we should change the label to "# Courses Passed"

PR #2133, opened 2016-12-15
Added Facet on number of courses completed

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/810

@pdpinch 
screen shot 2016-12-16 at 12 42 22 am

### DIFF
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -1,6 +1,7 @@
 """
 Provides functionality for serializing a ProgramEnrollment for the ES index
 """
+from dashboard.api import get_info_for_course, CourseStatus
 from dashboard.api_edx_cache import CachedEdxUserData
 from dashboard.models import CachedEnrollment
 from dashboard.utils import MMTrack
@@ -38,6 +39,29 @@ class UserProgramSearchSerializer:
         return serialized_enrollments
 
     @classmethod
+    def count_courses_passed(cls, program, mmtrack):
+        """
+        Returns number of passed courses.
+
+        Args:
+            program (Program): A program object.
+            mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program
+        Returns:
+            num_complete_courses (int): A number of passed courses.
+        """
+        num_courses_passed = 0
+        for course in program.course_set.all():
+            course_info_dict = get_info_for_course(course, mmtrack)
+            if course_info_dict:
+                expected_statuses = list(
+                    filter(lambda c: c['status'] in [CourseStatus.PASSED], course_info_dict["runs"])
+                )
+                if len(expected_statuses) > 0:
+                    num_courses_passed += 1
+
+        return num_courses_passed
+
+    @classmethod
     def serialize(cls, program_enrollment):
         """
         Serializes a ProgramEnrollment object
@@ -52,10 +76,17 @@ class UserProgramSearchSerializer:
             edx_user_data
         )
         enrollments_qset = CachedEnrollment.user_course_qset(user, program=program)
+        num_courses_passed = cls.count_courses_passed(
+            program,
+            mmtrack
+        )
+
         return {
             'id': program.id,
             'enrollments': cls.serialize_enrollments(mmtrack, enrollments_qset),
             'grade_average': mmtrack.calculate_final_grade_average(),
             'is_learner': is_learner(user, program),
             'email_optin': user.profile.email_optin,
+            'num_courses_passed': num_courses_passed,
+            'total_courses': program.course_set.count()
         }
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -1,7 +1,6 @@
 """
 Provides functionality for serializing a ProgramEnrollment for the ES index
 """
-from dashboard.api import get_info_for_course, CourseStatus
 from dashboard.api_edx_cache import CachedEdxUserData
 from dashboard.models import CachedEnrollment
 from dashboard.utils import MMTrack
@@ -39,29 +38,6 @@ class UserProgramSearchSerializer:
         return serialized_enrollments
 
     @classmethod
-    def count_courses_passed(cls, program, mmtrack):
-        """
-        Returns number of passed courses.
-
-        Args:
-            program (Program): A program object.
-            mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program
-        Returns:
-            num_complete_courses (int): A number of passed courses.
-        """
-        num_courses_passed = 0
-        for course in program.course_set.all():
-            course_info_dict = get_info_for_course(course, mmtrack)
-            if course_info_dict:
-                expected_statuses = list(
-                    filter(lambda c: c['status'] in [CourseStatus.PASSED], course_info_dict["runs"])
-                )
-                if len(expected_statuses) > 0:
-                    num_courses_passed += 1
-
-        return num_courses_passed
-
-    @classmethod
     def serialize(cls, program_enrollment):
         """
         Serializes a ProgramEnrollment object
@@ -76,10 +52,6 @@ class UserProgramSearchSerializer:
             edx_user_data
         )
         enrollments_qset = CachedEnrollment.user_course_qset(user, program=program)
-        num_courses_passed = cls.count_courses_passed(
-            program,
-            mmtrack
-        )
 
         return {
             'id': program.id,
@@ -87,6 +59,6 @@ class UserProgramSearchSerializer:
             'grade_average': mmtrack.calculate_final_grade_average(),
             'is_learner': is_learner(user, program),
             'email_optin': user.profile.email_optin,
-            'num_courses_passed': num_courses_passed,
+            'num_courses_passed': mmtrack.count_courses_passed(),
             'total_courses': program.course_set.count()
         }
--- a/dashboard/serializers_test.py
+++ b/dashboard/serializers_test.py
@@ -15,7 +15,6 @@ from courses.factories import (
     CourseRunFactory,
 )
 from dashboard.api_edx_cache import CachedEdxUserData
-from dashboard.api import CourseStatus
 from dashboard.factories import (
     CachedCertificateFactory,
     CachedCurrentGradeFactory,
@@ -220,7 +219,7 @@ class UserProgramSearchSerializerTests(TestCase):
         Tests that full ProgramEnrollment serialization works as expected when user
         has passed a course.
         """
-        with patch.object(UserProgramSearchSerializer, 'count_courses_passed', return_value=0):
+        with patch.object(MMTrack, 'count_courses_passed', return_value=0):
             Profile.objects.filter(pk=self.profile.pk).update(email_optin=True)
             self.profile.refresh_from_db()
             program = self.program_enrollment.program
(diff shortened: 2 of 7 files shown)


dashboard/serializers.py
```diff
@@ -25,6 +25,25 @@ def serialize_enrollments(cls, enrollments):
...
+
+        Args:
+            program (Program): A program object.
+            mmtrack (MMTrack): mmtrack utility.
+        Returns:
+            num_complete_courses: A number of complete courses.
+        """
+        num_complete_courses = 0
+        for course in program.course_set.all():
+            course_run = course.first_unexpired_run()
```
amir-qayyum-khan: @giocalitri @noisecapella this code is correct for getting edX course id.

I want to get number of courses complete here

noisecapella: I believe this logic is correct

dashboard/serializers.py
```diff
@@ -25,6 +25,25 @@ def serialize_enrollments(cls, enrollments):
...
+            if course_run and mmtrack.has_passed_course(course_run.edx_course_key):
+                num_complete_courses = num_complete_courses + 1
```
amir-qayyum-khan: ok

amir-qayyum-khan:
@noisecapella please review?

Ferdi:
@amir-qayyum-khan , thanks !  why is max number of courses == 3 ?     Maybe it should be equal to the number of courses in that particular program ?

amir-qayyum-khan:
yes @Ferdi,  i have 3 courses in selected program on my machine. this max
value depends on number of courses in program.

dashboard/serializers.py
```diff
@@ -25,6 +25,25 @@ def serialize_enrollments(cls, enrollments):
...
+        Returns:
+            num_complete_courses: A number of passed courses.
+        """
+        num_courses_passed = 0
+        for course in program.course_set.all():
+            course_run = course.first_unexpired_run()
+            if course_run and mmtrack.has_passed_course(course_run.edx_course_key):
+                num_courses_passed += 1
+
+        return num_courses_passed
```
gsidebo: i don't think the logic for this method is correct. for each run, we're running `course.first_unexpired_run()` to get a course run and check if it has passed. if a user has passed a course, that run is in the past, or in other words, it has expired. 

also, `MMTrack` already populates a list of every course run key in a given program (https://github.com/mitodl/micromasters/blob/master/dashboard/utils.py#L57), so if we need to know the number of passed course runs, that method can be implemented in `MMTrack` with very little work. if we need to know the number of _courses_ passed, we can implement that in `MMTrack` as well and avoid extra queries. we can change this line (which i linked above):

```python
self.course_ids = set(CourseRun.objects.filter(course__program=self.program).exclude(
                Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
            ).values_list("edx_course_key", flat=True))
```

and instead implement it like this:

```python
# Maps a CourseRun's edx_course_key to its parent Course id
self.edx_key_course_map = dict(
    CourseRun.objects.filter(course__program=program).exclude(
        Q(edx_course_key__isnull=True) | Q(edx_course_key__exact='')
    ).values_list("edx_course_key", "course__id")
)
self.course_ids = self.edx_key_course_map.keys()
```

we can then implement `num_courses_passed` in `MMTrack` by looping through `self.course_ids`, checking `self.has_passed_course(course_id)`, and getting the count of unique course id's in `self.edx_key_course_map`. @noisecapella what do you think?

dashboard/serializers.py
```diff
@@ -25,6 +25,25 @@ def serialize_enrollments(cls, enrollments):
...
+
+        Args:
+            program (Program): A program object.
+            mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program
+        Returns:
+            num_complete_courses: A number of passed courses.
+        """
+        num_courses_passed = 0
+        for course in program.course_set.all():
+            course_run = course.first_unexpired_run()
```
noisecapella: It looks like I was wrong, `first_unexpired_run()` is not the function we should use here. Can you use `get_info_for_course` and get the status from the dict returned from that function? We should have only one place where the decision of whether a course is passed or not is made

dashboard/serializers.py
noisecapella: I agree that `first_unexpired_run()` was the wrong function but I'd rather not do any heavy refactoring in this PR. We should use the output of `get_info_for_course` so that the decision on whether a course is passed or not is made in one place. A refactor would be good in a future PR though, I just prefer to keep the refactoring and feature work separate to make reviewing easier and to reduce the possibility of bugs slipping through

amir-qayyum-khan:
Done @noisecapella @gsidebo

dashboard/serializers.py
```diff
@@ -38,6 +39,29 @@ def serialize_enrollments(cls, mmtrack, enrollments):
...
+        for course in program.course_set.all():
+            course_info_dict = get_info_for_course(course, mmtrack)
+            if course_info_dict:
+                expected_statuses = list(
+                    filter(lambda c: c['status'] in [CourseStatus.PASSED], course_info_dict["runs"])
+                )
+                if len(expected_statuses) > 0:
+                    num_courses_passed += 1
+
+        return num_courses_passed
```
gsidebo: @noisecapella continuing the discussion from [here](https://github.com/mitodl/micromasters/pull/2133#discussion_r92888017)...

first of all i would stress that my suggestion was not a refactor. it was a change to some existing code to support a new feature. second, if my suggestion seemed heavy, i probably didn't explain myself well enough. here is the diff to implement my suggestion in its entirety: https://gist.github.com/gsidebo/812ecb6480656d772bbc85eb83217e8b

here are the diff stats for my changes on top of the current branch. i disagree that this is in any way heavy:
```
 dashboard/serializers.py | 30 +-----------------------------
 dashboard/utils.py       | 17 ++++++++++++++---
 2 files changed, 15 insertions(+), 32 deletions(-)
```

i tested locally and it works as expected. your approach uses `get_info_for_course`, which is used to serialize a course and all of its course runs. the work to loop through and serialize course runs is unnecessary for the purposes of counting the number of passed courses. you'll also see that my suggestion uses `MMTrack.has_passed_course`, which is the exact method that `get_info_for_course` uses to determine if a course is passed ([here](https://github.com/mitodl/micromasters/blob/master/dashboard/api.py#L247) and [here](https://github.com/mitodl/micromasters/blob/master/dashboard/api.py#L263)). `get_info_for_course` doesn't decide whether a course run is passed; `MMTrack.has_passed_course` does, and `get_info_for_course` makes use of it.

gsidebo: it doesn't count course runs twice. below i pasted some commands and their results. you can try them yourself if you like. `self.edx_key_course_map` is a dict that maps a course run edx_course_key to the parent course's id. `count_courses_passed` creates a set of the `Course.id`'s that have been passed, ensuring uniqueness. again, i want to stress that we shouldn't confuse "changing existing code to meet new requirements" with "refactoring". those are not the same thing. if you are concerned with some particular risks with the change i'm suggesting to `MMTrack`, please let me know what that is specifically. i can identify 2 specific problems with the current approach. (1) we are using `get_info_for_course` for a purpose that it wasn't intended, and that function is performing a lot of work that doesn't need to be done. (2) it should not be the responsibility of a serializer class to calculate how many courses a student has passed. `MMTrack` is the class that determines if a single course run passed, so it makes sense that it should be able to tell us how many courses have been passed. if you still have questions/concerns, it might be better to chat in person

```shell
(micromasters) ~/d/micromasters ❯❯❯ git status --short                                                       
 M dashboard/serializers.py
 M dashboard/utils.py
(micromasters) ~/d/micromasters ❯❯❯ curl -s -X POST "http://192.168.99.100:9100/micromasters/_search?q=profile.username:fake.kenzi.cunningham" | jq '.["hits"]["hits"][0]["_source"]["program"]["num_courses_passed"]'
2
(micromasters) ~/d/micromasters ❯❯❯ mmmanage alter_data --action=add_past_passed_run --username=staff --program-title='Analog' --course-title='100' --add-if-exists
Course run changed: (id: 2479, edx_course_key: course-v1:MITx+Analog+Learning+100+Jan_2015)
User: staff; Course: Analog Learning 100
(micromasters) ~/d/micromasters ❯❯❯ docker-compose run web ./manage.py recreate_index                        
(micromasters) ~/d/micromasters ❯❯❯ curl -s -X POST "http://192.168.99.100:9100/micromasters/_search?q=profile.username:fake.kenzi.cunningham" | jq '.["hits"]["hits"][0]["_source"]["program"]["num_courses_passed"]'
2
(micromasters) ~/d/micromasters ❯❯❯ mmmanage alter_data --action=set_course_to_passed --username=kenzi --program-title='Analog' --course-title='300'
Course run changed: (id: 2483, edx_course_key: course-v1:MITx+Analog+Learning+300+Oct_2016)
User: fake.kenzi.cunningham; Course: Analog Learning 300
(micromasters) ~/d/micromasters ❯❯❯ docker-compose run web ./manage.py recreate_index                        
(micromasters) ~/d/micromasters ❯❯❯ curl -s -X POST "http://192.168.99.100:9100/micromasters/_search?q=profile.username:fake.kenzi.cunningham" | jq '.["hits"]["hits"][0]["_source"]["program"]["num_courses_passed"]'
3
(micromasters) ~/d/micromasters ❯❯❯ mmmanage alter_data --action=add_past_passed_run --username=staff --program-title='Analog' --course-title='300' --add-if-exists
Course run changed: (id: 2485, edx_course_key: course-v1:MITx+Analog+Learning+300+Jan_2015)
User: staff; Course: Analog Learning 300
(micromasters) ~/d/micromasters ❯❯❯ docker-compose run web ./manage.py recreate_index                        
(micromasters) ~/d/micromasters ❯❯❯ curl -s -X POST "http://192.168.99.100:9100/micromasters/_search?q=profile.username:fake.kenzi.cunningham" | jq '.["hits"]["hits"][0]["_source"]["program"]["num_courses_passed"]'
3
```

gsidebo: i also don't understand "do this refactoring in a separate PR, to keep this one small and focused on the issue". the changes i'm suggesting are _directly_ related to this issue, and is 4 lines of code shorter (if you add the docstring on `count_courses_passed` which i forgot in my gist)

amir-qayyum-khan:
@gsidebo done, updated with logic that you suggested, it worked fine on tests.

If a course have multiple course run and there are mode the one of these courses with statues passed then it is counting whole course as 1. Which is good

dashboard/utils_test.py
```diff
@@ -451,3 +460,27 @@ def test_get_current_grade(self):
         # case when the grade is not available from edx
         with patch('edx_api.grades.models.CurrentGrades.get_current_grade', return_value=None):
             assert mmtrack.get_current_grade("course-v1:MITx+8.MechCX+2014_T1") is None
+
+    def count_courses_passed_normal(self):
```
noisecapella: Can you add a test for the case where there are two passed course runs in a course?

dashboard/serializers_test.py
```diff
@@ -202,9 +209,30 @@ def test_full_program_user_serialization_staff(self):
...
             'is_learner': False,
-            'email_optin': True
+            'email_optin': True,
+            'num_courses_passed': 1,
+            'total_courses': 1
         }
 
+    def test_full_program_user_serialization_with__no_passed_course(self):
+        """
+        Tests that full ProgramEnrollment serialization works as expected when user
```
noisecapella: Can you update this docstring?

dashboard/utils_test.py
```diff
@@ -451,3 +460,27 @@ def test_get_current_grade(self):
...
+            user=self.user,
+            program=self.program,
+            edx_user_data=self.cached_edx_user_data
+        )
+        # this is a verified enrollment and a verified certificate from edx
+        assert mmtrack.count_courses_passed() == 1
+
+    def count_courses_passed_fa(self):
+        """
+        Test for count_courses_passed_normal method in case of a normal program
```
noisecapella: Please update this docstring

dashboard/utils_test.py
```diff
@@ -451,3 +460,27 @@ def test_get_current_grade(self):
...
+        """
+        mmtrack = MMTrack(
+            user=self.user,
+            program=self.program_financial_aid,
+            edx_user_data=self.cached_edx_user_data
+        )
+        # this is a verified enrollment and a verified certificate from edx
```
noisecapella: This is not true for FA programs. Can you delete or adjust the comment?

noisecapella:
Functionality works fine, feel free to merge after addressing comments :+1:

PR #2142, opened 2016-12-15
Added minimal .editorconfig

rhysyngsun:
#### What's this PR do?

Adds a `.editorconfig` file, which informs editors/ides of some basic code styles. In this case I'd only added settings for Python and JS files to be space-indented with 4 and 2 spaces respectively. I also put in no trailing newline at EOF for Python since our linter enforces that too.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*.py]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.js]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+insert_final_newline = true


.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*.py]
+indent_style = spaces
```
noisecapella: This should be `space`

aliceriot:
i'm in favor of trailing newlines, if only to avoid that little 🚫 sign on github

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*.py]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = false
+
+[static/**.js]
```
aliceriot: is there a reason we wouldn't want to just do `*.js` here?

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*.py]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = false
```
noisecapella: Can we have this be `true` here and in the `js` below too?

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*.py]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[s*.js]
```
noisecapella: Typo here

noisecapella:
:+1:

PR #2145, opened 2016-12-15
Remove `foo: Function = () =>` annotations

aliceriot:
#### What's this PR do?

We had a lot of arrow functions declared on classes, annotated with the `Function` type, like this:

```js
class Foo {
  bar: Function = () => (
    'potatoes'
  );
}
```

Turns out that annotating functions like that is sort of equivalent to annotating them with the `any` type, meaning that the return types and so on aren't checked. Not good!

We had to do this previously because at a certain point Flow didn't properly support annotating that syntax, but now it can.

Anyway, I removed any annotation of an arrow function with `Function` and fixed all the flow errors which cropped up in that process (quite a few).

Read over the changes, make sure my regex to remove the `Function` annotations wasn't messed up somehow, and make sure the changes make sense, in general.

#### testing

I think just make sure there are no regressions.

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -41,7 +41,7 @@ const personalTabValidator = combineValidators(
 
 export default class PersonalTab extends React.Component {
   props: {
-    addProgramEnrollment:     Function,
+    addProgramEnrollment:     (p: number) => void,
     currentProgramEnrollment: AvailableProgram,
     errors:                   ValidationErrors,
     nextStep:                 () => void,
--- a/static/js/components/ProfileProgressControls.js
+++ b/static/js/components/ProfileProgressControls.js
@@ -6,28 +6,29 @@ import { saveProfileStep } from '../util/profile_edit';
 import { FETCH_PROCESSING } from '../actions';
 import type { Profile, SaveProfileFunc } from '../flow/profileTypes';
 import type { UIState } from '../reducers/ui';
+import type { Validator } from '../lib/validation/profile';
 
 export default class ProfileProgressControls extends React.Component {
   props: {
-    nextUrl?:      string,
-    prevUrl?:      string,
-    nextBtnLabel: string,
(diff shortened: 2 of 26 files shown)


static/js/components/ProfileProgressControls.js
```diff
@@ -9,25 +9,25 @@ import type { UIState } from '../reducers/ui';
...
+    isLastTab:               boolean,
+    nextBtnLabel:            string,
+    nextUrl?:                string,
+    prevUrl?:                string,
+    profile:                 Profile,
+    profilePatchStatus:      ?string,
+    programIdForEnrollment?:  ?number,
+    saveProfile:             SaveProfileFunc,
+    ui:                      UIState,
+    validator:               Function,
```
noisecapella: Can we use the `Validator` type here?

static/js/components/ProfileProgressControls.js
```diff
@@ -9,25 +9,25 @@ import type { UIState } from '../reducers/ui';
...
-    nextBtnLabel: string,
-    isLastTab:    boolean,
-    validator:    Function,
-    profile:      Profile,
-    profilePatchStatus: ?string,
-    ui:           UIState,
-    saveProfile:  SaveProfileFunc,
-    programIdForEnrollment: ?number,
-    addProgramEnrollment: Function,
+    addProgramEnrollment:    Function,
```
noisecapella: Can you make a more specific type here?

aliceriot:
comments should all be addressed!

static/js/components/PersonalTab.js
```diff
@@ -41,7 +41,7 @@ const personalTabValidator = combineValidators(
 
 export default class PersonalTab extends React.Component {
   props: {
-    addProgramEnrollment:     Function,
+    addProgramEnrollment:     (p: number) => void,
```
noisecapella: Can you do this update in `EducationTab` and `EmploymentTab` too?

PR #2146, opened 2016-12-15
Store a smaller version of avatars and use on leaners page

### LINKED ISSUE #1889: Images in learner search should all be the same size
On micromasters-rc:

screenshot from 2016-11-29 17-12-32

### LINKED ISSUE #2055: Search: create thumbnails of user profile images
In the search view, we force the user to download a large number of sometimes very large profile images. 

We should create standard, optimized thumbnails for these, to improve search performance.

noisecapella:
Fixes #2055
Fixes #1889

#### What's this PR do?
- Creates a new `image_small` field which stores a smaller version of the avatar
- Adds `image_small` to the serializer readonly fields and populates it when `image` is updated. `image_small` should not be updateable by itself
- Store it when indexing the profiles

- Make sure you have at least two users, one who is staff for a program and one who is a learner for that program. At least the learner should have an avatar.
- Run the migrations, then run `recreate_index`
- Log in as the staff user and go to `/learners` for the program and view the avatars. If you inspect the avatar image it should be `64x64` in size.
- Log in as the learner and update your avatar to something else
- Log in as the staff user again. Go to `/learners` and view the avatars. You should see the new avatar. If you inspect the image it should be `64x64` in size.


profiles/util_test.py
```diff
@@ -1,62 +1,131 @@
...
+        full_image_file.seek(0)
 
-        too_long_url = '{}.jpg'.format('a' * 150)
-        assert len(util.profile_image_upload_uri(None, too_long_url)) == 100
+        with patch('profiles.util.make_small_dimensions', return_value=(thumb_width, thumb_height)) as mocked:
+            thumb_file = util.make_thumbnail(full_image_file)
+            thumb_image = Image.open(thumb_file)
+            mocked.assert_called_with(width, height)
+            assert thumb_image.width == thumb_width
+            assert thumb_image.height == thumb_height
```
giocalitri: I think you need a test for the path too long even without the filename

profiles/views_test.py
```diff
@@ -397,3 +400,102 @@ def test_upload_image(self):
         assert resp.status_code == 200, resp.content.decode('utf-8')
         assert profile.education.count() == 1
         assert profile.work_history.count() == 1
+
+    @ddt.data(
+        [True, False],
+        [True, False]
```
giocalitri: why running twice with the same input?

profiles/views_test.py
```diff
@@ -397,3 +400,102 @@ def test_upload_image(self):
...
+
+        # save old thumbnail
+        backup_thumb_bytes = profile.image_small.file.read()
+        profile.image_small.file.seek(0)
+
+        # format patch using multipart upload
+        resp = self.client.patch(self.url1, data={
+            'image_small': image_file
+        }, format='multipart')
+        assert resp.status_code == 200, resp.content.decode('utf-8')
```
giocalitri: why this request is returning 200 and not 400?

profiles/migrations/0025_populate_image_small.py
```diff
@@ -0,0 +1,28 @@
...
+
+def populate_image_small(apps, schema_editor):
+    """
+    Populate image_small with thumbnail of image if it exists
+    """
+    Profile = apps.get_model('profiles.Profile')
+    for profile in Profile.objects.all():
+        if profile.image and not profile.image_small:
+            thumbnail = make_thumbnail(profile.image.file)
+            profile.image_small.save("{}.jpg".format(uuid4().hex), thumbnail)
```
noisecapella: Just in case the name gets used in a way where it would collide with the other image names. I don't know if it matters

profiles/models_test.py
```diff
@@ -37,16 +46,36 @@ def test_student_id_doesnt_change_on_save(self):
...
+    """
+    Tests for image fields
+    """
+    def test_upload_to(self):
+        """
+        Image upload_to should have a function which creates a path
+        """
+        now_mock = Mock(return_value=datetime.now(tz=pytz.UTC))
+        # pin the timestamps used in creating the URL
+        with patch('profiles.util.datetime', now=now_mock):
```
noisecapella: Sure

profiles/util.py
```diff
@@ -2,11 +2,17 @@
...
 
+from PIL import Image
 import pytz
 
-IMAGE_FILENAME_MAX_LENGTH = 64
+# This is the Django ImageField max path size
+IMAGE_PATH_MAX_LENGTH = 100
+
+# Max dimension of either height or width for a small image
+IMAGE_SMALL_MAX_DIMENSION = 64
```
noisecapella: I talked with @roberthouse54 and he doesn't see an immediate need for this to be configurable

profiles/util.py
```diff
@@ -29,15 +35,93 @@ def split_name(name):
...
+        path_format = "profile/{name}-{timestamp}{suffix}{ext}"
+
+        path_without_name = path_format.format(
+            timestamp=timestamp.strftime("%Y-%m-%dT%H%M%S-%z"),
+            suffix=suffix,
+            ext=ext,
+            name='',
+        )
+        if len(path_without_name) >= IMAGE_PATH_MAX_LENGTH:
+            raise Exception("path is longer than max length even without name: {}".format(path_without_name))
```
noisecapella: If it fails here then `suffix` is too large. Since the end user should not have control over `suffix` I think an exception is the way to go here. Otherwise we would need to cut off `name` or `timestamp`

profiles/util.py
```diff
@@ -29,15 +35,93 @@ def split_name(name):
...
+    upload_to handler for Profile.image
+    """
+    return _generate_upload_to_uri()(instance, filename)
+
+
+def profile_image_upload_uri_small(instance, filename):
+    """
+    upload_to handler for Profile.image_small
+    """
+    return _generate_upload_to_uri("_small")(instance, filename)
```
noisecapella: The functions are passed to `upload_to` on the model field. Also, they need to be accessible by `makemigrations` and using `_upload_to` directly confuses it

profiles/util.py
```diff
@@ -29,15 +35,93 @@ def split_name(name):
...
+        BytesIO:
+            A jpeg image which is a thumbnail of full_size_image
+    """
+    in_memory_image = full_size_image
+    pil_image = Image.open(in_memory_image)
+    pil_image.thumbnail(make_small_dimensions(pil_image.width, pil_image.height), Image.ANTIALIAS)
+    image_small_buffer = BytesIO()
+    pil_image.save(image_small_buffer, "JPEG", quality=90)
+    image_small_buffer.seek(0)
+    in_memory_image.seek(0)
```
noisecapella: It looks like we don't, I'll remove it

profiles/util_test.py
noisecapella: Sure

profiles/views_test.py
noisecapella: A patch of a readonly field is ignored with a 200. It does not cause a validation error

noisecapella:
Comments addressed

giocalitri:
did you push the new code? I still see some old stuff..

noisecapella:
Ah, I knew I forgot something

profiles/views_test.py
```diff
@@ -397,3 +401,101 @@ def test_upload_image(self):
...
+        resp = self.client.patch(self.url1, content_type="application/json", data=json.dumps(patch_data))
+        assert resp.status_code == 200
+
+        profile.refresh_from_db()
+        assert bool(profile.image) == image_already_exists
+        assert bool(profile.image_small) == thumb_already_exists
+
+    @ddt.data(
+        [True, False],
+        [True, False]
```
giocalitri: This is twice the same test: am I wrong?

noisecapella: You're right. This was meant to be every combination of true, false for the two arguments. I fixed this for the other test but I overlooked this one

profiles/models_test.py
```diff
@@ -37,16 +46,36 @@ def test_student_id_doesnt_change_on_save(self):
...
+    """
+    Tests for image fields
+    """
+    def test_upload_to(self):
+        """
+        Image upload_to should have a function which creates a path
+        """
+        # pin the timestamps used in creating the URL
+        with patch('profiles.util.datetime', autospec=True) as mocked_datetime:
+            mocked_datetime.now = Mock(return_value=datetime.now(tz=pytz.UTC))
```
giocalitri: you do not need to do this: by default the `new_callable` in a `patch` is a `MagicMock`, so you can simply define the return value for the function you want.
`mocked_datetime.now.return_value = datetime.now(tz=pytz.UTC)`

giocalitri:
this can be merged 👍

PR #2149, opened 2016-12-16
Add more Google Analytics events

### LINKED ISSUE #1919: Send a Google Analytics event when the user profile is complete
Having a unique tracking event for profile completion will allow us to detect users who start the profile process but don't complete. 

In addition, the marketing team would like to track a goal in Google Analytics that's tied to the user completing their profile. They can use `/dashboard` but that's less than ideal because it will be hit multiple times by the same user. 

We can add a simple event corresponding to the URL `/profile/complete` that fires when the user clicks the "I'm Done" button on `/profile/professional` and their form is successfully validated. 

However, we shouldn't put the time in to do this until we know someone will have the time to do the analysis. 

Part of #132

### LINKED ISSUE #2040: Collect more GA events on signup
We should collect more GA events to provide a more detailed funnel

aliceriot:

closes #1919 
closes #2040 

#### What's this PR do?

This adds a couple of new google analytics events, adding events for profile completion (clicking "I'm done") and for completing each form field.

You'll have to make a google analytics account and then set the `GA_TRACKING_ID` variable to the acocunt it. Then go through filling out the profile, and make sure that events show up which reference the form fields and have your username in the 'label' field.

### DIFF
--- a/static/js/components/ProfileProgressControls.js
+++ b/static/js/components/ProfileProgressControls.js
@@ -1,4 +1,5 @@
 // @flow
+/* global SETTINGS:false */
 import React from 'react';
 
 import SpinnerButton from './SpinnerButton';
@@ -7,6 +8,7 @@ import { FETCH_PROCESSING } from '../actions';
 import type { Profile, SaveProfileFunc } from '../flow/profileTypes';
 import type { UIState } from '../reducers/ui';
 import type { Validator } from '../lib/validation/profile';
+import { sendGAEvent } from '../lib/google_analytics';
 
 export default class ProfileProgressControls extends React.Component {
   props: {
@@ -40,6 +42,10 @@ export default class ProfileProgressControls extends React.Component {
     saveProfileStep.call(this, validator, isLastTab).then(() => {
       if (programIdForEnrollment && addProgramEnrollment) {
         addProgramEnrollment(programIdForEnrollment);
+      }
+
+      if ( isLastTab ) {
+        sendGAEvent('profile-form', 'completed', SETTINGS.user.username);
       }
       this.context.router.push(nextUrl);
     });
--- a/static/js/lib/google_analytics.js
+++ b/static/js/lib/google_analytics.js
@@ -1,0 +1,30 @@
+// @flow
+/* global SETTINGS:false */
+import ga from 'react-ga';
+import R from 'ramda';
+
+const makeGAEvent = (category, action, label, value) => ({
+  category: category,
+  action: action,
+  label: label,
+  value: value
+});
+
+const removeNilValue = R.when(
+  R.compose(R.isNil, R.prop('value')),
+  R.dissoc('value')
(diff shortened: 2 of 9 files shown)


static/js/components/ProfileProgressControls.js
```diff
@@ -40,6 +42,10 @@ export default class ProfileProgressControls extends React.Component {
       if (programIdForEnrollment && addProgramEnrollment) {
         addProgramEnrollment(programIdForEnrollment);
       }
+
+      if ( isLastTab ) {
+        sendGAEvent('profile-form', 'completed', SETTINGS.user.username);
```
singingwolfboy: Could we ever reach this point as an anonymous user? i.e. where `SETTINGS.user` will be `null`?

static/js/lib/google_analytics.js
```diff
@@ -0,0 +1,30 @@
...
+export const sendGAEvent = (category: string, action: string, label: string, value?: number) => {
+  ga.event(formatGAEvent(category, action, label, value));
+};
+
+const formatKeyset = R.join("-");
+
+export const sendFormFieldEvent = (keySet: Array<string>) => sendGAEvent(
+  'profile-form-field',
+  `completed-${formatKeyset(keySet)}`,
+  SETTINGS.user.username
```
singingwolfboy: Could this function ever be called as an anonymous user? i.e. where `SETTINGS.user` will be `null`?

aliceriot: No, this is called only when filling out the profile form, which unauthenticated users can't get to (same for your comment above).

pdpinch:
@singingwolfboy are you reviewing this? If so, please assign yourself.

singingwolfboy:
@pdpinch: I was not intending to do a thorough review of this pull request, I just wanted to make sure that this issue was addressed, because I made the change to serialist the user information to SETTINGS.user. In the future, is there a good way to make it more explicit that a comment or two is not intended to be a comprehensive review?

pdpinch:
You can just leave a comment.

static/js/util/profile_edit.js
```diff
@@ -159,6 +162,7 @@ export function boundDateField(keySet: string[], label: string, omitDay: boolean
   let onBlur = () => {
     updateValidationVisibility(keySet);
     updateProfileValidation(profile, validator);
+    sendFormFieldEvent(keySet);
```
rhysyngsun: It might be worth adding tests to cover these calls, although this PR raises the overall coverage, the addition of these lines reduces coverage in this file by ~2%.

aliceriot: I'm more interested in how useful the test is than in the coverage number, but I agree it would be good to add them here.

aliceriot:
@rhysyngsun I added some tests!

rhysyngsun:
:+1:

PR #2152, opened 2016-12-16
Fix a deletion bug with the profile work history

### LINKED ISSUE #2150: bug in the work history deletion code
I found a bug, which I'm having a little trouble repro-ing, but basically when I clicked delete on the 2nd work history entry it actually ended up deleting the first one. I believe I understand the code underlying the problem.

aliceriot:

closes #2150 

#### What's this PR do?

For some reason we weren't passing an `index` parameter to the function to delete a work history entry, which meant that, when we called the function, the value was undefined. With that value we do something more-or-less like:

```js
const deleteEntry = index => {
  profile.work_history.splice(index, 1);
};
```

turns out that if you pass `undefined` as the first argument to `Array.prototype.splice` JavaScript will helpfully assume that you mean zero. This meant that it was possible to click the second work history entry and end up deleting the first one.

Anyway, this fixes the problem by 1) passing in the right paramter and 2) getting rid of some indirection which was originally implemented because of the user page / profile tab split but which has since become unnecessary.

Add a work history entry, and make sure that clicking the second entry in the list does not delete the first entry.

I guess also try to make sure you can repro the bug on master, to ensure it's really being fixed.


static/js/components/EmploymentForm.js
```diff
@@ -205,7 +226,7 @@ class EmploymentForm extends ProfileFormFields {
     let endDateText = () => (
       _.isEmpty(position.end_date) ? "Current" : formatMonthDate(position.end_date)
     );
-    let deleteEntry = () => this.openWorkDeleteDialog();
+    let deleteEntry = () => this.openWorkDeleteDialog(index);
```
aliceriot: this is the change which fixed the bug

noisecapella:
Functionality looks good, looking at code...

static/js/containers/UserPage_test.js
```diff
@@ -666,7 +666,9 @@ describe("UserPage", function() {
         return renderComponent(`/learner/${username}`, userActions).then(([, div]) => {
           let updatedProfile = _.cloneDeep(USER_PROFILE_RESPONSE);
           updatedProfile.username = SETTINGS.user.username;
-          updatedProfile.work_history.splice(0,1);
+          // the second entry will be the first one display in the UI
+          // (work history entries are sorted by date before they are rendered)
+          updatedProfile.work_history.splice(1,1);
```
noisecapella: Can we programatically get the index here instead of using a hardcoded value?

aliceriot: I suppose we could, I could try to get the company name from the UI and look for that in the entries.

static/js/components/EmploymentForm.js
```diff
@@ -65,15 +61,38 @@ class EmploymentForm extends ProfileFormFields {
...
+    setWorkDialogIndex(index);
+    setWorkDialogVisibility(true);
   };
 
-  deleteWorkHistoryEntry: Function = (): void => {
-    return deleteWorkHistoryEntry.call(this);
+  deleteWorkHistoryEntry: Function = (): Promise<*> => {
+    const { saveProfile, profile, ui } = this.props;
+    let clone = _.cloneDeep(profile);
+    if ( ui.deletionIndex ) {
```
noisecapella: What if `ui.deletionIndex` is `0`?

aliceriot: ah good point

static/js/components/EmploymentForm.js
```diff
@@ -65,15 +61,38 @@ class EmploymentForm extends ProfileFormFields {
...
+      setWorkDialogIndex,
+      setWorkDialogVisibility,
+      validator,
+    } = this.props;
+    let clone = {
+      ...profile,
+      work_history: [...profile.work_history, generateNewWorkHistory()]
+    };
+    updateProfile(clone, validator, true);
+    setWorkDialogIndex(clone.work_history.length - 1);
```
aliceriot: sure, I just copy and pasted what was already there, happy to do a little refactoring now

aliceriot:
comments addressed

PR #2159, opened 2016-12-19
delete checked in JavaScript dependencies

aliceriot:
I'm not sure why these were added in the first place, they are totally superfluous (both bootstrap and jQuery are in `package.json`).

Anyway, this deletes two checked-in minified JS files that were added to `static/js/financial_aid`.

Verify that you can still visit the financial aid review page (`/financial_aid/review/${programID}`) and that everything looks the same.


noisecapella:
Can you delete `.eslintignore` too? It seems like it's only there for those two files haha

noisecapella:
We should also remove the jquery and bootstrap JS files linked to in `base_error.html`, maybe in another PR

aliceriot:
lol I deleted `.eslintignore`. 

I'm down to remove those script tags here, bootstrap is sourced in `entry/public.js` and jQuery in `entry/public_style.js`, which should both already be pulled in on the error page, right?

noisecapella:
The error page extends `base.html` so if the review template has it, the error page should too. Which bundle provides these?

aliceriot:
^^ see my comment above

aliceriot:
jquery and bootstrap are `require`'ed in the bundle for the financial aid page.

aliceriot:
`base_error` looks fine with the script tags deleted.

Issue #2135, opened 2016-12-15
Profile: replace country/state/city fields with address field


pdpinch:
Part of #1796 and #1821 

Currently we collect Country of Residence, State, and City. We will replace this with one field labelled “Current Address”. This field will have hint text “Ex. 100 Main Street, Anytown, 01234, United States of America”. This field will use the Google Places API, which basically allows the user to type their address, then select it from an autocomplete. It will accept the text entry and display results in UTF-8 character set, but on the back end will submit the Street Address, City, Region/State, Postal Code and Country in CP-1252 (provided by the Google Places API.

![image](https://cloud.githubusercontent.com/assets/430126/21227532/7c671d6a-c2a8-11e6-8ee4-8d9ef510cf02.png)

### Acceptance criteria

- [ ] Must work in both initial profile creation (/profile/personal) and in profile editing (/learner)
- [ ] Populates the following fields on `Profile`: `address1`, `city`, `state_or_territory`, `postal_code`, `country`.
  - Optionally, since each address line is limited to 40 characters, `address2` and `address3` are available for overflow.

rhysyngsun:
A few tips based on my research:

- Specify to the Google API that we want output in English.
- Specify that the API returns a precise address (https://developers.google.com/places/web-service/autocomplete#place_types)

noisecapella:
We should have some language telling the user what we're using the address for, and we also need to tell them which parts of their address we're sharing in their public profile

pdpinch:
@singingwolfboy this is In Progress now, right? I'll move it.

pdpinch:
Pricing details are at https://developers.google.com/maps/pricing-and-plans/

Since registration is free for the portal, I believe we fall under the Standard plan. If we start restricting access based on payment (as opposed to enrollment) then that could change. 

For production, we'll getan API key tied to a credit card so we can go over the rate limits if/when necessary.

PR #2161, opened 2016-12-19
Clean up how API keys are passed into templates

singingwolfboy:
We're currently using two different Javascript APIs that require API keys (Google Analytics and Smartlook), and with #2135 it looks like we'll be adding a third (Google Maps, required to use the Google Places Javascript API). This pull request tries to clean up our existing API usage, to make it easier and more understandable to add more API keys in the future.

Note that this pull request does make one important change to the behavior of the website. Previously, the Google Analytics Javascript would be loaded regardless of whether a tracking ID was set in an environment variable. This pull request now checks to see if we have a tracking ID, and only loads the Google Analytics Javascript if that tracking ID is defined.


amir-qayyum-khan:
looks good, is their any view test of this object APIKEYS?

amir-qayyum-khan:
👍

singingwolfboy:
@amir-qayyum-khan no, we don't seem to have any tests for template contexts, and I think that's OK.

noisecapella:
@singingwolfboy Please delete your branches after you merge the PR!

singingwolfboy:
I need to find a Chrome extension or something that will do that for me, automatically...

singingwolfboy:
This looks promising: [Auto branch delete for Github](https://chrome.google.com/webstore/detail/auto-branch-delete-for-gi/cadicbdpppmnmfaepeiolnoeomkoickn). Source is here: https://github.com/robin-drexler/github-auto-delete-branch-chrome-extension A little rough around the edges, but it'll do.

PR #2166, opened 2016-12-19
Updated redirect to preserve request URI

blarghmatey:
Added `$request_uri` suffix to Nginx redirect.

N/A

#### What's this PR do?
Fixes regression in redirects from micromasters.herokuapp.com where it is no longer preserving the request URI

config/nginx.conf.erb

Use Charles Proxy to map the PR build to micromasters.herokuapp.com and see if the redirect maintains the path.

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -1,6 +1,12 @@
 # This is the version used in development environments
 server {
     listen 8079;
+    server_name micromasters.herokuapp.com;
+    return 301 https://micromasters.mit.edu$request_uri;
+}
+
+server {
+    listen 8079 default_server;
     root /src;
 
     location / {
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -18,7 +18,20 @@ http {
     server_tokens off;
 
     log_format l2met 'measure#nginx.service=$request_time request_id=$http_x_request_id';
-    access_log logs/nginx/access.log l2met;
+    log_format apm '"$time_local" client=$remote_addr '
+                   'method=$request_method request="$request" '
+                   'request_length=$request_length '
+                   'status=$status bytes_sent=$bytes_sent '
+                   'body_bytes_sent=$body_bytes_sent '
+                   'referer=$http_referer '
+                   'user_agent="$http_user_agent" '
+                   'upstream_addr=$upstream_addr '
+                   'upstream_status=$upstream_status '
+                   'request_time=$request_time '
+                   'upstream_response_time=$upstream_response_time '
+                   'upstream_connect_time=$upstream_connect_time '
+                   'upstream_header_time=$upstream_header_time';
+    access_log logs/nginx/access.log apm;
     error_log logs/nginx/error.log;
 
     include mime.types;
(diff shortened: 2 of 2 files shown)


config/nginx.conf.erb
```diff
@@ -28,13 +28,16 @@ http {
...
-        if ($host = 'micromasters.herokuapp.com') {
-            return 301 'https://micromasters.mit.edu';
-        }
+    server {
+        listen <%= ENV["PORT"] %>;
+        server_name micromasters.mit.edu;
+        root /app;
 
         location ^~ /static/(.*$) {
+            expires max;
```
rhysyngsun: Is this `expires max;` change intended to be part of this redirect fix?

config/nginx.conf.erb
```diff
@@ -27,14 +40,17 @@ http {
...
+        listen <%= ENV["PORT"] %>;
         server_name _;
         root /app;
 
-        if ($host = 'micromasters.herokuapp.com') {
-            return 301 'https://micromasters.mit.edu';
-        }
-
         location ^~ /static/(.*$) {
+            expires max;
```
giocalitri: what is this?

config/nginx.conf.erb
```diff
@@ -18,7 +18,20 @@ http {
...
+                   'body_bytes_sent=$body_bytes_sent '
+                   'referer=$http_referer '
+                   'user_agent="$http_user_agent" '
+                   'upstream_addr=$upstream_addr '
+                   'upstream_status=$upstream_status '
+                   'request_time=$request_time '
+                   'upstream_response_time=$upstream_response_time '
+                   'upstream_connect_time=$upstream_connect_time '
+                   'upstream_header_time=$upstream_header_time';
+    access_log logs/nginx/access.log apm;
```
giocalitri: do we need to change the logs also in dev?

pdpinch:
What's the best way to test this? I was going to try to figure out the IP for the PR build and then do some hacking of my hosts file. But I'm getting a 404 from the PR build?

config/nginx.conf.erb
blarghmatey: That will set the expires and cache-control headers for static assets served through Nginx.

config/nginx.conf.erb
blarghmatey: It's not required, but we can if it would be valuable for developers to have that extra info. Most of that detail is already captured in the default Nginx logs for that container.

blarghmatey:
@pdpinch the 404 has been resolved. The best way to test it is with Charles Proxy and the [Map Remote](https://www.charlesproxy.com/documentation/tools/map-remote/) tool.

config/nginx.conf.erb
```diff
@@ -27,15 +40,20 @@ http {
...
 
-        if ($host = 'micromasters.herokuapp.com') {
-            return 301 'https://micromasters.mit.edu';
-        }
-
-        location ^~ /static/(.*$) {
-            try_files $uri $uri/ staticfiles/$1$args staticfiles/$1/$args =404;
+        location ~* /static/(.*$) {
+            expires max;
+            try_files $uri $uri/ staticfiles/$1 staticfiles/$1/ =404;
```
noisecapella: FYI I found that I needed `staticfiles` to start with a slash, `/staticfiles/$1` for example

config/nginx.conf
```diff
@@ -1,6 +1,12 @@
 # This is the version used in development environments
 server {
     listen 8079;
+    server_name micromasters.herokuapp.com;
+    return 301 https://micromasters.mit.edu$request_uri;
+}
+
+server {
+    listen 8079 default_server;
```
giocalitri: is still good after @noisecapella changes to this file?

config/nginx.conf.erb
```diff
@@ -27,14 +40,17 @@ http {
...
+        listen <%= ENV["PORT"] %> default_server;
         server_name _;
         root /app;
 
-        if ($host = 'micromasters.herokuapp.com') {
-            return 301 'https://micromasters.mit.edu';
-        }
-
         location ~* /static/(.*$) {
+            expires max;
```
giocalitri: is this still valid?

config/nginx.conf
noisecapella: I'm not sure these changes are necessary in this file since this file is only for dev environments

config/nginx.conf.erb
blarghmatey: Yes, that is still valid.

config/nginx.conf
blarghmatey: Correct on both points. I had added it in for testing locally. It doesn't hurt to leave it in, but it's not required for dev environments.

PR #2167, opened 2016-12-19
Address field with Google Places Autocomplete

### LINKED ISSUE #2135: Profile: replace country/state/city fields with address field

singingwolfboy:
Fixes #2135. Replaces the "country", "state", and "city" fields for the user's current residence with a single "current address" field. This field has an autocomplete that is powered by the [Google Places API](https://developers.google.com/places/). This field is implemented using the [React-Geosuggest](https://github.com/ubilabs/react-geosuggest) component.

### Setup
In order to test this pull request, you'll need to get an API key from Google. Start by going to the [Google API Console](https://console.developers.google.com), sign in with a Google account, and create a project if you don't already have one. Next, go to the "Credentials" section, and click on the "Create Credentials" button. You want to create an API key, not an OAuth client key or a service account key. Google will give you a key, which is a long string of random letters and numbers. Put this into your `.env` file, like this:
```ini
GOOGLE_API_KEY=mY-API-kEy-123
```

Next, you need to add permissions to your API key. In the Google API Console, go to the "Library" section, and enable the following three APIs: [Google Maps Geocoding API](https://developers.google.com/maps/documentation/geocoding/start), [Google Places API Web Service](https://developers.google.com/places/web-service/), and [Google Maps Javascript API](https://developers.google.com/maps/documentation/javascript/). For each one, type in the name in the search box, click on the result, and click the "Enable" button at the top of the page (next to the API's name).

Now you should be able to successfully use this API key for the Google Places autocomplete!

### Screenshots!
Placeholder text:
placeholder text

Autocomplete:
autocomplete

Autocomplete item selected:
autocomplete item selected

Text is automatically updated:
updated text


roberthouse54:
Looks good, except why is the "Where are you from" header sticking out to the left?

singingwolfboy:
It's because it's a header. That's the way it currently looks on production -- it's unrelated to this pull request.

static/scss/geosuggest.scss
```diff
@@ -0,0 +1,69 @@
+// https://github.com/ubilabs/react-geosuggest/blob/master/src/geosuggest.css
```
aliceriot: is this file copied from a library? if so, you should be able to just `require` it, in the way we do most of the rest of the library CSS (see `static/js/entry/style.js` for an example).

static/js/util/profile_edit.js
```diff
@@ -214,6 +215,53 @@ export function boundCheckbox(keySet: string[], label: string|React$Element<*>):
...
+      <Geosuggest initialValue={initial}
+        id={id} label={label} placeholder={placeholder} types={types}
+        onSuggestSelect={onSuggestSelect}
+      />
+      <span className="validation-error-text">
+        {_.get(errors, keySet)}
+      </span>
+    </div>
+  );
+}
```
aliceriot: A couple things in this function could probably be pulled out as little helper methods - if it's feasible to do that I think it's a good idea (especially for things like `addressComponentTypeMap`) because otherwise a bunch of things are being unnecessarily redeclared on every function call.

static/scss/geosuggest.scss
singingwolfboy: I actually already tried importing the file from the library in `static/js/entry/style.js`, but then discovered that we needed to make some changes to the styling to make it resemble the Material UI input fields. I figured that copying the file into the repo and making changes there was the best solution, but we could also import the default styles and then override them as necessary. Do you have a preference?

aliceriot: I'd prefer if we imported it and just overrode what we need to

pdpinch:
I've created a development and a production key, tied to our mitx devops google account. Contact me or Tobias to get the API key.

static/js/components/PersonalForm.js
```diff
@@ -74,34 +79,17 @@ export default class PersonalForm extends ProfileFormFields {
               {...this.defaultInputComponentProps()}
             />
           </Cell>
+          <Cell col={12}>
+            {this.boundGeosuggest(addressMapping, "Current address",
+              {
+                id: "current-home",
+                placeholder: "Example: 100 Main Street, Anytown, 01234, United States",
+                types: ["geocode"]
```
rhysyngsun: Small comment: I think this should be `address` instead of `geocode`. Reason being we need to be sure that we are guaranteed a precise street address. I'm fairly certain in my early testing that `geocode` accepted addresses like "Boston, MA". Worth verifying again though.

singingwolfboy: [According to Google's documentation,](https://developers.google.com/maps/documentation/javascript/reference#AutocompletionRequest) `geocode` is the right value to use for addresses.

static/js/util/profile_edit_test.js
```diff
@@ -556,6 +558,54 @@ describe('Profile Editing utility functions', () => {
     });
   });
 
+  describe.only('bound Geosuggest', () => {
```
noisecapella: Be careful of `.only` here, we don't currently check for this during CI

static/js/components/PersonalForm.js
rhysyngsun: Yup, appears I was looking at a different set of docs for a similar API on a different google service.

static/js/util/profile_edit_test.js
aliceriot: now we do! rebase!

static/js/util/profile_edit.js
```diff
@@ -220,6 +223,106 @@ export function boundCheckbox(keySet: string[], label: string|React$Element<*>):
...
+    hasExistingAddress,
+    R.compose(R.join(", "), R.props(R.values(addressMapping))),
+    R.always(""),
+  );
+
+  const initial = formatInitialAddress(profile);
+
+  const hasErrors = (errors) => (
+    R.any(R.pathSatisfies(hasValue, R.__, errors), R.values(addressMapping))
+  );
```
aliceriot: this is pretty similar to `hasAddressProps` above, maybe the two could be rewritten to reuse some code, something like

```js
const _addressPropCheck = R.flip(R.pathSatisifes(hasValue, R.__));

const addressPropCheck = checkFunc => (
  R.compose(checkFunc(R.__, R.values(addressMapping)), _addressPropCheck)
);

const hasAddressProps = addressPropCheck(R.all);

const hasErrors = addressPropCheck(R.any);
```

I think that works! haha.

static/js/util/test_utils.js
```diff
@@ -159,6 +160,74 @@ export const localStorageMock = (init: any = {}) => {
...
+    };
+
+    window.google = google;
+  }
+
+  cleanup() {
+    this.sandbox.restore();
+    delete window.google;
+  }
+}
```
singingwolfboy: My concern is that most of our pages do _not_ load the Google Maps API, but it's a common-enough library that other libraries may detect it and work differently depending on if the library is loaded or not. I think it's worthwhile to test _without_ a `google` variable in the global context, to more closely simulate how most of our pages will act.

That being said, this is a very minor concern, and I rather doubt that any of the libraries we're using will detect the presence of the `google` variable. I'm willing to be overruled on this if you think that the reduced boilerplate code is worth the imperfect testing environment.

PR #2172, opened 2016-12-20
Fix static asset handling

noisecapella:
None
#### What's this PR do?
- Updates nginx config to properly handle static assets for production, and removes static asset handling for dev, where Django will handle all static assets
- Removes `static-map` directive for `uwsgi`

When we turned on Cloudfront `STATIC_URL` was changed to an absolute URL from a relative one. This caused Django not to serve static assets anymore, and our nginx config wasn't matching static URLs properly. With this change nginx should handle all static file handling for production.

In the PR build go to https://micromasters-ci-pr-2172.herokuapp.com/cms/. You should see a broken login page and you should see cloudfront URLs for JS static assets in script tags, starting with `FAKE.`

In the PR build go to https://micromasters-ci-pr-2172.herokuapp.com/static/wagtailadmin/js/vendor/jquery-2.2.1.js. You should see a valid JS file and not a 404 error.

Also make sure your dev environment is still functioning properly, including Django Debug toolbar and `/cms/`

### DIFF
--- a/uwsgi.ini
+++ b/uwsgi.ini
@@ -25,7 +25,7 @@ pidfile=/tmp/micromasters-mast.pid
 vacuum=True
 enable-threads = true
 single-interpreter = true
-static-map = /static=./static
+static-map = /static=./staticfiles
 offload-threads = 2
 thunder-lock =
 if-env = HTTP_AUTH_CREDENTIALS
--- a/uwsgi.ini
+++ b/uwsgi.ini
@@ -25,6 +25,7 @@ pidfile=/tmp/micromasters-mast.pid
 vacuum=True
 enable-threads = true
 single-interpreter = true
+static-map = /static=./staticfiles
 offload-threads = 2
 thunder-lock =
 if-env = HTTP_AUTH_CREDENTIALS
--- a/uwsgi.ini
+++ b/uwsgi.ini
@@ -25,7 +25,6 @@ pidfile=/tmp/micromasters-mast.pid
 vacuum=True
 enable-threads = true
 single-interpreter = true
-static-map = /static=./staticfiles
 offload-threads = 2
 thunder-lock =
 if-env = HTTP_AUTH_CREDENTIALS
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -9,15 +9,4 @@ server {
         uwsgi_pass_request_headers on;
         uwsgi_pass_request_body on;
     }
-
-    location ~* /static/(.*$) {
-        try_files $uri $uri/ staticfiles/$1 staticfiles/$1/ =404;
-    }
-
-    location /static/debug_toolbar/ {
-        include uwsgi_params;
-        uwsgi_pass web:8077;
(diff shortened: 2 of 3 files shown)


uwsgi.ini
```diff
@@ -25,7 +25,7 @@ pidfile=/tmp/micromasters-mast.pid
 vacuum=True
 enable-threads = true
 single-interpreter = true
-static-map = /static=./static
+static-map = /static=./staticfiles
```
blarghmatey: I don't think this is necessarily the proper fix. We can actually remove this line entirely now that Nginx is serving the static assets. The static-map is only a best effort setting. http://uwsgi-docs.readthedocs.io/en/latest/StaticFiles.html#mode-3-using-static-file-mount-points

In the current configuration this shouldn't even be doing anything.

giocalitri: I still would like to be able to use UWSGI directly

giocalitri:
I cannot understand what kind of error I should see in the current master that this PR addresses

noisecapella:
@giocalitri 
go to
https://micromasters-rc.herokuapp.com/static/wagtailadmin/js/vendor/jquery-2.2.1.js

Then go to 
https://micromasters-ci-pr-2172.herokuapp.com/static/wagtailadmin/js/vendor/jquery-2.2.1.js

It's broken on RC because the `CLOUDFRONT_DIST` environment variable is set there, which is changing `STATIC_URL`

giocalitri:
why are you talking about the `/cms/` in the PR description?

giocalitri:
anyway the test you suggest works, so 👍

noisecapella:
Going to `/cms/` tests that the wagtail admin static assets are loading correctly

Anyway I think I screwed up the test. I didn't actually set `CLOUDFRONT_DIST` on the PR build, so more work needed

giocalitri:
> Going to `/cms/` tests that the wagtail admin static assets are loading correctly

But this currently works fine in my current dev environment

noisecapella:
@giocalitri If you set `CLOUDFRONT_DIST` to something then go to `/cms/`, you should see the problem

pdpinch:
This doesn't look right for `/cms` and `/admin` for me, but maybe that's because of how cloudfront is configured?

screenshot 2016-12-21 15 24 46
screenshot 2016-12-21 15 24 34

noisecapella:
@pdpinch Yes, `CLOUDFRONT_DIST` is defined with a fake distribution id. This causes `STATIC_URL` to be an absolute URL which means that it turns off Django's static file handling. But you can still go to the static files if you type in the URL manually, which means that nginx is serving them correctly.

noisecapella:
If you take one of those static URLs and use it with `micromasters-rc` where this bug is not fixed and `CLOUDFRONT_DIST` is set, for example https://micromasters-rc.herokuapp.com/static/wagtailadmin/js/vendor/jquery-2.2.1.js you should see a 404 error

giocalitri:
I think this works fine 👍

PR #2174, opened 2016-12-20
Remove static and webpack-related config

noisecapella:
None

#### What's this PR do?
Removes some config from the nginx config files. It seems like nginx should not know about webpack at all, and it seems like uwsgi is handling references to `/static/`. @blarghmatey can you take a look and see if this makes sense?

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -10,18 +10,7 @@ server {
         uwsgi_pass_request_body on;
     }
 
-    location @webpack {
-        proxy_pass http://watch:8078;
-    }
-
     location ^~ /static/(.*$) {
-        try_files $uri $uri/ staticfiles/$1 staticfiles/$1/ @webpack =404;
-    }
-
-    location /static/debug_toolbar/ {
-        include uwsgi_params;
-        uwsgi_pass web:8077;
-        uwsgi_pass_request_headers on;
-        uwsgi_pass_request_body on;
+        try_files $uri $uri/ staticfiles/$1 staticfiles/$1/ =404;
     }
 }
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -9,8 +9,4 @@ server {
         uwsgi_pass_request_headers on;
         uwsgi_pass_request_body on;
     }
-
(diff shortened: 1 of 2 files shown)


config/nginx.conf.erb
```diff
@@ -34,10 +34,6 @@ http {
             return 301 'https://micromasters.mit.edu';
         }
 
-        location ^~ /static/(.*$) {
-            try_files $uri $uri/ staticfiles/$1$args staticfiles/$1/$args =404;
-        }
-
```
blarghmatey: We don't want to remove this because having Nginx serve the static assets is much more efficient than delegating that to uWSGI or Django.

config/nginx.conf
```diff
@@ -9,19 +9,4 @@ server {
         uwsgi_pass_request_headers on;
         uwsgi_pass_request_body on;
     }
-
-    location @webpack {
-        proxy_pass http://watch:8078;
-    }
-
```
blarghmatey: This can be removed if all assets that would be served by webpack are addressed directly in the HTML.

config/nginx.conf
```diff
@@ -9,19 +9,4 @@ server {
...
-    location ^~ /static/(.*$) {
-        try_files $uri $uri/ staticfiles/$1 staticfiles/$1/ @webpack =404;
```
blarghmatey: If we are going to remove the webpack named location then the `@webpack` reference can be deleted but it makes sense to continue using Nginx to serve the static assets so that it is more akin to production.

config/nginx.conf
```diff
@@ -9,19 +9,4 @@ server {
...
-    }
-
-    location /static/debug_toolbar/ {
-        include uwsgi_params;
-        uwsgi_pass web:8077;
-        uwsgi_pass_request_headers on;
-        uwsgi_pass_request_body on;
-    }
```
blarghmatey: If we leave the `/static` location then it is necessary to have the debug_toolbar special case, otherwise all assets under that path will return a 404.

noisecapella:
Going to do this in the other PR

Issue #2163, opened 2016-12-19
Investigate slow page loads


pdpinch:
get a snapshot of performance now and then look into why it is so slow. 

- [ ] investigate our javascript bundle: how much is actually needed for each page?
- [ ] investigate chrome loading to see where the hot spots are
- [ ] consider webpack2

aliceriot:
I wrote a patch which can be used to get a production environment running locally (minified builds and so on). This is more useful for performance testing and getting relevant numbers than a development build - the patch is available here: https://github.com/aliceriot/mm_patches/blob/master/mock_production_mode.patch

I'm trying to use this to get some baseline numbers about our application's performance (mainly looking at the dashboard code right now) using the chrome 'Timeline' developer feature.

It looks like, on the dashboard, all of our JavaScript and CSS is done loading by about 700ms or so, and then we get first paint (anything rendered on the screen) 1740ms, which is a pretty long time. The interim looks to be mainly occupied with evaluating the `dashboard.js` bundle.

One thing we might try is these two webpack plugins:

https://babeljs.io/docs/plugins/transform-react-inline-elements/
https://babeljs.io/docs/plugins/transform-react-constant-elements/

I'm going to install them and see if there's a difference.

aliceriot:
I wanted to get some more reliable numbers for the time until the document is
loaded. Chrome makes it easy to get this information in the timeline tool.

on master, with just the `mock-production` patch applied:

| time to 'loaded' event (seconds) |
| ---------------------------------|
|2.5|
|2.75|
|2.37|
|2.43|
|2.5|
|2.83|
|2.58|
|3.04|
|2.39|
|2.55|
| average: 2.59 |

This gives us a baseline we can compare things to.

aliceriot:
First I wanted to investigate the effect of adding the two babel transforms I mentioned above. Here are some load times, on a branch where those plugins are added and configured, and the `mock-production` patch has been applied:

| time to 'loaded' event (seconds) |
| ------------------------------------------- |
| 2.99 |
| 2.71 |
| 2.85 |
| 2.86 |
| 2.93 |
| 3.04 |
| 2.46 |
| 2.57 | 
| 2.51 |
| 2.53 |
| average: 2.745 |

so that's not an improvement! darn.

aliceriot:
time to `loaded` event with chrome's 'low end device' throttling enabled:

| | master | babel react optimizations |
| -- | ---------------------------------- | -- |
| | 6.84 | 6.09 |
[...]

PR #2175, opened 2016-12-20
Add two new babel plugins

aliceriot:
This adds two babel plugins which are supposed to optimize React a little bit. Look at the relevant issue: https://github.com/mitodl/micromasters/issues/2163 to see some results comparing this branch with `master`.

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -1,8 +1,8 @@
 var webpack = require('webpack');
 var path = require("path");
-var sharedConfig = require(path.resolve("./webpack.config.shared.js"));
 var R = require('ramda');
 var BundleTracker = require('webpack-bundle-tracker');
+const { config, babelSharedLoader } = require(path.resolve("./webpack.config.shared.js"));
 
 const hotEntry = (host, port) => (
   `webpack-hot-middleware/client?path=http://${host}:${port}/__webpack_hmr&timeout=20000&reload=true`
@@ -12,15 +12,12 @@ const insertHotReload = (host, port, entries) => (
   R.map(R.compose(R.flatten, v => [v].concat(hotEntry(host, port))), entries)
 );
 
-const devConfig = {
+Object.assign(config, {
   context: __dirname,
   output: {
     path: path.resolve('./static/bundles/'),
     filename: "[name].js"
   },
-  module: sharedConfig.module,
-  sassLoader: sharedConfig.sassLoader,
-  resolve: sharedConfig.resolve,
   plugins: [
     new webpack.DefinePlugin({
       'process.env': {
@@ -37,11 +34,17 @@ const devConfig = {
     new BundleTracker({filename: './webpack-stats.json'})
   ],
   devtool: 'source-map'
-};
+});
+
+const devConfig = config;
+
+devConfig.module.loaders = [
+  babelSharedLoader, ...config.module.loaders
+];
 
 const makeDevConfig = (host, port) => (
   Object.assign({}, devConfig, { 
-    entry: insertHotReload(host, port, sharedConfig.entry),
+    entry: insertHotReload(host, port, devConfig.entry),
   })
 );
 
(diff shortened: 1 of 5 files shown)


aliceriot:
this is spitting out some warnings which are causing codecov to fail, even though none of the tests are failing. investigating.

noisecapella:
"The error output found" message is a bit misleading FYI. It's for the tests themselves, not code coverage, even though it's displayed right after the code coverage message

.babelrc
```diff
@@ -8,5 +8,7 @@
       "react-hot-loader/babel",
       "transform-object-rest-spread",
       "transform-class-properties",
+      "transform-react-constant-elements",
```
noisecapella: The docs say that this should be enabled only in production because it makes warning messages more cryptic. Is there a way to do that cleanly?

aliceriot:
@noisecapella yeah that's what I was trying to explain above. I've noticed it happening a lot since we added codecov.

.babelrc
aliceriot: I think so - we can tell the `babel-loader` webpack plugin to use certain babel plugins (plugins within plugins...)

I had read that in the documentation too, I'll try to get that set up now.

aliceriot:
@noisecapella I made some changes to the webpack configuration that lets us use different babel configurations for dev and prop. lmk what you think!

webpack.config.dev.js
```diff
@@ -12,15 +12,12 @@ const insertHotReload = (host, port, entries) => (
   R.map(R.compose(R.flatten, v => [v].concat(hotEntry(host, port))), entries)
 );
 
-const devConfig = {
+const devConfig = Object.assign(config, {
```
noisecapella: We should only do `x = Object.assign({}, ...)` or `Object.assign(x, ...)`, `y = Object.assign(x, ...)` is confusing because `x` is assigned to directly. This will set `devConfig` equal to `config` here. Also `devConfig.module.loaders` below will be unaffected because this is a shallow clone.

In any case I think it's fine here to just modify `config`, nothing else is going to be using it.

webpack.config.prod.js
```diff
@@ -1,18 +1,24 @@
...
+
+prodBabelConfig.query.plugins.push(
+  "transform-react-constant-elements",
+  "transform-react-inline-elements"
+);
+
+const prodConfig = Object.assign({}, config);
+prodConfig.module.loaders = [prodBabelConfig, ...config.module.loaders];
+
+module.exports = Object.assign(prodConfig, {
```
noisecapella: Same comment here about the `y = Object.assign(x, ...)` syntax

aliceriot: here it's being exported directly though, not bound to a variable.

webpack.config.dev.js
aliceriot: I'm not following what you said about `devConfig.module.loaders` below, do you mean that the `devConfig` loaders won't be affected? They are, otherwise the app wouldn't boot (because `babel-loader` wouldn't have been added).

noisecapella: That's fine with me, can we remove the `Object.assign` then?

aliceriot: I need some way to add the development specific config though, what exactly are you suggesting?

webpack.config.dev.js
```diff
@@ -12,15 +12,12 @@ const insertHotReload = (host, port, entries) => (
...
 
-const devConfig = {
+Object.assign(config, {
   context: __dirname,
   output: {
     path: path.resolve('./static/bundles/'),
     filename: "[name].js"
   },
-  module: sharedConfig.module,
-  sassLoader: sharedConfig.sassLoader,
```
noisecapella: Do you know where this functionality was moved to?

aliceriot: the sassLoader? I refactored the webpack configuration a little while ago to add the commons chunk plugin and this was removed from the shared config then - I had just missed this line here.

aliceriot: I can't remember what I changed that made me not need to add it in the shared config, I think it was that a newer version of the sass loader deprecated that way of doing things or something.

aliceriot:
Ok I think I fixed the warning output, things should be sorted now

aliceriot:
(the build should pass, I mean)

PR #2179, opened 2016-12-20
Added course semester facet

### LINKED ISSUE #1710: Search: Semester facet
image

## Acceptance Criteria:

- [ ] 2nd facet, after course
- [ ] user can only choose one value at a time ("and" facet)
- [ ] add course run to index, if it isn't already. 
- [ ] mapping of runs to labels: 
  1T2016 = "Spring 2016"
  2T2016 = "Summer 2016"
  3T2016 = "Fall 2016"
  etc. 
- [ ] if the run is ambiguous, fail over to using dates. @pdpinch will spec them. 
- [ ] Selecting a semester without selecting a course, will return all students who are enrolled in any course during the selected semester.
- [ ] unlike other facets, we would like to order these facets in reverse chronological order, instead of ordering by hits. e.g. 
  2016 - Fall
  2016 -  Summer
   2016 - Spring
  2015 -Fall
  etc. 
  This list is limited to the first 5. Clicking "see more" would reveal the whole list.

gsidebo:

Closes #1710 

#### What's this PR do?

Adds the course semester facet to the learner's search

dashboard/serializers.py, courses/utils.py, then static/js/components/LearnerSearch.js

1. Seed the DB with your user as staff
1. Recreate index
1. Go to learner's search page

There are a couple hacks in here to get searchkit working properly. I'll add details in the comments.

ss 2016-12-20 at 15 57 27

### DIFF
--- a/courses/utils.py
+++ b/courses/utils.py
@@ -1,0 +1,91 @@
+"""
+Utility functions for the courses app
+"""
+import re
+from opaque_keys.edx.keys import CourseKey
+from opaque_keys import InvalidKeyError
+
+NUMBER_SEASON_MAP = {
+    1: 'Spring',
+    2: 'Summer',
+    3: 'Fall'
+}
+
+
+def parse_edx_key(edx_course_key):
+    """
+    Attempts to parse a CourseRun's edx_course_key as an edX opaque key and return the portion
+    of the key that indicates the course's semester/year
+
+    Args:
+        edx_course_key (str): An edX course key (CourseRun.edx_course_key)
+
+    Returns:
+        str: The 'run' portion of a parsed CourseKey (opaque_keys.edx.keys.CourseKey#run),
+        eg: '1T2016'. Return None if the parse fails.
+    """
+    try:
+        course_key = CourseKey.from_string(edx_course_key)
+    except InvalidKeyError:
+        return None
+    return course_key.run if course_key else None
+
+
+def get_year_season_from_edx_key(parsed_course_key):
+    """
+    Verifies that a parsed course key matches our expected format and returns a year/season
+
+    Args:
+        parsed_course_key (str): A semester/year indicator from a course key (eg: '1T2016')
+
+    Returns:
+        tuple: (Year, Season number)
+    """
+    found = re.search('^[1-3]T[1|2][0-9]{3}$', parsed_course_key)
+    if found:
+        season_year_tuple = found.group().split('T')
+        return int(season_year_tuple[1]), int(season_year_tuple[0])
+    return ()
+
+
+def get_year_season_from_dates(course_run):
+    """
+    Args:
+        course_run (CourseRun): A CourseRun object
+
+    Returns:
+        tuple: (Year, Season number)
+    """
+    if not course_run.start_date:
+        return ()
+    start_month = course_run.start_date.month
+    if 1 <= start_month <= 4:
+        season_num = 1
+    elif 5 <= start_month <= 8:
+        season_num = 2
+    else:
+        season_num = 3
+    return course_run.start_date.year, season_num
+
+
+def get_year_season_from_course_run(course_run):
+    """
+    Attempts to get year/season info from a CourseRun
+
+    Args:
+        course_run (CourseRun): A CourseRun object
+
+    Returns:
+        tuple: (Year, Season string), eg: (2016, 'Spring'). Returns () if year/season can't
+        be determined
+    """
+    parsed_course_key = parse_edx_key(course_run.edx_course_key)
+    season_year_tuple = ()
+    if parsed_course_key:
+        season_year_tuple = get_year_season_from_edx_key(parsed_course_key)
+    if not season_year_tuple:
+        season_year_tuple = get_year_season_from_dates(course_run)
+    if season_year_tuple:
+        return season_year_tuple[0], NUMBER_SEASON_MAP[season_year_tuple[1]]
+    else:
+        return season_year_tuple
--- a/dashboard/serializers_test.py
+++ b/dashboard/serializers_test.py
@@ -2,6 +2,7 @@
 Test cases for the UserProgramSearchSerializer
 """
 from unittest.mock import patch
+from datetime import datetime
 import ddt
 import faker
 import pytz
@@ -110,6 +111,7 @@ class UserProgramSearchSerializerTests(TestCase):
         non_fa_cached_edx_data = CachedEdxUserData(cls.user, program=program)
         non_fa_mmtrack = MMTrack(cls.user, program, non_fa_cached_edx_data)
         cls.serialized_enrollments = UserProgramSearchSerializer.serialize_enrollments(non_fa_mmtrack, cls.enrollments)
+        cls.semester_enrollments = UserProgramSearchSerializer.serialize_semester_enrollments(cls.enrollments)
         cls.program_enrollment = ProgramEnrollment.objects.create(user=cls.user, program=program)
         # create a financial aid program
         cls.fa_program, _ = create_program()
@@ -135,6 +137,7 @@ class UserProgramSearchSerializerTests(TestCase):
         cls.fa_serialized_enrollments = (
             UserProgramSearchSerializer.serialize_enrollments(fa_mmtrack, cls.fa_enrollments)
         )
+        cls.fa_semester_enrollments = UserProgramSearchSerializer.serialize_semester_enrollments(cls.fa_enrollments)
 
     def test_full_program_user_serialization(self):
         """
@@ -146,6 +149,7 @@ class UserProgramSearchSerializerTests(TestCase):
         assert UserProgramSearchSerializer.serialize(self.program_enrollment) == {
             'id': program.id,
             'enrollments': self.serialized_enrollments,
+            'semester_enrollments': self.semester_enrollments,
             'grade_average': 75,
             'is_learner': True,
             'email_optin': True,
@@ -164,6 +168,7 @@ class UserProgramSearchSerializerTests(TestCase):
         assert UserProgramSearchSerializer.serialize(self.program_enrollment) == {
             'id': program.id,
             'enrollments': self.serialized_enrollments,
+            'semester_enrollments': self.semester_enrollments,
             'grade_average': 75,
             'is_learner': True,
             'email_optin': email_optin_flag,
@@ -183,6 +188,7 @@ class UserProgramSearchSerializerTests(TestCase):
         expected_result = {
             'id': self.fa_program.id,
             'enrollments': self.fa_serialized_enrollments,
+            'semester_enrollments': self.fa_semester_enrollments,
             'grade_average': 95,
             'is_learner': True,
             'email_optin': True,
@@ -207,6 +213,7 @@ class UserProgramSearchSerializerTests(TestCase):
         assert UserProgramSearchSerializer.serialize(self.program_enrollment) == {
             'id': program.id,
             'enrollments': self.serialized_enrollments,
+            'semester_enrollments': self.semester_enrollments,
             'grade_average': 75,
             'is_learner': False,
             'email_optin': True,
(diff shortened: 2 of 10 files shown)


gsidebo:
Details about the hacky-ness of this PR:

1. I'm unaware of searchkit-sponsored way to sort by one field in a set of nested documents and display another field. It would be ideal to index like this `[{'semester': '2016 - Fall', 'semester_sort': '2016_3']` and display `semester` while sorting the options by `semester_sort` but I just couldn't find a way to do it.
1. The recommended way to apply sorting to aggregated buckets is using `bucketsTransform` (see the [MenuFilter doc](http://docs.searchkit.co/stable/docs/components/navigation/menu.html), 'Props' section). I was delighted to find out that there is a bug in searchkit (gitter discussion, open github issue) where the `bucketsTransform` prop is effectively ignored for the `MenuFilter` component. This is why I created `PatchedMenuFilter`. A side perk for this hack is that it gets rid of the "All" option in the UI, which we've never wanted anyway.
1. `sortSemesterBuckets` [[source](https://github.com/mitodl/micromasters/pull/2179/files#diff-19b75ce18fd1eaefee48fc255d15b45bR54)] is the function that sorts the semester options based on values like "2016 - Fall" and "2015 - Summer". The big risk here is that if the index value is changed, this function could break and we wouldn't necessarily know about it. If we go with this approach, we might want to think about ways to mitigate that risk.

Ferdi:
@gsidebo are you doing all this to sort the facets  ?  And would it be easier / cleaner to sort by nr of hits ?

courses/utils.py
```diff
@@ -0,0 +1,50 @@
...
+        return int(season_year_tuple[1]), int(season_year_tuple[0])
+    return None, None
+
+
+def get_year_season_from_dates(course_run):
+    if not course_run.start_date:
+        return None, None
+    start_month = course_run.start_date.month
+    if 1 <= start_month <= 4:
+        season_num = 1
```
noisecapella: Why not just use the season strings directly instead of `NUMBER_SEASON_MAP`?

gsidebo: yeah, this is 'legacy'. i had an indexing structure that i thought would work for sorting these results directly in elasticsearch, but that didn't end up working out with searchkit. i can change these functions to return the season directly instead of the integer

requirements.txt
```diff
@@ -35,9 +36,11 @@ jsonfield==1.0.3
 kombu==3.0.37             # via celery, django-server-status
 newrelic==2.72.1.53
 oauthlib==2.0.0           # via python-social-auth, requests-oauthlib
+pbr==1.10.0               # via stevedore
 pillow==3.4.2
 psycopg2==2.6
 pyjwt==1.4.2              # via python-social-auth
+pymongo==3.4.0            # via edx-opaque-keys
```
noisecapella: Haha I didn't expect this from edx-opaque-keys

gsidebo: nor would i!

courses/utils.py
noisecapella: I see, it's to match the number when parsing in the other function

search/indexing_api.py
```diff
@@ -254,6 +254,9 @@ def program_enrolled_user_mapping():
             'ancestors': NOT_ANALYZED_STRING_TYPE,
             'value': NOT_ANALYZED_STRING_TYPE,
             'order': LONG_TYPE
+        }},
+        'semester_enrollments': {'type': 'nested', 'properties': {
+            'semester': NOT_ANALYZED_STRING_TYPE
```
noisecapella: Can you give me an example value for this field? Is the semester meant to be part of the program object?

gsidebo:
@Ferdi Most of the hacky bits are accommodating the sort, yes. If we ignore the sort, the hacky bits pretty much go away. It's up to us to decide how reasonable the hack is. I personally think it's small/contained enough to be reasonable, but I wanted more eyes on it. We do get a side benefit out of this like I said above - we eliminate the 'All' option for the facet, which we didn't want in the first place.

gsidebo:
@aliceriot @noisecapella any general feelings about the approach here? are we ok with this? 

another approach would be to index the values as sortable terms (eg: "2016_3", "2016_1" instead of "2016 - Fall", "2016_Spring") and create a new component that accepts a translation _function_ rather than a translation object (like the one we're using for country codes). that way the sort would be applied automatically and "2016_3" would be translated to "2016 - Fall" on the fly. probably should be an option in searchkit to begin with, but :man_shrugging:

noisecapella:
I haven't done a thorough review of this but it looks reasonable to me at a glance

static/js/components/LearnerSearch.js
```diff
@@ -41,10 +42,28 @@ const makeSearchkitTranslations: () => Object = () => {
...
+      return _.sortBy(buckets, [
+        (bucket) => {
+          let yearSeasonArray = bucket.key.split(/[^\w]+/);
+          return yearSeasonArray[0] + SEMESTER_SEASON_NUMBER_MAP[yearSeasonArray[0]];
+        }
+      ]).reverse();
+    } else {
+      return buckets;
+    }
+};
```
aliceriot: Is there a reason why you need the `buckets.length > 0` control flow? my suggestion would be something like this:

```js
const sortSemesterBuckets = R.compose(R.reverse, R.sortBy(R.compose(
  str => `${str}${SEMESTER_SEASON_NUMBER_MAP[str]}`,
  R.head,
  R.split(/[^\w]+/),
  R.prop('key')
)));
```

which might make it a little clearer.

gsidebo:
thanks @noisecapella @aliceriot. this is ready to review in earnest. either one of you can feel free to do it (don't think we'll need both)

courses/models.py
```diff
@@ -4,7 +4,6 @@
 import logging
 from datetime import datetime
 import urllib.parse
-
```
noisecapella: Can you revert this whitespace change?

courses/utils.py
```diff
@@ -0,0 +1,81 @@
...
+    3: 'Fall'
+}
+
+
+def parse_edx_key(edx_course_key):
+    """
+    Args:
+        edx_course_key (str): An edX course key (CourseRun.edx_course_key)
+
+    Returns:
```
noisecapella: Can you put the type right after `Returns:` instead of within parentheses in the description? It helps my IDE parse these if it matches the schema exactly

courses/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        tuple: (Year, Season string)
+    """
+    parsed_course_key = parse_edx_key(course_run.edx_course_key)
+    if parsed_course_key:
+        season_year_tuple = get_year_season_from_edx_key(parsed_course_key)
+    else:
+        season_year_tuple = get_year_season_from_dates(course_run)
+    return season_year_tuple[0], NUMBER_SEASON_MAP[season_year_tuple[1]]
```
noisecapella: This fails if the tuple is `None, None`. To produce the error I had a course with the course key = `course-v1:MIT+GEO102+GEO_T1`. We should assume that some course keys will be parsed by `parse_edx_key` but don't actually correspond to meaningful year/season numbers

courses/utils.py
```diff
@@ -0,0 +1,81 @@
...
+def get_year_season_from_dates(course_run):
+    """
+    Args:
+        course_run (CourseRun): A CourseRun object
+
+    Returns:
+        tuple: (Year, Season number)
+    """
+    if not course_run.start_date:
+        return None, None
```
noisecapella: Can we add a test for this case?

courses/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        parsed_course_key (str): A semester/year indicator from a course key (eg: '1T2016')
+
+    Returns:
+        tuple: (Year, Season number)
+    """
+    found = re.search('^[1-3]T[1|2][0-9]{3}$', parsed_course_key)
+    if found:
+        season_year_tuple = found.group().split('T')
+        return int(season_year_tuple[1]), int(season_year_tuple[0])
+    return None, None
```
noisecapella: Can we add a test for this case?

courses/utils.py
```diff
@@ -0,0 +1,81 @@
...
+from opaque_keys import InvalidKeyError
+
+NUMBER_SEASON_MAP = {
+    1: 'Spring',
+    2: 'Summer',
+    3: 'Fall'
+}
+
+
+def parse_edx_key(edx_course_key):
```
noisecapella: Can you add a brief description saying what this function is meant for?

dashboard/serializers_test.py
```diff
@@ -388,3 +396,52 @@ def test_course_enrollments_serialized_unique(self):
...
+            profile = ProfileFactory.create()
+        cls.user = profile.user
+        cls.program = ProgramFactory.create()
+        courses = CourseFactory.create_batch(7, program=cls.program)
+        runs = list()
+        # Add course runs that have a valid, edX-parse-able course key
+        runs.append(CourseRunFactory.create(course=courses[0], edx_course_key='course-v1:MITx+course1+1T2016'))
+        runs.append(CourseRunFactory.create(course=courses[1], edx_course_key='course-v1:MITx+course3+3T2016'))
+        # Add course runs that have an invalid key; they will be serialized based on their start_date
+        fall_2016_dt = datetime.strptime('10/01/2016', "%m/%d/%Y")
```
noisecapella: Why are we parsing the datetime this way? Also, you can pass the string to the ORM and it will parse it into a date or datetime depending on the field

dashboard/serializers_test.py
```diff
@@ -388,3 +396,52 @@ def test_course_enrollments_serialized_unique(self):
...
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        with mute_signals(post_save):
+            profile = ProfileFactory.create()
+        cls.user = profile.user
+        cls.program = ProgramFactory.create()
+        courses = CourseFactory.create_batch(7, program=cls.program)
+        runs = list()
```
noisecapella: If these runs are just relevant to one test we should create the data in that test, so that the input data and the assertions are close together

courses/utils.py
gsidebo: i don't understand what you mean here. i copied the format of the docstring directly from elsewhere in the codebase, and the type (`tuple`) is not within parenthesis

dashboard/serializers_test.py
gsidebo: silly oversight. i copied some snippet from one of my notebooks. i'll change this

courses/utils.py
gsidebo: good catch. i meant to test for this and whiffed

gsidebo:
@noisecapella should be ready now

courses/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Attempts to parse a CourseRun's edx_course_key as an edX opaque key and return the portion
+    of the key that indicates the course's semester/year
+
+    Args:
+        edx_course_key (str): An edX course key (CourseRun.edx_course_key)
+
+    Returns:
+        The 'run' portion of a parsed CourseKey (opaque_keys.edx.keys.CourseKey#run),
```
noisecapella: Can you add the return type like this: `str: The 'run' portion of a parsed CourseKey...`? These docstrings are parsed by PyCharm so it helps to be specific about syntax

gsidebo: done

courses/utils.py
```diff
@@ -0,0 +1,91 @@
...
+        parsed_course_key (str): A semester/year indicator from a course key (eg: '1T2016')
+
+    Returns:
+        tuple: (Year, Season number)
+    """
+    found = re.search('^[1-3]T[1|2][0-9]{3}$', parsed_course_key)
+    if found:
+        season_year_tuple = found.group().split('T')
+        return int(season_year_tuple[1]), int(season_year_tuple[0])
+    return ()
```
noisecapella: Can you add an assertion to cover this case? I got to this path using `course-v1:MITX+MITx_Digital_learning_300+3Tabc`

noisecapella:
Functionality works great

PR #2183, opened 2016-12-21
Updated alter_data commands and fixed various issues

gsidebo:

No tickets. 

#### What's this PR do?

Fixes a bunch of outstanding issues with alter_data and adds a few simple new options. Here's a list of the fixes:

1. Changed the 'action' parameter to be a positional argument b/c it's required
1. Added to ability to target specific course runs if that's what the user wants to do. (eg: `alter_data set_to_failed --username=staff --course-run-key='MITx+Analog+Learning+100+Dec_2016'`)
1. Made usage details and example commands more helpful
1. Renamed a bunch of functions/variables to be more clear
1. Got rid of the '--add-if-exists' option for adding past failed/past runs. It wasn't helping, so that's done by default now

Run through a few of the example commands printed by `alter_data examples`, tweak a few arguments here and there, and check on the results either using `alter_data course_info --username=YOUR_USER --course-title='SOME_COURSE_TITLE'` or loading the dashboard page in the app

### DIFF
--- a/seed_data/lib.py
+++ b/seed_data/lib.py
@@ -5,9 +5,9 @@ import re
 from datetime import datetime, timedelta
 from functools import wraps
 import pytz
-
+from django.db.utils import IntegrityError
 from django.contrib.auth.models import User
-from courses.models import Program, Course
+from courses.models import Program, Course, CourseRun
 from dashboard.api_edx_cache import CachedEdxDataApi
 from dashboard.models import CachedCertificate, CachedEnrollment, CachedCurrentGrade, UserCacheRefreshTime
 from financialaid.factories import TierProgramFactory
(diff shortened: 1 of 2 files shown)


gsidebo:
This code centers around creating and altering test data, and as such it isn't critical to create unit tests. We may want to test this code down the line, but for now please ignore the codecov checks

noisecapella:
It looks like this PR includes #2048 squashed into it. Can you rebase to remove it?

gsidebo:
@noisecapella just rebased again. does everything look ok? commit `51809f417e2340a5824546076943daf8b9770e95`, which looks to be associated with 2048, is still in the git log

seed_data/lib.py
```diff
@@ -36,54 +36,62 @@ class ModelFinder(object):
...
         elif len(objects) == 0:
             raise Exception(
                 'No {} found with the given params ({})'.format(
-                    self.model_cls.__name__,
+                    cls.model_cls.__name__,
                     params
                 )
             )
         elif len(objects) > 1:
+            exc_text = "Multiple {} records found with the given params ({}). These parameters need to " + \
```
noisecapella: Can you change the string concat to use the literal concat syntax instead, ie: `"abc" "def"` becomes `"abcdef"`

seed_data/management/commands/alter_data.py
```diff
@@ -177,55 +173,39 @@ def add_future_run(user=None, course=None, now=None):  # pylint: disable=unused-
...
-    CachedEnrollmentHandler(user).set_or_create(course_run=chosen_past_run)
-    CachedCertificateHandler(user).set_or_create(course_run=chosen_past_run, grade=grade)
-
-    return chosen_past_run
+    ungraded_past_run = get_past_ungraded_course_run(user, course, now)
+    return set_to_passed(user=user, course_run=ungraded_past_run, grade=grade)
 
 
-def _formatted_datetime(dt, dt_format='%m/%d/%Y'):
+def _formatted_datetime(dt, dt_format='%m/%d/%Y %H:%M:%S'):
```
noisecapella: Why not use ISO-8601 here?

gsidebo: i could. i had it in this format for the benefit of jared and any other non-dev folks who might want to see `course_info` output (I figured it would be a little prettier than ISO). since i'm adding time into the output i think it makes sense to go ahead and change it

noisecapella:
:+1:

PR #2185, opened 2016-12-21
Upgrade Wagtail to 1.8

singingwolfboy:
Follow-up to #1635. [ChangeLog is here.](https://github.com/wagtail/wagtail/blob/master/CHANGELOG.txt) Doesn't look like there's any big changes that will affect us in this upgrade.

This pull request is not urgent. It is only to keep us up-to-date on the upgrade treadmill.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -26,4 +26,4 @@ redis==2.10.5
 requests
 static3==0.5.1
 uwsgi
-wagtail==1.7
+wagtail==1.8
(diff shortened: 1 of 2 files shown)


rhysyngsun:
@singingwolfboy how do i manually test this? I assume at least migrate and then exercise the cms admin pages and public pages? Is there anything we do in production w/ wagtail I need to replicate locally for testing?

singingwolfboy:
@rhysyngsun Yeah, I think that's a good manual test. I don't think there's anything special we do with Wagtail in production.

rhysyngsun:
:+1: from manually exercising the cms this looks fine to me

PR #2187, opened 2016-12-21
Added model for Coupon

noisecapella:
Part of #2192

#### What's this PR do?
Adds a `Coupon` model


ecommerce/models.py
```diff
@@ -154,3 +154,53 @@ def save(self, *args, **kwargs):
...
+    """
+    CREATED = 'created'
+    USED = 'used'
+    DELETED = 'deleted'
+    STATUSES = [CREATED, USED, DELETED]
+
+    PERCENT_DISCOUNT = 'percent-discount'
+    FIXED_DISCOUNT = 'fixed-discount'
+
+    AMOUNT_TYPES = [PERCENT_DISCOUNT, FIXED_DISCOUNT]
```
giocalitri: nit: remove space before this so they can be grouped together

ecommerce/models.py
```diff
@@ -154,3 +154,53 @@ def save(self, *args, **kwargs):
...
+
+    COURSE_RUN = 'course-run'
+    COURSE = 'course'
+    PROGRAM = 'program'
+    PRODUCT_TYPES = [PROGRAM, COURSE, COURSE_RUN]
+
+    order = ForeignKey(Order, null=True)
```
giocalitri: is the coupon applied to the `order`? I thought it was going to be applied to the course_run so to the `line`

ecommerce/models.py
```diff
@@ -154,3 +154,53 @@ def save(self, *args, **kwargs):
...
+    status = CharField(
+        choices=[(status, status) for status in STATUSES],
+        max_length=30,
+    )
+    amount_type = CharField(
+        choices=[(type, type) for type in AMOUNT_TYPES],
```
giocalitri: do not use `type` as variable: maybe just add a `_` at the end.

ecommerce/models.py
```diff
@@ -154,3 +154,53 @@ def save(self, *args, **kwargs):
...
+        max_length=30,
+    )
+    product_type = CharField(
+        choices=[(type, type) for type in PRODUCT_TYPES],
```
giocalitri: same here for `type`

ecommerce/models.py
```diff
@@ -154,3 +154,53 @@ def save(self, *args, **kwargs):
...
+    amount = DecimalField(decimal_places=2, max_digits=20)
+    expiration_date = DateTimeField(null=True)
+
+    def __str__(self):
+        """Description for Coupon"""
+        return (
+            "Coupon with product_type={product_type}, product_id={product_id},"
+            " amount_type={amount_type}, amount={amount}, order={order}".format(
+                product_type=self.product_type,
+                product_id=self.product_id,
```
giocalitri: where has `product_id` been defined?

ecommerce/models.py
noisecapella: I think we want coupons to apply to program, course, or course run, but you're right that this should be line in any case

noisecapella:
@giocalitri I talked with @pdpinch and there are some things I missed so I'm rewriting this model. I'll let you know when it's updated

noisecapella:
@giocalitri Can you look at this again?

ecommerce/models.py
```diff
@@ -154,3 +156,97 @@ def save(self, *args, **kwargs):
...
+        choices=[(_type, _type) for _type in AMOUNT_TYPES],
+        max_length=30,
+    )
+    # Either a number from 0 to 1 representing a percent, or a fixed value for discount
+    amount = DecimalField(decimal_places=2, max_digits=20)
+
+    # Number of people this coupon can be redeemed by
+    num_coupons_available = IntegerField(null=False)
+    # Number of times a person can redeem a coupon
+    num_redemptions = IntegerField(null=False)
```
giocalitri: the total number is the product of `num_coupons_available` * `num_redemptions`? How will we validate this?

ecommerce/models.py
```diff
@@ -154,3 +156,97 @@ def save(self, *args, **kwargs):
...
+
+    # The coupon code used for redemption by the purchaser in the user interface.
+    # If blank, the purchaser may not redeem this coupon through the user interface,
+    # though it may be redeemed in their name by an administrator.
+    coupon_code = TextField(null=True, blank=True)
+    product_type = CharField(
+        choices=[(_type, _type) for _type in PRODUCT_TYPES],
+        max_length=30,
+    )
+    product_id = IntegerField(null=False)
```
giocalitri: what's the `product_id`?

ecommerce/models.py
```diff
@@ -154,3 +156,97 @@ def save(self, *args, **kwargs):
...
+
+class UserCoupon(Model):
+    """
+    Model for uses of a coupon by a user.
+    """
+    coupon = ForeignKey(Coupon, on_delete=SET_NULL)
+    user = ForeignKey(settings.AUTH_USER_MODEL, on_delete=SET_NULL)
+
+    # Number of times left the user can use the coupon in a purchase
+    available_redemptions = IntegerField()
```
giocalitri: Isn't this a copy of `Coupon. num_redemptions`? Do we need to validate that this number cannot be higher that the other? Do we need to default to the other (not sure it is possible to)?

ecommerce/models.py
noisecapella: The primary key of either `Course`, `CourseRun`, or `Program` depending on what `product_type` is

noisecapella: I'll add a comment for this

giocalitri: I do not like this solution: what about having 3 different foreign keys to the proper models and add a `save()` method that checks that only one of them can be `not null`?

giocalitri: or anything else that assure an actual constraint

ecommerce/models.py
noisecapella: We will need to do this at the point where the user is requesting to redeem the coupon. I'm thinking of something like this pseudocode:

    def redeem_coupon(coupon, user):
        with transaction.atomic():
            try:
                user_coupon = UserCoupon.objects.get(user=user, coupon=coupon)
            except UserCoupon.DoesNotExist:
                if coupon.num_coupons_available == 0:
                    raise ValidationError("No coupons left")
                    
                coupon.num_coupons_available -= 1
                coupon.save()
                user_coupon = UserCoupon.objects.create(user=user, coupon=coupon)
            
            if user_coupon.available_redemptions == 0:
                raise ValidationError("Coupon has no more redemptions left")
            user_coupon.available_redemptions -= 1
            user_coupon.save()

        order = create_unfulfilled_order()
        RedeemedCoupon.objects.create(order=order, coupon=coupon)
        order.status = 'fulfilled'
        order.save()

ecommerce/models.py
noisecapella: The difference is, `Coupon.num_redemptions` is the number of redemptions possible for any particular user, and `UserCoupon.available_redemptions` is the number of redemptions left for that specific user.

In general, the only field we should be mutating in `Coupon` is `num_coupons_available`, all other fields should generally be set at creation time only.

ecommerce/models.py
noisecapella: Sure, I'll make the three foreign keys

noisecapella: Should I keep `product_type` so the code knows which foreign key to use? Do you think it is redundant?

giocalitri: Maybe we can keep it just to help enforce the association: if the user select `program` in `product_type`, they cannot put a foreign key to `course_run`

giocalitri: Actually we should put all this logic in validators, not in the `save()`

noisecapella: Yeah, the `save` function won't run if the user updates via `Coupon.objects.filter(...).update`

noisecapella: Regarding `product_type` I'm not sure it adds much. If they put a foreign key to `course_run` and the key is not a `CourseRun` they will already get a validation error

ecommerce/models.py
giocalitri: Ok, so we need a way to automatically set this number based on the others and not let people do it, not even from the Admin

ecommerce/models.py
giocalitri: what key?

ecommerce/models.py
noisecapella: We can make the field read-only in the admin but there's not much we can do to prevent people from modifying it in the shell if they want.

Speaking of, I'll add admin interfaces for these models

giocalitri: yeah

ecommerce/models.py
noisecapella: If a user sets `self.course_run = program` they will already get an error on save, even without checking `product_type`

giocalitri: yeah, but you still need to write the validators for having only one foreign key

noisecapella: Right, but we don't need `product_type` for that. We can just count the null values of `self.course_run`, `self.course`, `self.program` and make sure that number is always `1`

giocalitri: OK

ecommerce/models.py
```diff
@@ -154,3 +160,93 @@ def save(self, *args, **kwargs):
...
+    amount = DecimalField(decimal_places=2, max_digits=20)
+
+    # Number of people this coupon can be redeemed by
+    num_coupons_available = IntegerField(null=False)
+    # Number of times a person can redeem a coupon
+    num_redemptions = IntegerField(null=False)
+    # After this time the coupons will not be redeemable
+    expiration_date = DateTimeField(null=True)
+    # If true, coupons are not presently redeemable
+    disabled = BooleanField(default=False)
```
singingwolfboy: It's generally a good idea to avoid double negatives, and the way you've named this field, the way to check if the coupon is active is by saying `if not disabled` -- a double-negative. I'd recommend naming this field `enabled` instead and setting the default to `True`.

ecommerce/models.py
```diff
@@ -154,3 +160,93 @@ def save(self, *args, **kwargs):
...
+            product=product,
+            num_coupons_available=self.num_coupons_available,
+        )
+
+
+class UserCoupon(Model):
+    """
+    Model for uses of a coupon by a user.
+    """
+    coupon = ForeignKey(Coupon, on_delete=SET_NULL, null=True)
```
singingwolfboy: So a `UserCoupon` indicates a coupon that can _only_ be redeemed by a particular user, or set of users. Is that right? If so, this model seems problematic -- you have to create a `Coupon` model to hold all the basic information about the coupon, and then a `UserCoupon` model to restrict how that coupon can be used. Any queries you run to fetch all `Coupon` models will also return coupons that have `UserCoupon` restrictions, even though they need to be treated differently from coupons that do _not_ have those restrictions. (You can specify in your database query to join on the `UserCoupon` table and only return results that do _not_ have a `UserCoupon`, but that is much more complex and error-prone. Simpler is better.)

I'd recommend defining an abstract model that contains the basic fields of a coupon: amount, amount-type, program/course/run, etc. Then, have two concrete models that inherit from it: broad-use coupons and restricted-use coupons. That way, it's not possible to accidentally use the wrong type of coupon in the wrong context.

ecommerce/models.py
```diff
@@ -154,3 +160,93 @@ def save(self, *args, **kwargs):
...
+    PERCENT_DISCOUNT = 'percent-discount'
+    FIXED_DISCOUNT = 'fixed-discount'
+
+    AMOUNT_TYPES = [PERCENT_DISCOUNT, FIXED_DISCOUNT]
+
+    # The coupon code used for redemption by the purchaser in the user interface.
+    # If blank, the purchaser may not redeem this coupon through the user interface,
+    # though it may be redeemed in their name by an administrator.
+    coupon_code = TextField(null=True, blank=True)
+    # One and only one of these three foreign keys must be set to not null
```
singingwolfboy: I'd recommend checking out Django's [generic relations](https://docs.djangoproject.com/en/1.10/ref/contrib/contenttypes/#generic-relations) -- it seems like the perfect tool for this use-case.

ecommerce/models.py
singingwolfboy: I recommend changing the name from `num_redemptions` to `num_redemptions_per_user`. It's more accurate, and clarifies the distinction between this field and `num_coupons_available`.

ecommerce/models.py
```diff
@@ -154,3 +160,93 @@ def save(self, *args, **kwargs):
...
+
+    # percent or fixed discount
+    amount_type = CharField(
+        choices=[(_type, _type) for _type in AMOUNT_TYPES],
+        max_length=30,
+    )
+    # Either a number from 0 to 1 representing a percent, or a fixed value for discount
+    amount = DecimalField(decimal_places=2, max_digits=20)
+
+    # Number of people this coupon can be redeemed by
```
singingwolfboy: All of these inline comments can be specified as `help_text` on the models, which will make them show up in the Django admin. [Documentation is here.](https://docs.djangoproject.com/en/1.10/ref/models/fields/#help-text)

ecommerce/models.py
```diff
@@ -154,3 +160,93 @@ def save(self, *args, **kwargs):
...
+    num_redemptions = IntegerField(null=False)
+    # After this time the coupons will not be redeemable
+    expiration_date = DateTimeField(null=True)
+    # If true, coupons are not presently redeemable
+    disabled = BooleanField(default=False)
+
+    def __str__(self):
+        """Description for Coupon"""
+        product = self.course_run or self.course or self.program
+        return "Coupon {amount_type} {amount} for {product}, {num_coupons_available} left".format(
```
singingwolfboy: Nitpick: I'd suggest adding more information here about the current state of the coupon. For example:
```python
tags = []
uses_left = self.num_coupons_available
if uses_left > 1:
    tags.append("{num} uses left".format(num=uses_left)
elif uses_left == 1:
    tags.append("1 use left")
else:
    tags.append("fully used")
exp_date = self.expiration_date or datetime.max
if datetime.now() > exp_date:
    tags.append("expired")
if self.disabled:
    tags.append("disabled")
return "Coupon {type} {amount} for {product} ({tags})".format(
    type=self.amount_type,
    amount=self.amount,
    product=product,
    tags=", ".join(tags),
)
```

ecommerce/models.py
```diff
@@ -154,3 +160,93 @@ def save(self, *args, **kwargs):
...
+    """
+    order = ForeignKey(Order, on_delete=SET_NULL, null=True)
+    coupon = ForeignKey(Coupon, on_delete=SET_NULL, null=True)
+
+    class Meta:
+        unique_together = ('order', 'coupon',)
+
+    def __str__(self):
+        """Description for RedeemedCoupon"""
+        return "RedeemedCoupon for order {order}, coupon {coupon}".format(
```
singingwolfboy: If the `__str__()` method on `Coupon` contains the word "coupon", you don't need to _also_ include the word "coupon" here.

ecommerce/models.py
noisecapella: Sure

ecommerce/models.py
noisecapella: A `UserCoupon` is a coupon which has already been redeemed by a user one or more times. Admittedly the name could be better, but I don't think an inheritance relationship fits here since the `UserCoupon` only links a `Coupon` to a `User` and stores how many times the user has used it.

noisecapella:
Ok, comments should be addressed now. @giocalitri I removed `UserCoupon` because @singingwolfboy said it was redundant, since we can get the same information from `RedeemedCoupon`. I added some validation in the `clean` method

ecommerce/factories.py
```diff
@@ -87,3 +89,41 @@ class CoursePriceFactory(DjangoModelFactory):
...
+    amount_type = FuzzyChoice(Coupon.AMOUNT_TYPES)
+    coupon_code = FuzzyChoice([None, FuzzyText().fuzz()])
+    num_redemptions_per_user = FuzzyInteger(1, 10)
+    num_coupons_available = FuzzyInteger(1, 100)
+
+    class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,old-style-class
+        model = Coupon
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
```
singingwolfboy: I think you're trying to be too clever, here. Factory Boy is best used with reasonable, static values that can be easily extended and overwritten when the factory is used in a test. Trying to override the `_create()` method usually leads to awkward interactions down the line, when trying to use this factory with others. Here's what I'd suggest instead:

```python
import factory
from factory.django import DjangoModelFactory


class CouponFactory(DjangoModelFactory):
    coupon_code = FuzzyText()
    amount_type = 'percent-discount'
    amount = 0.5
    num_coupons_available = 15
    num_redemptions_per_user = 2

    class Meta:
        model = Coupon

    content_object = factory.SubFactory(ProgramFactory)

    @factory.post_generation
    def content_type(self, create, extracted, **kwargs):
        return self.content_object.__class__

    @factory.post_generation
    def object_id(self, create, extracted, **kwargs):
        return self.content_object.id

    class Params:
        percent = factory.Trait(
            amount_type='percent-discount',
            amount=0.5, # 50% off
        )
        fixed = factory.Trait(
            amount_type='fixed-discount',
            amount=100, # $100 off
        )
        program = factory.Trait(
            content_object=factory.SubFactory(ProgramFactory)
        )
        course = factory.Trait(
            content_object=factory.SubFactory(CourseFactory)
        )
        course_run = factory.Trait(
            content_object=factory.SubFactory(CourseRunFactory)
        )
```

This defines some reasonable, static defaults while easily allowing the caller to override them using the [trait](https://factoryboy.readthedocs.io/en/latest/reference.html#traits) system. For example, to make a coupon for a course instead of a program, just run:
```python
coupon = CouponFactory(course=True)
```
Or, if you already have a course object that you want to generate a coupon for:
```python
coupon = CouponFactory(content_object=my_course)
```

By default, these coupons are for 50% off. To adjust the percentage, just change it in the caller:
```python
coupon = CouponFactory(amount=0.3)  # 30% off
```
Or to change it to a fixed amount, use the `fixed` trait:
```python
coupon = CouponFactory(fixed=True, amount=150)  # $150 off
```
You can even use the `percent` trait for clarity:
```python
coupon = CouponFactory(percent=True, amount=0.8)  # 80% off
```

And of course, all the other properties can be easily overridden as well, as per usual for Factory Boy factories:
```python
coupon1 = CouponFactory(coupon_code="mr.r0bot")
coupon2 = CouponFactory(num_coupons_available=9000, num_redemptions_per_user=10)
coupon3 = CouponFactory(activation_date=datetime(2018, 1, 1, 0, 0, 0), expiration_date=datetime(2019, 1, 1, 0, 0, 0))
coupon4 = CouponFactory(enabled=False)
```

noisecapella: sure

ecommerce/models.py
```diff
@@ -154,3 +161,116 @@ def save(self, *args, **kwargs):
...
+        null=True,
+        blank=True,
+        help_text="If set, the coupons will not be redeemable after this time",
+    )
+    activation_date = DateTimeField(
+        null=True,
+        blank=True,
+        help_text="If set, the coupons will not be redeemable before this time",
+    )
+    enabled = BooleanField(default=True, help_text="If true, coupons are presently redeemable")
```
singingwolfboy: I would suggest adding a `is_redeemable` utility method to this class:

```python
def is_redeemable(self):
    if not self.enabled:
        return False
    start = self.activation_date or datetime.min
    end = self.expiration_date or datetime.max
    return start <= datetime.now() <= end
```

It encapsulates checking all the conditions that could cause this coupon to not be redeemable at this point in time. If we add other possible conditions in the future, those could be rolled into this method as well.

noisecapella: I'd like to wait on that until we add this functionality in a future PR

singingwolfboy: Seems reasonable to me.

ecommerce/factories.py
noisecapella: I removed the `content_type` and `object_id` pieces, these are automatically handled by the `GenericForeignKey`

noisecapella:
Any other comments?

ecommerce/factories.py
```diff
@@ -87,3 +94,36 @@ class CoursePriceFactory(DjangoModelFactory):
...
         model = CoursePrice
+
+
+class CouponFactory(DjangoModelFactory):
+    """Factory for Coupon"""
+    coupon_code = FuzzyText()
+    amount_type = 'percent-discount'
+    amount = FuzzyDecimal(0, 1)
+    num_coupons_available = FuzzyInteger(0, 15)
+    num_redemptions_per_user = FuzzyInteger(0, 5)
```
singingwolfboy: Is `0` a reasonable value for these fields? If not, it shouldn't be included in the FuzzyInteger range.

noisecapella: Sure, I'll update it to 1

PR #2191, opened 2016-12-21
Added slug to FrequesntlyAskedQuestion

### LINKED ISSUE #1878: Add ids to the FAQs so we can create links to individual questions
Although I think this is a good idea, no one on the support team has asked for it.

annagav:
Fix #1878
#### What's this PR do?
Added unique SlugField for FrequesnltyAskedQuestion in cms in order to reference questions in the url.
Before you run the migration, make to have few existing faqs. Make sure to exit the editing mode for that page in wagtail (otherwise you will be seeing saved draft that will not show you any updates that took place in the models).
Then run the migration. Now when editing the `FrequesntlyAskedQuestion` objects you should see a third field `slug` that is populated with a generated slug from the question itself.

screen shot 2016-12-21 at 4 41 27 pm


giocalitri:
Please fix linting

Ferdi:
@annagav I'm not sure you need to have a separate field in db, deal with migrations etc.  Can you generate and ID on the fly   based on the item_id in the database and the slugify version of the title ?

annagav:
@Ferdi, I thought it would be nice if the user had the choice to edit the slug if they wanted to. And this way it gets nicer url without all the item_id numbers.

annagav:
Added pilcrow link that appears on hover:

screen shot 2016-12-22 at 3 55 03 pm

cms/models.py
```diff
@@ -303,12 +305,23 @@ class FrequentlyAskedQuestion(Orderable):
     faqs_page = ParentalKey(CategorizedFaqsPage, related_name='faqs', null=True)
     question = models.TextField()
     answer = RichTextField()
+    slug = models.SlugField(unique=True, default=None, blank=True)
+
+    def save(self, *args, **kwargs):
+        if not self.slug:
```
noisecapella: Can you add a test for this code?

cms/models.py
```diff
@@ -303,12 +305,23 @@ class FrequentlyAskedQuestion(Orderable):
...
+            for x in itertools.count(1):
+                self.slug = orig = slugify(self.question)
+                if not FrequentlyAskedQuestion.objects.filter(slug=self.slug).exists():
```
noisecapella: It seems like `self.slug` is always going to be `slugify(self.question)`, without any number concatenated to it. Can you change it to check with the number appended to it, if necessary?

cms/migrations/0027_populate_slug_values_faqs.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.utils.text import slugify
+import itertools
+
+
+def gen_unique_slug(apps, schema_editor):
+    FrequentlyAskedQuestion = apps.get_model('cms', 'FrequentlyAskedQuestion')
+    for row in FrequentlyAskedQuestion.objects.all():
+        if not row.slug:
+            for x in itertools.count(1):
```
singingwolfboy: Looks like you're trying to do an infinite loop, and break when you find a unique slug. The `itertools` module is great, but it's hiding your intent here. A `while` loop will more clearly express the intent here:

```python
if not row.slug:
    base_slug = slugify(row.question)
    slug = base_slug
    slug_is_unique = not FrequentlyAskedQuestion.objects.filter(slug=slug).exists()
    count = 1
    while not slug_is_unique:
        slug = "{base}-{count}".format(base=base_slug, count=count)
        slug_is_unique = not FrequentlyAskedQuestion.objects.filter(slug=slug).exists()
        count += 1
    row.slug = slug
    slug.save()
```

cms/models.py
```diff
@@ -303,12 +305,23 @@ class FrequentlyAskedQuestion(Orderable):
     faqs_page = ParentalKey(CategorizedFaqsPage, related_name='faqs', null=True)
     question = models.TextField()
     answer = RichTextField()
+    slug = models.SlugField(unique=True, default=None, blank=True)
+
+    def save(self, *args, **kwargs):
+        if not self.slug:
+            for x in itertools.count(1):
```
singingwolfboy: Same comment as above about using a `while` loop.

cms/models.py
annagav: It will add a number only if there is a collision with another question (in the next 3 lines of code).

noisecapella: But how will it tell if the slug with the number causes a collision? Since `self.slug` is defined in the previous line it will never contain the slug with the number, just the slug itself

annagav: you are right, the original assignment should be out side the loop.

cms/migrations/0027_populate_slug_values_faqs.py
```diff
@@ -0,0 +1,33 @@
...
+        if not row.slug:
+            slug = orig_slug = slugify(row.question)
+            slug_is_unique = not FrequentlyAskedQuestion.objects.filter(slug=orig_slug).exists()
+            count = 1
+            while not slug_is_unique:
+                slug = "{orig}-{count}".format(orig=orig_slug, count=count)
+                slug_is_unique = not FrequentlyAskedQuestion.objects.filter(slug=slug).exists()
+                count += 1
+            row.slug = slug
+        row.save()
```
singingwolfboy: This line should be indented by one. There's no need to save _every_ row, only the rows that didn't previously have slugs.

singingwolfboy:
Can you combine the two migration files into one? There's no reason to have two separate migrations for this pull request.

cms/models.py
annagav: added a test in models_test.py

annagav:
@singingwolfboy, I merged the migrations into one file.

PR #2197, opened 2016-12-22
Added access control header for static assets

blarghmatey:
**Added access control header for static assets**
CMS and Admin are generating CORS errors when served through
cloudfront. Adding access control header for all static assets.

### DIFF
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -35,6 +35,7 @@ http {
         }
 
         location ~* /static/(.*$) {
+            add_header Access-Control-Allow-Origin *;
             try_files $uri $uri/ /staticfiles/$1 /staticfiles/$1/ =404;
         }
 


pdpinch:
Can someone explain what the problem was and how this fixes it?

blarghmatey:
The problem was that wagtail was attempting to use Javascript to load static assets from Cloudfront but there were no access control headers set so those requests were failing due to CORS violations. Adding the `Access-Control-Allow-Origin: *` header to all static asset responses will resolve the CORS issues by telling the browser that it is OK to retrieve those assets.

pdpinch:
So Cloudfront was requesting the Wagtail (and django admin) static files from nginx, but nginx wasn't responding with an access-control header. So Cloudfront couldn't fill it's cache?

I'm still wondering why we would be using our CDN for these fonts instead of directing people to use Google's CDN. But since these are infrequently-requested pages, maybe it's not a huge cost.

blarghmatey:
I think that the fonts in question are part of the default Wagtail setup and so we didn't configure the font loading. It's not so much that Cloudfront wasn't filling the cache as it is that the fonts were being requested via Javascript which is where the CORS issue comes into play.

PR #2204, opened 2016-12-22
Fix a bug with startProfileEdit

aliceriot:
This is most likely an edge case, and I don't think that the bug had much possibility of user-facing impact, but nonetheless I think it's a good thing to clean up.

Ok so, there are basically a few places we want to dispatch the `startProfileEdit` action. 

1. when the user visits `/profile` on a fresh, full page load
2. when the user opens a 'new education' or 'new job' dialog
3. when the user opens the 'personal' dialog on `/learner`
4. when the user is redirected back to `/profile`, because of an incomplete profile.

cases 2 and 3 are already good, no worries there. We had an issue with 1, where we had the following on `ProfilePage`:

```js
componentDidMount() {
  dispatch(startProfileEdit(SETTINGS.user.username}));
}
```

so that `startProfileEdit` would be dispatched every time that component mounted. The problem was that we also have, in `App.js`, the following:

```js
fetchUserProfile(username) {
  const { userProfile, dispatch, location: { pathname } } = this.props;
  if (userProfile.getStatus === undefined) {
    dispatch(fetchUserProfile(username)).then(() => {
      if ( PROFILE_REGEX.test(pathname) ) {
	dispatch(startProfileEdit(SETTINGS.user.username));
      }
    });
  }
}
```

We call `fetchUserProfile` on `componentDidMount` and `componentDidUpdate` for `App.js`. So a common thing that would happen is that the user could visit `/profile`, and `componentDidMount` on `ProfilePage` would fire while this `fetchUserProfile` method was still in-flight. `PROFILE_REGEX` tests if the user is on `/profile[/foo]`, so we would go ahead a fire a second `startProfileEdit` action. In this PR I removed `componentDidMount` from `ProfilePage` to get around this.

But why were we doing that in the first place?

On `App` we have the following method:

```js
requireCompleteProfile() {
  const {
    userProfile,
    userProfile: { profile },
    location: { pathname },
    dispatch,
    ui: { profileStep }
  } = this.props;
  const [ complete, step, errors] = validateProfileComplete(profile);
  const username = SETTINGS.user ? SETTINGS.user.username : null;
  const idealStep = currentOrFirstIncompleteStep(profileStep, step);

  if (
    userProfile.getStatus === FETCH_SUCCESS &&
    !PROFILE_REGEX.test(pathname) &&
    (!complete || !profile.filled_out)
  ) {
    dispatch(updateProfileValidation(username, errors));
    this.context.router.push(`/profile/${idealStep}`);
  }
}
```

Basically this method ensures that a user will fill out their profile before they can do anything else. So if the profile is downloaded, and we're not already somewhere in the profile flow, and the profile is not complete, well, then we want to dispatch the validation and go to the relevant step in the profile.

Then, the user should be shown the relevant profile step, and the messages for the invalid fields should be highlighted. You'll notice in reading the method above, however, that we never check if `startProfileEdit` had already been dispatched or not - if it hasn't we're basically dropping the validation messages on the floor. The `componentDidMount` method was added to `ProfilePage` in an attempt to get this behavior (user is redirected, validation errors are sitting there waiting) but it wasn't implemented correctly.

What we want to do is dispatch `startProfileEdit`, dispatch an action to ensure all validation errors are visible, and then dispatch the validation errors themselves, before we do the route transition.

So to fix this, I've changed the last few lines of the method above to look like this:

```js
  if (
    userProfile.getStatus === FETCH_SUCCESS &&
    !PROFILE_REGEX.test(pathname) &&
    (!complete || !profile.filled_out)
  ) {
    dispatch(startProfileEdit(SETTINGS.user.username));
    dispatch(updateValidationVisibility(username, ALL_ERRORS_VISIBLE));
    dispatch(updateProfileValidation(username, errors));
    this.context.router.push(`/profile/${idealStep}`);
  }
}
```

which should fix it - now before we even boot up `ProfilePage` we've already got our edit state and validation errors how we want them.

To test this, do the following:

1. visit `/profile`, and ensure that just loading the page causes `startProfileEdit` to be dispatched once and only once.
2. use the django shell to delete a field on your user's profile, one that is on the personal form (like `preferred_name`). Then visit another page on the dashboard (like learner). When you are redirected to `/profile/personal` you should see:
    - just one `startProfileEdit` action being dispatched
    - the validation errors for your deleted field should already be visible.

I think that's it, whew.


noisecapella:
Looks good to me :+1:

Issue #1791, opened 2016-11-16
For staff mail to learners, set the return address to be help desk


pdpinch:
Currently, the return address on bulk staff mail to learners is `no-reply@mit.edu`

For purposes of getting mail to be opened, it would be better if the name in the from were the staff member's name and the address was the help desk address, e.g. `Peter Pinch <micromasters-support@mit.edu>`

PR #2206, opened 2016-12-23
For staff mail to learners, pointed the return address to be help desk

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/1791

@pdpinch @noisecapella

### DIFF
--- a/mail/api.py
+++ b/mail/api.py
@@ -30,7 +30,7 @@ class MailgunClient:
         """
         email_from = settings.MAILGUN_FROM_EMAIL
         if sender_name:
-            email_from = "{sender_name} {email}".format(sender_name=sender_name, email=email_from)
+            email_from = "{sender_name} <{email}>".format(sender_name=sender_name, email=email_from)
         return {'from': email_from}
 
     @classmethod
@@ -75,7 +75,7 @@ class MailgunClient:
         return body, recipients
 
     @classmethod
-    def send_bcc(cls, subject, body, recipients):
+    def send_bcc(cls, subject, body, recipients, sender_name=None):
         """
         Sends a text email to a BCC'ed list of recipients.
 
@@ -94,7 +94,7 @@ class MailgunClient:
             subject=subject,
             text=body
         )
-        return cls._mailgun_request(requests.post, 'messages', params)
+        return cls._mailgun_request(requests.post, 'messages', params, sender_name=sender_name)
 
     @classmethod
     def send_batch(cls, subject, body, recipients, chunk_size=settings.MAILGUN_BATCH_CHUNK_SIZE, sender_name=None):
--- a/mail/api.py
+++ b/mail/api.py
@@ -29,7 +29,7 @@ class MailgunClient:
             json: json object containing email from info.
         """
         email_from = settings.EMAIL_SUPPORT
-        if sender_name:
+        if sender_name is not None:
             email_from = "{sender_name} <{email}>".format(sender_name=sender_name, email=email_from)
         return {'from': email_from}
 
@@ -97,7 +97,14 @@ class MailgunClient:
         return cls._mailgun_request(requests.post, 'messages', params, sender_name=sender_name)
 
     @classmethod
-    def send_batch(cls, subject, body, recipients, chunk_size=settings.MAILGUN_BATCH_CHUNK_SIZE, sender_name=None):
+    def send_batch(
+        cls,
+        subject,
+        body,
+        recipients,
+        chunk_size=settings.MAILGUN_BATCH_CHUNK_SIZE,
+        sender_name=None
+    ):  # pylint: disable=too-many-arguments
         """
         Sends a text email to a list of recipients (one email per recipient) via batch.
 
@@ -111,7 +118,6 @@ class MailgunClient:
         Returns:
             list: List of requests.Response HTTP response from Mailgun
         """
-        # pylint: disable=too-many-arguments
         body, recipients = cls._recipient_override(body, recipients)
         responses = []
 
--- a/mail/api.py
+++ b/mail/api.py
@@ -28,7 +28,7 @@ class MailgunClient:
         Returns:
             json: json object containing email from info.
         """
-        email_from = settings.MAILGUN_FROM_EMAIL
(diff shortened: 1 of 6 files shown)


mail/api.py
```diff
@@ -18,14 +18,23 @@ class MailgunClient:
...
+            sender_name (str): email's sender name.
+
+        Returns:
+            json: json object containing email from info.
         """
-        return {'from': settings.MAILGUN_FROM_EMAIL}
+        email_from = settings.MAILGUN_FROM_EMAIL
+        if sender_name:
+            email_from = "{sender_name} <{email}>".format(sender_name=sender_name, email=email_from)
+        return {'from': email_from}
```
giocalitri: Looking at @pdpinch example, I think the `sender_name` is not `settings.MAILGUN_FROM_EMAIL`

giocalitri:
I am not sure the implemented solution is complete: I think we need input from @pdpinch

mail/api.py
amir-qayyum-khan: yes it is request.user.profile => first + last name

giocalitri: sorry, what I meant is that the sender email is not `settings.MAILGUN_FROM_EMAIL`

mail/api.py
```diff
@@ -18,14 +18,23 @@ class MailgunClient:
...
         Base params for Mailgun request. This a method instead of an attribute to allow for overrides.
+
+        Args:
+            sender_name (str): email's sender name.
+
+        Returns:
+            json: json object containing email from info.
         """
-        return {'from': settings.MAILGUN_FROM_EMAIL}
+        email_from = settings.MAILGUN_FROM_EMAIL
```
pdpinch: Should we just change this to use `settings.MICROMASTERS_SUPPORT_EMAIL` ?

amir-qayyum-khan:
Done @pdpinch @giocalitri

profiles/util.py
```diff
@@ -134,3 +134,25 @@ def make_thumbnail(full_size_image, max_dimension):
...
+        profile (Profile): User's profile object.
+
+    Returns:
+        str: full name from profile.
+    """
+    if not profile:
+        return None
+
+    first = profile.first_name or profile.user.username
+    last = " {}".format(profile.last_name) if profile.last_name else ''
```
giocalitri: @pdpinch could you please take a look at this logic and say if it is fine to not consider the preferred name?

profiles/util.py
```diff
@@ -134,3 +134,25 @@ def make_thumbnail(full_size_image, max_dimension):
     pil_image.save(buffer, "JPEG", quality=90)
     buffer.seek(0)
     return buffer
+
+
+def full_name(profile):
```
giocalitri: this should take in input the `User` object and not the profile.

profiles/util.py
giocalitri: you need to remove the space from here and add it later in case the last name is available.

anyway you can write this like:
` last = " {}".format(profile.last_name or '')`

profiles/util_test.py
```diff
@@ -145,3 +150,54 @@ def test_make_thumbnail(self):
             mocked.assert_called_with(width, height, thumbnail_size)
             assert thumb_image.width == thumb_width
             assert thumb_image.height == thumb_height
+
+
+@pytest.mark.django_db
```
giocalitri: question: what does this do?

mail/api.py
```diff
@@ -18,14 +18,23 @@ class MailgunClient:
...
+
+        Args:
+            sender_name (str): email's sender name.
+
+        Returns:
+            json: json object containing email from info.
         """
-        return {'from': settings.MAILGUN_FROM_EMAIL}
+        email_from = settings.EMAIL_SUPPORT
+        if sender_name:
```
giocalitri: this should be:
`if sender_name is not None:`

mail/api.py
```diff
@@ -97,11 +106,12 @@ def send_batch(cls, subject, body, recipients, chunk_size=settings.MAILGUN_BATCH
             body (str): Text email body
             recipients (list): A list of recipient emails
             chunk_size (int): The maximum amount of emails to be sent at the same time
+            sender_name (str): email sender name.
 
         Returns:
             list: List of requests.Response HTTP response from Mailgun
         """
-
+        # pylint: disable=too-many-arguments
```
giocalitri: this should probably go next to the list of arguments

mail/api_test.py
```diff
@@ -40,6 +40,19 @@ def test_from_address(self, mock_post):
...
+        Test that the sender name and 'from' address for our emails is set correctly.
+        """
+        sender = 'tester'
+        MailgunClient.send_bcc('email subject', 'email body', ['will_be_ignored@example.com'], sender_name=sender)
+        _, called_kwargs = mock_post.call_args
+        self.assertEqual(
+            called_kwargs['data']['from'],
+            '{sender_name} <mailgun_from_email@example.com>'.format(sender_name=sender)
+        )
+
```
giocalitri: This test is not enough: given that you added an extra argument to many functions, you should test all the modified functions with the extra argument.

profiles/util_test.py
amir-qayyum-khan: was getting some exception on access test db in these tests and in exception it suggest me to use this.

giocalitri: I read [here](http://pytest-django.readthedocs.io/en/latest/helpers.html) that the reason is that the test needs the database, so the solution is to use the django `TestCase` and not the unittest one.

profiles/util.py
pdpinch: Since this is the return address from a staff member, I think it's best to be formal and use first name -- not preferred name.

amir-qayyum-khan:
Done @giocalitri

pdpinch:
Any chance we can merge this today?

mail/api.py
giocalitri: any reason why this has not changed?

giocalitri:
Just one last question: but feel free to merge as soon as you answer/address that question 👍

PR #2207, opened 2016-12-26
Used FinalGrade model to fetch final grade info on mmtrack

### LINKED ISSUE #1974: Final grades: modify mmtrack
part of #1856

We need to modify the MMTrack to pull the final grades from the proper model defined in #1957
the call to the final grades  need to return the grade and the status (passed or not passed) 
e.g. return tuple of (grade, pass)

amir-qayyum-khan:
- closes #1974

#### What's this PR do?
- Uses FinalGrade model to fetch users final grade if he has passed the course.
- To enable sue flag in .env file ``FEATURE_FINAL_GRADE_ALGO='v1'``

### DIFF
--- a/dashboard/api_test.py
+++ b/dashboard/api_test.py
@@ -2,7 +2,12 @@
 Tests for the dashboard api functions
 """
 from datetime import datetime, timedelta
-from unittest.mock import patch, MagicMock, PropertyMock
+from unittest.mock import (
+    MagicMock,
+    Mock,
+    PropertyMock,
+    patch,
+)
 
 import ddt
 import pytz
@@ -303,7 +308,7 @@ class CourseRunTest(CourseTests):
         with patch(
             'courses.models.CourseRun.has_frozen_grades',
             new_callable=PropertyMock
-        ) as frozen_mock, override_settings(FEATURES={"FINAL_GRADE_ALGO": grade_algo}):
+        ) as frozen_mock, override_settings(FEATURES={"FINAL_GRADE_ALGORITHM": grade_algo}):
             frozen_mock.return_value = has_frozen_grades
             run_status = api.get_status_for_courserun(crun, self.mmtrack)
         assert run_status.status == expected_status
@@ -438,10 +443,9 @@ class CourseRunTest(CourseTests):
         assert run_status.status == expected_status
         assert run_status.course_run == crun
 
-    @patch('grades.models.FinalGrade.objects.get', autospect=True)
     @patch('courses.models.CourseRun.is_upgradable', new_callable=PropertyMock)
     @patch('courses.models.CourseRun.has_frozen_grades', new_callable=PropertyMock)
-    def test_not_paid_in_past_grade_frozen(self, froz_mock, upgr_mock, final_grades_mock):
+    def test_not_paid_in_past_grade_frozen(self, froz_mock, upgr_mock):
         """
         test for get_status_for_courserun for course
         not paid but that is past for a course with grades already frozen
@@ -464,22 +468,21 @@ class CourseRunTest(CourseTests):
         )
 
         # case when the user has not an entry in the frozen grades
-        final_grades_mock.side_effect = FinalGrade.DoesNotExist
+        self.mmtrack.extract_final_grade.side_effect = FinalGrade.DoesNotExist
         run_status = api.get_status_for_courserun(crun, self.mmtrack)
         assert run_status.status == api.CourseRunStatus.CURRENTLY_ENROLLED
         # just resetting the mock does not remove the side effect
-        final_grades_mock.side_effect = None
+        self.mmtrack.extract_final_grade.side_effect = None
 
         # case in a final grade with status passed=True is returned
-        grade = MagicMock(passed=True)
-        final_grades_mock.return_value = grade
+        grade = Mock(passed=True)
+        self.mmtrack.extract_final_grade.return_value = grade
         run_status = api.get_status_for_courserun(crun, self.mmtrack)
         assert run_status.status == api.CourseRunStatus.CAN_UPGRADE
-        final_grades_mock.reset_mock()
 
         # case in a final grade with status passed=False is returned
-        grade = MagicMock(passed=False)
-        final_grades_mock.return_value = grade
+        grade = Mock(passed=False)
+        self.mmtrack.extract_final_grade.return_value = grade
         run_status = api.get_status_for_courserun(crun, self.mmtrack)
         assert run_status.status == api.CourseRunStatus.NOT_PASSED
 
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -3,7 +3,9 @@ Utility functions and classes for the dashboard
 """
 import logging
 from decimal import Decimal
+
 from django.core.exceptions import ImproperlyConfigured
+from django.conf import settings
 from django.db import transaction
 from django.db.models import Max, Min, Q
 
@@ -12,6 +14,8 @@ from dashboard.api_edx_cache import CachedEdxUserData
 from ecommerce.models import Line
 from financialaid.constants import FinancialAidStatus
 from financialaid.models import FinancialAid, TierProgram
+from grades.constants import FinalGradeStatus
+from grades.models import FinalGrade
 
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


giocalitri:
This can be reviewed and merged only after #2126 has been merged and the final grades for past course runs have been frozen

giocalitri:
We need to reimplement this PR using feature flags (see PR #2227 ).

The idea is that we need to support both the current code and the new one at the same time and un only one of the two based on the feature flag.

In this way we can merge this code before we are done with freezing the grades and enable the new code only when we are ready.

giocalitri:
so, the idea behind the code for this PR is the following: you need to keep the old code as is (tests included) and at the beginning of every function/method that needs the new logic, do something like:
```
if settings.FEATURES.get('THE_NAME_OF_THE_FEATURE_FLAG'):
    ...
    # the new code to pull the final grade
    return final_grade 
    ...
```

IMPORTANT: do not change the old tests and try to not mix the old code and the new code, because it will be easier later to remove the old code when we do not need it any more.

amir-qayyum-khan:
Done @giocalitri can you take a look. Implemented feature flag algo version.

dashboard/utils.py
```diff
@@ -38,6 +44,7 @@ class MMTrack:
     financial_aid_min_price = None
     financial_aid_max_price = None
     financial_aid_date_documents_sent = None
+    final_grades = {}
```
singingwolfboy: Based on the code in the `extract_final_grade` method, this is actually a cache. Can you use [Django's cache framework](https://docs.djangoproject.com/en/1.10/topics/cache/) instead of implementing a cache yourself?

dashboard/utils.py
```diff
@@ -167,6 +174,32 @@ def has_verified_cert(self, course_id):
...
+        try:
+            final_grade = FinalGrade.objects.get(
+                course_run__edx_course_key=course_id,
+                user=self.user,
+                status=FinalGradeStatus.COMPLETE,
+            )
+        except FinalGrade.DoesNotExist:
+            log.exception('No final grade found for user %s in course %s', self.user.username, course_id)
+            raise
+        grade = MMTrackGrade(grade=final_grade.grade, passed=final_grade.passed)
```
singingwolfboy: Why even bother to turn this into a `MMTrackGrade` object? Why not just return the `FinalGrade` object?

dashboard/utils.py
```diff
@@ -179,6 +212,12 @@ def has_passed_course(self, course_id):
         Returns:
             bool: whether the user has passed the course
         """
+        version = settings.FEATURES.get("FINAL_GRADE_ALGO", "v0")
```
singingwolfboy: Can you use a more meaningful variable name than `version`? Also, please use the name `FINAL_GRADE_ALGORITHM` instead of `FINAL_GRADE_ALGO`: it's clearer.

dashboard/utils_test.py
```diff
@@ -363,9 +367,108 @@ def test_is_enrolled_mmtrack_fa(self):
...
+        mmtrack = MMTrack(
+            user=self.user,
+            program=self.program,
+            edx_user_data=self.cached_edx_user_data
+        )
+
+        assert mmtrack.has_passed_course(course_id) is True
+        extr_grade_mock.assert_called_once_with(mmtrack, course_id)
+
+        extr_grade_mock.reset_mock()
```
singingwolfboy: If you have to call `reset_mock()`, you're trying to test too much in one test function. Please split this into two different tests.

dashboard/utils_test.py
```diff
@@ -363,9 +367,108 @@ def test_is_enrolled_mmtrack_fa(self):
...
-    def test_has_passed_course_normal(self):
+    @override_settings(FEATURES={"FINAL_GRADE_ALGO": "v1"})
+    @patch('dashboard.utils.MMTrack.extract_final_grade', autospec=True)
+    def test_has_passed_course(self, extr_grade_mock):
+        """
+        Test for has_passed_course method.
+        V1 new behavior soon to be default
+        """
+        course_id = "course-v1:odl+FOO101+CR-FALL15"
+        grade_mock = MagicMock(passed=True)
```
singingwolfboy: Why does this need to be a `MagicMock` rather than a regular `Mock`?

dashboard/utils_test.py
```diff
@@ -363,9 +367,108 @@ def test_is_enrolled_mmtrack_fa(self):
...
+        grade_mock = MagicMock(grade=0.57)
+        extr_grade_mock.return_value = grade_mock
+
+        mmtrack = MMTrack(
+            user=self.user,
+            program=self.program,
+            edx_user_data=self.cached_edx_user_data
+        )
+
+        assert round(mmtrack.get_final_grade(course_id)) == 57.0
```
singingwolfboy: Why are you calling `round()` here?

dashboard/utils_test.py
```diff
@@ -363,9 +367,108 @@ def test_is_enrolled_mmtrack_fa(self):
...
+        extr_grade_mock.assert_called_once_with(mmtrack, course_id)
+
+        extr_grade_mock.reset_mock()
```
singingwolfboy: Please split these tests.

dashboard/utils_test.py
```diff
@@ -363,9 +367,108 @@ def test_is_enrolled_mmtrack_fa(self):
...
+
+        mmtrack = MMTrack(
+            user=self.user,
+            program=self.program,
+            edx_user_data=self.cached_edx_user_data
+        )
+
+        res = mmtrack.extract_final_grade(course_id)
+        assert res.grade == grade_mock.grade
+        assert res.passed == grade_mock.passed
```
singingwolfboy: Please compare with real values, rather than variables.
```python
assert res.grade == 0.97
assert res.passed == True
```

dashboard/utils_test.py
```diff
@@ -363,9 +367,108 @@ def test_is_enrolled_mmtrack_fa(self):
...
+        get_grades_mock.assert_called_once_with(
+            course_run__edx_course_key=course_id,
+            status=FinalGradeStatus.COMPLETE,
+            user=self.user
+        )
+        get_grades_mock.reset_mock()
```
singingwolfboy: Please split these tests.

dashboard/api_test.py
```diff
@@ -458,22 +457,21 @@ def test_not_paid_in_past_grade_frozen(self, froz_mock, upgr_mock, final_grades_
         )
 
         # case when the user has not an entry in the frozen grades
-        final_grades_mock.side_effect = FinalGrade.DoesNotExist
+        self.mmtrack.configure_mock(**{'extract_final_grade.side_effect': FinalGrade.DoesNotExist})
```
singingwolfboy: This can be more clearly written as:
```python
self.mmtrack.extract_final_grade.side_effect = FinalGrade.DoesNotExist
```

You never need to use `configure_mock()` if you're only setting one thing in the mock. Just set it directly. (This also applies for all the `configure_mock()` calls below.)

dashboard/utils.py
giocalitri: The reason why I use a name tuple and not a `FinalGrade` object directly is because I cache the object and I thought it is not a good idea to cache the object from the database.

giocalitri: I will add this information as comment in the code

dashboard/utils.py
giocalitri: ok

dashboard/utils_test.py
giocalitri: No reason: changing it

dashboard/utils_test.py
giocalitri: because `0.57 * 100` in `get_final_grade` returns `56.9999...`
I talked to George and we agreed that we should leave the code as is int he function (the front end takes care of it) and round in the test.

singingwolfboy: Can you add that explanation as a comment in the code?

dashboard/utils.py
giocalitri: using the django cache frameworks adds a roundtrip to the backend system and I am not sure it is worth it if we move to a not in memory cache. (now it is in memory cache)
not sure: your call4

dashboard/utils_test.py
giocalitri: sure

dashboard/api_test.py
giocalitri: ok

dashboard/utils_test.py
giocalitri: ok

dashboard/utils_test.py
giocalitri: ok

dashboard/utils_test.py
giocalitri: ok

dashboard/utils_test.py
giocalitri: I left the following assertion because I want to test that there is only one call to the database, so I need to make 2 calls anyway, but I put the test with the exception in another test

giocalitri:
I think I addressed or replied to all the comments.

dashboard/utils.py
singingwolfboy: You can use a [local memory cache with Django.](https://docs.djangoproject.com/en/1.10/topics/cache/#local-memory-caching) That seems like the best of both worlds.

dashboard/utils_test.py
singingwolfboy: OK, that's reasonable.

dashboard/utils.py
giocalitri: that is what we use right now as backend for the cache, but if we switch to another backend it will mean that we will need to make another roundtrip to the other backend for this data and, if this is the case, I prefer to rely on the database and the automatic django query caching.

giocalitri:
@singingwolfboy I just removed the cache: can you please review again?

Issue #2134, opened 2016-12-15
Profile: ask users with non-Latin names to enter a Latin first name and last name


pdpinch:
Part of #1821 and #1796 

Pearson's systems require that users first and last name be delivered in a Roman character set (CP-252)

We will ask for first name and last name as we do now, and will allow users to enter this in UTF-8 characters as we do currently. If the user enters a non-latin name, we will expand an area under the name fields to explicitly ask for their name in “latin characters” (CP-1252). (from https://goo.gl/r2qznf) 

![image](https://cloud.githubusercontent.com/assets/430126/21227388/bc690104-c2a7-11e6-9018-531a446a4ec2.png)

### Acceptance criteria

- [x] Must work in both initial profile creation (/profile/personal) and in profile editing (/learner)
- [x] Data should be stored in the `romanized_first_name` and `romanized_last_name` fields of `Profile`.

PR #2215, opened 2016-12-28
Ask users with non-Latin names to enter a Latin first name and last name

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/2134

#### What's this PR do?
- If first name or last name is not in cp-1252 ask user to add romanized name
- romanized name can only be in cp-1252. Added validation for it.

@pdpinch 
screen shot 2017-01-03 at 4 00 53 pm

### DIFF
--- a/static/js/components/PersonalForm.js
+++ b/static/js/components/PersonalForm.js
@@ -39,6 +39,37 @@ export default class PersonalForm extends ProfileFormFields {
     ui:                     UIState,
   };
 
+  renderRomanizedFields = () => {
+    const { profile } =  this.props;
+    const english = /^[A-Za-z]*$/;
+    let firstName = _.get(profile, ["first_name"], "");
+    let lastName = _.get(profile, ["last_name"], "");
+    let rFirstName = _.get(profile, ["romanized_first_name"], "");
+    let rLastName = _.get(profile, ["romanized_last_name"], "");
+    const showRomanizedFields = (
+      !english.test(firstName) || !english.test(lastName) || rFirstName || rLastName
+    );
+
+    if (showRomanizedFields) {
+      return (
+        <Cell col={12}>
+          <section className="romanized-name">
+            <h3>Plese enter your name in Latin characters</h3>
+            <Grid className="profile-form-grid">
+              <Cell col={6}>
+                {this.boundTextField(["romanized_first_name"], "Given name")}
+              </Cell>
+              <Cell col={6}>
+                {this.boundTextField(["romanized_last_name"], "Family name")}
+              </Cell>
+            </Grid>
+          </section>
+        </Cell>
+      );
+    }
+    return null;
+  }
+
   render() {
     const whyWeAskThis = 'Some program sponsors and employers offer benefits or scholarships ' +
       'to learners with specific backgrounds.';
@@ -57,6 +88,7 @@ export default class PersonalForm extends ProfileFormFields {
           <Cell col={6}>
             {this.boundTextField(["last_name"], "Family name")}
           </Cell>
+          {this.renderRomanizedFields()}
           <Cell col={12}>
             {this.boundTextField(["preferred_name"], "Nickname / Preferred name")}
           </Cell>
--- a/static/js/components/PersonalForm.js
+++ b/static/js/components/PersonalForm.js
@@ -9,6 +9,7 @@ import SelectField from './inputs/SelectField';
 import CountrySelectField from './inputs/CountrySelectField';
 import StateSelectField from './inputs/StateSelectField';
 import ProfileFormFields from '../util/ProfileFormFields';
+import { showRomanizedFields } from '../util/profile_edit';
 import type {
   Profile,
   SaveProfileFunc,
@@ -41,16 +42,7 @@ export default class PersonalForm extends ProfileFormFields {
 
   renderRomanizedFields = () => {
     const { profile } =  this.props;
-    const english = /^[A-Za-z]*$/;
-    let firstName = _.get(profile, ["first_name"], "");
-    let lastName = _.get(profile, ["last_name"], "");
-    let rFirstName = _.get(profile, ["romanized_first_name"], "");
-    let rLastName = _.get(profile, ["romanized_last_name"], "");
-    const showRomanizedFields = (
-      !english.test(firstName) || !english.test(lastName) || rFirstName || rLastName
-    );
-
-    if (showRomanizedFields) {
+    if (showRomanizedFields(profile)) {
       return (
         <Cell col={12}>
           <section className="romanized-name">
--- a/static/js/components/PersonalForm.js
+++ b/static/js/components/PersonalForm.js
@@ -9,7 +9,7 @@ import SelectField from './inputs/SelectField';
 import CountrySelectField from './inputs/CountrySelectField';
 import StateSelectField from './inputs/StateSelectField';
 import ProfileFormFields from '../util/ProfileFormFields';
-import { showRomanizedFields } from '../util/profile_edit';
+import { shouldRenderRomanizedFields } from '../util/profile_edit';
 import type {
   Profile,
   SaveProfileFunc,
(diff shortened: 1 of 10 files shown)


noisecapella:
I think `latin` needs to be capitalized here

static/js/components/PersonalForm.js
```diff
@@ -39,6 +39,37 @@ export default class PersonalForm extends ProfileFormFields {
     ui:                     UIState,
   };
 
+  renderRomanizedFields = () => {
+    const { profile } =  this.props;
+    const english = /^[A-Za-z0-9]*$/;
```
rhysyngsun: I think we want to allow characters that fit in CP-1252 beyond the alphanumeric range so that we can accept most European names without the need for the user to convert. That's roughly the ranges `0x0020` - `0x007F` and `0x00A0` - `0x00FF` in Unicode (the gap at the begging and in the middle excludes control characters and some characters that map oddly into unicode). There are also a few non-punctuation characters in the 128-159 range of CP-1252 (the middle gap) that might be worth including as well.

For more details: https://en.wikipedia.org/wiki/Windows-1252#Code_page_layout

amir-qayyum-khan: right now i am doing same, if name is not english then i am showing input box to add latin char names

static/js/components/PersonalForm.js
```diff
@@ -39,6 +39,37 @@ export default class PersonalForm extends ProfileFormFields {
...
 
+  renderRomanizedFields = () => {
+    const { profile } =  this.props;
+    const english = /^[A-Za-z]*$/;
+    let firstName = _.get(profile, ["first_name"], "");
+    let lastName = _.get(profile, ["last_name"], "");
+    let rFirstName = _.get(profile, ["romanized_first_name"], "");
+    let rLastName = _.get(profile, ["romanized_last_name"], "");
+    const showRomanizedFields = (
+      !english.test(firstName) || !english.test(lastName) || rFirstName || rLastName
```
amir-qayyum-khan: @rhysyngsun here if first name or last name is not in english then asking user to Romanized it or use already have Romanized first and last name

rhysyngsun: This code looks fine to me - my comment before was regarding the regex being used, it only allows lowercase/uppercase characters a-z whereas the original issue requested characters in the CP-1252 range (which is what Pearson requires), which includes many characters with diacritical marks. For example, the last name `Müller` is a common last name in Germany and fits in CP-1252, but doesn't pass the regex above.

My previous comment detailed the Unicode ranges I think are worth validating against.

static/js/components/PersonalForm.js
```diff
@@ -39,6 +39,37 @@ export default class PersonalForm extends ProfileFormFields {
...
+    );
+
+    if (showRomanizedFields) {
+      return (
+        <Cell col={12}>
+          <section className="romanized-name">
+            <h3>Plese enter your name in Latin characters</h3>
+            <Grid className="profile-form-grid">
+              <Cell col={6}>
+                {this.boundTextField(["romanized_first_name"], "Given name")}
```
rhysyngsun: These fields need some validation to ensure the user doesn't enter non-CP-1252 characters in them.

static/js/components/PersonalForm.js
amir-qayyum-khan: for names like Müller user have to provide romanized alternative

rhysyngsun: My understanding was that `Müller` was acceptable for Latin characters, specifically the `ü` falls within the CP-1252 character set that the ticket specifies so it shouldn't be required to convert that to a regular `u`.

Maybe @pdpinch can clarify? Do we want the romanized names to only be in the range of `a-zA-Z`?

static/js/components/PersonalForm.js
amir-qayyum-khan: Done added validation @rhysyngsun

amir-qayyum-khan:
@rhysyngsun done with (cp1252)[http://www.cp1252.com/] validation integration.

static/js/constants.js
```diff
@@ -12,6 +12,8 @@ export const ISO_8601_FORMAT = 'YYYY-MM-DD';
 export const DASHBOARD_FORMAT = 'M/D/Y';
 export const DASHBOARD_MONTH_FORMAT = 'MM[/]YYYY';
 
+export const CP1252_REGIX = /^[\u0020-\u00FF]*$/;
```
singingwolfboy: Please name this `CP1252_REGEX` instead of `CP1252_REGIX`. "Regex" is short for "regular expression". I don't know what "regix" means.

static/js/components/PersonalForm.js
```diff
@@ -39,6 +40,28 @@ export default class PersonalForm extends ProfileFormFields {
     ui:                     UIState,
   };
 
+  renderRomanizedFields = () => {
+    const { profile } =  this.props;
+    if (showRomanizedFields(profile)) {
+      return (
+        <Cell col={12}>
+          <section className="romanized-name">
+            <h3>Plese enter your name in Latin characters</h3>
```
singingwolfboy: Typo: "Please", not "Plese". Also, I'm not sure that everyone will know what "Latin characters" are. Maybe we should provide some examples, and an explanation for why we are asking this. I would recommend:

> Please enter your name using only Latin characters. For example, convert "è" to "e", "ä" to "a", "ñ" to "n", and so on. This is necessary because some of our systems can only handle Latin characters.

static/js/components/PersonalForm.js
```diff
@@ -39,6 +40,28 @@ export default class PersonalForm extends ProfileFormFields {
     ui:                     UIState,
   };
 
+  renderRomanizedFields = () => {
+    const { profile } =  this.props;
+    if (showRomanizedFields(profile)) {
```
singingwolfboy: The name for this function is a bit ambiguous: it's called `renderRomanizedFields`, but if the Romanized fields shouldn't be shown, then the function won't actually render anything at all. Could you split this up into two functions, like this?

```js
renderRomanizedFields = () => {
  return (
    <Cell col={12}>
      ...
    </Cell>
  );
}

maybeRenderRomanizedFields = () => {
  if (shouldRenderRomanizedFields(this.props.profile) {
    return renderRomanizedFields();
  } else {
    return null;
  }
}
```

Also, `showRomanizedFields` and `renderRomanizedFields` sound like they do exactly the same thing. I would also recommend renaming `showRomanizedFields` to `shouldRenderRomanizedFields`, to make the distinction clearer.

static/js/util/profile_edit.js
```diff
@@ -236,3 +237,16 @@ export function saveProfileStep(validator: Validator|UIValidator, isLastStep: bo
...
+ * Returns true when first name or last name has non CP-1252 string(s).
+ */
+export function showRomanizedFields(profile: Profile): boolean {
+  let firstName = _.get(profile, ["first_name"], "");
+  let lastName = _.get(profile, ["last_name"], "");
+
+  if (!CP1252_REGIX.test(firstName) || !CP1252_REGIX.test(lastName)) {
+    return true;
+  }
+  return false;
```
singingwolfboy: This function could be greatly simplified:

```js
export function shouldRenderRomanizedFields(profile: Profile): boolean {
  const { first_name: firstName, last_name: lastName } = profile;
  return CP1252_REGEX.test(firstName) || CP1252_REGEX.test(lastName);
}
```

static/js/components/PersonalForm.js
singingwolfboy: Actually, now that I've read [the Wikipedia page for CP-1252](https://en.wikipedia.org/wiki/Windows-1252), looks like my examples are incorrect: è, ä, and ñ are all present in that character set. Maybe we can't provide good examples, after all...

amir-qayyum-khan:
Done @singingwolfboy

static/js/util/profile_edit.js
amir-qayyum-khan: ``` const { first_name: firstName, last_name: lastName } = profile;``` this cases issues in personal tab tests where profile is none.

static/js/components/PersonalForm.js
amir-qayyum-khan: @roberthouse54 any thoughts?

static/js/lib/validation/profile.js
```diff
@@ -78,6 +80,15 @@ export const personalValidation = (profile: Profile) => {
   if (!moment(profile.date_of_birth).isBefore(moment(), 'day')) {
     errors.date_of_birth = personalMessages.date_of_birth;
   }
+  if (shouldRenderRomanizedFields(profile)) {
+    if (!profile.romanized_first_name || !CP1252_REGEX.test(profile.romanized_first_name)) {
+      errors.romanized_first_name = "CP-1252 string is required";
```
rhysyngsun: I think this needs a more human-friendly error message; most people won't understand what they need to do.

Maybe something like "Latin first name is required" (doesn't sound the best to me, but I think it's clearer). Maybe different error messages for blank vs. invalid characters would alleviate that. @roberthouse54 any thoughts?

static/js/lib/validation/profile_test.js
```diff
@@ -86,6 +86,57 @@ describe('Profile validation functions', () => {
...
+      assert.deepEqual(personalValidation(profile), errors);
+    });
+
+    it('should error when romanized_last_name is non cp-1252', () => {
+      let profile = {
+        ...USER_PROFILE_RESPONSE,
+        first_name: 'عامر',
+        romanized_first_name: 'test',
+        romanized_last_name: 'عامر'
+
```
rhysyngsun: Nit: extra space

static/js/lib/validation/profile.js
roberthouse54: Yes, let's use "Latin first name is required" etc.

roberthouse54: I think we can use the same error message in both cases (blank and invalid characters)

amir-qayyum-khan:
Done @rhysyngsun @roberthouse54

rhysyngsun:
@amir-qayyum-khan I think the tests for the error messages will need to be fixed too.

amir-qayyum-khan:
Done @rhysyngsun

static/js/components/PersonalForm.js
roberthouse54: The CP-1252 character set is hard to describe, so I think we should avoid trying to be overly specific here. Just use the text in the mock.

rhysyngsun:
:+1:

PR #2217, opened 2016-12-28
Check for OSError during image migration

### LINKED ISSUE #2216: Fix handling of broken image URLs during thumbnail migration

noisecapella:
Fixes #2216

#### What's this PR do?
Silences `OSError` if it occurs during an image thumbnail conversion. In this case it leaves the `image_small` field `null` and moves on to the next one.


rhysyngsun:
:+1:

PR #2218, opened 2016-12-28
Added medium sized image to profile model and REST API

### LINKED ISSUE #2213: Make medium size profile images
This should be very similar to the work done in #2146

noisecapella:
Fixes #2213

#### What's this PR do?
Makes `image_medium` for the `Profile` model and the `/api/v0/profiles/.../` REST API. 

Run the migrations then verify that you see a new `image_medium` URL in `/api/v0/profiles/<username_here>/` and that it is 128x128 pixels. 

@roberthouse54 What pixel dimensions should this image have?


noisecapella:
If both dimensions of an image are smaller than 128, the image size will not change. The image will not be resized bigger than it is

I'm going to wait for @roberthouse54's opinion on if 128 is a good value

PR #2220, opened 2016-12-28
Implement lazy loading for UserChip

noisecapella:
Part of #2213

#### What's this PR do?
Removes CSS hover rule for user chip and replaces it with explicit state in redux. This means the DOM elements are only mounted when the user hovers over the name, which means the image on the UserChip is only loaded on first mouseenter.

Make sure hovering over the name still produces the user chip. Hovering over the user chip itself should also keep the user chip visible.


rhysyngsun:
:+1:

PR #2221, opened 2016-12-29
Export exam authorizations to Pearson

### LINKED ISSUE #2076: Exams: Export exam authorizations to Pearson
In order to allow a user to schedule an exam we need to SFTP an exam authorization file (EAD) to Pearson containing rows for each user authorized. This is dependent on the preceding steps #1795 and #1797 having been completed for the given user's profile.

- [x] Data model to track exam authorization state
- [x] Method to create the EAD file locally
  - `ExamSeriesCode` will be tied to the program most likely *(waiting on verification)*
  - We **will** be using modules, they will be associated with the course. Waiting on confirmation of the values, but we we just store as an opaque string with maybe some light validation on the admin ui.
- [x] Celery task to generate and upload the EAD file

Open Questions:

- [x] What is the trigger on our end for a user being authorized? Do we currently track this somewhere?
- [x] I believe we need a flag on the `Course` record to signify that the course requires an exam. We also need to map the course to a Pearson exam code. Is there a case for not having these two be the same (i.e. presence of an exam code means the course has an exam)? I can't think of any. There's probably an argument for *not* putting this field on `Course` as well because it is Pearson-specific.
- [x] Pearson requires a range of eligibility dates (start/end) for authorizations. These can be updated, but we need a sensible default. This probably has some business logic implications (i.e. if someone took a course 5 years ago would we want to allow them to test for that?)
- [x] Are we supporting exam accommodations? If so, how is this expected to work? At the course level or per-user or both?

Deadline: 12/16/16

rhysyngsun:

Fixes #2076 

#### What's this PR do?

This PR implements exam authorization for Pearson. This involves generating a TSV file and uploading it to their FTP server. This builds on TSV writer and SFTP upload code written in PR #2060 for issue #1795.

Also cleaned up some of the tests around existing Pearson code so they're more thorough / legible.

If you review this, please ping me and I'll share the detailed docs on their API (I believe they're confidential so can't be shared here).

exams/pearson.py

In a Django shell:

```python
    from factory.django import mute_signals
    from exams.factories import ExamProfileFactory
    from exams.tasks import export_exam_authorizations
    OFFSET=1200
    with mute_signals(post_save):
        for x in range(50):
            profile = ProfileFactory.create(user____sequence=OFFSET+x)
            ExamAuthorizationFactory.create(user=profile.user)
    export_exam_authorizations()
```

Then in another terminal session session: `sftp -P 2022 odl@localhost:share` (password: 123).

You can `ls` to see the file that was exported and use `get` to download it to your host working directory so you can inspect the contents.


courses/factories.py
```diff
@@ -17,6 +17,7 @@ class ProgramFactory(DjangoModelFactory):
     title = fuzzy.FuzzyText(prefix="Program ")
     live = fuzzy.FuzzyAttribute(FAKE.boolean)
     description = fuzzy.FuzzyText()
+    exam_series_code = factory.LazyAttribute(lambda _: "DEDP_MicroMasters")
```
rhysyngsun: Good catch, probably a holdover of my trying to generate fuzzy values for this.

exams/pearson.py
```diff
@@ -23,6 +23,7 @@
     delimiter='\t',
 )
 
+PEARSON_DATE_FORMAT = "%Y/%m/%d"
```
giocalitri: you seem to have a lot of constants in this module. Consider move them to a different `constants.py` module.

exams/pearson.py
```diff
@@ -141,6 +149,19 @@ def _profile_country_to_alpha3(profile):
 ], field_prefix='profile')
 
 
+write_ead_file = _tsv_writer([
+    ('AuthorizationTransactionType', 'operation'),
+    ('ClientAuthorizationID', 'id'),
+    ('ClientCandidateID', 'user.profile.student_id'),
+    ('ExamSeriesCode', 'course.program.exam_series_code'),
+    ('Modules', 'course.exam_module'),
+    ('Accommodations', lambda _: ''),
```
giocalitri: why this and not just `('Accommodations', ''),` ? (not sure I understood the reason for this code)

exams/pearson.py
```diff
@@ -141,6 +149,19 @@ def _profile_country_to_alpha3(profile):
...
+    ('EligibilityApptDateFirst', lambda exam_auth: _format_date(exam_auth.date_first_eligible)),
+    ('EligibilityApptDateLast', lambda exam_auth: _format_date(exam_auth.date_last_eligible)),
+    ('LastUpdate', lambda exam_auth: _format_datetime(exam_auth.updated_on)),
+])
```
giocalitri: I think you should move this code inside a function like 
```
def get_ead_file_writer(<args>):
    return _tsv_writer([
         ...
```
In this way it is a bit more clear that you get a configured writer. Plus I am not convinced that this code should be executed at import time. 
If I am missing something let me know.

courses/factories.py
```diff
@@ -17,6 +17,7 @@ class ProgramFactory(DjangoModelFactory):
     title = fuzzy.FuzzyText(prefix="Program ")
     live = fuzzy.FuzzyAttribute(FAKE.boolean)
     description = fuzzy.FuzzyText()
+    exam_series_code = "DEDP_MicroMasters"
```
giocalitri: I do not think this should have this constant: maybe a fuzzy text is better or if you really need a static text, choose something more generic.

courses/factories.py
```diff
@@ -56,6 +57,7 @@ class CourseFactory(DjangoModelFactory):
 
     description = fuzzy.FuzzyText()
     prerequisites = fuzzy.FuzzyText(prefix="Course requires ")
+    exam_module = "14.73x"
```
giocalitri: same comment as before. Try to build a fake code using `fuzzy`

courses/models.py
```diff
@@ -26,6 +26,7 @@ class Program(models.Model):
     description = models.TextField(blank=True, null=True)
     financial_aid_availability = models.BooleanField(default=False, null=False)
     ga_tracking_id = models.CharField(max_length=255, blank=True, default="")
+    exam_series_code = models.CharField(max_length=20, null=True, blank=True)
```
giocalitri: please add a comment explaining what this field is

exams/pearson.py
rhysyngsun: Strings here are treated as property paths, so `''` would currently trigger an error. Pearson wanted us to add this column, but we don't have a defined use case for it yet, so it gets left blank.  `_writer` takes either a string property path or a callable here hence the lambda returning an empty string. An alternative would be to update `_writer` to recognize `None` values here, then the csv writer would convert `None` to `''`.

courses/models.py
```diff
@@ -65,6 +66,7 @@ class Course(models.Model):
     thumbnail = models.ImageField(null=True, blank=True)
     description = models.TextField(blank=True, null=True)
     prerequisites = models.TextField(blank=True, null=True)
+    exam_module = models.CharField(max_length=20, null=True, blank=True)
```
giocalitri: please add a comment explaining what this field is

exams/pearson.py
rhysyngsun: I had the same thought, but have done it in a separate branch that includes a refactor of all the pearson code into its own module: https://github.com/mitodl/micromasters/compare/nl/2076-exam-auth-exports...nl/refactor-pearson-to-module

I'm happy to include those changes in this PR, but I had figured a separate PR just to refactor was easier on reviewers.

exams/factories.py
```diff
@@ -20,3 +32,27 @@ class ExamProfileFactory(DjangoModelFactory):
...
+    course = SubFactory(CourseFactory)
+
+    operation = FuzzyChoice(
+        [value[0] for value in ExamAuthorization.OPERATION_CHOICES]
+    )
+    status = FuzzyChoice(
+        [value[0] for value in ExamAuthorization.STATUS_CHOICES]
+    )
+    date_first_eligible = factory.LazyAttribute(
+        lambda x: FAKE.date_time_this_year(before_now=False, after_now=True, tzinfo=pytz.utc).date()
```
giocalitri: isn't `tzinfo=pytz.UTC` ? (maybe it is equivalent)

exams/factories.py
```diff
@@ -20,3 +32,27 @@ class ExamProfileFactory(DjangoModelFactory):
...
+    )
+    date_last_eligible = factory.LazyAttribute(
+        lambda x: x.date_first_eligible + timedelta(days=10)
```
giocalitri: please use a more clear variable and not `x`

exams/models.py
```diff
@@ -35,3 +38,61 @@ class ExamProfile(Model):
...
+    Tracks state of an exam authorization
+    """
+    OPERATION_ADD = 'add'
+    OPERATION_DELETE = 'delete'
+    OPERATION_UPDATE = 'update'
+
+    OPERATION_CHOICES = (
+        (OPERATION_ADD, 'Add'),
+        (OPERATION_DELETE, 'Update'),
+        (OPERATION_UPDATE, 'Delete'),
```
giocalitri: what about `.capitalize()` for these 3 constants?

courses/models.py
rhysyngsun: Done

exams/models.py
```diff
@@ -35,3 +38,61 @@ class ExamProfile(Model):
 
     def __str__(self):
         return 'Exam Profile "{0}" with status "{1}"'.format(self.id, self.status)
+
+
+class ExamAuthorization(TimestampedModel):
+    """
+    Tracks state of an exam authorization
```
giocalitri: Nit: what is an exam authorization? (I might just need to read some docs, but a better overview would be nice)

exams/factories.py
rhysyngsun: We have both usages in our code, are we standardizing towards one over the other? https://github.com/mitodl/micromasters/search?utf8=%E2%9C%93&q=pytz.utc

exams/models.py
```diff
@@ -35,3 +38,61 @@ class ExamProfile(Model):
...
+        (STATUS_PENDING, 'Sync Pending'),
+        (STATUS_IN_PROGRESS, 'Sync in Progress'),
+        (STATUS_FAILED, 'Sync Failed'),
+        (STATUS_SUCCESS, 'Sync Suceeded'),
+    )
+
+    user = models.ForeignKey(
+        User,
+        related_name='exam_authorizations'
+    )
```
giocalitri: for this field and the following: does this not fit in one line?

exams/models.py
rhysyngsun: It's the record we need to send to Pearson authorizing an individual for a specific exam. This model exists so we can track state on our side.

exams/models.py
```diff
@@ -35,3 +38,61 @@ class ExamProfile(Model):
...
+
+    course = models.ForeignKey(
+        'courses.Course',
+        related_name='exam_authorizations'
+    )
+
+    operation = models.CharField(
+        max_length=30,
+        null=False,
+        choices=OPERATION_CHOICES
```
giocalitri: No default?

exams/models.py
```diff
@@ -35,3 +38,61 @@ class ExamProfile(Model):
...
+    )
+    status = models.CharField(
+        max_length=30,
+        null=False,
+        choices=STATUS_CHOICES
```
giocalitri: No default?

exams/models_test.py
```diff
@@ -16,3 +19,15 @@ def test_exam_profile_str(self):  # pylint: disable=no-self-use
...
+class ExamAuthorizationTest(TestCase):
+    """Tests for ExamAuthorizationss"""
+
+    def test_exam_authorization_str(self):  # pylint: disable=no-self-use
+        """
+        Test method ExamAuthorization.__str__ prints correctly
+        """
+        auth = ExamAuthorization(id=1, status=ExamProfile.PROFILE_IN_PROGRESS, user_id=2)
+
+        assert str(auth) == 'Exam Authorization "1" with status "in-progress" for user 2'
```
giocalitri: nit: I personally think that we do not need to test the `__str__` method, but feel free to leave

exams/models.py
rhysyngsun: Done

exams/models.py
rhysyngsun: Done

exams/models.py
rhysyngsun: Done

exams/pearson_test.py
```diff
@@ -59,29 +62,47 @@ def test_profile_country_to_alpha3_invalid_country(self):  # pylint: disable=no-
...
+    def tsv_lines(self):  # pylint: disable=missing-docstring
+        return self.tsv_value.splitlines()
+
+    @property
+    def tsv_header(self):  # pylint: disable=missing-docstring
+        return self.tsv_lines[0]
+
+    @property
+    def tsv_rows(self):  # pylint: disable=missing-docstring
+        return self.tsv_lines[1:]
```
giocalitri: Please add a (even minimal) docstring for these functions

exams/pearson_test.py
```diff
@@ -1,32 +1,35 @@
 """
 Tests for exams.pearson module
 """
-from datetime import datetime
+from datetime import date, datetime
 from unittest import mock
 from unittest.mock import patch
 
 import io
```
giocalitri: nit:why this is separated from the other groups of imports?

exams/pearson_test.py
```diff
@@ -99,16 +120,11 @@ def test_write_cdd_file(self):  # pylint: disable=no-self-use
 
         with mute_signals(post_save):
             exam_profiles = [ExamProfileFactory.create(**kwargs)]
+            exam_profiles[0].profile.updated_on = datetime(2016, 5, 15, 15, 2, 55)
```
giocalitri: why `exam_profiles[0].profile.updated_on = datetime(2016, 5, 15, 15, 2, 55)`?

rhysyngsun: `updated_on` is `auto_now=True` so I wrote the test to set it afterwards since the value in `kwargs` is discarded.

exams/tasks.py
```diff
@@ -55,3 +55,37 @@ def export_exam_profiles():
...
+
+
+@async.task
+def export_exam_authorizations():
+    """
+    Sync any outstanding profiles
+    """
+    exam_authorizations = (ExamAuthorization.objects
+                           .filter(status=ExamAuthorization.STATUS_PENDING)
+                           .prefetch_related('user__profile', 'course__program'))
```
giocalitri: nit: this can be written like
```
exam_authorizations = ExamAuthorization.objects.filter(
    status=ExamAuthorization.STATUS_PENDING).prefetch_related('user__profile', 'course__program')
```

exams/tasks.py
```diff
@@ -55,3 +55,37 @@ def export_exam_profiles():
...
+    file_prefix = datetime.now(pytz.utc).strftime('ead-%Y%m%d%H_')
```
giocalitri: why this prefix is not in a constant somewhere?

exams/pearson_test.py
rhysyngsun: For me, it's easier to scan imports for where I need to edit/add an import. The style I'm used to, is imports sorted by PEP8, separated by `import` and `from .. import ..` styles. I haven't seen a consistent organization of our imports aside from minimal PEP8 adherence (sometimes I've found imports to be in jumbled order within import blocks), but I'm happy to make adjustments to the style I'm writing in to conform to the project.

Do we have a prescribed import organization method that's more specific than PEP8?

exams/tasks.py
```diff
@@ -55,3 +55,37 @@ def export_exam_profiles():
...
+
+        # upload to SFTP server
+        pearson.upload_tsv(tsv.name)
+
+    # update records to reflect the successful upload
+    if valid_auths:
+        exam_auth_ids = [exam_auth.id for exam_auth in valid_auths]
+        (ExamAuthorization.objects
+         .filter(id__in=exam_auth_ids)
+         .update(status=ExamAuthorization.STATUS_IN_PROGRESS))
```
giocalitri: nit: please update this statement to not use parenthesis

exams/pearson_test.py
rhysyngsun: Done

exams/tasks.py
giocalitri: what if there is an error at any point? Shouldn't we log something?

exams/tasks_test.py
```diff
@@ -1,57 +1,136 @@
...
         Verify that export_exam_profiles makes calls to export the pending profiles
         """
-        from exams.tasks import export_exam_profiles
+        assert ExamProfile.objects.count() == 10
 
-        with mute_signals(post_save):
-            exam_profiles = ExamProfileFactory.create_batch(10, status=ExamProfile.PROFILE_PENDING)
+        # use side_effect to assert at call-time because query return values
+        # mutate by the time export_exam_profiles returns
+        def side_effect(tsv, qs):  # pylint: disable=missing-docstring
```
giocalitri: I think you can use the comment above as docstring

exams/tasks_test.py
```diff
@@ -1,57 +1,136 @@
...
-        assert hasattr(cdd_writer_mock.call_args[0][0], 'write')  # was first arg a file-like object?
-        assert isinstance(cdd_writer_mock.call_args[0][1], QuerySet)
-        assert all([isinstance(ep, ExamProfile) for ep in cdd_writer_mock.args[0][1]])
+        assert write_cdd_file_mock.call_count == 1
+
+        invalid_profiles = ExamProfile.objects.filter(status=ExamProfile.PROFILE_INVALID)
+        in_progress_profiles = ExamProfile.objects.filter(status=ExamProfile.PROFILE_IN_PROGRESS)
+
+        assert len(invalid_profiles) == 5
+        assert len(in_progress_profiles) == 5
```
giocalitri: nit: you can use `.count()` given that the variables are query sets

exams/tasks_test.py
```diff
@@ -1,57 +1,136 @@
...
+
+        # invalid profiles should be marked as such
+        for exam_profile in self.invalid_profiles:
+            assert exam_profile in invalid_profiles
```
giocalitri: nit: maybe a different name for the local variable could be more clear

exams/tasks_test.py
rhysyngsun: Done

exams/tasks_test.py
```diff
@@ -1,57 +1,136 @@
...
+    def setUpTestData(cls):
+        with mute_signals(post_save):
+            cls.exam_auths = ExamAuthorizationFactory.create_batch(10, status=ExamAuthorization.STATUS_PENDING)
+
+    @patch('exams.pearson.upload_tsv')
+    def test_export_exam_authorizations(self, upload_tsv_mock):  # pylint: disable=no-self-use
+        """
+        Verify that export_exam_authorizations exports pending exam auths
+        """
+        assert ExamAuthorization.objects.count() == 10
```
giocalitri: what is `10`?

exams/tasks_test.py
```diff
@@ -1,57 +1,136 @@
...
+
+        # use side_effect to assert at call-time because query return values
+        # mutate by the time export_exam_authorizations returns
+        def side_effect(tsv, qs):  # pylint: disable=missing-docstring
```
giocalitri: I think you can use the above comment as docstring

exams/pearson.py
giocalitri: No, if you have it in another PR it is fine to leave it there

exams/pearson.py
giocalitri: makes sense

exams/factories.py
giocalitri: IF they are equivalent, then I do not care, otherwise we should use `pytz.UTC`

exams/tasks.py
rhysyngsun: Done

exams/tasks.py
rhysyngsun: Done

exams/models.py
giocalitri: Perfect: please update the docstring with your comment

exams/tasks.py
rhysyngsun: Done

exams/pearson_test.py
giocalitri: I usually group together the `import ...` and the `from ... import ... ` for the same groups. We do not have a preference about whether to put `import ...` first or `from ... import ... ` but I think we are pretty consistent about the grouping together without spaces.
Anyway, this is super minor nit

exams/tasks_test.py
rhysyngsun: Done

exams/tasks_test.py
rhysyngsun: Done and ditto for the other test.

exams/pearson_test.py
giocalitri: then please move  `datetime(2016, 5, 15, 15, 2, 55)` to a variable or just use `.now(tz=pytz.UTC)`

giocalitri:
I did a first pass and the code looks overall in a pretty good shape.

exams/tasks_test.py
rhysyngsun: Prefixed the instance attribute with `expected_`.

exams/factories.py
rhysyngsun: Fixed

exams/tasks_test.py
rhysyngsun: I'm asserting that only the 10 authorizations created in this test's setup are present. That's probably overly defensive and if some bad state has somehow leaked the other assertions should catch that. I'll remove it.

exams/models.py
rhysyngsun: Done

exams/factories.py
rhysyngsun: Done

courses/models.py
rhysyngsun: Done

exams/pearson_test.py
rhysyngsun: Done

exams/pearson_test.py
rhysyngsun: I'll adjust them then, as much as I have my own preferences, I much more don't like having islands of personal code styles dotted though code.

exams/models.py
rhysyngsun: Done, ditto for the other existing model in this file.

exams/factories.py
rhysyngsun: They're the same: https://git.launchpad.net/pytz/tree/src/pytz/__init__.py#n245

courses/factories.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: I agree with that. Not being able to provide a docstring with the current style is another ding against that design.

courses/factories.py
rhysyngsun: Done

exams/factories.py
giocalitri: ok

exams/tasks_test.py
giocalitri: I think my comment is more about why exactly `10`: I think you should assert equal to the length of the elements in the setup or something else easily identifiable

rhysyngsun: Makes sense, adjusted for that.

exams/pearson.py
rhysyngsun: I'm probably going to go in the direction of converting these to classes instead because it's going to be easier to mock and reason about.

exams/pearson.py
```diff
@@ -55,90 +67,147 @@ def _get_field_mapper(field):
...
                 valid_rows.append(row)
             except InvalidTsvRow:
                 log.exception("Invalid tsv row")
                 invalid_rows.append(row)
 
         return (valid_rows, invalid_rows)
 
-    return _writer
 
+class CDDWriter(TSVWriter):
```
rhysyngsun: @giocalitri Here is the revised class-based writers that I refactored based on your comment [here](https://github.com/mitodl/micromasters/pull/2221#discussion_r94607920).

exams/tasks.py
```diff
@@ -2,56 +2,128 @@
...
+
+def _backoff(retries):
+    """
+    Exponential backoff for retried tasks
+    """
+    return settings.EXAMS_SFTP_BACKOFF_BASE ** retries
+
+
+@async.task(bind=True, max_retries=3)
+def export_exam_profiles(self):
```
rhysyngsun: @giocalitri I added some exception handling / logging here. In the case of certain exceptions that are retryable (`SSHException`, `ConnectionException`), those are re-reraised as `RetryableSFTPException` in `exams.pearson.upload_tsv` and trigger a task retry with exponential backoff. The default factor base of 5 will means these 3 reitries will spread over 2.5 minutes.

Otherwise, other exceptions will get logged and the task exits (and acks). I think that's acceptable because this task will eventually get run via celerybeat and the profiles that didn't get synced due to a previous error will sync on the next run (assuming the issue was either transient or resolved by us in the interim).

courses/factories.py
```diff
@@ -17,6 +17,7 @@ class ProgramFactory(DjangoModelFactory):
     title = fuzzy.FuzzyText(prefix="Program ")
     live = fuzzy.FuzzyAttribute(FAKE.boolean)
     description = fuzzy.FuzzyText()
+    exam_series_code = factory.LazyFunction(lambda: FAKE.lexify("????_MicroMasters"))
```
giocalitri: nice solution

exams/factories.py
```diff
@@ -1,13 +1,25 @@
 """
 Factories for exams
 """
+from datetime import timedelta
+import pytz
```
giocalitri: nit: this is not part of the standard library

exams/models.py
```diff
@@ -30,8 +36,65 @@ class ExamProfile(Model):
...
+    Tracks state of an exam authorization
+
+    Exam Authorization record are sent to Pearson authorizing an individual for a specific exam.
+    This model exists so we can track state on our side and corresponds to a row in a Pearson EAD file.
+    """
+    OPERATION_ADD = 'add'
+    OPERATION_DELETE = 'delete'
+    OPERATION_UPDATE = 'update'
+
+    OPERATION_CHOICES = tuple((op, op.capitalize()) for op in [
```
giocalitri: I am pretty sure you do not need to use `tuple` explicitly, but feel free to leave as is

exams/pearson.py
```diff
@@ -1,18 +1,22 @@
 """
 Pearson specific exam code
 """
-from operator import attrgetter
+from collections import OrderedDict
 import csv
 import logging
+from operator import attrgetter
```
giocalitri: nit, please group together the `import ...` and the `from ... import ...`

exams/pearson.py
```diff
@@ -1,18 +1,22 @@
...
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from paramiko import SSHException
 import pycountry
 import pysftp
+from pysftp.exceptions import ConnectionException
```
giocalitri: nit: please group together the imports

exams/pearson.py
```diff
@@ -55,90 +67,147 @@ def _get_field_mapper(field):
...
+        The second value is a str property path (e.g. "one.two.three") or
+        a callable that when passed a row returns a computed field value
+
+        Arguments:
+            fields (List): list of (str, str|callable) tuples
+            field_prefix (str): path prefix to prefix field lookups with
+        """
+        self.fields = OrderedDict(fields)
+        self.columns = self.fields.keys()
+        self.field_mappers = {column: _get_field_mapper(field) for column, field in self.fields.items()}
```
giocalitri: isn't `self.field_mappers` == the `fields` passed as argument to the `__init__`

exams/pearson.py
```diff
@@ -55,90 +67,147 @@ def _get_field_mapper(field):
...
+        Maps a row object to a row dict
+
+        Args:
+            row: the row to map to a dict
+
+        Returns:
+            dict:
+                row mapped to a dict using the field mappers
+        """
+        if self.prefix_mapper:
```
giocalitri: please be explicit if here you mean `if self.prefix_mapper is not None:`

exams/pearson.py
```diff
@@ -36,6 +41,13 @@
 log = logging.getLogger(__name__)
 
 
+def _format_date(date):
```
giocalitri: why `_format_date`, `_format_datetime` and `_get_field_mapper` are not part of `TSVWriter`? Are they used outside this module?

exams/pearson.py
```diff
@@ -55,90 +67,147 @@ def _get_field_mapper(field):
...
+
+        Returns:
+            str:
+                the alpha3 country code
+        """
+        # Pearson requires ISO-3166 alpha3 codes, but we store as alpha2
+        try:
+            country = pycountry.countries.get(alpha_2=profile.country)
+        except KeyError as exc:
+            raise InvalidProfileDataException() from exc
```
giocalitri: did not know about this syntax!

exams/pearson.py
```diff
@@ -154,14 +223,17 @@ def upload_tsv(file_path):
...
+            host=str(settings.EXAMS_SFTP_HOST),
+            port=int(settings.EXAMS_SFTP_PORT),
+            username=str(settings.EXAMS_SFTP_USERNAME),
+            password=str(settings.EXAMS_SFTP_PASSWORD),
+            cnopts=cnopts,
+        ) as sftp:
+            with sftp.cd(settings.EXAMS_SFTP_UPLOAD_DIR):
+                sftp.put(file_path)
+    except (ConnectionException, SSHException) as ex:
+        raise RetryableSFTPException() from ex
```
giocalitri: please, try to wrap only the statements for which you want to handle exceptions in a `try/except`, even if you have to write a `try/except` for each statement. 
in case of problems it will be much easier to find the origin of the failure.

exams/pearson_test.py
```diff
@@ -57,31 +80,55 @@ def test_profile_country_to_alpha3_invalid_country(self):  # pylint: disable=no-
             profile = ProfileFactory.create()
         profile.country = 'XXXX'
         with self.assertRaises(InvalidProfileDataException):
-            _profile_country_to_alpha3(profile)
+            CDDWriter.profile_country_to_alpha3(profile)
+
 
-    def test_write_profiles_cdd_no_profiles(self):  # pylint: disable=no-self-use
+class ExamPearsonProfileTest(TestCase):
```
giocalitri: I think you need to reorganize all the tests here around the new classes you wrote and make tests first for the base class and its methods and then test that the subclasses are instantiated in the right way.

exams/tasks.py
```diff
@@ -2,56 +2,128 @@
...
+                ExamProfile.objects.filter(
+                    id__in=exam_profile_ids).update(status=ExamProfile.PROFILE_IN_PROGRESS)
+
+            # update records to reflect invalid profile
+            if invalid_profiles:
+                exam_profile_ids = [exam_profile.id for exam_profile in invalid_profiles]
+                ExamProfile.objects.filter(
+                    id__in=exam_profile_ids).update(status=ExamProfile.PROFILE_INVALID)
+    except:  # pylint: disable=bare-except
+        log.exception('Unexpected exception updating ExamProfile.status')
```
giocalitri: like in another comment, please wrap with `try/except` only the lines you want to handle.

exams/factories.py
rhysyngsun: Done

giocalitri:
please rebase master

exams/pearson.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: Done

exams/tasks.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: Done

exams/pearson.py
rhysyngsun: Done

giocalitri:
functionality looks good, but the content of the file contains some weird newlines:

```
AuthorizationTransactionType    ClientAuthorizationID   ClientCandidateID       ExamSeriesCode  Modules Accommodations  EligibilityApptDateFirst        EligibilityApptDateLast LastUpdate^M
add     5       133     CgiZ_MicroMasters       88.69x          2017/10/20      2017/10/30      2017/01/05 17:08:57^M
update  8       136     osog_MicroMasters       18.63x          2017/09/18      2017/09/28      2017/01/05 17:08:57^M
delete  9       137     RNAN_MicroMasters       29.05x          2017/11/02      2017/11/12      2017/01/05 17:08:57^M
delete  15      143     WMug_MicroMasters       08.14x          2017/02/23      2017/03/05      2017/01/05 17:08:57^M
update  16      144     zNSi_MicroMasters       65.26x          2017/07/19      2017/07/29      2017/01/05 17:08:57^M
update  19      147     OAuv_MicroMasters       54.13x          2017/08/15      2017/08/25      2017/01/05 17:08:57^M
add     22      150     qLfD_MicroMasters       18.68x          2017/12/15      2017/12/25      2017/01/05 17:08:57^M
update  26      154     pTOq_MicroMasters       72.16x          2017/05/27      2017/06/06      2017/01/05 17:08:57^M
delete  27      155     PjTr_MicroMasters       68.35x          2017/08/03      2017/08/13      2017/01/05 17:08:57^M
delete  31      159     ICgD_MicroMasters       97.50x          2017/06/07      2017/06/17      2017/01/05 17:08:57^M
add     41      169     tULK_MicroMasters       43.81x          2017/09/03      2017/09/13      2017/01/05 17:08:57^M
delete  43      171     QjCG_MicroMasters       01.96x          2017/10/28      2017/11/07      2017/01/05 17:08:57^M
update  51      179     njSM_MicroMasters       24.43x          2017/04/27      2017/05/07      2017/01/05 17:08:57^M
delete  53      181     rjxH_MicroMasters       23.14x          2017/12/12      2017/12/22      2017/01/05 17:08:57^M
```

rhysyngsun:
`\r\n` newlines are expected and required. They're the default for `csv` module dialects so I didn't bother to specify it: https://docs.python.org/3/library/csv.html#csv.Dialect.lineterminator

exams/pearson_test.py
rhysyngsun: Done

exams/pearson.py
```diff
@@ -36,109 +41,173 @@
...
+        a callable that when passed a row returns a computed field value
+
+        Arguments:
+            fields (List): list of (str, str|callable) tuples
+            field_prefix (str): path prefix to prefix field lookups with
+        """
+        self.fields = OrderedDict(fields)
+        self.columns = self.fields.keys()
+        self.field_mappers = {column: self.get_field_mapper(field) for (column, field) in fields}
+        self.prefix_mapper = attrgetter(field_prefix) if field_prefix is not None else None
```
giocalitri: nit: please move the `__init__` at the beginning of the class

exams/pearson.py
```diff
@@ -36,109 +41,173 @@
...
+            ('Address2', 'address2'),
+            ('Address3', 'address3'),
+            ('City', 'city'),
+            ('State', 'state_or_territory'),
+            ('PostalCode', 'postal_code'),
+            ('Country', self.profile_country_to_alpha3),
+            ('Phone', 'phone_number'),
+            ('PhoneCountryCode', 'phone_country_code'),
+            ('LastUpdate', lambda profile: self.format_datetime(profile.updated_on)),
+        ], field_prefix='profile')
```
giocalitri: nit: `__init__` at the beginning of the class

exams/tasks.py
```diff
@@ -2,56 +2,130 @@
...
-             .filter(id__in=exam_profile_ids)
-             .update(status=ExamProfile.PROFILE_INVALID))
+        try:
+            # update records to reflect the successful upload
+            if valid_profile_ids:
+                ExamProfile.objects.filter(
+                    id__in=valid_profile_ids).update(status=ExamProfile.PROFILE_IN_PROGRESS)
+
+            # update records to reflect invalid profile
+            if invalid_profile_ids:
```
giocalitri: could `valid_profile_ids` and `invalid_profile_ids` be `True` at the same time?

giocalitri:
Just 2 final comments. Feel free to merge after addressing them and after tests pass again. 👍

exams/tasks.py
rhysyngsun: Yes they could - I'm not quite sure what you're alluding to but now that I'm reading my implementation here again, this `try..except` inside the atomic transaction is not a good thing - it hides from Django that the transaction broke. I'd propose wrapping the entire transaction block instead. If you feel we need per-query logging on which one errored that we don't otherwise get in the traceback I can also wrap/log/reraise each query.

giocalitri: ok, go for it

rhysyngsun: I moved the transaction inside the try..except and added a comment explaining this section.

PR #2225, opened 2016-12-29
Use image_medium for profile images

### LINKED ISSUE #2214: Use medium size profile images, and change user card on learners page to lazy load images
Blocked by #2213

noisecapella:
Fixes #2214 

#### What's this PR do?
Uses the medium size profile image URL in all profile images

View your profile image and confirm that it's 128x128


rhysyngsun:
It looks like the profile image in the header navigation is using the medium image instead of the small, is that intentional?

rhysyngsun:
:+1:

PR #2228, opened 2016-12-29
Show spinner only on currently active button

### LINKED ISSUE #2140: Spinners appearing in two places on some buttons
Specifically:
 - if one pay now button is clicked, all pay now buttons have the spinner while it is loading
 - if an education or employment entry is deleted, the spinner appears on the save button of the profile as well

noisecapella:
Fixes #2140

#### What's this PR do?
Stores state on whether the button was recently clicked. Only buttons which are recently clicked show the spinner, otherwise the button is disabled.

Set your network throttling to simulate a slow connection, and configure your dashboard to show more than one course which has the enroll link. Mocking the functions in api.js might be useful here. Then click one of the enroll buttons. Only that button should have the spinner, and the other enroll buttons should be disabled. When the API call completes, the buttons should all be enabled again.

screenshot from 2016-12-29 16-36-07

screenshot from 2016-12-29 16-35-46
The dot on top here is actually part of the spinner

### DIFF
--- a/static/js/components/SpinnerButton.js
+++ b/static/js/components/SpinnerButton.js
@@ -34,15 +34,14 @@ export default class SpinnerButton extends React.Component {
 
   isDisabled = () => (this.props.disabled || this.props.spinning) ? true : undefined;
 
-  // wrap onClick to return undefined if disabled and to set recentlyClicked if clicked
-  onClickWrapper = () => (
-    this.isDisabled() ? undefined : (...args) => {
-      this.setState({
-        recentlyClicked: true
-      });
-      return this.props.onClick(...args);
-    }
-  );
+  // If button is not disabled and has an onClick handler, make sure to set recentlyClicked
+  // so we display the spinner
+  onClick = (...args) => {
+    this.setState({
+      recentlyClicked: true
+    });
+    return this.props.onClick(...args);
+  };
 
   render() {
     let {
@@ -67,7 +66,7 @@ export default class SpinnerButton extends React.Component {
       className={className}
       disabled={this.isDisabled()}
       {...otherProps}
-      onClick={this.onClickWrapper()}
+      onClick={this.isDisabled() ? undefined : this.onClick}
     >
       {children}
     </ComponentVariable>;
--- a/static/js/components/SpinnerButton.js
+++ b/static/js/components/SpinnerButton.js
@@ -1,15 +1,36 @@
 import React from 'react';
 import Spinner from 'react-mdl/lib/Spinner';
 
+type SpinnerButtonProps = {
+  spinning: bool,
+  component: React.Component<*, *, *>,
+  className?: string,
+  onClick?: Function,
+  children?: any,
+  disabled?: ?bool,
+};
+
 export default class SpinnerButton extends React.Component {
-  props: {
-    spinning: bool,
-    component: React.Component<*, *, *>,
-    className?: string,
-    onClick?: Function,
-    children?: any,
-    disabled?: ?bool,
-  };
+  props: SpinnerButtonProps;
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      // This keeps track of if a button was recently clicked, to indicate
+      // that props.spinning is relevant to this button. If a button was not
+      // clicked but spinning=true, it will be ignored.
+      recentlyClicked: false
+    };
+  }
+
+  componentWillReceiveProps(nextProps: SpinnerButtonProps) {
+    if (!nextProps.spinning && this.props.spinning) {
+      // spinning has finished, so reset the state
+      this.setState({
+        recentlyClicked: false
+      });
+    }
+  }
 
   render() {
     let {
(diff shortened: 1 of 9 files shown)


aliceriot:
functionality looks great! looking at the code now.

static/js/components/SpinnerButton.js
```diff
@@ -21,19 +42,32 @@ export default class SpinnerButton extends React.Component {
...
 
+    if (onClick) {
+      let oldOnClick = onClick;
+      onClick = (...args) => {
+        this.setState({
+          recentlyClicked: true
+        });
+        return oldOnClick(...args);
+      };
+    }
```
aliceriot: I think it would be nice to pull this out into a separate method (`this.onClick` or something)

static/js/components/SpinnerButton_test.js
```diff
@@ -71,11 +78,64 @@ describe("SpinnerButton", () => {
...
+      />
+    );
+    assert.isFalse(wrapper.state().recentlyClicked);
     let buttonProps = wrapper.find("button").props();
+    buttonProps.onClick("args");
+    assert.isTrue(onClick.calledWith("args"));
+    assert.isTrue(wrapper.state().recentlyClicked);
+  });
 
+  it('is disabled if spinning is true but recentlyClicked is false', () => {
```
aliceriot: maybe it would be better to title this test `'does not show the spinner if spinning is true but recentlyClicked is false'`.

static/js/components/UserPageAboutMeDialog_test.js
```diff
@@ -88,13 +89,12 @@ describe('UserPageAboutMeDialog', () => {
...
     renderDialog({
       profilePatchStatus: FETCH_PROCESSING
     });
-    let saveButton = getDialog().querySelector(".save-button");
-    assert(saveButton.className.includes("disabled-with-spinner"));
-    assert(saveButton.querySelector(".mdl-spinner"));
-    TestUtils.Simulate.click(saveButton);
-    assert.equal(saveProfile.callCount, 0);
+    // assert that inFlight is true
+    console.log("calls", dialogActionsSpy.calls);
```
aliceriot: stray `console.log`

static/js/containers/FinancialAidCalculator_test.js
```diff
@@ -113,20 +114,18 @@ describe('FinancialAidCalculator', () => {
...
       return renderComponent('/dashboard', DASHBOARD_SUCCESS_ACTIONS).then(() => {
-        let confirmDialog = document.querySelector('.skip-financial-aid-dialog-wrapper');
-        let skipButton = confirmDialog.querySelector('.skip-button');
-
-        assert.equal(skipButton.className.includes('disabled-with-spinner'), activity);
-        assert(skipButton.innerHTML.includes(activity ? 'mdl-spinner' : 'Pay Full Price'));
-        TestUtils.Simulate.click(skipButton);
-        assert.equal(skipFinancialAidStub.calledWith(program.id), !activity);
+        // assert inFlight arg
+        console.log(dialogActionsSpy.callCount);
```
aliceriot: another, haha

noisecapella:
@aliceriot I fixed the tests, looking at the other review comments now...

noisecapella:
Comments addressed!

static/js/components/SpinnerButton.js
```diff
@@ -32,47 +32,42 @@ export default class SpinnerButton extends React.Component {
...
+
+  // wrap onClick to return undefined if disabled and to set recentlyClicked if clicked
+  onClickWrapper = e => (
+    this.isDisabled() ? undefined : (...args) => {
+      this.setState({
+        recentlyClicked: true
+      });
+      return this.props.onClick(...args);
+    }
+  );
```
aliceriot: Maybe I'm missing something, but why do we need to call this function at render time? It seems like we could just write a simple `onClick` handler:

```js
onClick = (...args) => {
  const { onClick } = this.props;
  this.setState({ recentlyClicked: true });
  onClick(...args);
};
```

then down in `render` we could do the ternary in the `onClick` prop of the `ComponentVariable`, like:

```js
onClick = { this.isDisabled() ? undefined : this.onClick }
```

that approach feels a little cleaner to me, not sure what your thoughts are.

noisecapella: Sure, I assumed that your comment was moving the whole thing including the if statement, but this works too

noisecapella:
How's this?

static/js/components/SpinnerButton.js
```diff
@@ -1,46 +1,74 @@
...
+  componentWillReceiveProps(nextProps: SpinnerButtonProps) {
+    if (!nextProps.spinning && this.props.spinning) {
+      // spinning has finished, so reset the state
+      this.setState({
+        recentlyClicked: false
+      });
+    }
+  }
+
+  isDisabled = () => (this.props.disabled || this.props.spinning) ? true : undefined;
```
aliceriot: I think this could be written a little bit more simply, as just

```js
this.props.disabled || this.props.spinning || undefined;
```

PR #2239, opened 2017-01-03
Added find_test.sh

noisecapella:
None

#### What's this PR do?
Just a shell script to use wildcards in looking up test names for JS tests

### DIFF
--- a/find_test.sh
+++ b/find_test.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+FILE=""
+if [[ "$1" != "" ]]
+then
+    FILE=$(find -name "$1")
+
+    if [[ "$FILE" == "" ]]
+    then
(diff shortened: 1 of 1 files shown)


aliceriot:
idk if this is helpful, but I have a pair of zsh functions that are kind of handy for doing something along these lines:

```zsh
function setjstest() {
    jstest=$(ag -g test.js | fzf-tmux)
    echo 'running '$jstest
    djstest $jstest
    export lasttest=$jstest
}

function jstest() {
    if [ -z ${lasttest+1} ]; then
        setjstest
    else
        echo 'running '$jstest
        djstest $lasttest
    fi
}
```

you use `setjstest` to pick a test you want to run, and `jstest` to re-run the last test you picked. `jstest` will run the test picker if a test isn't already set.

Issue #1941, opened 2016-12-02
Replace static links in React code with ones from server


noisecapella:
This will let us use links to a CDN instead of our own server

PR #2240, opened 2017-01-03
add phone number to profile

### LINKED ISSUE #2137: Profile: Telephone Number
Part of #1821 and #1796 

This is the only totally new thing we need to collect. Add this to the bottom of the Sign Up page 1 as a two part field: 1). Country Code dropdown pre-selected the user’s country, and 2) The telephone number. TBD whether we want a link “Why we ask this”.

image

### Acceptance criteria

- [x] Must work in both initial profile creation (/profile/personal) and in profile editing (/learner)
- [x] Data should be stored in `phone_number` and `phone_country_code` of `Profile`. The country code can be up to 3 digits.

aliceriot:

closes #2137 

#### What's this PR do?

this adds a telephone number input (and relevant validation) to the profile.

If you open the app and go to `/profile/personal` there should be a telephone input. It should require the input, and it should complain about any input which is not a valid telephone number.

phone_numbah


profiles/serializers.py
```diff
@@ -228,7 +228,11 @@ class Meta:  # pylint: disable=missing-docstring
             'image_medium',
             'about_me',
             'romanized_first_name',
-            'romanized_last_name'
+            'romanized_last_name',
+            'phone_number',
```
rhysyngsun: `'phone_number'` in here twice.

static/js/util/profile_edit.js
```diff
@@ -220,6 +221,57 @@ export function boundCheckbox(keySet: string[], label: string|React$Element<*>):
...
+  updateProfile,
+  validator,
+  newPhoneNumber,
+  countryObj
+) => {
+  let currentPhoneNumber = _.get(profile, keySet);
+  if ( currentPhoneNumber !== newPhoneNumber ) {
+    let clone = _.cloneDeep(profile);
+    _.set(clone, keySet, newPhoneNumber);
+    _.set(clone, countryKeySet, countryObj.iso2);
```
rhysyngsun: Probably a misnomer on my part with the field name that lead you in this direction, but this should actually be the calling code (e.g. "1" for North America). Looks like `dialCode` is what we'd want here.

static/js/util/profile_edit.js
```diff
@@ -220,6 +221,57 @@ export function boundCheckbox(keySet: string[], label: string|React$Element<*>):
...
+    updateProfileValidation(profile, validator);
+    sendFormFieldEvent(keySet);
+  };
+
+  let currentCountry = R.toLower(R.defaultTo("", _.get(profile,  ['country'])));
+  return (
+    <div className={`bound-telephone ${validationErrorSelector(errors, keySet)}`}>
+      <ReactTelInput
+        defaultCountry={currentCountry}
+        flagsImagePath='/static/images/flags.png'
```
rhysyngsun: How does this factor into the CDN work we've done? Would be nice to make use of that, although I don't know if we have a methodology for using it in JS yet.

static/js/util/profile_edit.js
```diff
@@ -220,6 +221,57 @@ export function boundCheckbox(keySet: string[], label: string|React$Element<*>):
...
+    updateValidationVisibility,
+  } = this.props;
+
+  let onBlur = () => {
+    updateValidationVisibility(keySet);
+    updateProfileValidation(profile, validator);
+    sendFormFieldEvent(keySet);
+  };
+
+  let currentCountry = R.toLower(R.defaultTo("", _.get(profile,  ['country'])));
```
rhysyngsun: This is probably worth a discussion outside this PR, but code that uses both Ramda and lodash feels weird to me. Long-term we might want to figure out if we try to standardize around usage of one. This could probably be `R.toLower(R.pathOr("", ['country'], profile))` if you want to go pure-Ramda.

aliceriot:
node-sass! always trouble with that one. try blowing away and rebuilding your watch container?

rhysyngsun:
Functionality is good, I like the validation (i.e. the error on an invalid area code).

aliceriot:
I used google's [libphonenumber](https://github.com/googlei18n/libphonenumber), which is pretty nice!

static/js/util/profile_edit.js
aliceriot: I didn't think of that, maybe @noisecapella has an idea of how it would interact?

static/js/util/profile_edit.js
aliceriot: do you think we still need to save that if the country code is already in the phone number field? when the user types a number into the input it will format it automatically with a country code.

profiles/serializers.py
aliceriot: oops lol

static/js/util/profile_edit.js
aliceriot: Ah, I forgot about `pathOr`! I'm all for Ramda but I think there are a few things that we want to keep lodash around for.

static/js/util/profile_edit.js
noisecapella: It wouldn't, we still need to do the work of telling the JS side what our CDN is and attaching it as a prefix #1941

aliceriot: OK cool, I'll leave that off for now then.

static/js/util/profile_edit.js
aliceriot: @rhysyngsun any thoughts on this?

rhysyngsun: @aliceriot sorry for some reason I did not see this comment. These data fields in the profile are for sending to Pearson, so that's why they're structured that way (see https://github.com/mitodl/micromasters/blob/master/exams/pearson_test.py#L79). For example, if your full US phone number is `1-978-123-4567`, then the code we already have in place to sync this data to Pearson will expect `phone_number = '978-123-4567'` and `phone_country_code = '1'`. Otherwise, if you think sorting them in an alternative way is better, we can discuss that, but the corresponding Pearson profile sync code will need to be adjusted accordingly.

aliceriot: Ah OK, I'll see about making the frontend compatible with that.

profiles/serializers.py
```diff
@@ -228,7 +228,10 @@ class Meta:  # pylint: disable=missing-docstring
             'image_medium',
             'about_me',
             'romanized_first_name',
-            'romanized_last_name'
+            'romanized_last_name',
+            'about_me',
```
singingwolfboy: You have `about_me` listed twice. Is that intentional?

static/js/util/profile_edit.js
aliceriot: OK so I think I can see a way to make this work, but it's going to be a messy hack. the input I'm using is not meant to be used this way at all - so I'm going to have to chop out the country code in the `onChange` handler, record it separately, and then pass the value of pasting them together in as the value to the input. Which do you think would be easier to change, how it's done on the python side of things or here?

profiles/serializers.py
aliceriot: haha, no it is not! thanks for catching that.

static/js/util/profile_edit.js
aliceriot: If the above wasn't clear, the reason is that if I type in a US phone number the input component will format the number as `+1 (207) 702-0020` automatically. There doesn't seem to be any granularity in the component options (like I can't turn off adding the country code) other than completely disabling all the formatting.

rhysyngsun: Since the output formatting varies across countries I don't feel putting the burden of parsing that on the backend is the right choice since it's an implementation detail of the frontend, but I also don't want to make a mess working around this.

I think an easier path would be to parse the two parts out via regex? Reading the sources for the input I think `/\+([0-9]+)(.+)/` would probably do the trick (http://regexr.com/3f1a3) and is much less hacky than what you're describing. It should be easy to regenerate that string to initialize the input as well.

aliceriot: Haha, sorry, that's exactly what I'm describing.

aliceriot: I think it's yucky and inelegant but if it's more work on the backend we can do it here. Is there a reason, though, that we couldn't in theory just record the whole phone number, area code and all, and ship that to pearson?

aliceriot: Just wondering if the requirement is our data model or theirs.

rhysyngsun: I suppose we can, but up until now the strategy has been to reflect their data model in ours for the fields we didn't have yet.

If we want to go this route the current `phone_number` column on profile needs to get widened by enough chars that it can fit a 3-digit country code + whatever extra formatting might be present (probably the leading `+` and whatever delimeter falls between the country code and the rest of the number).

If you go this route, could you open a new ticket to fix the profile export to Pearson to address this change in data structuring?

aliceriot: OK, cool, thanks for the info. I think I'm going to see about getting the frontend strategy to work before trying anything like that.

rhysyngsun: One slight amendment to that previous comment: I suppose the column with change is only really required if there is an output of the tool you're using that exceeds 35 chars.

aliceriot: based on https://en.wikipedia.org/wiki/Telephone_numbering_plan it seems like phone numbers can't exceed 30 digits, but there is a little bit of formatting that the input includes (although we could easily use a regex to take out whitespace, parentheses, dashes, etc).

aliceriot:
I updated things to handle separating the country code from the rest of the number on the frontend

aliceriot:
@rhysyngsun I think this should be sorted, I removed the code to split up the number and I added a migration to widen the phone number field.

aliceriot:
@rhysyngsun I don't seem to be able to reproduce that error, is it still happening for you?

aliceriot:
@rhysyngsun that issue should be fixed, working on fixing the python tests now.

aliceriot:
I think I fixed it? lol

pdpinch:
Does this change apply to `/learner/{username}` as well as `/profile/personal`, so users can update their phone number?

If it's a lot of work, I'm ok with opening another issue. I'd really like to see this deployed this week.

aliceriot:
We use the same component in both places

aliceriot:
(for the whole form, I mean)

exams/pearson_test.py
```diff
@@ -224,6 +224,25 @@ def test_profile_country_to_alpha3_invalid_country(self):  # pylint: disable=no-
         with self.assertRaises(InvalidProfileDataException):
             CDDWriter.profile_country_to_alpha3(profile)
 
+    def test_profile_phone_number_functions(self):
+        """
+        A profile with an invalid phone number should raise an InvalidProfileDataException
+        """
+        for case in ['', None, 'bad string', '120272727']:
```
rhysyngsun: This is probably a good use case for `ddt`.

static/js/components/PersonalTab_test.js
```diff
@@ -24,6 +24,7 @@ describe("PersonalTab", () => {
             ui={{selectedProgram: selectedProgram}}
             dispatch={store.dispatch}
             uneditedProfile={USER_PROFILE_RESPONSE}
+            profile={USER_PROFILE_RESPONSE}
```
rhysyngsun: Just curious when this needed to be added for this PR?

exams/pearson_test.py
aliceriot: What is ddt?

rhysyngsun: Data Driven Tests, it's a python module we use to parameterize tests on different data sets. It also tries to give each run of the test a meaningful name based on the data:

Docs: http://ddt.readthedocs.io/en/latest/example.html

Example usage in one of our tests: https://github.com/mitodl/micromasters/blob/master/exams/pearson_test.py#L367

In this case, you'd split the false cases into a `ddt`-based test and keep the positive test case at the end of the method in a regular test method.

static/js/components/PersonalTab_test.js
aliceriot: Sorry this took forever to answer! Ran into a bunch of issues with `yarn` that I couldn't get past (in order to get the tests running).

Basically, the phone number input adds a `'+'` to the input if it is empty (on page-load), which fires off it's `onChange` handler (and then calls the `updateProfile` function). Before we were not passing a `profile` to `PersonalTab`, which was fine, because we aren't testing here any functionality relating to the profile editing, but now this behavior of the empty phone number input meant that, in the personal tab test, that `onChange` handler was firing, and throwing an error because it couldn't find the `updateProfile` prop. Passing a `profile` prop ensures that the phone number input has a value, so it doesn't try to do this thing.

aliceriot: Does that make sense? haha

aliceriot:
questions addressed, switched to using `ddt`.

static/js/components/PersonalTab_test.js
rhysyngsun: Yup, makes sense.

rhysyngsun:
I merged that fix for the loop last night so that should be covered, but I don't think you rebased since so I can't test that.

Otherwise :+1:

aliceriot:
updated to use one migration instead of two

rhysyngsun:
:+1: assuming tests pass

exams/pearson.py
```diff
@@ -142,6 +143,14 @@ def write(self, tsv_file, rows):
         return (valid_rows, invalid_rows)
 
 
+def split_phone_number(num):
+    """
+    split a phone number into the calling code and the rest of the number
+    """
+    reg = re.compile(r'^\+(\d*)\s?(.*)')
```
singingwolfboy: Several comments here:

* If you're going to compile the regular expression, you're better off doing so outside of the function declaration, so that it can be reused between function calls
* This regular expression could probably benefit from adding a few comments using the [`re.VERBOSE`](https://docs.python.org/2/library/re.html#re.VERBOSE) flag
* The fact that you're using `findall` and then returning the first item in this list makes it difficult to understand what this function is really doing. At first, I thought that this function was misnamed, and was only extracting the calling code from the phone number. Might want to use named captures to make this more clear.
* You're using `*` in several places where you probably want `+` instead. `*` matches 0 or more times, `+` matches 1 or more times. Only `*` if the thing you're matching is optional.

I'd suggest something more like this:

```python
import re

PHONE_NUMBER_RE = re.compile(r"""
    ^    # start of string
    \+   # literal plus sign
    (?P<calling_code>\d+)  # named capture: 1 or more digits
    \s?  # an optional whitespace character
    (?P<number>.+)         # named capture: anything else
""", re.VERBOSE)

def split_phone_number(phone_number):
    """
    Split a phone number into the calling code and the rest of the number
    """
    match = PHONE_NUMBER_RE.search(phone_number)
    if not match:
        msg = "{phone_number} is not a valid phone number".format(
            phone_number=phone_number
        )
        raise ValueError(msg)
    return match.group("calling_code"), match.group("number")
```

static/js/components/PersonalForm.js
```diff
@@ -152,6 +152,10 @@ export default class PersonalForm extends ProfileFormFields {
             </Cell>
           </Grid>
         </section>
+        <section>
+          <h3>Telephone</h3>
+          { this.boundTelephoneInput(['phone_number']) }
```
singingwolfboy: This telephone input isn't accessible. The `h3` tag above should be marked as a label for this input.

static/js/util/profile_edit.js
```diff
@@ -1,11 +1,12 @@
 // @flow
 import React from 'react';
 import _ from 'lodash';
-import R from 'ramda';
 import TextField from 'material-ui/TextField';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 import Checkbox from 'material-ui/Checkbox';
 import Geosuggest from 'react-geosuggest';
+import R from 'ramda';
```
singingwolfboy: Out of curiosity, why are you moving this input further down the list? Seems like Lodash and Ramda should be grouped together.

aliceriot:
@singingwolfboy given I already got a 👍 from Nathan, do you want me to address those comments?

singingwolfboy:
@aliceriot I'm fine with pushing it off until later, if this is time-sensitive. However, the accessibility issue is an important one, and unlikely to be addressed properly if we leave it for later. I was just browsing through this PR and saw these issues, so I commented on them.

aliceriot:
I think we want to get this merged asap (@rhysyngsun has some work rebased off of it) so I'm going to go ahead and merge. I'll file an issue for improving the accessibility, and hopefully be able to circle back on it in the next day or so.

singingwolfboy:
Alright, that sounds reasonable to me.

PR #2246, opened 2017-01-04
Split up URLs into respective apps

### LINKED ISSUE #2245: Split up urls.py so that URLs for views are defined in their modules

noisecapella:
Fixes #2245 

#### What's this PR do?
Breaks up `micromasters/urls.py` and creates `urls.py` for each app which defines views

I made the tests and the refactoring separate commits. You can verify that there is no change to the URLs by checking out the test commit, running the tests, then checking out the refactor commit and running the tests again.

### DIFF
--- a/courses/urls.py
+++ b/courses/urls.py
@@ -1,0 +1,17 @@
+"""URLs for courses and programs"""
+
+from django.conf.urls import include, url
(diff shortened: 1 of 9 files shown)


rhysyngsun:
:+1:

PR #2247, opened 2017-01-04
update to webpack 2, add route-based JS splitting

aliceriot:

part of #2163 

#### What's this PR do?

This upgrades webpack up to version 2, and implements route-based splitting for the JavaScript bundles. This means that every route has a separate bundle which contains the dependencies which are unique to that route. In particular, this will mean that most users will never have to download searchkit, which is humongous! hooray!

there's some info on the issue about the actual filesizes that this achives.

Make sure that everything in the app still works.

Implementing this feature requires the use of webpack-specific functionality, in particular webpack's implementation of the `import()` TC39 proposal: https://github.com/tc39/proposal-dynamic-import

This functionality is not implemented by nodejs (it is transpiled by babel and webpack into `require` statements which webpack resolves). So, this code cannot directly be run by nodejs.

There are two options here. I have isolated this code in it's own file, so it's possible to just not touch it from the node environment we use for running tests. This is what I have implemented - we declare the routes for the application twice, once to implement the lazy loading functionality, which we build with webpack, and once for testing, which runs on nodejs. 

The other possibility would be that we start running the tests through webpack - I don't know how feasible this is technically. Anyway, if there's a better was to do this I would love to know! haha.


noisecapella:
What's the difference between `import(...)` and `require(...)`?

noisecapella:
I'm getting this when running `yarn install`:

    Incorrect peer dependency "webpack@^1.12.6 || ^2.1.0-beta".

Not sure if this is a problem

noisecapella:
The functionality works fine, but there's now a delay when switching from one page to the next. If I turn on 2G throttling, go to `/profile/professional`, then click next, there's some amount of time where the button has a spinner. Then the spinner stops and returns to a button while the rest of the page loads. When the JS file finishes loading I'm immediately moved to the next page even if I had moved elsewhere in the meantime

aliceriot:
hmmm not sure what to do about that, I'll poke around a little bit. One thing might be implementing some sort of background thing that starts pulling down the rest of the JavaScript that might be required once the page finishes loading.

aliceriot:
`import()` is likely going to be part of the ECMAScript standard (it's a stage-3 TC39 proposal right now, you can read the details [here](https://github.com/tc39/proposal-dynamic-import)). `require` is, for us in this case, something that will be handled by webpack. We could implement the functionality I've implemented using `import()` with `require.ensure`, which lets you do a similar form of asynchronous module loading.

aliceriot:
do you think the page load issue is something that would be handled by the CDN / gzip compression? wondering if I should start looking into ways around that.

noisecapella:
Is it possible to just split apart the `/learners` page from all the other pages? I feel like there isn't much extra which is not shared between the different pages, and nobody with a bad internet connection goes between `/learners` and etc anyway

I'm fine keeping it the way it is FYI, just wanted to bring it up for discussion so that we figure out what the tradeoffs are

webpack.config.shared.js
```diff
@@ -13,35 +13,44 @@ module.exports = {
...
+      extensions: ['.js', '.jsx'],
       alias: {
         react: path.resolve('./node_modules/react')
       }
     }
   },
-
   babelSharedLoader: {
-    text: /\.jsx?$/,
+    test: /\.jsx?$/,
```
noisecapella: Haha I wonder why this didn't affect us before

aliceriot: no idea haha

aliceriot:
Yes, we should be able to do that - I can test out just splitting out the learners page and see what that gets us. One other possibility is adding an `onRouteEnter` callback which would start pulling down the JavaScript for the rest of the (user-facing) site on `/dashboard`.

static/js/dashboard_routes.js
```diff
@@ -0,0 +1,99 @@
...
+  throw new Error(`Dynamic page loading failed: ${error}`);
+};
+
+const loadRoute = cb => (
+  module => cb(null, module.default)
+);
+
+const username = SETTINGS.user ? SETTINGS.user.username : '';
+
+export const routes = {
```
noisecapella: Is it possible to have a function with these states returned as react components, then translate it via map to this dictionary? Or the other way around for the test routes?

aliceriot: Unfortunately no, the calls to `import` have to be statically known to webpack before any of the code runs in order to figure out where the module boundaries should be (and to then do the splitting). We can't use this to derive the test routes from because the test routes do not run in a webpack environment, so when `nodejs` tries to parse the files it will complain about non-top level import statements. It's messy I know :(

aliceriot: One solution that would allow us to potentially unify them is to use webpack to run the tests - but I'm not sure how complicated and messy that would be (my thought is probably very messy)

noisecapella: Yeah, I'm fine to leave it as is

static/js/dashboard_routes.js
```diff
@@ -0,0 +1,99 @@
...
+  path: "/",
+  component: App,
+  childRoutes: [
+    {
+      path: 'dashboard',
+      getComponent(location, cb) {
+        import('./containers/DashboardPage')
+          .then(loadRoute(cb, false))
```
noisecapella: It looks like `loadRoute` only takes one argument, can we remove the `, false` here and below?

static/js/dashboard_routes.js
```diff
@@ -0,0 +1,99 @@
...
+
+const username = SETTINGS.user ? SETTINGS.user.username : '';
+
+export const routes = {
+  path: "/",
+  component: App,
+  childRoutes: [
+    {
+      path: 'dashboard',
+      getComponent(location, cb) {
```
noisecapella: Can you rename `location` to `nextState` to match name in the documentation [here](https://github.com/ReactTraining/react-router/blob/master/docs/API.md#getcomponentnextstate-callback)?

noisecapella:
I'm getting an error when signing onto the PR build:
screenshot from 2017-01-05 13-34-21

static/js/dashboard_routes.js
aliceriot: Oh oops, I was basing this on a tutorial that had that.

aliceriot:
huh it's making http (i.e. not https) requests for some reason

aliceriot:
It looks like it's making a request to `http://micromasters-ci-pr-2247.herokuapp.com:8078/0-bd8514f6fecf9932b36a.js`

is there something I need to update in the django side of things for this maybe?

noisecapella:
Hmm... our Django code just differentiates between the `http://server:8078/name.js` links and `http://server.herokuapp.com/static/js/bundle/name.js` links. Since that link is being generated by webpack it's probably not aware that the `__webpack_public_path__` being used is only meant for dev hot reloading purposes.

noisecapella:
I guess we can make a `SETTINGS.public_path` in Django which depends on whether it's dev or production and then use it to specify `__webpack_public_path__`

aliceriot:
I think we can set it in the webpack config too (in `output`), not sure if that will override the `__webpack_public_path__` declaration

noisecapella:
The problem is we still need to get the hostname part from Django since it's determined at runtime

aliceriot:
ohh hm

from the docs it looks like I can set a path like `/static/bundle`

aliceriot:
hm well it seems like `__webpack_public_path__` overrides what's in the config file

aliceriot:
what if we just set `__webpack_public_path__` to be `SETTINGS.host`? then we could use that to do the prod or development setting.

aliceriot:
@noisecapella I think that I fixed the issue with `__webpack_public_path__`, things seem to be loading correctly on the PR build. Do I have to do anything else for compatibility with the CDN changes?

noisecapella:
No, any work to get the front end to work with the CDN is in a future PR for #1941  so this should be fine

aliceriot:
cool :)

ui/templatetags/render_bundle.py
```diff
@@ -12,20 +12,23 @@
...
     for chunk in get_loader('DEFAULT').get_bundle(bundle_name):
         chunk_copy = dict(chunk)
-        if settings.DEBUG and settings.USE_WEBPACK_DEV_SERVER:
-            chunk_copy['url'] = "{host_url}/{bundle}".format(
-                host_url=webpack_dev_server_url(request),
-                bundle=chunk['name'],
-            )
-        else:
-            chunk_copy['url'] = static("bundles/{bundle}".format(bundle=chunk['name']))
+        chunk_copy['url'] = "{host_url}/{bundle}".format(
```
noisecapella: Can you use `urljoin` here, from `urllib.parse`?

aliceriot: sure, should we have already been using that?

noisecapella: Yeah, it just takes care of the case where `host_url` contains a trailing slash already

aliceriot: gotcha

static/js/entry/public.js
```diff
@@ -1,5 +1,5 @@
 /* global SETTINGS: false $:false jQuery: false CURRENT_PAGE_URL: false */
-__webpack_public_path__ = `http://${SETTINGS.host}:8078/`;  // eslint-disable-line no-undef, camelcase
+__webpack_public_path__ = `${SETTINGS.public_path}/`;  // eslint-disable-line no-undef, camelcase
```
noisecapella: Can we add the trailing slash in the Python side and remove it here and in the other places we define `__webpack_public_path__`?

aliceriot: yeah that's a good idea

webpack.config.shared.js
noisecapella: Yeah I guess from the documentation that it basically whitelists the files it processes, but we don't really have files that aren't of this file type right now anyway

webpack.config.shared.js
```diff
@@ -13,35 +13,47 @@ module.exports = {
...
+        path.join(__dirname, "static/js"),
+        "node_modules"
+      ],
+      extensions: ['.js', '.jsx'],
       alias: {
         react: path.resolve('./node_modules/react')
       }
+    },
+    performance: {
+      hints: false
```
noisecapella: What does this do?

aliceriot: It disables some very loud warnings about bundle size that webpack added recently

aliceriot: basically it prints a warning any time there is a bundle which is over 250kb (pretty common in development) to both the command line and to the JavaScript console in the browser

(I got too annoyed by it haha)

noisecapella:
Did you get any results from splitting out the `/learners` page?

aliceriot:
oh! I forgot to do that. I'm going to iron out all these public path issues and then I'll give it a try.

aliceriot:
OK I believe all the public path stuff is sorted. Looking into just splitting out the search page

aliceriot:
this is what the filesizes look like right now:

```
total 3.5M
drwxr-xr-x 2 alice users  12K Jan  6 13:32 ./
drwxr-xr-x 6 alice users 4.0K May 24  2016 ../
-rw-r--r-- 1 alice  1000 221K Jan  6 13:32 0-aab124863da28fcbca8d.js
-rw-r--r-- 1 alice  1000  16K Jan  6 13:32 1-b3245ff90346fe6b425e.js
-rw-r--r-- 1 alice  1000 335K Jan  6 13:32 2-8464e318da043ca0795e.js
-rw-r--r-- 1 alice  1000 136K Jan  6 13:32 3-11ee87501db22eaa55c0.js
-rw-r--r-- 1 alice  1000  22K Jan  6 13:32 4-9131dcdd6b590e9736c2.js
-rw-r--r-- 1 alice  1000 1.4M Jan  6 13:32 common-11956c8af79b67ca0ca4.js
-rw-r--r-- 1 alice  1000 505K Jan  6 13:32 dashboard-aa690f004260d45a3801.js
-rw-r--r-- 1 alice  1000  23K Jan  6 13:32 financial_aid-73377da4f8a4786bd005.js
-rw-r--r-- 1 alice  1000  59K Jan  6 13:32 public-9bcae94bff5ad8e9b88e.js
-rw-r--r-- 1 alice  1000  25K Jan  6 13:32 sentry_client-ddf3d503787f3a581b79.js
-rw-r--r-- 1 alice  1000 671K Jan  6 13:32 style-ea55ec165567ca9d5fe0.js
-rw-r--r-- 1 alice  1000 147K Jan  6 13:32 style_public-6735286d87c6b4a656d7.js
-rw-r--r-- 1 alice  1000 4.1K Jan  6 13:32 zendesk_widget-015539f191d1041e2254.js
```

here's what they look like when I add `import` statements so that all the containers (except for `LearnerSearchPage`) get pulled into the dashboard module:

```
total 3.5M
drwxr-xr-x 2 alice users  12K Jan  6 13:40 ./
drwxr-xr-x 6 alice users 4.0K May 24  2016 ../
-rw-r--r-- 1 alice  1000 334K Jan  6 13:35 0-9d93ec531ad429529d44.js
-rw-r--r-- 1 alice  1000 1.4M Jan  6 13:35 common-72671d242dccd0709d58.js
-rw-r--r-- 1 alice  1000 859K Jan  6 13:35 dashboard-157f3777136d022fccdf.js
-rw-r--r-- 1 alice  1000 2.6K Jan  6 13:35 financial_aid-81d7118bac5c39ca9afa.js
-rw-r--r-- 1 alice  1000  59K Jan  6 13:35 public-bb36a8348fd02726bb29.js
-rw-r--r-- 1 alice  1000  25K Jan  6 13:35 sentry_client-722f317a060c9a9c711e.js
-rw-r--r-- 1 alice  1000 671K Jan  6 13:35 style-432e6202fb17c7785ea4.js
-rw-r--r-- 1 alice  1000 147K Jan  6 13:35 style_public-e793e842c847ad1975e5.js
-rw-r--r-- 1 alice  1000 4.1K Jan  6 13:35 zendesk_widget-72ef84efa421c09cd29f.js
```

so dashboard goes from about 500kb to around 850kb. what do we think? should we go with this approach?

noisecapella:
Is it possible to get a further breakdown of what makes up each bundle?

aliceriot:
like which modules and such? I can post the stats file, gimme a sec

noisecapella:
Mostly I'm just curious what `0` and `3` in the first listing contain, which large components aren't being required in the other pages

aliceriot:
here's a stats file with all the splits: https://gist.github.com/aliceriot/bfbbe11cb0166b33db3e651c45dc3f9b

and here's another with just the learner search page pulled out: https://gist.githubusercontent.com/aliceriot/1b76c9ba8a59c9c69d534950712b936c/raw/09f47969943e5da72272f6a86577f387a6674016/webpack-stats-just-search.json

the analyzer here is pretty handy: https://webpack.github.io/analyse

noisecapella:
@aliceriot It looks like the learner search split file is truncated. Can you upload it again?

noisecapella:
It looks like the `financial_aid` bundle shares a lot of stuff with one other chunk, which gets put into `common`. It might be worth investigating splitting that off too

aliceriot:
which chunk is it shared with?

noisecapella:
3, I think that's the main dashboard page

noisecapella:
I might be overestimating how much is shared actually. It seems like the large shared stuff is only reported in `common`.

noisecapella:
Ooh, lodash is 497KiB, and from their docs it seems like it's designed to support importing submodules. Maybe that's a target for later

aliceriot:
OK I added import statements to just have the learner search page split. Do you want to take a final look over the code?

noisecapella:
yep

Issue #2065, opened 2016-12-09
Anonymize production data for use in dev and testing environments


pdpinch:
_Part of #2270. This description is being updated to reflect recent discussions._

We want to use data as is it exists in production to test our app as accurately as possible, but we don't want our users' personal information to exist in multiple, possibly-less-secure databases. We should look into methods/tools for creating a postgres data dump in a way that obscures specific columns in specific tables.

The current list of fields we know we want to scrub as of 1/11/2017:

- username
- email
- first, last name
- auth tokens (we may just want to delete these entirely as part of the scrubbing process)

**Acceptance criteria:** 

1. We have a final, exhaustive list of every piece of data we want to scrub.
1. We have some proven method for dumping postgres data in a way that obscures specific columns in specific tables. If this isn't possible, we can label this as wontfix and explore other options.

Here's the original description from @pdpinch:

> Sometimes developers need production data for testing. But we don't want to proliferate any personal data. We could use a script for dumping and scrubbing data. 
> 
> data to be scrubbed:
> 
> - [ ] e-mail addresses
> - [ ] user tokens

pdpinch:
It would also be nice if we could find a way to remove the tokens but also avoid getting lots of errors when celery tries to update the cached edx data.

aliceriot:
Optimally this is part of the default development environment - it's nice not to have to fiddle with data setup when you're trying to test something or repro a bug.

gsidebo:
Some reasons why it's still relevant/important to generate our own test data for manual testing:

1. Size/Speed. A prod data dump is basically guaranteed to be a great deal larger than our own test data. Reindexing on the prod dump that Tobias sent me took 73 seconds last time I ran it, and that's a single example of the possible performance penalties. I do not want that to be part of my dev workflow.
1. Precision. We cannot guarantee that prod data covers edge cases that we know and care about for testing.
1. Local development and PR review complexity. When we want to test something like dashboard course states or grades in learners search, the steps to match whatever condition is needed to develop and test a feature are typically pretty simple. That's because we understand certain things about the user and course data. Doing the same with scrubbed prod data would involve mangling user/course/etc. data to match our required conditions. That process could become complex and time-consuming.
[...]

PR #2249, opened 2017-01-04
Added Facet by Degree

### LINKED ISSUE #859: Facet by Degree

annagav:
Fix #859
#### What's this PR do?
Add facet by degree (education level)
Go to learner's search page as staff for a program.

screen shot 2017-01-04 at 5 08 54 pm

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -34,14 +34,23 @@ import EmailCompositionDialog from './EmailCompositionDialog';
 import type { Option } from '../flow/generalTypes';
 import type { Email } from '../flow/emailTypes';
 import type { AvailableProgram } from '../flow/enrollmentTypes';
+import { EDUCATION_LEVELS } from '../constants';
 
-const makeSearchkitTranslations: () => Object = () => {
+const makeCountryNameTranslations: () => Object = () => {
   let translations = {};
   for (let code of Object.keys(iso3166.data)) {
     translations[code] = iso3166.data[code].name;
     for (let stateCode of Object.keys(iso3166.data[code].sub)) {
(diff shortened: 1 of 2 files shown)


static/js/components/LearnerSearch.js
```diff
@@ -45,6 +47,14 @@ const makeSearchkitTranslations: () => Object = () => {
   return translations;
 };
 
+const makeDegreeTranslations: () => Object = () => {
```
gsidebo: can you change the name of `makeSearchkitTranslations` to `makeCountryNameTranslations`, and the `searchkitTranslations` variable name to `countryNameTranslations`? `makeSearchkitTranslations` was only a relevant name because it was the only translation we were applying to any of our searchkit results. now that we are creating another set of translations, we should get more specific with the naming

pdpinch:
Can we test this with production data?

static/js/components/LearnerSearch.js
annagav: Sure

gsidebo:
@pdpinch since we don't yet have a straightforward path to test individual PRs with production data, it might be easier to test this in a release candidate. would you be satisfied with that?

as a side note, it seems that there is enough demand for testing with prod data that we should commit resources to creating that 'straightforward path' that i mentioned above. scrubbing/dumping prod data (https://github.com/mitodl/micromasters/issues/2065) is only one piece of that

pdpinch:
Based on the screenshot, it looks like we have an "All Filter." That's not useful -- let's remove it.

annagav:
Using Gavin's PatchedMenuFilter removes the `All` filter.

screen shot 2017-01-06 at 3 49 23 pm

PR #2250, opened 2017-01-04
Added coupon APIs

### LINKED ISSUE #2237: Coupons: APIs
Part of #2158 

There will be one coupon-specific API which handles GET and POST methods, on `/api/v0/coupons/`:
 - GET will provide the user a list of coupons associated with the user
 - POST will associate a coupon with a user, if it's redeemable by that user. The POST will take a coupon code.

noisecapella:
Fixes #2237 

#### What's this PR do?
Adds REST APIs for getting information about coupons and to attach a user to a coupon.

Create a coupon using the Django shell using the `CouponFactory`. Go to `/api/v0/coupons/`. You should see an empty list. Add the coupon to your user using `UserCoupon.objects.create(coupon=coupon, user=user)` in the shell. Look at `/api/v0/coupons/` again and you should see one coupon.

In the shell delete the `UserCoupon` by running `UserCoupon.objects.filter(user=user, coupon=coupon).delete()`. Go to `/api/v0/coupons/<code>/users/` and POST to that API with `{"username": "your-username-here"}`. In the shell verify that the `UserCoupon` was created via `UserCoupon.objects.filter(user=user, coupon=coupon).exists()`

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -379,3 +379,40 @@ def is_coupon_redeemable(coupon, user):
         return any((mmtrack.has_verified_cert(run.edx_course_key) for run in course.courserun_set.all()))
 
     return True
+
+
+def pick_coupons(user):
+    """
+    Choose the coupon which would be used in redemptions by the user.
+
+    The heuristic is currently:
+     - choose attached coupons over automatic coupons
+     - choose the coupon which has been most recently attached, or most recently modified
+
+    Args:
+        user (django.contrib.auth.models.User): A user
+
+    Returns:
+        Coupon: The coupon which will be used by the user when redeeming runs in a program
+    """
+    sorted_attached_coupons = list(
+        Coupon.objects.filter(usercoupon__user=user).order_by('-usercoupon__updated_on')
+    )
+    sorted_automatic_coupons = list(
+        Coupon.objects.filter(Coupon.is_automatic_qset()).order_by('-updated_on')
+    )
+
+    # At this point there should only be coupons the user has attached (opted into by clicking a link)
+    # or automatic coupons, which there should only be a few. So the next iterations should not
+    # affect many rows in the DB.
+
+    coupons = []
+    # Only one coupon per program
+    program_ids = set()
+    for coupon in sorted_attached_coupons + sorted_automatic_coupons:
+        program_id = coupon.program.id
+        if program_id not in program_ids and is_coupon_redeemable(coupon, user):
+            coupons.append(coupon)
+            program_ids.add(program_id)
+
+    return coupons
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -5,6 +5,7 @@ from base64 import b64encode
 from datetime import datetime
 import hashlib
 import hmac
+from itertools import chain
 import logging
 from urllib.parse import quote_plus
 import uuid
@@ -395,12 +396,8 @@ def pick_coupons(user):
     Returns:
         Coupon: The coupon which will be used by the user when redeeming runs in a program
     """
-    sorted_attached_coupons = list(
-        Coupon.user_coupon_qset(user).order_by('-usercoupon__updated_on')
-    )
-    sorted_automatic_coupons = list(
-        Coupon.is_automatic_qset().order_by('-updated_on')
-    )
+    sorted_attached_coupons = Coupon.user_coupon_qset(user).order_by('-usercoupon__updated_on')
+    sorted_automatic_coupons = Coupon.is_automatic_qset().order_by('-updated_on')
 
     # At this point there should only be coupons the user has attached (opted into by clicking a link)
     # or automatic coupons, which there should only be a few. So the next iterations should not
@@ -409,7 +406,7 @@ def pick_coupons(user):
     coupons = []
     # Only one coupon per program
     program_ids = set()
-    for coupon in sorted_attached_coupons + sorted_automatic_coupons:
+    for coupon in chain(sorted_attached_coupons, sorted_automatic_coupons):
         program_id = coupon.program.id
         if program_id not in program_ids and is_coupon_redeemable(coupon, user):
             coupons.append(coupon)
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -396,10 +396,10 @@ def pick_coupons(user):
         Coupon: The coupon which will be used by the user when redeeming runs in a program
     """
     sorted_attached_coupons = list(
-        Coupon.objects.filter(usercoupon__user=user).order_by('-usercoupon__updated_on')
+        Coupon.user_coupon_qset(user).order_by('-usercoupon__updated_on')
     )
     sorted_automatic_coupons = list(
-        Coupon.objects.filter(Coupon.is_automatic_qset()).order_by('-updated_on')
+        Coupon.is_automatic_qset().order_by('-updated_on')
     )
 
     # At this point there should only be coupons the user has attached (opted into by clicking a link)
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -314,7 +314,7 @@ class Coupon(TimestampedModel):
                 True if the coupon is not automatic and it has already been redeemed by someone else
         """
         return (
-            not Coupon.objects.filter(id=self.id).filter(self.is_automatic_qset()) and
+            not Coupon.objects.filter(id=self.id).filter(self.is_automatic_qset()).exists() and
             RedeemedCoupon.objects.filter(
                 coupon=self,
                 order__status=Order.FULFILLED,
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -313,10 +313,13 @@ class Coupon(TimestampedModel):
             bool:
                 True if the coupon is not automatic and it has already been redeemed by someone else
         """
-        return not self.is_automatic and RedeemedCoupon.objects.filter(
-            coupon=self,
-            order__status=Order.FULFILLED,
-        ).exclude(order__user=user).exists()
+        return (
+            not Coupon.objects.filter(id=self.id).filter(self.is_automatic_qset()) and
+            RedeemedCoupon.objects.filter(
+                coupon=self,
+                order__status=Order.FULFILLED,
+            ).exclude(order__user=user).exists()
+        )
 
     def clean(self):
         """Validate amount and content_object"""
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -28,6 +28,7 @@ from django.db.models.fields import (
 from django.db.models.fields.related import (
     ForeignKey,
 )
+from django.db.models.query_utils import Q
 import pytz
 
 from courses.models import (
@@ -247,6 +248,20 @@ class Coupon(TimestampedModel):
             raise ImproperlyConfigured("content_object expected to be one of Program, Course, CourseRun")
 
     @property
+    def program(self):
+        """Get the program which all course keys for the coupon belong to"""
+        obj = self.content_object
+        if isinstance(obj, Program):
+            return obj
+        elif isinstance(obj, Course):
+            return obj.program
+        elif isinstance(obj, CourseRun):
+            return obj.course.program
+        else:
+            # Should probably not get here, clean() should take care of validating this
+            raise ImproperlyConfigured("content_object expected to be one of Program, Course, CourseRun")
+
+    @property
     def is_valid(self):
         """Returns true if the coupon is enabled and also within the valid date range"""
         now = datetime.now(tz=pytz.UTC)
@@ -258,12 +273,12 @@ class Coupon(TimestampedModel):
             return False
         return True
 
-    @property
-    def is_automatic(self):
+    @classmethod
+    def is_automatic_qset(cls):
         """
         Returns true if the coupon would be redeemed automatically without input from the user.
         """
-        return self.coupon_type == self.DISCOUNTED_PREVIOUS_COURSE
+        return Q(coupon_type=cls.DISCOUNTED_PREVIOUS_COURSE)
 
     def user_has_redemptions_left(self, user):
         """
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -28,7 +28,6 @@ from django.db.models.fields import (
 from django.db.models.fields.related import (
     ForeignKey,
 )
-from django.db.models.query_utils import Q
 import pytz
 
 from courses.models import (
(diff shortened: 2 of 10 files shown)


ecommerce/models.py
```diff
@@ -223,6 +235,85 @@ class Coupon(Model):
...
+        """
+        Returns true if the coupon is redeemable for the user.
+        """
+        if (
+                not self.is_valid or                      # coupon must be enabled and within valid date range
+                course_key not in self.course_keys or     # coupon must apply to a course run we care about
+                self.num_coupons_available == 0 or        # coupon must have a redemption available
+                not self.user_has_redemptions_left(user)  # coupon must not be used up for that user
+        ):
+            return False
```
rhysyngsun: Reading through the logic in this method, it looks like unless conditions are met for this `False` return, all other code paths return `True` so the logic below is fairly moot (i.e., might as well return here). Is this future-proofing or did I miss something?

noisecapella: No, looks like I missed something. It should return `False` below if `has_paid_on_edx` returns `False`

noisecapella:
@gsidebo You may want to review #2282 first

ecommerce/views.py
```diff
@@ -152,3 +164,54 @@ def post(self, request, *args, **kwargs):
...
+    permission_classes = (IsAuthenticated,)
+
+    def get_queryset(self):
+        """List coupons which a user is allowed to see"""
+        return [
+            coupon for coupon in Coupon.objects.all()
+            if is_coupon_redeemable(coupon, self.request.user) and (
+                coupon.is_automatic or UserCoupon.objects.filter(user=self.request.user, coupon=coupon).exists()
+            )
+        ]
```
giocalitri: this is a very expensive piece of code, especially long term. we should find a way to limit least the initial query.

ecommerce/views.py
```diff
@@ -152,3 +164,54 @@ def post(self, request, *args, **kwargs):
...
+    """
+    permission_classes = (IsLoggedInUser, IsAuthenticated,)
+    authentication_classes = (
+        SessionAuthentication,
+        TokenAuthentication,
+    )
+
+    def post(self, request, *args, **kwargs):
+        """Attach a coupon to a user"""
+        code = self.kwargs['code']
```
giocalitri: you can put `code` as an `arg` (or `kwarg` if you need a default value) in the definition of the function and avoid this code

giocalitri:
I did this:

```
In [1]: from ecommerce.factories import CouponFactory
   ...:
   ...: from courses.models import *
   ...:

In [2]: Program.objects.all()
Out[2]: <QuerySet [<Program: Analog Learning>, <Program: Digital Learning>]>

In [3]: Program.objects.exclude(id__in=(1,2)).delete()
Out[3]: (0, {})

In [4]: Program.objects.all()[1]
Out[4]: <Program: Digital Learning>

In [5]: prog = _

In [6]: CouponFactory.create(content_object=prog)
Out[6]: <Coupon: Coupon percent-discount 0.27 of type standard for Digital Learning>
```

and a request to
```
http://192.168.99.100:8079/api/v0/coupons/
```

raises 

```
web_1     | [2017-01-12 18:20:40] ERROR 226 [django.request] exception.py:124 - [001e7c77548c] - Internal Server Error: /api/v0/coupons/
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 39, in inner
web_1     |     response = get_response(request)
web_1     |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
web_1     |     response = self._get_response(request)
web_1     |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 217, in _get_response
web_1     |     response = self.process_exception_by_middleware(e, request)
web_1     |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 215, in _get_response
web_1     |     response = response.render()
web_1     |   File "/usr/local/lib/python3.5/site-packages/django/template/response.py", line 109, in render
web_1     |     self.content = self.rendered_content
web_1     |   File "/usr/local/lib/python3.5/site-packages/rest_framework/response.py", line 72, in rendered_content
web_1     |     ret = renderer.render(self.data, accepted_media_type, context)
web_1     |   File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py", line 701, in render
web_1     |     context = self.get_context(data, accepted_media_type, renderer_context)
web_1     |   File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py", line 678, in get_context
web_1     |     'filter_form': self.get_filter_form(data, view, request),
web_1     |   File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py", line 609, in get_filter_form
web_1     |     queryset = view.get_queryset()
web_1     |   File "./ecommerce/views.py", line 184, in get_queryset
web_1     |     coupon for coupon in Coupon.objects.all()
web_1     |   File "./ecommerce/views.py", line 185, in <listcomp>
web_1     |     if is_coupon_redeemable(coupon, self.request.user) and (
web_1     |   File "./ecommerce/api.py", line 355, in is_coupon_redeemable
web_1     |     programenrollment__program__course__courserun__edx_course_key__in=coupon.course_keys,
web_1     |   File "./ecommerce/models.py", line 246, in course_keys
web_1     |     raise ImproperlyConfigured("content_object expected to be one of Program, Course, CourseRun")
web_1     | django.core.exceptions.ImproperlyConfigured: content_object expected to be one of Program, Course, CourseRun
```

noisecapella:
I'm not able to reproduce that, how many other `Coupon`s are on your system?

ecommerce/views.py
noisecapella: Sure

ecommerce/permissions.py
```diff
@@ -31,3 +32,18 @@ def has_permission(self, request, view):
...
+    """
+    Confirms that the username in the request body is the same as the logged in user's.
+    """
+
+    def has_permission(self, request, view):
+        """
+        Returns true if the username in the request body matches the logged in user.
+        """
+        try:
+            return request.data['username'] == get_social_username(request.user)
```
noisecapella: Shouldn't it be the edX one, since that's the only username the user would know about? That's what `SETTINGS.user.username` is

ecommerce/views.py
noisecapella: On a related note, this API will only return zero or one coupons. I talked to @pdpinch and we need to make a decision somewhere which coupon to display. I'll add a function in `api.py` to calculate that, and try to do filtering to avoid n+1 queries

noisecapella: Nevermind what I said about zero or one coupons. I have some changes to make and then I'll ping you to look at it again

noisecapella:
Ok, that should address all comments

giocalitri:
You said yesterday that something else was going to change in the logic of the code: what has changed?

noisecapella:
I added functionality to pick the best coupon for each program. Previously it listed all redeemable coupons, but we realized we need to choose one for the user and give that information to the frontend. The new logic should mostly be in the `pick_coupon` commit

ecommerce/api.py
```diff
@@ -379,3 +379,40 @@ def is_coupon_redeemable(coupon, user):
...
+        user (django.contrib.auth.models.User): A user
+
+    Returns:
+        Coupon: The coupon which will be used by the user when redeeming runs in a program
+    """
+    sorted_attached_coupons = list(
+        Coupon.objects.filter(usercoupon__user=user).order_by('-usercoupon__updated_on')
+    )
+    sorted_automatic_coupons = list(
+        Coupon.objects.filter(Coupon.is_automatic_qset()).order_by('-updated_on')
```
giocalitri: I think that `Coupon.is_automatic_qset()` needs to be at least renamed given that it does not return a queryset.

BUT, given that it is used always in a `.filter`, I think it should be refactored to do the actual filter and return a queryset, so here and anywhere else can be just be refined.
This is up to you, but if you choose not to refactor, please rename it.

ecommerce/api.py
```diff
@@ -379,3 +379,40 @@ def is_coupon_redeemable(coupon, user):
...
+    )
```
giocalitri: do you actually need to make a list out of these 2 query sets? I am not sure if it has an actual impact on the performances, but it looks to me that it is unnecessary.

ecommerce/models.py
```diff
@@ -28,6 +28,7 @@
 from django.db.models.fields.related import (
     ForeignKey,
 )
+from django.db.models.query_utils import Q
```
giocalitri: nit: I think that the [suggested way](https://docs.djangoproject.com/en/1.10/topics/db/queries/#complex-lookups-with-q-objects) to make this import is 
`from django.db.models import Q`

ecommerce/models.py
```diff
@@ -247,6 +248,20 @@ def course_keys(self):
             raise ImproperlyConfigured("content_object expected to be one of Program, Course, CourseRun")
 
     @property
+    def program(self):
+        """Get the program which all course keys for the coupon belong to"""
```
giocalitri: Nit: not sure I understand what this docstring means. 
I know what this method does, but I do not think that this is the right way to explain it.

ecommerce/models.py
```diff
@@ -258,12 +273,12 @@ def is_valid(self):
             return False
         return True
 
-    @property
-    def is_automatic(self):
+    @classmethod
+    def is_automatic_qset(cls):
         """
         Returns true if the coupon would be redeemed automatically without input from the user.
```
giocalitri: the docstring needs to be updated

ecommerce/models.py
```diff
@@ -298,10 +313,13 @@ def another_user_already_redeemed(self, user):
...
-        return not self.is_automatic and RedeemedCoupon.objects.filter(
-            coupon=self,
-            order__status=Order.FULFILLED,
-        ).exclude(order__user=user).exists()
+        return (
+            not Coupon.objects.filter(id=self.id).filter(self.is_automatic_qset()).exists() and
+            RedeemedCoupon.objects.filter(
+                coupon=self,
+                order__status=Order.FULFILLED,
+            ).exclude(order__user=user).exists()
```
giocalitri: I am not sure you actually need to make the extra query (`RedeemedCoupon.objects.filter`) given that the first filter is `self`. But It looks neat, so it is up to you.

ecommerce/permissions.py
giocalitri: I think the the username in the request is always the local username and not the edX one. 
Am I missing something here?

ecommerce/views.py
```diff
@@ -152,3 +165,47 @@ def post(self, request, *args, **kwargs):
...
+        SessionAuthentication,
+        TokenAuthentication,
+    )
+    permission_classes = (IsAuthenticated,)
+
+    def get_queryset(self):
+        """List coupons which a user is allowed to see"""
+        return pick_coupons(self.request.user)
+
+    serializer_class = CouponSerializer
```
giocalitri: nit: please move this before the `get_queryset`

ecommerce/views.py
```diff
@@ -152,3 +165,47 @@ def post(self, request, *args, **kwargs):
...
+        SessionAuthentication,
+        TokenAuthentication,
+    )
+
+    def post(self, request, code, *args, **kwargs):
+        """Attach a coupon to a user"""
+        with transaction.atomic():
+            coupon = get_object_or_404(Coupon, coupon_code=code)
+            if not is_coupon_redeemable(coupon, self.request.user):
+                raise Http404
```
giocalitri: Question: no need for an explanation here?

ecommerce/views.py
```diff
@@ -152,3 +165,47 @@ def post(self, request, *args, **kwargs):
...
+
+            UserCoupon.objects.get_or_create(
+                coupon=coupon,
+                user=self.request.user,
+            )
+
+            return Response(status=HTTP_200_OK)
```
giocalitri: I know that this API does not need an actual content, but probably at least a message is something we should return in the content.

giocalitri:
Is the PR description about how to test this still valid?

noisecapella:
Yes it should be

ecommerce/models.py
noisecapella: Sure, I'll reword it

ecommerce/api.py
noisecapella: I wanted to make sure one list was considered before the other one. I think I would need a `UNION ALL` or to add another field in order to use with an `ORDER BY` clause in order to do this in SQL

ecommerce/api.py
noisecapella: Will do

ecommerce/permissions.py
noisecapella: The only username we expose to the user is the edX one. For example `SETTINGS.user.username` is the edX username of the logged in user and it's also the one used in `/api/v0/profiles/<username>/`. This is to not confuse the two in our API

ecommerce/models.py
noisecapella: I don't understand where the redundant query is. `coupon=self` limits the query to only that coupon, and `order__status = FULFILLED` and `order.user != user` limit the query to orders fulfilled by other users

ecommerce/api.py
giocalitri: Or you can just process the two queryset one after the other without translating them in lists...

ecommerce/views.py
noisecapella: What explanation should we have here?

ecommerce/models.py
giocalitri: I think I misread the query (even if I read multiple times). 
Please ignore.

noisecapella:
Comments addressed

ecommerce/api.py
noisecapella: Sure, I'll use itertools to chain them together instead

ecommerce/views.py
giocalitri: why the user is receiving a 404... 
actually I wonder if this needs a 400 instead...
anyway this is pretty minor

noisecapella: I was thinking a 404 would be appropriate since it's basically saying, coupon not found. And the distinction between coupon not found and invalid or unredeemable coupon is irrelevant to the user

giocalitri:
I just notices that multiple requests to `/api/v0/coupons/<code>/users/` with the same code return all a 200 (without modifying the status).
Looking at the code this is the right behavior, but I wonder if:
1- at the first post we need to return a 201
2- at the following posts we need to return a 200

This is pretty minor, so it is up to you.

giocalitri:
👍

noisecapella:
I'm not sure that telling the user whether we created the attachment or just confirmed it exists is relevant to them. That API should be idempotent so that multiple clicks on links will be handled correctly

giocalitri:
well, the behavior would still be idempotent, but from a technical point of view, the right status code is 201, not 200.
anyway, it does not matter that much.

PR #2251, opened 2017-01-04
Fixed test names

noisecapella:
None

#### What's this PR do?
Renames tests to have them run, and fixes ESTestCase to reindex before `setUpTestCase` so that signals can connect to Elasticsearch


rhysyngsun:
:+1:

PR #2255, opened 2017-01-05
Change default log level to INFO

### LINKED ISSUE #2030: Increase default Django log level or filter out template exceptions
We are still getting many template exceptions and I think at this point they all come from third party libraries so there's not much we can do to fix the exceptions at the source

noisecapella:
Fixes #2030

#### What's this PR do?
Changes the default log level

Go into the heroku shell and load an object from the ORM. You shouldn't see any SQL statements output

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -324,8 +324,8 @@ else:
     ADMINS = ()
 
 # Logging configuration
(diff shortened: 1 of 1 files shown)


gsidebo:
@noisecapella right now, when i go into a shell and run queries for objects in -ci and -rc, i don't see any SQL output. should i?

noisecapella:
You should not, that's expected because the SQL statements are only on the DEBUG level

PR #2257, opened 2017-01-05
Upgrade Ramda to 0.23

singingwolfboy:
This upgrade will allow us to use the `R.forEachObjIndexed` function, which @aliceriot suggested I use in an improvement for #2167. [ChangeLog is here.](https://github.com/ramda/ramda/issues/2029) I expect that this is a very low-risk upgrade.

For manual testing, it should be sufficient to load up the application and click around a bit, particularly the Javascript-heavy pages on our site. If everything seems to work, it's probably fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "node-sass": "^4.1.0",
     "object.entries": "1.0.3",
     "postcss-loader": "^1.1.1",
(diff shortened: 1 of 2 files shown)


aliceriot:
functionality looks good!

PR #2259, opened 2017-01-05
Added NODE_MODULES_CACHE

### LINKED ISSUE #2258: Add `NODE_MODULES_CACHE: false` to app.json
This will fix the intermittent deploy failures due to yarn cache problems

noisecapella:
Fixes #2258

#### What's this PR do?
Turns off the node modules cache for heroku builds which fixes problems with yarn install cache deployment

Make sure deployment worked

### DIFF
--- a/app.json
+++ b/app.json
@@ -128,6 +128,10 @@
       "default": "MicroMasters",
       "description": "Application identifier in New Relic."
     },
+    "NODE_MODULES_CACHE": {
+      "description": "If false, disables the node_modules cache to fix yarn install",
+      "value": "false"
+    },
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -128,6 +128,10 @@
       "default": "MicroMasters",
       "description": "Application identifier in New Relic."
     },
+    "NODE_MODULES_CACHE": {
+      "description": "If false, disables the node_modules cache to fix yarn install",
+      "value": false
```
rhysyngsun: `env` values always need to be strings: https://devcenter.heroku.com/articles/app-json-schema#env

noisecapella: Ah, good catch

PR #2261, opened 2017-01-05
Added Facet by Company

### LINKED ISSUE #863: Facet by Company
Company is currently a free text field, where the user can enter multiple values. We can make this a simple flat text facet, although the list may be very long.

annagav:
Fix #863
#### What's this PR do?
Added a MultiSelect 'Company' filter to search for users that currently work at a company.
To test this, run the `recreate_index` command. Go to learners search and test the Company filter.
To test that a user does not appear under company name for a previous position, you can add another previous position for a user who currently works for 'Google':
```
from profiles.models import Profile, Employment
p= Profile.objects.filter(first_name="Rose",last_name="Li")[0]
Employment.objects.create(
  profile=u,
  city="Delisle", 
  state_or_territory="CA-MB", country= "CA", 
  company_name="Volvo",
  position= "Mechanic", industry="Automotive",
  end_date='2014-05-07',
  start_date="2013-08-08")
```
screen shot 2017-01-05 at 5 55 11 pm

screen shot 2017-01-05 at 5 56 08 pm

### DIFF
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -181,13 +181,26 @@ def serialize_program_enrolled_user(program_enrollment):
         'email': user.email
     }
     try:
-        serialized['profile'] = ProfileSerializer(user.profile).data
+        serialized['profile'] = current_work_history(ProfileSerializer(user.profile).data)
     except Profile.DoesNotExist:
         # Just in case
         pass
 
     serialized['program'] = UserProgramSearchSerializer.serialize(program_enrollment)
     return serialized
+
+
+def current_work_history(profile):
+    """
+    Remove work_history objects that are not current
+    """
+    current_employment = []
+    for work in profile['work_history']:
+        if not work['end_date']:
+            current_employment.append(work)
+
+    profile['work_history'] = current_employment
+    return profile
 
 
 def refresh_index():
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -181,7 +181,7 @@ def serialize_program_enrolled_user(program_enrollment):
         'email': user.email
     }
     try:
-        serialized['profile'] = current_work_history(ProfileSerializer(user.profile).data)
+        serialized['profile'] = filter_current_work(ProfileSerializer(user.profile).data)
     except Profile.DoesNotExist:
         # Just in case
         pass
@@ -190,17 +190,17 @@ def serialize_program_enrolled_user(program_enrollment):
     return serialized
 
 
-def current_work_history(profile):
+def filter_current_work(profile):
     """
     Remove work_history objects that are not current
+    Args:
+        profile (dict): serialized user Profile
+    Returns:
+        dict: Profile with filtered current work_history list
     """
-    current_employment = []
-    for work in profile['work_history']:
-        if not work['end_date']:
-            current_employment.append(work)
-
-    profile['work_history'] = current_employment
-    return profile
+    mod_profile = dict(profile)
+    mod_profile['work_history'] = [work for work in mod_profile['work_history'] if not work['end_date']]
+    return mod_profile
 
 
 def refresh_index():
--- a/static/js/components/search/ModifiedMultiSelect.js
+++ b/static/js/components/search/ModifiedMultiSelect.js
@@ -2,28 +2,29 @@
  * Created by anna on 1/9/17.
  */
 import React from 'react';
-
-const Select = require('react-select');
-
+import Select from 'react-select';
+import _ from 'lodash';
 
 
 export default class ModifiedMultiSelect extends React.Component {
 
   constructor(props){
     super(props);
-    this.handleChange = this.handleChange.bind(this)
+    this.handleChange = this.handleChange.bind(this);
   }
 
   handleChange(selectedOptions = []) {
-    this.props.setItems(selectedOptions.map(el => el.value))
+    this.props.setItems(selectedOptions.map(el => el.value));
   }
   render(){
-    const { placeholder, clearable = true, items, selectedItems = [], disabled, showCount, setItems } = this.props
+    const { placeholder, clearable = true, items, selectedItems = [], disabled, showCount } = this.props;
 
     const options = items.map((option) => {
       let label = option.title || option.label || option.key;
-      if (showCount) label += ` (${option.company_name_count.doc_count}) `
-      return { value: option.key, label}
+      if (showCount && !_.isNil(option.company_name_count)){
+        label += ` (${option.company_name_count.doc_count}) `;
+      }
+      return { value: option.key, label };
     });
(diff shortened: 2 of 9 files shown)


pdpinch:
@annagav I know this is still WIP, but I noticed that the screenshots aren't exactly what Rob described in https://github.com/mitodl/micromasters/issues/863#issuecomment-268630134

In addition to the type-ahead, can we display the top 20 hits for the facet? This is important, as it helps users understand what data exists. With just the type-ahead, they have no idea what to type.

annagav:
The doc_count per company shows the total amount of 'work_history' instances corresponding to that company. So in the case where a user had two positions at the same company, it will be added as two occurrences  to the result. Need to use a reverse nested aggregation in the filter that would enable aggregating on parent docs from nested documents.

Ferdi:
@annagav , just to make sure we are on the same page. The result from this filter should answer the question:   "Show me the list of learners who *currently* work at Microsoft".   Is that what this does ?

annagav:
Not currently. This is the behavior we are trying to achieve.

annagav:
@Ferdi, sorry I misunderstood the question. The filter shows all the learners that have worked or currently work at a company.

Ferdi:
Ok, but that's not useful.  We should try to do what I described

Ferdi:
That's not good though. Can we change as I described ?

annagav:
@Ferdi, sounds good, I will filter the 'work_history' objects that do not have end date.

annagav:
The new query that filters out only current jobs did not work with searchkit. So I ended up indexing only current work_history objects.

search/indexing_api.py
```diff
@@ -190,6 +190,19 @@ def serialize_program_enrolled_user(program_enrollment):
...
 
 
+def filter_current_work(profile):
+    """
+    Remove work_history objects that are not current
+    """
+    current_employment = []
+    for work in profile['work_history']:
+        if not work['end_date']:
+            current_employment.append(work)
```
noisecapella: It looks like this case is uncovered FYI. Can you add a test that asserts that only current work history entries end up in the Elasticsearch index?

static/js/components/search/WorkHistoryFilter.js
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+
+  defineAccessor() {
+    return this._accessor;
+  }
+
+  render() {
+    return (
+      <RefinementListFilter
```
noisecapella: It looks like the test is complaining about a lack of an `id` prop. Can you add one? Something like `company_name` should work. You may need to update the accessor above to account for this change

annagav: I added an id prop to RefinementListFilter.

noisecapella:
If I refresh the browser after selecting a facet I get an error in the console:
screenshot from 2017-01-17 13-01-57

annagav:
@noisecapella, should be fixed now

noisecapella:
I'm getting this error:
screenshot from 2017-01-17 16-33-49

noisecapella:
Nevermind, it went away after I restarted the server

search/indexing_api.py
```diff
@@ -190,6 +190,19 @@ def serialize_program_enrolled_user(program_enrollment):
     return serialized
 
 
+def filter_current_work(profile):
```
noisecapella: Can you add arguments and a return value to the docstring?

search/indexing_api.py
```diff
@@ -190,6 +190,19 @@ def serialize_program_enrolled_user(program_enrollment):
...
+    """
+    Remove work_history objects that are not current
+    """
+    current_employment = []
```
noisecapella: It looks like we can use a list comprehension here instead, can you modify the function to return one?

static/js/components/search/ModifiedMultiSelect.js
```diff
@@ -0,0 +1,41 @@
...
+  constructor(props){
+    super(props);
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(selectedOptions = []) {
+    this.props.setItems(selectedOptions.map(el => el.value));
+  }
+  render(){
+    const { placeholder, clearable = true, items, selectedItems = [], disabled, showCount, setItems } = this.props;
```
noisecapella: Can you remove `setItems` here? It looks unused

static/js/components/search/ModifiedMultiSelect.js
```diff
@@ -0,0 +1,41 @@
...
+
+    const options = items.map((option) => {
+      let label = option.title || option.label || option.key;
+      if (showCount && !_.isNil(option.company_name_count)){
+        label += ` (${option.company_name_count.doc_count}) `;
+      }
+      return { value: option.key, label};
```
noisecapella: Can you add a space after `label`?

static/js/components/search/WorkHistoryFilter.js
```diff
@@ -0,0 +1,58 @@
...
+import ModifiedMultiSelect from './ModifiedMultiSelect';
+import type { AvailableProgram } from '../../flow/enrollmentTypes';
+
+export default class WorkHistoryFilter extends SearchkitComponent{
+  props: {
+    currentProgramEnrollment: AvailableProgram,
+  };
+
+  _accessor = new AnonymousAccessor(query => {
+
```
noisecapella: Can you add a comment here describing what this accessor is doing?

annagav:
@noisecapella I addressed comments

search/indexing_api.py
```diff
@@ -193,13 +193,12 @@ def serialize_program_enrolled_user(program_enrollment):
 def filter_current_work(profile):
     """
     Remove work_history objects that are not current
+    Args:
+        profile: serialized profile
```
noisecapella: Can you change `profile` here and below to be `Profile`, to exactly match the profile type?

search/indexing_api.py
```diff
@@ -193,13 +193,12 @@ def serialize_program_enrolled_user(program_enrollment):
...
+    Returns:
+        profile: profile with filtered current work_history list
     """
-    current_employment = []
-    for work in profile['work_history']:
-        if not work['end_date']:
-            current_employment.append(work)
-
-    profile['work_history'] = current_employment
+    profile['work_history'] = [work for work in profile['work_history'] if not work['end_date']]
```
noisecapella: I forgot to mention this earlier but we should make a defensive copy of `profile` so that the original is not modified. Can you add a `profile = dict(profile)` before this line?

search/indexing_api.py
annagav: done

Issue #2211, opened 2016-12-27
MicroMasters Adwords Remarketing Tag Insertion


ChrisLanfranco:
Request: Insert Google Adwords Remarketing Tag right before the </body> for the following pages: 

https://micromasters.mit.edu/
https://micromasters.mit.edu/dedp/
https://micromasters.mit.edu/scm/

Maria Cruz wants to start tracking website visitors for adword remarketing purposes for the MicroMasters program. Inserting this tag will allow us to start tracking these visitors. Attached is the tag. Let me know if you have questions.

Thank you!
Chris L
[Google Adword Remarketing Tag.docx](https://github.com/mitodl/micromasters/files/674481/Google.Adword.Remarketing.Tag.docx)

pdpinch:
@ChrisLanfranco can you explain what the user value of remarketing is? And also how we will know if this is working?

ChrisLanfranco:
@pdpinch 

Google Adwords remarketing is conducted through Google Display Network. The value in it allows us to show ads to users who have already visited our site while browsing the web. 

So in theory, we can improve our conversation rate by spending ad dollars that would have gone to marketing to brand new visitors, towards remarketing to people that have searched for and are familiar with our program already.

How it works:

Add the Adwords tracking url to pages we want to track users on

I create a remarketing list on Adwords with certain rules:

- URL contains category
- Search terms used by visitors 
- Referral urls
- ETC

I can setup rules to target particular visitors to remarket to.

Example: I’m targeting visitors to remarket to by what keywords they searched for to find our site. 

If the user fits the criteria, they’re cookied and added within a list in the Adwords account. Allowing us to remarket/show ads to them within Google Display Network sites. 

Adwords has a verification test to make sure the tag is working properly as well.

pdpinch:
Thanks for the explanation. Definitely seems valuable. Can you point us to the docs on the verification test?

ChrisLanfranco:
https://support.google.com/adwords/answer/3103509?hl=en

I know how to verify on my end if you want to stick the tag within the pages. If you want to verify it, I'll be more than happy to give you admin if you don't already to the adwords account. I'm pretty sure you'll need that. Thx again Peter!

amir-qayyum-khan:
@ChrisLanfranco i need to add below code as it on header of these pages ?
please explain what do you mean by ``right before the for the following pages:``?
should this cover (https://micromasters.mit.edu/) all site, i mean header of all site?

```js
[...]

PR #2263, opened 2017-01-06
Added Adwords Remarketing Tag Insertion

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2211
@pdpinch 

### Note:
set .env variable ```ADWORDS_CONVERSION_ID``` to make it work


ui/templates/base.html
```diff
@@ -58,6 +58,19 @@
         smartlook('init', '{{ APIKEYS.SMARTLOOK }}');
     </script>
     {% endif %}
+    <script type="text/javascript">
+      /* <![CDATA[ */
+      var google_conversion_id = 935224753;
```
rhysyngsun: This value should probably be in an environment variable in case it changes.

ui/templates/base.html
```diff
@@ -58,6 +58,19 @@
         smartlook('init', '{{ APIKEYS.SMARTLOOK }}');
     </script>
     {% endif %}
+    <script type="text/javascript">
```
rhysyngsun: I believe this template is applied to all pages, but the original ticket asks for it to only show up on 3. Is that going to be an issue?

amir-qayyum-khan: @rhysyngsun please see this thread and https://github.com/mitodl/micromasters/issues/2211#issuecomment-270929044

amir-qayyum-khan:
Done @rhysyngsun

micromasters/settings.py
```diff
@@ -437,6 +437,7 @@ def get_var(name, default):
 STATUS_TOKEN = get_var("STATUS_TOKEN", "")
 HEALTH_CHECK = ['CELERY', 'REDIS', 'POSTGRES', 'ELASTIC_SEARCH']
 
+ADWORDS_CONVERSION_ID = get_var("ADWORDS_CONVERSION_ID", "")
```
rhysyngsun: Should this get added to `app.json` as an optional var so it shows up in the Heroku settings dashboard?

rhysyngsun:
@amir-qayyum-khan code looks good, feel free to merge after adding to app.json and it deploys correctly.

PR #2267, opened 2017-01-06
Document how to get a Google API key

singingwolfboy:
Follow-up to #2167. This pull request provides developer-facing documentation for how to get a Google API key, so that they can use features like the Geosuggest element in their own local development setup.


blarghmatey:
👍

PR #2269, opened 2017-01-06
Upgrade test dependencies

### LINKED ISSUE #2268: Pytest issue
See https://travis-ci.org/mitodl/micromasters/jobs/189607827

This error was caused by this test:

    def test_no_more_coupons(self):
        """If user has no redemptions left the coupon should not be redeemable"""
        coupon = CouponFactory.create(content_object=self.program)
        with patch('ecommerce.api.Coupon.user_has_redemptions_left', autospec=True) as _user_has_redemptions:
            _user_has_redemptions.return_value = False
            assert is_coupon_redeemable(coupon, self.user) is False
        assert _user_has_redemptions.call_count == 1
        _user_has_redemptions.assert_called_with(self.user)

It looks like a pytest issue. Upgrading to 3.0.5 fixes it

noisecapella:
Fixes #2268

#### What's this PR do?
Upgrades test dependencies which incidentally fixes a pytest issue. I skipped upgrading Django since that conflicts with `requirements.txt` and it seems like it would require some extra work

Since these are only test dependencies we shouldn't need to do manual testing here, if the tests still pass

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -49,7 +49,7 @@ pytest-django==3.1.2
 pytest-pep8==1.0.6
 pytest-pylint==0.6.0
 pytest==3.0.5
-requests==2.12.4          # via bpython, codecov
+requests==2.11.1          # via bpython, codecov
 semantic-version==2.6.0
 simplegeneric==0.8.1      # via ipython
 six==1.10.0               # via astroid, bpython, nplusone, pip-tools, prompt-toolkit, pylint, pytest-pylint, traitlets
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -6,61 +6,62 @@
 #
 apipkg==1.4               # via execnet
 argparse==1.4.0           # via codecov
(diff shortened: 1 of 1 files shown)


test_requirements.txt
```diff
@@ -6,61 +6,62 @@
...
 pytest-cache==1.0         # via pytest-pep8
 pytest-cov==2.4.0
-pytest-django==3.0.0
+pytest-django==3.1.2
 pytest-pep8==1.0.6
 pytest-pylint==0.6.0
-pytest==3.0.3
-requests==2.11.1          # via bpython, codecov
+pytest==3.0.5
+requests==2.12.4          # via bpython, codecov
```
singingwolfboy: Looks like Travis is complaining that the version of `requests` specified here, and the version specified in the `requirements.txt` file, are different. I guess these will need to be kept in lockstep with each other.

noisecapella: Yes, I'll revert those changes then

PR #2274, opened 2017-01-09
Fixed progress widget ignores prior (passed) runs issue

amir-qayyum-khan:
- Fixed https://github.com/mitodl/micromasters/issues/2271

#### What's this PR do?
- Check past runs of course i.e if we have passed run in past it will count course as passed.

@pdpinch

### DIFF
--- a/static/js/components/ProgressWidget_test.js
+++ b/static/js/components/ProgressWidget_test.js
@@ -81,7 +81,44 @@ export const program = {
       "title": "EDX Demo course",
       "description": "",
       "id": 3
-    }
+    },
+    {
+      "prerequisites": "",
+      "runs": [],
+      "position_in_program": 2,
+      "title": "EDX Demo course",
+      "description": "",
+      "id": 4
+    },
+    {
+      "prerequisites": "",
+      "runs": [
+        {
+          "position": 1,
+          "title": "Gio Test Course #14",
+          "course_id": "course-v1:odl+GIO101+FALL14",
+          "status": STATUS_NOT_PASSED,
+          "id": 5,
+          "course_start_date": "2016-08-22T11:48:27Z",
+          "fuzzy_start_date": "Fall 2016",
+          "course_end_date": "2016-09-09T10:20:10Z"
+        },
+        {
+          "position": 2,
+          "title": "Gio Test Course #16",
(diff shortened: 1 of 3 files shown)


static/js/components/ProgressWidget_test.js
```diff
@@ -41,11 +41,21 @@ export const program = {
...
-          "status": STATUS_NOT_PASSED,
+          "status": STATUS_PASSED,
           "id": 2,
           "course_start_date": "2016-08-22T11:48:27Z",
-          "fuzzy_start_date": "Fall 2017",
+          "fuzzy_start_date": "Fall 2016",
           "course_end_date": "2016-09-09T10:20:10Z"
+        },
+        {
+          "position": 2,
```
singingwolfboy: You've added a second run of this course, but the first is passing and the second is not. Based on the way the code used to be, this automated test would pass either way.

Can you modify another one of the courses in this program to have multiple runs, and set the status of the first course run to `STATUS_NOT_PASSED`? That would be an appropriate test of this change.

static/js/util/util.js
```diff
@@ -386,7 +386,7 @@ export function programCourseInfo(program: Program): Object {
   if (program.courses) {
     totalCourses = program.courses.length;
     passedCourses = program.courses.filter(
-      course => course.runs.length > 0 && course.runs[0].status === STATUS_PASSED
+      course => _.some(course.runs, ["status", STATUS_PASSED])
```
singingwolfboy: This works, but I had to look up the documentation for [`_.some`](https://lodash.com/docs#some) to understand what's going on. Can you add a comment to explain this?
```js
// check if any course run has a `status` property set to STATUS_PASSED
course => _.some(course.runs, ["status", STATUS_PASSED])  
```

amir-qayyum-khan: it check if in an array any object has status=== STATUS_PASSED then it will return true. No further check.

singingwolfboy: Yes, I understand that. However, it's not very clear just by looking at the code. That's why I'm asking you to add a comment, as I demonstrated in my comment above.

amir-qayyum-khan:
Done @singingwolfboy

singingwolfboy:
Looks good to me! 👍

Issue #2158, opened 2016-12-19
Coupons 


pdpinch:
Design document at https://goo.gl/CMlzHd

We have four near-term use cases for creating coupons for learners:

1. 100% discount coupons for learners who have passed an economics course in the past and will participate in the Pearson exam pilot.
2. 100% discount coupons so members of certain organizations can take all DEDP MicroMasters courses for free.
3. fixed discount for learners who have previously paid for verified certificates
4. Fixed price coupon for affiliated staff.

### Acceptance criteria

- [x] #2192 model for managing coupons
   - type (percentage off; fixed discount)
   - level (program level, course level or (maybe) course-run)
   - for the future, expiration date
- [x] #2222 model to map a user to a coupon
- [x] dashboard: update pricing to reflect coupon 
  - [x] #2322 populate redux store with coupon info API 
  - [x] #2356 calculate prices in frontend 
  - [x] #2355 coupon messaging 
  - [x] #2403 display calculated prices in UI
  - [x] #2335 When price is $0, Pay now should become Enroll now 
  - [x] #2334 hide financial aid options if user has a 100% off coupon for whole program
- [x] #2328 Order Summary page
- [x] #2237 integration with checkout and orders
  - [x] #2321 allow $0 checkouts in checkout API 
  - [x] #2325 redeem coupon in checkout
- [x] method for creating coupons (separate from distributing)
- [x] distribution of  coupons (will be handled by course team)
- [x] #2323 learner interface for applying coupon 
  - [x] #2405 handle anonymous users
- [x] #2400 admin interface for creating coupons
- [x] #2399 fixed price coupons 
- [x] #2195 create audit table to preserve any changes to coupons

noisecapella:
Pseudo-code for redeeming a coupon with models from #2192:

    def redeem_coupon(coupon, user):
        with transaction.atomic():
            if coupon.num_coupons_available == 0:
                raise ValidationError("No coupons left")
            num_redemptions = RedeemedCoupon.objects.filter(order__user=user, coupon=coupon).count()
            if num_redemptions >= coupon.num_redemptions_per_user:
                raise ValidationError("Coupon has no more redemptions left")

            order = create_unfulfilled_order(...)
            RedeemedCoupon.objects.create(order=order, coupon=coupon)
            order.status = 'fulfilled'
            order.save()

pdpinch:
Updated the acceptance criteria to include new issues that George wrote. Also added the two new uses cases, for a total of 4.

[...]

PR #2282, opened 2017-01-09
Added UserCoupon, removed num_... fields, added helper properties

noisecapella:
Part of #2158 

#### What's this PR do?
Updates coupon models to handle ability to associate a `Coupon` with `User` without having them redeem it

There's no user facing functionality yet


ecommerce/models.py
```diff
@@ -243,18 +286,44 @@ def clean(self):
...
 
     def __str__(self):
         """Description for Coupon"""
-        return "Coupon {amount_type} {amount} for {product}, {num_coupons_available} left".format(
+        return "Coupon {amount_type} {amount} of type {type} for {product}".format(
             amount_type=self.amount_type,
             amount=self.amount,
             product=self.content_object,
-            num_coupons_available=self.num_coupons_available,
+            type=self.coupon_type,
```
giocalitri: please do not use reserved words

giocalitri:
ok... just the questions

ecommerce/models.py
```diff
@@ -176,6 +179,10 @@ class Coupon(Model):
 
     AMOUNT_TYPES = [PERCENT_DISCOUNT, FIXED_DISCOUNT]
 
+    STANDARD = 'standard'
+    DISCOUNTED_PREVIOUS_COURSE = 'discounted-previous-course'
```
noisecapella: Yes

ecommerce/models.py
```diff
@@ -223,6 +228,44 @@ class Coupon(Model):
...
+    def course_keys(self):
+        """Get the course keys which the coupon can be redeemed with"""
+        obj = self.content_object
+        if isinstance(obj, Program):
+            return CourseRun.objects.filter(course__program=obj).values_list('edx_course_key', flat=True)
+        elif isinstance(obj, Course):
+            return CourseRun.objects.filter(course=obj).values_list('edx_course_key', flat=True)
+        else:
+            # This must be a CourseRun. This is validated in clean()
+            return [obj.edx_course_key]
```
noisecapella: Sure, seems reasonable

ecommerce/models.py
```diff
@@ -223,6 +228,44 @@ class Coupon(Model):
...
+    @property
+    def is_valid(self):
+        """Returns true if the coupon is enabled and also within the valid date range"""
+        now = datetime.now(tz=pytz.UTC)
+        if not self.enabled:
+            return False
+        if self.activation_date is not None and now < self.activation_date:
+            return False
+        if self.expiration_date is not None and now > self.expiration_date:
+            return False
```
noisecapella: Sounds good

ecommerce/models.py
```diff
@@ -223,6 +228,44 @@ class Coupon(Model):
...
+        return True
+
+    @property
+    def is_automatic(self):
+        """
+        Returns true if the coupon would be redeemed automatically without input from the user.
+        """
+        return self.coupon_type == self.DISCOUNTED_PREVIOUS_COURSE
```
noisecapella: Because there may be other coupons which are automatically applied but not necessarily due to having purchased the previous course. Keeping `is_automatic` separate from `DISCOUNTED_PREVIOUS_COURSE` means `is_automatic` can change to handle other types of coupons in the future

giocalitri: ok

ecommerce/models.py
noisecapella: Sure

ecommerce/models.py
```diff
@@ -223,6 +228,44 @@ class Coupon(Model):
...
+
+    def user_has_redemptions_left(self, user):
+        """
+        Has the coupon already been redeemed for all possible runs?
+        """
+        runs_purchased = set(Line.objects.filter(order__user=user).values_list("course_key", flat=True))
+        return not set(self.course_keys).issubset(runs_purchased)
```
noisecapella: I should fix the docstring, it should say "Has the coupon **not** already been redeemed for all possible runs". The `Line` filter also does not filter on `Order.status` = `fulfilled`, so I'll fix that too. Does that clear this up? The expression there basically says, if the all runs the coupon is for have already been purchased, then the coupon cannot be redeemed. The `not` is because `user_has_redemptions_left` is the opposite case

noisecapella:
Ok, that should be all of them I think

giocalitri:
👍 after tests pass

PR #2283, opened 2017-01-09
Remove jsonfield dependency

singingwolfboy:
Now that Django has a built-in [JSONField type](https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#jsonfield), we don't need to depend on the [jsonfield](https://pypi.python.org/pypi/jsonfield) library. This pull request removes that dependency.

To test this pull request, I suggest dumping your database and re-running all the database migrations to be sure they work successfully. I did that on my computer, and had no problems.

### DIFF
--- a/dashboard/migrations/0001_initial.py
+++ b/dashboard/migrations/0001_initial.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 from django.conf import settings
 from django.db import migrations, models
 import django.db.models.deletion
-import jsonfield.fields
+from django.contrib.postgres.fields import JSONField
 
 
 class Migration(migrations.Migration):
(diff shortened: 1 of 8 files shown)


noisecapella:
I don't think this is a good idea. The field used by the jsonfield dependency isn't the same postgres field as the jsonb field in the django json field type. I'm also generally against modifying our migrations unless we really need to. What happens if people need to roll back migrations at some point?

singingwolfboy:
Theoretically, the Django JSONField type should behave identically to the `jsonfield` type. However, theory and practice don't always align, so I understand where you're coming from.  Reducing dependencies is good, but reducing migration-related risk is also good. I'll close this pull request for now, and we can always come back to it later if we decide it's a good idea.

noisecapella:
They do not behave identically. The Django JSONField type uses the `jsonb` postgres type, the jsonfield type uses the `json` postgres type (source: https://github.com/bradjasper/django-jsonfield/issues/57#issuecomment-103937355). One of the reasons we migrated in the first place was to get query functionality that we didn't have with the json type

PR #2289, opened 2017-01-09
Added coupon functions

noisecapella:
Part of #2158 
Rebased on #2282 

#### What's this PR do?
Adds functions to `dashboard/utils.py` and `ecommerce/api.py` to test coupon redeemability

There's no user facing piece yet

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -19,15 +19,23 @@ import pytz
 from rest_framework.exceptions import ValidationError
 
 from backends.edxorg import EdxOrgOAuth2
-from courses.models import CourseRun
-from dashboard.api_edx_cache import CachedEdxDataApi
+from courses.models import (
+    CourseRun,
+    Program,
+)
+from dashboard.api_edx_cache import (
+    CachedEdxDataApi,
+    CachedEdxUserData,
+)
 from dashboard.models import ProgramEnrollment
+from dashboard.utils import MMTrack
 from ecommerce.exceptions import (
     EcommerceEdxApiException,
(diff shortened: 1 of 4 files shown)


ecommerce/api.py
```diff
@@ -310,3 +318,48 @@ def enroll_user_on_success(order):
 
     if exceptions:
         raise EcommerceEdxApiException(exceptions)
+
+
+def is_coupon_redeemable_for_run(coupon, user, course_key):
+    """
+    Returns true if the coupon is redeemable for the user for a course run.
+    """
```
rhysyngsun: Could you add `Args`, `Returns`, etc for this and the other method added here?

ecommerce/api.py
```diff
@@ -310,3 +318,48 @@ def enroll_user_on_success(order):
...
+        mmtrack = MMTrack(
+            user,
+            program,
+            edx_user_data,
+        )
+
+        # For this coupon type the user must have already purchased a course run on edX
+        return any((mmtrack.has_verified_cert(run.edx_course_key) for run in course.courserun_set.all()))
+
+    return True
```
rhysyngsun: I'm on the fence as to whether the default behavior should be for a valid coupon rather than defaulting to `False` and returning `True` for specific cases. I can see that making this code far less elegant though.

ecommerce/api.py
noisecapella: Sure

noisecapella:
Anything else @rhysyngsun?

PR #2290, opened 2017-01-09
Change status for enrollment to audit, since it's used in FA programs

noisecapella:
None

#### What's this PR do?
Changes the enrollment status in test data from `verified` to `audit` for test data, since this course run is used in FA programs and users are always `audit` on edX for FA programs


giocalitri:
why is this change necessary?

noisecapella:
It's not necessary, it just makes the test data more realistic since FA courses are never "verified" on edX

giocalitri:
ok 👍

PR #2296, opened 2017-01-09
remove extraneous about_me serializer fields

singingwolfboy:
These fields were introduced in #1862, but they aren't actually necessary. Because these classes inherit from `ModelSerializer`, they automatically generate these fields, so there's no need to specify them (except in the `Meta.fields` list, where they're already specified).

### DIFF
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -9,7 +9,6 @@ from rest_framework.serializers import (
     IntegerField,
     ModelSerializer,
     SerializerMethodField,
-    CharField,
 )
 
 from profiles.api import get_social_username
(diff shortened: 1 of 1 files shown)


giocalitri:
I am not sure I understood the explanation in this PR.

singingwolfboy:
`ProfileSerializer` and `ProfileFilledOutSerializer` both inherit from [`ModelSerializer`](http://www.django-rest-framework.org/api-guide/serializers/#modelserializer), which introspects a Django model class and automatically generates serializer fields based on the model fields. Because of this introspection, there's no need to declare these serializer fields manually, the way this code was doing. All that's necessary is to make sure that `amount_me` is referenced in the `fields` list, and it already is listed there.

@giocalitri is that clearer?

giocalitri:
ok, now it is clear: what I understood was that the field was defined in the base class (and I could not find it for obvious reasons)

giocalitri:
this looks fine to me 👍

PR #2300, opened 2017-01-10
Don't need to make pylint disable missing-docstring for serializer Meta

singingwolfboy:
[pylint-django](https://github.com/landscapeio/pylint-django) takes care of suppressing this

### DIFF
--- a/cms/serializers.py
+++ b/cms/serializers.py
@@ -26,7 +26,7 @@ class FacultyImageSerializer(serializers.ModelSerializer):
         rendition = image.get_rendition('fill-500x385')
         return RenditionSerializer(rendition).data
 
-    class Meta:  # pylint: disable=missing-docstring
+    class Meta:
         model = Image
         fields = ('alt', 'rendition',)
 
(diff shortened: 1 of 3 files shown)


giocalitri:
if I grep the code in your branch, there are several other places where that pylint disable is still present next to Meta classes

singingwolfboy:
@giocalitri Unfortunately, Pylint is only smart enough to figure this out for Django models and DRF serializers. It doesn't know about Factory Boy factories, which also have a Meta class. [I started building a pylint plugin to teach it about Factory Boy](https://github.com/singingwolfboy/pylint-factory), based on `pylint-django`, but it's not working yet.

giocalitri:
ok... that is fine. 
I did not give my 👍 before you merged though :-)
(no problem for this PR)

singingwolfboy:
Oops! I saw you switched the label to "Waiting on Author", and for some reason my brain parsed that as "Waiting on Author to Merge". 🤦‍♂️

PR #2302, opened 2017-01-10
Allow users to pay after end of the course.

### LINKED ISSUE #2253: Dashboard: allow user to pay after the end of the course, until the "upgrade" deadline"
Some courses (typically financial aid courses) can be paid for even after the course has ended. 

We should rely on the course's upgrade deadline for the logic for displaying the payment button, and not the course end date.

giocalitri:
closes #2253 

#### What's this PR do?
Removes the restriction about the fact that the payment must be done by the end of the course. Now the only limit is the `upgrade_deadline`, if present.

`dashboard/api.py`

Set the course to `past` and with `upgrade_deadline` in the future and verify that you can still see the pay button.
See the code for special cases with the dates and final grades existence.

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -271,6 +271,17 @@ class CourseRun(models.Model):
             raise ImproperlyConfigured('Missing freeze_grade_date')
         return datetime.now(pytz.utc) > self.freeze_grade_date
 
+    @property
+    def has_frozen_grades(self):
+        """
+        Checks if the grades for the course have been frozen
+        """
+        try:
+            freeze_status = self.courserungradingstatus
+        except self._meta.model.courserungradingstatus.RelatedObjectDoesNotExist:
+            return False
+        return freeze_status.status == FinalGradeStatus.COMPLETE
+
     @classmethod
     def get_freezable(cls):
         """
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -11,6 +11,7 @@ import pytz
 from courses.models import Program
 from dashboard.api_edx_cache import CachedEdxDataApi, CachedEdxUserData
 from dashboard.utils import MMTrack
+from grades.models import FinalGrade
 
 log = logging.getLogger(__name__)
 
@@ -34,7 +35,7 @@ class CourseStatus:
     def all_statuses(cls):
         """Helper to get all the statuses"""
         return [cls.PASSED, cls.NOT_PASSED, cls.CURRENTLY_ENROLLED, cls.PAID_BUT_NOT_ENROLLED,
-                cls.CAN_UPGRADE, cls.OFFERED, cls.WILL_ATTEND, cls.MISSED_DEADLINE]
+                cls.CAN_UPGRADE, cls.OFFERED, cls.WILL_ATTEND, cls.MISSED_DEADLINE, ]
 
 
 class CourseRunStatus:
@@ -258,7 +259,7 @@ def get_info_for_course(course, mmtrack):
     elif run_status.status == CourseRunStatus.CAN_UPGRADE:
         _add_run(run_status.course_run, mmtrack, CourseStatus.CAN_UPGRADE)
     elif run_status.status == CourseRunStatus.PAID_BUT_NOT_ENROLLED:
-        _add_run(run_status.course_run, mmtrack, CourseRunStatus.PAID_BUT_NOT_ENROLLED)
+        _add_run(run_status.course_run, mmtrack, CourseStatus.PAID_BUT_NOT_ENROLLED)
 
     # add all the other runs with status != NOT_ENROLLED
     # the first one (or two in some cases) has been added with the logic before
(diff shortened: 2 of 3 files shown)


noisecapella:
I have `None` for `upgrade_deadline` and `enrollment_end` and `end_date` is `2017-01-01`. I'm getting 'No future courses are currently scheduled' in the dashboard. If I then set `end_date` to `None` I see the enroll and pay buttons. I think some logic in the dashboard API needs to be updated

noisecapella:
After adding the `CachedEnrollment` it works fine for me. Looking at code...

courses/models.py
```diff
@@ -263,6 +263,17 @@ def can_freeze_grades(self):
...
         return datetime.now(pytz.utc) > self.freeze_grade_date
 
+    @property
+    def has_frozen_grades(self):
+        """
+        Checks if the grades for the course have been frozen
+        """
+        try:
+            freeze_status = self.courserungradingstatus
+        except self._meta.model.courserungradingstatus.RelatedObjectDoesNotExist:
```
noisecapella: Does `CourseRunGradingStatus.DoesNotExist` not work here?

dashboard/api.py
```diff
@@ -293,19 +294,53 @@ def get_status_for_courserun(course_run, mmtrack):
...
-        if course_run.is_past:
-            status = CourseRunStatus.NOT_PASSED
+        elif course_run.has_frozen_grades:
+            status = CourseRunStatus.CHECK_IF_PASSED
+        # this last check needs to be done as last one
+        elif course_run.is_past:
+            status = CourseRunStatus.CURRENTLY_ENROLLED
         else:
+            raise ImproperlyConfigured(
+                'The course {0} results not being current, past or future at the same time'.format(
```
noisecapella: Can you change this to `The course {0} results should not be current, past, or future at the same time`?

dashboard/api.py
```diff
@@ -293,19 +294,53 @@ def get_status_for_courserun(course_run, mmtrack):
...
+                    final_grade = None
+                    try:
+                        final_grade = FinalGrade.objects.get(user=mmtrack.user, course_run=course_run)
+                    except FinalGrade.DoesNotExist:
+                        # this is a very special case that happens if the user has logged in
+                        # for the first time after we have already frozen the final grades
+                        # this should be eventually changed to freeze the final grade sync
+                        status = CourseRunStatus.CURRENTLY_ENROLLED
+                        log.exception(
+                            'The user "%s" has not a final grade for the course run "%s" '
```
noisecapella: Can you change the wording to `The user "%s" doesn't have a final grade for the course...`?

dashboard/api_test.py
```diff
@@ -372,7 +383,32 @@ def test_enrolled_upgradable(self):
...
+            end=None,
+            enr_start=None,
+            enr_end=None,
+            edx_key="course-v1:MITx+8.MechCX+2014_T1"
+        )
+        with self.assertRaises(ImproperlyConfigured):
+            api.get_status_for_courserun(crun, self.mmtrack)
+
+    @ddt.data(
+        (False, None, api.CourseRunStatus.MISSED_DEADLINE),
```
noisecapella: Should this be `None` be here? It doesn't look like `has_frozen_grades` will ever return that

dashboard/api_test.py
```diff
@@ -387,9 +423,59 @@ def test_not_paid_not_passed(self):
...
+            upgr_mock.return_value = is_upgradable
+            froz_mock.return_value = has_frozen_grades
+            run_status = api.get_status_for_courserun(crun, self.mmtrack)
+        assert run_status.status == expected_status
+        assert run_status.course_run == crun
+
+    @patch('grades.models.FinalGrade.objects.get', autospect=True)
+    @patch('courses.models.CourseRun.is_upgradable', new_callable=PropertyMock)
+    @patch('courses.models.CourseRun.has_frozen_grades', new_callable=PropertyMock)
+    def test_not_paid_in_past_grade_frozen(self, upgr_mock, froz_mock, final_grades_mock):
```
noisecapella: I might be reading this wrong but aren't `upgr_mock` and `froz_mock` in the wrong order?

dashboard/api_test.py
```diff
@@ -869,6 +979,13 @@ def test_past_course_runs(self):
         )
 
         result = api.get_user_program_info(self.user, self.edx_client)
+        # extract the right program from the result
+        program_result = None
```
noisecapella: What's the significance of this addition here?

courses/models.py
giocalitri: no, it would require to import `CourseRunGradingStatus` triggering a circular import error

dashboard/api.py
giocalitri: will go with `The course {0} results are not either current, past, or future at the same time`
is it ok?

dashboard/api.py
giocalitri: sure

dashboard/api_test.py
giocalitri: the `@ddt.unpack` needs a value in the tuple, does not?

dashboard/api_test.py
giocalitri: no, you are right

dashboard/api.py
noisecapella: Sounds good to me

dashboard/api_test.py
giocalitri: this integration test has 2 programs generated via the `setUpTestData` and the previous version of the test was assuming that the relevant one was the first in the result (and that is true) but there is no guarantee that this is always the case

dashboard/api_test.py
noisecapella: I think `@ddt.unpack` looks fine here but I'm wondering about the `None` in `False, None, api.CourseRunStatus.MISSED_DEADLINE` above. Is that intentional? It seems like `has_frozen_grades` should return either `True` or `False`, not `None`

giocalitri: it is irrelevant what it does return, because the code should never reach that point.
Should I make the test fail intentionally in that case?

noisecapella: No, if the value is not used feel free to leave it as is

giocalitri:

PR #2304, opened 2017-01-10
Filter filled_out fields

### LINKED ISSUE #2301: Search: don't index users whose profiles aren't filled out
This just checks the `filled_out` field. We should still assume new fields will be created which will be null even if `filled_out` is true

noisecapella:
Fixes #2301 

#### What's this PR do?
Adds filtering so that only `filled_out` profiles are seen via the search API.

With one account, create a profile which is partially filled out. Complete the personal step and go to the next step, but do not go any further.

With a staff account, view the learners search. You should not see any learners in the results.

Log back in with the first account and complete your profile. Then log back in as staff and view the learners search. You should see one learner in the results, the one you just created.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -79,6 +79,13 @@ def create_search_obj(user, search_param_dict=None, filter_on_email_optin=False)
         user,
         filter_on_email_optin=filter_on_email_optin
     ))
+    # Filter so that only filled_out profiles are seen
+    search_obj = search_obj.filter(Q(
+        'bool',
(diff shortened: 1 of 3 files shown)


gsidebo:
might be out of scope for this PR, but if all of a Program's users have `filled_out=False`, then you get this old horrible results page:

ss 2017-01-10 at 13 03 14

if you don't think that's in scope for this PR, this is otherwise good to go 👍

PR #2309, opened 2017-01-10
Add --reuse-db flag to speed up running Python tests locally

### LINKED ISSUE #2307: Add --reuse-db to pytest.ini
This will tell pytest to tell Django to reuse the test database so that we don't run all migrations each time the tests are run. The migrations will be run if the test database does not yet exist. This greatly speeds up running Python tests locally. There's a potential for migration conflicts if migrations change, but the user can just remove `--reuse-db` and rerun the tests to fix these problems

noisecapella:
Fixes #2307 

#### What's this PR do?
Adds --reuse-db flag to speed up running Python tests locally.

Run the Python tests twice, maybe on a single file. The second run should be notably faster than the first. No tests should fail.


giocalitri:
this looks good to me 👍

PR #2310, opened 2017-01-10
don't set 'use-credentials'

aliceriot:
This option was set during development / testing and snuck into the merged PR

### DIFF
--- a/webpack.config.prod.js
+++ b/webpack.config.prod.js
@@ -19,7 +19,7 @@ module.exports = Object.assign(prodConfig, {
     path: path.resolve('./static/bundles/'),
     filename: "[name]-[chunkhash].js",
     chunkFilename: "[id]-[chunkhash].js",
-    crossOriginLoading: "use-credentials",
+    crossOriginLoading: "anonymous",
   },
 
   plugins: [


webpack.config.prod.js
```diff
@@ -19,7 +19,6 @@ module.exports = Object.assign(prodConfig, {
     path: path.resolve('./static/bundles/'),
     filename: "[name]-[chunkhash].js",
     chunkFilename: "[id]-[chunkhash].js",
-    crossOriginLoading: "use-credentials",
```
rhysyngsun: I think we actually need to set this to "anonymous" because the default is `false`, which disables it entirely.

aliceriot: cool, will do

rhysyngsun:
Assuming the tests pass :+1:

aliceriot:
Do you think we should confirm it works ok with cloudfront first? Looking into setting it up on the PR build to test.

aliceriot:
alright, so the cloudfront stuff is set up on the PR build now, trying to get that working to confirm the fix is good

aliceriot:
@noisecapella So the `CLOUDFRONT_DIST` env var is set, and it seems like django is picking up on that, but none of the static assets are being fetched. Do you know if there's any other configuration I need to do for it?

aliceriot:
It seems like it is working now! I think before an issue was that heroku for some reason was not doing the python portion of the build, not sure.

But @rhysyngsun do you want to confirm? I think just confirm that if you go to the PR review app you can use the learner search page.

rhysyngsun:
:+1: Confirmed the js bundles are loading without error from CF on the PR build.

PR #2316, opened 2017-01-11
Fixed seed_db commands to work with indexing/grade changes

gsidebo:

No ticket

#### What's this PR do?

2 recent changes caused some problems for the `seed_db` command:

1. Omitting users with `filled_out=False` from search results (this exposed a minor bug in the script)
1. Final grade freezing

Unseed and re-seed the db with your user as the `staff-user` parameter (`seed_db --staff-user=YOUR_USERNAME`), and check out the /learners search results. All of the seed data should be there

Details about the bugs being fixed:

- `Profile` deserialization wasn't properly setting `filled_out=True` for all of the test users
- The `FinalGrade` table needed to be cleared of fake course runs before the fake programs could be deleted

### DIFF
--- a/seed_data/lib_test.py
+++ b/seed_data/lib_test.py
@@ -1,0 +1,76 @@
+"""
+Tests for library functions used by seed_db and alter_data commands
+"""
+
+from django.test import TestCase
+from micromasters.factories import UserFactory
+from courses.factories import ProgramFactory, CourseFactory, CourseRunFactory
+from seed_data.lib import (
+    CourseFinder,
+    CourseRunFinder,
+    UserFinder,
+)
+
+
+class UserFinderTests(TestCase):
+    """Test cases for UserFinder"""
+    @classmethod
+    def setUpTestData(cls):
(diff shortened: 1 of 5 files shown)


gsidebo:
The unit tests I added are not related to this PR. I did it to stop codecov from complaining

seed_data/lib_test.py
```diff
@@ -0,0 +1,73 @@
...
+class UserFinderTests(TestCase):
+    """Test cases for UserFinder"""
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = UserFactory.create(username='username1', email='email1@example.com')
+        UserFactory.create(username='username2', email='email2@example.com')
+
+    def test_user_finder_success(self):
+        """Tests that UserFinder will return a desired user"""
+        found_users = []
```
rhysyngsun: Nit: I think this would be a little cleaner if you moved all the finds into the literal instead of appending. Same for the other tests here.

gsidebo:
@rhysyngsun should be ready

PR #2324, opened 2017-01-11
Revert "Deleted .babelrc"

aliceriot:
fixin a thing

as long as the JS tests run without an error the fix is good


giocalitri:
👍

giocalitri:
I would have used

PR #2327, opened 2017-01-11
Filtered out coupons redeemed by another user

### LINKED ISSUE #2326: Coupons: prevent users from redeeming a coupon already redeemed by someone else
Part of #2158

I missed this case when implementing `is_coupon_redeemable`. We should check that no `RedeemedCoupon` is present for the coupon and any other user, with the exception of automatic coupons which are allowed to be redeemed by many other users.

noisecapella:
Fixes #2326 
#### What's this PR do?
Changes `user_has_redemptions_left` function to return false if another user has already redeemed the `Coupon` at least once, and the `Coupon` is not automatic

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -266,13 +266,41 @@ class Coupon(Model):
 
     def user_has_redemptions_left(self, user):
         """
-        Has the coupon not already been redeemed for all possible runs?
+        Has the coupon not already been redeemed by another user, and has the user not already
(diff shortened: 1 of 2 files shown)


rhysyngsun:
@noisecapella this has no user-facing components to test, correct?

noisecapella:
Right, there is no REST API yet

rhysyngsun:
Thought so :+1:

PR #2330, opened 2017-01-11
Added timestamps for Coupon-related models

### LINKED ISSUE #2329: Coupons: add created_at and modified_at fields
We should add these fields to each of the three coupon models

noisecapella:
Fixes #2329

#### What's this PR do?
Adds timestamp fields to coupon-related models


ecommerce/models.py
```diff
@@ -231,6 +231,9 @@ class Coupon(Model):
     )
     enabled = BooleanField(default=True, help_text="If true, coupons are presently redeemable")
 
+    created_on = DateTimeField(auto_now_add=True)
```
rhysyngsun: I think we have a standard `TimestampedModel` in `micromasters.models` that has these, is there a reason not to use that?

noisecapella: No, I'll change this to use them

Issue #2077, opened 2016-12-12
Exams: Trigger exam authorizations when users pass and authorize existing users who already passed


rhysyngsun:
Part of #1796 

We need to ensure that the exam authorizations are properly triggered when a user passes a course in edX. We also want users who previously passed a course to get retroactively authorized to take the exam

Criteria for authorizing a user:
- Passed a course in edX.
- Paid for the course.

### Tasks:
- When a user pays for the class, we should: 
  - [ ] Trigger the `ExamProfile` record to be created
  - [ ] Trigger the `ExamAuthorization` record to be created.
- Where we currently determine that the user has passed we should:
  - [ ] Trigger the `ExamAuthorization` record to be created.

### Conditions:
- When we trigger the `ExamProfile` record to be created the following conditions must be met:
  - [ ] Learner must have paid for the course.
  - [ ] `Program.exam_series_code` must be set for the course's program
  - [ ] `Course.exam_module` must be set for the course
- When we trigger the `ExamAuthorization` record to be created the following conditions need to be met:
  - [ ] Learner must have passed the course.
  - [ ] Learner must have paid for the course.
  - [ ] `Program.exam_series_code` must be set for the course's program
  - [ ] `Course.exam_module` must be set for the course

**We won't worry about retroactively authorizing users in this issue**

Open Questions:

- [x] Since it's related to other issues we're working on, is there anything we need to handle special around users passing a course multiple times? At the very least I can ensure we won't have an error if that happens.

pdpinch:
Answer to the open question: the logic to determine whether a use has passed a course should take the max of all the user's enrollments.

rhysyngsun:
@pdpinch I added a requirement for max retakes because I saw that this was on one of the documents:

> Students can only take the same module a maximum of two attempts

rhysyngsun:
Looks like discussion of the eligibility dates happened over here: https://github.com/mitodl/micromasters/issues/2076#issuecomment-266932121

pdpinch:
Since this is blocked and you've picked up some other work, I'm going to move this back to ready. You (or anyone) can pick it up again if it's unblocked.

rhysyngsun:
@pdpinch I've updated the description based on our whiteboarding, could you verify this is correct?

amir-qayyum-khan:
@rhysyngsun 

- So when ever user paid for a course we will authorize him for exam along with exam profile creation.
[...]

Issue #2286, opened 2017-01-09
Exams: restrict learners to 2 exam attempts


pdpinch:
A learner would not be allowed to schedule a 3rd exam (create another `ExamAuthorization` record) if they already had two authorizations with `exam_taken` set to true. That is it to say, the learner is allowed two authorization for each course payment.

rhysyngsun:
Clarification needed: is this a maximum of 2 retakes or a maximum of 2 attempts (i.e. 1 initial attempt + 1 retake).

rhysyngsun:
This would require handling the exam registration results so that we can tell when an authorization is redeemed for a scheduled exam.

pdpinch:
Does it require processing the registration results? Couldn't we just key it off of grades? (assuming that we deal with #3084 appropriately).

rhysyngsun:
@pdpinch we could key it off of grade processing, there's already an `exam_taken` flag on the auth record we could use.

pdpinch:
I updated the total count in the description based on an e-mail from DEDP. 

I also noted that for each course payment, two attempts are allowed.

rhysyngsun:
@pdpinch it's sufficient to only support FA courses for the number of paid courses, correct?

pdpinch:
> @pdpinch <https://github.com/pdpinch> it's sufficient to only support FA courses for the number of paid courses, correct?
> 

Yes, that's fine.

Issue #2279, opened 2017-01-09
Coupons: support Pearson exams pilot


pdpinch:
Generate and apply 100% off coupons for learners / exams in pilot

To support the Pearson pilot, we would like to generate 100% off coupons for anyone who has passed one of the DEDP courses in the pilot. We will apply the coupons for them, and automate the orders, so their dashboard will display as "paid" and they can schedule the exam when it is ready. 

Eligible learners will be notified of this opportunity by the course team (ideally, using the learner search)

### acceptance criteria:

- [ ] generate list of eligible learners
- [ ] generate coupons for eligible leaners for the appropriate course
- [ ] apply the coupon to the user
- [ ] apply the coupon to the course

noisecapella:
Can a user apply two or more coupons to the same course run?

noisecapella:
Is there anything I need to do for this or is this all administrative work?

pdpinch:
How might we handle people who haven't signed up the portal yet, but might do so in order to participate in the pilot?

PR #2331, opened 2017-01-12
Triggered exam authorizations when users enrolled or passed course

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2077

#### how it work?
- When user pay and enroll in course we create ExamProfile and ExamAuthorization
- When user pass the course in grade section we authorize him for exam
- A command to force authorize users who passed course.

@pdpinch @rhysyngsun

### DIFF
--- a/exams/util.py
+++ b/exams/util.py
@@ -1,6 +1,7 @@
 """Exam related helpers"""
 import logging
 import datetime
+import pytz
 
 from dashboard.api_edx_cache import CachedEdxUserData
 from dashboard.utils import MMTrack
@@ -33,7 +34,7 @@ def get_mmtrack(user, program):
     )
 
 
-def exam_authorization(mmtrack, course_run):
+def authorize_for_exam(mmtrack, course_run):
     """
     Authorize user for exam if he has paid for course and passed course.
 
@@ -45,22 +46,25 @@ def exam_authorization(mmtrack, course_run):
         course_run and
         course_run.edx_course_key and
         course_run.course and
+        mmtrack.program.exam_series_code and
+        course_run.course.exam_module and
         mmtrack.has_passed_course(course_run.edx_course_key) and
         mmtrack.has_paid(course_run.edx_course_key) and
-        not ExamAuthorization.objects.filter(user=mmtrack.user, course=course_run.course).exists()
+        not ExamAuthorization.objects.filter(
+            user=mmtrack.user,
+            course=course_run.course,
+            date_first_eligible__lte=datetime.datetime.now(tz=pytz.UTC),
+            date_last_eligible__gte=datetime.datetime.now(tz=pytz.UTC)
+        ).exists()
     )
 
     if ok_for_exam:
-        exam_profile, created = ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)
-
-        if created or exam_profile.status != ExamProfile.PROFILE_SUCCESS:
-            ExamProfile.objects.filter(profile=mmtrack.user.profile).update(status=ExamProfile.PROFILE_SUCCESS)
-
+        ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)
         ExamAuthorization.objects.create(
             user=mmtrack.user,
             course=course_run.course,
-            date_first_eligible=datetime.datetime.now(),
-            date_last_eligible=add_year(datetime.datetime.now())
+            date_first_eligible=datetime.datetime.now(tz=pytz.UTC),
+            date_last_eligible=add_year(datetime.datetime.now(tz=pytz.UTC))
         )
         log.info(
             'user "%s" is authorize for exam the for course id "%s"',
--- a/exams/util.py
+++ b/exams/util.py
@@ -1,0 +1,69 @@
+"""Exam related helpers"""
+import logging
+import datetime
+
+from dashboard.api_edx_cache import CachedEdxUserData
+from dashboard.utils import MMTrack
+from exams.models import (
+    ExamProfile,
+    ExamAuthorization
+)
+from seed_data.utils import add_year
+
+
+log = logging.getLogger(__name__)
+
+
+def get_mmtrack(user, program):
+    """
+    Creates mmtack object for given user.
+
+    Args:
+        user (User): a Django user.
+        program (programs.models.Program): program where the user is enrolled.
+
+    Returns:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program
+    """
+    edx_user_data = CachedEdxUserData(user, program=program)
+    return MMTrack(
+        user,
+        program,
+        edx_user_data
+    )
+
+
+def exam_authorization(mmtrack, course_run):
+    """
+    Authorize user for exam if he has paid for course and passed course.
+
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        course_run (CourseRun): A CourseRun object.
+    """
+    ok_for_exam = (
+        course_run and
+        course_run.edx_course_key and
+        course_run.course and
+        mmtrack.has_passed_course(course_run.edx_course_key) and
+        mmtrack.has_paid(course_run.edx_course_key) and
+        not ExamAuthorization.objects.filter(user=mmtrack.user, course=course_run.course).exists()
+    )
+
+    if ok_for_exam:
+        exam_profile, created = ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)
+
+        if created or exam_profile.status != ExamProfile.PROFILE_SUCCESS:
+            ExamProfile.objects.filter(profile=mmtrack.user.profile).update(status=ExamProfile.PROFILE_SUCCESS)
+
+        ExamAuthorization.objects.create(
+            user=mmtrack.user,
+            course=course_run.course,
+            date_first_eligible=datetime.datetime.now(),
+            date_last_eligible=add_year(datetime.datetime.now())
+        )
+        log.info(
+            'user "%s" is authorize for exam the for course id "%s"',
+            mmtrack.user.username,
+            course_run.edx_course_key
+        )
--- a/financialaid/api_test.py
+++ b/financialaid/api_test.py
@@ -42,14 +42,17 @@ def create_program(create_tiers=True, past=False):
     Returns:
         courses.models.Program: A new program
     """
+    end_date = None
     program = ProgramFactory.create(
         financial_aid_availability=True,
         live=True
     )
     course = CourseFactory.create(program=program)
-    end_date = datetime.now(tz=pytz.UTC) + timedelta(days=100)
+
     if past:
         end_date = datetime.now(tz=pytz.UTC) - timedelta(days=100)
+    else:
+        end_date = datetime.now(tz=pytz.UTC) + timedelta(days=100)
 
     course_run = CourseRunFactory.create(
         end_date=end_date,
--- a/financialaid/api_test.py
+++ b/financialaid/api_test.py
@@ -36,7 +36,7 @@ from roles.models import Role
 from roles.roles import Staff, Instructor
 
 
-def create_program(create_tiers=True):
+def create_program(create_tiers=True, past=False):
     """
     Helper function to create a financial aid program
     Returns:
@@ -47,7 +47,12 @@ def create_program(create_tiers=True):
         live=True
     )
     course = CourseFactory.create(program=program)
+    end_date = datetime.now(tz=pytz.UTC) + timedelta(days=100)
+    if past:
+        end_date = datetime.now(tz=pytz.UTC) - timedelta(days=100)
+
     course_run = CourseRunFactory.create(
+        end_date=end_date,
         enrollment_end=datetime.now(tz=pytz.UTC) + timedelta(hours=1),
         course=course
     )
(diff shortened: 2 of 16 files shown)


seed_data/management/commands/authorization_user_exam.py
```diff
@@ -0,0 +1,62 @@
...
+                        enrollment.course_run.edx_course_key and
+                        enrollment.course_run.course and
+                        mmtrack.has_passed_course(enrollment.course_run.edx_course_key) and
+                        not ExamAuthorization.objects.filter(user=user, course=enrollment.course_run.course).exists()
+                    )
+                    if ok_for_exam:
+                        ExamAuthorization.objects.create(
+                            user=user,
+                            course=course,
+                            date_first_eligible=datetime.datetime.now() + datetime.timedelta(years=1),
```
amir-qayyum-khan: is date for date_first_eligible and date_last_eligible should be same?

rhysyngsun: `date_first_eligible` should be now, `date_last_eligible` can be left as you have it written here.

pdpinch:
Can we add a parameter to pass a username to this function, which would force an authorization whether or not the user is eligible? (I'm anticipating testing the SSO flow, and also potential user support issues)

FYI @rhysyngsun

rhysyngsun:
@pdpinch is there any special logic needed to support the beta period? Meaning, we had decided to make the authorizations valid for a year, so I'm not certain if that's the case for learners we retroactively authorize via this script. I believe after the beta, that users will have to enroll *and* pay so would that mean we generate authorizations with a shorter eligibility period for beta and then let the regular behavior kick in afterwards?

pdpinch:
We've decided to handle the exams pilot (aka beta) with coupons. See #2279 

So, as far as I know, we don't need any special logic.

rhysyngsun:
Ok, so do we not need this script then?

pdpinch:
Don't we still need to authorize learners who became eligible in the past?

rhysyngsun:
@pdpinch yes, but once they're authorized they can schedule, so I don't see where the coupons factor in.

Ferdi:
Just curious, why django command ?  And not cron or something that does not requires someone manually doing it ?

amir-qayyum-khan:
@nlevesque @pdpinch my questions are 

- when to trigger  exam authorizations. I dont this their is a point in system which triggers when user pass a course?
- 2nd for students who passed course in past do we need a django command or a celery task they will run after a specific time and see if their is a leaner who passed a course but not  authorize. then it will authorize him?

ecommerce/api.py
```diff
@@ -315,6 +321,19 @@ def enroll_user_on_success(order):
...
+        exam_profile, created = ExamProfile.objects.get_or_create(profile=order.user.profile)
+
+        if created or exam_profile.status != ExamProfile.PROFILE_SUCCESS:
+            exam_profile.update(status=ExamProfile.PROFILE_SUCCESS)
+
+        if not ExamAuthorization.objects.filter(user=order.user, course=enrollment.course_run.course).exists():
+            ExamAuthorization.objects.create(
+                user=order.user,
+                course=enrollment.course_run.course,
+                status=ExamAuthorization.STATUS_SUCCESS,
```
amir-qayyum-khan: @rhysyngsun this approach is right?

seed_data/management/commands/authorization_user_exam.py
```diff
@@ -0,0 +1,99 @@
...
+    def handle(self, *args, **options):
+        """
+        Trigger exam authorizations for user for a course if he has passed it.
+        """
+        program_id = options.get("program_id", None)
+        username = options.get("username", None)
+
+        programs = (
+            Program.objects.filter(id=program_id)
+            if program_id else Program.objects.filter(exam_series_code__isnull=False, live=True)
```
amir-qayyum-khan: Assuming that user with ``exam_series_code`` are only who can be authorized for exam.
Am i right?

rhysyngsun: @amir-qayyum-khan sorry I didn't specify this before (I thought I had), but yes this is only valid courses where `Program.exam_series_code` and `Course.exam_module` are both set.

I'm also don't think we need this command anymore since we're planning on doing this retroactively for users by marking as paid via a coupon.

The best way forward with this code is probably what you mentioned in chat yesterday which is to create a task that checks all these conditions and then creates the `ExamProfile` and `ExamAuthorization` if they don't already exist and to call that task when they enroll and when we calculate the final grade.

Does that make sense? I want to be sure you're not wasting your time in the wrong direction so if it's easier to discuss in chat please ping me.

amir-qayyum-khan: this command can be use for https://github.com/mitodl/micromasters/pull/2331#issuecomment-272174328

amir-qayyum-khan: we dont need task, task is something run in queue (background). I think we need centralize code. i made a new class for that purpose ``exam/util.py``

rhysyngsun: That's fine, I agree on the centralized code, although I think this command itself should be in the `exams` app rather than `seed_data` since it doesn't have anything to do with `seed_data`.

amir-qayyum-khan: signal is good idea https://github.com/mitodl/micromasters/issues/2077#issuecomment-272468053

exams/util.py
```diff
@@ -0,0 +1,69 @@
...
+        mmtrack.has_passed_course(course_run.edx_course_key) and
+        mmtrack.has_paid(course_run.edx_course_key) and
+        not ExamAuthorization.objects.filter(user=mmtrack.user, course=course_run.course).exists()
+    )
+
+    if ok_for_exam:
+        exam_profile, created = ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)
+
+        if created or exam_profile.status != ExamProfile.PROFILE_SUCCESS:
+            ExamProfile.objects.filter(profile=mmtrack.user.profile).update(status=ExamProfile.PROFILE_SUCCESS)
```
rhysyngsun: Don't set `status=ExamProfile.PROFILE_SUCCESS`, it should only be in a PENDING state here if it was just created and that's the default value so I think you can just drop these two lines.

exams/util.py
```diff
@@ -0,0 +1,69 @@
...
+
+        ExamAuthorization.objects.create(
+            user=mmtrack.user,
+            course=course_run.course,
+            date_first_eligible=datetime.datetime.now(),
```
rhysyngsun: Needs a TZ.

exams/util.py
```diff
@@ -0,0 +1,69 @@
...
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        course_run (CourseRun): A CourseRun object.
+    """
+    ok_for_exam = (
+        course_run and
+        course_run.edx_course_key and
+        course_run.course and
+        mmtrack.has_passed_course(course_run.edx_course_key) and
+        mmtrack.has_paid(course_run.edx_course_key) and
+        not ExamAuthorization.objects.filter(user=mmtrack.user, course=course_run.course).exists()
```
rhysyngsun: I'm guessing you're trying to ensure that we don't create multiple authorizations for the same course run, which is something I didn't think about. 

It is possible to have multiple authorizations. For example, DEDP is going to allow up to 2 retakes (issue for this is #2286). @pdpinch is it possible that both retakes would occur on the same course run (I'm suspecting yes and that that will be the primary use case for retakes)?

In order to avoid the duplicates we might need to modify this to check for an authorization where `now` is between `date_first_eligible` and `date_last_eligible`.

exams/util.py
```diff
@@ -0,0 +1,69 @@
...
+    Authorize user for exam if he has paid for course and passed course.
+
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        course_run (CourseRun): A CourseRun object.
+    """
+    ok_for_exam = (
+        course_run and
+        course_run.edx_course_key and
+        course_run.course and
```
rhysyngsun: Additionally, both `program.exam_series_code` and `course.exam_module` need to be nonempty values.

exams/util.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    edx_user_data = CachedEdxUserData(user, program=program)
+    return MMTrack(
+        user,
+        program,
+        edx_user_data
+    )
+
+
+def exam_authorization(mmtrack, course_run):
```
rhysyngsun: I would probably rename this to something like `authorize_for_exam` or something like that. The current name is a little ambiguous.

exams/util.py
pdpinch: Yes, you're right. The most likely case for retakes is that they will be on the same course run.

amir-qayyum-khan:
@rhysyngsun done with changes and added tests

seed_data/utils_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+from seed_data.utils import add_year
+
+
+class AddYearTests(TestCase):
+    """Tests for add year method"""
+
+    def test_add_year(self):  # pylint: disable=no-self-use
+        """assert add year method works"""
+        now = datetime.datetime.now(tz=pytz.UTC)
```
rhysyngsun: I would be *very* careful with this, if there's a bug in `add_year` (like the two cases I pointed out above) this test could fail during certain times of certain years. I think you should prefer explicit testing of specific dates that do/don't include a leap day between that date and +1 year.

seed_data/management/commands/authorization_user_exam.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        for program in programs:
+            users_qset = (
```
rhysyngsun: Nit: Could also be written this way:

```python
users_qset = ProgramEnrollment.objects.filter(program=program)

if username:
    users_qset = users_qset.filter(user__username=username)
```

rhysyngsun: Same for the `Program` query above.

financialaid/api_test.py
```diff
@@ -47,7 +47,12 @@ def create_program(create_tiers=True):
         live=True
     )
     course = CourseFactory.create(program=program)
+    end_date = datetime.now(tz=pytz.UTC) + timedelta(days=100)
+    if past:
```
rhysyngsun: Nit: `if..else` here would read better

exams/util.py
```diff
@@ -0,0 +1,73 @@
...
+    ExamProfile,
+    ExamAuthorization
+)
+from seed_data.utils import add_year
+
+
+log = logging.getLogger(__name__)
+
+
+def get_mmtrack(user, program):
```
rhysyngsun: This might be better located in dashboard/urls.py

exams/util.py
```diff
@@ -0,0 +1,73 @@
...
+
+def authorize_for_exam(mmtrack, course_run):
+    """
+    Authorize user for exam if he has paid for course and passed course.
+
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        course_run (CourseRun): A CourseRun object.
+    """
+    ok_for_exam = (
```
rhysyngsun: @amir-qayyum-khan sorry I didn't notice this before. I think this logic needs to be different based on the requirement to create the `ExamProfile` when the users first pays (I still think my comment just above this about refactoring is valuable here as well):

```python
is_exam_course = (
    course_run and 
    course_run.edx_course_key and 
    course_run.course and 
    mmtrack.program.exam_series_code and
    course_run.course.exam_module
)

if is_exam_course:
    if mmtrack.has_paid(course_run.edx_course_key):
        ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)
    if (
        mmtrack.has_passed_course(course_run.edx_course_key) and
        not ExamAuthorization.objects.filter(
             user=mmtrack.user,
             course=course_run.course,
             date_first_eligible__lte=datetime.datetime.now(tz=pytz.UTC),
             date_last_eligible__gte=datetime.datetime.now(tz=pytz.UTC)
         ).exists()
    ):
        ExamAuthorization.objects.create(
             user=mmtrack.user,
             course=course_run.course,
             date_first_eligible=datetime.datetime.now(tz=pytz.UTC),
             date_last_eligible=add_year(datetime.datetime.now(tz=pytz.UTC))
         )
```

seed_data/utils_test.py
```diff
@@ -0,0 +1,25 @@
+"""Tests for utils file"""
+import datetime
+import pytz
+from django.test import TestCase
+
+from seed_data.utils import add_year
+
+
+class AddYearTests(TestCase):
```
rhysyngsun: Can be `unittest.TestCase` instead since you don't need the DB.

amir-qayyum-khan:
Done @rhysyngsun

seed_data/utils_test.py
rhysyngsun: @amir-qayyum-khan not sure if you missed this, but this test will definitely fail down the road. Particularly on the next leap day `day = 29`, but there won't be a 29th on the next year. I think you need to remove this test and assert specific behaviors around specific dates in and around the leap day. Particularly:

- Feb 28, 2015
- Feb 28, 2016
- Feb 29, 2016

amir-qayyum-khan:
Done @rhysyngsun

PR #2345, opened 2017-01-12
Fixed infinite loop in profile validation (#2344)

### LINKED ISSUE #2344: Application hangs if you previously completed profile but now are considered incomplete
This is needed for #2240 and #2308 which add several new fields. This is a race condition present in App.js where react gets stuck in an infinite loop trying to redirect the user to /profile/personal to complete the missing fields and the application goes unresponsive.

rhysyngsun:

Fixes #2344 

#### What's this PR do?

This fixes an infinite cycle in the profile validation in `App.js`. It doesn't happen in production currently, but when we add more profile fields users would see it.

`App.js`

Register as a new user, go to the profile page and be sure you can edit your profile.

### DIFF
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -74,7 +74,6 @@ class App extends React.Component {
       this.fetchUserProfile(SETTINGS.user.username);
     }
     this.fetchEnrollments();
-    this.requireCompleteProfile();
   }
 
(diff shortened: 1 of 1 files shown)


aliceriot:
functionality is nice and infinite-loop free!

PR #2352, opened 2017-01-13
fix program selection

### LINKED ISSUE #2347: Enrolled program not selected if redirected to /profile/personal from /dashboard
This redirect happens when we add new required profile fields and the user tries to go to /dashboard/.

The program option does not have anything selected, but it should be selecting the program you're already enrolled in so you don't have to select it again.

**NOTE: if you go directly to /profile/personal** you will not see this issue**

aliceriot:

closes #2347

#### What's this PR do?

This PR fixes a small UI bug with the program selector that we show on `/profile/personal`. 

Basically, if a user returns to `/profile/personal` to change something, and they have already enrolled in a program, we don't want the program selector to be empty. If the program selector is empty, when the user presses 'save' we will show a validation error - if the user is already enrolled in a program this is inappropriate.

I made a change in https://github.com/mitodl/micromasters/commit/72504a58efe6d679e3e3f69f7d76e5a67a83e2f4 that was supposed to address this. My earlier fix made sure that, after we fetch the program enrollments, we check if the user is on `/profile` and, if so, we dispatch the action to pre-set the user's first enrolled program, which would prevent the validation error from showing up.

There is a bug in this fix, though. In the case that the user had a validation error which would cause them to be redirected from `/dashboard` -> `/profile`, the request to fetch enrollment information could come back while the user was still on the dashboard. Then the `pathname` check would fail, and so the user would still see the validation error when they tried to save (yucky). This wasn't an issue if the user visited `/profile/personal` right away, because then the profile regex pathname check would pass.

Anyway, this logic shouldn't really be in `<App />` anyway, so I've moved it into the relevant component (`<PersonalTab />`). This should ensure that any time we render the personal tab we perform the same actions (to dispatch the `setProgram` action and ensure the user doesn't see a needless validation error).

Blank out a field on a user profile which already has program enrollments in the django shell. Then:

1. Visit `/dashboard` and confirm you get redirected to `/profile/personal` and, when you arrive, a program enrollment is already selected.

2. Visit `/profile/personal` directly, and confirm that a program enrollment is already selected.

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -68,7 +68,15 @@ export default class PersonalTab extends React.Component {
     dispatch(setProfileStep(PERSONAL_STEP));
   }
 
-  onChange = (selection: Option): void => {
+  componentDidUpdate() {
+    const { currentProgramEnrollment, setProgram } = this.props;
+    const selectedProgram = this.getSelectedProgramId();
+    if ( currentProgramEnrollment && !selectedProgram ) {
+      setProgram(currentProgramEnrollment);
+    }
+  }
+
+  onProgramSelect = (selection: Option): void => {
     const {
       programs,
       setProgram,
@@ -91,7 +99,7 @@ export default class PersonalTab extends React.Component {
     return (
       <Select
         value={this.getSelectedProgramId()}
-        onChange={this.onChange}
+        onChange={this.onProgramSelect}
         clearable={false}
         className={`program-selectfield ${validationErrorSelector(errors, ['program'])}`}
         errorText={_.get(errors, "program")}
--- a/static/js/containers/ProfilePage_test.js
+++ b/static/js/containers/ProfilePage_test.js
@@ -26,6 +26,7 @@ import {
   setProgram,
   SET_PROFILE_STEP,
   SET_TOAST_MESSAGE,
+  SET_PROGRAM,
 } from '../actions/ui';
 import { USER_PROFILE_RESPONSE, PROGRAMS } from '../test_constants';
 import {
@@ -58,6 +59,8 @@ describe("ProfilePage", function() {
     SET_PROFILE_STEP,
   ];
 
+  const SUCCESS_SET_PROGRAM_ACTIONS = SUCCESS_ACTIONS.concat(SET_PROGRAM);
+
   const REDIRECT_ACTIONS = SUCCESS_ACTIONS.concat([
     SET_PROFILE_STEP,
   ]);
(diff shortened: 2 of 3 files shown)


static/js/components/PersonalTab.js
```diff
@@ -68,7 +68,15 @@ export default class PersonalTab extends React.Component {
     dispatch(setProfileStep(PERSONAL_STEP));
   }
 
-  onChange = (selection: Option): void => {
+  componentDidUpdate() {
+    const { currentProgramEnrollment } = this.props;
+    const selectedProgram = this.getSelectedProgramId();
+    if ( currentProgramEnrollment && !selectedProgram ) {
+      this.onProgramSelect({ value: currentProgramEnrollment.id, label: '' });
```
gsidebo: this smells a bit to me. we're directly calling a function that is intended to handle the `onChange` event for a select box? shouldn't we just call `setProgram` directly?

static/js/containers/ProfilePage_test.js
```diff
@@ -189,19 +177,20 @@ describe("ProfilePage", function() {
           work_history: [],
         };
         helper.profileGetStub.withArgs(SETTINGS.user.username).returns(Promise.resolve(updatedProfile));
-        return renderComponent(`/profile/${step}`, SUCCESS_ACTIONS).then(([, div]) => {
-          // close all switches
-          if (step === 'personal') {
-            return confirmSaveButtonBehavior(updatedProfile, {div: div}, true);
-          }
-          return confirmSaveButtonBehavior(updatedProfile, {div: div});
+        let actions = step === PERSONAL_STEP ? SUCCESS_ACTIONS.concat(SET_PROGRAM) : SUCCESS_ACTIONS;
```
gsidebo: `SUCCESS_ACTIONS.concat(SET_PROGRAM)` is used 4 times. can we just define a new constant for it? `SUCCESS_AND_SET_PROGRAM_ACTIONS` or something? that name sucks. if you think of a better one, go for it. otherwise you can ignore this

static/js/containers/ProfilePage_test.js
```diff
@@ -256,7 +245,8 @@ describe("ProfilePage", function() {
     [EMPLOYMENT_STEP, 'EmploymentTab'],
   ]) {
     it(`sends the right props to tab components for step ${step}`, () => {
-      return renderComponent(`/profile/${step}`, SUCCESS_ACTIONS).then(([wrapper]) => {
+      let actions = step === PERSONAL_STEP ? SUCCESS_ACTIONS.concat(SET_PROGRAM) : SUCCESS_ACTIONS;
```
gsidebo: general comment: it's a bummer that you need to update all of these tests that have nothing to do with your functionality. this has happened to me a ton, and probably everyone else that has changed anything action-related in this codebase. it would be really nice to be able to listen for a _subset_ of actions (ie: "i care that actions A, B, and C were executed, and i'm not really concerned about the other ones") rather than always requiring an exact list of actions all the time

static/js/containers/ProfilePage_test.js
```diff
@@ -189,19 +177,20 @@ describe("ProfilePage", function() {
...
-          if (step === 'personal') {
-            return confirmSaveButtonBehavior(updatedProfile, {div: div}, true);
-          }
-          return confirmSaveButtonBehavior(updatedProfile, {div: div});
+        let actions = step === PERSONAL_STEP ? SUCCESS_ACTIONS.concat(SET_PROGRAM) : SUCCESS_ACTIONS;
+        return renderComponent(`/profile/${step}`, actions).then(([, div]) => {
+          return confirmSaveButtonBehavior(
+            updatedProfile,
+            {div: div},
+            step === PERSONAL_STEP ? [REQUEST_ADD_PROGRAM_ENROLLMENT] : []
```
gsidebo: it seems a little odd that we're _adding_ a program enrollment for users that already have that enrollment but have some profile info missing. i can't think of a way that would bite us but it's worth bringing up

static/js/components/PersonalTab.js
aliceriot: The onProgramSelect is passed in as an onChange handler but it doesn't receive an event as an argument, just an Option object. I wanted to reuse the code that's in onProgramSelect here, but if you think we shouldn't re-use that function I'm fine with changing it.

static/js/containers/ProfilePage_test.js
aliceriot: sure

static/js/containers/ProfilePage_test.js
aliceriot: AFAIK the program enrollment API is idempotent, I think that was built that way for this reason, so it should be ok? haha, famous last words.

static/js/containers/ProfilePage_test.js
aliceriot: Yeah, I agree that would make a *lot* of these tests easier. We wrote the redux-asserts library, so we could definitely add that functionality if we wanted!

static/js/components/PersonalTab.js
gsidebo: yeah, it happens to do what we want, but it smells. the fact that we're sort of cheating and creating a fake `Option` with an empty `label` value is one indicator of that. at this point in execution, all we want to do is make sure we have a selected program set in the state. the concept of an Option with a label and a value is irrelevant. i think we should split `onProgramSelect` like so:

```javascript
setProgramById= (programId: number?): void => {
    const {
      programs,
      setProgram,
    } = this.props;
    if ( programId ) {
      let selected = programs.find(program => program.id === parseInt(programId));
      setProgram(selected);
    } else {
      setProgram(null);
    }
}

onProgramSelect = (selection: Option): void => {
    let programId = selection && selection.value ? parseInt(selection.value) : null;
    this.setProgramById(programId);
}
```

`setProgramById` might be relevant to other components too. i feel like it would be better if we just called the action with a simple program id in the argument (instead of finding the right object in the `programs` list) but that's a bit more work, perhaps for a different PR

aliceriot: Oh, sorry, I went with the original suggestion, the `componentDidUpdate` function just pulls `setProgram` out of `this.props` and calls it with `currentProgramEnrollment` as an argument.

aliceriot:
comments addressed

PR #2362, opened 2017-01-17
Replace hardcoded Adwords Conversion ID with variable reference

singingwolfboy:
Follow-up from #2263. We should use variable references, not hardcode IDs.


amir-qayyum-khan:
👍

amir-qayyum-khan:
please merge it

PR #2364, opened 2017-01-17
Properly rename migration file based on what it does

singingwolfboy:
Is this a good idea, or do we want to leave the migration files untouched after they merge? @aliceriot @rhysyngsun @blarghmatey


singingwolfboy:
This was introduced in #2240.

rhysyngsun:
I think this only would've got applied to CI at this point, but we could roll the old one back.

aliceriot:
what's the benefit? just a little clarity?

rhysyngsun:
This SO answer has a method, however since deploys automatically run the migration I don't think it would work for us: http://stackoverflow.com/a/30427833/140999

singingwolfboy:
@aliceriot Yes. Particularly for future migrations that declare a dependency on this one.

singingwolfboy:
Sounds like we don't actually want to do this.

PR #2366, opened 2017-01-17
Fix toast message loop

### LINKED ISSUE #2359: Infinite loop on order receipt page after purchasing a course
When working on a coupon issue I ran into this bug where the browser hung due to infinite stack. From the stack trace it looks like `CourseDescription` is involved and `react-router` too, but I haven't done much more investigating yet. 

Note:
 - this shouldn't affect order fulfillment, just the order receipt page showing correctly
 - I haven't been able to reproduce this on micromasters-rc, so it might not be an issue in production

screenshot from 2017-01-13 17-37-27

noisecapella:
Fixes #2359 

#### What's this PR do?
Fixes an infinite loop regarding toast messages. If a toast message is already present other messages will be ignored.

Go to the URL with `/dashboard/?status=receipt&course_key=course_key`, where `course_key` is the edX course key, encoded as a query parameter. The course key can refer to any run which is present in the dashboard API. You should see an order success toast message which goes away after 5 seconds. Also go to `/dashboard/?status=cancel` to verify that the cancellation message works too. If you go to these URLs when you're checked out on master you should get the toast loop which causes the browser to become unresponsive.

### DIFF
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -100,31 +100,35 @@ class DashboardPage extends React.Component {
   }
 
   handleOrderSuccess = (course: Course): void => {
-    const { dispatch } = this.props;
+    const { dispatch, ui: { toastMessage } } = this.props;
     let firstRun: CourseRun = course.runs.length > 0 ? course.runs[0] : null;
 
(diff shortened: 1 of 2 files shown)


aliceriot:
Can repro on master

aliceriot:
and can confirm the fix fixes it! nice.

PR #2371, opened 2017-01-17
Added test for auto migrations (#2365)

### LINKED ISSUE #2365: Test for autogenerated migration names
We would like our migrations to have meaningful names going forward, this should be easy to enforce via a  test script so we don't have to manually be vigilant on catching these.

rhysyngsun:

Fixes #2365 

#### What's this PR do?

This PR adds a test for migrations that didn't have a `--name` specified when they're created.

Also refactored detect_missing_migrations.sh to the script directory I created here and ensured ./test_suite.sh runs flow while I was in there.

scripts/test/no_auto_migrations.sh

Run the script above, see that it prints nothing and exits `0`. Add a dummy auto migration file like `touch profiles/migrations/00300_auto_dummy_test_file.py` and run the script again. It fail print a message indicating the offending filename and exit nonzero.


rhysyngsun:
@noisecapella maybe you could review since this touches the tox config?

PR #2377, opened 2017-01-18
scroll to error when profile page mounts (handle the redirect properly)

### LINKED ISSUE #2376: page should scroll to error when the user is redirected due to a missing field
thought that this was implemented, but it's not!

aliceriot:

closes #2376

#### What's this PR do?

Makes sure that we scroll to any existing validation errors when the `ProfilePage` component mounts.

Go into the django shell and delete a field on your profile. Then visit `/dashboard`. You should be redirected to the profile page, and the page should scroll to ensure you can see the error.


noisecapella:
Sorry for the delay, got caught up in something else :+1:

PR #2378, opened 2017-01-18
Added logic to calculate prices including coupons on frontend

### LINKED ISSUE #2356: Coupons: calculate prices in frontend
We need to combine the price in `/api/v0/course_prices/` with the discount information in `/api/v0/coupons/` in a front end utility function, and use it when displaying prices for each course.

noisecapella:
Fixes #2356 

#### What's this PR do?
Adds functions to calculate prices based on coupons (`static/js/util/coupon.js`).

This doesn't connect with anything in the UI yet


static/js/flow/programTypes.js
```diff
@@ -40,13 +65,7 @@ export type CourseRun = {
   course_end_date?: string,
   course_upgrade_deadline?: string,
   price?: number,
-};
-
-export type FinancialAid = {
```
aliceriot: was this an extraneous type?

static/js/test_constants.js
```diff
@@ -1,6 +1,13 @@
 /* global SETTINGS: false */
+// @flow
```
aliceriot: 💯 guess I left this off when I created this file...oops!

static/js/util/coupon.js
```diff
@@ -0,0 +1,85 @@
+// @flow
```
aliceriot: do you think it would make sense for this to live in `lib` instead of `util`?

static/js/util/coupon.js
```diff
@@ -0,0 +1,85 @@
...
+
+export const calculatePrices = (programs: Dashboard, prices: CoursePrices, coupons: Coupons): CalculatedPrices => {
+  let couponLookup: Map<number, Coupon> = new Map();
+  let priceLookup: Map<number, CoursePrice> = new Map();
+  for (const coupon of coupons) {
+    couponLookup.set(coupon.program_id, coupon);
+  }
+  for (const price of prices) {
+    priceLookup.set(price.program_id, price);
+  }
```
aliceriot: it might be nice to pull these two loops out into some helper functions, something like:

```js
const populateMap = R.curry((getKey, objArray) => {
  let map = new Map();
  objArray.forEach(obj => map.set(getKey(obj), obj));
});

const makeCouponMap: (c: Coupons) => Map<number, Coupon> = populateMap(
  R.prop('program_id'))
);

const makePriceMap: (c: Coupons) => Map<number, Coupon> = populateMap(
  R.prop('program_id'))
);
```

hum just realized both of those are using the `program_id` prop, but you'd probably still want to annotate the return types separately, maybe you could just do:

```js
const makeProgramIDMap = populateMap(R.prop('program_id'));

let couponLookup: Map<number, Coupon> = makeProgramIDMap(coupons);
let priceLookup: Map<number, Coupon> = makeProgramIDMap(prices);
```

just a thought!

aliceriot: Could also make use potentially of `new Map([ [id, obj], ... ])` as well - that might be more elegant than what I suggested.

static/js/util/coupon.js
```diff
@@ -0,0 +1,85 @@
...
+    courses: program.courses.map(course => ({
+      id: course.id,
+      runs: course.runs.map(run => ({
+        id: run.id,
+        price: mockableCalculateRunPrice(
+          run.id, course.id, program.id, priceLookup.get(program.id), couponLookup.get(program.id)
+        ),
+      }))
+    }))
+  }));
```
aliceriot: this is also something that strikes me as good to make a little more 'flat', three nested `.map`s is kinda hard to read. you could separate it out into three separate functions, an inner one that deals with `runs`, another that deals with `courses`, etc.

static/js/util/coupon.js
aliceriot: For instance, something like:

```js
const populateProgramIdMap = R.compose(
  entries => new Map(entries), R.map(obj => [obj.program_id, obj])
);
```

would do it I think!

static/js/flow/programTypes.js
noisecapella: Yes, it wasn't being used and there's also `FinancialAidUserInfo`

static/js/util/coupon.js
noisecapella: Yeah, that makes more sense

noisecapella:
How's that?

static/js/lib/coupon.js
```diff
@@ -0,0 +1,91 @@
...
+
+  let calcPriceForCourse = (course, program) => ({
+    id: course.id,
+    runs: course.runs.map(run => calcPriceForRun(run, course, program))
+  });
+
+  let calcPriceForProgram = program => ({
+    id: program.id,
+    courses: program.courses.map(course => calcPriceForCourse(course, program))
+  });
```
aliceriot: I think for writing these the ramda functions `over` and `lensProp` might be really helpful, `over` basically lets you 'edit' an object by supplying an update func and a lens that focuses on the property that you want to update. The return value will be a clone of the object, with the value of the focused prop set to the value of the update function called with that prop's original value as an argument.

You could do something like:

```js
let calcPriceforProgram = program => (
  R.over(R.lensProp('courses'), R.map(calcPriceForCourse(program)), program)
);
```

(assuming the argument order for `calcPriceForCourse` could be switched and updated to be a curried function). You could use a similar pattern for the `calcPriceForCourse` and `calcPriceForRun` function as well.

static/js/lib/coupon_test.js
```diff
@@ -0,0 +1,151 @@
...
+import * as couponFuncs from './coupon';
+import {
+  makeCoupon,
+  makeCoupons,
+  makeCoursePrice,
+  makeCoursePrices,
+  makeDashboard,
+  makeProgram,
+} from '../factories';
+
```
aliceriot: extra blank line

static/js/factories/index.js
```diff
@@ -0,0 +1,105 @@
+// @flow
```
aliceriot: what do you think about calling this `factories/dashboard.js` or similar? Just thinking we might eventually want to have separate modules for profiles and so on.

static/js/lib/coupon.js
```diff
@@ -0,0 +1,91 @@
...
+    // coupon doesn't match
+    return startingPrice;
+  }
+};
+
+// import to allow mocking in tests
+import {
+  calculateRunPrice as mockableCalculateRunPrice,
+  calculateDiscount as mockableCalculateDiscount
+} from './coupon';
```
aliceriot: why do we have to do this in order to mock? I know we do for the API and stuff too, just wondering why it's necessary.

static/js/factories/index.js
noisecapella: Sure

static/js/factories/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+const makeCounter = (): (() => number) => {
+  let gen = (function*() {
+    let i = 1;
+    while (true) {  // eslint-disable-line no-constant-condition
+      yield i;
+      i += 1;
+    }
+  })();
+  // $FlowFixMe: Flow doesn't know that this always returns a number
```
noisecapella: Actually I don't think Flow can ever know about this. `next()` returns an object like `{value: 'value', done: false}`, but if the iterator ends then it returns `{done: true}` with no `value`, so it may not be a number. Flow's not smart enough to solve the halting problem :)

aliceriot: haha, not gonna worry about it then!

static/js/lib/coupon.js
noisecapella: Here's an explanation: http://stackoverflow.com/questions/35753797/stub-module-function-called-from-the-same-module

Basically in Python when a function is defined in the module level it's put into a dictionary of names in the Python module, and every reference to the name is a lookup from that dictionary. This lets the mock work by just swapping out the value for the name in the mock. The functions defined in ES6 are just treated as local variables, even if they are exported, so they can't be swapped out by sinon. The lines here explicitly say, use the function value we exported, rather than the local variable

aliceriot: Gotcha, so it's a variable scope problem - there's a closure that the module keeps with it which sinon can't normally reach inside. Cool!

noisecapella: Yeah, I'll investigate if there's a cleaner way to do this so we don't have the ugly mocked name. It looks like we can possibly make use of `exports` and reassign the variable that way

aliceriot: hm, or we could use `_function` for the local variable and `function` for the exported one - I don't have too much of an issue with `mockableFooFunction` anyhow.

noisecapella: Yeah, I'll go with the `_function`. Using `exports` directly isn't working and ES6 complains about having local variables and `import` using the same name

static/js/lib/coupon.js
noisecapella: I'm not sure that works here since at each level I only want a dict with the `id`, `price`, and the list of children, instead of cloning the existing `program`

static/js/lib/coupon.js
```diff
@@ -0,0 +1,91 @@
...
+  Dashboard,
+} from '../flow/dashboardTypes';
+
+export const calculatePrices = (programs: Dashboard, prices: CoursePrices, coupons: Coupons): CalculatedPrices => {
+  type HasProgramId = {
+    program_id: number
+  };
+  function makeLookup<T: HasProgramId>(obj: Array<T>): Map<number, T> {
+    return new Map(obj.map((value: T) => [value.program_id, value]));
+  }
```
noisecapella: It's hard to do for the other three functions since they reference the lookup via closure. `calcPriceForRun` would need to take 5 arguments if it were outside the function

static/js/lib/coupon.js
aliceriot: Oh gotcha, we don't want the other properties to tag along? Well, a simple way around that would be using `R.pick`, but if you wanna stick with what's here that's fine too.

static/js/lib/coupon.js
aliceriot: fair enough

static/js/lib/coupon.js
noisecapella: Ok, I'll leave it the way it is just in the interest of less work for me :)

aliceriot: haha, sounds good

static/js/lib/api.js
```diff
@@ -129,34 +133,33 @@ export function fetchJSONWithCSRF(input: string, init: Object = {}, loginOnError
     respJson.errorStatusCode = statusCode;
     return Promise.reject(respJson);
   });
-}
+};
+
+// allow mocking in tests
+export { _fetchJSONWithCSRF as fetchJSONWithCSRF };
+import { fetchJSONWithCSRF } from './api';
```
aliceriot: oh nice, this is cleaner!

static/js/components/dashboard/CourseSubRow.js
```diff
@@ -1,4 +1,3 @@
-// @flow
```
aliceriot: is this still being type-checked? missed this earlier

aliceriot: (goes for other files too)

noisecapella: No, I ran into a bunch of flow errors when doing the other flow related changes. I'll see what files I can undo this change on

aliceriot: gotcha

noisecapella:
I removed some of the `@flow` removals and fixed some errors

aliceriot:
awesome, I'm fine with leaving `// @flow` out of test files (it sometimes just gets in the way) but not actual code.

PR #2386, opened 2017-01-19
Mocked indexing functions in most tests

### LINKED ISSUE #2254: Fix MockedESTestCase
Since we don't use Elasticsearch in most of our Python tests we should mock Elasticsearch somehow to prevent communication with it. We need to do it in a way that doesn't interfere with the functionality of the app, and a test speed increase would be nice too

giocalitri:
closes #2254 
#### What's this PR do?
This PR introduces a new version of `ESTestCase` where the indexing functions are mocked and renames the old `ESTestCase` to `ESTestCaseWithIndex`. 
`ESTestCaseWithIndex` now drops the index after every single Test Class, to avoid the side effect of having some test classes inheriting from `django.test.TestCase` and still working because the index was created by a previous test.
The only tests that use `ESTestCaseWithIndex` are the ones in the `search` app.

`search.base`

Start your docker environment with a `docker-compose up` so you can monitor the logs.
In another shell you should run first 
`docker-compose run web tox -- -k 'not search'`
that will run all the tests but the ones in the `search` directory. You should not see any activity in the elasticsearch logs.

Then run `docker-compose run web tox -- search/*` to run the `search` tests and you should see activity in the elasticsearch logs.

Then run `docker-compose run web tox ` to run all the tests to see that everything works.

Finally, after running the tests before, you should modify a test class that uses `ESTestCase` to use `django.test.TestCase`. The tests should fail with an error related to the lack of the index.

### DIFF
--- a/search/base.py
+++ b/search/base.py
@@ -4,23 +4,23 @@ Base test classes for search
 from unittest.mock import patch
 
 from django.test import (
-    override_settings,
     TestCase,
 )
 
-from search.indexing_api import recreate_index
+from search import indexing_api
+from search.indexing_api import recreate_index, delete_index
 
 
 class ESTestCase(TestCase):
     """
-    Set up ES index on setup
+    Test class for test cases that need a live ES index
     """
 
     @classmethod
-    def setUpTestData(cls):
+    def setUpClass(cls):
         # Make sure index exists when signals are run.
         recreate_index()
-        super().setUpTestData()
+        super().setUpClass()
 
     def setUp(self):
         # Make sure index exists when signals are run.
@@ -31,31 +31,34 @@ class ESTestCase(TestCase):
         recreate_index()
         super().setUp()
 
+    @classmethod
+    def tearDownClass(cls):
+        delete_index()
+        super().tearDownClass()
 
-@override_settings(ELASTICSEARCH_URL="fake")
+
 class MockedESTestCase(TestCase):
     """
-    Mock ES signals
+    Test class that mocks the MicroMasters indexing API to avoid unnecessary ES index operations
     """
-
-    def setUp(self):
-        self.patchers = [
-            patch('search.api.get_conn', autospec=True),
-            patch('search.api.bulk', autospec=True, return_value=(0, [])),
-            patch('search.api.Mapping', autospec=True),
-        ]
-        for patcher in self.patchers:
+    @classmethod
+    def setUpClass(cls):
+        cls.patchers = []
+        for name, val in indexing_api.__dict__.items():
+            if callable(val):
+                cls.patchers.append(patch('search.indexing_api.{0}'.format(name), autospec=True))
+        for patcher in cls.patchers:
             patcher.start()
         try:
-            super(MockedESTestCase, self).setUp()
+            super().setUpClass()
         except:
-            for patcher in self.patchers:
+            for patcher in cls.patchers:
                 patcher.stop()
             raise
 
-    def tearDown(self):
-        try:
-            super(MockedESTestCase, self).tearDown()
-        finally:
-            for patcher in self.patchers:
-                patcher.stop()
+    @classmethod
+    def tearDownClass(cls):
+        for patcher in cls.patchers:
+            patcher.stop()
+
+        super().tearDownClass()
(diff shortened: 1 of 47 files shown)


search/base.py
```diff
@@ -4,23 +4,23 @@
...
     TestCase,
 )
 
-from search.indexing_api import recreate_index
+from search import indexing_api
+from search.indexing_api import recreate_index, delete_index
 
 
-class ESTestCase(TestCase):
+class ESTestCaseWithIndex(TestCase):
```
gsidebo: naming nitpick: if i see a class that inherits from `ESTestCase`, that suggests to me that it's a test case that uses/tests elasticsearch. to me it makes more sense to rename `ESTestCase` to `MockedESTestCase`, and `ESTestCaseWithIndex` to `ESTestCase`. most/all of the test classes that inherit from the mocked ES class don't care if an ES index exists at all. we only need to mock it to handle calls to our indexing API from signals. in that sense i think it's confusing for a test class that doesn't test any kind of ES interaction to inherit from a class called `ESTestCase`. i understand that rename would touch a ton of files. if you like the idea and pycharm's rename function gets this done with minimal fuss, go for it. i'm happy to review the much-larger PR. otherwise you can ignore

search/indexing_api.py
```diff
@@ -33,6 +33,24 @@
...
+    "remove_user",
+    "serialize_program_enrolled_user",
+    "refresh_index",
+    "program_enrolled_user_mapping",
+    "create_program_enrolled_user_mapping",
+    "create_mappings",
+    "clear_index",
+    "delete_index",
+    "recreate_index",
+]
```
gsidebo: i did a quick google search for iterating through all of a modules functions and saw [this SO answer](http://stackoverflow.com/questions/21885814/how-to-iterate-through-a-modules-functions#21885920). i think that might get us what we want without needing to manually specify `__all__` as we're doing here

search/indexing_api.py
```diff
@@ -304,6 +322,16 @@ def clear_index():
     create_mappings()
 
 
+def delete_index():
```
gsidebo: can you remove the repeated code in `clear_index` and call this function instead?

search/base.py
```diff
@@ -31,31 +31,33 @@ def setUp(self):
...
+        delete_index()
+        super().tearDownClass()
+
 
-@override_settings(ELASTICSEARCH_URL="fake")
-class MockedESTestCase(TestCase):
+class ESTestCase(TestCase):
     """
-    Mock ES signals
+    Mock user indexing
```
gsidebo: can you change this description to something like: "Test class that mocks the MicroMasters indexing API to avoid unnecessary ES index operations"

search/base.py
```diff
@@ -4,23 +4,23 @@
...
 
-from search.indexing_api import recreate_index
+from search import indexing_api
+from search.indexing_api import recreate_index, delete_index
 
 
-class ESTestCase(TestCase):
+class ESTestCaseWithIndex(TestCase):
     """
     Set up ES index on setup
```
gsidebo: can you change this description to something like: "Test class for test cases that need a live ES index"

search/base.py
giocalitri: I went down this road because renaming every single test class that uses `ESTestCase` is going to be a lot of work, unless I find some kind of magic unix command to do it...

search/base.py
giocalitri: sure

search/base.py
giocalitri: sure

search/indexing_api.py
giocalitri: ok

search/indexing_api.py
giocalitri: I thought about this and decided to just copy the code because the `clear_index` will still need `conn` and `index_name`, so either I make `delete_index` return these two variables (but it does not make sense to me, given the scope of `delete_index` or I keep the statements anyway... and at this point I just save 1 line... 
not sure... this is up to you

search/base.py
giocalitri: ok... I might have found the magic command

giocalitri:
I think I addressed or replied all the comments.

gsidebo:
looks great! 👍

Issue #2080, opened 2016-12-12
Exams: Validate exam authorizations were accepted by Pearson


rhysyngsun:
This validates that exam authorizations sent by the mechanism for #2076 were successfully processed and acknowledged by Pearson. The files are referred to as Exam Authorization Confirmations (EAC). Upon validation of acceptance it would then be possible for the user to SSO over to the Pearson site and complete exam scheduling for the given exam.

- [x] Download the EAC files from the SFTP server
- [x] Process the EAC files for success/failures
  - With a success, user should be ready to SSO and complete exam scheduling
  - With a failure, email sent to `MICROMASTERS_ADMIN_EMAIL`
- [x] Remove processed files from SFTP server

Open Questions:

- [x] What do we want to trigger on failure of an exam authorization? Probably an email to someone?

rhysyngsun:
@pdpinch questions over here (not immediate priority)

pdpinch:
For exam authorization errors, I would suggest 1) raising an exception so that it's caught by Sentry and logged and 2) e-mail the address in MICROMASTERS_ADMIN_EMAIL

amir-qayyum-khan:
@pdpinch @rhysyngsun 

- How many records will be in one EAC file?
- Currently i am sending summery of failures for each file as email to ``MICROMASTERS_ADMIN_EMAIL``, is this approach right?

pdpinch:
1. Why does the number of records matter? The number of records in an EAC file should correspond to the number in the EAD file. For the pilot, this could be as many as 4500, although probably will be much less. For the May exam offering, it looks like it could be around 3000. 
2. Sending the summary of failures to MICROMASTERS_ADMIN_EMAIL sounds good. Are they also logged somehow?
3. I defer to @rhysyngsun but I think we'd want to keep the file around in the case of any error. We aren't worried about filename collisions, right? 
4. @rhysyngsun can confirm, but I think we will only get a response from Pearson once every 24 hours.

amir-qayyum-khan:
thanks @pdpinch 

> Why does the number of records matter? The number of records in an EAC file should correspond to the number in the EAD file. For the pilot, this could be as many as 4500, although probably will be much less. For the May exam offering, it looks like it could be around 3000.

This was for my information.

> Sending the summary of failures to MICROMASTERS_ADMIN_EMAIL sounds good. Are they also logged somehow?

Yes i am logging failure.

rhysyngsun:
[...]

Issue #1797, opened 2016-11-17
Exams: Validate the user profiles were accepted by Pearson


pdpinch:
Part of #1796 

Pearson will ingest user profiles from a Candidate Data Document delivered over sftp (see #1795). The results of the ingestion will be reported back in the same sftp site, in a VCDD file (VUE Candidate Data Confirmation). We will want to update our database with the current status of user profiles and also notify us of any errors. 

- [ ] Download VCDC files from the SFTP server
- [ ] Process the VCDC files for success/failures
    - Send email to `MICROMASTERS_ADMIN_EMAIL` for failures
- [ ] Remove processed files from SFTP server

rhysyngsun:
@pdpinch I've detailed out this ticket, it's still in `In Design`, is there any reason it can't be `Ready` at this point?

pdpinch:
LGTM

PR #2387, opened 2017-01-19
Added base reader class for Pearson TSV responses

rhysyngsun:

None - this PR is for common work needed in #1797 and #2080.

#### What's this PR do?

This PR implements a base TSV file reader in the style of the writers implemented in `exams/pearson/writers.py`.

exams/pearson/readers.py

There is no user-facing code for this yet, so passing tests are all we need here.

Base Class

### DIFF
--- a/exams/pearson/constants.py
+++ b/exams/pearson/constants.py
@@ -1,10 +1,6 @@
-"""
-Pearson-related constants
-"""
+"""Pearson-related constants"""
 
 # Pearson TSV constants
-PEARSON_CSV_DIALECT = 'pearsontsv'
-
 PEARSON_DATE_FORMAT = "%Y/%m/%d"
 PEARSON_DATETIME_FORMAT = "%Y/%m/%d %H:%M:%S"
 
@@ -16,3 +12,8 @@ PEARSON_UPLOAD_REQUIRED_SETTINGS = [
     "EXAMS_SFTP_PASSWORD",
     "EXAMS_SFTP_UPLOAD_DIR",
 ]
+
+# Common options for Pearson TSV readers/writers
+PEARSON_DIALECT_OPTIONS = {
+    'delimiter': '\t',
+}
--- a/exams/pearson/readers.py
+++ b/exams/pearson/readers.py
@@ -1,0 +1,110 @@
+"""
+Readers for data Pearson exports
+"""
+import csv
+from datetime import datetime
+
+import pytz
+
+from exams.pearson.constants import (
+    PEARSON_DATETIME_FORMAT,
+    PEARSON_DIALECT_OPTIONS,
+)
+from exams.pearson.exceptions import InvalidTsvRowException
+
+
+class BaseTSVReader(object):
+    """
+    Base class for TSV file readers
+
+    It handles the high-level mapping and reading of the data.
+    Subclasses specify how the fields map.
+    """
+    def __init__(self, field_mappers, read_as_cls):
+        """
+        Initializes a new TSV writer
+
+        The first value of each fields tuple is the destination field name.
+        The second value is a str property path (e.g. "one.two.three") or
+        a callable that when passed a row returns a computed field value
+
+        Usage:
+
+            # maps column 'ABC' to field 'abc' and casts it to an int
+            # maps column 'DEF' to field 'def' and leaves it as a str
+            # Initializes each row into the namedtuple
+            Record = namedtuple('Record', ['abc', 'def'])
+
+            BaseTSVReader([
+                ('ABC', 'abc', int),
+                ('DEF', 'def'),
+            ], Record)
+
+        Arguments:
+            field_mappers (list(tuple)): list of tuple field mappers
+            read_as_cld (cls): class to instantiate row data
+        """
+        self.read_as_cls = read_as_cls
+        self.field_mappers = field_mappers
+
+    @classmethod
+    def parse_datetime(cls, dt):
+        """
+        Parses a datetime from Pearson's format
+        """
+        return datetime.strptime(dt, PEARSON_DATETIME_FORMAT).replace(tzinfo=pytz.UTC)
+
+    @classmethod
+    def map_cell(cls, row, source, target, transformer=None):
+        """
+        Maps an individual cell of a row
+
+        Args:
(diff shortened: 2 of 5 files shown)


exams/pearson/constants.py
```diff
@@ -16,3 +17,9 @@
     "EXAMS_SFTP_PASSWORD",
     "EXAMS_SFTP_UPLOAD_DIR",
 ]
+
+# Custom csv dialect for Pearson
+csv.register_dialect(
+    PEARSON_CSV_DIALECT,
+    delimiter='\t',
+)
```
giocalitri: does this mean that you configure the `cvs` globally here at application load time?

exams/pearson/readers.py
```diff
@@ -0,0 +1,100 @@
...
+            Record = namedtuple('Record', ['abc', 'def'])
+
+            BaseTSVReader([
+                ('ABC', 'abc', int),
+                ('DEF', 'def'),
+            ], Record)
+
+        Arguments:
+            field_mappers (list(tuple)): list of tuple field mappers
+            read_as_cld (cls): class to instantiate row data
```
giocalitri: is this docstring supposed to be here? I think it could make more sense to move it at the class level... but I am open to discussion.

exams/pearson/readers.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        self.read_as_cls = read_as_cls
+        self.field_mappers = field_mappers
+
+    @classmethod
+    def parse_datetime(cls, dt):
+        """
+        Parses a datetime from Pearson's format
+        """
+        return datetime.strptime(dt, PEARSON_DATETIME_FORMAT).replace(tzinfo=pytz.UTC)
```
giocalitri: what if the parsed string is not a valid one? should we just raise the default exception?

exams/pearson/writers_test.py
```diff
@@ -186,7 +187,7 @@ def test_write_skips_invalid_rows(self):  # pylint: disable=no-self-use
 
 
 @ddt.ddt
-class CDDWriterTest(TSVWriterTestCase):
+class CDDWriterTest(TSVWriterTestCase, TestCase):
```
giocalitri: Please do not use `django.test.TestCase`: use `search.base.MockedESTestCase` or `ESTestCase`: `django.test.TestCase`could have some side effects now that we mock all the indexing functions.

exams/pearson/readers.py
rhysyngsun: Yes, I think that would be best to let the `ValueError` raise. If it doesn't parse that means Pearson made a breaking change to their API and I'd want it reported to sentry and to abort the processing.

exams/pearson/readers.py
rhysyngsun: I believe generally we follow this which says either way is acceptable: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

In practice, so far we're using the style used in this PR [here](https://github.com/mitodl/micromasters/blob/bf1ec5f0e9c38855190288b895ccf215e8c5b062/dashboard/api_edx_cache.py#L30) and [here](https://github.com/mitodl/micromasters/blob/4b000f07106e814bfe546cb4b92e55dbaad7403f/dashboard/utils.py#L43) so I think if this is something we want to change, that's fair for discussion, but I don't think in the context of this PR.

exams/pearson/writers_test.py
rhysyngsun: I'll make those changes, but are these new testing best practices documented somewhere? Meaning, if it wasn't you reviewing this, how would this have got caught?

exams/pearson/readers.py
```diff
@@ -0,0 +1,100 @@
...
+        Maps a row object to a row dict
+
+        Args:
+            row: the row to map to a dict
+
+        Returns:
+            object:
+                row mapped to an object using the field mappers and read_as_cls
+        """
+        def map_cell(source, target, transformer=None):
```
rhysyngsun: Not particularly, I can move it to a classmethod.

exams/pearson/readers.py
giocalitri: ok

exams/pearson/readers.py
giocalitri: ok

exams/pearson/writers_test.py
giocalitri: theoretically after my PR about the tests optimizations in case of problems you should see the tests fail, so you are fine here, but you are right, we should document what I just said.
Feel free to leave as is if you want.

exams/pearson/constants.py
rhysyngsun: I was, I'm not a huge fan of how Python implements dialects in global state like this, but that's how the API is designed :-1: . Since we're only configuring 1 setting now and it can be passed to the writer/writer constructors instead I've refactored this to use an options `dict` instead to avoid the global code.

exams/pearson/readers.py
```diff
@@ -54,6 +54,27 @@ def parse_datetime(cls, dt):
         """
         return datetime.strptime(dt, PEARSON_DATETIME_FORMAT).replace(tzinfo=pytz.UTC)
 
+    @classmethod
+    def map_cell(cls, row, source, target, transformer=None):
```
giocalitri: Just one question: why is this defined as a `classmethod` if you access it from another method via `self`

exams/pearson/writers_test.py
rhysyngsun: I think since these tests mute the `post_save` signal that we never hit indexing code in these tests so it shouldn't be an issue.

exams/pearson/readers.py
rhysyngsun: `self` isn't used here so there's no need for it to be an instance method and have to add `# pylint: disable=no-self-use` clutter

exams/pearson/writers_test.py
giocalitri: ok

exams/pearson/constants.py
giocalitri: ok

giocalitri:
👍 when tests pass

PR #2388, opened 2017-01-19
Redeem coupon during checkout

### LINKED ISSUE #2325: Coupons: redeem coupon in checkout
Part of #2158 

 - This should add a `RedeemedCoupon` to attach a coupon to an order, which should prevent other people from redeeming the coupon, unless it's an automatic coupon
 - This should include calculating the price of the item based on the coupon

noisecapella:
Fixes #2325 

#### What's this PR do?
- Adds a `RedeemedCoupon` between an order and coupon when the order is created
- Lowers the price in the backend when order is fulfilled

The `RedeemedCoupon` should only prevent other users from purchasing when the order is fullfiled

 - Create a coupon locally in the Django shell using `CouponFactory`. Set `content_object` to a relevant run, course, or program which is available for purchase. Make sure that the discount will make the final price $0, to make it easier to test.
 - Attach your user to the coupon using the API. POST `/api/v0/coupons/<coupon.coupon_code>/users/` with `{username: 'your_edx_username'}` to do this, or just create `UserCoupon` in the Django shell
 - Purchase the course. The price will show up as the regular price of the course (inc financial aid), the coupon is not accounted for in the front end yet. The coupon should take effect when you actually make the purchase though. Because the coupon is for the full remaining price you should immediately be redirected to the dashboard with an order success message.

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -468,8 +468,8 @@ def calculate_run_price(course_run, user):
     enrollment = get_object_or_404(ProgramEnrollment, program=program, user=user)
     price = get_formatted_course_price(enrollment)['price']
     coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
-    coupon = None
-    if len(coupons) > 0:
-        coupon = coupons[0]
-        price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
+    if not coupons:
+        return price, None
+    coupon = coupons[0]
+    price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
     return price, coupon
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -39,6 +39,7 @@ from ecommerce.models import (
     Coupon,
     Line,
     Order,
+    RedeemedCoupon,
 )
 from financialaid.api import get_formatted_course_price
 from financialaid.models import (
@@ -104,27 +105,25 @@ def get_purchasable_course_run(course_key, user):
 
 
 @transaction.atomic
-def create_unfulfilled_order(course_id, user):
+def create_unfulfilled_order(course_key, user):
     """
     Create a new Order which is not fulfilled for a purchasable course run. If course run is not purchasable,
     it raises an Http404
 
     Args:
-        course_id (str):
+        course_key (str):
             A course key
         user (User):
             The purchaser of the course run
     Returns:
         Order: A newly created Order for the CourseRun with the given course_id
     """
-    course_run = get_purchasable_course_run(course_id, user)
-    enrollment = get_object_or_404(ProgramEnrollment, program=course_run.course.program, user=user)
-    price_dict = get_formatted_course_price(enrollment)
-    price = price_dict['price']
+    course_run = get_purchasable_course_run(course_key, user)
+    price, coupon = calculate_run_price(course_run, user)
     if price < 0:
         log.error(
             "Price to be charged for course run %s for user %s is less than zero: %s",
-            course_id,
+            course_key,
             get_social_username(user),
             price,
         )
@@ -137,10 +136,12 @@ def create_unfulfilled_order(course_id, user):
     )
     Line.objects.create(
         order=order,
-        course_key=course_run.edx_course_key,
+        course_key=course_key,
         description='Seat for {}'.format(course_run.title),
         price=price,
     )
+    if coupon is not None:
+        RedeemedCoupon.objects.create(order=order, coupon=coupon)
     order.save_and_log(user)
     return order
 
@@ -397,7 +398,8 @@ def is_coupon_redeemable(coupon, user):
 
 def pick_coupons(user):
     """
-    Choose the coupon which would be used in redemptions by the user.
+    Choose the coupons which would be used in redemptions by the user. There should be at most one coupon
+    per program in the output.
 
     The heuristic is currently:
      - choose attached coupons over automatic coupons
@@ -407,7 +409,7 @@ def pick_coupons(user):
         user (django.contrib.auth.models.User): A user
 
     Returns:
-        Coupon: The coupon which will be used by the user when redeeming runs in a program
+        list of Coupon: The coupons which will be used by the user when redeeming runs in a program
     """
     sorted_attached_coupons = Coupon.user_coupon_qset(user).order_by('-usercoupon__updated_on')
     sorted_automatic_coupons = Coupon.is_automatic_qset().order_by('-updated_on')
@@ -426,3 +428,48 @@ def pick_coupons(user):
             program_ids.add(program_id)
 
     return coupons
+
+
+def calculate_coupon_price(coupon, price, course_key):
+    """
+    Calculate the adjusted price given a coupon
+
+    Args:
+        coupon (Coupon): A coupon
+        price (decimal.Decimal): A price
+        course_key (str): An edX course key
+
+    Returns:
+        decimal.Decimal: An adjusted price
+    """
+    if course_key not in coupon.course_keys:
+        return price
+    if coupon.amount_type == Coupon.PERCENT_DISCOUNT:
+        return price * (1-coupon.amount)
+    elif coupon.amount_type == Coupon.FIXED_DISCOUNT:
+        return price - coupon.amount
+    else:
+        return price
+
+
+def calculate_run_price(course_run, user):
+    """
+    Calculate the price of a course given the coupons and financial aid available to the user.
+
+    Args:
+        course_run (CourseRun): A course run
+        user (django.contrib.auth.models.User): A user
+
+    Returns:
+        (decimal.Decimal, Coupon):
+            The adjusted of the course, and the coupon used if any
+    """
+    program = course_run.course.program
+    enrollment = get_object_or_404(ProgramEnrollment, program=program, user=user)
+    price = get_formatted_course_price(enrollment)['price']
+    coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
+    coupon = None
+    if len(coupons) > 0:
+        coupon = coupons[0]
+        price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
+    return price, coupon
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -808,7 +808,7 @@ class PriceTests(MockedESTestCase):
         course_run, _ = create_purchasable_course_run()
         price = Decimal('0.3')
         coupon = CouponFactory.create(content_object=course_run)
-        # Use manager to get around validation
+        # Use manager to skip validation, which usually prevents setting content_object to an arbitrary object
         Coupon.objects.filter(id=coupon.id).update(amount_type='xyz')
         coupon.refresh_from_db()
         assert calculate_coupon_price(coupon, price, course_run.edx_course_key) == price
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -748,7 +748,8 @@ class PriceTests(MockedESTestCase):
         coupon = CouponFactory.create()
         UserCoupon.objects.create(coupon=coupon, user=user)
         discounted_price = 5
-        fa_price = get_formatted_course_price(course_run.course.program.programenrollment_set.first())['price']
+        program_enrollment = course_run.course.program.programenrollment_set.first()
+        fa_price = get_formatted_course_price(program_enrollment)['price']
         with patch('ecommerce.api.calculate_coupon_price', autospec=True) as _calculate_coupon_price:
             _calculate_coupon_price.return_value = discounted_price
             assert calculate_run_price(course_run, user) == (fa_price, None)
@@ -774,7 +775,8 @@ class PriceTests(MockedESTestCase):
         with patch('ecommerce.api.calculate_coupon_price', autospec=True) as _calculate_coupon_price:
             _calculate_coupon_price.return_value = discounted_price
             assert calculate_run_price(course_run, user) == (discounted_price, coupon)
-        fa_price = get_formatted_course_price(course_run.course.program.programenrollment_set.first())['price']
+        program_enrollment = course_run.course.program.programenrollment_set.first()
+        fa_price = get_formatted_course_price(program_enrollment)['price']
         _calculate_coupon_price.assert_called_with(coupon, fa_price, course_run.edx_course_key)
 
     def test_percent_discount(self):
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -203,10 +203,8 @@ class PurchasableTests(MockedESTestCase):
         """
         course_run, user = create_purchasable_course_run()
         ProgramEnrollment.objects.filter(program=course_run.course.program, user=user).delete()
-        try:
+        with self.assertRaises(Http404):
             create_unfulfilled_order(course_run.edx_course_key, user)
-        except Http404:
-            pass
 
     def test_already_purchased(self):
         """
@@ -762,10 +760,8 @@ class PriceTests(MockedESTestCase):
         """
         course_run, user = create_purchasable_course_run()
         ProgramEnrollment.objects.filter(program=course_run.course.program, user=user).delete()
-        try:
+        with self.assertRaises(Http404):
             calculate_run_price(course_run, user)
-        except Http404:
-            pass
 
     def test_calculate_run_price_coupon(self):
         """
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -736,7 +736,7 @@ class PickCouponTests(MockedESTestCase):
             _is_coupon_redeemable.assert_any_call(coupon, self.user)
 
 
-class PriceTests(ESTestCase):
+class PriceTests(MockedESTestCase):
     """
     Tests for calculating prices
     """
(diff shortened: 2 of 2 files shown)


giocalitri:
when I tried to purchase a course as suggested in the PR description, I get a `404`.
I thought it was a problem with the configuration of the courses, but from the logs it does not seem that the view is even reached.

```
web_1     | [2017-01-23 13:33:21] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.002) SELECT "wagtailcore_site"."id", "wagtailcore_site"."hostname", "wagtailcore_site"."port", "wagtailcore_site"."site_name", "wagtailcore_site"."root_page_id", "wagtailcore_site"."is_default_site", CASE WHEN ("wagtailcore_site"."port" = 8079 AND "wagtailcore_site"."hostname" = '192.168.99.100') THEN 0 WHEN ("wagtailcore_site"."is_default_site" = true AND "wagtailcore_site"."hostname" = '192.168.99.100') THEN 1 WHEN "wagtailcore_site"."is_default_site" = true THEN 2 ELSE 3 END AS "match", "wagtailcore_page"."id", "wagtailcore_page"."path", "wagtailcore_page"."depth", "wagtailcore_page"."numchild", "wagtailcore_page"."title", "wagtailcore_page"."slug", "wagtailcore_page"."content_type_id", "wagtailcore_page"."live", "wagtailcore_page"."has_unpublished_changes", "wagtailcore_page"."url_path", "wagtailcore_page"."owner_id", "wagtailcore_page"."seo_title", "wagtailcore_page"."show_in_menus", "wagtailcore_page"."search_description", "wagtailcore_page"."go_live_at", "wagtailcore_page"."expire_at", "wagtailcore_page"."expired", "wagtailcore_page"."locked", "wagtailcore_page"."first_published_at", "wagtailcore_page"."latest_revision_created_at" FROM "wagtailcore_site" INNER JOIN "wagtailcore_page" ON ("wagtailcore_site"."root_page_id" = "wagtailcore_page"."id") WHERE ("wagtailcore_site"."hostname" = '192.168.99.100' OR "wagtailcore_site"."is_default_site" = true) ORDER BY "match" ASC; args=(8079, '192.168.99.100', 0, True, '192.168.99.100', 1, True, 2, 3, '192.168.99.100', True)
web_1     | [2017-01-23 13:33:22] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.001) SELECT "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined" FROM "auth_user" WHERE "auth_user"."id" = 1; args=(1,)
web_1     | [2017-01-23 13:33:22] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.002) SELECT "courses_courserun"."id", "courses_courserun"."title", "courses_courserun"."edx_course_key", "courses_courserun"."enrollment_start", "courses_courserun"."start_date", "courses_courserun"."enrollment_end", "courses_courserun"."end_date", "courses_courserun"."upgrade_deadline", "courses_courserun"."freeze_grade_date", "courses_courserun"."fuzzy_start_date", "courses_courserun"."fuzzy_enrollment_start_date", "courses_courserun"."enrollment_url", "courses_courserun"."prerequisites", "courses_courserun"."course_id" FROM "courses_courserun" INNER JOIN "courses_course" ON ("courses_courserun"."course_id" = "courses_course"."id") INNER JOIN "courses_program" ON ("courses_course"."program_id" = "courses_program"."id") WHERE ("courses_program"."live" = true AND "courses_courserun"."edx_course_key" = 'course-v1:odl+MX+2016'); args=(True, 'course-v1:odl+MX+2016')
web_1     | [2017-01-23 13:33:22] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.000) SELECT "courses_course"."id", "courses_course"."program_id", "courses_course"."position_in_program", "courses_course"."title", "courses_course"."thumbnail", "courses_course"."description", "courses_course"."prerequisites", "courses_course"."exam_module" FROM "courses_course" WHERE "courses_course"."id" = 7; args=(7,)
web_1     | [2017-01-23 13:33:22] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.001) SELECT "courses_program"."id", "courses_program"."title", "courses_program"."live", "courses_program"."description", "courses_program"."financial_aid_availability", "courses_program"."ga_tracking_id", "courses_program"."exam_series_code" FROM "courses_program" WHERE "courses_program"."id" = 1; args=(1,)
web_1     | [2017-01-23 13:33:22] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.001) SELECT "courses_courserun"."id", "courses_courserun"."title", "courses_courserun"."edx_course_key", "courses_courserun"."enrollment_start", "courses_courserun"."start_date", "courses_courserun"."enrollment_end", "courses_courserun"."end_date", "courses_courserun"."upgrade_deadline", "courses_courserun"."freeze_grade_date", "courses_courserun"."fuzzy_start_date", "courses_courserun"."fuzzy_enrollment_start_date", "courses_courserun"."enrollment_url", "courses_courserun"."prerequisites", "courses_courserun"."course_id" FROM "courses_courserun" INNER JOIN "courses_course" ON ("courses_courserun"."course_id" = "courses_course"."id") INNER JOIN "courses_program" ON ("courses_course"."program_id" = "courses_program"."id") INNER JOIN "ecommerce_courseprice" ON ("courses_courserun"."id" = "ecommerce_courseprice"."course_run_id") WHERE ("courses_program"."live" = true AND "courses_program"."financial_aid_availability" = true AND "courses_courserun"."edx_course_key" = 'course-v1:odl+MX+2016' AND "ecommerce_courseprice"."is_valid" = true); args=(True, True, 'course-v1:odl+MX+2016', True)
web_1     | [2017-01-23 13:33:22] DEBUG 437 [django.db.backends] utils.py:90 - [a85ecf9cd161] - (0.001) SELECT "wagtailredirects_redirect"."id", "wagtailredirects_redirect"."old_path", "wagtailredirects_redirect"."site_id", "wagtailredirects_redirect"."is_permanent", "wagtailredirects_redirect"."redirect_page_id", "wagtailredirects_redirect"."redirect_link" FROM "wagtailredirects_redirect" WHERE (("wagtailredirects_redirect"."site_id" = 2 OR "wagtailredirects_redirect"."site_id" IS NULL) AND "wagtailredirects_redirect"."old_path" = '/api/v0/checkout'); args=(2, '/api/v0/checkout')
web_1     | [2017-01-23 13:33:22] WARNING 437 [django.request] base.py:152 - [a85ecf9cd161] - Not Found: /api/v0/checkout/
nginx_1   | 192.168.99.1 - - [23/Jan/2017:13:33:22 +0000] "POST /api/v0/checkout/ HTTP/1.1" 404 34 "http://192.168.99.100:8079/dashboard/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:50.0) Gecko/20100101 Firefox/50.0" "-"
web_1     | [pid: 437|app: 0|req: 1/1] 192.168.99.1 () {50 vars in 1235 bytes} [Mon Jan 23 13:33:21 2017] POST /api/v0/checkout/ => generated 23 bytes in 613 msecs (HTTP/1.1 404) 4 headers in 131 bytes (1 switches on core 0)
```

I am pretty sure if this is a problem with my environment, but we should probably understand why this is happening.

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+        coupon.refresh_from_db()
+        assert calculate_coupon_price(coupon, price, course_run.edx_course_key) == price
+
+    def test_coupon_allowed(self):
+        """
+        Assert that the price is not adjusted if the coupon doesn't affect the course run
+        """
+        course_run, _ = create_purchasable_course_run()
+        price = Decimal('0.3')
+        coupon = CouponFactory.create()
```
singingwolfboy: Can you be a little more explicit, here? Why doesn't the coupon affect the course run? It could be as simple as adding a line like this:
```python
assert coupon.content_object != course_run
```

ecommerce/api.py
```diff
@@ -137,10 +136,12 @@ def create_unfulfilled_order(course_id, user):
     )
     Line.objects.create(
         order=order,
-        course_key=course_run.edx_course_key,
+        course_key=course_key,
         description='Seat for {}'.format(course_run.title),
         price=price,
     )
+    if coupon is not None:
+        RedeemedCoupon.objects.create(order=order, coupon=coupon)
```
singingwolfboy: `RedeemedCoupon` has a `UNIQUE` constraint in the database on order and coupon. What happens if the code gets to this point and there's already a row in the database for this? If it shouldn't raise an exception, you can use `get_or_create()` instead of `create()`. If it should, you should explicitly handle the exception somewhere, and have a test for it.

ecommerce/api.py
```diff
@@ -426,3 +428,48 @@ def pick_coupons(user):
...
+
+    Returns:
+        (decimal.Decimal, Coupon):
+            The adjusted of the course, and the coupon used if any
+    """
+    program = course_run.course.program
+    enrollment = get_object_or_404(ProgramEnrollment, program=program, user=user)
+    price = get_formatted_course_price(enrollment)['price']
+    coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
+    coupon = None
```
singingwolfboy: Nitpick: instead of setting `coupon = None` and then maybe populating the `coupon` variable, I would suggest returning early. Like this:

```python
coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
if not coupons:
    return price, None
coupon = coupons[0]
price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
return price, coupon
```

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+    def test_no_program_enrollment(self):
+        """
+        If a user is not enrolled a 404 should be raised when getting the price
+        """
+        course_run, user = create_purchasable_course_run()
+        ProgramEnrollment.objects.filter(program=course_run.course.program, user=user).delete()
+        try:
+            calculate_run_price(course_run, user)
+        except Http404:
+            pass
```
singingwolfboy: Definitely use either [`self.assertRaises()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) or [`pytest.raises()`](http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions).

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+
+    def test_calculate_run_price_coupon(self):
+        """
+        If there is a coupon calculate_run_price should use calculate_coupon_price to get the discounted price
+        """
+        course_run, user = create_purchasable_course_run()
+        coupon = CouponFactory.create(content_object=course_run)
+        UserCoupon.objects.create(coupon=coupon, user=user)
+        discounted_price = 5
+        with patch('ecommerce.api.calculate_coupon_price', autospec=True) as _calculate_coupon_price:
```
singingwolfboy: Rather than mocking `calculate_coupon_price`, it would be a better test if you set an explicit price for the course, set an explicit discount for the coupon, and assert an explicit return value for the result of `calculate_run_price`.

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+    def test_percent_discount(self):
+        """
+        Assert the price with a percent discount
+        """
+        course_run, _ = create_purchasable_course_run()
+        price = Decimal(5)
+        coupon = CouponFactory.create(
+            content_object=course_run, amount_type=Coupon.PERCENT_DISCOUNT, amount=Decimal("0.3")
+        )
+        assert calculate_coupon_price(coupon, price, course_run.edx_course_key) == price * (1 - coupon.amount)
```
singingwolfboy: Used fixed values in assertions when possible. Rather than comparing with `price * (1 - coupon.amount)`, compare with `Decimal("3.5")`.

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+    def test_fixed_discount(self):
+        """
+        Assert the price with a fixed discount
+        """
+        course_run, _ = create_purchasable_course_run()
+        price = Decimal(5)
+        coupon = CouponFactory.create(
+            content_object=course_run, amount_type=Coupon.FIXED_DISCOUNT, amount=Decimal("1.5")
+        )
+        assert calculate_coupon_price(coupon, price, course_run.edx_course_key) == price - coupon.amount
```
singingwolfboy: Rather than comparing with `price - coupon.amount`, compare with `Decimal("3.5")`.

noisecapella:
Who is doing the review? Did either of you do a functionality test?

giocalitri:
> Who is doing the review? Did either of you do a functionality test?

I reviewed the code, but I am unable to functionally test it

singingwolfboy:
@noisecapella I noticed that several other coupon-related pull requests were dependent on this pull request, so I wanted to help review this one. I'll attempt a functionality test now.

ecommerce/api.py
```diff
@@ -137,10 +136,12 @@ def create_unfulfilled_order(course_id, user):
...
     Line.objects.create(
         order=order,
-        course_key=course_run.edx_course_key,
+        course_key=course_key,
         description='Seat for {}'.format(course_run.title),
         price=price,
     )
+    if coupon is not None:
+        RedeemedCoupon.objects.create(order=order, coupon=coupon)
     order.save_and_log(user)
```
noisecapella: The whole function is wrapped in a transaction already

ecommerce/api.py
```diff
@@ -426,3 +428,48 @@ def pick_coupons(user):
...
+
+    Returns:
+        decimal.Decimal: An adjusted price
+    """
+    if course_key not in coupon.course_keys:
+        return price
+    if coupon.amount_type == Coupon.PERCENT_DISCOUNT:
+        return price * (1-coupon.amount)
+    elif coupon.amount_type == Coupon.FIXED_DISCOUNT:
+        return price - coupon.amount
```
noisecapella: I'm not sure, I'll talk to @pdpinch about it

ecommerce/api_test.py
noisecapella: Will do

ecommerce/api.py
giocalitri: ok

ecommerce/api.py
noisecapella: `order` is created in this function so there shouldn't be a row in the database already

ecommerce/api.py
```diff
@@ -426,3 +428,48 @@ def pick_coupons(user):
...
+        (decimal.Decimal, Coupon):
+            The adjusted of the course, and the coupon used if any
+    """
+    program = course_run.course.program
+    enrollment = get_object_or_404(ProgramEnrollment, program=program, user=user)
+    price = get_formatted_course_price(enrollment)['price']
+    coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
+    coupon = None
+    if len(coupons) > 0:
+        coupon = coupons[0]
```
noisecapella: `pick_coupons` only returns one coupon per program, so if there's a coupon in the list it's the only one

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+    """
+
+    def test_calculate_run_price_no_coupons(self):
+        """
+        If there are no coupons for this program the price should be what get_formatted_course_price returned
+        """
+        course_run, user = create_purchasable_course_run()
+        # This coupon is for a different program
+        coupon = CouponFactory.create()
+        UserCoupon.objects.create(coupon=coupon, user=user)
```
noisecapella: `UserCoupon` only has the two fields, so what's the point of a factory class?

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+        discounted_price = 5
+        fa_price = get_formatted_course_price(course_run.course.program.programenrollment_set.first())['price']
+        with patch('ecommerce.api.calculate_coupon_price', autospec=True) as _calculate_coupon_price:
+            _calculate_coupon_price.return_value = discounted_price
+            assert calculate_run_price(course_run, user) == (fa_price, None)
+        assert _calculate_coupon_price.called is False
```
noisecapella: Both seem equally readable to me

ecommerce/api_test.py
noisecapella: I want to make sure `calculate_coupon_price` is mocked so that it verifies that `calculate_run_price` is using it. What you're describing will only test that particular fixed discount case, but if I mock `calculate_coupon_price` then I know it works for the percent discount or the fixed price coupon that's coming in the future

ecommerce/api_test.py
noisecapella: Why are fixed values preferable here?

noisecapella:
Can someone do a functional test please?

aliceriot:
I will do a functional test!

aliceriot:
functionality looks :100: !

noisecapella:
Thanks! @giocalitri Any other comments?

giocalitri:
👍

PR #2392, opened 2017-01-19
Attach user to coupon

### LINKED ISSUE #2323: Coupons: attach user to coupon
Part of #2158

If the user clicks on a coupon link (a link to the dashboard with a certain query parameter) these things should happen:

- the dashboard should POST the attachment to `/api/v0/coupons/<code>/users/` with their username to attach the user to the coupon. This call will return `200` even if the attachment already exists so there's no harm doing it twice
 - We should show the modal here, using the information from the coupon API to determine the message: https://projects.invisionapp.com/share/NB7VM1GSG#/screens/213871691
- remove the query parameter from the URL so that they don't see the modal again if they refresh

singingwolfboy:
Fixes #2323.

#### What's this PR do?
When the user visits the dashboard with a `coupon` query parameter in the URL, that query parameter is assumed to be a coupon code. This PR detects that query parameter, and tries to attach the currently logged-in user to that coupon. If the coupon is invalid or doesn't exist, we pop a toast notification that it didn't work. If the coupon exists and is valid for the user, we pop a notification to confirm that they have successfully applied the coupon.

* Try applying a coupon code for a coupon that doesn't exist, and confirm that you get a notification.
* Then make a coupon that is invalid, and confirm you get the same notification when you try to apply it.
* Make a percentage discount coupon that can be successfully applied, and apply it. You should get a confirmation notification, with information about the coupon you just applied.
* Try again with a fixed discount coupon that can be successfully applied. Make sure that the information in the confirmation notification is accurate.

fixed discount notification
percent discount notification
coupon failure notification

### DIFF
--- a/static/js/actions/ui.js
+++ b/static/js/actions/ui.js
@@ -105,6 +105,9 @@ export const setConfirmSkipDialogVisibility = createAction(SET_CONFIRM_SKIP_DIAL
 export const SET_DOCS_INSTRUCTIONS_VISIBILITY = 'SET_DOCS_INSTRUCTIONS_VISIBILITY';
 export const setDocsInstructionsVisibility = createAction(SET_DOCS_INSTRUCTIONS_VISIBILITY);
 
+export const SET_COUPON_NOTIFICATION_VISIBILITY = 'SET_COUPON_NOTIFICATION_VISIBILITY';
+export const setCouponNotificationVisibility = createAction(SET_COUPON_NOTIFICATION_VISIBILITY);
+
 export const SET_NAV_DRAWER_OPEN = 'SET_NAV_DRAWER_OPEN';
 export const setNavDrawerOpen = createAction(SET_NAV_DRAWER_OPEN);
 
--- a/static/js/components/CouponNotificationDialog.js
+++ b/static/js/components/CouponNotificationDialog.js
@@ -1,0 +1,102 @@
+// @flow
+import React from 'react';
+import Dialog from 'material-ui/Dialog';
+import Button from 'react-mdl/lib/Button';
+import {
+  COUPON_CONTENT_TYPE_PROGRAM,
+  COUPON_AMOUNT_TYPE_FIXED_DISCOUNT,
+  COUPON_AMOUNT_TYPE_PERCENT_DISCOUNT,
+} from '../constants';
+import type { Coupon } from '../flow/couponTypes';
+import type { AvailableProgram } from '../flow/enrollmentTypes';
+import type { Course } from '../flow/programTypes';
+
+type CouponNotification = {
+  coupon:               Coupon,
+  couponProgram:        ?AvailableProgram,
+  couponCourse:         ?Course,
+  open:                 boolean,
+  setDialogVisibility:  (v: boolean) => void,
+};
+
+const CouponNotificationDialog = (
+  { coupon, couponProgram, couponCourse, open, setDialogVisibility }: CouponNotification
+) => {
+  const {
+    amount,
+    amount_type: amountType,
+    content_type: contentType,
+    program_id: programId,
+    object_id: objectId,
+  } = coupon;
+  let programName;
+  if ( couponProgram ) {
+    programName = `the ${couponProgram.title} MicroMasters program`;
+  } else {
+    programName = `program ID ${programId}`;
+  }
+
+  let title, message;
+  if ( contentType === COUPON_CONTENT_TYPE_PROGRAM ) {
+    if ( amountType === COUPON_AMOUNT_TYPE_PERCENT_DISCOUNT ) {
+      title = `Coupon applied: ${amount.times(100)}% off each course!`;
+      message = <p>
+        This coupon gives <strong>a discount
+        of { amount.times(100).toString() }% off</strong> the price
+        of <strong>each</strong> course in { programName }.
+      </p>;
+    } else if ( amountType === COUPON_AMOUNT_TYPE_FIXED_DISCOUNT ) {
+      title = `Coupon applied: $${amount} off each course!`;
+      message = <p>
+        This coupon gives <strong>a discount
+        of ${ amount.toString() } off</strong> the price
+        of <strong>each</strong> course in { programName }.
+      </p>;
+    }
+  } else {
+    let courseName;
+    if ( couponCourse ) {
+      courseName = `${couponCourse.title} in ${programName}`;
+    } else {
+      courseName = `course ID ${objectId} in ${programName}`;
+    }
+    if ( amountType === COUPON_AMOUNT_TYPE_PERCENT_DISCOUNT ) {
+      title = `Coupon applied: ${amount.times(100)}% off!`;
+      message = <p>
+        This coupon gives <strong>a discount
+        of { amount.times(100).toString() }% off</strong> the price
+        of { courseName }.
+      </p>;
+    } else if ( amountType === COUPON_AMOUNT_TYPE_FIXED_DISCOUNT ) {
+      title = `Coupon applied: $${amount} off!`;
+      message = <p>
+        This coupon gives <strong>a discount
+        of ${ amount.toString() } off</strong> the price
+        of { courseName }.
+      </p>;
+    }
+  }
+
+  const okButton = <Button
+    type='ok'
+    key='ok'
+    className="primary-button ok-button"
+    onClick={() => setDialogVisibility(false)}>
+    OK
+  </Button>;
+
+  return <Dialog
+    title={title}
+    titleClassName="dialog-title"
+    contentClassName="dialog coupon-notification-dialog"
+    className="coupon-notification-dialog-wrapper"
+    actions={okButton}
+    open={open}
+    onRequestClose={() => setDialogVisibility(false)}
+  >
+    {message}
+    <p>The discount will be automatically applied when you enroll.</p>
+  </Dialog>;
+};
+
+export default CouponNotificationDialog;
--- a/static/js/components/CouponNotificationDialog.js
+++ b/static/js/components/CouponNotificationDialog.js
@@ -2,6 +2,8 @@
 import React from 'react';
 import Dialog from 'material-ui/Dialog';
 import Button from 'react-mdl/lib/Button';
+
+import { formatPrice, formatPercent } from '../util/util';
 import {
   COUPON_CONTENT_TYPE_PROGRAM,
   COUPON_AMOUNT_TYPE_FIXED_DISCOUNT,
@@ -39,10 +41,10 @@ const CouponNotificationDialog = (
   let title, message;
   if ( contentType === COUPON_CONTENT_TYPE_PROGRAM ) {
     if ( amountType === COUPON_AMOUNT_TYPE_PERCENT_DISCOUNT ) {
-      title = `Coupon applied: ${amount.times(100)}% off each course!`;
+      title = `Coupon applied: ${amount.times(100).toDecimalPlaces(0).toString()}% off each course!`;
       message = <p>
         This coupon gives <strong>a discount
-        of { amount.times(100).toString() }% off</strong> the price
+        of { amount.times(100).toDecimalPlaces(0).toString() }% off</strong> the price
         of <strong>each</strong> course in { programName }.
       </p>;
     } else if ( amountType === COUPON_AMOUNT_TYPE_FIXED_DISCOUNT ) {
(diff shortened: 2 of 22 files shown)


roberthouse54:
Let's just change the copy slightly.

Header:

"Coupon applied: $50 off each course!" 
or
"Coupon applied: 100% off each course!"

Paragraphs:

"This coupon gives **a discount of** [100% off | $50 off] the price of **each** course in the Data, Economics, and Development Policy MicroMasters program. 

The discount will be automatically applied when you enroll."

static/js/containers/DashboardPage.js
```diff
@@ -270,6 +314,51 @@ class DashboardPage extends React.Component {
...
+    const {
+      programs,
+      ui,
+      coupons,
+    } = this.props;
+    if ( !coupons.activeCoupon ) {
+      return null;
+    }
+    let couponProgram = null, couponCourse = null;
+    if ( programs.availablePrograms ) {
```
noisecapella: You should be able to skip this check, it will always be a list of some kind because the initial value is an empty list

static/js/containers/DashboardPage.js
```diff
@@ -270,6 +314,51 @@ class DashboardPage extends React.Component {
...
+      couponProgram = programs.availablePrograms.find(
+        program => program.id === coupons.activeCoupon.program_id
+      );
+    }
+    if (
+      couponProgram &&
+      couponProgram.courses &&
```
noisecapella: If `couponProgram` is of type `AvailableProgram`, there isn't a `courses` field. You'll need to use the dashboard API for that, or add it to the programs API, or change the dialog text to not use the course or run title

noisecapella:
Functionality looks great

static/js/test_constants.js
```diff
@@ -706,11 +710,28 @@ export const COURSE_PRICES_RESPONSE: CoursePrices = deepFreeze(DASHBOARD_RESPONS
...
+    amount: "0.55",
+    amount_type: COUPON_AMOUNT_TYPE_PERCENT_DISCOUNT,
+    content_type: COUPON_CONTENT_TYPE_PROGRAM,
+    coupon_code: "success-coupon",
+    object_id: 3,
+    program_id: 3,
+  }
+});
+
+export const COUPON = R.evolve(
```
singingwolfboy: done

static/js/containers/DashboardPage.js
singingwolfboy: skipped

static/js/containers/DashboardPage.js
singingwolfboy: I switched it to use the dashboard API.

static/js/actions/ui.js
```diff
@@ -105,6 +105,9 @@ export const setConfirmSkipDialogVisibility = createAction(SET_CONFIRM_SKIP_DIAL
 export const SET_DOCS_INSTRUCTIONS_VISIBILITY = 'SET_DOCS_INSTRUCTIONS_VISIBILITY';
 export const setDocsInstructionsVisibility = createAction(SET_DOCS_INSTRUCTIONS_VISIBILITY);
 
+export const SET_COUPON_NOTIFICATION_VISIBILITY = 'SET_COUPON_NOTIFICATION_VISIBILITY';
```
noisecapella: Can you add this action to the test in `ui_test.js`?

static/js/reducers/coupons.js
```diff
@@ -43,6 +46,8 @@ export const coupons = (state: CouponsState = INITIAL_COUPONS_STATE, action: Act
     };
   case RECEIVE_FETCH_COUPONS_FAILURE:
     return { ...state, fetchGetStatus: FETCH_FAILURE};
+  case SET_ACTIVE_COUPON:
```
noisecapella: Instead of using `activeCoupon` can we just refetch from the coupon list API? I think this will help with race conditions, although there's still this possibility:

    GET                         response
    POST    GET    response

Which would cause stale data. But I think it would be a good start to refetch from the coupon list API

noisecapella: In any case this race condition will be triggered by my future PR so I'll take care of this edge case

static/js/actions/ui.js
singingwolfboy: done

static/js/components/CouponNotificationDialog.js
```diff
@@ -0,0 +1,102 @@
...
+      </p>;
+    } else if ( amountType === COUPON_AMOUNT_TYPE_FIXED_DISCOUNT ) {
+      title = `Coupon applied: $${amount} off each course!`;
+      message = <p key='message'>
+        This coupon gives <strong>a discount
+        of ${ amount.toString() } off</strong> the price
+        of <strong>each</strong> course in { programName }.
+      </p>;
+    }
+  } else {
```
noisecapella: What about `COUPON_CONTENT_TYPE_COURSERUN`? Is both course and courserun handled here?

static/js/components/CouponNotificationDialog.js
```diff
@@ -0,0 +1,102 @@
...
+    title={title}
+    titleClassName="dialog-title"
+    contentClassName="dialog coupon-notification-dialog"
+    className="coupon-notification-dialog-wrapper"
+    actions={okButton}
+    open={open}
+    onRequestClose={() => setDialogVisibility(false)}
+  >
+    {message}
+    <p key='enroll'>The discount will be automatically applied when you enroll.</p>
```
noisecapella: Can you remove `key` here? It's only necessary if you have an array of react components or a very complicated nested structure

static/js/containers/DashboardPage.js
```diff
@@ -207,6 +217,38 @@ class DashboardPage extends React.Component {
...
+      return;
+    }
+
+    if ( coupons.fetchPostStatus !== undefined ) {
+      // If we've already launched a coupon fetch, don't launch another one
+      return;
+    }
+
+    dispatch(attachCoupon(query.coupon)).then(result => {
+      if ( !result.coupon ) {
```
noisecapella: Would this ever happen?

static/js/reducers/coupons_test.js
```diff
@@ -21,15 +22,19 @@ import {
 } from '../actions';
 import rootReducer from '../reducers';
 import * as api from '../lib/api';
+import type { CouponsState } from '../reducers/coupons';
+import type { AssertReducerResultState } from '../flow/reduxTypes';
+import { createAssertReducerResultState } from '../util/test_utils';
 
-describe('financial aid reducers', () => {
-  let sandbox, store, dispatchThen;
+describe('coupon reducers', () => {
```
noisecapella: Copy and paste strikes again :)

static/js/components/CouponNotificationDialog.js
singingwolfboy: Coupons for course runs are not handled. I was under the impression that we only really needed to handle coupons for programs for this pull request, and I was doing extra work by handling coupons for courses. If we need to handle course runs, as well, we'll need to figure out how to phrase the notification to make it clearer.

Can we merge this pull request without course run support for now, and add it in the future if necessary? This pull request is getting pretty big even without it.

static/js/components/CouponNotificationDialog.js
singingwolfboy: removed.

static/js/containers/DashboardPage.js
singingwolfboy: I guess not -- I was being overly defensive. Removed.

static/js/components/CouponNotificationDialog.js
noisecapella: Sure, can you file an issue so we can work on it later?

singingwolfboy: Done: #2440.

static/js/lib/api.js
```diff
@@ -243,15 +244,20 @@ export function addCourseEnrollment(courseId: string) {
 }
 
 export function getCoupons(): Promise<Coupons> {
-  return fetchJSONWithCSRF('/api/v0/coupons/');
+  return fetchJSONWithCSRF('/api/v0/coupons/').then(coupon => (
+    // turn `amount` from string into decimal
+    R.evolve({amount: Decimal}, coupon)
```
noisecapella: This should be an array, not an object:

screenshot from 2017-01-26 10-45-44

singingwolfboy: fixed

noisecapella:
Can you update the coupon test in `api_test.js` to test the parsing of the coupon response? Specifically that amount becomes a decimal

static/js/containers/DashboardPage.js
```diff
@@ -229,6 +237,36 @@ class DashboardPage extends React.Component {
...
+      // If there's no coupon code in the URL query parameters,
+      // there's nothing to do.
+      return;
+    }
+
+    if ( coupons.fetchPostStatus !== undefined ) {
+      // If we've already launched a coupon fetch, don't launch another one
+      return;
+    }
+
```
noisecapella: Can you add an additional `if` statement here? We should wait until the original GET will succeed before doing the POST to prevent a race condition. Can you return if `coupons.fetchGetStatus !== FETCH_SUCCESS`?

singingwolfboy: This function isn't dependent on the value of `state.coupons.coupons`, so it doesn't matter if it executes before or after that GET request returns. I suppose that you could have a race condition like this:

1. launch first `fetchCoupons()` API request
2. launch `attachCoupon()` API request
3. `attachCoupon()` returns, launch second `fetchCoupons()` API request
4. second `fetchCoupons()` returns, updates Redux store with accurate information
5. first `fetchCoupons()` _finally_ returns, updates Redux store with stale information

If that's the race condition you're thinking of, wouldn't it make more sense to abort the first `fetchCoupons()` API request in step 3, instead of waiting for it to finish?

noisecapella: Yes, that's the race condition. We can't abort fetches, at least not with the fetch API: http://stackoverflow.com/questions/31061838/how-do-i-cancel-an-http-fetch-request

singingwolfboy: OK, I wasn't aware you can't abort fetches -- good to know. Rather than ensuring that `fetchGetStatus === FETCH_SUCCESS`, wouldn't it be better to ensure that `fetchGetStatus !== FETCH_PROCESSING`? The fetch can fail, and we don't want that failure to prevent future fetches.

noisecapella: Sure, but `fetchGetStatus` will be `undefined` initially so that should be handled too

static/js/lib/api.js
```diff
@@ -243,15 +244,23 @@ export function addCourseEnrollment(courseId: string) {
 }
 
 export function getCoupons(): Promise<Coupons> {
-  return fetchJSONWithCSRF('/api/v0/coupons/');
+  return fetchJSONWithCSRF('/api/v0/coupons/').then(coupons => {
+    if ( !coupons ) {
```
noisecapella: What is this for? `coupons` should always return a list of something, which is always truthy in javascript

singingwolfboy: It's necessary for the tests. Many of them stub out `fetchJSONWithCSRF` to return `undefined`:
```js
fetchJSONStub.returns(Promise.resolve());
```

static/js/containers/DashboardPage_test.js
```diff
@@ -326,4 +342,82 @@ describe('DashboardPage', () => {
...
+          message: "This coupon code is invalid or does not exist.",
+          icon: TOAST_FAILURE
+        });
+      });
+    });
+
+    // This test is disabled because I can't figure out a good way to simulate
+    // this race condition in an automated test. If someone else can pick it
+    // up and fix it, that would be great!
+    xit('without a race condition', () => {  // eslint-disable-line mocha/no-skipped-tests
```
singingwolfboy: When I run this test, it says that it's not receiving the `RECEIVE_GET_PROGRAM_ENROLLMENTS_SUCCESS` action, which is odd -- I don't know why the `getPrograms()` API would be unable to return if the `getCoupons()` API is slow. More concerning is that the `REQUEST_ATTACH_COUPON` action and `RECEIVE_ATTACH_COUPON_SUCCESS` actions apparently _are_ being received in this test -- I was trying to write it in such a way that the code would return early so that `attachCoupon()` would never get called, but apparently it's still getting called. (As an aside, is there an easy way to assert that a particular action is _not_ called with `renderComponent`?)

If we can find a way to write this test correctly, so that it asserts on the right thing and passes reliably, then we should do so. Otherwise, I think we should just delete this test and leave the race condition avoidance code untested. @noisecapella: what do you think? Can you see how to fix this?

PR #2402, opened 2017-01-20
Remove readonly_fields for Coupon admin

### LINKED ISSUE #2400: Coupons: update admin to allow creating coupons
Part of #2158

This should just be a matter of changing the fields we mark as readonly

noisecapella:
Fixes #2400
Rebased on #2388 

#### What's this PR do?
Allows creating coupons in the Django admin

Create a coupon in the Django admin

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -80,7 +80,6 @@ class ReceiptAdmin(admin.ModelAdmin):
 class CouponAdmin(admin.ModelAdmin):
     """Admin for Coupon"""
     model = Coupon
-    readonly_fields = get_field_names(Coupon)
 
 
 class RedeemedCouponAdmin(admin.ModelAdmin):


ecommerce/api.py
```diff
@@ -426,3 +428,48 @@ def pick_coupons(user):
...
+    """
+    program = course_run.course.program
+    enrollment = get_object_or_404(ProgramEnrollment, program=program, user=user)
+    price = get_formatted_course_price(enrollment)['price']
+    coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
+    coupon = None
+    if len(coupons) > 0:
+        coupon = coupons[0]
+        price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
+    return price, coupon
```
gsidebo: personally i think this function should accept an optional `coupon` argument, and the work to pick a coupon for a user and course_run should be moved to a different function. this function is doing too many things, and returning `price` and `coupon` is awkward (not to mention unexpected for a function called `calculate_run_price`). lines 121-122 would be changed to something like this:

```python
course_run = get_purchasable_course_run(course_key, user)
coupon = find_course_run_coupon(course_run, user)
price = calculate_run_price(course_run, user, coupon=coupon)
```

ecommerce/api.py
```diff
@@ -426,3 +428,48 @@ def pick_coupons(user):
...
+
+    Returns:
+        (decimal.Decimal, Coupon):
+            The adjusted of the course, and the coupon used if any
+    """
+    program = course_run.course.program
+    enrollment = get_object_or_404(ProgramEnrollment, program=program, user=user)
+    price = get_formatted_course_price(enrollment)['price']
+    coupons = [coupon for coupon in pick_coupons(user) if coupon.program == program]
+    coupon = None
```
gsidebo: you can use `micromasters.utils.first_matching_item` for this logic

```python
coupon = first_matching_item(pick_coupons(user), lambda coupon: coupon.program == program]
```

ecommerce/api_test.py
```diff
@@ -238,32 +245,32 @@ def test_less_or_equal_to_zero(self):
...
             autospec=True,
             return_value=course_run,
         ) as get_purchasable, patch(
-            'ecommerce.api.get_formatted_course_price',
+            'ecommerce.api.calculate_run_price',
             autospec=True,
-            return_value=price_dict,
-        ) as get_price:
+            return_value=price_tuple,
+        ) as _calculate_run_price:
```
gsidebo: general question: why are we patching this function? is it really difficult for us to manually set up the data to make this work? as it is, this is another patch that makes the test more difficult to understand and would fail on a refactor that keeps the exact same behavior but changed implementation

ecommerce/api_test.py
```diff
@@ -289,6 +296,14 @@ def test_create_order(self):
         del dict_before['modified_at']
         assert data_before == dict_before
 
+        if has_coupon:
+            assert RedeemedCoupon.objects.count() == 1
```
gsidebo: you can assert `redeemed_coupon is not None` after you call `.first()` instead of this `.count()` assertion

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+class PriceTests(MockedESTestCase):
+    """
+    Tests for calculating prices
+    """
+
+    def test_calculate_run_price_no_coupons(self):
+        """
+        If there are no coupons for this program the price should be what get_formatted_course_price returned
+        """
+        course_run, user = create_purchasable_course_run()
```
gsidebo: every test case begins with a call to `create_purchasable_course_run`, with the only variance being that some cases discard the `user` return value. can you move this into `setUp` or `setUpTestData` and assign `course_run` and `user` to `self`?

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+        # This coupon is for a different program
+        coupon = CouponFactory.create()
+        UserCoupon.objects.create(coupon=coupon, user=user)
+        discounted_price = 5
+        fa_price = get_formatted_course_price(course_run.course.program.programenrollment_set.first())['price']
+        with patch('ecommerce.api.calculate_coupon_price', autospec=True) as _calculate_coupon_price:
```
gsidebo: we're already asserting below that `calculate_run_price(course_run, user) == (fa_price, None)`, effectively asserting that no coupon was found/used. i don't see the value in patching this function and asserting that it wasn't called

ecommerce/api_test.py
```diff
@@ -719,3 +734,92 @@ def test_not_redeemable(self):
...
+        """
+        course_run, user = create_purchasable_course_run()
+        coupon = CouponFactory.create(content_object=course_run)
+        UserCoupon.objects.create(coupon=coupon, user=user)
+        discounted_price = 5
+        with patch('ecommerce.api.calculate_coupon_price', autospec=True) as _calculate_coupon_price:
+            _calculate_coupon_price.return_value = discounted_price
+            assert calculate_run_price(course_run, user) == (discounted_price, coupon)
+        fa_price = get_formatted_course_price(course_run.course.program.programenrollment_set.first())['price']
+        _calculate_coupon_price.assert_called_with(coupon, fa_price, course_run.edx_course_key)
```
gsidebo: i think it would be better just to create an actual coupon with a fixed discount instead of patching `calculate_coupon_price`. seems pretty straightforward. either way, i think we should get rid of this `assert_called_with`. we've already tested that the function is invoked and returns values we expect. all i see this doing is making it more difficult for us to change implementation

noisecapella:
I think you're looking at the PR I rebased on rather than the changes in just this PR. The commit for this PR is just a one liner

gsidebo:
:man_facepalming:

PR #2404, opened 2017-01-20
Final exam card stuff

### LINKED ISSUE #2407: Exams: Dashboard card with Exam status
The front end and status API portion of #2198

aliceriot:
this closes #2407


dashboard/utils.py
```diff
@@ -277,6 +282,40 @@ def count_courses_passed(self):
...
 
+    def get_pearson_exam_status(self):
+        """
+        Get the pearson exam status for the user / program combo
+        """
+        if self.program.exam_series_code is None:
+            return ""
+        else:
+            user = self.user
+            exam_profile = ExamProfile.objects.filter(profile=user.profile)
```
rhysyngsun: To make this a little cleaner I would probably do:

```python
try:
    exam_profile = ExamProfile.objects.get(profile=user_profile)
except ExamProfile.DoesNotExist:
    return ExamProfile.PROFILE_ABSENT
```

Then you can ditch lines 295-296 and all the `[0]` shenanigans.

dashboard/utils.py
```diff
@@ -277,6 +282,40 @@ def count_courses_passed(self):
...
+                return ExamProfile.PROFILE_ABSENT
+
+            elif (exam_profile[0].status == ExamProfile.PROFILE_PENDING or
+                  exam_profile[0].status == ExamProfile.PROFILE_IN_PROGRESS):
+                return ExamProfile.PROFILE_IN_PROGRESS
+
+            elif exam_profile[0].status == ExamProfile.PROFILE_INVALID:
+                return ExamProfile.PROFILE_INVALID
+
+            elif exam_profile[0].status == ExamProfile.PROFILE_SUCCESS:
```
rhysyngsun: I would probably put in an `else` clause here to be exhaustive. You could log an error in there and default to something sensible (maybe `PROFILE_INVALID`?). Just in case some other value gets in there and this code doesn't get updated.

dashboard/factories_test.py
```diff
@@ -22,6 +23,7 @@ def test_verified_enroll_factory_fa_create(self):  # pylint: disable=no-self-use
         assert Line.objects.count() == 0
         with mute_signals(post_save):
             user = UserFactory.create()
+            ProfileFactory(user=user)
```
rhysyngsun: I would call `.create()` to keep the factory invocation style consistent.

dashboard/utils.py
aliceriot: oh that is nicer!

aliceriot:
@rhysyngsun I just finished writing tests for `get_pearson_exam_status` using `ddt`. I've got a little more JS work to do, so I'm going to get going on that.

dashboard/utils.py
```diff
@@ -277,6 +282,47 @@ def count_courses_passed(self):
...
+        if self.program.exam_series_code is None:
+            return ""
+        else:
+            user = self.user
+            try:
+                exam_profile = ExamProfile.objects.get(profile=user.profile)
+            except ExamProfile.DoesNotExist:
+                return ExamProfile.PROFILE_ABSENT
+
+            if (exam_profile.status == ExamProfile.PROFILE_PENDING or
```
rhysyngsun: Could be simplified to `exam_profiles.status in (ExamProfile.PROFILE_PENDING, ExamProfile.PROFILE_IN_PROGRESS)`

dashboard/utils.py
```diff
@@ -277,6 +282,47 @@ def count_courses_passed(self):
                 passed_courses.add(self.edx_key_course_map[course_id])
         return len(passed_courses)
 
+    def get_pearson_exam_status(self):  # pylint: disable=too-many-return-statements
+        """
+        Get the pearson exam status for the user / program combo
+        """
+        if self.program.exam_series_code is None:
+            return ""
+        else:
```
rhysyngsun: Code style nit: I'd probably drop the `else` so you don't have to indent the majority of the function impl.

dashboard/utils.py
```diff
@@ -277,6 +282,47 @@ def count_courses_passed(self):
                 passed_courses.add(self.edx_key_course_map[course_id])
         return len(passed_courses)
 
+    def get_pearson_exam_status(self):  # pylint: disable=too-many-return-statements
+        """
+        Get the pearson exam status for the user / program combo
+        """
+        if self.program.exam_series_code is None:
```
rhysyngsun: This is one of those cases where this could have an empty value so I think you need to change this is `if not self.program.exam_series_code` so it bails on both `None` and `""`. Also, this needs to bail if `course.exam_module` is empty. Amir wrote some code that encapsulates these conditions [here](https://github.com/mitodl/micromasters/blob/master/exams/utils.py#L24-L31) that could probably be extracted into it's own function. **note: I don't think you want the paid condition in this common function, since Rob's screenshots including prompting the user to pay** If they haven't paid, they won't have an `ExamAuthorization` record, which I believe you covered in logic below.

dashboard/utils.py
aliceriot: oh right, early return make it unnecessary

dashboard/utils.py
aliceriot: OK, I don't quite understand what you mean by the paid condition - how should I separate that out?

rhysyngsun: Line 30 in the linked code, would it be hepful if I addressed this in a small upstream PR?

aliceriot: I've started working on it, but I'm definitely a bit confused!

aliceriot:
@rhysyngsun I think this is ready for review again!

exams/utils.py
```diff
@@ -13,6 +13,19 @@
...
+    """
+    Check if the user is authorized for an exam for a course run
+    """
+    return (
+        course_run and
+        course_run.edx_course_key and
+        course_run.course and
+        course_run.course.exam_module and
+        mmtrack.program.exam_series_code
+    ) is not None
```
aliceriot: gotcha, I'll do the cast to `bool`. Before it was returning strings and stuff which was causing some strange behavior, haha.

aliceriot:
@rhysyngsun comments addressed!

static/js/components/dashboard/FinalExamCard.js
```diff
@@ -0,0 +1,141 @@
...
+      <div className="address-info">
+        <div className="address">
+          <span className="name">
+            { getPreferredName(profile) }
+          </span>
+          <span>
+            { _.get(profile, ['address']) }
+          </span>
+          <span>
+            { getLocation(profile) }
```
rhysyngsun: This doesn't appear to display the zip code in my tests.

static/js/components/dashboard/FinalExamCard.js
```diff
@@ -0,0 +1,141 @@
...
+          </span>
+          <span>
+            Phone: { _.get(profile, ['phone_number']) }
+          </span>
+        </div>
+        { editProfileButton(navigateToProfile) }
+      </div>
+    </div>
+    We will notify you when you become eligible to schedule course exams.
```
rhysyngsun: Nit: the alignment of the left edge of this sentence is off by a few pixels with the box above.

selection_019

static/js/components/dashboard/FinalExamCard.js
```diff
@@ -0,0 +1,141 @@
...
+
+export default class FinalExamCard extends React.Component<void, Props, void> {
+  render () {
+    const {
+      profile,
+      program,
+      navigateToProfile
+    } = this.props;
+
+    switch (program.pearson_exam_status) {
```
rhysyngsun: I was testing all the permutations of `ExamAuthorization.status` in the UI and noticed that in the case it failed it still displays as "We will notify you when you become eligible to schedule course exams." I think in this case we may want to display a message for them to contact support (although we'd already be alerted to it elsewhere). cc @pdpinch

rhysyngsun: Also, just verifying this is the case, I assume you left out a case for a successful `ExamAuthorization` because that will get added  in #2408, correct? If that's the case, I agree with that approach.

static/js/components/dashboard/FinalExamCard_test.js
```diff
@@ -0,0 +1,105 @@
...
+      navigateToProfile: navigateToProfileStub
+    };
+  });
+
+  let stringStrip = R.compose(R.join(" "), _.words);
+
+  let commonText = `You must take a proctored exam for each course. Exams may
+be taken at any Pearson test center. Before you can take an exam, you have to
+pay for the course and pass the online work.`;
+      // <MuiThemeProvider muiTheme={getMuiTheme()}>
```
rhysyngsun: Can this be removed if it's not needed anymore?

aliceriot: oh yes, haha

static/js/components/dashboard/FinalExamCard.js
aliceriot: Yes, that's correct.

static/js/components/dashboard/FinalExamCard.js
aliceriot: Do you mean that the left margin of the address should be the same as the margin between 'Your Pearson Testing...' and the left side of the grey box?

aliceriot:
@rhysyngsun comments addressed

static/js/components/dashboard/FinalExamCard.js
rhysyngsun: I meant visually the sentence below the gray box seems to jut out a bit. Like I said, it's just a nit and a very small visual one at that.

aliceriot: Oh I see, I misunderstood your comment. I'll add a little padding there.

rhysyngsun:
Code looks good, but I'm having difficulties with my local env testing, hopefully get back to you before EOD on that.

Issue #2390, opened 2017-01-19
Google Conversion iframe has added a space at the top of the page


roberthouse54:
The iframe containing the Google Leads conversion frame has added a space at the top of the page (above the micromasters navbar). See screenshot below:

![image](https://cloud.githubusercontent.com/assets/20047260/22118503/12d989dc-de46-11e6-8155-23af5c918465.png)

pdpinch:
@amir-qayyum-khan can you take a look at this?

1. Why does adwords remarketing require an iframe?
2. Can we put it someplace else?

FYI @ChrisLanfranco

pdpinch:
@roberthouse54 do you know if this is on every page? One solution would be to remove the remarketing tag from this particular template. It's an admin view and there's no point in spending ad dollars selling MM to MIT faculty.

roberthouse54:
It seems to be on all pages, including the marketing site. I don't know about the solution.

amir-qayyum-khan:
@pdpinch @roberthouse54  this seem solution 
- http://stackoverflow.com/questions/11939653/google-adwords-remove-iframe-added-by-tracking-conversion-code
- http://stackoverflow.com/questions/18467529/google-remarketing-tag-iframe-height-issue

PR #2280, opened 2017-01-09
Respect DNT request header

singingwolfboy:
The "Do Not Track" request header indicates that a user does not wish to be tracked by the website they are visiting. This commit checks the DNT header to respect this wish. See http://donottrack.us for more information.

Note that this pull request may affect our analytics numbers.

PR #2410, opened 2017-01-23
Removed 13px Adwords iframe height and added Adwords tags to only home page and program page

amir-qayyum-khan:
- fixed https://github.com/mitodl/micromasters/issues/2390

#### What's this PR do?
- Moved Adwords tag to bottom and given white background.
- Limited Adwords tag only to home and program pages.

@pdpinch @roberthouse54 
screen shot 2017-01-23 at 4 21 16 pm

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -41,6 +41,8 @@
     and a new path to a master’s degree from MIT. Learn more ">
 {% endblock %}
 
+{% block adwords %} {% include "adwords.html" %} {% endblock %}
+
 {% block content %}
 
 <div id="home-page">
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -41,8 +41,6 @@
     and a new path to a master’s degree from MIT. Learn more ">
 {% endblock %}
 
-{% block adwords %} {% include "adwords.html" %} {% endblock %}
-
 {% block content %}
 
 <div id="home-page">
@@ -236,6 +234,7 @@
     </main>
   </div>
 
+  {% include "adwords.html" %}
   {% include "footer.html" %}
 
   <div class="modal fade modal-fade-in-scale-up" id="modal-about"
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -26,8 +26,6 @@
   <meta itemprop="image" content="{{ page.background_image.file.url|safe }}">
 {% endblock %}
 
-{% block adwords %} {% include "adwords.html" %} {% endblock %}
-
 {% block body_class %}template-programpage{% endblock %}
 {% block content %}
 {% include "header.html" %}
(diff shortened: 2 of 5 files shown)


roberthouse54:
I am not seeing this in the deployed version (or am I missing something?) Also, I thought you were suggesting leaving it at the top and just adding "display: none"? Did that not work?

roberthouse54:
Also, I just want to check... the goal is to have this on all pages in the app? Or just on the marketing site?

amir-qayyum-khan:
@roberthouse54 
- deployed version? this is PR is not merged yet.
- yes i was, people suggested that that hiding iframe dont change the functionality of adwords. But i thought that they are using it i may have a purpose.

roberthouse54:
I clicked the "View Deployment" link. I then checked the page source and I did not see the code. Maybe I am missing something.

I am just wondering why, instead of hiding the div, you instead moved it to the bottom of the page with a white background.

amir-qayyum-khan:
>I clicked the "View Deployment" link. I then checked the page source and I did not see the code. Maybe I am missing something.

it seems environment variable is not fix for https://micromasters-ci-pr-2410.herokuapp.com/. locally it is working  


>I am just wondering why, instead of hiding the div, you instead moved it to the bottom of the page with a white background.

i did is because of views like these:
- `This may actually stop some browsers loading the tracking iframe, so I'd suggest testing this across all the common browsers before deploying to a live site.`
- `Google do a lot of checking for simulated fake data. I wouldn't be suprised if google filtered out results when the iframe is hidden. There is a reason they put the iframe there.`
- `the problem is I don't know and I wouldn't like to get banned. After some time of digging I found a solution that seems to work (insert script in a hidden div)`

roberthouse54:
OK, I am now able to see this on the deployment. I do not see any programs, so I cannot test the Program page. But I assume the program page would have a white line separating the "Instructors" slideshow and the footer, both of which have dark backgrounds.

And again... just to clarify: this is only on the home page and the program page, correct?

amir-qayyum-khan:
- yes, this is only on home and program

roberthouse54:
Yes, but if you go to an actual micromasters site, you will see that there is an Instructors section at the bottom, which has a black background.

image

amir-qayyum-khan:
@roberthouse54 i made color transparent, so that it will be work on both black and white

roberthouse54:
Can you post a screenshot of how this looks on the program page when there are some faculty in a program?

roberthouse54:
I think a slightly better solution might be to change the background to match the footer background color (#000), as in your current implementation, "transparent" means it will inherit the page background which is white. So on the program page, I think there will be a white stripe between the Instructor section and the footer. As long as this always shows up immediately above the footer, I think a black background would be preferable.

amir-qayyum-khan:
Done @roberthouse54

roberthouse54:
looks good 👍

singingwolfboy:
Since you're changing this code anyway, can you change the template variable `APIKEYS. ADWORDS_CONVERSION_ID` to `APIKEYS.GOOGLE_ADWORDS`, the way I've done in #2280? The `APIKEYS` dictionary should have keys named after the API that is being used.

Issue #2284, opened 2017-01-09
Dashboard: contact course team link


pdpinch:
Depends on #2375 

One feature we'd like to offer MicroMasters students is the ability to quickly and easily contact the course team. 

To start with, we'll offer a link to send mail to the course team from the dashboard. Each course potentially has a unique e-mail. 

- [ ] Buttons appear for all users, one associated with each course
- [ ] If user hasn't paid for the specific course, on click they get an upsell message
- [ ] If user has paid for the course, on click they get compose mail pop-up, similar to learner search
- [ ] model for associating courses with e-mail address(es)
- [ ] modify dashboard course buttons to group "Contact course team" with "View on edX" (could be a ︙menu, or could be a horizontal list of links before the course description)
- [ ] from address should be learner's e-mail address

pdpinch:
Can we use the micromasters_verified in the profile to indicate who is eligible for this?

gsidebo:
@pdpinch I'm pretty positive that the `verified_micromaster_user` property in the Profile doesn't tell us anything about whether or not a student has paid for any courses. I don't really know what it tells us, to be honest. 

Just to confirm, if a user is currently enrolled in Course A and B in some program, and they've paid for A, but not B, they would only be shown the 'contact' link for Course A?

gsidebo:
Seems like some courses will have a course-team-specific email and others won't. Is this a possible scenario? If that's the case, I think it might be a good idea to have a program-wide email address and require that it be set for each program. Courses with no course team email would fall back to that. What do you think?

pdpinch:
I made some changes to the description to try to answer your questions above in https://github.com/mitodl/micromasters/issues/2284#issuecomment-271984487. I agree that `verified_micromaster_user` won't be helpful. 

Both of our current course teams have course-specific e-mail addresses, so I don't think it's worth the engineering effort to have a fall back like you suggest in https://github.com/mitodl/micromasters/issues/2284#issuecomment-271986283

gsidebo:
Sounds good. The email address should be a required field then. I'll try to send some simple test emails via mailgun to make sure it behaves the way we expect with specific 'from' addresses

pdpinch:
I'm ok with making the e-mail address required, but it also seems like you could just omit the feature (per course) if there is no e-mail set. 

Amir already did some work to customize the from address for maligun in #2206

gsidebo:
[...]

PR #2421, opened 2017-01-23
Added course team contact email API endpoint

### LINKED ISSUE #2375: Create API endpoint to contact course team
Supports #2284 

The contact course team link will need to use an API endpoint to send messages to course teams. This issue covers that work. The endpoint on the front and back end should send/accept a course id and a user id.

Acceptance criteria:

1. Model has been changed/created to accommodate course team emails, and future courses are required to specify
1. Some mechanism to set correct course team emails for existing courses in production (can probably be done as part of the migration)
1. API endpoint created and tested in the back end
1. ~~API endpoint created and tested in the front end (a la `static/js/lib/api.js:sendSearchResultMail`)~~ (moving this to a different issue #2420)

gsidebo:

Closes #2375 

#### What's this PR do?

Adds an API endpoint in the backend to send an email to a course team

Probably `mail/api.py`

The following snippet can be run (with 'YOUR_USERNAME' and 'SOME_PROGRAM_TITLE' substituted for your preferred username and program title) to send a course contact email. You should run this with `MAILGUN_RECIPIENT_OVERRIDE` set in your .env file so you receive the email.

```python
from django.test.client import Client
#
user = User.objects.get(username='YOUR_USERNAME')
course = Course.objects.filter(program__title='SOME_PROGRAM_TITLE').first()
course.contact_email = 'courseteam@example.com'
course.save()
url = reverse('course_team_mail_api', kwargs={'course_id': course.id})
client = Client()
client.force_login(user)
resp = client.post(
    url,
    data={'email_subject': 'API Email Subject', 'email_body': 'API Email Body'},
)
resp.json()
```


giocalitri:
Can this be reviewed?

mail/views.py
```diff
@@ -64,11 +67,47 @@ def post(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-
...
+    queryset = Course.objects.all()
+
+    def post(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-self-use
+        """
+        POST method handler
+        """
+        user = request.user
+        course = self.get_object()
+        if not ProgramEnrollment.objects.filter(user=user, program=course.program).exists():
+            raise ValidationError('The user sending the course team message must be enrolled in the given program.')
```
giocalitri: This needs to go inside a custom permission class and should not raise a `ValidationError`

giocalitri: @pdpinch Also I do not think this is enough as a restriction: I think that this could be easily exploitable by someone to create an account and start to spam other users.
We should, for example, restrict it to users who paid for at least a course run or something like this.

pdpinch: The UI is actually limited to paid learners (see [the UI in the epic](https://github.com/mitodl/micromasters/issues/2284#issuecomment-274866731). So it would be appropriate to apply this limitation to the API as well. 

How much more work does this make?

gsidebo: A couple things:

1. You could only spam a course team email using this view, so at the very least it reduces the risk.
1. It shouldn't be too much trouble to add the restriction on the API side. I'll work on that

mail/api.py
```diff
@@ -157,14 +176,45 @@ def send_financial_aid_email(cls, acting_user, financial_aid, subject, body):
...
+        Returns:
+            requests.Response: HTTP Response from Mailgun
+       """
+        if not course.contact_email:
+            message = (
+                'Course team contact email attempted for course without contact_email'
+                '(id: {}, title: "{}")'.format(course.id, course.title)
+            )
+            log.error(message)
+            raise ImproperlyConfigured(message)
```
gsidebo: there are other parts of the codebase that follow this pattern (eg: `financialaid/api.py` lines 40, 111). are you sure that we don't need to log the error explicitly?

giocalitri: it is up to you, but it will end up with a log.error and a log.exception with the same message

mail/api_test.py
```diff
@@ -150,18 +153,43 @@ def test_send_batch_chunk(self, mock_post):
...
 
     @override_settings(MAILGUN_RECIPIENT_OVERRIDE=None)
-    def test_send_individual_email(self, mock_post):
+    @data(None, 'Tester')
+    def test_send_individual_email(self, sender_name, mock_post):
         """
         Test that MailgunClient.send_individual_email() sends an individual message
         """
-        MailgunClient.send_individual_email('email subject', 'email body', 'a@example.com')
+        MailgunClient.send_individual_email('email subject', 'email body', 'a@example.com', sender_name=sender_name)
```
gsidebo: that's just the "to" address. i only changed this line to include `sender_name` in the kwargs since `send_individual_email` now accepts `sender_name`

giocalitri: maybe call the method with the keywords, so it is a bit more clear what's going on

mail/permissions.py
```diff
@@ -18,3 +21,34 @@ def has_permission(self, request, view):
...
+            return False
+        # Make sure the user has paid for any course run for the given course
+        edx_user_data = CachedEdxUserData(user)
+        mmtrack = MMTrack(
+            user,
+            program,
+            edx_user_data
+        )
+        course_run_keys = obj.courserun_set.values_list('edx_course_key', flat=True)
+        return any([mmtrack.has_paid(course_run_key) for course_run_key in course_run_keys])
```
gsidebo: @giocalitri obviously this is doing a bunch of work that isn't needed, but this is the quickest way to check that a user has paid for a course implementation-wise. i'll open different issue to look into cases like these where we want to use specific functionality that exists in `MMTrack`, but we could avoid some of its overhead

giocalitri: ok

giocalitri:
Functionality is fine.

PR #2428, opened 2017-01-24
APIKEYS.GOOGLE_ADWORDS

singingwolfboy:
@amir-qayyum-khan, can we add this commit to #2410?

### DIFF
--- a/ui/context_processors.py
+++ b/ui/context_processors.py
@@ -15,7 +15,7 @@ def api_keys(request):
         "APIKEYS": {
             "GOOGLE": settings.GOOGLE_API_KEY,
             "GOOGLE_ANALYTICS": settings.GA_TRACKING_ID,
+            "GOOGLE_ADWORDS": settings.ADWORDS_CONVERSION_ID,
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
looks good.

ui/context_processors.py
```diff
@@ -15,7 +15,7 @@ def api_keys(request):
         "APIKEYS": {
             "GOOGLE": settings.GOOGLE_API_KEY,
             "GOOGLE_ANALYTICS": settings.GA_TRACKING_ID,
+            "GOOGLE_ADWORDS": settings.ADWORDS_CONVERSION_ID,
```
amir-qayyum-khan: should not we change ``settings.ADWORDS_CONVERSION_ID`` to ``GOOGLE_ADWORDS`` as well?

singingwolfboy: If we do that, we'll have to change how the environment variables are populated, which is a larger change that involves devops. In addition, in the template we access this variable in the `APIKEYS` dict, and that name provides additional context for what `GOOGLE_ADWORDS` means. We don't have namespaces in environment variables.

Issue #2429, opened 2017-01-24
FinancialAidSkipView needs a serializer


singingwolfboy:
Visiting https://micromasters.mit.edu/api/v0/financial_aid_skip/1/ while logged in triggers a 500 exception on production. Reproducing this locally produces an `AssertionError` with the following error message:

> 'FinancialAidSkipView' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.

Traceback:

```
File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py" in inner
  39.             response = get_response(request)

File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py" in _legacy_get_response
  249.             response = self._get_response(request)

File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  217.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  215.                 response = response.render()

File "/usr/local/lib/python3.5/site-packages/django/template/response.py" in render
  109.             self.content = self.rendered_content

File "/usr/local/lib/python3.5/site-packages/rest_framework/response.py" in rendered_content
  72.         ret = renderer.render(self.data, accepted_media_type, context)

File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py" in render
  701.         context = self.get_context(data, accepted_media_type, renderer_context)

File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py" in get_context
  635.         raw_data_put_form = self.get_raw_data_form(data, view, 'PUT', request)

File "/usr/local/lib/python3.5/site-packages/rest_framework/renderers.py" in get_raw_data_form
  548.                     serializer = view.get_serializer(instance=instance)

File "/usr/local/lib/python3.5/site-packages/rest_framework/generics.py" in get_serializer
  109.         serializer_class = self.get_serializer_class()

File "/usr/local/lib/python3.5/site-packages/rest_framework/generics.py" in get_serializer_class
  126.             % self.__class__.__name__

Exception Type: AssertionError at /api/v0/financial_aid_skip/1/
Exception Value: 'FinancialAidSkipView' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.
```

singingwolfboy:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/208583737/

[...]

PR #2430, opened 2017-01-24
Display toast notification for API failure when skipping financial aid

singingwolfboy:
Related to #2429. If the user receives an API failure when trying to skip financial aid, show a toast notification with an error message.

I'm not sure the best way to write tests for this change. It doesn't seem like the `skipFinancialAid` is tested at all.
screen shot 2017-01-24 at 2 41 05 pm

### DIFF
--- a/static/js/actions/financial_aid.js
+++ b/static/js/actions/financial_aid.js
@@ -71,6 +71,7 @@ export const skipFinancialAid = (programId: number): Dispatcher<*> => {
       },
       () => {
         dispatch(receiveSkipFinancialAidFailure());
+        return Promise.reject();
       });
   };
 };
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -306,6 +306,12 @@ class DashboardPage extends React.Component {
     const { dispatch } = this.props;
     dispatch(skipFinancialAid(programId)).then(() => {
       this.setConfirmSkipDialogVisibility(false);
+    }).catch(() => {
+      this.setConfirmSkipDialogVisibility(false);
+      dispatch(setToastMessage({
+        message: "Failed to skip financial aid.",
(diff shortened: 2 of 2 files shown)


pdpinch:
You say:

> It doesn't seem like the skipFinancialAid is tested at all.

Do you mean in terms of an integration test?

singingwolfboy:
`skipFinancialAid` is defined in `js/containers/DashboardPage.js`, and passed to the `FinancialAidCard` component. This component has tests, but it doesn't actually test the `skipFinancialAid` function, since it's not defined in that component. This _should_ be tested in `js/containers/DashboardPage_test.js`, but there's no test in that file about what should happen when that function is called.

aliceriot:
@singingwolfboy have you been able to test and confirm that this is working? for me, locally, I was unable to get it to work without making this change:

```
diff --git a/static/js/actions/financial_aid.js b/static/js/actions/financial_aid.js
index dc438418..e2c268fd 100644
--- a/static/js/actions/financial_aid.js
+++ b/static/js/actions/financial_aid.js
@@ -71,6 +71,7 @@ export const skipFinancialAid = (programId: number): Dispatcher<*> => {
       },
       () => {
         dispatch(receiveSkipFinancialAidFailure());
+        return Promise.reject();
       });
   };
 };
```

aliceriot:
( it's easy to test by changing the return value of the `skipFinancialAid` function in `lib/api.js` to return `Promise.reject()` )

singingwolfboy:
Oh, I must have forgotten to commit that change! I recall making it when I was testing it -- I don't know why it didn't make it into the commit. Thanks for testing! I've made that change, and rebased on top of the latest `master`.

aliceriot:
awesome, taking a look now

aliceriot:
(assuming tests pass and so on)

PR #2434, opened 2017-01-25
Toast role alert

singingwolfboy:
Set `role="alert"` on toast notifications. This will cause screen readers to read them first when the page is loaded. See http://rawgit.com/w3c/aria/master/aria/aria.html#alert for more info.

### DIFF
--- a/static/js/components/Toast.js
+++ b/static/js/components/Toast.js
@@ -30,7 +30,7 @@ export default class Toast extends React.Component {
   render() {
     const { children, open } = this.props;
 
-    return <div className={`toast ${open ? 'open' : ''}`}>
+    return <div role="alert" className={`toast ${open ? 'open' : ''}`}>
       {children}
     </div>;
   }


static/js/components/Toast.js
```diff
@@ -30,7 +30,7 @@ export default class Toast extends React.Component {
   render() {
     const { children, open } = this.props;
 
-    return <div className={`toast ${open ? 'open' : ''}`}>
+    return <div role="alert" className={`toast ${open ? 'open' : ''}`}>
```
aliceriot: would it be a good idea to only put the `alert` role on if the toast is currently active? similarly to what we do with the `className` for it.

singingwolfboy: Oh, good point! However, wouldn't it be even better not to render the `<Toast>` component at all if it's not active?

aliceriot: That sounds reasonable to me, we could move the toast rendering function into a helper method and have the `render` function just do something like

```js
return open ? this.toast() : null;
```

which would take care of it.

Wondering what @noisecapella thinks about this.

noisecapella: Fine with me

singingwolfboy:
Closing in favor of #2437.

PR #2437, opened 2017-01-25
Don't render closed Toast

singingwolfboy:
Builds off of #2434. When there isn't a `toastMessage` in the Redux state, don't render the `<Toast>` component at all. This pull request also includes the `role="alert"` change from #2434.

This pull request should have no user-visible impact at all. Make sure that you can still get toast notifications through the app.

### DIFF
--- a/static/js/components/Toast.js
+++ b/static/js/components/Toast.js
@@ -19,13 +19,6 @@ export default class Toast extends React.Component {
     }
   }
 
-  componentDidUpdate() {
-    const { onTimeout, timeout } = this.props;
-    if (onTimeout) {
-      setTimeout(onTimeout, timeout);
-    }
-  }
-
   render() {
     const { children } = this.props;
 
--- a/static/js/components/Toast.js
+++ b/static/js/components/Toast.js
@@ -3,7 +3,6 @@ import React from 'react';
 export default class Toast extends React.Component {
   props: {
     children: any,
-    open: boolean,
     timeout: number,
     onTimeout: () => void,
   };
@@ -13,24 +12,24 @@ export default class Toast extends React.Component {
   };
 
   componentDidMount() {
-    const { onTimeout, open, timeout } = this.props;
+    const { onTimeout, timeout } = this.props;
 
-    if (open && onTimeout) {
+    if (onTimeout) {
       setTimeout(onTimeout, timeout);
     }
   }
 
-  componentDidUpdate(prevProps: Object) {
+  componentDidUpdate() {
     const { onTimeout, timeout } = this.props;
-    if (!prevProps.open && this.props.open && onTimeout) {
+    if (onTimeout) {
       setTimeout(onTimeout, timeout);
     }
   }
 
   render() {
-    const { children, open } = this.props;
+    const { children } = this.props;
 
-    return <div className={`toast ${open ? 'open' : ''}`}>
+    return <div role="alert" className="toast">
       {children}
     </div>;
   }
--- a/static/js/constants.js
+++ b/static/js/constants.js
@@ -88,8 +88,8 @@ export const FA_PENDING_STATUSES = [
 ];
 export const FA_APPROVED_STATUSES = [FA_STATUS_APPROVED, FA_STATUS_AUTO_APPROVED];
 
-export const TOAST_SUCCESS = 'success';
-export const TOAST_FAILURE = 'failure';
+export const TOAST_SUCCESS = 'done';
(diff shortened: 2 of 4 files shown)


aliceriot:
why not just pull these changes into your other PR? (or close that one).

singingwolfboy:
@aliceriot I made a new pull request because this has become a larger change with a different focus. I closed #2434.

aliceriot:
ok cool

aliceriot:
hm, I was thinking we would want to keep the code in `App.js` as-is and just make a few small changes to the Toast component - so that `<App />` doesn't have to have logic around when to write the component to the DOM and when not to, and just passes props to the `<Toast />`.

We could just move the existing (on master) code to return the toast into a little helper method, and then use a ternary to either call that method or return null.

something like this:

```js
// Toast.js
class Toast ... {
  
  renderAlert () {
    const { children } = this.props;
    return <div role="alert" className="toast">
      { children }
    </div>;
  }
  
  render () {
    const { open } = this.props;
    return open ? this.renderAlert() : null;
  }
}
```

Actually, what we should really do is refactor the `<Toast />` to take the contents of the `toastMessage` object (from `state.ui`) and take care of the rest from there (moving that logic for picking the icon and so on out of `<App.js />`).

what do you think? it strikes me as cleaner overall to have `<App />` just pass down props and not worry about the display logic of its children.

singingwolfboy:
@aliceriot Your suggestions sound reasonable to me -- but the scope of this change is growing larger and larger. Maybe we shouldn't try to fix what isn't broken? The original impetus for making this change was to make the toast notifications more accessible for screen readers, and I don't want to stray _too_ far away from that.

That being said, if you want to make a pull request with the change you're suggesting, I'd be happy to review it. :grin:

aliceriot:
Haha, ok. Maybe we should just re-open your other PR to add the role and circle back on this later?

static/js/components/Toast.js
```diff
@@ -13,24 +12,24 @@ export default class Toast extends React.Component {
...
     }
   }
 
-  componentDidUpdate(prevProps: Object) {
+  componentDidUpdate() {
     const { onTimeout, timeout } = this.props;
-    if (!prevProps.open && this.props.open && onTimeout) {
+    if (onTimeout) {
       setTimeout(onTimeout, timeout);
     }
```
aliceriot: could this change result in `setTimeout` being called multiple times?

static/js/constants.js
```diff
@@ -88,8 +88,8 @@ export const FA_PENDING_STATUSES = [
 ];
 export const FA_APPROVED_STATUSES = [FA_STATUS_APPROVED, FA_STATUS_AUTO_APPROVED];
 
-export const TOAST_SUCCESS = 'success';
-export const TOAST_FAILURE = 'failure';
+export const TOAST_SUCCESS = 'done';
+export const TOAST_FAILURE = 'error';
```
aliceriot: is this changed to coincide with the `name` prop for the `Icon` below?

aliceriot:
we're going to go ahead and merge this change, and come back at some point to refactor it more fully.

static/js/constants.js
singingwolfboy: Correct. The only thing that the old code was doing with these constants was comparing them, like this:

```js
if (toastMessage.icon === TOAST_FAILURE) {
  icon = <Icon name="error" key="icon "/>;
} else if (toastMessage.icon === TOAST_SUCCESS) {
  icon = <Icon name="done" key="icon" />;
}
```

So I figure, might as well use `done` and `error` directly.

static/js/components/Toast.js
singingwolfboy: I don't believe so. Any case in the past where `open` was `false`, now this component simply won't be rendered, so the code won't be called at all. But if you can think of a way that `setTimeout` might be called multiple times, let me know.

static/js/constants.js
aliceriot: makes sense! haha

static/js/components/Toast.js
aliceriot: Well, my suspicion is that we could run into an edge case where componentDidUpdate() is called multiple times - I don't know how much of a concern that would be. But, in any case, I think because we are now render the component on-demand we can just lean on componentDidMount and delete the componentDidUpdate method.

singingwolfboy: OK, I've deleted the `componentDidUpdate()` method. Can you review this again?

PR #2442, opened 2017-01-26
Add unique constraint on coupon code

### LINKED ISSUE #2441: Coupons: coupon code should have a unique constraint
Coupons do not require a coupon code, but there should be only one coupon with a given coupon code in the database

noisecapella:
Fixes #2441 

#### What's this PR do?
Adds a unique constraint on coupon code. Note that a unique constraint doesn't apply to null values


giocalitri:
Just a question: how we'll generate coupon codes to be sure that they are unique?

Anyway 👍 as soon as #2401 is merged

noisecapella:
Currently the user has to use the Django Admin. At some point we would create a script to generate them, maybe a few one time scripts

giocalitri:
if I am not wrong, this can be merged

noisecapella:
Yes, I missed the :+1:

PR #2443, opened 2017-01-26
hide financial aid card if 100% coupon is present

### LINKED ISSUE #2334: Coupons: hide financial aid options if user has a 100% off coupon for whole program
Part of #2158 

For 100% off coupons for a program we need to:
- [ ] change the user-coupon attachment code to additionally skip financial aid
- [ ] Change the UI to omit whatever messaging we have that the user skipped financial aid
- [ ] need a way to skip the financial aid step on the user's behalf

aliceriot:

this closes #2334 

#### What's this PR do?

This checks to see if a user has a 100% off coupon before we show the financial aid card. If they do, we dont show it, and we automatically submit a request to 'skip' financial aid (provided they do not already have a financial aid object).

I think you want to:

1. confirm that without adding a coupon the financial aid card still appears.

2. add a 100% discount coupon to your user and ensure that your user does not currently have a financial aid.

3. visit `/dashboard`.

4. confirm that a request to skip financial aid was issued, and that the financial aid card UI does not show up.

I think that's it!

### DIFF
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -6,6 +6,7 @@ import { connect } from 'react-redux';
 import Loader from '../components/Loader';
 import _ from 'lodash';
 import moment from 'moment';
+import R from 'ramda';
 
 import { calculatePrices } from '../lib/coupon';
 import {
@@ -19,6 +20,8 @@ import {
 } from '../actions';
 import {
   COUPON_CONTENT_TYPE_COURSE,
+  COUPON_CONTENT_TYPE_PROGRAM,
+  COUPON_AMOUNT_TYPE_PERCENT_DISCOUNT,
   TOAST_SUCCESS,
   TOAST_FAILURE,
   STATUS_OFFERED,
@@ -105,6 +108,11 @@ class DashboardPage extends React.Component {
 
   componentDidUpdate() {
     this.updateRequirements();
+
+    let program = this.getCurrentlyEnrolledProgram();
+    if ( this.shouldSkipFinancialAid() && program !== undefined ) {
+      this.skipFinancialAid(program.id);
+    }
   }
 
   componentWillUnmount() {
@@ -303,17 +311,40 @@ class DashboardPage extends React.Component {
     dispatch(setDocumentSentDate(newDate));
   };
 
+  getCurrentlyEnrolledProgram = () => {
+    const { currentProgramEnrollment, dashboard } = this.props;
+    if (
+      currentProgramEnrollment !== null &&
+      dashboard !== null
+    ) {
+      return dashboard.programs.find(program => (
+        program.id === currentProgramEnrollment.id
+      ));
+    } else {
+      return undefined;
+    }
+  }
+
   skipFinancialAid = programId => {
-    const { dispatch } = this.props;
-    dispatch(skipFinancialAid(programId)).then(() => {
-      this.setConfirmSkipDialogVisibility(false);
-    }).catch(() => {
-      this.setConfirmSkipDialogVisibility(false);
-      dispatch(setToastMessage({
-        message: "Failed to skip financial aid.",
-        icon: TOAST_FAILURE,
-      }));
-    });
+    const { dispatch, financialAid } = this.props;
+
+    let program = this.getCurrentlyEnrolledProgram();
+    if (
+      program &&
+      program.financial_aid_user_info &&
+      program.financial_aid_user_info.has_user_applied === false &&
+      financialAid.fetchSkipStatus === undefined
+    ) {
+      dispatch(skipFinancialAid(programId)).then(() => {
+        this.setConfirmSkipDialogVisibility(false);
+      }).catch(() => {
+        this.setConfirmSkipDialogVisibility(false);
+        dispatch(setToastMessage({
+          message: "Failed to skip financial aid.",
+          icon: TOAST_FAILURE,
+        }));
+      });
+    }
   };
 
   setConfirmSkipDialogVisibility = bool => {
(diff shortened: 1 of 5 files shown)


noisecapella:
It looks like there's an issue with `setState` somewhere according to the error output

aliceriot:
yeah, I'm not sure where it's coming from

static/js/containers/DashboardPage.js
```diff
@@ -265,11 +265,28 @@ class DashboardPage extends React.Component {
     dispatch(setDocumentSentDate(newDate));
   };
 
+  getCurrentlyEnrolledProgram = () => {
+    const { currentProgramEnrollment, dashboard } = this.props;
+
+    return dashboard.programs.find(program => (
+      program.id === currentProgramEnrollment.id
```
noisecapella: Is `currentProgramEnrollment` ever undefined?

static/js/containers/DashboardPage.js
```diff
@@ -325,24 +343,29 @@ class DashboardPage extends React.Component {
...
-          setDocsInstructionsVisibility={this.setDocsInstructionsVisibility}
-          ui={ui}
-          financialAid={financialAid}
-        />;
-      }
 
       const calculatedPrices = calculatePrices(dashboard.programs, prices.coursePrices, coupons.coupons);
 
+      if ( program.financial_aid_availability ) {
+        if ( calculatedPrices.get(program.id) === 0 ) {
```
noisecapella: This is should be the run id... It would be nice if flow had the ability to tell them apart

aliceriot: which run would I use for that?

noisecapella: Hmm... we should probably just get the coupon being used and check that it's a 100% off coupon

aliceriot: OK I can work on doing that. If I'm understanding correctly I need to check two things, that the coupon is a program-level coupon and that the coupon is a 100% coupon. If it's not a program coupon we would still want to show financial aid, right?

noisecapella: Yes, if the 100% coupon is only for one course then the other course runs will still have a non-zero price and therefore financial aid might be helpful.

Alternatively we can check `calculatedPrices.get(run.id) == 0` for every run in the current program. What do you think?

pdpinch:
Does this behavior apply to 100% off course coupons, or just 100% off program coupons?

aliceriot:
@pdpinch George and I discussed this a bit above

static/js/containers/DashboardPage.js
aliceriot: hm, I think the latter might be a better solution - if someone has a 100% for each course we wouldn't want to show the financial aid card, right?

noisecapella: Yes I agree

aliceriot:
@noisecapella I figured out that `setState` issue! I was kicking off an method that could issue a request (the one to skip the financial aid) from the `render` method of `DashboardPage`. This is bad because then `render` has side-effects, and is no longer a pure function of props and state. This can in certain situations lead to infinite loops and the like.

Anyway, I pulled the functionality out into a few helper methods, and we now perform the same check in `componentDidUpdate`.

aliceriot:
@noisecapella this should be ready for review again

static/js/containers/DashboardPage.js
```diff
@@ -371,6 +397,22 @@ class DashboardPage extends React.Component {
...
+      prices,
+      coupons,
+    } = this.props;
+
+    return calculatePrices(dashboard.programs, prices.coursePrices, coupons.coupons);
+  };
+
+  allPricesZero = () => {
+    let allPricesZero = true;
+    this.calculatedPrices().forEach(price => allPricesZero = allPricesZero && price === 0);
```
noisecapella: I think you need to filter on program here, to make sure you're only considering the currently selected program

aliceriot: hm, what if I changed the `calculatedPrices` function to return only the prices for the selected program?

noisecapella: Do you mean `this.calculatedPrices()`? That would be fine with me

aliceriot: yeah, exactly. ok I'll do that.

noisecapella: Also, it seems like `R.all` would avoid the need for an explicit `allPricesZero` variable here

aliceriot:
@noisecapella tests fixed, should be ready for review again

noisecapella:
Can you rebase?

static/js/containers/DashboardPage.js
```diff
@@ -382,6 +411,17 @@ class DashboardPage extends React.Component {
     this.context.router.push("/learner");
   };
 
+  shouldSkipFinancialAid = (): boolean => {
+    return R.compose(
+      R.any(coupon => (
+        coupon.content_type === COUPON_CONTENT_TYPE_PROGRAM &&
```
noisecapella: I think you also need to make sure that `coupon.program_id` equals the currently selected program id

aliceriot: oh yeah haha

aliceriot: I added a check for that

noisecapella:
Can you add or adjust a test to check that coupons for other programs are ignored?

static/js/containers/DashboardPage_test.js
```diff
@@ -203,27 +209,51 @@ describe('DashboardPage', () => {
...
-      return renderComponent('/dashboard', DASHBOARD_SUCCESS_ACTIONS).then(([wrapper]) => {
-        assert.include(wrapper.text(), `Pay Now $123`);
+        return renderComponent('/dashboard', DASHBOARD_SUCCESS_ACTIONS).then(([wrapper]) => {
+          assert.include(wrapper.text(), `Pay Now $123`);
+          assert.isTrue(calculatePricesStub.calledWith(dashboard, coursePrices, []));
+        });
+      });
 
-        assert.isTrue(calculatePricesStub.calledWith(dashboard, coursePrices, []));
+      it('should issue a request to skip if all the course prices are 0', () => {
```
noisecapella: Can we add a test that verifies that we don't try to skip if `has_user_applied` is already true?

aliceriot:
@noisecapella I added some more tests!

aliceriot:
@noisecapella this PR is for issue 2334 and the PR number is 2443! Whoa!

PR #2444, opened 2017-01-26
Freeze grade sync in case the course run has already frozen grades.

### LINKED ISSUE #2391: Final Grades: Freeze synchronously the final grade
Freeze synchronously the final grade if the course has completed the final grade freeze and the user visits the dashboard.

The reason is that some user could log in for the first time after the course has been already frozen.

giocalitri:
closes #2391 

#### What's this PR do?
If the course run has already final grades and the user does not have one for the course run, a final grade freeze is performed synchronously for the user in the course run.

`dashboard/api.py`

The manual test is pretty complicated. 

Set in your `.env` the variable `FEATURE_FINAL_GRADE_ALGORITHM=v1` 

To really test this, you should create 2 real courses on the edX instance with a real enrollment for your user and connect one to a course on a Financial Aid program and the other to a non Financial Aid program. 
Remember to create a verified enrollment for the non Financial Aid Program.

The reason why you need to create a real course is because the freezing of a grade will trigger a refresh of the cache even if the cache is fresh, so there is no way to avoid the request to edX.

Once this is set, there are 2 parts to test.

For the first part, use the Financial Aid Program. 
Set the course run to be past, but with the upgrade deadline in the future.
Then create an entry in `grades.models.CourseRunGradingStatus` with `course_run=<your_course_run>` and `status='complete'` to simulate the status of the grades already frozen.
Then refresh the dashboard. The status for the course run should be "Failed" (unless the current grade on edX is `>>0`), but now you should have an entry in `grades.models.FinalGrade` for your user in the course_run.

For the second part, use the non Financial Aid Program. 
Set the course run to be past. The upgrade deadline is not important because the enrollment is already verified on edX (so the user paid).
Then create an entry in `grades.models.CourseRunGradingStatus` with `course_run=<your_course_run>` and `status='complete'` to simulate the status of the grades already frozen.
Then refresh the dashboard. The status for the course run should be "Failed" (the user does not have a certificate), but now you should have an entry in `grades.models.FinalGrade` for your user in the course_run.

**NOTE**: there is a way to avoid creating courses on edX and it is commenting [this line](https://github.com/mitodl/micromasters/blob/caa9081f18d73fd5a09f122c29adcf2bdef9dc1e/grades/api.py#L147) and put a dumb statement: it is where the refresh of the cache is called. 
If you do this then you can set everything locally on MM.

### DIFF
--- a/dashboard/api_test.py
+++ b/dashboard/api_test.py
@@ -29,13 +29,14 @@ from dashboard.api_edx_cache import CachedEdxDataApi
 from dashboard.factories import CachedEnrollmentFactory, CachedCurrentGradeFactory, UserCacheRefreshTimeFactory
 from dashboard.utils import MMTrack
 from grades.constants import FinalGradeStatus
+from grades.exceptions import FreezeGradeFailedException
 from grades.models import FinalGrade, CourseRunGradingStatus
 from micromasters.factories import UserFactory
 from micromasters.utils import is_subset_dict
 from search.base import MockedESTestCase
 
 
-# pylint: disable=too-many-lines
+# pylint: disable=too-many-lines, too-many-arguments
 class StatusTest(MockedESTestCase):
     """
     Tests for the different status classes
@@ -292,12 +293,16 @@ class CourseRunTest(CourseTests):
         (False, api.CourseRunStatus.CHECK_IF_PASSED, 'v0')
     )
     @ddt.unpack
-    def test_check_if_passed(self, has_frozen_grades, expected_status, grade_algo):
-        """test for get_status_for_courserun for a finished course if enrolled"""
+    def test_check_if_passed_if_has_frozen_grade(self, has_frozen_grades, expected_status, grade_algo):
+        """
+        test for get_status_for_courserun for a finished course if enrolled
+        and in case the user has a final grade (if the grade_algorithm is v1)
+        """
         self.mmtrack.configure_mock(**{
             'is_enrolled.return_value': True,
             'is_enrolled_mmtrack.return_value': True,
-            'has_paid.return_value': True
+            'has_paid.return_value': True,
+            'extract_final_grade.return_value': Mock()
         })
         # create a run that is past
         crun = self.create_run(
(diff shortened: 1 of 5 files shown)


dashboard/api_test.py
```diff
@@ -466,23 +505,53 @@ def test_not_paid_in_past_grade_frozen(self, froz_mock, upgr_mock):
...
+        # case when the user has not an entry in the frozen grades the code tries to create it
         self.mmtrack.extract_final_grade.side_effect = FinalGrade.DoesNotExist
+        # in case the created frozen grade that says the user passed
+        freeze_mock.return_value = Mock(passed=True)
         run_status = api.get_status_for_courserun(crun, self.mmtrack)
-        assert run_status.status == api.CourseRunStatus.CURRENTLY_ENROLLED
-        # just resetting the mock does not remove the side effect
-        self.mmtrack.extract_final_grade.side_effect = None
+        assert run_status.status == api.CourseRunStatus.CAN_UPGRADE
+        # in case the created frozen grade that says the user did not pass
```
singingwolfboy: Each of these "in case" comments sounds like they're testing something different. If that's true, then they should each be a separate test function.

grades/api_test.py
```diff
@@ -200,43 +203,64 @@ def test_get_users_without_frozen_final_grade(self):
...
+            assert final_grade is None
+        else:
+            with self.assertRaises(FreezeGradeFailedException):
+                api.freeze_user_final_grade(
+                    self.user, self.run_no_fa, raise_on_exception=raise_on_exception)
         assert mock_refr.called is False
         assert mock_get_fg.called is False
         assert FinalGrade.objects.filter(user=self.user, course_run=self.run_no_fa).exists() is False
 
         # case without freeze date
```
singingwolfboy: This also looks like it should be a separate test.

dashboard/api_test.py
```diff
@@ -466,23 +505,53 @@ def test_not_paid_in_past_grade_frozen(self, froz_mock, upgr_mock):
...
+        # create a run that is past
+        crun = self.create_run(
+            start=self.now-timedelta(weeks=52),
+            end=self.now-timedelta(weeks=45),
+            enr_start=self.now-timedelta(weeks=62),
+            enr_end=self.now-timedelta(weeks=53),
+            edx_key="course-v1:MITx+8.MechCX+2014_T1"
+        )
 
         # case in a final grade with status passed=True is returned
```
singingwolfboy: These cases should be separate tests.

singingwolfboy:
I manually tested this pull request, with @giocalitri's help. (He's right, it was very complicated!) It seems to work as expected, although we discovered that Celery was throwing an exception that sounds like it _might_ be related to this work: https://gist.github.com/singingwolfboy/c54001c12249120e1f03c4a7bda9fe05 It also might be completely independent, though. @giocalitri will investigate.

giocalitri:
I looked into the problem and I think it is unrelated from this PR.
I will work on that next

Issue #2420, opened 2017-01-23
Create API endpoint to contact course team (front-end)


gsidebo:
Supports #2284 
Created as a separate issue, previously part of #2375

Acceptance criteria: Course team contact API endpoint created and tested in the front end (a la `static/js/lib/api.js:sendSearchResultMail`)

PR #2446, opened 2017-01-26
Changed front-end handling of emails

gsidebo:

No specific ticket - it is one of two issues that will address #2420 
⚠️️ **REBASED ON #2421** ⚠️️

#### What's this PR do?

Restructures the email portion of the app state to accommodate multiple email types

static/js/containers/LearnerSearchPage.js

With `MAILGUN_RECIPIENT_OVERRIDE` set to your email address, send a few emails from the /learners page using the "Email These Learners" button. Try a couple different filters and make sure the email you receive contains the right recipients. Also check for normal behavior of the email dialog

This doesn't change any functionality. I put this up as a separate PR due to the amount of code it changes

### DIFF
--- a/static/js/actions/email.js
+++ b/static/js/actions/email.js
@@ -3,7 +3,12 @@ import type { Dispatch } from 'redux';
 import { createAction } from 'redux-actions';
 
 import type { Dispatcher } from '../flow/reduxTypes';
-import type { EmailSendResponse } from '../flow/emailTypes';
+import type {
+  EmailSendResponse,
+} from '../flow/emailTypes';
+import {
+  SEARCH_EMAIL_TYPE,
+} from '../components/email/constants';
 import * as api from '../lib/api';
 
 export const START_EMAIL_EDIT = 'START_EMAIL_EDIT';
@@ -33,13 +38,13 @@ export function sendSearchResultMail(
   searchRequest: Object
 ): Dispatcher<EmailSendResponse> {
   return (dispatch: Dispatch) => {
-    dispatch(initiateSendEmail());
+    dispatch(initiateSendEmail(SEARCH_EMAIL_TYPE));
     return api.sendSearchResultMail(subject, body, searchRequest).
       then(response => {
-        dispatch(sendEmailSuccess());
+        dispatch(sendEmailSuccess(SEARCH_EMAIL_TYPE));
         return Promise.resolve(response);
       }).catch(error => {
-        dispatch(sendEmailFailure(error));
+        dispatch(sendEmailFailure({type: SEARCH_EMAIL_TYPE, error: error}));
       });
   };
 }
--- a/static/js/components/email/constants.js
+++ b/static/js/components/email/constants.js
@@ -1,0 +1,2 @@
+export const SEARCH_EMAIL_TYPE = 'searchResultEmail';
+export const COURSE_EMAIL_TYPE = 'courseTeamEmail';
--- a/static/js/reducers/email.js
+++ b/static/js/reducers/email.js
@@ -1,4 +1,5 @@
 // @flow
+import _ from 'lodash';
 import {
   START_EMAIL_EDIT,
   UPDATE_EMAIL_EDIT,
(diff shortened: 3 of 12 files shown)


gsidebo:
@aliceriot beware of the rebase! i think this PR only changes JS files

aliceriot:
haha, I saw the notice! thanks for making sure though :D

aliceriot:
functionality looks good! looking at code

static/js/reducers/email.js
```diff
@@ -15,38 +16,57 @@ import {
...
 };
 
-export const email = (state: EmailState = INITIAL_EMAIL_STATE, action: Action) => {
+function updatedState(state: AllEmailsState, emailType: string, updateObject: Object) {
+  _.merge(state[emailType], updateObject);
+  return _.cloneDeep(state);
+}
+
+export const email = (state: AllEmailsState = INITIAL_ALL_EMAILS_STATE, action: Action) => {
+  let emailType;
```
aliceriot: what's the reasoning behind using this variable? seems like it's just adding boilerplate.

static/js/reducers/email.js
```diff
@@ -15,38 +16,57 @@ import {
...
+
   switch (action.type) {
   case START_EMAIL_EDIT:
-    return { ...state, email: { ...NEW_EMAIL_EDIT, query: action.payload } };
+    emailType = action.payload;
```
aliceriot: wouldn't `action.payload` have type `Action`, and not type `string`? Maybe I'm missing something here.

aliceriot: oh nvm I see, for some actions `action.payload` is the type and for some it's `action.payload.type`.

static/js/reducers/email.js
```diff
@@ -15,38 +16,57 @@ import {
...
+    return updatedState(state, emailType, { inputs: NEW_EMAIL_EDIT });
   case UPDATE_EMAIL_EDIT:
-    return { ...state, email: action.payload };
+    emailType = action.payload.type;
+    return updatedState(state, emailType, { inputs: action.payload.inputs });
```
aliceriot: I think it might be clearer to just be explicit, using `action.payload.type` here instead of `emailType`.

aliceriot: Or! You could write a little wrapper function:

```js
const getType = payload => (
  typeof payload === "string" ? payload : payload.type
);
```

and then call your update function like this:

```js
return updatedState(state, getType(action.payload), { inputs: action.payload.inputs });
```

static/js/reducers/email.js
```diff
@@ -15,38 +16,57 @@ import {
...
+export const NEW_EMAIL_EDIT: EmailInputs = {
   subject:    null,
   body:       null,
-  query:      null,
 };
 
-export const email = (state: EmailState = INITIAL_EMAIL_STATE, action: Action) => {
+function updatedState(state: AllEmailsState, emailType: string, updateObject: Object) {
+  _.merge(state[emailType], updateObject);
+  return _.cloneDeep(state);
```
aliceriot: Just tried it out, and I see why you're not doing it! It seems like `_.merge` does a deep merge (which we want) but it mutates the first argument (which is yucky).

I think we generally do not want to mutate the previous state - it might be better if we do the `_.cloneDeep` of the previous state first, so something like:

```js
return _.merge(_.cloneDeep(state), {[emailType]: updateObject});
```

static/js/components/email/constants.js
```diff
@@ -0,0 +1,2 @@
+export const SEARCH_EMAIL_TYPE = 'searchResultEmail';
+export const COURSE_EMAIL_TYPE = 'courseTeamEmail';
```
aliceriot: Generally we do constants like `SEARCH_RESULT_EMAIL`, is there a strong reason for camelCase here?

static/js/reducers/email.js
gsidebo: that works! my intent with setting the variable was just to make it clear to someone reading the code that the type could be found in different parts of the payload depending on the action. that helper function does the job a little better

aliceriot: cool, sorry for leaving all the comments as I worked through the code! haha

static/js/actions/email.js
```diff
@@ -33,13 +38,13 @@ export function sendSearchResultMail(
...
-    dispatch(initiateSendEmail());
+    dispatch(initiateSendEmail(SEARCH_EMAIL_TYPE));
     return api.sendSearchResultMail(subject, body, searchRequest).
       then(response => {
-        dispatch(sendEmailSuccess());
+        dispatch(sendEmailSuccess(SEARCH_EMAIL_TYPE));
         return Promise.resolve(response);
       }).catch(error => {
-        dispatch(sendEmailFailure(error));
+        dispatch(sendEmailFailure({type: SEARCH_EMAIL_TYPE, error: error}));
```
gsidebo: yeah. when we implement something that allows the containers to reuse these functions, the `emailType` would be specified ahead of time and would replace this reference

static/js/components/email/constants.js
gsidebo: it's a string constant, so i think it's appropriate to represent it with caps+snake case (we also do this in `static/js/constants.js`)

static/js/reducers/email.js
gsidebo: actually, `return _.merge(_.cloneDeep(state), {[emailType]: updateObject});` would replace `state['searchResultEmal']` entirely instead of merging some limited set of properties into it. let me know what you think of the alternative i pushed

static/js/components/email/constants.js
aliceriot: makes sense

static/js/reducers/email.js
aliceriot: Are you sure? In my tests (just in the `nodejs` repl) it seems to be working

aliceriot: I did:

```
var _ = require('lodash')
let test = { inputType: { foo: 'bar', biz: 'boz'} }
_.merge(_.cloneDeep(test), {['inputType']: { foo: 'baz' }})
```

the `_.merge` line evaluated to:

```
{ inputType: { foo: 'baz', biz: 'boz' } };
```

so it seems like the `{ foo: 'baz' }` object was correctly merged with the `{ foo: 'bar', biz: 'boz' }` object.

aliceriot: I'm fine with what you have btw, in the interest of merging soon

PR #2422, opened 2017-01-24
Validated exam authorizations were accepted by Pearson 

amir-qayyum-khan:
#### What are the relevant tickets?
fixes https://github.com/mitodl/micromasters/issues/2080

#### What's this PR do?
- A task that open ftp and process Exam Authorization Confirmation files (EAC) files.
- In case of success or failure of ``Exam Authorization`` it updates status of ExamAuthorization object.

#### Note:
- Flag need to set is ```EXAMS_SFTP_TEMP_DIR=(to a dir in you local machine)```

@pdpinch

PR #2447, opened 2017-01-26
Pearson response processing

### LINKED ISSUE #1797: Exams: Validate the user profiles were accepted by Pearson

rhysyngsun:
_**Before starting review on this, please message me so I can share the documentation with you**_

---

Fixes #1797 and #2422

#### What's this PR do?

This PR primarily adds processing of two types of Pearson result files. Each row in these files is a result of a corresponding row in a file we uploaded. The results are asynchronous and since Pearson only processes our uploads every 12 hours it can be a while between upload and results. You can think of this as RPC-over-SFTP. Pearson places these files on their SFTP server so we have to periodically poll and check for new ones. They also need to be removed once we've processed them.

The two types of files we're immediately interested in:

- VCDC - results to CDD (profile data) files we upload. Corresspond to `ExamProfile` records.
- EAD - results to EAD (exam authorization) files we upload. Correspond to `ExamAuthorization` records.

Both of these records go through a status lifecycle roughly `PENDING` -> `IN_PROGRESS` -> (`SUCCESS` | `FAILURE`). This code handles processing response files from Pearson to determine if our original request was a success or failure.

A slight complication is that these result files are contained in ZIP files and there's no documented convention of which result files show up in which zip files. The approach I've taken here is to basically walk all the zip files and then walk their contents file-by-file, handling the files that we currently support parsing for. ZIP files we don't consume fully are left on the server (in case we change our mind down the road and want to re-parse all historic data of that type). In my observations, the VCDC and EAD files always come together in the same zip file with no other file types.

The meat of this implementation is in two places:

- `exams/pearson/readers.py` - implementation of VCDC- and EAC-specific readers. These parse TSV files into a list of rows, in this case using `namedtuple`s (could be classes, but that's overkill).
- `exams/pearson/download.py` - this code walks the SFTP server and downloads each zip file into a temporary directory. Each TSV file within is then extracted locally, parsed, and processed based on the filename prefix (e.g.` vcdc-2016-02-15-a.dat` is a VCDC file). Errors are logged and emailed.

To test first ensure you're enrolled in a course and have a fully filled out profile and then you can run this in shell (keep it open!):

```python

course = Course.objects.get(id=<COURSE_ID>)

# fudge the records to sync
ea = ExamAuthorization.objects.create(user=user, course=course)

tasks.export_exam_profiles()
tasks.export_exam_authorizations()
```

Then, run this command to simulate the processing of the files you just exported:

`docker-compose run web ./manage.py simulate_sftp_responses --poll`

That command will randomly assign success/failure response, write them out to a response file and put that file in a zip file on the local sftp container. This is basically a dumb simulation of what Pearson will do. You can check the sftp server to verify these files are in place :

```shell
sftp -P 2022 odl@localhost # password: 123
cd results
ls
```

Then you can finally run the real deal of this PR in the first shell you opened (you won't see much output since it all going to the logging module):

```python
tasks.batch_process_pearson_zip_files()
ea.refresh_from_db()
ep.refresh_from_db()
print((ea.id, ea.status))
print((ep.id, ep.status))
```

That should process and remove the zip files from the SFTP server and the respective record statuses should now be either success or failure matching what `simulate_sftp_responses` output (our model constants differ from the response file constants).

Processing

### DIFF
--- a/exams/management/commands/simulate_sftp_responses.py
+++ b/exams/management/commands/simulate_sftp_responses.py
@@ -1,0 +1,176 @@
+"""Command to simulate Pearson responses"""
+import csv
+import io
+import os
+import random
+import time
+import zipfile
+from datetime import datetime
+
+import pytz
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from exams.pearson.constants import (
+    PEARSON_DATETIME_FORMAT,
+    PEARSON_FILE_TYPE_EAC,
+    PEARSON_FILE_TYPE_VCDC,
+)
+from exams.pearson import sftp
+
+
+class Command(BaseCommand):
+    """Simulates Pearson responses"""
+    help = 'Simulates the SFTP backend by generating random responses'
+
+    def add_arguments(self, parser):  # pylint: disable=no-self-use
+        """Configure command args"""
+        parser.add_argument(
+            '--ratio-success',
+            dest='ratio',
+            type=float,
+            default=0.5,
+        )
+        parser.add_argument(
+            '--interval',
+            dest='interval',
+            type=int,
+            default=5,
+        )
+        parser.add_argument(
+            '--poll',
+            action='store_true',
+            dest='poll',
+            default=False,
+        )
+        parser.add_argument(
+            '--keep-files',
+            action='store_true',
+            dest='keep',
+            default=False,
+        )
+
+    def handle(self, *args, **options):
+        """Handle the command"""
+        ratio = options['ratio']
+        poll = options['poll']
+        interval = options['interval']
+        keep = options['keep']
+
+        if poll:
+            while True:
+                self.handle_poll(ratio, keep)
+
+                self.stdout.write('Next poll in {} seconds'.format(interval))
+                time.sleep(interval)
+        else:
+            self.handle_poll(ratio, keep)
+
+    def handle_poll(self, ratio, keep):
+        """Handle a poll interval"""
+        self.stdout.write('Checking sftp server')
+
+        with sftp.get_connection() as sftp_conn:
+            with sftp_conn.cd(settings.EXAMS_SFTP_UPLOAD_DIR):
+                paths = [path for path in sftp_conn.listdir() if sftp_conn.isfile(path)]
+
+                self.stdout.write('Found {} file(s)'.format(len(paths)))
+
+                for path in paths:
+                    time.sleep(2)  # ensures unique filename timestamps
+                    self.stdout.write('Found a file: {}/{}'.format(sftp_conn.pwd, path))
+
+                    if path.startswith('ead'):
+                        self.handle_ead(sftp_conn, path, ratio)
+
+                    elif path.startswith('cdd'):
+                        self.handle_cdd(sftp_conn, path, ratio)
+                    else:
+                        continue
+
+                    if not keep:
+                        sftp_conn.remove(path)
+
+    def handle_ead(self, sftp_conn, remote_path, ratio):
+        """Handle an EAD file"""
+        now = datetime.now(pytz.utc)
+        result_file = io.StringIO()
+        writer = csv.DictWriter(result_file, [
+            'ClientAuthorizationID',
+            'ClientCandidateID',
+            'Status',
+            'Date',
+            'Message',
+        ], delimiter='\t')
+        writer.writeheader()
+        with sftp_conn.open(remote_path, mode='r') as eac_file:
+            for row in csv.DictReader(eac_file, delimiter='\t'):
+                cid = row['ClientCandidateID']
+                aid = row['ClientAuthorizationID']
+                error = random.random() > ratio
+                status = 'Error' if error else 'Accepted'
+                self.stdout.write('Marking authorization {aid} for profile {cid} as {status}'.format(
+                    aid=aid,
+                    cid=cid,
+                    status=status,
+                ))
+                writer.writerow({
+                    'ClientAuthorizationID': aid,
+                    'ClientCandidateID': cid,
+                    'Status': status,
+                    'Date': now.strftime(PEARSON_DATETIME_FORMAT),
+                    'Message': 'Invalid ExamSeriesCode' if error else '',
+                })
+
+        self.write_zip(
+            sftp_conn,
+            result_file.getvalue(),
+            now.strftime('{}-%Y-%m-%d.dat'.format(PEARSON_FILE_TYPE_EAC)),
+            now
+        )
+
+    def handle_cdd(self, sftp_conn, remote_path, ratio):
+        """Handle a CDD file"""
+        now = datetime.now(pytz.utc)
+        result_file = io.StringIO()
+        writer = csv.DictWriter(result_file, [
+            'ClientCandidateID',
+            'Status',
+            'Date',
+            'Message',
+        ], delimiter='\t')
+        writer.writeheader()
+        with sftp_conn.open(remote_path, mode='r') as cdd_file:
+            for row in csv.DictReader(cdd_file, delimiter='\t'):
+                cid = row['ClientCandidateID']
+                error = random.random() > ratio
+                status = 'Error' if error else 'Accepted'
+                self.stdout.write('Marking profile {cid} as {status}'.format(
+                    cid=cid,
+                    status=status,
+                ))
+                writer.writerow({
+                    'ClientCandidateID': cid,
+                    'Status': 'Error' if error else 'Accepted',
+                    'Date': now.strftime(PEARSON_DATETIME_FORMAT),
+                    'Message': 'Invalid Address' if error else '',
+                })
+
+        self.write_zip(
+            sftp_conn,
+            result_file.getvalue(),
+            now.strftime('{}-%Y-%m-%d.dat'.format(PEARSON_FILE_TYPE_VCDC)),
+            now
+        )
+
+    def write_zip(self, sftp_conn, data, filename, now):  # pylint: disable=no-self-use
+        """Write a zip file to the sftp server"""
+        zip_path = os.path.join(
+            settings.EXAMS_SFTP_TEMP_DIR,
+            now.strftime('ORGNAME-NS-%Y-%m-%d-%H%M%S.zip'),
+        )
+        with zipfile.ZipFile(zip_path, 'w') as zf:
+            zf.writestr(filename, data)
+
+        with sftp_conn.cd(settings.EXAMS_SFTP_RESULTS_DIR):
+            sftp_conn.put(zip_path)
--- a/exams/pearson/download.py
+++ b/exams/pearson/download.py
@@ -1,0 +1,248 @@
+"""Pearson SFTP download implementation"""
+import logging
+import os
+import zipfile
+from contextlib import contextmanager
+
+from django.conf import settings
+from paramiko import SSHException
+
+from exams.pearson.constants import (
+    EAC_SUCCESS_STATUS,
+    PEARSON_FILE_TYPE_EAC,
+    PEARSON_FILE_TYPE_VCDC,
+    VCDC_SUCCESS_STATUS,
+)
+from exams.pearson.exceptions import RetryableSFTPException
+from exams.pearson.readers import (
+    EACReader,
+    VCDCReader,
+)
+from exams.pearson import utils
+from exams.models import (
+    ExamAuthorization,
+    ExamProfile,
+)
+
+log = logging.getLogger(__name__)
+
+
+@contextmanager
+def locally_extracted(zip_file, member):
+    """
+    Context manager for temporarily extracting a zip file member to the local filesystem
+
+    Args:
+        zip_file (zipfile.ZipFile): the zip file to extract from
+        member (str): the name of the file to extract
+
+    Yields:
+        file: the extracted file object
+    """
+    extracted_file_path = zip_file.extract(member, path=settings.EXAMS_SFTP_TEMP_DIR)
+    try:
+        # csv.reader requires files to be opened in text mode, not binary
+        with open(extracted_file_path, 'r') as extracted_file:
+            yield extracted_file
+    finally:
+        if os.path.exists(extracted_file_path):
+            os.remove(extracted_file_path)
+
+
+class ArchivedResponseProcessor:
+    """
+    Handles fetching and processing of files stored in a ZIP archive on Pearson SFTP
+    """
+    def __init__(self, sftp):
+        self.sftp = sftp
+
+    def fetch_file(self, remote_path):
+        """        Fetches a remote file and returns the local path
+
+        Args:
+            remote_path (str): the remote path of the file to fetch
+
+        Returns:
+            str: the local path of the file
+        """
+        local_path = os.path.join(settings.EXAMS_SFTP_TEMP_DIR, remote_path)
+
+        self.sftp.get(remote_path, localpath=local_path)
+
+        return local_path
+
+    def filtered_files(self):
+        """
+        Walks a directory and yields files that match the pattern
+
+        Yields:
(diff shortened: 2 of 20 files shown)


noisecapella:
Functional test worked!

noisecapella:
Can you rebase by the way?

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+                    'Status': 'Error' if error else 'Accepted',
+                    'Date': now.strftime(PEARSON_DATETIME_FORMAT),
+                    'Message': 'Invalid Address' if error else '',
+                })
+
+        self.write_zip(sftp_conn, result_file.getvalue(), now.strftime('vcdc-%Y-%m-%d.dat'), now)
+
+    def write_zip(self, sftp_conn, data, filename, now):  # pylint: disable=no-self-use
+        """Write a zip file to the sftp server"""
+        zip_path = now.strftime('/tmp/ORGNAME-NS-%Y-%m-%d-%H%M%S.zip')
```
noisecapella: Is `ORGNAME` a placeholder? Should we put our own organization name here?

noisecapella: Also we should use `settings.EXAMS_SFTP_TEMP_DIR` here instead of `/tmp`

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def handle(self, *args, **options):
+        """Handle the command"""
+        ratio = options['ratio']
+        poll = options['poll']
+        interval = options['interval']
+        keep = options['keep']
+
+        if poll:
+            while True:
```
noisecapella: It looks like this is uncovered, not sure how much we care about testing this management command though

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+                    time.sleep(2)  # ensures unique filename timestamps
+                    self.stdout.write('Found a file: {}/{}'.format(sftp_conn.pwd, path))
+
+                    if path.startswith('ead'):
+                        self.handle_ead(sftp_conn, path, ratio)
+
+                    elif path.startswith('cdd'):
+                        self.handle_cdd(sftp_conn, path, ratio)
+                    else:
+                        continue
```
noisecapella: This line is uncovered

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+                    'Date': now.strftime(PEARSON_DATETIME_FORMAT),
+                    'Message': 'Invalid Address' if error else '',
+                })
+
+        self.write_zip(sftp_conn, result_file.getvalue(), now.strftime('vcdc-%Y-%m-%d.dat'), now)
+
+    def write_zip(self, sftp_conn, data, filename, now):  # pylint: disable=no-self-use
+        """Write a zip file to the sftp server"""
+        zip_path = now.strftime('/tmp/ORGNAME-NS-%Y-%m-%d-%H%M%S.zip')
+        with zipfile.ZipFile(zip_path, 'w') as zf:
```
noisecapella: This function is uncovered, just FYI. Feel free to skip if you don't think it's worth testing

rhysyngsun:
@noisecapella rebased

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+                    status=status,
+                ))
+                writer.writerow({
+                    'ClientCandidateID': cid,
+                    'Status': 'Error' if error else 'Accepted',
+                    'Date': now.strftime(PEARSON_DATETIME_FORMAT),
+                    'Message': 'Invalid Address' if error else '',
+                })
+
+        self.write_zip(sftp_conn, result_file.getvalue(), now.strftime('vcdc-%Y-%m-%d.dat'), now)
```
noisecapella: Can you use `PEARSON_FILE_TYPE_VCDC` here?

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+                ))
+                writer.writerow({
+                    'ClientAuthorizationID': aid,
+                    'ClientCandidateID': cid,
+                    'Status': status,
+                    'Date': now.strftime(PEARSON_DATETIME_FORMAT),
+                    'Message': 'Invalid ExamSeriesCode' if error else '',
+                })
+
+        self.write_zip(sftp_conn, result_file.getvalue(), now.strftime('eac-%Y-%m-%d.dat'), now)
```
noisecapella: Can you use `PEARSON_FILE_TYPE_EAC` here?

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+    extracted_file_path = zip_file.extract(member, path=settings.EXAMS_SFTP_TEMP_DIR)
+    try:
+        with open(extracted_file_path, 'r') as extracted_file:
+            yield extracted_file
+    finally:
+        if os.path.exists(extracted_file_path):
+            os.remove(extracted_file_path)
+
+
+class ArchivedResponseProcesser(object):
```
noisecapella: Feel free to leave it but we don't need `(object)` here

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def fetch_file(self, remote_path):
+        """
+        Fetches a remote file and returns the local path
+
+        Args:
+            remote_path (str): the remote path of the file to fetch
+
+        Returns:
+            local_path (str): the local path of the file
```
noisecapella: This should just be `str: the local path of the file`

exams/pearson/download.py
noisecapella: Also can you rename to `Processor`?

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        local_path = os.path.join(settings.EXAMS_SFTP_TEMP_DIR, remote_path)
+
+        self.sftp.get(remote_path, localpath=local_path)
+
+        return local_path
+
+    def filtered_files(self):
+        """
+        Walks a directory and yeilds files that match the pattern
```
noisecapella: Should be `yields`

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+
+        Args:
+            pattern (re): the regex pattern to mathc the filename on
```
noisecapella: Should be `match`, but I don't see an argument here. Also we should have a `Yields` description here (not exactly sure how that works compared to `Returns`)

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+                    except SSHException as exc:
+                        raise RetryableSFTPException(
+                            "Error trying to remove file from sftp: {}".format(remote_path)
+                        ) from exc
+                    except:  # pylint: disable=bare-except
+                        log.exception("Error processing file: %s", remote_path)
+                    finally:
+                        if os.path.exists(local_path):
+                            os.remove(local_path)
+        except SSHException as exc:
```
noisecapella: What exception would get handles here which wouldn't get handled in the inner except block?

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def process_zip(self, local_path):
+        """
+        Process a single zip file
+
+        Args:
+            local_path (str): path to the zip file on the local filesystem
+
+        Returns:
+            (bool): True if all files processed successfully
```
noisecapella: Should be `bool: True if all...`

exams/management/commands/simulate_sftp_responses.py
rhysyngsun: There isn't an easy way to terminate this in a test. I think there are some packages that implement timeout-based short-circuiting but I don't think all that is worth pulling in just to test this small section of code.

exams/management/commands/simulate_sftp_responses.py
rhysyngsun: It's just a placeholder, TBH, the naming conventions of the real-world files aren't documented so I want to be sure there's no depedency on that.

exams/pearson/download.py
rhysyngsun: Good point, I missed that when I added the outer one, I'll change the inner to a vanilla `raise`.

exams/management/commands/simulate_sftp_responses.py
noisecapella: Sounds good

exams/management/commands/simulate_sftp_responses.py
noisecapella: Sounds good

exams/pearson/download.py
rhysyngsun: From [this](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), it looks like it should describe what an individual `yield` statement yields (in this case, the tuple).

exams/tasks.py
```diff
@@ -135,3 +129,18 @@ def export_exam_authorizations(self):
...
+def batch_process_pearson_zip_files():
+    """
+    Fetch zip files from pearsons sftp periodically.
+    """
+    try:
+        with sftp.get_connection() as sftp_connection:
+            processor = download.ArchivedResponseProcesser(sftp_connection)
+            processor.process()
+    except ImproperlyConfigured as exc:
+        log.warning('batch_process_pearson_zip_files got an ImproperlyConfigured exception: %s', exc)
```
noisecapella: Shouldn't we use `log.exception` here?

exams/utils.py
```diff
@@ -13,9 +16,35 @@
...
+        raise ImproperlyConfigured('EXAMS_SFTP_BACKOFF_BASE must be an integer') from ex
+
+
 def course_has_exam(mmtrack, course_run):
     """
     Check if the user is authorized for an exam for a course run
+
+    Args:
+        mmtrack (dashboard.utils.MMTrack): mmtrack for the user/program
+        cource_run (cources.models.CourseRun): CourseRun to check against
```
noisecapella: `course_run`

noisecapella: And `courses.models.CourseRun`

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+@contextmanager
+def locally_extracted(zip_file, member):
+    """
+    Context manager for temporarily extracting a zip file member to the local filesystem
+
+    Args:
+        zip_file (zipfile.ZipFile): the zip file to extract from
+        member (str|zipfile.ZipInfo): the name or ZipInfo of the member to extract
```
rhysyngsun: Not in any current code, but the pysftp api accepts it. If you don't feel that's worth documenting I can simplify to just `str`.

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,158 @@
...
+
+                    elif path.startswith('cdd'):
+                        self.handle_cdd(sftp_conn, path, ratio)
+                    else:
+                        continue
+
+                    if not keep:
+                        sftp_conn.remove(path)
+
+    def handle_ead(self, sftp_conn, remote_path, ratio):
```
noisecapella: This should be `handle_eac` right?

exams/pearson/download.py
noisecapella: Yeah, let's use `str` and we can add back `ZipInfo` if we decide to use that later on

exams/tasks.py
rhysyngsun: This task will be run on cron, my thought was to use a warning because it might not want exceptions coming up in all environments, thoughts?

noisecapella: I feel like if it's an `ImproperlyConfigured` then it's something we should be alerted to, even if it's a false alarm on rc or ci or other environments

rhysyngsun: OK, agreed on that. I'm going to add a feature flag for this task so that we only get errors in environments that we've explicitly turned this on but not configured this for.

exams/management/commands/simulate_sftp_responses.py
rhysyngsun: No, EAC is the response to the EAD file this method is processing.

exams/pearson/readers.py
```diff
@@ -108,3 +110,40 @@ def read(self, tsv_file):
...
+            ('Status', 'status'),
+            ('Date', 'date', self.parse_datetime),
+            ('Message', 'message'),
+        ], VCDCResult)
+
+
+EACResult = namedtuple('EACResult', ['exam_authorization_id', 'candidate_id', 'status', 'message'])
+
+
+class EACReader(BaseTSVReader):
```
rhysyngsun: We could, but it's worth adding so we have that info should we need it.

exams/pearson/download.py
```diff
@@ -0,0 +1,248 @@
...
+
+    Yields:
+        file: the extracted file object
+    """
+    extracted_file_path = zip_file.extract(member, path=settings.EXAMS_SFTP_TEMP_DIR)
+    try:
+        with open(extracted_file_path, 'r') as extracted_file:
```
rhysyngsun: I recall getting an error trying to read it in binary mode, let me give it another shot and at the very least I'll add a comment explaining it if I am still getting the error.

rhysyngsun: Ok, so here's the deal, `csv.DictReader` can take an str or an object that responds to the iterator protocol and returns strings. Opening in binary mode makes the file object's iterator return bytes so I'd need to add some intermediate code  like `data = [str(line) for line in extracted_file]` which to me makes the code a bit messier. The file should always be text so I don't see an issue opening the file in this way anyway.

noisecapella: Sure, let's leave it as is then

rhysyngsun:
@noisecapella sorry for the wait, ready for review again

exams/management/commands/simulate_sftp_responses.py
```diff
@@ -0,0 +1,176 @@
...
+        """Handle an EAC file"""
```
noisecapella: This should be `EAD` I think

exams/pearson/download.py
```diff
@@ -0,0 +1,251 @@
...
+        self.sftp.get(remote_path, localpath=local_path)
+
+        return local_path
+
+    def filtered_files(self):
+        """
+        Walks a directory and yields files that match the pattern
+
+        Args:
+            pattern (re): the regex pattern to match the filename on
```
noisecapella: Can you remove this arg from the docstring since there's no corresponding argument?

exams/pearson/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def is_zip_file(filename):
+    """
+    Checks if a filename looks like a zip file
+
+    Args:
+        filename (str): filename to check
+    Returns:
+        (bool): True if the file is a zip file
```
noisecapella: Can you change to `bool: True if ...`?

exams/pearson/download.py
noisecapella: Looks like you missed this

rhysyngsun:
Done and squashed, merge pending thaw.

Issue #2445, opened 2017-01-26
[exam] Command `authorization_user_exam` crashed because of using user id instead of object 


amir-qayyum-khan:
see https://github.com/mitodl/micromasters/blob/master/exams/management/commands/authorization_user_exam.py#L60

```python
Traceback (most recent call last):
  File "manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/app/exams/management/commands/authorization_user_exam.py", line 72, in handle
    authorize_for_exam(mmtrack, enrollment.course_run)
  File "/app/exams/utils.py", line 36, in authorize_for_exam
    ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)
AttributeError: 'int' object has no attribute 'profile'
```

@pdpinch

amir-qayyum-khan:
Issue ripple from issue on #2331 
RC #2435

PR #2448, opened 2017-01-27
Fixed exam authorization command and refactoring

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/2445
- Fixed ripples of #2331

#### What's this PR do?
- Refactored the code of exam authorization command
- Added exceptions and logging to see behaviour of code
- Added new tests for authorization code.

### How to test:
- In .env add `FEATURE_FINAL_GRADE_ALGORITHM=v1`
- Set program.exam_series_code
- Set course.exam_module
- Pay for course
- Create final grade http://192.168.99.100:8079/admin/grades/finalgrade/
- Run command ```docker-compose run web ./manage.py authorization_user_exam --username=(value) --program-id=(value)```
- Check weather ( ExamAuthoriztion) object created 
  - ``docker-compose run web ./manage.py shell``
  - ``from exams.models import ExamAuthorization``
  -  ``ExamAuthorization.objects.all()``

#### To create dummy data:
```docker-compose run web ./manage.py shell```

```shell
from dashboard.factories import CachedEnrollmentFactory
from ecommerce.factories import (OrderFactory, LineFactory)
from exams.models import ExamAuthorization

user = User.objects.get(username='amirq')
# use same  `edx_course_key` for final grade entry
course_run = CourseRun.objects.get(edx_course_key='course-v1:MITx+Digital+Learning+100+Jan_2016')
CachedEnrollmentFactory.create(user=user, course_run=course_run)
# enter  `total_price_paid` according to income threshold
order = OrderFactory.create(user=user, status='fulfilled', total_price_paid=1200)
LineFactory.create(order=order, course_key=course_run.edx_course_key)
ExamAuthorization.objects.all()
```

@pdpinch @giocalitri

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,21 @@ You should now be able to do the following:
 1. Click "Sign in with edX.org" and sign in by authorizing an edX client. If you're
  running edX locally, this will be the client you created in the steps above.
 
+#### 7) Authorize for exam manually
+ We can authorize user(s) for exam(s) manually using django command only if they paid and 
+ passed selected course(s). **Note:** ``course.exam_module`` and ``program.exam_series_code`` must be set.
+ 
+ **params**
+ - **--username:** (optional) If provided, command will look for given user only, 
+    and authorize him if he matches the critaria.
+ - **--program-id:** (optional) if provided, command will look for given program only, 
+    and authorize all users who matches the critaria.
+ - **--all:** (optional) if provided, command will authorize all eligible users for exams.
+ 
+```shell 
+    docker-compose run web ./manage.py authorization_user_exam --username=(value) --program-id=(value)
+```
+
 ## Wagtail CMS (Content Management System)
 
 The CMS can be found at `/cms/`. Use the CMS to manage the content of the program pages and, by extension, the home 
--- a/courses/utils.py
+++ b/courses/utils.py
@@ -89,3 +89,16 @@ def get_year_season_from_course_run(course_run):
         return season_year_tuple[0], NUMBER_SEASON_MAP[season_year_tuple[1]]
     else:
         return season_year_tuple
+
+
+def is_blank(text):
+    """
+    Returns true of string is blank i.e None or empty.
+
+    Args:
+        text(str): any string
+
+    Returns:
+        bool: True if input is empty or none
+    """
+    return not (text and text.strip())
--- a/courses/utils_test.py
+++ b/courses/utils_test.py
@@ -2,13 +2,21 @@
 Test cases for course utils
 """
 from datetime import datetime
+
+from ddt import (
+    ddt,
+    data,
+    unpack
+)
 from django.test import TestCase
 from courses.utils import (
-    get_year_season_from_course_run
+    get_year_season_from_course_run,
+    is_blank
 )
 from courses.factories import CourseRunFactory
 
 
+@ddt
 class CourseUtilTests(TestCase):
     """
     Test cases for course utils
@@ -34,3 +42,13 @@ class CourseUtilTests(TestCase):
         assert get_year_season_from_course_run(test_run4) == (2016, 'Fall')
         assert get_year_season_from_course_run(unparseable_test_run1) == ()
         assert get_year_season_from_course_run(unparseable_test_run2) == ()
+
+    @data(
+        (None, True),
+        ("", True),
+        ("Foo", False),
+    )
+    @unpack
+    def test_is_blank(self, text, expected):
+        """Test is_blank"""
+        assert is_blank(text) is expected
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -19,7 +19,6 @@ from financialaid.models import FinancialAid, TierProgram
 from grades.constants import FinalGradeStatus
 from grades.models import FinalGrade
 from exams.models import ExamProfile, ExamAuthorization
-from exams.utils import course_has_exam
 
 
 log = logging.getLogger(__name__)
@@ -362,7 +361,7 @@ class MMTrack:
             course__program=self.program
         )
 
-        if not any(course_has_exam(self, cr) for cr in course_runs_for_program):
+        if not any(course_run.has_exam for course_run in course_runs_for_program):
             return ""
 
         user = self.user
--- a/exams/management/commands/authorization_user_exam.py
+++ b/exams/management/commands/authorization_user_exam.py
@@ -2,19 +2,15 @@
 Authorize for exam of passed users
 """
 from django.core.management import BaseCommand
-from django.db.models import Q
 
-from courses.models import Program
-from dashboard.models import CachedEnrollment, ProgramEnrollment
-from dashboard.utils import get_mmtrack
-from exams.utils import authorize_for_exam
+from exams.utils import bulk_authorize_for_exam
 
 
 class Command(BaseCommand):
     """
     Authorizations of exam of passed users
     """
-    help = "Trigger exam authorizations when users already passed course"
+    help = "Trigger exam authorization when for users who have paid and passed course(s) "
 
     def add_arguments(self, parser):
         parser.add_argument(
@@ -32,41 +28,27 @@ class Command(BaseCommand):
             type=int
         )
         parser.add_argument(
-            "--course-key",
-            dest="course_key",
-            default=None,
-            help="Edx course key where we want to authorize users who passed course(s).",
-            type=int
+            "--all",
+            action='store_true',
+            help="Authorize eligible users for exams."
         )
 
     def handle(self, *args, **options):
         """
         Trigger exam authorizations for user for a course if he has passed and paid for it.
         """
-        # pylint: disable=too-many-locals
-        program_id = options.get("program_id", None)
-        username = options.get("username", None)
-        course_key = options.get("course_key", None)
+        all_allowed = options.get("all")
+        program_id = options.get("program_id")
+        username = options.get("username")
 
-        programs = Program.objects.filter(exam_series_code__isnull=False, live=True).filter(~Q(exam_series_code=""))
-        if program_id:
-            programs = programs.filter(id=program_id)
-
-        for program in programs:
-            users_qset = ProgramEnrollment.objects.filter(program=program)
-            if username:
-                users_qset = users_qset.filter(user__username=username)
-
-            users = users_qset.values_list('user', flat=True)
-
-            for user in users:
-                enrollments_qset = CachedEnrollment.user_course_qset(user, program=program)
-                if course_key:
-                    enrollments_qset = enrollments_qset.filter(course_run__edx_course_key=course_key)
-
-                mmtrack = get_mmtrack(user, program)
-
-                for enrollment in enrollments_qset:
-                    # if user has passed and paid for the course
-                    # and not already authorized for exam the create authorizations.
-                    authorize_for_exam(mmtrack, enrollment.course_run)
+        if all_allowed:
+            bulk_authorize_for_exam()
+            self.stdout.write(self.style.SUCCESS('Done with authorization process'))
+        elif program_id or username:
+            bulk_authorize_for_exam(
+                program_id=program_id,
+                username=username
+            )
+            self.stdout.write(self.style.SUCCESS('Done with authorization process'))
+        else:
+            self.stdout.write(self.style.ERROR('Incomplete arguments. Please use --help to see options'))
--- a/exams/management/commands/test_authorization_user_exam.py
+++ b/exams/management/commands/test_authorization_user_exam.py
@@ -1,0 +1,54 @@
+"""authorization command test"""
+
+from unittest.mock import patch
+from django.test import TestCase
+from django.db.models.signals import post_save
+from factory.django import mute_signals
+
+from financialaid.api_test import create_program
+from exams.management.commands import authorization_user_exam
+from profiles.factories import ProfileFactory
+
+
+class AuthorizationCommandTests(TestCase):
+    """Authorization command tests"""
+    @classmethod
+    def setUpTestData(cls):
+        with mute_signals(post_save):
+            profile1 = ProfileFactory.create()
+            profile2 = ProfileFactory.create()
+
+        cls.program, _ = create_program(past=True)
+        cls.users = [profile1.user, profile2.user]
+
+        cls.command = authorization_user_exam.Command()
+
+    def test_bulk_exam_authorization(self):
+        """For all users in any program"""
+        with patch('exams.management.commands.authorization_user_exam.bulk_authorize_for_exam') as mocked_bafe:
+            self.command.handle("authorization_user_exam", all=True)
+
+        mocked_bafe.assert_called_once_with()
+
+    def test_bulk_exam_authorization_given_user_name(self):
+        """For given user"""
+        user1 = self.users[0]
+        with patch('exams.management.commands.authorization_user_exam.bulk_authorize_for_exam') as mocked_bafe:
+            self.command.handle("authorization_user_exam", username=user1.username)
+
+        mocked_bafe.assert_called_once_with(program_id=None, username=user1.username)
+
+    def test_bulk_exam_authorization_given_program(self):
+        """For given program id"""
+        with patch('exams.management.commands.authorization_user_exam.bulk_authorize_for_exam') as mocked_bafe:
+            self.command.handle("authorization_user_exam", program_id=self.program.id)
+
+        mocked_bafe.assert_called_once_with(program_id=self.program.id, username=None)
+
+    def test_bulk_exam_authorization_given_program_username(self):
+        """For given program id and user name"""
+        user1 = self.users[0]
+        with patch('exams.management.commands.authorization_user_exam.bulk_authorize_for_exam') as mocked_bafe:
+            self.command.handle("authorization_user_exam", program_id=self.program.id, username=user1.username)
+
+        mocked_bafe.assert_called_once_with(program_id=self.program.id, username=user1.username)
--- a/exams/signals.py
+++ b/exams/signals.py
@@ -1,6 +1,7 @@
 """
 Signals for exams
 """
+import logging
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
@@ -9,10 +10,14 @@ from dashboard.utils import get_mmtrack
 from exams.models import ExamProfile
 from exams.utils import (
     authorize_for_exam,
-    is_eligible_for_exam
+    ExamAuthorizationException,
+    is_eligible_for_exam,
 )
 from grades.models import FinalGrade
 from profiles.models import Profile
+
+
+log = logging.getLogger(__name__)
 
 
 @receiver(post_save, sender=Profile, dispatch_uid="update_exam_profile")
@@ -29,7 +34,14 @@ def update_exam_authorization_final_grade(sender, instance, **kwargs):  # pylint
     Signal handler to trigger an exam profile and authorization for FinalGrade creation.
     """
     mmtrack = get_mmtrack(instance.user, instance.course_run.course.program)
-    authorize_for_exam(mmtrack, instance.course_run)
+    try:
+        authorize_for_exam(mmtrack, instance.course_run)
+    except ExamAuthorizationException:
+        log.exception(
+            'Unable to authorize user: %s for exam on course_id: %s',
+            instance.user.username,
+            instance.course_run.course.id
+        )
 
 
 @receiver(post_save, sender=CachedEnrollment, dispatch_uid="update_exam_authorization_cached_enrollment")
--- a/exams/signals_test.py
+++ b/exams/signals_test.py
@@ -2,6 +2,8 @@
 Tests for exam signals
 """
 from datetime import datetime, timedelta
+
+from unittest.mock import patch
 from django.contrib.auth.models import User
 from django.db.models.signals import post_save
 from django.test import override_settings
@@ -17,11 +19,11 @@ from dashboard.factories import (
     CachedCurrentGradeFactory,
     CachedEnrollmentFactory,
 )
+from exams.utils_test import create_order
 from exams.models import (
     ExamProfile,
     ExamAuthorization
 )
-from exams.utils_test import create_order
 from financialaid.api_test import create_program
 from grades.factories import FinalGradeFactory
 from profiles.factories import ProfileFactory
@@ -96,11 +98,41 @@ class ExamSignalsTest(MockedESTestCase):
         )
 
         # assert Exam Authorization and profile created.
-        self.assertTrue(ExamProfile.objects.filter(profile=self.profile).exists())
-        self.assertTrue(ExamAuthorization.objects.filter(
+        assert ExamProfile.objects.filter(profile=self.profile).exists() is True
+        assert ExamAuthorization.objects.filter(
             user=self.profile.user,
             course=self.course_run.course
-        ).exists())
+        ).exists() is True
+
+    @override_settings(FEATURES={"FINAL_GRADE_ALGORITHM": "v1"})
+    def test_update_exam_authorization_final_grade_when_user_not_paid(self):
+        """
+        Verify that update_exam_authorization_final_grade is called and log exception when
+        FinalGrade saves user dont match exam authorization criteria
+        """
+        with mute_signals(post_save):
+            # muting signal for CachedEnrollment. Because CachedEnrollment and FinalGrade both omits
+            # signal, we want to see behaviour of FinalGrade here
+            CachedEnrollmentFactory.create(user=self.profile.user, course_run=self.course_run)
+
+        with patch("exams.signals.log") as log:
+            FinalGradeFactory.create(
+                user=self.profile.user,
+                course_run=self.course_run,
+                passed=True,
+            )
+
+        log.exception.assert_called_with(
+            'Unable to authorize user: %s for exam on course_id: %s',
+            self.profile.user.username,
+            self.course_run.course.id
+        )
+        # assert Exam Authorization and profile created.
+        assert ExamProfile.objects.filter(profile=self.profile).exists() is False
+        assert ExamAuthorization.objects.filter(
+            user=self.profile.user,
+            course=self.course_run.course
+        ).exists() is False
 
     def test_update_exam_authorization_cached_enrollment(self):
         """
--- a/exams/utils.py
+++ b/exams/utils.py
@@ -6,7 +6,14 @@ import pytz
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.db.models import Q
 
+from courses.models import Program
+from dashboard.models import (
+    CachedEnrollment,
+    ProgramEnrollment
+)
+from dashboard.utils import get_mmtrack
 from exams.models import (
     ExamProfile,
     ExamAuthorization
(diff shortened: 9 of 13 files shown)


README.md
```diff
@@ -230,6 +230,22 @@ You should now be able to do the following:
...
+ 
+ **params**
+ - **--username:** (optional) If provided, command will look for given user only, 
+    and authorize him if he matches the critaria.
+ - **--program-id:** (optional) if provided, command will look for given program only, 
+    and authorize all users who matches the critaria.
+ - **--course-key:**  (optional) if provided, command will look for given course only, 
+    and authorize all users who matches critaria.
+ 
+```python
```
rhysyngsun: This should be ` ```shell ` instead.

exams/signals.py
```diff
@@ -29,10 +29,10 @@ def update_exam_authorization_final_grade(sender, instance, **kwargs):  # pylint
     authorize_for_exam(mmtrack, instance.course_run)
 
 
-@receiver(post_save, sender=CachedEnrollment, dispatch_uid="update_exam_authorization_cached_enrollment")
-def update_exam_authorization_cached_enrollment(sender, instance, **kwargs):  # pylint: disable=unused-argument
+@receiver(post_save, sender=CachedCurrentGrade, dispatch_uid="update_exam_authorization_grades")
```
rhysyngsun: Why is this changing to `CachedCurrentGrade`, could you give some context? My understanding was that `CachedCurrentGrade` was the non-final grade in our system so tt seems we lose the behavior of doing this when the user enrolls.

exams/utils.py
```diff
@@ -59,3 +67,53 @@ def authorize_for_exam(mmtrack, course_run):
...
+                course_run.edx_course_key
+            )
+
+
+def bulk_authorize_for_exam(program_id=None, username=None, course_key=None):
+    """
+    Authorize user(s) for exam(s).
+
+    Args:
+        program_id(str): Program id (optional) if you want authorization on spcific program exams
```
rhysyngsun: Typo: spcific

exams/utils.py
```diff
@@ -59,3 +67,53 @@ def authorize_for_exam(mmtrack, course_run):
...
+        return
+
+    for program in programs:
+        program_enrollment_qset = ProgramEnrollment.objects.filter(program=program)
+        if username:
+            program_enrollment_qset = program_enrollment_qset.filter(user__username=username)
+
+        if not program_enrollment_qset.exists():
+            log.info('[authorization_user_exam] User(s) not exist for exam authorization.')
+            return
```
rhysyngsun: Shouldn't this be a `continue`?

exams/utils.py
```diff
@@ -59,3 +67,53 @@ def authorize_for_exam(mmtrack, course_run):
...
+
+        for program_enrollment in program_enrollment_qset:
+            user = program_enrollment.user
+            enrollments_qset = CachedEnrollment.user_course_qset(user, program=program)
+            if course_key:
+                enrollments_qset = enrollments_qset.filter(course_run__edx_course_key=course_key)
+
+            if not enrollments_qset.exists():
+                log.info('[authorization_user_exam] Course(s) not exist for exam authorization.')
+                return
```
rhysyngsun: Shouldn't this be a `continue`?

exams/utils.py
```diff
@@ -59,3 +67,53 @@ def authorize_for_exam(mmtrack, course_run):
...
+def bulk_authorize_for_exam(program_id=None, username=None, course_key=None):
+    """
+    Authorize user(s) for exam(s).
+
+    Args:
+        program_id(str): Program id (optional) if you want authorization on spcific program exams
+        username(str): User name (optional) whom you want to authorize for exam(s)
+        course_key(str): Edx course key  (optional) where you want to authorize user(s)
+    """
+    programs = Program.objects.filter(exam_series_code__isnull=False, live=True).filter(~Q(exam_series_code=""))
```
rhysyngsun: I would suggest refactoring these filters a little so it's a little easier to read:

```python
Program.objects.filter(live=True).exclude(exam_series_code__isnull=True).exclude(exam_series_code__exact="")
```

exams/utils.py
```diff
@@ -59,3 +67,53 @@ def authorize_for_exam(mmtrack, course_run):
...
+            if not enrollments_qset.exists():
+                log.info('[authorization_user_exam] Course(s) not exist for exam authorization.')
+                return
+
+            mmtrack = get_mmtrack(user, program)
+
+            for enrollment in enrollments_qset:
+                # if user has passed and paid for the course
+                # and not already authorized for exam the create authorizations.
+                authorize_for_exam(mmtrack, enrollment.course_run)
```
rhysyngsun: I'm not sure we want to authorize the user for *every* course_run the user ever took. @pdpinch I know you mentioned we want to check all the course runs, but do we want to do authorize for all runs or do something more targeted like only authorize for the most recent course run the user is eligible to take an exam for?

financialaid/api_test.py
```diff
@@ -36,7 +36,7 @@
 from search.base import MockedESTestCase
 
 
-def create_program(create_tiers=True, past=False):
+def create_program(create_tiers=True, past=False, exam_series_code="", exam_module=""):
```
rhysyngsun: I think `None` might be a better default here

exams/signals.py
amir-qayyum-khan: actually we determine if user is pass or not using a process that involves current grade 
https://github.com/mitodl/micromasters/blob/master/dashboard/utils.py#L223-L233

I think i should also add a signal for cached certificates as well, because for not financial_aid programs

rhysyngsun: Are we fixing all these signals in the same PR? I'd advocate for addressing any changes to signals separately so this review doesn't drag out.

amir-qayyum-khan: i changes the title of PR and description is updated. I think we should fix every issue on exam authorization triggering here. To avoid complexity because code is already merge and think it has some issues in both signals and command

rhysyngsun: @amir-qayyum-khan could you update the description with some details on why you're making these changes around signals and how I can manually test everything here?

exams/utils.py
rhysyngsun: I discussed with @pdpinch earlier and he agreed that we should only authorize once per course in this command and that the most recently passed one is probably the most reasonable one to use.

exams/signals.py
amir-qayyum-khan: Done updated

exams/utils.py
amir-qayyum-khan: So what if course is ongoing, should i look for status passed as well?

amir-qayyum-khan:
@rhysyngsun removed signal code here

exams/management/commands/authorization_user_exam.py
```diff
@@ -48,25 +37,9 @@ def handle(self, *args, **options):
...
+        bulk_authorize_for_exam(
+            program_id=program_id,
+            username=username
+        )
 
-                for enrollment in enrollments_qset:
-                    # if user has passed and paid for the course
-                    # and not already authorized for exam the create authorizations.
-                    authorize_for_exam(mmtrack, enrollment.course_run)
+        print("Done with authorization process")
```
giocalitri: please do not use "print" statements: this is an example of how to proper write output of management commands (also with a nice output) taken from
https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/
```
self.stdout.write(self.style.SUCCESS('Successfully closed poll "%s"' % poll_id))
```

exams/utils.py
```diff
@@ -59,3 +65,59 @@ def authorize_for_exam(mmtrack, course_run):
...
+            course_ids = set()
+
+            for edx_course_key in mmtrack.edx_key_course_map.keys():
+                course_ids.add(mmtrack.edx_key_course_map[edx_course_key])
+
+            # get latest course_run from given course
+            for course_id in course_ids:
+                enrollments_qset = CachedEnrollment.user_course_qset(user, program=program).filter(
+                    course_run__course__id=course_id
+                ).order_by('course_run__end_date')
```
giocalitri: from what I read this should be ordered `desc`

exams/utils_test.py
```diff
@@ -23,43 +27,49 @@
...
-            course_run=course_run,
-            data={
-                "passed": True,
-                "percent": 0.9,
-                "course_key": course_run.edx_course_key,
-                "username": cls.user.username
-            }
-        )
+        CachedEnrollmentFactory.create(user=cls.user, course_run=course_run)
+        with mute_signals(post_save):
```
giocalitri: why there is this `mute_signals`?

amir-qayyum-khan: it need to update changes status to in progress

exams/utils_test.py
```diff
@@ -107,3 +117,111 @@ def test_exam_authorization_when_not_passed_course(self):
...
+    """Tests for authorization_user_exam command operations"""
+    @classmethod
+    def set_data(cls, user, program, course_run):
+        """
+        creates user records like CachedGrades, CachedEnrollments and CachedCertificate
+        """
+        with mute_signals(post_save):
+            ProgramEnrollment.objects.create(user=user, program=program)
+        CachedEnrollmentFactory.create(user=user, course_run=course_run)
+        with mute_signals(post_save):
```
giocalitri: why the `mute_signals` here?

exams/utils_test.py
giocalitri: sorry I do not understand: what's the signal handler hooked to `CachedCurrentGrade`>

exams/utils_test.py
```diff
@@ -107,3 +117,111 @@ def test_exam_authorization_when_not_passed_course(self):
...
+            profile1 = ProfileFactory.create()
+            profile2 = ProfileFactory.create()
+
+        cls.program, cls.tier_programs = create_program(
+            past=True,
+            exam_series_code="TestExam",
+            exam_module="SampleCourse"
+        )
+        cls.course_run = course_run = cls.program.course_set.first().courserun_set.first()
```
giocalitri: you do not need 2 variables here

exams/utils_test.py
```diff
@@ -107,3 +117,111 @@ def test_exam_authorization_when_not_passed_course(self):
...
+        cls.users = users = [profile1.user, profile2.user]
```
giocalitri: and not even here you need 2 variables

financialaid/api_test.py
```diff
@@ -64,7 +66,7 @@ def create_program(create_tiers=True, past=False):
         is_valid=True
     )
     tier_programs = None
-    if create_tiers:
+    if fa_availability and create_tiers:
```
giocalitri: why this change?

giocalitri:
In addition to the comments I already made, I think that `bulk_authorize_for_exam` should be split in multiple functions and each of them should address one step of `bulk_authorize_for_exam`,
but we can open a different issue for this.

giocalitri:
I am doing functional testing and the command is ending (apparently with success) without doing anything (because of missing configurations): I do not think this should happen.
I am not sure if this should be part of this PR or not.

financialaid/api_test.py
amir-qayyum-khan: to create Non fa program

giocalitri:
I also think that the management command by default should not do anything without parameters and if someone wants to run for everything, she should specify a `--all` argument, while without arguments the script should just fail and return the `--help`.
Or just make 2 different management commands, one for all the users and one for the specific program and/or user.
Not sure this belongs to this PR.

exams/utils.py
```diff
@@ -59,3 +65,59 @@ def authorize_for_exam(mmtrack, course_run):
...
+            log.info('[authorization_user_exam] User(s) not exist for exam authorization.')
+            continue
+
+        for program_enrollment in program_enrollment_qset:
+            user = program_enrollment.user
+            mmtrack = get_mmtrack(user, program)
+            course_ids = set()
+
+            for edx_course_key in mmtrack.edx_key_course_map.keys():
+                course_ids.add(mmtrack.edx_key_course_map[edx_course_key])
```
giocalitri: cannot you just do
`course_ids = set(mmtrack.edx_key_course_map.values())`
or am I missing something?

giocalitri: I might be wrong, but I think you should filter also by courses that are finished. You might just filter by course_run that have a frozen final grade, (but this should not be done in this PR, probably)

amir-qayyum-khan: ok

amir-qayyum-khan:
@giocalitri  i am logging success or failiure

giocalitri:
`exam.utils.authorize_for_exam` needs to be refactored to raise exceptions in case of missing configurations.

giocalitri:
> @giocalitri i am logging success or failiure

I am not sure which comment are you referring to, but if it is 

>I am doing functional testing and the command is ending (apparently with success) without doing anything (because of missing configurations): I do not think this should happen.
I am not sure if this should be part of this PR or not.

I think you are wrong: 
1. You use `log.info` that by default will be silenced (and it is silenced on my local env by default)
2. The code was stopping in my case inside `authorize_for_exam` because `ok_for_exam` was `False and there there is no logging of any kind.

giocalitri:
Ok: I mocked my way through the code and I got the result.
I think that it will be @pdpinch decision if any of the suggestions above need to be implemented in this PR:
I advice against shipping without any kind of explicit error returned or raised by the management command.

exams/management/commands/authorization_user_exam.py
```diff
@@ -32,41 +28,29 @@ def add_arguments(self, parser):
...
+            bulk_authorize_for_exam()
+            self.stdout.write(self.style.SUCCESS('Done with authorization process'))
+        elif program_id or username:
+            bulk_authorize_for_exam(
+                program_id=program_id,
+                username=username
+            )
+            self.stdout.write(self.style.SUCCESS('Done with authorization process'))
+        else:
+            self.stdout.write(self.style.SUCCESS('No param is provided.'))
```
giocalitri: This should be modified to not return a success message (there is no success here): this should return the `--help` option.

amir-qayyum-khan: done

exams/management/commands/authorization_user_exam.py
```diff
@@ -32,41 +28,29 @@ def add_arguments(self, parser):
...
         )
 
     def handle(self, *args, **options):
         """
         Trigger exam authorizations for user for a course if he has passed and paid for it.
         """
-        # pylint: disable=too-many-locals
+        all = options.get("all", 'n')
         program_id = options.get("program_id", None)
         username = options.get("username", None)
```
giocalitri: question: are you sure you need to specify the default if there is a `default=` option when you add the arguments before?

exams/management/commands/authorization_user_exam.py
```diff
@@ -32,41 +28,29 @@ def add_arguments(self, parser):
...
-            "--course-key",
-            dest="course_key",
-            default=None,
-            help="Edx course key where we want to authorize users who passed course(s).",
-            type=int
+            "--all",
+            dest="all",
+            default='n',
+            help="Authorize eligible users for exams. Can be y or n",
+            type=str
```
giocalitri: this should just be an argument without value, meaning I should be able to specify `--all` and not `--all y`

exams/management/commands/authorization_user_exam.py
amir-qayyum-khan: yes, you are right, we dont need

exams/management/commands/authorization_user_exam.py
amir-qayyum-khan: Done

exams/signals_test.py
```diff
@@ -38,7 +38,11 @@ def setUpTestData(cls):
         with mute_signals(post_save):
             cls.profile = profile = ProfileFactory.create()
 
-        program, _ = create_program(past=True)
+        program, _ = create_program(
+            past=True,
+            exam_series_code="TestExam",
+            exam_module="SampleCourse"
+        )
```
giocalitri: is this a leftover of the PR split or do you need it?

exams/utils.py
```diff
@@ -59,3 +65,84 @@ def authorize_for_exam(mmtrack, course_run):
...
+    """
+    Authorize user(s) for exam(s).
+
+    Args:
+        program_id(str): Program id (optional) if you want authorization on specific program exams
+        username(str): User name (optional) whom you want to authorize for exam(s)
+    """
+    programs = Program.objects.filter(
+        live=True
+    ).exclude(exam_series_code__isnull=True).exclude(exam_series_code__exact="")
```
giocalitri: nit: you can use the `Q` operator here and use only one `.exclude`

giocalitri:
you need to write a test for the management command just to assert that the right function with the right parameters is called.

giocalitri:
@pdpinch please test this and see if the behavior (erroneous cases included) is good enough.

This need to be refactored anyway to remove the logging and raise exceptions to be explicit about where the code stopped, but we can open an issue about this.

exams/signals_test.py
amir-qayyum-khan: nop, these conditions must meet for user to be authorize for exam.

giocalitri: I understand, but this file is `signal_test` and you have not modified the file `signal`, so why there is a change here?

dashboard/utils.py
```diff
@@ -371,6 +370,18 @@ def get_pearson_exam_status(self):  # pylint: disable=too-many-return-statements
...
+        """
+        Check if the user is authorized for an exam for a course run
+        """
+        return bool(
+            course_run and
+            course_run.edx_course_key and
+            course_run.course and
+            course_run.course.exam_module and
+            self.program.exam_series_code
+        )
```
giocalitri: why this is here and it is not a property of the `CourseRun` model?

exams/signals_test.py
giocalitri: you changed `exams/signals.py` but there are no new tests for the changes

exams/utils.py
```diff
@@ -59,7 +68,105 @@ def authorize_for_exam(mmtrack, course_run):
...
+                user=mmtrack.user.username,
+                course_id=course_run.edx_course_key
+            )
+            raise ExamAuthorizationException(errors_message)
+    else:
+        errors_message = message_not_eligible_template.format(
+            user=mmtrack.user.username,
+            course_id=course_run.edx_course_key
+        )
+        raise ExamAuthorizationException(errors_message)
```
giocalitri: please refactor this code in the following way (please verify that it actually works):
```
def authorize_for_exam(mmtrack, course_run):
    """
    Authorize user for exam if he has paid for course and passed course.
    Args:
        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
        course_run (CourseRun): A CourseRun object.
    """
    if not mmtrack.course_has_exam(course_run) and mmtrack.has_paid(course_run.edx_course_key):
        raise ExamAuthorizationException(
            message_not_eligible_template.format(
                user=mmtrack.user.username,
                course_id=course_run.edx_course_key
            )
        )

    # if user paid for a course then create his exam profile if it is not creaated yet.
    ExamProfile.objects.get_or_create(profile=mmtrack.user.profile)

    # if user passed the course and currently not authorization for that run then give
    # her authorizations.
    ok_for_authorization = (
        mmtrack.has_passed_course(course_run.edx_course_key) and
        not ExamAuthorization.objects.filter(
            user=mmtrack.user,
            course=course_run.course,
            date_first_eligible__lte=now,
            date_last_eligible__gte=now
        ).exists()
    )

    if not ok_for_authorization:
        raise ExamAuthorizationException(
            message_not_passed_or_exist_template.format(
                user=mmtrack.user.username,
                course_id=course_run.edx_course_key
            )
        )

    ExamAuthorization.objects.create(
        user=mmtrack.user,
        course=course_run.course,
        date_first_eligible=now,
        date_last_eligible=add_year(now)
    )
    log.info(
        '[Exam authorization] user "%s" is authorize for exam the for course id "%s"',
        mmtrack.user.username,
        course_run.edx_course_key
    )
```

exams/utils.py
```diff
@@ -59,7 +68,105 @@ def authorize_for_exam(mmtrack, course_run):
...
+        exam_series_code__isnull=False
+    )
+    if program_id:
+        programs = programs.filter(id=program_id)
+
+    if not programs.exists():
+        if program_id:
+            raise CommandError('[Exam authorization] Invalid program id: %s', program_id)
+        else:
+            raise CommandError('[Exam authorization] Program(s) are not available for exam authorization.')
```
giocalitri: `CommandError` is an exception to be used in the management commands, not here.

exams/utils.py
```diff
@@ -59,7 +68,105 @@ def authorize_for_exam(mmtrack, course_run):
...
+    Args:
+        program(courses.models.Program): Program object.
+        username(str): User name (optional) whom you want to authorize for exam(s)
+    """
+    program_enrollment_qset = ProgramEnrollment.objects.filter(program=program)
+    if username:
+        program_enrollment_qset = program_enrollment_qset.filter(user__username=username)
+
+    if not program_enrollment_qset.exists():
+        if username:
```
giocalitri: please be explicit
`if username is not None`

exams/utils.py
```diff
@@ -59,7 +68,105 @@ def authorize_for_exam(mmtrack, course_run):
...
+
+def authorize_for_exam_given_program(program, username=None):
+    """
+    Authorize user(s) for exam on given program.
+    Args:
+        program(courses.models.Program): Program object.
+        username(str): User name (optional) whom you want to authorize for exam(s)
+    """
+    program_enrollment_qset = ProgramEnrollment.objects.filter(program=program)
+    if username:
```
giocalitri: please be explicit
`if username is not None`

exams/utils.py
```diff
@@ -59,7 +68,105 @@ def authorize_for_exam(mmtrack, course_run):
...
+        program_enrollment_qset = program_enrollment_qset.filter(user__username=username)
+
+    if not program_enrollment_qset.exists():
+        if username:
+            raise CommandError('[Exam authorization] Invalid username: %s', username)
+        else:
+            raise CommandError('[Exam authorization] Eligible users do not exist for exam authorization.')
```
giocalitri: `CommandError` is an exception to be used in the management commands, not here.

exams/management/commands/test_authorization_user_exam.py
```diff
@@ -0,0 +1,148 @@
...
+        with self.assertRaises(CommandError) as ce:
+            self.command.handle("authorization_user_exam", username="invalid_user")
+
+        assert ce.exception.args[0] == '[Exam authorization] Invalid username: %s', "invalid_user"
+        for user in self.users:
+            self.assertFalse(ExamProfile.objects.filter(profile=user.profile).exists())
+            self.assertFalse(ExamAuthorization.objects.filter(
+                user=user,
+                course=self.course_run.course
+            ).exists())
```
giocalitri: I think that these tests are not necessary: the only thing to be done to test the management command is to assert that when the command is called, the right backend function is called.
So you can mock the entire backend function and just assume that it works.

dashboard/utils.py
amir-qayyum-khan: yes it is not part of `CourseRun`, but for course to have exam its program must have  exam_series_code.

exams/signals_test.py
amir-qayyum-khan: this change is because signal uses `exams.utils.authorize_for_exam` and we change few things there

amir-qayyum-khan:
Done @giocalitri

exams/management/commands/test_authorization_user_exam.py
```diff
@@ -0,0 +1,86 @@
...
+        cls.program, _ = create_program(past=True)
+        cls.course_run = cls.program.course_set.first().courserun_set.first()
+        cls.users = [profile1.user, profile2.user]
+
+        for user in cls.users:
+            with mute_signals(post_save):
+                ProgramEnrollment.objects.create(user=user, program=cls.program)
+            cls.set_data(user, cls.course_run)
+
+        cls.command = authorization_user_exam.Command()
```
giocalitri: do you still need all this data in the set up?

dashboard/utils.py
```diff
@@ -372,6 +371,18 @@ def get_pearson_exam_status(self):  # pylint: disable=too-many-return-statements
...
```
amir-qayyum-khan: just to check if course has exam we need bool statement that will give us true or false

dashboard/utils.py
amir-qayyum-khan: you are right @giocalitri , i just moved it to course_run

dashboard/utils.py
amir-qayyum-khan: Fixed there was issue in logic

courses/utils.py
```diff
@@ -89,3 +89,16 @@ def get_year_season_from_course_run(course_run):
         return season_year_tuple[0], NUMBER_SEASON_MAP[season_year_tuple[1]]
     else:
         return season_year_tuple
+
+
+def is_blank(string):
```
giocalitri: please do not use reserved keywords (`string` is a module of the standard library)

courses/utils_test.py
```diff
@@ -34,3 +42,13 @@ def test_get_year_season_from_course_run(self):
...
         assert get_year_season_from_course_run(unparseable_test_run1) == ()
         assert get_year_season_from_course_run(unparseable_test_run2) == ()
+
+    @data(
+        (None, True),
+        ("", True),
+        ("Foo", False),
+    )
+    @unpack
+    def test_is_blank(self, string, expected):
```
giocalitri: please do not use reserved keywords

dashboard/utils.py
```diff
@@ -362,7 +361,7 @@ def get_pearson_exam_status(self):  # pylint: disable=too-many-return-statements
             course__program=self.program
         )
 
-        if not any(course_has_exam(self, cr) for cr in course_runs_for_program):
+        if not any(cr.has_exam for cr in course_runs_for_program):
```
giocalitri: nit: use a more readable variable

amir-qayyum-khan:
Done @giocalitri

giocalitri:
Can you please rebase master?

pdpinch:
@amir-qayyum-khan can you update the name (and maybe the description) of this to better describe what's covered by this?

Also, it looks like this needs a rebase

giocalitri:
The code looks fine: doing a manual test

exams/utils.py
```diff
@@ -79,44 +79,151 @@ def authorize_for_exam(mmtrack, course_run):
...
+        ~Q(exam_series_code__exact=""),
+        live=True,
+        exam_series_code__isnull=False
+    )
+    if program_id is not None:
+        programs = programs.filter(id=program_id)
+
+    if not programs.exists():
+        if program_id is not None:
+            raise ExamAuthorizationException('[Exam authorization] Invalid program id: %s', program_id)
```
giocalitri: I did a test without setting the `program.exam_series_code` and the management command correctly raises an exception at this point, but the message is confusing because the program is right, it is the `exam_series_code` that is missing.
Can you please update this error message saying that either the `program id` is wrong or the `exam_series_code` is missing?

giocalitri:
If I do not set a `course.exam_module` I do not get any error and the command seems to run fine but I have nothing in `ExamAuthorization`

exams/utils.py
amir-qayyum-khan: or should i make it, `program: 'x' is not eligible for exam`

giocalitri: better to be explicit in this case because `program: 'x' is not eligible for exam` is still ambiguous

giocalitri:
ok: functionality looks decent at the moment, but I think we should improve the tool to list all the users that have been authorized to have a better understanding of what's going on.
Please open an issue for this.

amir-qayyum-khan:
Done @giocalitri https://github.com/mitodl/micromasters/issues/2535

giocalitri:
what about the message and the other problem?

amir-qayyum-khan:
i am working on those, thats why i did not change the PR status

amir-qayyum-khan:
Done

giocalitri:
👍

giocalitri:
please squash

PR #2451, opened 2017-01-27
testing out solutions to yarn build problem

aliceriot:
This just upgrades `node-sass` and `sass-loader`, which gets rid of two of the unmet peer dependency errors we were getting with yarn + heroku. We don't _need_ to do this but I think it would be a good idea.


aliceriot:
@noisecapella I opened this to try to resolve the build issues, but I think it would be a good idea to merge anyhow (even though the build is now fixed).

PR #2406, opened 2017-01-20
Add handling for next parameter

noisecapella:
#### What are the relevant tickets?
Fixes #2385 

#### What's this PR do?
In the homepage the `next` query parameter is set in the link to login to edx. If the `next` query parameter is set for the link it is used as the redirect URI after the signin is completed. Since Django sets `?next` if the user attempts to go somewhere that requires a login, this should allow the user to just click the login link to get back where they were, including all relevant query parameters.

#### How should this be manually tested?
Log out, then go to `/dashboard/?xy=z`. You should be redirected to the home page. Click the login link to log in, then you should be redirected to edX and be directed back to end up at `/dashboard/?xy=z`.

PR #2458, opened 2017-01-30
Don't needlessly set a `next` query param

singingwolfboy:
Follow-up from #2406. If the URL doesn't have a `next` query parameter, don't set a `next` query parameter when logging into Micromasters.

### DIFF
--- a/static/js/containers/SignupDialog.js
+++ b/static/js/containers/SignupDialog.js
@@ -23,7 +23,11 @@ const SignupDialog = ({
   open,
   setDialogVisibility,
 }: signupProps) => {
-  let next = URI(window.location.search).query(true).next;
+  let loginUrl = URI('/login/edxorg');
+  const nextUrl = URI(window.location.search).query(true).next;
+  if ( nextUrl ) {
+    loginUrl = loginUrl.setSearch("next", nextUrl);
+  }
   return <Dialog
     titleClassName="dialog-title"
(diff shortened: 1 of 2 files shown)


pdpinch:
what problem were we solving with this? @noisecapella ?

singingwolfboy:
@pdpinch If you visit the home page without a `next` URL query parameter, the "Continue with edX" button gets a `next=undefined` URL query parameter, which it shouldn't have.
screen shot 2017-01-30 at 11 15 46 am

pdpinch:
I understand the side effect you're addressing DB. I was wondering why we made the change in #2406 at all?

singingwolfboy:
@pdpinch This is so that if an anonymous user tries to access the Micromasters site with a coupon code in the URL, it will be retained throughout the login process.

noisecapella:
Thanks for catching this. Code looks good, doing a functional review...

pdpinch:
So I'm late to this party, but our plan was just to tell users to make sure they logged in first before trying to use the coupon URL. Does this actually work with coupons URLs?

pdpinch:
This isn't on the critical path to getting coupons working. Not a priority at the moment.

singingwolfboy:
@pdpinch I'm not quite sure what you're asking. This should work with _any_ `next` parameter in the URL, whether it's related to coupons or not.

PR #2459, opened 2017-01-30
Alert anonymous user if they try to use a coupon

### LINKED ISSUE #2405: Coupons: tell anonymous user to login first
When a user links to a coupon but they haven't logged in, we should return a message asking them to do so (and create an account if necessary) 

Related to #2323

singingwolfboy:
Fixes #2405.

#### What's this PR do?
If an anonymous user visits the home page with a `coupon` query parameter in the URL, this code will pop a toast-like alert dialog, informing the user that they must log in to use their coupon. If the user does log in, the `coupon` query parameter is retained through the login process, so that it can be automatically applied after the user is logged in.

Log out of the app, and visit the home page with a `coupon` query parameter in the URL. For example: `http://192.168.99.100:8079/?coupon=fake`. Verify that the toast-like notification appears, and that it can be dismissed by clicking the "x". (Unlike other toast notifications in the app, it will not fade away by itself.)

Verify that you can log in, either using the links in the toast-like notification, or using the standing "Log In" and "Sign Up" buttons. When you do, verify that the coupon code is retained: when you land on the dashboard, you should get confirmation dialog that the coupon has been applied, or a toast message that coupon code is invalid.

coupon login notification

### DIFF
--- a/cms/util.py
+++ b/cms/util.py
@@ -5,14 +5,23 @@ from urllib.parse import urlparse, parse_qs
 
 
 def get_coupon_code(request):
-    coupon_code = request.GET.get("coupon", None)
-    if coupon_code:
-        return coupon_code
-    next_url = request.GET.get("next", None)
+    """
+    Get coupon code from an HttpRequest
+
+    Args:
+        request (django.http.request.HttpRequest): An HttpRequest
+
+    Returns:
+        str: A coupon code or None if none found
+    """
+    next_url = request.GET.get("next")
     if not next_url:
         return None
     parsed = urlparse(next_url)
-    coupons = parse_qs(parsed.query).get("coupon", None)
+    path = parsed.path
+    if path not in ("/dashboard", "/dashboard/"):
+        return None
+    coupons = parse_qs(parsed.query).get("coupon")
     if coupons:
         return coupons[0]
     return None
--- a/cms/util.py
+++ b/cms/util.py
@@ -1,0 +1,18 @@
+"""
+Utility functions for CMS models
+"""
+from urllib.parse import urlparse, parse_qs
+
+
+def get_coupon_code(request):
+    coupon_code = request.GET.get("coupon", None)
+    if coupon_code:
+        return coupon_code
+    next_url = request.GET.get("next", None)
+    if not next_url:
+        return None
+    parsed = urlparse(next_url)
+    coupons = parse_qs(parsed.query).get("coupon", None)
+    if coupons:
+        return coupons[0]
+    return None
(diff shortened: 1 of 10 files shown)


static/js/containers/SignupDialog.js
```diff
@@ -24,7 +24,11 @@ const SignupDialog = ({
   setDialogVisibility,
 }: signupProps) => {
   let loginUrl = URI('/login/edxorg');
-  const nextUrl = URI(window.location.search).query(true).next;
+  const urlQuery = URI(window.location.search).query(true);
+  let nextUrl = urlQuery.next;
+  if ( !nextUrl && urlQuery.coupon ) {
+    nextUrl = `/dashboard/?coupon=${urlQuery.coupon}`;
```
noisecapella: You need to encode this: `encodeURIComponent`

static/js/containers/SignupDialog_test.js
```diff
@@ -37,16 +37,38 @@ describe("SignupDialog", () => {
...
     assert.equal(link.getAttribute('href'), `/login/edxorg${queryParams}`);
   });
 
+  it('also checks the coupon query param', () => {
+    window.location = "http://fake/?coupon=whee";
+    helper.store.dispatch(setDialogVisibility(true));
+    renderDialog();
+
+    const link = document.body.querySelector(".signup-dialog a");
+    const expectedNext = "/dashboard/?coupon=whee";
```
noisecapella: Can you change `whee` to include a space or other encodable character?

singingwolfboy: According to [the URLconf for the UserCouponsView](https://github.com/mitodl/micromasters/blob/master/ecommerce/urls.py#L14), the only allowable characters for a coupon are letters, numbers, dashes, and periods. I've changed the code and the test to do all URL encoding using `urijs`, to make sure that any edge cases are handled by that library.

noisecapella:
I think we need to handle `/dashboard/?coupon=code` but it looks like this PR looks for `/?coupon=code`. If the user is anonymous and redirected back to `/` you should be able to find the original URL in `next`

cms/util.py
```diff
@@ -0,0 +1,18 @@
+"""
+Utility functions for CMS models
+"""
+from urllib.parse import urlparse, parse_qs
+
+
+def get_coupon_code(request):
+    coupon_code = request.GET.get("coupon", None)
```
noisecapella: Can we skip this case and just look in `next`? We should only have one URL for a coupon link

PR #2452, opened 2017-01-27
Completed implementation for front-end course team email API

gsidebo:
#### What are the relevant tickets?

Closes #2420 
⚠️️  REBASED ON #2446 ⚠️️

#### What's this PR do?

- Completes implementation for the front-end API for course team emails
- Makes `EmailCompositionDialog` a generic component for use with any type of email (not just search result emails)

#### How should this be manually tested?

Just review the code. The actual end-to-end implementation of the course team email dialog will be in a the PR for #2284. If tests pass and code looks good, this should be ready to merge

#### Any background context you want to provide?

This was done as a separate PR to limit the PR size

PR #2460, opened 2017-01-31
Added course contact email link to the student dashboard

### LINKED ISSUE #2284: Dashboard: contact course team link

gsidebo:

Closes #2284 
⚠️️ REBASED ON #2452 ⚠️️

#### What's this PR do?

Adds the course contact email link to the student dashboard

Probably static/js/containers/DashboardPage.js

1. Seed the DB and set your staff user
1. Set `MAILGUN_RECIPIENT_OVERRIDE` to your email
1. Set any/all 'Analog Learning' courses to have a `contact_email`
1. Enroll in the courses that you've set a contact email for (`alter_data` can be used)
1. Reindex
1. Check the student dashboard and send an email using the "Contact Course Team" link
1. Check that the email you've received is correct

One of the features that #2284 calls for is a message to unpaid users that they can only contact the course team once they've paid (https://github.com/mitodl/micromasters/issues/2284#issuecomment-274866731). I don't have the graphic for that modal, and this PR is sizable enough already that I think it makes sense to create a new issue to implement that new message/modal. 

ss 2017-01-31 at 07 36 50


aliceriot:
functionality looks great!

gsidebo:
the giant python test errors got by me due to this being a mostly-front-end PR. im fixing them up right now

static/js/util/util.js
```diff
@@ -455,3 +455,30 @@ export function getUserDisplayName(profile: Profile): string {
...
 }
+
+function intersperse(arr, elementFunc) {
+  let resultArray = [], i=0;
+  if (i < arr.length)
+    resultArray.push(arr[i++]);
+  while (i < arr.length)
+    resultArray.push(elementFunc(i), arr[i++]);
+  return resultArray;
+}
```
aliceriot: I think you might be able to use `R.chain` productively here - something like:

```js
const intersperse = R.curry((cb, xs) => {
  let inter = x => [x, cb(x)];
  return R.chain(inter, xs);
});
```

or in a more point-free form:

```js
const intersperse = cb => {
  let inter = x => [x, cb(x)];
  return R.chain(inter);
};
```

static/js/util/util.js
```diff
@@ -455,3 +455,30 @@ export function getUserDisplayName(profile: Profile): string {
...
+  return intersperse(components, (i) => ( <span key={`separator-${i}`}>{ separator }</span> ));
+}
+
+/**
+ * Returns the value of an object's property by name, or a default if (a) the property name
+ * is null/undefined or (b) the property doesn't exist in the object.
+ */
+export function getPropertyOrDefault(obj: Object, property: ?string, defaultValue: any) {
+  return _.get(obj, property, defaultValue);
+}
```
aliceriot: Do we need this function? Just curious why we wouldn't just use `_.get` straight up.

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -23,32 +24,35 @@ import {
...
-    };
-  };
-
-  it('shows the course title and a link to view the course on edX', () => {
+  let getElements = (renderedComponent) => ({
+    titleText: renderedComponent.find(".course-title").text(),
+    edxLink: renderedComponent.find("a.view-edx-link"),
+    contactLink: renderedComponent.find("a.contact-link"),
+    detailsText: renderedComponent.find(".details").text()
+  });
```
aliceriot: :100:

static/js/components/dashboard/CourseDescription_test.js
```diff
@@ -23,32 +24,35 @@ import {
...
+
+  let renderCourseDescription = (courseRun, courseTitle, canContactCourseTeam) => (
+    shallow(
+      <CourseDescription
+        courseRun={courseRun}
+        courseTitle={courseTitle}
+        canContactCourseTeam={R.isNil(canContactCourseTeam) ? true : canContactCourseTeam}
```
aliceriot: maybe a default argument of `false` would be a little clearer for the `canContactCourseTeam` arg?

static/js/components/dashboard/CourseRow.js
```diff
@@ -45,6 +46,16 @@ export default class CourseRow extends React.Component {
       []
   );
 
+  hasPaidForAnyCourseRun = (course: Course): boolean => (
+    (course.runs.length > 0) ?
+      R.any(R.propEq('has_paid', true), course.runs) :
+      false
```
aliceriot: `R.any` is safe to call on `[]`, it will return false in that case (so we don't need the `course.runs.length > 0` check)

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -28,25 +31,49 @@ type EmailDialogProps = {
...
+      <Cell col={1} align={"middle"} className="subheading-recipient">TO:</Cell>
+      <Cell col={11}><h5 className="subheading rounded">{ subheading }</h5></Cell>
+    </Grid>
+  </div>
+);
+
+const emailTypeRendererMap = {
+  [SEARCH_EMAIL_TYPE]: renderSimpleSubheading,
+  [COURSE_EMAIL_TYPE]: renderRoundedSubheading
+};
```
aliceriot: It might be nice to use a `Map` here, then down below where you're pulling the render function out of here you could just do:

```js
let renderFunc = emailTypeRenderedMap.get(renderType) || renderSimpleSubheading;
```

(taking advantage of that fact that if you call `.get` with a key the map in question does not contain it returns `undefined`)

aliceriot:
cool - I'm looking over the JS rn.

static/js/containers/LearnerSearchPage.js
```diff
@@ -68,15 +76,15 @@ class LearnerSearchPage extends React.Component {
...
+    const { dispatch, searchResultEmail: { inputs, params } } = this.props;
     let errors = emailValidation(inputs);
     dispatch(updateEmailValidation({type: SEARCH_EMAIL_TYPE, errors: errors}));
     if ( R.isEmpty(errors) ) {
       dispatch(
         sendSearchResultMail(
           inputs.subject || '',
           inputs.body || '',
-          searchKit.query.query
+          params.query
```
aliceriot: this is a cleaner way to do this!

static/js/components/email/EmailCompositionDialog.js
gsidebo: > if you call .get with a key the map in question does not contain it returns undefined

an Object does the same thing. `emailTypeRendererMap['nonsense']` == `undefined`. i jumped through these hoops mostly b/c of flow typing. `emailTypeRendererMap[renderType]` fails because `renderType` can be `undefined`. i want to get the default function if the key doesn't exist _or_ the key is null/undefined. addressing your comment above, i created for a couple reasons: (1) `getPropertyOrDefault` is a little bit clearer about intent than using `_.get` directly, and (2) there's a lot of functional overlap between ramda and lodash, so on the very very remote chance that we decided we wanted to get rid of lodash (or just import specific functions), we could change the implementation of that function. neither are terribly compelling reasons. i think it's likely for a developer to need to check docs whether they see this function or `_.get` with a third arg, and it's just a bit nicer to have that function documented locally, so i'm slightly leaning towards keeping it. let me know what you think

static/js/util/util.js
gsidebo: i created this for a couple reasons: (1) getPropertyOrDefault is a little bit clearer about intent than using _.get directly, and (2) there's a lot of functional overlap between ramda and lodash, so on the very very remote chance that we decided we wanted to get rid of lodash (or just import specific functions), we could change the implementation of this function. neither are terribly compelling reasons. i think it's likely for a developer to need to check docs whether they see this function or _.get with a third arg, and it's just a bit nicer to have that function documented locally, so i'm slightly leaning towards keeping it. let me know what you think

aliceriot: Hm, well we do use `_.get` in a bunch of places with a third argument already, so I'm not so sure this would give us a big win in terms of swapping out lodash functionality for lodash. I'm also not so opinionated on this, so if you wanna keep it go for it.

static/js/components/dashboard/CourseRow.js
```diff
@@ -47,9 +47,7 @@ export default class CourseRow extends React.Component {
   );
 
   hasPaidForAnyCourseRun = (course: Course): boolean => (
-    (course.runs.length > 0) ?
-      R.any(R.propEq('has_paid', true), course.runs) :
-      false
+    R.any(R.propEq('has_paid', true), course.runs) || false
```
aliceriot: I don't think we need `|| false` here, unless I'm missing something - if any elements in `course.runs` fail the predicate function (`R.propEq('has_paid', true)`) or if `course.runs` is `[]`, then we'll be doing `false || false // -> false`, and in cases where one of the elements passes we'll be doing `true || false // -> true`.

gsidebo: yup, you're absolutely right. just fixed

aliceriot:
Looks like there's an error due to `ramda` not being imported on `DashboardPage` ?

PR #2465, opened 2017-01-31
Correctly export user profiles with blank romanized name fields

### LINKED ISSUE #2463: Exams: Profile export should fall back to standard name fields if roman ones are empty
Basically, [these two](https://github.com/mitodl/micromasters/blob/master/exams/pearson/writers.py#L145-L146) exported TSV fields need to get converted to a function that falls back to `profile.first_name` and `profile.last_name` if `profile.romanized_first_name` and `profile.romanized_last_name` respectively are blank. You can see an example of how such a function would be written in the last 4 fields in that writer's configuration (`Country`, `Phone`, etc).

singingwolfboy:
Fixes #2463. When exporting a profile for Pearson, fall back on the 'first_name' and 'last_name' fields, if the 'romanized_first_name' and 'romanized_last_name' fields aren't set.

### DIFF
--- a/exams/pearson/writers_test.py
+++ b/exams/pearson/writers_test.py
@@ -195,53 +195,41 @@ class CDDWriterTest(TSVWriterTestCase, TestCase):
         self.cdd_writer = CDDWriter()
         super().setUp()
 
-    def test_first_name(self):  # pylint: disable=missing-docstring,no-self-use
+    @ddt.data(
+        ("Jekyll", None, "Jekyll"),
+        (None, "Hyde", "Hyde"),
+        ("Jekyll", "Hyde", "Hyde"),
+    )
+    @ddt.unpack  # pylint: disable=no-self-use
+    def test_first_name(self, unromanized, romanized, expected):
+        """
+        Test that the `first_name` method prefers the `romanized_first_name`
+        field, and falls back on `first_name` field.
+        """
         with mute_signals(post_save):
             profile = ProfileFactory(
-                first_name="Jekyll",
-                romanized_first_name=None,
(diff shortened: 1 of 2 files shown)


exams/pearson/writers_test.py
```diff
@@ -195,13 +195,60 @@ def setUp(self):  # pylint: disable=missing-docstring
         self.cdd_writer = CDDWriter()
         super().setUp()
 
+    def test_first_name(self):  # pylint: disable=missing-docstring,no-self-use
```
rhysyngsun: We usually put docstrings on all our tests. Also, I think you could refactor these into two test methods and throw `ddt` at it with `name`, `romanized`, `expected` values so you don't have to write so many docstrings. :-)

singingwolfboy: I'm not convinced that we _need_ to put docstrings on every test -- if a test is short, and the function name is descriptive, that's enough. I've made #2466 to remove the pylint requirement, and I'm happy to argue the point there, if you want. :)

I certainly _could_ use `ddt` here, but I feel like that would make it difficult to understand what the tests are actually doing. `ddt` is best used when you're testing a piece of functionality that should have the same output regardless of what sort of input you throw at it. These tests are asserting that the output should be different based on the input.

noisecapella: Why not have the expected output as an extra parameter in the ddt arguments?

singingwolfboy: Alright, I switched to `ddt`. Does that look better?

rhysyngsun:
:+1: assuming tests pass

PR #2466, opened 2017-01-31
Configure pylintrc to be more accepting

singingwolfboy:
Double-underscore methods and `setUp`/`tearDown` methods for tests don't _always_ need docstrings, and our code is so littered with `# pylint: disable=no-self-use` that we might as well just turn off the check.

No manual testing needed. If the tests pass on Travis, it's good to go.


noisecapella:
I feel like we should keep docstrings for tests. They aren't always necessary but it can be useful to describe what the test is intended for. IMO it's better to have the docstring rule and disable docstring checks for certain trivial tests than to omit the rule and put that burden on the reviewer to decide which tests need docstrings

singingwolfboy:
@noisecapella I've changed this pull request so that it doesn't require docstrings on `setUp` and `tearDown` methods, but it _does_ require them on test methods. Does that work for you?

noisecapella:
Yes, that's fine with me

amir-qayyum-khan:
is their any issue for this task?

singingwolfboy:
@amir-qayyum-khan: No, I unilaterally decided that the code could use some cleaning.

PR #2478, opened 2017-02-01
Fixing code to run with v1 grades agorithm

giocalitri:
N/A
#### What's this PR do?
Fixed the code to work with the `v1` grade algorithm

Run the tests with the environment variable:
`FEATURE_FINAL_GRADE_ALGORITHM=v0`
and
`FEATURE_FINAL_GRADE_ALGORITHM=v1`

The tests should pass in both cases.

### DIFF
--- a/dashboard/utils_test.py
+++ b/dashboard/utils_test.py
@@ -16,21 +16,19 @@ import ddt
 from courses.factories import ProgramFactory, CourseFactory, CourseRunFactory
 from dashboard.api_edx_cache import CachedEdxUserData
 from dashboard.models import CachedEnrollment, CachedCertificate, CachedCurrentGrade
-from dashboard.utils import (
-    get_mmtrack,
-    MMTrack
-)
+from dashboard.utils import get_mmtrack, MMTrack
 from ecommerce.factories import CoursePriceFactory, LineFactory, OrderFactory
 from ecommerce.models import Order
-from financialaid.factories import TierProgramFactory, FinancialAidFactory
+from exams.factories import ExamProfileFactory, ExamAuthorizationFactory
+from exams.models import ExamProfile, ExamAuthorization
 from financialaid.constants import FinancialAidStatus
+from financialaid.factories import TierProgramFactory, FinancialAidFactory
 from grades.constants import FinalGradeStatus
+from grades.factories import FinalGradeFactory
 from grades.models import FinalGrade
 from micromasters.factories import UserFactory
 from micromasters.utils import load_json_from_file
 from search.base import MockedESTestCase
-from exams.models import ExamProfile, ExamAuthorization
-from exams.factories import ExamProfileFactory, ExamAuthorizationFactory
 
 
 @ddt.ddt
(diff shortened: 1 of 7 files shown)


dashboard/utils_test.py
```diff
@@ -645,6 +653,43 @@ def test_count_courses_passed_fa(self):
         )
         assert mmtrack.count_courses_passed() == 1
 
+    @override_settings(FEATURES={"FINAL_GRADE_ALGORITHM": "v1"})
+    def test_count_courses_passed_normal(self):
```
rhysyngsun: Is it worth suffixing this test with `_v1` just like the others or is your thought that we'll eventually sunset the `_v0` code and you don't want to have to rename everything down the road?

grades/factories.py
```diff
@@ -0,0 +1,26 @@
...
+
+FAKE = faker.Factory.create()
+
+
+class FinalGradeFactory(DjangoModelFactory):
+    """Factory for FinalGrade"""
+    user = SubFactory(UserFactory)
+    course_run = SubFactory(CourseRunFactory)
+    grade = randint(30, 100)/100.0
+    passed = FAKE.boolean()
```
rhysyngsun: For both `grade` and `passed` here, the values are random initialized staticly, is that the intent?

exams/utils.py
```diff
@@ -41,8 +41,12 @@ def authorize_for_exam(mmtrack, course_run):
 
         # if user passed the course and currently not authorization for that run then give
         # her authorizations.
+        try:
+            passed = mmtrack.has_passed_course(course_run.edx_course_key)
+        except:  # pylint: disable=bare-except
+            return
```
rhysyngsun: Is there a reason we don't log the exception here? I worry about this bare exception hiding quiet failures which will make it difficult to debug.

rhysyngsun:
Tests passed for me locally with the two `FEATURE_FINAL_GRADE_ALGORITHM` values. Code makes sense based on our discussion yesterday. A few questions above.

dashboard/utils_test.py
giocalitri: I marked as `v0` the ones that should be removed later when we make `v1` default

grades/factories.py
giocalitri: Not sure if I understood the question

exams/utils.py
giocalitri: actually there is not: I will add it.

grades/factories.py
rhysyngsun: The right operand is only evaluated once during the lifetime of the program, resulting in a fixed value across all calls to the factory. Example:

```python
In [8]: for u in User.objects.all()[:10]:
   ...:     g = FinalGradeFactory(user=u)
   ...:     print((g.passed, g.grade))
   ...: 
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)
(True, 0.91)

```

Is that intended?

giocalitri: no: this should be fixed

giocalitri:
I think I addressed the comments

PR #2485, opened 2017-02-02
update to webpack 2

aliceriot:
this just updates us to webpack 2 and updates some of the related packages (loaders and so on) as well.

if everything boots and works in development and on the PR build I think we're good.


package.json
```diff
@@ -11,27 +11,28 @@
     "autoprefixer": "^6.5.3",
     "awesome-phonenumber": "^1.1.3",
     "babel": "6.5.2",
-    "babel-core": "^6.18.2",
-    "babel-eslint": "^6.1.0",
-    "babel-loader": "^6.2.10",
-    "babel-plugin-jsx": "1.2.0",
+    "babel-core": "latest",
```
noisecapella: Does this mean we're relying on `yarn.lock` to set the version?

webpack.config.shared.js
```diff
@@ -57,7 +57,11 @@ module.exports = {
     exclude: /node_modules/,
     loader: 'babel-loader',
     query: {
-      "presets": ["latest", "react"],
+      "presets": [
+        ["env", { "modules": false }],
```
noisecapella: What does this do?

aliceriot: take a look [here](https://medium.com/webpack/webpack-2-and-beyond-40520af9067f#.8relf4lln) and [here](https://babeljs.io/docs/plugins/preset-env/#options) for info - basically what it does is cause Babel to not transform any `import` or `export` statements to `require` statements. This should allow for tree shaking to work, although (as [this issue](https://github.com/webpack/webpack/issues/2867) points out) tree shaking is pretty broken right now.

aliceriot: because tree shaking needs to be able to perform static analysis of ES6 modules to work

package.json
aliceriot: I suppose so, I just did `yarn add package@latest`, I didn't realize it hadn't set the version number here. I'll update it so that `package.json` has the right version number.

webpack.config.shared.js
noisecapella: Would tree shaking be turned on by this PR? Is tree shaking broken in a way that would affect us?

aliceriot: Yes, it's turned on by this PR (it wasn't before, it is very poorly documented which is quite frustrating). It won't have any effect because (as per the issue above) it is broken currently.

PR #2487, opened 2017-02-02
The Frozen grade should be taken in account before enything else in case they exist

giocalitri:
N/A
#### What's this PR do?
In case there are final grades, they are taken in account in the dashboard before checking anything else. The reason is because we cannot rely on the existence of cached data long term.

`dashboard/api.py`

Create a final grade for the user in a course run with a verified cached enrollment in a non financial aid program. The user should see if she passed (or not, depending on the final grade you created) the course on the dashboard.

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -292,11 +292,7 @@ def get_status_for_courserun(course_run, mmtrack):
     """
     get_grade_algorithm_version = settings.FEATURES.get("FINAL_GRADE_ALGORITHM", "v0")
     if get_grade_algorithm_version == "v1":
-        try:
-            mmtrack.extract_final_grade(course_run.edx_course_key)
-        except FinalGrade.DoesNotExist:
-            pass
-        else:
+        if mmtrack.has_paid_frozen_grade(course_run.edx_course_key):
             return CourseRunUserStatus(CourseRunStatus.CHECK_IF_PASSED, course_run)
     if not mmtrack.is_enrolled(course_run.edx_course_key):
         if mmtrack.has_paid(course_run.edx_course_key):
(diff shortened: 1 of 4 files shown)


noisecapella:
I set my grade to 56% but it shows up as 100% here (the value from edX):
screenshot from 2017-02-02 16-06-05

giocalitri:
@noisecapella I think this is ready again for review: if you already looked at the code, I am sorry but I had to modify a lot of logic, so you should probably re-review from scratch.

noisecapella:
The functionality looks good now. If the feature flag is `v0` I get the edX grade and passed values, and if it's `v1` then I get the final grade values

noisecapella:
When indexing users I get a lot of exceptions in Celery. For example:

    celery_1   | [2017-02-07 16:28:15,062: ERROR/Worker-9] No final grade found for user test_gio_3999 in course course-v1:MIT+GEO102+GEO_T1
    celery_1   | Traceback (most recent call last):
    celery_1   |   File "/src/dashboard/utils.py", line 195, in extract_final_grade
    celery_1   |     status=FinalGradeStatus.COMPLETE,
    celery_1   |   File "/usr/local/lib/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
    celery_1   |     return getattr(self.get_queryset(), name)(*args, **kwargs)
    celery_1   |   File "/usr/local/lib/python3.5/site-packages/nplusone/core/signals.py", line 23, in wrapped
    celery_1   |     ret = func(*args, **kwargs)
    celery_1   |   File "/usr/local/lib/python3.5/site-packages/django/db/models/query.py", line 385, in get
    celery_1   |     self.model._meta.object_name
    celery_1   | grades.models.DoesNotExist: FinalGrade matching query does not exist.

giocalitri:
@noisecapella the error with celery is likely to be caused by something totally independent: I fixed all of them with another PR but probably some other code has been introduced in the meantime.
Will try to debug.

PR #2488, opened 2017-02-02
Added tranformation for exam profile state (#2486)

### LINKED ISSUE #2486: Exams: ensure state values are in 2-character abbreviation format
For the their processing team, appears the data set we're using for state values represents those differently than expected. I had previously tested this but I think we missed this regression when we converted from the geocoding API back to vanilla form fields:

> There was an issue with the import of the State field in the CDD file. It looks like the field was populated with the data: US-MA which does not align with the required input for this field. In looking at the CDI document, the State field should only be populated by the 2-character state abbreviation.

Need to look into this for the US and Canada.

rhysyngsun:

Fixes #2486

#### What's this PR do?

For US/CA profiles, removes the country abbreviation prefix so we end up with just the 2-char state abbr.

`exams/pearson/writers.py`

Update your profile so you country is US/Canada with an appropriate state selected.
Run this against your user:

```python

# fudge the records to sync

tasks.export_exam_profiles()
```

Then fetch the file from the local sftp:

```shell
sftp -P 2022 odl@localhost # password: 123
cd results
ls
get cdd-<random-suffix>.dat
```

Check the file to ensure it has a 2-character state abbreviation for the `State` column

States


exams/pearson/writers.py
```diff
@@ -160,21 +161,60 @@ def __init__(self):
...
+        Returns:
+            str: 2-character state representation if country is US/CA otherwise None
+        """
+        if profile.country not in PEARSON_STATE_SUPPORTED_COUNTRIES:
+            return ''
+
+        state = profile.state_or_territory
+
+        if '-' in state:
+            _, state = state.split('-')
```
gsidebo: i think it makes sense to implement this split as part of `Profile` class. `Profile.territory_state_tuple` or something like that

this could also make sense for `split_phone_number`, but that could be considered for a different PR if you prefer

gsidebo:
looking at functionality now

rhysyngsun:
@gsidebo Refactored to move the impl into `Profile`. I chose to reimplement using a regex because I don't want this code to fail were the formatting of stored values to change down the road. Instead, it'll simply return a `(None, None)` and we'll catch this downstream as an invalid profile.

PR #2490, opened 2017-02-02
Show Coupon code on OrderSummary page

### LINKED ISSUE #2491: Don't show discount calculation when there is no coupon to apply
The discount calculation is displayed when there is no coupon to apply. 
The calculation has to be displayed when there is a coupon applied, even if the discount is $0, and the user has to pay the original price.

annagav:
Fix #2491

#### What's this PR do?
This PR changes the the logic behind the display of the discount calculation to be based on an existing coupon, instead of the discount that is applied.

Displays the coupon code in the calculation.

As a learner, try to 'PAY NOW' for a course without any coupons, with a coupon, and with a coupon that gives $0 discount.
screen shot 2017-02-02 at 12 41 12 pm

### DIFF
--- a/static/js/lib/coupon_test.js
+++ b/static/js/lib/coupon_test.js
@@ -45,9 +45,10 @@ describe('coupon utility functions', () => {
       let stubPrice = 5;
       let calculateRunPriceStub = sandbox.stub(couponFuncs, 'calculateRunPrice');
       calculateRunPriceStub.returns(stubPrice);
-      let actual = calculatePrice(run.id, course.id, price, coupons);
-      assert.equal(actual, stubPrice);
+      let [ actualCoupon, actualPrice ] = calculatePrice(run.id, course.id, price, coupons);
+      assert.equal(actualPrice, stubPrice);
       let coupon = coupons.find(coupon => coupon.program_id === program.id);
+      assert.deepEqual(actualCoupon, coupon);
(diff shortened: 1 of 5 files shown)


pdpinch:
Is this a blocker on distributing the 100% of coupons? I'm not sure what the "before" behavior is. 

We can still release v0.43.0. I'm asking if this blocks us giving the coupons to DEDP for distribution.

annagav:
@pdpinch, This does not block any coupon related activity.

static/js/lib/coupon_test.js
```diff
@@ -45,12 +45,10 @@ describe('coupon utility functions', () => {
       let stubPrice = 5;
       let calculateRunPriceStub = sandbox.stub(couponFuncs, 'calculateRunPrice');
       calculateRunPriceStub.returns(stubPrice);
-      let actual = calculatePrice(run.id, course.id, price, coupons);
-      assert.equal(actual, stubPrice);
+      let [ actualCoupon, actualPrice ] = calculatePrice(run.id, course.id, price, coupons);
+      assert.equal(actualPrice, stubPrice);
       let coupon = coupons.find(coupon => coupon.program_id === program.id);
-      assert.isTrue(calculateRunPriceStub.calledWith(
```
noisecapella: Can we keep this assertion?

PR #2492, opened 2017-02-02
Pearson SSO

### LINKED ISSUE #2408: Exams: SSO with Pearson
### Summary

Part of #2198 

SSO will be implementing visually to the user as one button. The technical implementation behind that button has a little complexity:

- User clicks "Schedule Exam"
- React client requests form data from server. The first two values are used to compute the third so they all need to be generated server-side. These values are:
  - `page_timestamp` - _timestamp for the current time_
  - `session_timeout` - _how long in seconds before the user's session expires on our end - `request.session.get_expiry_age()` seems the most future-proof_
  - `sso_digest` - _computed (details below)_
- React client generates an html form (pseudocode follows) and programmatically submits it.

**Note: the form is generated on-the-fly because the `page_timestamp` needs to be within a short delta of the current time when Pearson received the SSO request.**

### Pseudocode:

```html
<form action="{ settings.EXAMS_SSO_URL }" method="post">
   <input type="submit" name="submit" value="Schedule Exam"/>
    <input type="hidden" name="ACTION" value="scheduleExam" />
    <input type="hidden" name="CLIENT_CODE" value="{ settings.EXAMS_SSO_CLIENT_CODE }" />
    <input type="hidden" name="CLIENT_CANDIDATE_ID" value="{ profile.student_id }" />
    <input type="hidden" name="EXTERNAL_PAGE_TIMESTAMP" value="{ page_timestamp }" />
    <input type="hidden" name="EXTERNAL_ERROR_URL" value="<errorURL>" />
    <input type="hidden" name="EXTERNAL_LOGOUT_URL" value="<logoutURL>" />
    <input type="hidden" name="EXTERNAL_RETURN_URL" value="<returnURL>" />
    <input type="hidden" name="EXTERNAL_TIMEOUT_URL" value="<sessionTimeoutURL>" />
    <input type="hidden" name="EXTERNAL_SESSION_TIMEOUT" value="{ session_timeout }" />
    <input type="hidden" name="EXTERNAL_AUTH_HASH" value="{ sso_digest }>" />
</form>
``` 

#### Notes:
- The `EXTERNAL_*` urls values should be the absolute urls as specified in #2173.
- The two settings values aren't present yet and will need to be added (actual values will be communicated privately/securely).
- The message digest will need to be computed server-side because it contains a secret pass phrase.

### Message Digest Computation

The algorithm for computing the message digest takes the following as input:

  - `settings.EXAMS_SSO_PASSPHRASE`
  - `settings.EXAMS_SSO_CLIENT_CODE`
  - `profile.student_id`
  - `page_timestamp` - _(same value that ends up in the form)_
  - `session_timeout` - _(same value that ends up in the form)_

These values are then concatenated and `sha256` hashed. Pseudocode:

```python
import hashlib

input_val = ''.join([
    settings.EXAMS_SSO_PASSPHRASE,
    settings.EXAMS_SSO_CLIENT_CODE,
    timestamp,
    session_timeout,
    profile.student_id,
])
sso_digest = hashlib.sha256(input_val).hexdigest()

### PR DESCRIPTION
**Pull request by aliceriot**
closes #2408 

this implements the SSO functionality for Pearson.

basically, what you should do to test the functionality is:

1. get your user into a state where you'll be eligible to schedule an exam
    - make an `ExamProfile` for your user's profile and set status to `success`
    - do the same for an `ExamAuthorization`, ensuring that the `date_first_eligible` and `date_last_eligible` values imply that the `ExamAuthorization` is eligible right now.
    - ensure your courses and programs are set up correctly, courses should have an `exam_module` and programs should have an `exam_series_code`
2. get the correct environment variable settings
3. go to `/dashboard` and click the `schedule` button on the final exam card.

### DIFF
--- a/exams/api.py
+++ b/exams/api.py
@@ -1,0 +1,34 @@
+"""
+API for exams app
+"""
+import hashlib
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+
+def sso_digest(client_candidate_id, timestamp, session_timeout):
+    """
+    Compute the sso_digest value we need to send to pearson
+
+    Args:
+        client_candidate_id (int|str): id for the user, usually Profile.student_id
+        timestamp (int): unix timestamp
+        session_timeout (int): number of seconds the session will last
+
+    Returns:
+        str: the computed digest value
+    """
+    if settings.EXAMS_SSO_PASSPHRASE is None:
+        raise ImproperlyConfigured("EXAMS_SSO_PASSPHRASE is not configured")
+    if settings.EXAMS_SSO_CLIENT_CODE is None:
+        raise ImproperlyConfigured("EXAMS_SSO_CLIENT_CODE is not configured")
+
+    data = ''.join([
+        settings.EXAMS_SSO_PASSPHRASE,
+        settings.EXAMS_SSO_CLIENT_CODE,
+        str(timestamp),
+        str(session_timeout),
+        str(client_candidate_id),
+    ]).encode('iso-8859-1')
+    return hashlib.sha256(data).hexdigest()
--- a/exams/urls.py
+++ b/exams/urls.py
@@ -1,12 +1,16 @@
 """URLs for exams app"""
 from django.conf.urls import url
 
-from exams.views import PearsonCallbackRedirectView
+from exams.views import (
+    PearsonCallbackRedirectView,
+    PearsonSSO,
+)
 
 urlpatterns = [
     url(
-        r'^pearson/(?P<status>success|error|timeout|logout)/?$',
+        r'^pearson/(?P<status_code>success|error|timeout|logout)/?$',
         PearsonCallbackRedirectView.as_view(),
         name='pearson_callback'
     ),
+    url(r'^api/v0/pearson/sso/$', PearsonSSO.as_view(), name='pearson_sso_api'),
 ]
--- a/exams/views.py
+++ b/exams/views.py
@@ -1,12 +1,70 @@
 """
 Views for exams app
 """
+from datetime import datetime
+
+from urllib.parse import quote_plus
+import pytz
 from django.views.generic.base import RedirectView
+from django.core.exceptions import ImproperlyConfigured
+from rest_framework import (
+    authentication,
+    permissions,
+    status,
+)
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from exams.api import sso_digest
+from exams.models import ExamProfile
 
 
 class PearsonCallbackRedirectView(RedirectView):
     """
     Redirect from Pearson callbacks to dashboard
     """
-    def get_redirect_url(self, status):  # pylint: disable=arguments-differ
-        return "/dashboard?exam={status}".format(status=status)
+    def get_redirect_url(self, status_code):  # pylint: disable=arguments-differ
+        return "/dashboard?exam={status_code}".format(status_code=quote_plus(status_code))
+
+
+class PearsonSSO(APIView):
+    """
+    Views for the Pearson SSO API
+    """
+    authentication_classes = (
+        authentication.SessionAuthentication,
+        authentication.TokenAuthentication,
+    )
+    permission_classes = (permissions.IsAuthenticated, )
+
+    def get(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-self-use
+        """
+        Request for exam SSO parameters
+        """
+        profile = request.user.profile
+        student_id = profile.student_id
+
+        if not ExamProfile.objects.filter(
+                profile=profile,
+                status=ExamProfile.PROFILE_SUCCESS,
+        ).exists():
+            # UI should in theory not send a user here in this state,
+            # but it's not impossible so let's handle it politely
+            return Response(data={
+                'error': 'You are not ready to schedule an exam at this time',
+            }, status=status.HTTP_403_FORBIDDEN)
+
+        timestamp = int(datetime.now(pytz.utc).timestamp())
+        session_timeout = request.session.get_expiry_age()
+
+        try:
+            digest = sso_digest(student_id, timestamp, session_timeout)
+        except ImproperlyConfigured:
+            return Response(status=500)
+
+        return Response(data={
+            'sso_digest': digest,
+            'timestamp': timestamp,
+            'session_timeout': session_timeout,
+            'sso_redirect_url': request.build_absolute_uri('/'),
+        })
--- a/exams/views_test.py
+++ b/exams/views_test.py
@@ -1,7 +1,20 @@
 """
 Tests for exam views
 """
+from unittest.mock import patch
+
+from datetime import datetime
+import ddt
+from rest_framework import status
+from rest_framework.test import APITestCase
+from django.core.exceptions import ImproperlyConfigured
+from django.core.urlresolvers import reverse
+import pytz
+
+from exams.models import ExamProfile
 from micromasters.test import SimpleTestCase
+from micromasters.factories import UserFactory
+from search.base import MockedESTestCase
 
 
 class PearsonSSOCallbackTests(SimpleTestCase):
@@ -13,7 +26,7 @@ class PearsonSSOCallbackTests(SimpleTestCase):
         Test /pearson/success URL
         """
         response = self.client.get('/pearson/success/')
-        assert response.status_code == 302
+        assert response.status_code == status.HTTP_302_FOUND
         assert response.url == "/dashboard?exam=success"
 
     def test_error(self):
@@ -21,7 +34,7 @@ class PearsonSSOCallbackTests(SimpleTestCase):
         Test /pearson/error URL
         """
         response = self.client.get('/pearson/error/')
-        assert response.status_code == 302
+        assert response.status_code == status.HTTP_302_FOUND
         assert response.url == "/dashboard?exam=error"
 
     def test_timeout(self):
@@ -29,7 +42,7 @@ class PearsonSSOCallbackTests(SimpleTestCase):
         Test /pearson/error URL
         """
         response = self.client.get('/pearson/timeout/')
-        assert response.status_code == 302
+        assert response.status_code == status.HTTP_302_FOUND
         assert response.url == "/dashboard?exam=timeout"
 
     def test_logout(self):
@@ -37,7 +50,7 @@ class PearsonSSOCallbackTests(SimpleTestCase):
         Test /pearson/logout URL
         """
         response = self.client.get('/pearson/logout/')
-        assert response.status_code == 302
+        assert response.status_code == status.HTTP_302_FOUND
         assert response.url == "/dashboard?exam=logout"
 
     def test_not_found(self):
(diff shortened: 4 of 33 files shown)

### CODE REVIEW

File: exams/api.py [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+
+from micromasters.settings import (
+    EXAMS_SSO_PASSPHRASE,
+    EXAMS_SSO_CLIENT_CODE,
+)
+
+
+def sso_digest(student_id, timestamp, session_timeout):
+    """
+    Computer the sso_digest value we need to send to pearson
```
rhysyngsun: Add `Args` and `Returns` docstrings.

exams/urls.py
```diff
@@ -1,12 +1,16 @@
...
+    PearsonSSO,
+)
 
 urlpatterns = [
     url(
         r'^pearson/(?P<status>success|error|timeout|logout)/?$',
         PearsonCallbackRedirectView.as_view(),
         name='pearson_callback'
     ),
+    url(r'^api/v0/pearson/$', PearsonSSO.as_view(), name='pearson_sso_api'),
```
rhysyngsun: Could you namespace this a little more (e.g. `api/v0/pearson/sso/`) in case we end up adding other pearson apis down the road?

exams/views.py
```diff
@@ -10,3 +18,27 @@ class PearsonCallbackRedirectView(RedirectView):
...
+    )
+    permission_classes = (permissions.IsAuthenticated, )
+
+    def post(self, request, *args, **kargs):  # pylint: disable=unused-argument, no-self-use
+        """
+        handle the post request
+        """
+        # import ipdb
+        # ipdb.set_trace()
+        student_id = request.data.get('student_id')
```
rhysyngsun: None of these values should pull from the query, we can't trust user input and this could allow someone to authenticate as another user. Instead, you should pull the student id from the current user's profile. Timestamp can be `datetime.now(pytz.utc)` and `session_timeout` I believe I described on the original issue how to get.

Also, this method should check that there is at least an `ExamProfile` record for the user in a success state. If not it should return an error that the ui can display saying they're not ready for an exam yet (similar messaging to the same state in the exam card is fine I think).

exams/views.py
```diff
@@ -10,3 +18,27 @@ class PearsonCallbackRedirectView(RedirectView):
...
+        timestamp = request.data.get('timestamp')
+        session_timeout = request.data.get('session_timeout')
+
+        digest = sso_digest(student_id, timestamp, session_timeout)
+        return Response(digest)
```
rhysyngsun: This needs to return a dictionary of `timestamp`, `session_timeout`, and `digest` so you can use that in the form.

exams/views.py
aliceriot: Ah ok I just misunderstood and though I would generate those values on the frontend.

aliceriot:
@rhysyngsun I think this is more or less code complete now, although we still have to test it with Pearson to make sure everything's working alright.

exams/views.py
```diff
@@ -1,12 +1,63 @@
...
+        student_id = profile.student_id
+
+        if not ExamProfile.objects.filter(
+                profile=profile,
+                status=ExamProfile.PROFILE_SUCCESS,
+        ).exists():
+            # UI should in theory not send a user here in this state,
+            # but it's not impossible so let's handle it politely
+            return Response(data={
+                'error': 'You are not reday to schedule an exam at this time',
```
rhysyngsun: Typo: `ready`

static/js/lib/pearson.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+/* global SETTINGS:false */
+import R from 'ramda';
+
+export const staticFormEntries: Array<[string, string]> = [
+  ["ACTION", "scheduleExam"],
+  ["CLIENT_CODE", SETTINGS.EXAMS_SSO_CLIENT_CODE],
+  ["EXTERNAL_ERROR_URL", "/pearson/error"],
```
rhysyngsun: These urls need to be absolute.

static/js/lib/pearson_test.js
```diff
@@ -0,0 +1,57 @@
...
+      createFormInput(form, ['foo', 'bar']);
+      let input = form.querySelector('input');
+      assert.equal(input.name, 'foo');
+      assert.equal(input.type, 'hidden');
+      assert.equal(input.value, 'bar');
+    });
+  });
+
+  describe('generateSSOForm', () => {
+    let timestamp = Math.round(new Date().getTime() / 1000);
```
rhysyngsun: Nit: these first two could be `const`.

exams/views.py
aliceriot: haha oops

aliceriot:
@rhysyngsun I updated the callback urls

exams/api.py
```diff
@@ -0,0 +1,28 @@
...
+    Args:
+        client_candidate_id (int|str): id for the user, usually Profile.student_id
+        timestamp (int): unix timestamp
+        session_timeout (int): number of seconds the session will last
+
+    Returns:
+        str: the computed digest value
+    """
+    data = ''.join([
+        settings.EXAMS_SSO_PASSPHRASE,
```
rhysyngsun: Needs to handle these not being set.

static/js/lib/pearson.js
```diff
@@ -0,0 +1,55 @@
...
+  node.type = "hidden";
+  node.name = name;
+  node.value = value;
+  form.appendChild(node);
+});
+
+const createForm = () => {
+  let form = document.createElement('form');
+  document.body.appendChild(form);
+  form.action = SETTINGS.EXAMS_SSO_URL;
```
rhysyngsun: What do we do if this is not set? Because then the action url ends up being `/dashboard/null`?

static/js/lib/pearson.js
```diff
@@ -0,0 +1,55 @@
+// @flow
+/* global SETTINGS:false */
+import R from 'ramda';
+
+export const staticFormEntries: Array<[string, string]> = [
+  ["ACTION", "scheduleExam"],
+  ["CLIENT_CODE", SETTINGS.EXAMS_SSO_CLIENT_CODE],
```
rhysyngsun: Need to handle the situation where this setting isn't set.

aliceriot: sure, I'll add some error handling

exams/api.py
rhysyngsun: What this should do is raise `django.core.exceptions.ImproperlyConfigured` error if either of these are not set. The view should catch those and log with `log.exception` and return some sensible error response to the UI.

aliceriot: Cool, I'm going to add some error handling for the JavaScript stuff and then I'll add this.

exams/views.py
```diff
@@ -1,12 +1,64 @@
...
 
 
 class PearsonCallbackRedirectView(RedirectView):
     """
     Redirect from Pearson callbacks to dashboard
     """
-    def get_redirect_url(self, status):  # pylint: disable=arguments-differ
-        return "/dashboard?exam={status}".format(status=status)
+    def get_redirect_url(self, status_code):  # pylint: disable=arguments-differ
+        return "/dashboard?exam={status_code}".format(status_code=status_code)
```
noisecapella: I know this is technically unnecessary for this case, but can we use `quote_plus` for the query parameter?

exams/views.py
```diff
@@ -1,12 +1,64 @@
...
+                profile=profile,
+                status=ExamProfile.PROFILE_SUCCESS,
+        ).exists():
+            # UI should in theory not send a user here in this state,
+            # but it's not impossible so let's handle it politely
+            return Response(data={
+                'error': 'You are not ready to schedule an exam at this time',
+            }, status=status.HTTP_403_FORBIDDEN)
+
+        timestamp = int(datetime.now(pytz.utc).timestamp())
```
noisecapella: Is using seconds a requirement by pearson, or are milliseconds allowed?

exams/views_test.py
```diff
@@ -13,36 +22,109 @@ def test_success(self):
...
+
+
+@ddt.ddt
+class PearsonSSOViewTests(MockedESTestCase, APITestCase):
+    """
+    Tests for the SSO views
+    """
+    @classmethod
+    def setUpTestData(cls):
+        super(PearsonSSOViewTests, cls).setUpTestData()
```
noisecapella: You can use `super().setUpTestData` here

exams/views_test.py
```diff
@@ -13,36 +22,109 @@ def test_success(self):
...
+        # create an user
+        cls.user = UserFactory()
+
+    def setUp(self):
+        super(PearsonSSOViewTests, self).setUp()
```
noisecapella: Same comment on `super()` here

exams/views_test.py
```diff
@@ -13,36 +22,109 @@ def test_success(self):
...
+        """
+        Test issuing a GET request when user has an ExamProfile in PROFILE_SUCCESS status
+        """
+        ExamProfile.objects.create(
+            profile=self.user.profile,
+            status=ExamProfile.PROFILE_SUCCESS,
+        )
+
+        with patch('exams.views.sso_digest', return_value='test value'):
+            response = self.client.get("/api/v0/pearson/sso/")
```
noisecapella: Can you use `reverse('pearson_sso_api')` here and above?

exams/views_test.py
```diff
@@ -13,36 +22,109 @@ def test_success(self):
...
+            result = response.json()
+            assert response.status_code == status.HTTP_200_OK
+
+            timestamp = result['timestamp']
+            assert isinstance(timestamp, int)
```
noisecapella: Can we test that this is within 5 seconds of now? Do you think that would be worthwhile to test or is it unnecessary?

exams/views.py
rhysyngsun: Seconds are required

noisecapella:
Just some nitpicking, generally looks good to me

exams/views.py
aliceriot: sure

exams/views_test.py
aliceriot: I don't think it's so important, @rhysyngsun thoughts?

aliceriot:
OK all comments should be addressed, and I added a bunch of error handling stuff.

exams/views_test.py
rhysyngsun: I think it's worth testing...while I doubt we'd intentionally introduce such a regression, I'm not certain on whether or not it would present as an obvious error in production were a regression introduced.

Also, it's probably worth testing purely because if it is wrong the impact is very high - no one would be able to SSO to Pearson to schedule exams.

aliceriot: ok, I'll add a test for it

exams/views_test.py
```diff
@@ -13,36 +26,125 @@ def test_success(self):
...
+    def setUp(self):
+        super().setUp()
+        self.client.force_login(self.user)
+
+    def test_sso_get_no_exam_profile(self):
+        """
+        Test issuing a GET request when user has no ExamProfile
+        """
+        with patch('exams.views.sso_digest', return_value='test value'):
+            response = self.client.get("/api/v0/pearson/sso/")
```
noisecapella: Can you use `reverse("pearson_sso_api")` here?

exams/views_test.py
```diff
@@ -13,36 +26,125 @@ def test_success(self):
...
+        """
+        Test issuing a GET request when user has an ExamProfile in PROFILE_SUCCESS status
+        """
+        ExamProfile.objects.create(
+            profile=self.user.profile,
+            status=ExamProfile.PROFILE_SUCCESS,
+        )
+
+        with patch('exams.views.sso_digest', side_effect=ImproperlyConfigured):
+            response = self.client.get("/api/v0/pearson/sso/")
```
noisecapella: Can you use `reverse("pearson_sso_api")` here?

noisecapella:
:+1: from me

Issue #2426, opened 2017-01-24
Exams: Add entries to crontab for Pearson TSV tasks


rhysyngsun:
Part of #1796 

The tasks `exams.tasks.export_exam_profiles` and `exams.tasks_export_exam_authorizations` need to be put in the celery crontab. Pearson processes the files uploaded every 12 hours, but we don't know exactly when those hours fall so we should probably run these tasks once every hour to ensure the lag time is minimal.

Also, since these now will run all the time, we should catch `ImproperlyConfigured` being raised from `get_connection` and log it as a warning so the tasks don't fail and we end up with a sentry alert.

pdpinch:
@rhysyngsun is this a task for @mitodl/devops ?

rhysyngsun:
@pdpinch no, this is a MicroMasters application change, unless we deal with celery crontab differently in production.

amir-qayyum-khan:
@rhysyngsun 

- There are 2 task
  - https://github.com/mitodl/micromasters/blob/master/exams/tasks.py#L41
  - https://github.com/mitodl/micromasters/blob/master/exams/tasks.py#L96
- I need to schedule them to run after each hour. i.e add them [here](https://github.com/mitodl/micromasters/blob/master/micromasters/settings.py#L488)
- Add test like this https://github.com/mitodl/micromasters/blob/master/dashboard/tasks_test.py#L47

rhysyngsun:
@amir-qayyum-khan Correct, I don't think there's any value adding a test, since the task itself is already tested and otherwise you're just testing celerybeat (the test you linked to doesn't test celerybeat schedule anyway), which we should be confident in. I think code review and local testing should be sufficient here.

rhysyngsun:
@amir-qayyum-khan it just occurred to me that these tasks will raise an `ImproperlyConfigured` exception from `get_connection()` in some environments if some of the `settings` values are not set. Could you update the tasks to catch this exception and log it out (I think only as a warning, no need to alert on this).

rhysyngsun:
I added this to my PR as well, you can get an idea of what I'm talking about there: https://github.com/mitodl/micromasters/pull/2447/commits/791926f0a1232b41c552ceca227ee8551a074177

PR #2496, opened 2017-02-02
Added TSV tasks to celery crontab

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2426

#### What's this PR do?
- Adds existing tasks to celery crontab

@pdpinch @rhysyngsun

### DIFF
--- a/exams/pearson/upload.py
+++ b/exams/pearson/upload.py
@@ -28,10 +28,6 @@ def get_connection():
         if getattr(settings, key) is None:
             missing_settings.append(key)
 
-    pearson_exams_sync = settings.FEATURES.get("PEARSON_EXAMS_SYNC", False)
-    if not pearson_exams_sync:
-        missing_settings.append("FEATURE_PEARSON_EXAMS_SYNC")
-
     if missing_settings:
         raise ImproperlyConfigured(
             "The setting(s) {} are required".format(', '.join(missing_settings))
--- a/exams/pearson/upload.py
+++ b/exams/pearson/upload.py
@@ -28,6 +28,10 @@ def get_connection():
         if getattr(settings, key) is None:
             missing_settings.append(key)
 
+    pearson_exams_sync = settings.FEATURES.get("PEARSON_EXAMS_SYNC", False)
+    if not pearson_exams_sync:
+        missing_settings.append("FEATURE_PEARSON_EXAMS_SYNC")
+
     if missing_settings:
         raise ImproperlyConfigured(
             "The setting(s) {} are required".format(', '.join(missing_settings))
--- a/exams/pearson/upload_test.py
+++ b/exams/pearson/upload_test.py
@@ -33,7 +33,6 @@ EXAMS_SFTP_SETTINGS = {
 
 @ddt
 @override_settings(**EXAMS_SFTP_SETTINGS)
-@override_settings(FEATURES={"PEARSON_EXAMS_SYNC": True})
 @patch('pysftp.Connection')
 class PeasonUploadTest(SimpleTestCase):
     """
--- a/exams/pearson/upload_test.py
+++ b/exams/pearson/upload_test.py
@@ -33,6 +33,7 @@ EXAMS_SFTP_SETTINGS = {
 
 @ddt
 @override_settings(**EXAMS_SFTP_SETTINGS)
+@override_settings(FEATURES={"PEARSON_EXAMS_SYNC": True})
 @patch('pysftp.Connection')
 class PeasonUploadTest(SimpleTestCase):
     """
--- a/exams/tasks.py
+++ b/exams/tasks.py
@@ -43,6 +43,9 @@ def export_exam_profiles(self):
     """
     Sync any outstanding profiles
     """
+    if not settings.FEATURES.get("PEARSON_EXAMS_SYNC", False):
+        return None
+
     exam_profiles = ExamProfile.objects.filter(
         status=ExamProfile.PROFILE_PENDING).select_related('profile')
     file_prefix = datetime.now(pytz.utc).strftime(PEARSON_CDD_FILE_PREFIX)
@@ -100,6 +103,9 @@ def export_exam_authorizations(self):
     """
     Sync any outstanding profiles
     """
+    if not settings.FEATURES.get("PEARSON_EXAMS_SYNC", False):
+        return None
+
     exam_authorizations = ExamAuthorization.objects.filter(
         status=ExamAuthorization.STATUS_PENDING).prefetch_related('user__profile', 'course__program')
     file_prefix = datetime.now(pytz.utc).strftime(PEARSON_EAD_FILE_PREFIX)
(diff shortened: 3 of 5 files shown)


micromasters/settings.py
```diff
@@ -493,6 +493,14 @@ def get_var(name, default):
         'task': 'financialaid.tasks.sync_currency_exchange_rates',
         'schedule': crontab(minute=0, hour='3')
     },
+    'export_exam_profiles-every-1-hrs': {
+        'task': 'exams.tasks.export_exam_profiles',
+        'schedule': crontab(minute=0, hour='1')
```
rhysyngsun: This says "run every day at 1am". You need `crontab(minute=0, hour='*')` to run every hour.

rhysyngsun: http://www.nncron.ru/help/EN/working/cron-format.htm

amir-qayyum-khan:
Done @rhysyngsun

amir-qayyum-khan:
@rhysyngsun 
Done with testing, there are files on sftp it is not deleting them after use is this OK?

screen shot 2017-02-06 at 5 50 19 pm

exams/tasks_test.py
```diff
@@ -60,6 +61,25 @@ def test_task_retry(self, task, task_retry):
...
         return self
 
+    @data(
+        (export_exam_authorizations, 'export_exam_authorizations got an ImproperlyConfigured exception: %s'),
+        (export_exam_profiles, 'export_exam_profiles got an ImproperlyConfigured exception: %s')
+    )
+    @unpack
+    def test_task_improperly_config_logged(self, task, expected_warning_message):
+        """
+        Verify that when a retryable error occurs that the task retries
```
rhysyngsun: Please update this docstring.

exams/tasks_test.py
```diff
@@ -60,6 +61,25 @@ def test_task_retry(self, task, task_retry):
...
+        """
+        error = ImproperlyConfigured()
+        with patch("exams.tasks.log") as log, patch('exams.pearson.upload.upload_tsv') as upload_tsv_mock:
+            upload_tsv_mock.side_effect = error
+            task.delay()
+
+        log.warning.assert_called_with(
+            expected_warning_message, error
+        )
+        return self
```
rhysyngsun: Not sure why this `return self` is here, except that it is in the test method above. Can you please remove both?

exams/tasks.py
```diff
@@ -118,6 +121,8 @@ def export_exam_authorizations(self):
         try:
             # upload to SFTP server
             upload.upload_tsv(tsv.name)
+        except ImproperlyConfigured as ice:
+            log.warning('export_exam_authorizations got an ImproperlyConfigured exception: %s', ice)
```
rhysyngsun: After discussion with George [over here](https://github.com/mitodl/micromasters/pull/2447#discussion_r99610047), could you change this to a `log.exception`? Also, could add a check for the feature flag `PEARSON_EXAMS_SYNC` (e.g. `settings.FEATURES.get("PEARSON_EXAMS_SYNC", False)`) so that we can ensure this task doesn't execute and then error in environments we aren't explicitly turning this on and configuring it in?

amir-qayyum-khan:
Done @rhysyngsun

exams/pearson/constants.py
```diff
@@ -11,6 +11,7 @@
     "EXAMS_SFTP_USERNAME",
     "EXAMS_SFTP_PASSWORD",
     "EXAMS_SFTP_UPLOAD_DIR",
+    "PEARSON_EXAMS_SYNC"
```
rhysyngsun: Remove this.

exams/pearson/upload_test.py
```diff
@@ -22,12 +22,15 @@
 EXAMS_SFTP_USERNAME = 'username'
 EXAMS_SFTP_PASSWORD = 'password'
 EXAMS_SFTP_UPLOAD_DIR = 'tmp'
+PEARSON_EXAMS_SYNC = True
```
rhysyngsun: Remove

exams/pearson/upload_test.py
```diff
@@ -22,12 +22,15 @@
...
+
 EXAMS_SFTP_SETTINGS = {
     'EXAMS_SFTP_HOST': EXAMS_SFTP_HOST,
     'EXAMS_SFTP_PORT': EXAMS_SFTP_PORT,
     'EXAMS_SFTP_USERNAME': EXAMS_SFTP_USERNAME,
     'EXAMS_SFTP_PASSWORD': EXAMS_SFTP_PASSWORD,
     'EXAMS_SFTP_UPLOAD_DIR': EXAMS_SFTP_UPLOAD_DIR,
+    'PEARSON_EXAMS_SYNC': PEARSON_EXAMS_SYNC
```
rhysyngsun: Remove

amir-qayyum-khan:
Done @rhysyngsun

exams/pearson/upload.py
```diff
@@ -28,6 +28,10 @@ def get_connection():
         if getattr(settings, key) is None:
             missing_settings.append(key)
 
+    pearson_exams_sync = settings.FEATURES.get("PEARSON_EXAMS_SYNC", False)
```
rhysyngsun: This will result in a logged exception and the point of adding this feature flag was to avoid the exception in the first place. Could you please remove this and add only these lines to the beginning of both tasks:

```python
if not settings.FEATURES.get("PEARSON_EXAMS_SYNC", False):
    return
```

Make sure to add tests to ensure that the body of the function is / is not being run depending on the value of this flag.

amir-qayyum-khan: got it

amir-qayyum-khan: @rhysyngsun So in `upload_tsv` when `get_connection()` is None, what exception we should raise exception or just skip?

rhysyngsun: @amir-qayyum-khan what conditions are you seeing `get_connection()` returning `None`?

amir-qayyum-khan: yes

amir-qayyum-khan:
```python
    try:
        with get_connection() as sftp:
            with sftp.cd(settings.EXAMS_SFTP_UPLOAD_DIR):
                sftp.put(file_path)
    except SSHException as ex:
        raise RetryableSFTPException() from ex
    except AttributeError:
        # should we rise exception or skip.
        return
```

exams/tasks_test.py
```diff
@@ -58,9 +60,40 @@ def test_task_retry(self, task, task_retry):
...
+            export_exam_profiles,
+            'export_exam_profiles is improperly configured, please review require settings.'
+        )
+    )
+    @unpack
+    def test_task_improperly_config_logged(self, task, expected_warning_message):
+        """
+        Verify that when a ImproperlyConfigured error occurs that the task logs exception
+        """
+        error = ImproperlyConfigured()
```
rhysyngsun: Tiny nit: no need for the local here, could inline to the assignment on line 81.

PR #2498, opened 2017-02-02
Cap coupon-adjusted price to between 0 and the full price

### LINKED ISSUE #2438: Coupons: handle coupons which drop price below $0
We may want to just have these coupons set the price to $0, but we might also want to do some server side validation so that the coupon cannot set the price below $0

noisecapella:
Fixes #2438 

#### What's this PR do?
Caps the coupon adjusted price to have a minimum price of $0 and maximum price of the full price.

Create a fixed discount coupon and set it to -987654321. Apply it to the user, then check the dashboard and verify that the price is $0.

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -443,16 +443,20 @@ def calculate_coupon_price(coupon, price, course_key):
     Returns:
         decimal.Decimal: An adjusted price
     """
-    if course_key not in coupon.course_keys:
-        return price
-    if coupon.amount_type == Coupon.PERCENT_DISCOUNT:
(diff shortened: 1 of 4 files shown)


giocalitri:
the code looks fine and it works as expected.
Just one comment about the wording: maybe we should say that the coupon will give "maximum" the fixed discount and not like in the screenshot.
Anyway this should probably go in a different issue 👍 
screen shot 2017-02-03 at 1 21 39 pm

giocalitri:
please rebase and let the tests pass again before merging

PR #2508, opened 2017-02-06
Validate exam profile

### LINKED ISSUE #2339: Exams: validate profile before syncing to Pearson
To cover profiles that already exist in production, we should validate the data before we send to Pearson and if they're invalid mark them as such. We probably want to at least display a message on the dashboard and possibly email them.

Particularly, all the following fields need to be filled out and fall within the CP-1252 character set:

- `first_name`
- `last_name`
- `email`
- `address`
- `city`
- `state`
- `country`
- `postal_code`
- `phone_number`

As a special case, `first_name` and/or `last_name` can be outside the CP-1252 character set, however, `romanized_first_name` and `romanized_last_name` need to have values and within the CP-1252 range.

If a profile is invalid when we go to export it, set its `ExamProfile.status` value to `ExamProfile.PROFILE_INVALID`.

annagav:
Fix #2339
#### What's this PR do?
validates `ExamProfile` data before sending it to Pearson. If profile is invalid then sets status on the `ExamProfile` to 'invalid'.
```
user = User.objects.filter(username='username')[0]
user.profile.address=‘中华人民共和国’
user.profile.save()
```
now the ExamProfile status for this user should be set to 'pending'. Now you can the task:

```
from exams.tasks import *
result = export_exam_profiles.delay()
```
Now the ExamProfile status should be set to 'invalid', and the exported file should not contain this profile. The same should happen if any of the fields are not filled out (except for `postal_code` which is optional)

### DIFF
--- a/exams/tasks.py
+++ b/exams/tasks.py
@@ -19,7 +19,7 @@ from exams.pearson import (
     upload,
     writers,
 )
-from exams.utils import exponential_backoff
+from exams.utils import exponential_backoff, validate_profile
 from micromasters.celery import async
 
 PEARSON_CDD_FILE_PREFIX = "cdd-%Y%m%d%H_"
@@ -44,6 +44,16 @@ def export_exam_profiles(self):
         status=ExamProfile.PROFILE_PENDING).select_related('profile')
     file_prefix = datetime.now(pytz.utc).strftime(PEARSON_CDD_FILE_PREFIX)
 
+    valid_exam_profiles = []
+    for exam_profile in exam_profiles:
+        validated = validate_profile(exam_profile.profile)
+
+        if validated:
+            valid_exam_profiles.append(exam_profile)
+        else:
+            exam_profile.status = ExamProfile.PROFILE_INVALID
+            exam_profile.save()
+
     # write the file out locally
     # this will be written out to a file like: /tmp/cdd-20160405_kjfiamdf.dat
     with tempfile.NamedTemporaryFile(
@@ -53,7 +63,7 @@ def export_exam_profiles(self):
         mode='w',
     ) as tsv:
         cdd_writer = writers.CDDWriter()
-        valid_profiles, invalid_profiles = cdd_writer.write(tsv, exam_profiles)
+        valid_profiles, invalid_profiles = cdd_writer.write(tsv, valid_exam_profiles)
 
         # flush data to disk before upload
         tsv.flush()
--- a/exams/tasks_test.py
+++ b/exams/tasks_test.py
@@ -108,7 +108,7 @@ class ExamProfileTasksTest(TestCase):
             mutate by the time export_exam_profiles returns
             """
             assert hasattr(tsv, 'write')
-            assert isinstance(qs, QuerySet)
+            assert isinstance(qs, list)
 
             profiles = list(qs)
             assert len(profiles) == 10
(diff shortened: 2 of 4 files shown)


exams/utils.py
```diff
@@ -72,3 +73,33 @@ def authorize_for_exam(mmtrack, course_run):
                 mmtrack.user.username,
                 course_run.edx_course_key
             )
+
+
+def match_field(profile, field):
+    """
+    If a field is filled out match it to the CP-1252 character set.
+    """
+    pattern = r'^[\u0020-\u00FF]*$'
```
noisecapella: Where does this pattern come from?

exams/utils.py
```diff
@@ -72,3 +73,33 @@ def authorize_for_exam(mmtrack, course_run):
...
+    """
+    Make sure all the required fields fall within the CP-1252 character set
+    Args:
+        profile (Profile): user profile
+    """
+    fields = ['address', 'city', 'state_or_territory', 'country', 'phone_number']
+    optional = {'first_name': 'romanized_first_name', 'last_name': 'romanized_last_name'}
+
+    if not match_field(profile.user, 'email'):
+        return False
```
noisecapella: It looks like this line is uncovered, can you write a test for it?

exams/utils.py
```diff
@@ -72,3 +73,33 @@ def authorize_for_exam(mmtrack, course_run):
...
+    for key, value in optional.items():
+        if not match_field(profile, key):
+            fields.append(value)
```
noisecapella: This line is uncovered too

exams/tasks.py
```diff
@@ -46,6 +47,16 @@ def export_exam_profiles(self):
...
     file_prefix = datetime.now(pytz.utc).strftime(PEARSON_CDD_FILE_PREFIX)
 
+    valid_exam_profiles = []
+    for exam_profile in exam_profiles:
+        validated = validate_profile(exam_profile.profile)
+
+        if validated:
+            valid_exam_profiles.append(exam_profile)
+        else:
+            exam_profile.status = ExamProfile.PROFILE_INVALID
```
noisecapella: These two lines are uncovered. Can you write a test for this case?

annagav:
@noisecapella, I addressed all comments.

exams/utils.py
```diff
@@ -116,3 +117,33 @@ def authorize_for_exam(mmtrack, course_run):
...
+    reg = re.compile(pattern)
+    value = getattr(profile, field)
+    return (True if reg.match(value) else False) if value else False
+
+
+def validate_profile(profile):
+    """
+    Make sure all the required fields fall within the CP-1252 character set
+    Args:
+        profile (Profile): user profile
```
noisecapella: Can you add a `Returns` section to this docstring?

exams/tasks_test.py
```diff
@@ -143,6 +143,36 @@ def side_effect(tsv, qs):
         for exam_profile in self.expected_in_progress_profiles:
             assert exam_profile in in_progress_profiles
 
+    @patch('exams.tasks.validate_profile', return_value=False, autospec=True)
+    @patch('exams.pearson.upload.upload_tsv')
```
noisecapella: Can you add `autospec=True` here too?

exams/utils_test.py
```diff
@@ -147,3 +150,92 @@ def test_exam_authorization_when_not_passed_course(self):
...
+        ('汉字', 'Andrew', True),
+        ('', 'Andrew', True),
+        ('汉字', '', False),
+        ('', '', False)
+    )
+    @unpack
+    def test_romanized_name(self, name, romanized_name, result):
+        """
+        test romanized name optional/required
+        :return:
```
noisecapella: Can you remove this line?

Issue #2467, opened 2017-01-31
New Enroll Now UX


roberthouse54:
This issue includes the following changes:

1. Add a blue box at the top of the courses card that contains copy about course pricing. This will include the cost of courses, and copy informing users that they can audit classes for free if they want. 
2. For DEDP, if the user has not yet figured out their personal course pricing, this box will tell them they must do so. After the user has determined their personal course pricing, we will display the price in this box, and we will no longer display the personal pricing card.
3. Change the button copy to just say "Enroll Now," and remove the "Enroll and Pay later" link. The course price will no longer be shown on buttons.
3. When the user clicks "Enroll Now" they will see a modal giving them two choices. They can either pay for the course or they can audit for free. If the personal course price has not been set, the Pay Now button will be disabled. 

**Basic View (Non-DEDP program)**
![enroll_now1 2x](https://cloud.githubusercontent.com/assets/20047260/22556421/1220585a-e935-11e6-9bcf-59ebc931d3ac.png)

**Personal Course Pricing - Price Not Set**
![enroll_now2 2x](https://cloud.githubusercontent.com/assets/20047260/22556435/218be142-e935-11e6-9f69-337c29c1723e.png)

**Personal Course Pricing - Price Set**
![enroll_now3 2x](https://cloud.githubusercontent.com/assets/20047260/22556458/3626f600-e935-11e6-8749-f989b1310629.png)

**Enroll Modal**
![enroll_now4 2x](https://cloud.githubusercontent.com/assets/20047260/22556473/404ced88-e935-11e6-8547-d828d0c54133.png)

**Enroll Modal - Price Not Set**
![enroll_now5 2x](https://cloud.githubusercontent.com/assets/20047260/22599706/c017139c-ea05-11e6-829e-8879c930c639.png)

pdpinch:
There's a typo in the last screenshot, "Your need" should be "You need"

This comment can be removed if/when @roberthouse54 updates the screenshot.

singingwolfboy:
After discussing this with @roberthouse54, we agreed that the UX of the entire page could be simplified. Rather than having a card for reporting the financial aid price, a card for reporting the coupon, and a highlighted message with the final cost, we could consolidate all of that information into just the highlighted message.

roberthouse54:
We are going to need a simple UI for the case where a user has decided to audit the course. I think we can just add a "Pay Now" link after where it says "You are auditing this course". We will address this UI more clearly when we switch to the new course status design.

[...]

PR #2522, opened 2017-02-08
CoursePrice.price is a decimal

singingwolfboy:
#2523. Related to #2467.

#### What's this PR do?
This pull request modifies the course price API to return decimal values as strings rather than floating-point numbers, since Javascript/JSON does not have a decimal primitive type. This pull request also modifies the code that fetches from the course price API to automatically parse these strings into [`decimal.js-light`](http://mikemcl.github.io/decimal.js-light/) instances.

`financialaid/serializers.py` to verify that the back-end API serializes decimals correctly. Then `static/js/lib/api.js` to verify that the front-end code parses the serialized decimals correctly.

There should be **no user-visible changes** as a result of this pull request.

### DIFF
--- a/financialaid/serializers.py
+++ b/financialaid/serializers.py
@@ -10,7 +10,9 @@ from rest_framework.fields import (
     CharField,
     ChoiceField,
     FloatField,
-    IntegerField
+    IntegerField,
+    DecimalField,
+    BooleanField,
 )
 
 from courses.models import Program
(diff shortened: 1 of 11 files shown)


noisecapella:
Please use the PR template next time, it's an issue we brought up in retrospective. You don't have to fill out each section of the template.

We should also have an issue for each PR. Can you make an issue for this one?

singingwolfboy:
@noisecapella Thanks for the tip. I've created and linked an issue for this, and I've reformatted the PR description to use the template.

PR #2533, opened 2017-02-09
Skip if the status is not terminal

### LINKED ISSUE #2532: Coupons: reset financial aid if user has a 100% off coupon for whole program
in #2334 we hid the financial aid options for users who apply a 100% off coupon, but we (apparently) didn't address the case where the user had started the financial aid process and was in one of the states that blocks them from paying (e.g. "pending-docs"). 

This results in the user seeing the "Pay $0" button but it's greyed out. 

problem

A possible solution would be to set the financialaid status of these users to `reset`

noisecapella:
Fixes #2532 

#### What's this PR do?
Instead of just checking if the user is in some financial aid state, it explicitly checks if the user is approved or skipped already if the user has a 100% off coupon for a program.

Set up financial aid for a program so that it's pending documents or otherwise waiting for input from the reviewer. Create a 100% off program coupon. Apply that coupon and verify that the financial aid status for that user for that program is now skipped.


amir-qayyum-khan:
looks good to me 👍 💯

Issue #2501, opened 2017-02-03
add robots.txt


pdpinch:
I noticed that we don't have a robots.txt file because it's causing a number of 404 errors from cloudfront. 

We should probably block robots from the following directories:

/dashboard
/learner
/learners

We definitely want to allow crawling of these URLs and directories:

/
/dedp
/scm

I'm not sure if it's necessary to block robots from the API directories. We don't want to interfere with their indexing of /dedp and /scm

PR #2540, opened 2017-02-10
Added robots.txt file

amir-qayyum-khan:
Fixes https://github.com/mitodl/micromasters/issues/2501

#### What's this PR do?
- Adds [rebots.txt](http://www.robotstxt.org/robotstxt.html) file

- open url http://192.168.99.100:8079/robots.txt

#### Useful links:
- http://fredericiana.com/2010/06/09/three-ways-to-add-a-robots-txt-to-your-django-project/
- http://www.robotstxt.org/robotstxt.html

@pdpinch

### DIFF
--- a/ui/templates/robots.txt
+++ b/ui/templates/robots.txt
@@ -1,0 +1,15 @@
+User-agent: *
+Disallow: /dashboard/
+Disallow: /learner/
+Disallow: /learners/
+Disallow: /cms/
+Disallow: /profile/
+Disallow: /financial_aid/
+Disallow: /admin/
+Disallow: /logout/
+Disallow: /404/
+Disallow: /500/
+Disallow: /terms_of_service/
+Disallow: /profiles/
+Disallow: /pearson/
+Disallow: /api/
--- a/ui/templates/robots.txt
+++ b/ui/templates/robots.txt
@@ -7,9 +7,6 @@ Disallow: /profile/
 Disallow: /financial_aid/
 Disallow: /admin/
 Disallow: /logout/
-Disallow: /404/
-Disallow: /500/
-Disallow: /terms_of_service/
 Disallow: /profiles/
 Disallow: /pearson/
 Disallow: /api/
--- a/ui/urls.py
+++ b/ui/urls.py
@@ -3,7 +3,6 @@ URLs for ui
 """
 from django.conf.urls import url
 from django.contrib.auth import views as auth_views
-from django.views.generic import TemplateView
 
 from ui.url_utils import (
     DASHBOARD_URLS,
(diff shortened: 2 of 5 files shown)


ui/templates/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+Disallow: /dashboard/
+Disallow: /learner/
+Disallow: /learners/
+Disallow: /api/
```
amir-qayyum-khan: i dont think so apis are useful for crowlers or rebots? because they are session oriented. No one can use it without login.
@pdpinch @giocalitri

pdpinch: Makes sense to me. My only concern is whether the program pages (i.e. /dedp/ and /scm/) have a dependency on /api even for not logged in users.

pdpinch: Can you review urls.py for other routes? I'm thinking `/cms/`, `/profile/` and `/financial_aid/` should also be blocked

Is `/admin/` typically blocked?

pdpinch:
Can I get a quick second opinion from someone on @mitodl/devops ?

ui/templates/robots.txt
amir-qayyum-khan: no, need to add admin as well

ui/templates/robots.txt
```diff
@@ -0,0 +1,15 @@
...
+Disallow: /learner/
+Disallow: /learners/
+Disallow: /cms/
+Disallow: /profile/
+Disallow: /financial_aid/
+Disallow: /admin/
+Disallow: /logout/
+Disallow: /404/
+Disallow: /500/
+Disallow: /terms_of_service/
```
amir-qayyum-khan: @pdpinch should i disallow terms_of_service, errors page as well

pdpinch: No. Let's keep all of those. Robots need to get errors, and they know what to do with them.

ui/templates/robots.txt
```diff
@@ -0,0 +1,12 @@
+User-agent: *
+Disallow: /dashboard/
+Disallow: /learner/
+Disallow: /learners/
+Disallow: /cms/
+Disallow: /profile/
+Disallow: /financial_aid/
+Disallow: /admin/
+Disallow: /logout/
+Disallow: /profiles/
```
noisecapella: This doesn't exist, there is only `profile`

ui/urls.py
```diff
@@ -27,4 +28,5 @@
     url(r'^500/$', page_500, name='ui-500'),
     url(r'^learner/(?P<user>[-\w.]+)?/?', UsersView.as_view(), name='ui-users'),
     url(r'^{}$'.format(TERMS_OF_SERVICE_URL.lstrip("/")), terms_of_service, name='terms_of_service'),
+    url(r'^robots\.txt$', TemplateView.as_view(template_name='robots.txt', content_type='text/plain')),
```
noisecapella: `robots.txt` is not a template. Why not leave it as a static file and serve it via nginx?

blarghmatey: I would second this sentiment. There's no sense in having to go through Django to serve a text file.

amir-qayyum-khan:
@blarghmatey

blarghmatey:
For the robots.txt to be served in production you also need to add a location block to `nginx.conf.erb` as that is the config file that is used by Heroku.

amir-qayyum-khan:
Done add @blarghmatey

blarghmatey:
👍

Issue #2464, opened 2017-01-31
Format price to account for cents


noisecapella:
We need to adjust the `formatPrice` function to round to the nearest cent, or leave off the cents entirely when it's a whole number

pdpinch:
Would this affect form inputs, or just display? See #2470

singingwolfboy:
I recommend using [`decimal.js-light`](http://mikemcl.github.io/decimal.js-light/) where it makes sense to do so. We're already using this library in our codebase.

PR #2541, opened 2017-02-10
Round to the nearest cent, formatPrice util

amir-qayyum-khan:
Fixes https://github.com/mitodl/micromasters/issues/2464

#### What's this PR do?
- Rounded the nearest cent, or leave off the cents entirely when it's a whole number.

- See order summery or action btn 

@pdpinch @noisecapella

### DIFF
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -378,8 +378,12 @@ export function formatPrice(price: ?string|number|Decimal): string {
   if (price === null || price === undefined) {
     return '';
   } else {
-    const formatedPrice = Decimal(price).todp(2, Decimal.ROUND_DOWN).toFixed(2);
-    let dollars = `$${formatedPrice}`;
+    let formattedPrice: Decimal = Decimal(price);
+    if (!formattedPrice.isInteger()) {
+      formattedPrice =  formattedPrice.toFixed(2, Decimal.ROUND_HALF_UP);
+    }
+
+    let dollars = `$${formattedPrice}`;
     return dollars.replace(/\.00$/,'');
   }
 }
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -378,7 +378,9 @@ export function formatPrice(price: ?string|number|Decimal): string {
   if (price === null || price === undefined) {
     return '';
   } else {
-    return `$${price}`;
+    const formatedPrice = Decimal(price).todp(2, Decimal.ROUND_DOWN).toFixed(2);
+    let dollars = `$${formatedPrice}`;
+    return dollars.replace(/\.00$/,'');
   }
 }
 
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -379,12 +379,13 @@ export function formatPrice(price: ?string|number|Decimal): string {
     return '';
   } else {
     let formattedPrice: Decimal = Decimal(price);
-    if (!formattedPrice.isInteger()) {
+
+    if (formattedPrice.isInteger()) {
+      formattedPrice =  formattedPrice.toFixed(0);
+    } else {
       formattedPrice =  formattedPrice.toFixed(2, Decimal.ROUND_HALF_UP);
     }
-
-    let dollars = `$${formattedPrice}`;
-    return dollars.replace(/\.00$/,'');
+    return `$${formattedPrice}`;
   }
 }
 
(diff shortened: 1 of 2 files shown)


static/js/util/util_test.js
```diff
@@ -485,6 +485,10 @@ describe('utility functions', () => {
   describe('formatPrice', () => {
     it('format price', () => {
       assert.equal(formatPrice(20), "$20");
+      assert.equal(formatPrice(20.005), "$20");
+      assert.equal(formatPrice(20.6059), "$20.6");
```
noisecapella: This should be `20.61`

amir-qayyum-khan: one is ignorable, $20.6 or $20.61 is approximately  same difference is 0.1

pdpinch: In USD prices, we don't have prices that only go to one decimal place. It's either none, or two decimal places (hence "cents" for one-hundredth).

amir-qayyum-khan: got is

amir-qayyum-khan:
Done @noisecapella

static/js/util/util.js
```diff
@@ -378,7 +378,9 @@ export function formatPrice(price: ?string|number|Decimal): string {
   if (price === null || price === undefined) {
     return '';
   } else {
-    return `$${price}`;
+    const formatedPrice = Decimal(price).todp(2, Decimal.ROUND_DOWN).toFixed(2);
+    let dollars = `$${formatedPrice}`;
+    return dollars.replace(/\.00$/,'');
```
noisecapella: Why not just compare `Math.floor(price)` to `price` and see if they're equal?

static/js/util/util.js
```diff
@@ -378,7 +378,9 @@ export function formatPrice(price: ?string|number|Decimal): string {
   if (price === null || price === undefined) {
     return '';
   } else {
-    return `$${price}`;
+    const formatedPrice = Decimal(price).todp(2, Decimal.ROUND_DOWN).toFixed(2);
```
noisecapella: Why did you pick `ROUND_DOWN` here?

static/js/util/util.js
amir-qayyum-khan: not really understood here

static/js/util/util.js
amir-qayyum-khan: self assumption, change 10.6103 to 10.61 etc

noisecapella: According to the documentation the default rounding mode is `ROUND_HALF_UP` which seems more appropriate for this case. The problem with `ROUND_DOWN` is that `10.619` will become `10.61` instead of `10.62` which is closer. Why are we overriding the default rounding setting here?

Also, `toFixed(...)` will round the value. We shouldn't need a separate `round(...)` call here

noisecapella: Can you also rename the variable to `formattedPrice`?

static/js/util/util.js
noisecapella: I looked at the documentation and there is `isInteger()`. Please use that to determine whether to format the cents or just to have the dollar value.

amir-qayyum-khan:
Done @noisecapella

static/js/util/util.js
```diff
@@ -378,7 +378,13 @@ export function formatPrice(price: ?string|number|Decimal): string {
   if (price === null || price === undefined) {
     return '';
   } else {
-    return `$${price}`;
+    let formattedPrice: Decimal = Decimal(price);
+    if (!formattedPrice.isInteger()) {
+      formattedPrice =  formattedPrice.toFixed(2, Decimal.ROUND_HALF_UP);
```
noisecapella: Please leave out `Decimal.ROUND_HALF_UP`, it is the default value

static/js/util/util.js
```diff
@@ -378,7 +378,13 @@ export function formatPrice(price: ?string|number|Decimal): string {
...
+    }
+
+    let dollars = `$${formattedPrice}`;
+    return dollars.replace(/\.00$/,'');
```
noisecapella: Please don't use a regular expression in this function. Just use `formattedPrice.toFixed(2)` if `formattedPrice.isInteger()` else `formattedPrice.toFixed(0)`

amir-qayyum-khan:
Done @noisecapella

PR #2546, opened 2017-02-10
Added exception chaining for FreezeGradeFailedException (#2503)

### LINKED ISSUE #2503: Fix bare exceptions on grades/api.py
For the cases where we reraise the exception we should connect the original exception with the new one using the syntax `raise FreezeGradeFailedException(...) from ex`

rhysyngsun:
Fixes #2503 

#### What's this PR do?
Adds exception chaining to two raised exceptions so we get a full traceback.

This doesn't functionally change any behavior, but rather adds extra debug info to an exception we're already raising. Since these lines are covered a passing test should be sufficient.

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -126,7 +126,6 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
     Returns:
         None
     """
-    # pylint: disable=bare-except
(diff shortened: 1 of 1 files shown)


noisecapella:
Did you mean to merge this?

PR #2547, opened 2017-02-10
Don't send email on order cancellations, ignore duplicate cancellations

### LINKED ISSUE #2272: EcommerceException: Order 264 is expected to have status 'created'
@noisecapella - any idea what's happening here? Maybe a race condition? 

the odlcyber account received two emails with "decision=CANCEL" within the same minute. 

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/193474152/

```
DoesNotExist: Order matching query does not exist.
  File "ecommerce/api.py", line 271, in get_new_order_by_reference_number
    return Order.objects.get(id=order_id, status=Order.CREATED)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 385, in get
    self.model._meta.object_name

EcommerceException: Order 264 is expected to have status 'created'
(8 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 477, in dispatch
    response = self.handle_exception(exc)
  File "rest_framework/views.py", line 437, in handle_exception
    self.raise_uncaught_exception(exc)
  File "rest_framework/views.py", line 474, in dispatch
    response = handler(request, *args, **kwargs)
  File "ecommerce/views.py", line 97, in post
    order = get_new_order_by_reference_number(reference_number)
  File "ecommerce/api.py", line 273, in get_new_order_by_reference_number
    raise EcommerceException("Order {} is expected to have status 'created'".format(order_id))

EcommerceException: Order 264 is expected to have status 'created'
```

noisecapella:
Fixes #2272 

#### What's this PR do?
- Ignores duplicate Cybersource cancellation messages
- Doesn't sent email alerting to cancellations

This is difficult to test since we're not sure why Cybersource sends duplicate Cybersource cancellation emails. For the other part of this, it's probably easiest just to test this on RC since all the environment variables and Cybersource configuration is already done there

### DIFF
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -445,13 +445,15 @@ class ReferenceNumberTests(MockedESTestCase):
         course_run, user = create_purchasable_course_run()
         order = create_unfulfilled_order(course_run.edx_course_key, user)
 
-        for status in (status for status in Order.STATUSES if status != Order.CREATED):
-            order.status = status
-            order.save()
+        order.status = Order.FAILED
+        order.save()
 
-            with self.assertRaises(EcommerceException) as ex:
-                get_new_order_by_reference_number(make_reference_id(order))
-            assert ex.exception.args[0] == "Order {} is expected to have status 'created'".format(order.id)
+        with self.assertRaises(EcommerceException) as ex:
+            # change order number to something not likely to already exist in database
+            order.id = 98765432
+            assert not Order.objects.filter(id=order.id).exists()
+            get_new_order_by_reference_number(make_reference_id(order))
+        assert ex.exception.args[0] == "Unable to find order {}".format(order.id)
 
 
 class EnrollUserTests(MockedESTestCase):
--- a/ecommerce/views.py
+++ b/ecommerce/views.py
@@ -26,6 +26,7 @@ from ecommerce.api import (
     make_dashboard_receipt_url,
     pick_coupons,
 )
+from ecommerce.exceptions import EcommerceException
 from ecommerce.models import (
     Coupon,
     Order,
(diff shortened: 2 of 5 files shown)


ecommerce/api_test.py
```diff
@@ -445,13 +445,15 @@ def test_status(self):
...
+        order.status = Order.FAILED
+        order.save()
 
-            with self.assertRaises(EcommerceException) as ex:
-                get_new_order_by_reference_number(make_reference_id(order))
-            assert ex.exception.args[0] == "Order {} is expected to have status 'created'".format(order.id)
+        with self.assertRaises(EcommerceException) as ex:
+            # change order number to something not likely to already exist in database
+            order.id = 98765432
+            assert not Order.objects.filter(id=order.id).exists()
```
giocalitri: I would have done `assert Order.objects.filter(id=order.id).exists() is False`, but probably this is irrelevant.

ecommerce/views.py
```diff
@@ -148,29 +149,36 @@ def post(self, request, *args, **kwargs):
         receipt.order = order
         receipt.save()
 
-        if request.data['decision'] != 'ACCEPT':
-            # This may happen if the user clicks 'Cancel Order'
+        decision = request.data['decision']
+        if order.status == Order.FAILED and decision == 'CANCEL':
+            # This is a duplicate message, ignore since it's already handled
+            return Response()
```
giocalitri: this is a `200` by default, right? In any case please be explicit.

ecommerce/views.py
```diff
@@ -148,29 +149,36 @@ def post(self, request, *args, **kwargs):
...
+        elif order.status != Order.CREATED:
+            raise EcommerceException("Order {} is expected to have status 'created'".format(order.id))
+
+        if decision != 'ACCEPT':
```
giocalitri: is it reasonable to move all the CyberSource statuses in a proper class just to have them centralized? Feel free to postpone.

ecommerce/api_test.py
noisecapella: Ok, I'll leave it as is

ecommerce/views.py
noisecapella: Yes, I'll make it explicit

ecommerce/views.py
noisecapella: Sure

noisecapella:
That should be all of them

noisecapella:
Anything else?

giocalitri:
given that there is no manual test to be done 👍

PR #2548, opened 2017-02-10
Only certificates with status "downloadable" are passing certificates

### LINKED ISSUE #2537: modify MMTrack to consider only a certificate status of `downloadable` to be passing
Part of #2538

a certificate from edX doesn't represent a passing grade unless the status is "downloadable" (or "generating," which is a transient state). 

We need to modify MMTrack so that only certificates with `'status': 'downloadable'` are passing.

giocalitri:
closes #2537

#### What's this PR do?
Updates the rule for passing certificates to take in account the status `"downloadable"`.
Other statuses are considered not passing.

set the edx cache last refresh in the future
```
from dashboard.models import *
refr = UserCacheRefreshTime.objects.get(user__username='<your_username>')
refr.enrollment = refr.enrollment + timedelta(days=2)
refr.certificate = refr.enrollment
refr.current_grade = refr.enrollment
refr.save()
```

for the dashboard:

In a non financial aid program, create a verified cached enrollment and a verified certificate with status different from `"downloadable"` for a past course run. 
The course should be Failed.
Modifying just the status to `"downloadable"` in the cached data should turn the course to be Passed.

for the grades:
for each of the previous cases, in a shell run `grades.api._compute_grade_for_non_fa` and check that the result is consistent with the dashboard.

`dashboard/utils.py`
`grades/api.py`

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -167,16 +167,19 @@ class MMTrack:
         enrollment = self.enrollments.get_enrollment_for_course(course_id)
         return enrollment and enrollment.is_verified
 
-    def has_verified_cert(self, course_id):
+    def has_passing_certificate(self, course_id):
         """
-        Returns whether the user has a verified cert.
+        Returns whether the user has a passing certificate.
 
         Args:
             course_id (str): An edX course key
         Returns:
-            bool: whether the user has a verified cert meaning that they passed the course on edX
+            bool: whether the user has a passing certificate meaning that the user passed the course on edX
         """
-        return self.certificates.has_verified_cert(course_id)
+        if not self.certificates.has_verified_cert(course_id):
+            return False
+        certificate = self.certificates.get_verified_cert(course_id)
+        return certificate.status == 'downloadable'
 
     def extract_final_grade(self, course_id):
         """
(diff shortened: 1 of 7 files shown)


dashboard/utils.py
```diff
@@ -176,7 +176,10 @@ def has_verified_cert(self, course_id):
         Returns:
             bool: whether the user has a verified cert meaning that they passed the course on edX
```
singingwolfboy: This comment is no longer accurate. In fact, this function is named `has_verified_cert`, and that _name_ is no longer accurate. I would suggest renaming this function `has_downloadable_cert`.

giocalitri:
@singingwolfboy done

grades/api.py
```diff
@@ -55,8 +55,11 @@ def _compute_grade_for_non_fa(user_edx_run_data):
         UserFinalGrade: a namedtuple of (float, bool,) representing the final grade
             of the user in the course run and whether she passed it
     """
-    run_passed = user_edx_run_data.certificate is not None
-    grade = user_edx_run_data.certificate.grade if run_passed else user_edx_run_data.current_grade.percent
+    run_passed = user_edx_run_data.certificate is not None and user_edx_run_data.certificate.status == 'downloadable'
```
singingwolfboy: Nitpick: I would suggest breaking this into multiple lines to make it easier to read.

```python
run_passed = (
    user_edx_run_data.certificate is not None and
    user_edx_run_data.certificate.status == 'downloadable'
)
```

grades/api.py
```diff
@@ -55,8 +55,11 @@ def _compute_grade_for_non_fa(user_edx_run_data):
...
+    if user_edx_run_data.certificate is not None:
```
singingwolfboy: Nitpick: Why not just check `if user_edx_run_data.certificate:`? It's shorter.

PR #2555, opened 2017-02-13
Refactored email front-end code to use HOC pattern

### LINKED ISSUE #2468: Refactor email-related front-end code
Our first pass at providing email functionality to users was the search result email dialog, which was not designed for reuse. Implementing the course team contact email (#2284) was a time-sensitive issue, and a large amount of code had to be duplicated in order to make things work quickly. Any new use cases we will have for sending emails will need a similarly large-scale duplication. We should look into implementing email functionality  via composition ([example](https://facebook.github.io/react/docs/higher-order-components.html#convention-maximizing-composability))

gsidebo:
Closes #2468 

#### What's this PR do?
Refactors email front-end code to use a higher-order component and a single rendering of the email dialog.

Use the contact course team link on /dashboard and the search result email link on /learners

/static/js/components/email/EmailCompositionDialog.js

[Facebook docs discussing the pattern](https://facebook.github.io/react/docs/higher-order-components.html#convention-maximizing-composability)

### DIFF
--- a/static/js/components/email/EmailCompositionDialog.js
+++ b/static/js/components/email/EmailCompositionDialog.js
@@ -1,108 +1,97 @@
 // @flow
 import React from 'react';
 import Dialog from 'material-ui/Dialog';
-import R from 'ramda';
-import Grid, { Cell } from 'react-mdl/lib/Grid';
 
 import { FETCH_PROCESSING } from '../../actions';
 import { dialogActions } from '../inputs/util';
-import { getPropertyOrDefault } from '../../util/util';
-import { COURSE_EMAIL_TYPE, SEARCH_EMAIL_TYPE } from './constants';
-import type {
-  EmailState,
-  EmailValidationErrors
-} from '../../flow/emailTypes';
+import { isNilOrBlank } from '../../util/util';
+import type { EmailState } from '../../flow/emailTypes';
 
-const showValidationError = R.curry((getter, object: EmailValidationErrors) => {
-  let val = getter(object);
-  if (val !== undefined) {
-    return <span className="validation-error">{ val }</span>;
-  }
-});
+export default class EmailCompositionDialog extends React.Component {
+  props: {
+    dialogVisibility:           boolean,
+    activeEmail:                EmailState,
+    title?:                     string,
+    subheadingRenderer?:        () => React$Element<*>,
+    closeAndClearEmailComposer: () => void,
+    closeEmailComposerAndSend:  () => void,
+    updateEmailFieldEdit:       () => void
+  };
 
-const showSubjectError = showValidationError(R.prop('subject'));
+  showValidationError = (fieldName: string): ?React$Element<*> => {
+    const { activeEmail: { validationErrors } } = this.props;
+    let val = validationErrors[fieldName];
+    if (val !== undefined) {
+      return <span className="validation-error">{ val }</span>;
+    }
+  };
 
-const showBodyError = showValidationError(R.prop('body'));
-
-type EmailDialogProps = {
-  closeEmailDialog: () => void,
-  updateEmailEdit:  Function,
-  open:             boolean,
-  email:            EmailState,
-  sendEmail:        () => void,
-  title?:           string,
-  subheadingType?:  string
-};
-
-const renderSimpleSubheading = (subheading: string) => (
-  <div className="subheading-section">
-    <h5 className="subheading">{ subheading }</h5>
-  </div>
-);
-
-const renderRoundedSubheading = (subheading: string) => (
-  <div className="subheading-section">
-    <Grid noSpacing={true}>
-      <Cell col={1} align={"middle"} className="subheading-to">TO:</Cell>
-      <Cell col={11}><h5 className="subheading rounded">{ subheading }</h5></Cell>
-    </Grid>
-  </div>
-);
-
-const emailTypeRendererMap = {
-  [SEARCH_EMAIL_TYPE]: renderSimpleSubheading,
-  [COURSE_EMAIL_TYPE]: renderRoundedSubheading
-};
-
-const renderSubheading = (renderType: ?string, subheading: ?string) => {
-  if (R.isNil(subheading) || R.isEmpty(subheading)) {
-    return null;
-  } else {
-    let renderFunc = getPropertyOrDefault(emailTypeRendererMap, renderType, renderSimpleSubheading);
-    return renderFunc(subheading);
-  }
-};
-
-const EmailCompositionDialog = (props: EmailDialogProps) => {
-  const {
-    closeEmailDialog,
-    updateEmailEdit,
-    open,
-    email: { inputs, subheading, validationErrors, fetchStatus },
-    sendEmail,
-    title,
-    subheadingType
-  } = props;
-
-  return <Dialog
-    title={title || "New Email"}
-    titleClassName="dialog-title"
-    contentClassName="dialog email-composition-dialog"
-    className="email-composition-dialog-wrapper"
-    open={open}
-    actions={dialogActions(closeEmailDialog, sendEmail, fetchStatus === FETCH_PROCESSING, 'Send')}
-    onRequestClose={closeEmailDialog}
-  >
-    <div className="email-composition-contents">
-      { renderSubheading(subheadingType, subheading) }
-      <textarea
-        rows="1"
-        className="email-subject"
-        placeholder="Subject"
-        value={inputs.subject || ""}
-        onChange={updateEmailEdit('subject')}
-      />
-      { showSubjectError(validationErrors) }
-      <textarea
-        rows="7"
-        className="email-body"
-        placeholder="Type a message"
-        value={inputs.body || ""}
-        onChange={updateEmailEdit('body')}
-      />
-      { showBodyError(validationErrors) }
+  renderDefaultSubheading = (subheading: ?string): React$Element<*> => (
+    <div className="subheading-section">
+      <h5 className="subheading">{ subheading }</h5>
     </div>
-  </Dialog>;
-};
+  );
 
-export default EmailCompositionDialog;
+  renderSubheading = (): ?React$Element<*> => {
+    const {
+      activeEmail: { subheading },
+      subheadingRenderer
+    } = this.props;
+    if (isNilOrBlank(subheading)) {
+      return null;
+    } else {
+      let renderFunc = subheadingRenderer || this.renderDefaultSubheading;
+      return renderFunc(subheading);
+    }
+  };
+
+  render() {
+    if (!this.props.activeEmail) return null;
+
+    const {
+      activeEmail: { fetchStatus, inputs },
+      title,
+      dialogVisibility,
+      closeAndClearEmailComposer,
+      closeEmailComposerAndSend,
+      updateEmailFieldEdit
+    } = this.props;
+
+    return <Dialog
+      title={title || "New Email"}
+      titleClassName="dialog-title"
+      contentClassName="dialog email-composition-dialog"
+      className="email-composition-dialog-wrapper"
+      open={dialogVisibility}
+      actions={
+        dialogActions(
+          closeAndClearEmailComposer,
+          closeEmailComposerAndSend,
+          fetchStatus === FETCH_PROCESSING,
+          'Send'
+        )
+      }
+      onRequestClose={closeAndClearEmailComposer}
+    >
+      <div className="email-composition-contents">
+        { this.renderSubheading() }
+        <textarea
+          rows="1"
+          className="email-subject"
+          placeholder="Subject"
+          value={inputs.subject || ""}
+          onChange={updateEmailFieldEdit('subject')}
+        />
+        { this.showValidationError('subject') }
+        <textarea
+          rows="7"
+          className="email-body"
+          placeholder="Type a message"
+          value={inputs.body || ""}
+          onChange={updateEmailFieldEdit('body')}
+        />
+        { this.showValidationError('body') }
+      </div>
+    </Dialog>;
+  }
+}
--- a/static/js/components/email/hoc.js
+++ b/static/js/components/email/hoc.js
@@ -1,0 +1,114 @@
+import React from 'react';
+import R from 'ramda';
+
+import EmailCompositionDialog from './EmailCompositionDialog';
+import { getDisplayName, isNilOrBlank } from '../../util/util';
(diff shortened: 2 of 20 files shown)


noisecapella:
Can you change `#xxxx` to `Fixes #xxxx`?

static/js/components/email/hoc.js
```diff
@@ -0,0 +1,51 @@
...
+        emailOpenActions={emailOpenActions}
+      />;
+    }
+  }
+
+  if (WithEmailDialog.displayName && !WithEmailDialog.displayName.startsWith('WithEmailDialogs')) {
+    WithEmailDialog.displayName = `WithEmailDialogs(${getDisplayName(WrappedComponent)})`;
+  }
+  return WithEmailDialog;
+});
```
aliceriot: My understanding of the idea behind using an HOC for this was that we'd have something like a `WithEmail` HOC that handled dispatching all of the necessary actions, dealing with the API, etc. so that could be abstracted away from the rendering / UI code. 

It seems a bit of indirection to me to have the `EmailCompositionDialog`, which actually shoulders much of this responsibility, separate from this HOC. Does that make sense? I'm also not super into passing down `dispatch` as a prop - I think I've made my case for avoiding that pattern before though 😆

aliceriot: So to sum up I think this would be more straightforward if the API code (the methods on `EmailCompositionDialog` like `updateEmailFieldEdit` and `closeEmailComposerAndSend` were defined here. I think this would also simplify passing down the action dispatchers for opening the email and such things.

aliceriot: (i.e. it would obviate the need for some of the stuff in `email/lib.js`)

static/js/components/email/lib.js
```diff
@@ -0,0 +1,79 @@
...
+  },
+
+  emailSendAction: (emailState) => (
+    sendSearchResultMail(
+      emailState.inputs.subject || '',
+      emailState.inputs.body || '',
+      emailState.params.searchkit.buildQuery().query
+    )
+  )
+};
```
aliceriot: As I sorted indicated above, I tend to think that these parameters should either be props passed in to the HOC or arguments that are passed to it at call-time (like the `emailType` argument it already takes).

Right now I think declaring these in a separate file (away from where the HOC is actually going to be rendered on the page) is a bit of indirection we could do away with.

static/js/containers/DashboardPage.js
```diff
@@ -119,14 +111,16 @@ class DashboardPage extends React.Component {
...
-    courseTeamEmail:          EmailState,
+    email:                    AllEmailsState,
     documents:                DocumentsState,
     fetchDashboard:           () => void,
     orderReceipt:             OrderReceiptState,
     courseEnrollments:        CourseEnrollmentsState,
     financialAid:             FinancialAidState,
     location:                 Object,
     pearson:                  PearsonAPIState,
+    emailDialog:              React$Element<*>|null,
```
aliceriot: on a side note, I think it would be better to have the HOC render the dialog component, just so the wrapped component doesn't have to worry about it.

static/js/containers/DashboardPage.js
```diff
@@ -119,14 +111,16 @@ class DashboardPage extends React.Component {
...
+    emailOpenActions:         {[emailType: string]: () => void}
```
aliceriot: it also strikes me as a little cleaner to have the HOC only pass down the action which is relevant to the particular wrapped component (indicated by the `emailType` argument passed to the HOC), so that the wrapped component doesn't have to worry about the logic. Although maybe in the future we would want to support opening multiple different email types, but then again if we wanted to add that behavior later it would just be a small refactor.

static/js/reducers/ui.js
```diff
@@ -124,10 +135,29 @@ export const INITIAL_UI_STATE: UIState = {
...
+      },
+    };
+  case HIDE_DIALOG:
+    return {
+      ...state,
+      dialogVisibility: {
+        ...state.dialogVisibility,
+        [action.payload]: false
+      },
+    };
```
aliceriot: this is nice - it would be good to refactor all of the dialogs to use this system at some point (and avoid the like dozens of `showFooDialog` actions we have now).

could you remove the blank lines above and below these two `case` statements though?

static/js/containers/DashboardPage.js
gsidebo: yeah, this was done on purpose. we already have use cases for multiple email types in the same page. passing them in as an object that keys on the email type was the best way i could think of to do that. the export line will look something like this:

```javascript
export default R.compose(
  connect(mapStateToProps),
  withEmailDialog(COURSE_EMAIL_TYPE),
  withEmailDialog(LEARNER_EMAIL_TYPE)
)(DashboardPage);
```

`emailOpenActions` in that case will look like this:

```javascript
{
  'courseTeamEmail': //course team email open function,
  'learnerEmail': //learner-to-learner email open function,
}
```

The HOC adds to that object with each invocation

static/js/reducers/ui.js
gsidebo: yeah, that's the plan. i'm going to create a separate issue to do some of that work

aliceriot: oh yeah, definitely not in this PR, haha

static/js/containers/DashboardPage.js
aliceriot: yup, or `withEmailDialog([TYPE, TYPE])`, whatever makes sense.

static/js/components/email/hoc.js
```diff
@@ -0,0 +1,51 @@
...
+
+      return <WrappedComponent
+        {...this.props}
+        emailDialog={emailDialog}
+        emailOpenActions={emailOpenActions}
+      />;
+    }
+  }
+
+  if (WithEmailDialog.displayName && !WithEmailDialog.displayName.startsWith('WithEmailDialogs')) {
```
noisecapella: Not reviewing the PR but just spotted this. Isn't `displayName` always `WithEmailDialogs` at this point?

gsidebo: my original plan was to include multiple email dialogs on a page by invoking `withEmailDialog()` multiple times, but i'm realizing now that doesn't make as much sense as just invoking it once with an array of email types (if you need more than one). this is going to change. aside: if you're up for the annoyance, i'd love to have you as a second reviewer!

static/js/components/email/lib.js
gsidebo: there are a few reasons that i separated the functionality into 3 files. i'll try to sort out my thoughts and summarize

static/js/components/email/hoc.js
```diff
@@ -0,0 +1,116 @@
...
+            subheadingRenderer={emailConfigs[activeEmailType].renderSubheading}
+          />
+        );
+      }
+
+      render() {
+        let emailOpenActions = {};
+        for (let emailType of Object.keys(emailConfigs)) {
+          emailOpenActions[emailType] =
+            emailConfigs[emailType].emailOpenAction.bind(this, emailType);
```
aliceriot: proposal: I think it would be a bit cleaner if we could avoid the `this` binding here. We could do so by declaring the `emailOpenAction` functions as curried functions with the following type signature:

```
emailOpenAction :: openEmailComposer -> emailType -> ...args
```

then here we could do something like:

```js
let emailOpenActions = Object.assign({}, ...Object.keys(emailConfigs).map(emailType => ({
  [emailType]: emailConfigs[emailType].emailOpenAction(this.openEmailComposer, emailType)
})));
```

what do you think?

static/js/components/email/hoc.js
```diff
@@ -0,0 +1,120 @@
...
+            subheadingRenderer={emailConfigs[activeEmailType].renderSubheading}
+          />
+        );
+      }
+
+      render() {
+        let emailOpenActions = {};
+        for (let emailType of Object.keys(emailConfigs)) {
+          emailOpenActions[emailType] = this.openEmailComposer(emailType);
+        }
```
aliceriot: it seems like we could maybe avoid this step and just pass down `this.openEmailComposer`, trusting the the call-site to specify the right `emailType`

static/js/containers/DashboardPage.js
```diff
@@ -62,45 +58,42 @@ import {
...
-  pearsonSSOInProgress,
-  pearsonSSOFailure,
-  setPearsonError
-} from '../actions/pearson';
-import { generateSSOForm } from '../lib/pearson';
-import type { PearsonAPIState } from '../reducers/pearson';
+import type { CoursePricesState, DashboardState } from '../flow/dashboardTypes';
+import type {
+  AvailableProgram, AvailableProgramsState, CourseEnrollmentsState
+} from '../flow/enrollmentTypes';
```
aliceriot: if possible (and like not too much work) could you revert the changes to imports here to just reflect what's changed?

static/js/components/email/hoc.js
aliceriot: looks like you're already doing that! disregard this one.

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -1,108 +1,99 @@
...
+    const {
+      activeEmail: { subheading },
+      subheadingRenderer
+    } = this.props;
+    if (isNilOrBlank(subheading)) {
+      return null;
+    } else {
+      let renderFunc = subheadingRenderer !== undefined ?
+        subheadingRenderer :
+        this.renderDefaultSubheading;
```
aliceriot: I think my earlier comment got partially swallowed, but could this be expressed as `subheadingRenderer || this.renderDefaultSubheading` ?

gsidebo: 👍  i think i wrote it like this b/c flow was being annoying, but i checked this again and it works

aliceriot: ah flow, always there to make my day a little more difficult

static/js/util/util.js
```diff
@@ -480,10 +480,10 @@ export function renderSeparatedComponents(
...
- * Returns the value of an object's property by name, or a default if (a) the property name
- * is null/undefined or (b) the property doesn't exist in the object.
- */
-export function getPropertyOrDefault(obj: Object, property: ?string, defaultValue: any): any {
-  return _.get(obj, property, defaultValue);
+export function getDisplayName(WrappedComponent: ReactClass<*>) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'Component';
+}
+
+export function isNilOrBlank(input: ?string): boolean {
```
gsidebo: this was another "STFU flow" fix. i'll try this

static/js/components/email/hoc.js
```diff
@@ -0,0 +1,116 @@
...
+        let activeEmailType = this.getActiveEmailType();
+
+        if (!activeEmailType) return null;
+
+        return (
+          <EmailCompositionDialog
+            updateEmailFieldEdit={this.updateEmailFieldEdit}
+            closeAndClearEmailComposer={this.closeAndClearEmailComposer}
+            closeEmailComposerAndSend={this.closeEmailComposerAndSend}
+            dialogVisibility={dialogVisibility[EMAIL_COMPOSITION_DIALOG]}
```
gsidebo: this PR makes sure that only one email dialog exists. the contents of that email dialog change based on the email state and `currentlyActive`

static/js/components/email/hoc.js
```diff
@@ -0,0 +1,116 @@
...
+import EmailCompositionDialog from './EmailCompositionDialog';
+import { getDisplayName, isNilOrBlank } from '../../util/util';
+import { hideDialog } from '../../actions/ui';
+import {
+  startEmailEdit,
+  updateEmailEdit,
+  clearEmailEdit,
+  updateEmailValidation,
+} from '../../actions/email';
+import { showDialog } from '../../actions/ui';
```
gsidebo: good catch

static/js/containers/DashboardPage.js
gsidebo: done. sorry, in the early going i was experiencing some frustration due the scattershot imports

static/js/components/email/hoc.js
gsidebo: so take [this line](https://github.com/mitodl/micromasters/pull/2555/files#diff-19b75ce18fd1eaefee48fc255d15b45bR153) from `LearnerSearch.js` as an example:

```javascript
onClick={R.partial(openSearchResultEmailComposer, [this.searchkit])}
// would become...
onClick={R.partial(openSearchResultEmailComposer, [SEARCH_EMAIL_TYPE, this.searchkit])}
```

is that what you had in mind? i'm not sure about this. i understand the drawback of the current approach - we are partially satisfying the arguments of the function two times, which can make it confusing to figure out what the function needs at any given point during the run time. this change would introduce some problems too. the email type value on the container page would need to match the value at the call site, and there's no way we could 'enforce' those two values being the same. i would rather have the slightly more confusing invocation pattern than open the potential for a bug where the email types don't match. the former can be somewhat easily figured out with a debugger or console.log lines, but the latter could send someone on a very lengthy search across several files to figure out why their expected dialog isn't coming up. we know what type of email we want to open when this function is passed down to the wrapped component's props from the HOC, so i think it makes enough sense to satisfy that argument up front.

gsidebo:
@aliceriot i made some minor changes to the email types and the initial state values in response to the change in the way we're invoking the `withEmailDialog` HOC (this commit: https://github.com/mitodl/micromasters/pull/2555/commits/6195cad1fba6467361bae835993a4126df7c3ca0). since we're passing an object with the email types as keys, it made sense to use those same keys in the `email` state. as a result of that, i had to make a few changes to make flow happy.

static/js/components/email/hoc.js
aliceriot: oh right, that makes sense! haha

aliceriot:
sounds good, taking a look now!

static/js/dashboard_routes.js
```diff
@@ -30,8 +30,7 @@ export const routes = {
       path: 'dashboard',
       getComponent(nextState, cb) {
         import('./containers/DashboardPage')
-          .then(loadRoute(cb))
-          .catch(errorLoading);
+          .then(loadRoute(cb));
```
aliceriot: what's the reason for this change?

gsidebo: oops!

aliceriot: 😃

gsidebo: fixed

PR #2557, opened 2017-02-13
Check for open exchange API URL before requesting it

### LINKED ISSUE #2556: Financial aid task does not check for open exchange URL before making a request
I was seeing odd tracebacks when trying to run the Micromasters app locally:
```
celery_1   | [2017-02-13 15:32:03,082: ERROR/MainProcess] Task financialaid.tasks.sync_currency_exchange_rates[1b5e2284-c582-4f0f-bb95-1b123f3985cf] raised unexpected: MissingSchema("Invalid URL 'Nonelatest.json?app_id=': No schema supplied. Perhaps you meant http://Nonelatest.json?app_id=?",)
celery_1   | Traceback (most recent call last):
celery_1   |   File "/usr/local/lib/python3.5/site-packages/celery/app/trace.py", line 240, in trace_task
celery_1   |     R = retval = fun(*args, **kwargs)
celery_1   |   File "/usr/local/lib/python3.5/site-packages/celery/app/trace.py", line 438, in __protected_call__
celery_1   |     return self.run(*args, **kwargs)
celery_1   |   File "/src/financialaid/tasks.py", line 18, in sync_currency_exchange_rates
celery_1   |     resp = requests.get(CURRENCY_EXCHANGE_RATE_API_REQUEST_URL)
celery_1   |   File "/usr/local/lib/python3.5/site-packages/requests/api.py", line 70, in get
celery_1   |     return request('get', url, params=params, **kwargs)
celery_1   |   File "/usr/local/lib/python3.5/site-packages/requests/api.py", line 56, in request
celery_1   |     return session.request(method=method, url=url, **kwargs)
celery_1   |   File "/usr/local/lib/python3.5/site-packages/requests/sessions.py", line 461, in request
celery_1   |     prep = self.prepare_request(req)
celery_1   |   File "/usr/local/lib/python3.5/site-packages/requests/sessions.py", line 394, in prepare_request
celery_1   |     hooks=merge_hooks(request.hooks, self.hooks),
celery_1   |   File "/usr/local/lib/python3.5/site-packages/requests/models.py", line 294, in prepare
celery_1   |     self.prepare_url(url, params)
celery_1   |   File "/usr/local/lib/python3.5/site-packages/requests/models.py", line 354, in prepare_url
celery_1   |     raise MissingSchema(error)
celery_1   | requests.exceptions.MissingSchema: Invalid URL 'Nonelatest.json?app_id=': No schema supplied. Perhaps you meant http://Nonelatest.json?app_id=?
```

After tracing through the exception, I found that the problem was that the `sync_currency_exchange_rates` task was trying to make an HTTP request to an invalid URL, and the URL was invalid because `settings.OPEN_EXCHANGE_RATES_URL` was unset. If this setting is unset, the system should display a helpful and informative error message, instead of this misleading and confusing one.

singingwolfboy:
Fixes #2556

#### What's this PR do?
Check for `settings.OPEN_EXCHANGE_RATES_URL` and `settings.OPEN_EXCHANGE_RATES_APP_ID` when trying to sync currency exchange rates. If either is unset, fail with a meaningful error message rather than trying to fetch an invalid URL.

Try running the app with one or both of those settings unset. Wait for the `sync_currency_exchange_rates` task to trigger, and verify that the error is meaningful.


noisecapella:
Can you use `Fixes #xxxx` or `Closes #xxxx` if the linked issue should be closed when the PR is merged?

singingwolfboy:
@noisecapella Done.

financialaid/constants.py
```diff
@@ -52,11 +52,13 @@ class FinancialAidStatus:
         SKIPPED: "Skipped",
     }
 
-
-CURRENCY_EXCHANGE_RATE_API_REQUEST_URL = "{url}latest.json?app_id={app_id}".format(
-    url=settings.OPEN_EXCHANGE_RATES_URL,
-    app_id=quote_plus(settings.OPEN_EXCHANGE_RATES_APP_ID)
-)
+if settings.OPEN_EXCHANGE_RATES_URL and settings.OPEN_EXCHANGE_RATES_APP_ID:
```
gsidebo: looking at [this line](https://github.com/mitodl/micromasters/pull/2557/files#diff-705fc88e88815b293b37d0c2adf3cccaR16), it seems like `https://openexchangerates.org/` should always be our `OPEN_EXCHANGE_RATES_URL`, or at the very least the default. `OPEN_EXCHANGE_RATES_APP_ID` probably wouldn't be relevant for any other exchange rate API apart from `openexchangerates.org`. can you set that as the `OPEN_EXCHANGE_RATES_URL` default?

singingwolfboy: Frankly, I don't think this should be a variable at all -- it's not like we'll be able to just swap out which API we use. That's a larger change, though, that may not be in scope of this pull request. I've set a default for this variable in the `settings.py` file.

financialaid/tasks_test.py
```diff
@@ -3,14 +3,39 @@
...
+
+
+@override_settings(
+    OPEN_EXCHANGE_RATES_URL="https://openexchangerates.org/api/",
+    OPEN_EXCHANGE_RATES_APP_ID="fakeID123",
+)
+@patch(
+    'financialaid.tasks.CURRENCY_EXCHANGE_RATE_API_REQUEST_URL',
+    "https://openexchangerates.org/api/latest.json?app_id=fakeID123"
+)
```
gsidebo: is this patch necessary? seems redundant

singingwolfboy: Explicit is better than implicit. These tests are testing what happens when the constants are set, so I'm using `patch` to make sure that they're set the way we expect. I only had to add the env variables to the `travis-docker-compose.yml` file because other, unrelated tests were failing.

gsidebo: if we want to keep the patch on `financialaid.tasks.CURRENCY_EXCHANGE_RATE_API_REQUEST_URL`, then why are you using `override_settings` for `OPEN_EXCHANGE_RATES_URL` and `OPEN_EXCHANGE_RATES_APP_ID`?  we should have one or the other, right? sorry if i'm missing something here...

singingwolfboy: Because `CURRENCY_EXCHANGE_RATE_API_REQUEST_URL` isn't determined dynamically, it's determined once, when the Python application starts up. Changing the settings values won't automatically cause the constant to change.

Sowa rhr Mkw awnaw?

Issue #2470, opened 2017-02-01
Personal Pricing: validate income so it can only be an integer


pdpinch:
Currently the income field in the personal pricing calculator doesn't allow users to enter decimals. Some users haven't noticed this and ended up entering an income that is 100x what they meant to enter. For example, they type "12345.76" but the input field rejects the period and they end up submitting "123476"

The input field also rejects commas, which could lead to a similar outcome if the user is from a region where commas are used as a decimal separator. I haven't heard from any users who have made this error. I think we should continue to reject commas in the input field.

amir-qayyum-khan:
so what should we do here?
- alert user on pressing decimal?
- allow user  to add  decimal then convert it to integer on focus out or submitting to api

@pdpinch

pdpinch:
Sorry, we had discussed this at planning but I didn't update the issue. 

Anything other than an number should cause a validation error, "Please only use whole numbers."

PR #2559, opened 2017-02-14
Personal Pricing: Added validation on income so that it can only be an integer

amir-qayyum-khan:
- Fixes https://github.com/mitodl/micromasters/issues/2470#issuecomment-279479984

#### What's this PR do?
- Throws error when prices entered on 

(Required)
- Open dashboard and then `Cost Calculator`
 - Try to input string, alpha numeric
 - Try to enter decimal number
 - Try to enter special character 
 - Try to enter whole numbers

screen shot 2017-02-14 at 8 09 17 pm

screen shot 2017-02-14 at 8 12 35 pm

### DIFF
--- a/static/js/containers/FinancialAidCalculator.js
+++ b/static/js/containers/FinancialAidCalculator.js
@@ -23,7 +23,6 @@ import {
 import { createSimpleActionHelpers } from '../lib/redux';
 import { currencyOptions } from '../lib/currency';
 import { validateFinancialAid } from '../lib/validation/profile';
-import { sanitizeNumberString } from '../lib/validation/date';
 import type { AvailableProgram } from '../flow/enrollmentTypes';
 import type {
   FinancialAidState,
--- a/static/js/containers/FinancialAidCalculator.js
+++ b/static/js/containers/FinancialAidCalculator.js
@@ -57,7 +57,7 @@ const currencySelect = (update, current) => (
 
 const salaryUpdate = R.curry((update, current, e) => {
   let newEdit = R.clone(current);
-  newEdit.income = sanitizeNumberString(20, e.target.value);
+  newEdit.income = e.target.value
   update(newEdit);
 });
 
--- a/static/js/containers/FinancialAidCalculator.js
+++ b/static/js/containers/FinancialAidCalculator.js
@@ -56,7 +56,7 @@ const currencySelect = (update, current) => (
 
 const salaryUpdate = R.curry((update, current, e) => {
   let newEdit = R.clone(current);
-  newEdit.income = e.target.value
+  newEdit.income = e.target.value;
   update(newEdit);
 });
 
--- a/static/js/lib/validation/profile.js
+++ b/static/js/lib/validation/profile.js
@@ -296,18 +296,11 @@ const financialAidMessages: ErrorMessages = {
 
 export const validateFinancialAid = (edit: FinancialAidState): FinancialAidValidation => {
   let errors: FinancialAidValidation = findErrors(edit, R.keys(financialAidMessages), financialAidMessages);
-  const income = edit.income;
-  const not_whole_num = (
-    income && (
-      /[.]/.test(income) || /[a-z]/.test(income.toLowerCase())
-    )
-  );
-
   if (!edit.checkBox) {
     errors['checkBox'] = 'You must agree to these terms';
   }
 
-  if (not_whole_num) {
+  if (edit.income && !/^\d+$/.test(edit.income)) {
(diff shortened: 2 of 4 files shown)


static/js/containers/FinancialAidCalculator.js
```diff
@@ -56,7 +55,7 @@ const currencySelect = (update, current) => (
 
 const salaryUpdate = R.curry((update, current, e) => {
   let newEdit = R.clone(current);
-  newEdit.income = sanitizeNumberString(20, e.target.value);
+  newEdit.income = e.target.value
```
aliceriot: why not use the `filterPositiveInt` function we define in `util/util.js`? then we can keep the in-line behavior, preventing the user from typing a decimal in the first place.

you could do something like

```js
const sanitizeIncome = R.compose(filterPositiveInt, sanitizeNumberString(20));

newEdit.income = sanitizeIncome(e.target.value);
```

amir-qayyum-khan: this wont allow user to enter (.) i.e 2000.00 and see error message.
it changes 2000.00 to 200000. try ur self, copy 2000.00  and paste in field you will see it change into 200000

aliceriot: Oh, that makes sense. How about we add the error message but keep the `sanitizeNumberString` call?

amir-qayyum-khan: how? i think `newEdit.income` is what trigger validation/

aliceriot: just undo the change you made to line 59 in this file, keep it as-is on master. `sanitizeNumberString` basically just removes leading zeros.

amir-qayyum-khan: - It dont allow me to enter text and neither show the validation (please enter whole number)
- it dont allow me to add `.xy`
- it when you copy `555.50` into income it change it to `55550`

that the is reason @pdpinch opened this issue

aliceriot: ah ok now I see the issue. sorry for the lack of clarity! I wasn't following your explanations.

static/js/lib/validation/profile.js
```diff
@@ -299,5 +299,10 @@ export const validateFinancialAid = (edit: FinancialAidState): FinancialAidValid
   if (!edit.checkBox) {
     errors['checkBox'] = 'You must agree to these terms';
   }
+
+  if (edit.income && !/^\d+$/.test(edit.income)) {
```
aliceriot: I'm not sure this is exactly the regex we want - I think it might be more clear to test for non-digit characters, instead of `!`-ing a test for all digits.

so maybe something like `/\D+/` ?

aliceriot: `+` isn't even really necessary, I think `/\D/` would do the job.

amir-qayyum-khan: @aliceriot `!/^\d+$/.test(edit.income)` this give us all string which are not number, including `dot(.)` special characters.

Otherwise we have to cater those in our regex

aliceriot: will `/\D/` not do that?

aliceriot: `\D` matches any non-digit character

amir-qayyum-khan: nop,

aliceriot: I don't think you don't need the quantifiers or the anchors (`^`, `+`, `$`) either because all we care about is whether there is a non-digit character anywhere in the string

aliceriot: `!/^\d+$/.test(edit.income)` and `/\D/.test(edit.income)` are, as far as I can tell, totally equivalent - I'm just advocating for the latter because the regular expression is simpler.

amir-qayyum-khan: you r right fixed

static/js/lib/validation/profile.js
```diff
@@ -299,5 +299,10 @@ export const validateFinancialAid = (edit: FinancialAidState): FinancialAidValid
   if (!edit.checkBox) {
     errors['checkBox'] = 'You must agree to these terms';
   }
+
+  if (edit.income && /\D+/.test(edit.income)) {
+    errors['income'] = 'Please only use whole numbers.';
+  }
+
```
aliceriot: could you add a quick test of this behavior in the `validation/profile_test.js` file?

amir-qayyum-khan: Done @aliceriot

Issue #2565, opened 2017-02-14
settings.MIDDLEWARE_CLASSES deprecated in Django 1.10


rhysyngsun:
This setting was [deprecated](https://docs.djangoproject.com/en/1.10/ref/settings/#middleware-classes) in favor of `settings.MIDDLEWARE`. I ran into this implementing #2558 but didn't want to bite off the amount of testing this warrants.

Any usages of `django.utils.deprecation.MiddlewareMixin` in our code should get updated to the new middleware implementation.

rhysyngsun:
As a heads up on this for whomever takes this, the `social.apps.django_app.middleware.SocialAuthExceptionMiddleware` and `nplusone.ext.django.NPlusOneMiddleware` middlewares appear to not support Django post-1.9 style middleware.

PR #2568, opened 2017-02-14
Added runtime feature flags via cookie (#2558)

### LINKED ISSUE #2558: Exams: beta feature toggle
We need to test this in production and need a mechanism to toggle this on as a beta feature so that we can "soft-launch" it with internal users. Ideally, this is implemented generically so that we can reuse it for other feature rollouts in the future.

##### General mechanism:

 - [ ] Implement a custom [Django middleware](https://docs.djangoproject.com/en/1.10/topics/http/middleware/) that checks for the presence of certain feature flags and sets a cookie that describes which feature flags are set.
 - [ ] Implement a custom Django middleware that parses the cookie so we can use it in views and templates.
 - [ ] Apply some minimal effort to obfuscate the above (as long as it don't make things too complicated)

#### Exams-specific
- [ ] Implement a feature flag using the above mechanism for exams

rhysyngsun:
Fixes #2558 

#### What's this PR do?
Adds a general mechanism for enabling feature flags per user and implements one for exams ui.

It does so by setting a signed cookie to the bitmask of the feature(s) you've turned on. This is implemented as middleware and I've made an effort to keep the code as minimal and unobtrusive as possible.

**Note:** the middleware uses `django.utils.deprecation.MiddlewareMixin`. I briefly explored upgrading our settings to avoid usage of this, but ran into difficulties and it's probably better dealt with separately anyway. I logged #2565 to address it later and noted the issues I ran into.

- In `.env` set `FEATURE_FLAG_QS_PREFIX=MM`
- Create an `ExamProfile` record for your user.
- Login and go to http://localhost:8079/dashboard/?MM_FEATURE_EXAMS=1
- Ensure you can get redirected to /dashboard/ (no querystring) and that you see the exam card
- Add `?MM_FEATURE_CLEAR` to the url to clear the cookie and verify the card goes away.

ui/middleware.py

semaphore

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -557,6 +557,17 @@ FEATURES = {
     in get_all_config_keys() if key.startswith(MM_FEATURES_PREFIX)
 }
 
+MIDDLEWARE_FEATURE_FLAG_QS_PREFIX = get_var("MIDDLEWARE_FEATURE_FLAG_QS_PREFIX", None)
+MIDDLEWARE_FEATURE_FLAG_COOKIE_NAME = get_var('MIDDLEWARE_FEATURE_FLAG_COOKIE_NAME', 'MM_FEATURE_FLAGS')
+MIDDLEWARE_FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS = get_var('MIDDLEWARE_FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS', 60 * 60)
+
+
+if MIDDLEWARE_FEATURE_FLAG_QS_PREFIX:
+    MIDDLEWARE_CLASSES = MIDDLEWARE_CLASSES + (
+        'ui.middleware.QueryStringFeatureFlagMiddleware',
+        'ui.middleware.CookieFeatureFlagMiddleware',
+    )
+
 
 # django debug toolbar only in debug mode
 if DEBUG:
--- a/ui/middleware.py
+++ b/ui/middleware.py
@@ -1,0 +1,121 @@
+"""Common MicroMasters middleware"""
+from django.conf import settings
+from django import shortcuts
+from django.utils.deprecation import MiddlewareMixin
+
+from ui.utils import FeatureFlag
+
+
+class QueryStringFeatureFlagMiddleware(MiddlewareMixin):
+    """
+    Extracts feature flags from the query string
+    """
+
+    @classmethod
+    def get_flag_key(cls, suffix):
+        """
+        Determines the full key for a given feature flag suffix
+
+        Args:
+            suffix (str): suffix to append to the key prefix
+
+        Returns:
+            str: the full key value
+        """
+        return '{prefix}_FEATURE_{suffix}'.format(
+            prefix=settings.MIDDLEWARE_FEATURE_FLAG_QS_PREFIX,
+            suffix=suffix,
+        )
+
+    @classmethod
+    def encode_feature_flags(cls, data):
+        """
+        Encodes the set of feature flags from the request by creating a bit mask
+
+        Args:
+            data (dict): request query dict
+
+        Returns:
+            str: value encoded as a str
+        """
+        mask = 0
+        if data is None:
+            return str(mask)
+
+        for member in FeatureFlag:
+            if cls.get_flag_key(member.name) in data:
+                mask = mask | member.value
+        return str(mask)
+
+    def process_request(self, request):
+        """
+        Processes an individual request for the feature flag query parameters
+
+        Args:
+            request (django.http.request.Request): the request to inspect
+        """
+        prefix = self.get_flag_key('')
+        print(prefix)
+        if request.GET and any(key.startswith(prefix) for key in request.GET.keys()):
+            response = shortcuts.redirect(request.path)
+            if self.get_flag_key('CLEAR') in request.GET:
+                response.delete_cookie(settings.MIDDLEWARE_FEATURE_FLAG_COOKIE_NAME)
+            else:
+                response.set_signed_cookie(
+                    settings.MIDDLEWARE_FEATURE_FLAG_COOKIE_NAME,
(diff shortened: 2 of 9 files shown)


rhysyngsun:
@giocalitri @gsidebo @noisecapella I'd appreciate any weigh-in on the general approach here.

micromasters/settings.py
```diff
@@ -557,6 +559,8 @@ def get_all_config_keys():
     in get_all_config_keys() if key.startswith(MM_FEATURES_PREFIX)
 }
 
+FEATURE_FLAG_QS_PREFIX = get_var("FEATURE_FLAG_QS_PREFIX", None)
```
giocalitri: why don't you reuse the `FEATURES ` dictionary and add to that?

ui/middleware.py
```diff
@@ -0,0 +1,130 @@
+"""Common MicroMasters middleware"""
+from django.conf import settings
+from django import shortcuts
+from django.utils.deprecation import MiddlewareMixin
+
+from ui.utils import FeatureFlag
+
+FEATURE_FLAG_COOKIE_NAME = 'MM_FEATURE_FLAGS'
+FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS = 60 * 60
```
giocalitri: Do we need these to come from the settings to easily invalidate them?

ui/utils.py
```diff
@@ -0,0 +1,12 @@
...
+    Enum,
+    unique,
+)
+
+
+# Python 3.6: this should be upgraded to enum.Flag
+@unique
+class FeatureFlag(Enum):
+    """FeatureFlag enum"""
+    EXAMS = 1
```
giocalitri: what is this?

ui/middleware.py
rhysyngsun: Sure I'll add them.

micromasters/settings.py
rhysyngsun: I'll rename this with a different prefix, it was unintentional that this pollute that namespace.

ui/utils.py
rhysyngsun: Values should be powers of 2 so they can be used in the bitmask, I'll add that explanation to the docstring.

rhysyngsun:
@giocalitri addressed your feedback plus the conditional middleware activation you mentioned

micromasters/settings.py
```diff
@@ -557,6 +557,17 @@ def get_all_config_keys():
     in get_all_config_keys() if key.startswith(MM_FEATURES_PREFIX)
 }
 
+MIDDLEWARE_FEATURE_FLAG_QS_PREFIX = get_var("MIDDLEWARE_FEATURE_FLAG_QS_PREFIX", None)
```
giocalitri: nit: set it to `False` so we can explicitly set it to `True` in the env

micromasters/settings.py
```diff
@@ -557,6 +557,17 @@ def get_all_config_keys():
...
+MIDDLEWARE_FEATURE_FLAG_COOKIE_NAME = get_var('MIDDLEWARE_FEATURE_FLAG_COOKIE_NAME', 'MM_FEATURE_FLAGS')
+MIDDLEWARE_FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS = get_var('MIDDLEWARE_FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS', 60 * 60)
+
+
+if MIDDLEWARE_FEATURE_FLAG_QS_PREFIX:
```
giocalitri: check explicitly if it is `True`

micromasters/settings.py
```diff
@@ -557,6 +557,17 @@ def get_all_config_keys():
...
+    MIDDLEWARE_CLASSES = (
+        'ui.middleware.QueryStringFeatureFlagMiddleware',
+        'ui.middleware.CookieFeatureFlagMiddleware',
+    ) + MIDDLEWARE_CLASSES
```
giocalitri: if there is not a real need to be the first classes in the list, please put them as last.

giocalitri:
can you please update the "how to test" section?

micromasters/settings.py
giocalitri: sorry I misread the variable

rhysyngsun: This isn't a boolean variable, please see my comment below.

micromasters/settings.py
rhysyngsun: The possible values are either `None` or `str` (please see usage [here](https://github.com/mitodl/micromasters/pull/2568/files#diff-910e7ef6baffbc57d4ad11ad102ceb5aR26)). I coded it this way so that the feature is only enabled if the prefix has a non-empty value, that gives us two behaviors I think are desirable a) the prefix needs to be set to something other than `None` or `""`, this means the feature will never be enabled with no prefix b) it reduces configuration complexity by eliminating the need to set both the prefex environment variable **and** a boolean variable to enable the middleware. Disabling the middleware is then as simple as either setting the env var to `""` or unsetting it.

micromasters/settings.py
rhysyngsun: Done.

micromasters/settings.py
```diff
@@ -22,7 +22,7 @@
 from django.core.exceptions import ImproperlyConfigured
 
 
-VERSION = "0.47.1"
+VERSION = "0.47.0"
```
giocalitri: Not sure why this: you may want to rebase master.

micromasters/settings.py
giocalitri: I am sorry: like I wrote above, I misread the variable name.

giocalitri:
I am still running manual tests

micromasters/settings.py
rhysyngsun: Stale file in my editor, removed.

giocalitri:
it works as expected 👍

PR #2519, opened 2017-02-07
New course enrollment UX

singingwolfboy:
#### What are the relevant tickets?
Fixes #2467.

#### What's this PR do?
Changes the user experience for enrolling in courses through MicroMasters. Adds a message above the course listing in a program, explaining what the user's price is for these courses. Changes the buttons to say "Enroll Now" instead of displaying the course price. When the user clicks one of these "Enroll Now" buttons, the user sees a modal with two options: pay for the course, or audit it for free.

#### How should this be manually tested?
1. While enrolled in a program that has some courses that you are not yet enrolled in, visit the dashboard. Verify that the buttons to the right of the courses say "Enroll Now", and do not have any course price in the button text. Verify that there is no "Enroll and pay later" link. Click on the button, and verify that a dialog pops up with two choices: "Pay Now" and "Audit for Free & Pay Later". When clicked, the "Pay Now" button should redirect the user to the order summary page. When clicked, the "Audit" button should cause the user to be enrolled in the course on edX on the audit track.
2. Courses where you are already enrolled as an auditing user should instead have a button that says "Pay Now". Clicking on this button should not open a dialog, but should immediately redirect the user to the order summary page.
3. For a course without financial aid available, verify that the cost of the course is described in the blue box at the top of the course list. When the course enrollment dialog is opened, verify that the "Pay Now" button also includes the cost of the course.
4. Create a coupon for this course, and apply the coupon to your user account. Verify that the reduced price is correctly reflected in the text of the blue box above the course list, and in the "Pay Now" button.
5. For a course _with_ financial aid available, verify that the "Pay Now" button in the enrollment dialog is disabled before the user has calculated their financial aid. The text of the blue box should also direct the user to calculate their financial aid. Verify that the user can still audit the course.
6. After financial aid has been calculated, verify that the blue box and the "Pay Now" button should display the cost with financial aid applied. The "Pay Now" button should no longer be disabled.
7. For a course with both financial aid and a coupon, the price in the blue box and the "Pay Now" button should reflect both discounts.
8. Create a 100% off coupon for a program. Verify that clicking on the "Enroll Now" button on the dashboard for a course in that program does _not_ cause the enrollment dialog to appear, but instead redirects the user to the order summary page immediately.

#### Screenshots (if appropriate)
course-price-message
enroll dialog disabled
enroll dialog enabled

audit

PR #2571, opened 2017-02-15
dashboard return promises

### LINKED ISSUE #2570: Dashboard functions should return promises where possible
The dashboard has a lot of functions that call `dispatch`, which returns a Promise. These functions should return these promises, rather than returning `void`, so that callers can chain these promises together. This is especially important for handling API errors effectively.

singingwolfboy:
Fixes #2570.

#### What's this PR do?
Updates functions in `DashboardPage.js` to return Promise objects where possible. If a function generates multiple promises, a new Promise is returned that waits for all the others to complete (using `Promise.all()`).

No change in functionality -- I'm not sure this needs a manual test at all.

`static/js/containers/DashboardPage.js`


aliceriot:
Is this just a general refactor or are you trying to fix a specific error? Generally we return a promise from an API action dispatcher function if we need to chain off of it, and otherwise we haven't worried about it.

aliceriot:
Also not all actions return a promise. We use a middleware with redux (called [redux-thunk](https://github.com/gaearon/redux-thunk)) which lets us use async actions that return a promise. But a normal action (i.e. one we create with `createAction` doesn't do that.

aliceriot:
(some of the actions you're wrapping with `Promise.all` do not return a promise, and are simple synchronous actions)

singingwolfboy:
@aliceriot It's just a general refactor. I'm trying to promote good habits: in this case, returning promises.

For a more specific example of why I want to do this, I was working on #2519, and I wanted to add a specific error handler. If the user clicks on the button to audit a course, and the fetch request gets an error response, the app should notify the user of the failure instead of failing silently. However, my `.catch()` handler wasn't being called, and it took me a decent amount of debugging to figure out that the dashboard function that wraps `dispatch()` wasn't actually _returning_ the Promise from `dispatch()`. As a result, the exception wasn't chaining properly, and the app thought that there was no problem with the fetch.

We should get into better habits, so that it's easier to write correct code and harder to write incorrect code.

singingwolfboy:
@aliceriot if you can point out the synchronous actions, or show me how to determine which are synchronous and which are asynchronous, I'd be happy to clean up this PR to only return promises when there's a real promise to return.

aliceriot:
sure, I'll read through the code, but I'm not convinced we should make this change at all, in cases where we do want to chain off the promise we already return them.

singingwolfboy:
@aliceriot I'm happy to discuss this change, and potentially to be talked out of it. My goal is to avoid confusion, and when I was trying to write that error handler, I was very confused when it wasn't called.

aliceriot:
I think in general the action dispatcher functions we use to wrap the API (like `sendCourseTeamMail` from `actions/email`) are where you would want to do your chaining.

aliceriot:
Sorry, to clarify: I think all of our API-wrapping action dispatcher functions should return `Promise.resolve()` on success and `Promise.reject()` on error, but the same doesn't hold for wrapper functions at the container level.

aliceriot:
Also, as far as telling synchronous action creators from async ones, simple action creators return an object literal, of the form:

```js
const myActionCreator = someValue => ({
  type: SOME_TYPE, payload: someValue
});
```

and async ones return a function, like this:

```js
const myAsyncActionCreator = someValue => {
  return dispatch => {
    dispatch(startAPIThing());
    return api.getTheThing().then(
      () => {
        dispatch(apiSuccessAction());
        return Promise.resolve(); 
      },
      () => {
        dispatch(apiFailAction());
        return Promise.reject();
      }
    );
  }
}
```

aliceriot:
I forgot a thing! In JavaScript if you return a rejected promise from a function (or otherwise have a `Promise.reject()` without a function to call on rejection) it raises an error. So we don't really want to return `Promises` from things all over the place, but only in cases where we have error handling set up.

aliceriot:
^^ the above comment makes me think we don't want to make these kinds of changes at all. I do think that the difficulty that you (@db) pointed out w/r/t figuring out how to chain some sort of on-error behavior off of an API action indicates we may have a need to write some documentation around the patterns we use with `fetch` and `redux`.

@noisecapella do you have any thoughts on this?

noisecapella:
I agree with @aliceriot here about the rejected promises. I think we should update the functions individually at the point where there's a need for the promise

singingwolfboy:
Fair enough. Closing this pull request.

PR #2574, opened 2017-02-15
Added field to FinalGrade to save if user paid on edx

### LINKED ISSUE #2507: Financial Aid: freeze payment status for users with in not financial aid programs
We should freeze the payment status for the users in SCM together with the final grade

giocalitri:
closes #2507 

#### What's this PR do?
Adds the feature to freeze the payment status if done on edX. This is important because we will not need to rely on cached enrollments to determine if a user payed on edX.

set `FEATURE_FINAL_GRADE_ALGORITHM=v1` in your `.env`

For the dashboard: 

in a not Financial Aid program, take a course (**that run in the past**) without an enrollment and create a final grade (using the factory) paying attention to set the `passed=True` and `course_run_payed_on_edx=True`. The course should show up in the dashboard as passed. 
Modifying the `course_run_payed_on_edx=False` should turn the course to "missed deadline". 

For the grades api:

take a course in a Financial aid program, create a certificate with `certificate_type="honor"` and from a shell run `grades.api._compute_grade_for_fa`: this should return an object saying that the course has been payed on edX.
if you modify the `certificate_type="audit"` and run the same function, the returned object should say that the course has not been payed on edX.

`dashboard/utils.py`
`grades/api.py`

This should handle the case of old courses on edX where used that got `honor` enrollment and certificate should be considered as payed.


grades/models.py
```diff
@@ -39,6 +39,7 @@ class FinalGrade(TimestampedModel, AuditableModel):
         default=FinalGradeStatus.PENDING,
         max_length=30,
     )
+    course_run_payed_on_edx = models.BooleanField(default=False)
```
rhysyngsun: Spelling nit: "paid" is typically used over "payed" for past-tense.

grades/api_test.py
```diff
@@ -179,8 +211,9 @@ def test_compute_grade_for_non_fa_certs_status(self, certificate_type, status, e
 
         run_data = CachedEdxUserData(self.user).get_run_data(course_key)
         grade = api._compute_grade_for_non_fa(run_data)
-        assert grade.passed is expected_result
+        assert grade.passed is grade_result
         assert grade.grade == self.certificates.get(course_key).data.get('grade')
+        assert grade.payed_on_edx == payed_edx_result
```
rhysyngsun: Shouldn't this be `is` instead of `==`?

grades/api.py
```diff
@@ -36,10 +36,14 @@ def _compute_grade_for_fa(user_edx_run_data):
     if user_edx_run_data.certificate is not None:
         run_passed = user_edx_run_data.certificate.status == 'downloadable'
         grade = user_edx_run_data.certificate.grade
+        # the following line should be updated when
+        # we add support for honor enrollments and certificates in the edx-api-client
+        payed_on_edx = user_edx_run_data.certificate.certificate_type in ['honor', 'verified']
```
rhysyngsun: `['honor', 'verified']` should probably be a constant because it's used 3 times here and I would expect we want those to all be in sync?

grades/models.py
rhysyngsun: I also assume that we either have no or very few records in this table in production since this field isn't nullable (and hence PG will update all existing records)?

grades/api.py
giocalitri: I did not do it on purpose because there is another story to move this in the `edx-api-client` library as soon as we have some more info from people at edX.

grades/models.py
giocalitri: >Spelling nit: "paid" is typically used over "payed" for past-tense.
will fix

giocalitri: >I also assume that we either have no or very few records in this table in production since this field isn't nullable (and hence PG will update all existing records)?

Yes: most likely we will merge and deploy before we perform actual freezes.
Currently there are some in the DB in production, but I plan to remove them.

grades/api_test.py
giocalitri: will change

giocalitri:
@rhysyngsun What you found during your test was an actual bug that required some new logic.
I think I fixed everything but now if an user has a not paid final grade, she sees "Missed Deadline" on the dashboard and not "Failed"

grades/api.py
rhysyngsun: ok, makes sense

rhysyngsun:

PR #2575, opened 2017-02-15
Lower logging of unexceptional exception to debug

### LINKED ISSUE #2573: Exams: Signal on FinalGrade logs exceptions when a course isn't configured for exams
See https://github.com/mitodl/micromasters/blob/0742322e2a4f4274bfaddf43723963d3f692665b/exams/signals.py#L40

This exception was added to facilitate error reporting elsewhere, but in this signal it's just noise because the exceptions aren't truly exceptional.

rhysyngsun:
Fixes #2573 

#### What's this PR do?
Lowers a single `log.exception` to `log.debug`.

This is just a logging statement and the tests exercise that code, so if they pass we're good.

### DIFF
--- a/selenium_tests/base.py
+++ b/selenium_tests/base.py
@@ -259,6 +259,10 @@ FROM pg_stat_activity WHERE pid <> pg_backend_pid()""")
             if "'webkitURL' is deprecated. Please use 'URL' instead" in message:
                 continue
 
+            # warnings (e.g. deprecations) should not fail the tests
+            if entry['level'] in ["WARNING"]:
+                continue
+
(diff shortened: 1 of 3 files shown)


giocalitri:
👍 after tests pass

selenium_tests/base.py
```diff
@@ -259,6 +259,10 @@ def assert_console_logs(self):
             if "'webkitURL' is deprecated. Please use 'URL' instead" in message:
                 continue
 
+            # warngings (e.g. deprecations) should not fail the tests
```
noisecapella: `warning`

PR #2629, opened 2017-02-15
Validate prices and FA discounts management command

### LINKED ISSUE #2497: Validation for price + financial aid
We should have a management command to assert that all prices are present and equal to each other for a financial aid program, and that no financial aid discount causes the price to go lower than $0. We should also assert that no price or financial aid discount is negative

annagav:
Fix #2497

#### What's this PR do?
Create a  management command to validate `course_price`s and `discount_amount` for all live programs.

Run the command: ` ./manage.py validate_price`.
It should print errors if course prices for a live program are not equal to each other, or do not exist. Also, if a TierProgram has a discount that would bring the final price to a negative value.

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -41,10 +41,12 @@ from ecommerce.models import (
     Order,
     RedeemedCoupon,
 )
+from ecommerce.models import CoursePrice
 from financialaid.api import get_formatted_course_price
 from financialaid.models import (
     FinancialAid,
     FinancialAidStatus,
+    TierProgram
 )
 from profiles.api import get_social_username
 
@@ -480,3 +482,36 @@ def calculate_run_price(course_run, user):
     coupon = coupons[0]
     price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
     return price, coupon
+
+
+def validate_prices():
+    """
+    Validate prices and financial aid discounts
+    Returns:
+        list: List of validation errors
+    """
+    errors = []
+    programs = Program.objects.filter(live=True)
+    for program in programs:
+        runs = CourseRun.objects.filter(course__program=program)
+        course_prices = [CoursePrice.objects.filter(course_run=run, is_valid=True).first() for run in runs]
+        prices = [course_price.price if course_price else None for course_price in course_prices]
+
+        if not (len(prices) and all(prices[0] == price for price in prices) and prices[0] is not None):
+            errors.append('Invalid course prices for program {0}'.format(program.title))
+
+        elif program.financial_aid_availability:
+            tier = TierProgram.objects.filter(program=program, current=True).order_by("-discount_amount").first()
+            if tier:
+                if tier.discount_amount > prices[0]:
+                    errors.append('Discount is higher than course price for program {0}'.format(program.title))
+                if not TierProgram.objects.filter(discount_amount=0, program=program, current=True).exists():
+                    errors.append('Could not find 0 discount TierProgram for program {0}'.format(program.title))
+                if not TierProgram.objects.filter(income_threshold=0, program=program, current=True).exists():
+                    errors.append(
+                        'Could not find 0 income_threshold TierProgram for program {0}'.format(program.title)
+                    )
+            else:
+                errors.append('Could not find current TierProgram for program {0}'.format(program.title))
+
+    return errors
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -22,7 +22,7 @@ from rest_framework.exceptions import ValidationError
 from edx_api.enrollments import Enrollment
 
 from backends.pipeline_api import EdxOrgOAuth2
-from courses.factories import CourseRunFactory
+from courses.factories import CourseRunFactory, ProgramFactory
 from dashboard.models import (
     CachedCertificate,
     CachedEnrollment,
@@ -42,6 +42,7 @@ from ecommerce.api import (
     ISO_8601_FORMAT,
     make_reference_id,
     pick_coupons,
+    validate_prices,
 )
 from ecommerce.exceptions import (
     EcommerceEdxApiException,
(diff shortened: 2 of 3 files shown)


noisecapella:
Can you rename the management command to `validate_db`? We might extend it in the future to do more than validate prices

ecommerce/api.py
```diff
@@ -480,3 +482,28 @@ def calculate_run_price(course_run, user):
...
+        if not all(prices[0] == price for price in prices):
+            errors.append('Invalid course prices for program {0}'.format(program.title))
+
+        if program.financial_aid_availability:
+            tier = TierProgram.objects.filter(program=program, current=True).first()
+            if tier:
+                if tier.discount_amount > program.get_course_price():
+                    errors.append('Discount is higher than course price for program {0}'.format(program.title))
+            else:
+                errors.append('Could not find TierProgram for program {0}'.format(program.title))
```
noisecapella: It looks like this line is uncovered, can you add a test for it?

ecommerce/api.py
```diff
@@ -480,3 +482,28 @@ def calculate_run_price(course_run, user):
     coupon = coupons[0]
     price = calculate_coupon_price(coupon, price, course_run.edx_course_key)
     return price, coupon
+
+
+def validate_prices():
+    """
+    Validate prices and financial aid discounts
+    Returns:
+        (list): List of validation errors
```
noisecapella: You don't need parentheses around the return value type: `list: List of validation errors`

ecommerce/api.py
```diff
@@ -480,3 +482,28 @@ def calculate_run_price(course_run, user):
...
+    errors = []
+    programs = Program.objects.filter(live=True)
+    for program in programs:
+        runs = CourseRun.objects.filter(course__program=program)
+        course_prices = [CoursePrice.objects.filter(course_run=run, is_valid=True).first() for run in runs]
+        prices = [course_price.price if course_price else None for course_price in course_prices]
+        if not all(prices[0] == price for price in prices):
+            errors.append('Invalid course prices for program {0}'.format(program.title))
+
+        if program.financial_aid_availability:
```
noisecapella: I didn't say this in the issue but can you also verify that there is one `discount_amount` which is `0` and one `income_threshold` which is `0`?

ecommerce/api.py
```diff
@@ -480,3 +482,28 @@ def calculate_run_price(course_run, user):
...
+    Validate prices and financial aid discounts
+    Returns:
+        (list): List of validation errors
+    """
+    errors = []
+    programs = Program.objects.filter(live=True)
+    for program in programs:
+        runs = CourseRun.objects.filter(course__program=program)
+        course_prices = [CoursePrice.objects.filter(course_run=run, is_valid=True).first() for run in runs]
+        prices = [course_price.price if course_price else None for course_price in course_prices]
```
noisecapella: Can you also check for the case where there is no course price on any course run within the program?

ecommerce/api_test.py
```diff
@@ -861,3 +863,47 @@ def test_capped_coupon_price_above_full(self):
...
+
+    def test_courses_with_no_price(self):
+        """course does not have price"""
+        CourseRunFactory.create(
+            course__program=self.program,
+        )
+        assert validate_prices()[0] == 'Invalid course prices for program {0}'.format(self.program.title)
+
+    def test_fa_discount_too_large(self):
+        """discount greater then coures price"""
```
noisecapella: Should be `course`

annagav:
@noisecapella, I addressed all the comments.

ecommerce/api.py
```diff
@@ -480,3 +482,36 @@ def calculate_run_price(course_run, user):
...
+        list: List of validation errors
+    """
+    errors = []
+    programs = Program.objects.filter(live=True)
+    for program in programs:
+        runs = CourseRun.objects.filter(course__program=program)
+        course_prices = [CoursePrice.objects.filter(course_run=run, is_valid=True).first() for run in runs]
+        prices = [course_price.price if course_price else None for course_price in course_prices]
+
+        if not (all(prices[0] == price for price in prices) and prices[0]):
```
noisecapella: What is `prices[0]` checking for here? If `prices` is an empty list this will cause a `IndexError`, and we don't want to do anything different if the price is `0` or if it's greater than `0`

ecommerce/api.py
```diff
@@ -480,3 +482,36 @@ def calculate_run_price(course_run, user):
...
+            errors.append('Invalid course prices for program {0}'.format(program.title))
+
+        elif program.financial_aid_availability:
+            tier = TierProgram.objects.filter(program=program, current=True).order_by("-discount_amount").first()
+            if tier:
+                if tier.discount_amount > prices[0]:
+                    errors.append('Discount is higher than course price for program {0}'.format(program.title))
+                if not TierProgram.objects.filter(discount_amount=0, program=program, current=True).exists():
```
noisecapella: Can you move these cases out of the `if` block, since they don't depend on `tier`?

ecommerce/api.py
annagav: `prices` should not be empty. If there is no course price there should be None in the list for that course_run.

ecommerce/api.py
annagav: if there is not `tier` does it make sense to check if a `tier` with 0 discount exists?

noisecapella: Probably not, feel free to leave it as is

ecommerce/api.py
annagav: @noisecapella, I addressed this in the last commit.

ecommerce/api_test.py
```diff
@@ -861,3 +863,86 @@ def test_capped_coupon_price_above_full(self):
...
+            course__program=self.program,
+        )
+        CoursePrice.objects.filter(
+            is_valid=True,
+            course_run__course__program=self.program
+        ).delete()
+        assert validate_prices()[0] == 'Invalid course prices for program {0}'.format(self.program.title)
+
+    def test_fa_discount_too_large(self):
+        """discount greater then coures price"""
```
noisecapella: Typo: `course`

PR #2634, opened 2017-02-15
Make separate selenium container for tests

### LINKED ISSUE #2628: Tests: Python tests on travis should not pull selenium container
The current set-up adds unnecessary minutes to the tests.

noisecapella:
Fixes #2628

#### What's this PR do?
Creates a separate selenium container so that the grid container is not pulled with everything else when it's not needed.

Make sure you can run the tests locally and that you can see the website as before

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ matrix:
     - install:
       - env | grep TRAVIS > .env
       - env | grep CI >> .env
-      - docker-compose -f travis-docker-compose.yml build
       - docker-compose -f travis-docker-compose.yml run celery echo
       script: (docker-compose -f travis-docker-compose.yml up celery &); sleep 5; docker-compose -f travis-docker-compose.yml run web tox
       services:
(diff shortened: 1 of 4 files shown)


noisecapella:
cc @giocalitri

.travis.yml
```diff
@@ -7,7 +7,6 @@ matrix:
     - install:
       - env | grep TRAVIS > .env
       - env | grep CI >> .env
-      - docker-compose -f travis-docker-compose.yml build
```
giocalitri: why this (and the similar command later) is not necessary?

noisecapella: The build step is done automatically when `docker-compose run container` is done, if that container does not already exist. The way it's defined here it will build all containers which includes grid, but we want build only the containers that are used in the tests

travis-docker-compose.yml
```diff
@@ -31,31 +31,25 @@ services:
       BROKER_URL: redis://redis:6379/4
       CELERY_RESULT_BACKEND: redis://redis:6379/4
     env_file: .env
+
+  web:
+    image: micromasters_python
```
aliceriot: where does the `micromasters_python` image come from? is that tag automatically generated based on having the `python` container declared in the `docker-compose` file?

noisecapella: Yes, the image is specified here so we don't go through the docker build multiple times

aliceriot: nvm disregard this 😆

giocalitri:
👍

Issue #2560, opened 2017-02-14
Exams: environment variable to suppress payment requirement 


rhysyngsun:
This is a replacement for #2279. In order to facilitate the pilot period for exams we decided to suppress the payment requirement during that period instead of implementing this via coupons.

This feature flag should default to a value so that it does not require payment.

This code that performs this check is located here: https://github.com/mitodl/micromasters/blob/master/exams/utils.py#L91-L96

PR #2640, opened 2017-02-16
Exams: Added environment variable to suppress payment requirement

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2560

#### What's this PR do?
Adds feature flag `FEATURE_SUPPRESS_PAYMENT_FOR_EXAM=True` to suppress the payment check for exam authorization

-Authorize for course after passing it, no need to pay.

@pdpinch

### DIFF
--- a/exams/utils.py
+++ b/exams/utils.py
@@ -68,22 +68,24 @@ def is_eligible_for_exam(mmtrack, course_run):
     Returns:
         bool: whether user is eligible or not
     """
-    return course_run.has_exam and has_paid(mmtrack, course_run.edx_course_key)
+    return course_run.has_exam and _has_paid_for_exam(mmtrack, course_run.edx_course_key)
 
 
-def has_paid(mmtrack, edx_course_key):
+def _has_paid_for_exam(mmtrack, edx_course_key):
     """
     Returns True if user has paid for the course or payment check is suppressed using feature flag
     FEATURE_SUPPRESS_PAYMENT_FOR_EXAM.
+
     Args:
         mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
         edx_course_key (str): An edX course key (CourseRun.edx_course_key)
 
-    bool: whether user has paid or not.
+    Returns:
+        bool: whether user has paid or not.
     """
     # if FEATURE_SUPPRESS_PAYMENT_FOR_EXAM=True then bypass payment check
-    suppress_payment_for_exam = settings.FEATURES.get("SUPPRESS_PAYMENT_FOR_EXAM", True)
-    return mmtrack.has_paid(edx_course_key) or suppress_payment_for_exam
+    suppress_payment_for_exam = settings.FEATURES.get("SUPPRESS_PAYMENT_FOR_EXAM", False)
+    return suppress_payment_for_exam or mmtrack.has_paid(edx_course_key)
 
 
 def authorize_for_exam(mmtrack, course_run):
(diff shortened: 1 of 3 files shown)


exams/utils.py
```diff
@@ -66,9 +66,24 @@ def is_eligible_for_exam(mmtrack, course_run):
...
 
     Returns:
-        bool: whether use is eligible or not
+        bool: whether user is eligible or not
     """
-    return course_run.has_exam and mmtrack.has_paid(course_run.edx_course_key)
+    return course_run.has_exam and has_paid(mmtrack, course_run.edx_course_key)
+
+
+def has_paid(mmtrack, edx_course_key):
```
rhysyngsun: Could you prefix this with `_` so it's clear that this shouldn't be used elsewhere? Also, it should probably be named `_has_paid_for_exam` because this is specific to exams.

exams/utils.py
```diff
@@ -66,9 +66,24 @@ def is_eligible_for_exam(mmtrack, course_run):
...
+    """
+    Returns True if user has paid for the course or payment check is suppressed using feature flag
+    FEATURE_SUPPRESS_PAYMENT_FOR_EXAM.
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        edx_course_key (str): An edX course key (CourseRun.edx_course_key)
+
+    bool: whether user has paid or not.
```
rhysyngsun: Please add `Returns` section here.

exams/utils_test.py
```diff
@@ -146,6 +150,22 @@ def test_exam_authorization_when_not_paid(self):
             course=self.course_run.course
         ).exists() is False
 
+    @override_settings(FEATURES={"SUPPRESS_PAYMENT_FOR_EXAM": True})
+    def test_exam_authorization_when_payment_check_suppress(self):
+        """
+        test exam_authorization when user has passed and payment check is suppress.
```
rhysyngsun: Spelling: "suppressed"

exams/utils.py
```diff
@@ -66,9 +66,24 @@ def is_eligible_for_exam(mmtrack, course_run):
...
+    """
+    # if FEATURE_SUPPRESS_PAYMENT_FOR_EXAM=True then bypass payment check
+    suppress_payment_for_exam = settings.FEATURES.get("SUPPRESS_PAYMENT_FOR_EXAM", True)
+    return mmtrack.has_paid(edx_course_key) or suppress_payment_for_exam
```
rhysyngsun: Could you reverse the order of these so we don't check `mmtrack.has_paid()` if `suppress_payment_for_exam` is already True.

exams/utils.py
```diff
@@ -66,9 +66,24 @@ def is_eligible_for_exam(mmtrack, course_run):
...
+    Returns True if user has paid for the course or payment check is suppressed using feature flag
+    FEATURE_SUPPRESS_PAYMENT_FOR_EXAM.
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        edx_course_key (str): An edX course key (CourseRun.edx_course_key)
+
+    bool: whether user has paid or not.
+    """
+    # if FEATURE_SUPPRESS_PAYMENT_FOR_EXAM=True then bypass payment check
+    suppress_payment_for_exam = settings.FEATURES.get("SUPPRESS_PAYMENT_FOR_EXAM", True)
```
rhysyngsun: Change to `False` as we talked about.

exams/utils_test.py
```diff
@@ -123,6 +127,7 @@ def test_exam_authorization(self):
             course=self.course_run.course
         ).exists() is True
 
+    @override_settings(FEATURES={"SUPPRESS_PAYMENT_FOR_EXAM": False})
```
rhysyngsun: Don't need this.

amir-qayyum-khan: we need this, if user turn this feature on using `.env` file
```shell
FEATURES_SUPPRESS_PAYMENT_FOR_EXAM=True
```

here we are making sure , this test is for `FEATURES_SUPPRESS_PAYMENT_FOR_EXAM=False`

rhysyngsun: Ok

PR #2642, opened 2017-02-16
allow fetching dashboard info for other users

### LINKED ISSUE #2531: modify API endpoint for getting dashboard information for another user
part of #2480 

I need to be able to load the enrollment data for another user, so that the staff user can hit an API and get enrollment data for the learner whose profile they are trying to view.

aliceriot:

this closes #2531

#### What's this PR do?

This does just the backend work to support fetching dashboard information based on a username URL parameter. For now the only change in the frontend it to the `getDashboard` function and adding `SETTINGS.user.username` in when necessary.

Make sure that the dashboard and all the frontend stuff works fine. If it does, I think things should be good.

If you wanted you could do the following (but I think it's not essential):

1. edit `DashboardPage.js` or `App.js`, adding:

```js
import * as api from '../lib/api';

window.api = api
```

then, from the JavaScript console, call:

```js
api.getDashboard('fake_username');
```

this should return a 404. confirm getting your own username returns a 200.

then, find the username of a real learner in your database, and ensure your user does not have a `Staff` or `Instructor` role in a program that learner is enrolled in. then call:

```js
api.getDashboard('real_learner_username');
```

this should return a 404. then add a staff or instructor role to your user for that program, and try again. you should now get a 200 and the user's dashboard data.

### DIFF
--- a/dashboard/permissions.py
+++ b/dashboard/permissions.py
@@ -1,0 +1,44 @@
+"""
+Permission classes for the dashboard
+"""
+from django.contrib.auth.models import User
+from django.http import Http404
+from rest_framework.generics import get_object_or_404
+from rest_framework.permissions import BasePermission
+
+from backends.edxorg import EdxOrgOAuth2
+from roles.roles import (
+    Instructor,
+    Staff,
+)
+
+
+class CanReadIfStaffOrSelf(BasePermission):
+    """
+    Only staff on a program the learner is enrolled in can
+    see their dashboard. Learners can view their own dashboard.
+    """
+
+    def has_permission(self, request, view):
+        if request.user.is_anonymous():
+            raise Http404
+
+        user = get_object_or_404(
+            User,
+            social_auth__uid=view.kwargs['username'],
+            social_auth__provider=EdxOrgOAuth2.name
+        )
+
+        # if the user is looking for their own profile, they're good
+        if request.user == user:
+            return True
+
+        # if the user is looking for someone enrolled in a program they
+        # are staff on, they're good
+        if request.user.role_set.filter(
+                role__in=(Staff.ROLE_ID, Instructor.ROLE_ID),
+                program__programenrollment__user=user,
+        ).exists():
+            return True
+        else:
+            raise Http404
--- a/dashboard/permissions_test.py
+++ b/dashboard/permissions_test.py
@@ -1,0 +1,109 @@
+"""
+Tests for the dashboard permissions
+"""
+from unittest.mock import Mock
+from django.http import Http404
+from django.db.models.signals import post_save
+from factory.django import mute_signals
+import ddt
+
+from backends.edxorg import EdxOrgOAuth2
+from courses.factories import ProgramFactory
+from micromasters.factories import UserFactory
+from dashboard.models import ProgramEnrollment
+from dashboard.permissions import CanReadIfStaffOrSelf
+from search.base import MockedESTestCase
+from roles.models import Role
+from roles.roles import (
+    Instructor,
+    Staff,
+)
+
+
+@ddt.ddt
+class CanReadIfStaffOrSelfTests(MockedESTestCase):
+    """
+    Tests
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        with mute_signals(post_save):
+            cls.learner1 = UserFactory.create()
+            cls.learner1_username = 'learner1_username'
+            cls.learner1.social_auth.create(
+                provider=EdxOrgOAuth2.name,
+                uid=cls.learner1_username
+            )
+            cls.learner2 = UserFactory.create()
+            cls.learner2_username = 'learner2_username'
+            cls.learner2.social_auth.create(
+                provider=EdxOrgOAuth2.name,
+                uid=cls.learner2_username
+            )
+            cls.program = ProgramFactory.create()
+            cls.staff = UserFactory.create()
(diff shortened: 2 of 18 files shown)


noisecapella:
I investigated why the selenium test was failing and it looks like it's because you're using `User.username` but the client is passing the edX username. Can you change the view to look up the username in the social auth model?

dashboard/permissions.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Only staff on a program the learner is enrolled in can
+    see their dashboard. Learners can view their own dashboard.
+    """
+
+    def has_permission(self, request, view):
+        if request.user.is_anonymous():
+            raise Http404
+
+        user = get_object_or_404(User, username=view.kwargs['username'])
```
noisecapella: We should look up the user by the social auth username here too

dashboard/permissions_test.py
```diff
@@ -0,0 +1,96 @@
...
+        perm = CanReadIfStaffOrSelf()
+        with mute_signals(post_save):
+            user = UserFactory.create()
+            another_user = UserFactory.create()
+            request = Mock(user=user)
+            view = Mock(kwargs={'username': another_user.username})
+            with self.assertRaises(Http404):
+                perm.has_permission(request, view)
+
+    @ddt.data((Instructor), (Staff))
```
noisecapella: Can you remove the extra parentheses here?

dashboard/permissions_test.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_staff_can_read_learners(self, role):
+        """
+        A staff or instructor on a program a learner is also in should
+        be able to read their profile
```
noisecapella: Can you change to `read their dashboard`?

dashboard/views_test.py
```diff
@@ -80,6 +87,23 @@ def test_get_dashboard(self):
         assert len(result.data) == 2
         assert [self.program_1.id, self.program_2.id] == [res_item['id'] for res_item in result.data]
 
+    @ddt.data((Instructor), (Staff))
```
noisecapella: Can you remove the extra parentheses here?

static/js/actions/index.js
```diff
@@ -35,10 +36,10 @@ export const receiveDashboardFailure = (errorInfo: APIErrorInfo): Action => ({
 export const CLEAR_DASHBOARD = 'CLEAR_DASHBOARD';
 export const clearDashboard = () => ({ type: CLEAR_DASHBOARD });
 
-export function fetchDashboard(noSpinner: boolean = false): Dispatcher<Dashboard> {
+export function fetchDashboard(username: string, noSpinner: boolean = false): Dispatcher<Dashboard> {
   return (dispatch: Dispatch) => {
     dispatch(requestDashboard(noSpinner));
-    return api.getDashboard().
+    return api.getDashboard(username).
       then(dashboard => dispatch(receiveDashboardSuccess(dashboard))).
```
noisecapella: We need to update the reducer to handle multiple dashboards, else dashboards from other users will overwrite each other

dashboard/permissions.py
aliceriot: Hm, I think I was under the impression that `user.username` was the same thing. How would I go about doing that? If someone visits `/profile/${username}` is `username` guaranteed to be someone's edx username?

dashboard/permissions_test.py
aliceriot: huh, before I had to add those because of some error, but now it's not happening! lol

dashboard/permissions.py
noisecapella: Yes, the only username we expose to the outside user is their edX username. Usually this will be the same as `User.username` but if there's a conflict or an error when creating the social auth object it may create a separate user object with some hex digits appended to the username, ie `george6fb83a` for an edX username of `george`. You can look it up by joining the social auth table like this: `User.objects.filter(social_auth__uid='a_username_here')`

dashboard/permissions_test.py
aliceriot: oop good catch

static/js/actions/index.js
aliceriot: yes, I'm planning to do that work in a separate PR to keep the changeset small.

aliceriot: Provided that makes sense to you. All I've changed is on the backend we can now theoretically get the information for another user, but on the frontend all the calls to `getDashboard` use the user's own username. The next PR I do will change the reducer to support multiple dashboards (similar to the setup we have for profiles) and then after that one I will probably work on actually adding the dashboard support to the learner page (along with the UI work, some of which I've already started).

dashboard/permissions.py
aliceriot: OK, I'll make that change.

static/js/actions/index.js
noisecapella: Sounds good to me

dashboard/permissions.py
noisecapella: Actually you should also filter on the provider:

    User.objects.filter(social_auth_uid='a_username_here', social_auth__provider=EdxOrgOAuth2.name)

aliceriot: makes sense!

aliceriot:
@noisecapella I've changed things around to fetch users based on the edx social auth stuff (and refactored some tests as well). ready for review again!

PR #2644, opened 2017-02-16
Moved ipython to requirements.txt

giocalitri:
N/A

#### What's this PR do?
Moves ipython to requirements.txt
It also removes the test_requirements.in (no more pinned test requirements) and sets the pip cache environment variable to `/src/.cache`

1- the tests should pass on travis.
2- connect to the heroku build for this PR with the `./manage.py shell` and verify that ipython has been used (instead of the default python)


noisecapella:
Looks like the test is complaining about a double requirement

requirements.txt
```diff
@@ -32,21 +34,30 @@ elasticsearch-dsl==2.1.0
 elasticsearch==2.3.0
 factory-boy==2.8.1
 faker==0.7.3
-html5lib==0.999999999
+html5lib==0.999999
```
noisecapella: This looks accidental, can you revert?

test_requirements.txt
```diff
@@ -25,7 +25,7 @@ first==2.0.1              # via pip-tools
 greenlet==0.4.11          # via bpython
 ipdb==0.10.1
 ipython-genutils==0.1.0   # via traitlets
-ipython==5.1.0
+ipython==5.1.0            # via ipdb
```
noisecapella: Shoudl we move over ipdb too?

giocalitri:
Yep
this PR also shows that `pyup did not modify the right file

requirements.txt
giocalitri: this is because yup modified the wrong file: I will fix it

test_requirements.txt
giocalitri: ipdb is a debugger: I do not see the reason to move it. 
maybe I can keep ipython here with a strict requirement. I'll test something

noisecapella:
It looks like the selenium failures are actually pylint failures in the pylint directory, maybe because it's not using the same `mm_travis_web` image. Can you rebase on master and see if the problem persists?

giocalitri:
@noisecapella I rebased and this seems to have fixed the selenium tests

.travis.yml
```diff
@@ -7,6 +7,7 @@ matrix:
     - install:
       - env | grep TRAVIS > .env
       - env | grep CI >> .env
+      - echo "XDG_CACHE_HOME=/src/.cache" >> .env
```
noisecapella: Why not put this in `travis-docker-compose.yml`? I'm fine either way but just wondering

giocalitri: no reson: I will move it

giocalitri: it is probably one of the million tests I did on friday

PR #2658, opened 2017-02-21
Use testindex when running selenium tests

noisecapella:
None

#### What's this PR do?
Defines `ELASTICSEARCH_INDEX` when running selenium tests locally, so that running the tests won't clobber your existing index. This change isn't done on `.travis.yml` but it doesn't matter there since there is no other index

Run the selenium tests locally then start the docker container and verify that you can view the `/learners` page without a problem.


giocalitri:
👍

Issue #2564, opened 2017-02-14
Add Navigation - Sidebar


roberthouse54:
The goal of this issue is to add a navigation drawer (sidenav) that will be available on all pages in the app, on both desktop and mobile. The sidenav will have links to all pages, including the links that are currently in the user dropdown into the side nav. The "context switcher" will remain in the topnav, but move to the right side where the user dropdown is now. For mobile devices, the context switcher will be in the sidenav (as it is now).

The sidenav, when open, will cover the content on the page. The menu icon will open and close the sidenav. A chevron arrow will close the sidenav, or swiping left will also close the sidenav.

Staff users, faculty and other admin users who have links to admin pages, we see links to those pages above the "My Profile" link.

_In the future we may change this so the sidenav will be open by default if the device is wide enough to have it be open without covering the content of the page._

![navigation_new 2x](https://cloud.githubusercontent.com/assets/20047260/22991304/d082f5ce-f389-11e6-87c5-b9c0ae255037.jpg)

aliceriot:
@roberthouse54 do you have a mockup of what this should look like on a desktop?

roberthouse54:
Here is the admin version of the sidenav. Links to open in new tabs. On desktop it should look as shown (for now):

![image](https://cloud.githubusercontent.com/assets/20047260/23565316/13e49794-001b-11e7-8ca7-c7d42f80c10b.png)

aliceriot:
@roberthouse54 do you know if the icon you have for 'CMS' is a material-ui icon? I tried to find it but couldn't ;_;

roberthouse54:
It is not. I will have to make it as a custom icon.

roberthouse54:
No. Actually, let's switch this and use the material icon called "Description". I have updated the mock above ^^

aliceriot:
there also isn't a material icon for that graduation hat - currently we're using the one called 'dashboard':

![dasbasdfasdf](https://cloud.githubusercontent.com/assets/6207644/23757255/4168a780-04b4-11e7-9abd-17e0ef7be689.png)

aliceriot:
would it be ok to stick with the one we're using now?

roberthouse54:
Yup. I think I just forgot to update my PSD. That dashboard one is fine.

PR #2660, opened 2017-02-21
Restyle sidenav

roberthouse54:
https://github.com/mitodl/micromasters/issues/2564

#### What's this PR do?
This does the style changes required for #2564

Open and close the sidebar. Make sure the links work.

Check that the css looks like it is properly formatted.

This only partly completes #2564. The other thing required to complete that issue is to make the sidebar always visible on the page, and to remove the user dropdown in the navbar.

sidenav

### DIFF
--- a/static/scss/navbar.scss
+++ b/static/scss/navbar.scss
@@ -47,10 +47,6 @@
           cursor: pointer;
           background: none;
         }
-
-        .Select-menu-outer {
-        }
-
       }
 
       .name {
--- a/static/scss/navbar.scss
+++ b/static/scss/navbar.scss
@@ -49,7 +49,6 @@
         }
 
         .Select-menu-outer {
-          border: none;
         }
 
       }
@@ -77,7 +76,7 @@
     .links {
       display: flex;
       flex-direction: column;
-      border-bottom: thin solid rgba(255,255,255,.2);
+      border-bottom: thin solid $sidenav-separator;
     }
 
     .links, .logout-link {
@@ -239,7 +238,7 @@
   }
 
   .Select-menu-outer,.Select-control,.Select-value-label,.Select-menu-outer,.Select-option {
-    background-color: rgba(0,0,0,.1);
+    background-color: $select-menu-bg;
   }
 
   .Select-menu-outer,.Select-control,.Select-value-label,.Select-placeholder,.Select-option {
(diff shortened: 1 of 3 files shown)


static/scss/navbar.scss
```diff
@@ -243,13 +239,13 @@
...
   .Select-control {
-    border: 1px solid $program_selector_border !important;
+    border: none !important;
     box-shadow: none !important;
     cursor: pointer;
   }
 
   .Select-menu-outer,.Select-control,.Select-value-label,.Select-menu-outer,.Select-option {
-    background-color: $mm-brand-bg !important;
+    background-color: rgba(0,0,0,.1);
```
aliceriot: I think we should generally avoid inline colors and try to declare them all as variables - could you declare this in the `_variables.scss` file? (same for the other inline colors)

aliceriot:
I do think the program selector is somewhat hard to read:

fiplle

what do you think?

aliceriot:
I'm also wondering if the grey-on-black has enough contrast to meet accessibility guidelines, but I don't know if we have a quick way to test for that.

static/scss/navbar.scss
roberthouse54: sure

pdpinch:
You can plug the color numbers into the calculator at http://webaim.org/resources/contrastchecker/

We should aim for at least a WCAG AA pass. 

> I'm also wondering if the grey-on-black has enough contrast to meet accessibility guidelines, but I don't know if we have a quick way to test for that.

roberthouse54:
It passes WCAG AA, but I will lighten it anyway

roberthouse54:
I created variables for the colors, and lightened the font.

static/scss/navbar.scss
aliceriot: there's still an inline `rgba` color here and on line 261 below

static/scss/navbar.scss
```diff
@@ -83,19 +77,21 @@
     .links {
       display: flex;
       flex-direction: column;
-      border-bottom: thin solid $border-color;
+      border-bottom: thin solid rgba(255,255,255,.2);
```
aliceriot: also here

aliceriot:
cool, my thought was that the dropdown for the program selector is a little bit too indistinguishable from the background. here's a screenshot without the chrome debugger overlay:

border

what do you think? I think it is a bit hard to visually distinguish the selector drop-down from the background, but I'm not sure how much of an issue this is / if it's just to my eye. My thoughts are that maybe using a contrasting border for the selector dropdown could make it more obvious? idk, haha.

roberthouse54:
OK, I think I fixed all the things

Issue #2632, opened 2017-02-15
Sort the learner search results by column header


roberthouse54:
The purpose of this is to allow the user to sort columns by clicking on column headers, which is a very common pattern. This was supposed to be done as part of #2479, but we decided to move it out as a separate issue.

The default sort will be by last name. Clicking a column header will sort by that column. Clicking on a column name a second time will reverse the sort order of that column. We will remove the current sorting select method.

![e3e77510-e892-11e6-8e99-62bf8b4a66af](https://cloud.githubusercontent.com/assets/20047260/22996886/562cfb98-f39e-11e6-80d0-744a605a4af7.png)

noisecapella:
@amir-qayyum-khan Sorry that I haven't assigned this to myself earlier, it slipped my mind

amir-qayyum-khan:
its ok

PR #2663, opened 2017-02-21
Added search test

### LINKED ISSUE #2569: Write selenium test for learner search page

noisecapella:
Part of #2632 
Fixes #2569

#### What's this PR do?
Adds a selenium test for learner search. This will make it easier to detect problems that show up when something's not quite right with searchkit

The test should pass

The test runs migrations in `setUp` which is very slow. Each test takes about 35 seconds just to run migrations. At some point we should dump and restore the database, but that's another PR

### DIFF
--- a/selenium_tests/basic_test.py
+++ b/selenium_tests/basic_test.py
@@ -17,6 +17,12 @@ from selenium_tests.base import SeleniumTestsBase
 class BasicTests(SeleniumTestsBase):
     """Basic selenium tests for MicroMasters"""
 
+    def num_elements_on_page(self, selector, driver=None):
+        """Count hits from a selector"""
+        script = "return document.querySelectorAll({selector!r}).length".format(selector=selector)
+        driver = driver or self.selenium
+        return driver.execute_script(script)
+
     def test_zero_price_purchase(self):
         """
         Do a $0 purchase using a 100% off program-level coupon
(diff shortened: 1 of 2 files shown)


selenium_tests/basic_test.py
```diff
@@ -28,3 +40,54 @@ def test_zero_price_purchase(self):
...
+
+        Role.objects.create(
+            user=self.user,
+            program=self.program,
+            role=Staff.ROLE_ID,
+        )
+
+        page_size = settings.ELASTICSEARCH_DEFAULT_PAGE_SIZE
+        with mute_signals(post_save):
+            for _ in range((page_size * 2) - 5):
```
singingwolfboy: Can you explain what this calculation is for? (Ideally, add a comment in the code.)

selenium_tests/basic_test.py
```diff
@@ -28,3 +40,54 @@ def test_zero_price_purchase(self):
...
+        other_program = ProgramFactory.create(live=True)
+        ProgramEnrollment.objects.create(
+            user=self.user,
+            program=other_program,
+        )
+
+        # Update for new users and new role
+        index_program_enrolled_users(ProgramEnrollment.objects.iterator())
+        self.get("{}/learners".format(self.live_server_url))
+        assert self.selenium.execute_script("return document.querySelectorAll('.learner-result').length") == page_size
```
singingwolfboy: Can you write a utility function for this?

```python
def num_elements_on_page(self, selector, driver=None):
    script = "return document.querySelectorAll({selector!r}).length".format(selector=selector)
    driver = driver or self.selenium
    return driver.execute_script(script)
```

selenium_tests/basic_test.py
```diff
@@ -28,3 +40,54 @@ def test_zero_price_purchase(self):
...
+        self.selenium.find_elements_by_class_name('sk-pagination-option')[1].click()
+        # Verify that second page has 5 less elements
+        self.wait().until(
+            lambda driver: driver.execute_script(
+                "return document.querySelectorAll('.learner-result').length"
+            ) == page_size - 5
```
singingwolfboy: There's a lot of logic on this line. Can you break this up?

```python
def verify_num_elements(driver):
    actual_size = self.num_elements_on_page('.learner-result', driver=driver)
    return actual_size == page_size - 5

self.wait().until(verify_num_elements)
```

selenium_tests/basic_test.py
```diff
@@ -28,3 +40,54 @@ def test_zero_price_purchase(self):
...
+        )
+
+        # Go to profile and back to learners to verify that nothing breaks
+        self.selenium.find_element_by_class_name("user-menu").click()
+        self.wait().until(
+            lambda driver: "open" in driver.find_element_by_class_name("user-menu-dropdown").get_attribute("class")
+        )
+        self.selenium.find_element_by_css_selector(
+            ".user-menu-dropdown a[href='/learner/{}']".format(self.username)
+        ).click()
```
singingwolfboy: I think this is better:
```python
profile_link_selector = ".user-menu-dropdown a[href='/learner/{username}']".format(
    username=self.username
)
self.selenium.find_element_by_css_selector(profile_link_selector).click()
```

It would be even better if we were using Python 3.6 and could use f-strings for formatting...

selenium_tests/basic_test.py
noisecapella: I want to have enough so that there are two pages worth of results, but I also want the second page to be smaller than the first so we can assert the page changed correctly. I'll add a comment

noisecapella:
Comments addressed

PR #2666, opened 2017-02-22
Added redis django cache backend

### LINKED ISSUE #2659: Final Grades: change django cache backed to have async process work consistently
Currently the grades apis use the default django cache to save the status of a grading freeze process, but the cache is currently working per process and cannot be read across processes.

We should define an additional cache backend (redis) to be used currently only by the grades api.

This is necessary to run the grade freezing properly in production

giocalitri:
fixes #2659 

#### What's this PR do?
Adds a redis django cache backend to have a persistent cache to be used by the asynchronous tasks. 

The async tasks should just keep working (and there are tests for that).
If you want to test the persistence of the cache:
open the django shell, then:
```
from django.core.cache import caches
cache_redis = caches['redis']
cache_default = caches['default']
cache_redis.set('foo1', 'bar1')
cache_default.set('foo2', 'bar2')
```
if you type 
```
cache_redis.get('foo1')
cache_default.get('foo2')
```
both should give the right result back.
if you exit the django shell and enter again (you re-instanciate the docker instance)
```
cache_redis.get('foo1')
```
should give you the right result while the other should return `None`

### DIFF
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -9,7 +9,7 @@ import pytz
 from celery import group
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.core.cache import cache
+from django.core.cache import caches
 from django.db.models import Q
 from edx_api.client import EdxApi
 from social.apps.django_app.default.models import UserSocialAuth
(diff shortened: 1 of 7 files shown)


blarghmatey:
👍

PR #2667, opened 2017-02-22
Moved sorting UI to column headers

### LINKED ISSUE #2632: Sort the learner search results by column header

noisecapella:
Fixes #2632

#### What's this PR do?
Removes the sort select box and adds column headers to do the sorting instead

Go to the learner page and click the column headers. It should sort the results. You should also be able to refresh the browser and preserve the existing sort

screenshot from 2017-02-22 09-37-39

### DIFF
--- a/static/js/components/search/CustomSortingSelect.js
+++ b/static/js/components/search/CustomSortingSelect.js
@@ -1,58 +1,85 @@
+// @flow
 import React from 'react';
-import ReactDOM from 'react-dom';
-import { Select } from 'searchkit';
-import block from 'bem-cn';
+import Grid, { Cell } from 'react-mdl/lib/Grid';
 
-import type { Event } from '../../flow/eventType';
+import type { SearchSortItem } from '../../flow/searchTypes';
 
-export default class CustomSortingSelect extends Select {
-  showDropdown(sortSelectField: string, clickEvent: Event): void {
-    clickEvent.preventDefault();
-    let event: Event = document.createEvent('MouseEvents');
-    event.initMouseEvent('mousedown', true, true, window);
-    ReactDOM.findDOMNode(sortSelectField).dispatchEvent(event);
-  }
+export default class CustomSortingSelect extends React.Component {
+  props: {
+    items: Array<SearchSortItem>,
+    setItems: (keys: Array<string>) => void,
+    selectedItems: ?Array<string>,
+  };
 
-  renderOptions(
-    translate: Function,
-    showCount: number,
-    countFormatter: Function,
-    items: Array<Object>
-  ): Array<React$Element<*>> {
-    return items.map(item => {
-      let text = translate(item.label || item.title || item.key);
-      if (showCount && item.docCount !== undefined) {
-        text += ` (${countFormatter(item.docCount)})`;
-      }
+  toggleSort = ([defaultSort, otherSort]: [string, string]) => {
+    const { setItems, selectedItems } = this.props;
+    if (selectedItems && selectedItems[0] === defaultSort) {
+      setItems([otherSort]);
+    } else {
+      setItems([defaultSort]);
+    }
+  };
 
-      return <option key={item.key} value={item.key} disabled={item.disabled}>
-        { text }
-      </option>;
-    });
-  }
+  sortDirection = (keys: [string, string]) => {
+    let selectedItem = this.getSelectedItem(keys);
+    if (!selectedItem) {
+      return '';
+    }
+    let order;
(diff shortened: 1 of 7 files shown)


static/js/components/search/CustomSortingSelect.js
```diff
@@ -1,58 +1,85 @@
...
-      items,
-      disabled,
-      showCount,
-      translate,
-      countFormatter
-    } = this.props;
-    const bemBlocks = { container: block(mod) };
+    const nameKeys = ['name_a_z', 'name_z_a'];
+    const locationKeys = ['loc-a-z', 'loc-z-a'];
+    const gradeKeys = ['grade-high-low', 'grade-low-high'];
```
gsidebo: can we move these definition outside of the class or make them static class properties?

static/js/components/search/CustomSortingSelect.js
```diff
@@ -1,58 +1,85 @@
...
-      }
-
-      return <option key={item.key} value={item.key} disabled={item.disabled}>
-        { text }
-      </option>;
-    });
-  }
+import type { SearchSortItem } from '../../flow/searchTypes';
+
+export default class CustomSortingSelect extends React.Component {
```
gsidebo: this is no longer a select box, so it should be called something like `CustomSortingColumnHeaders`

static/js/components/search/CustomSortingSelect.js
```diff
@@ -1,58 +1,85 @@
...
+  props: {
+    items: Array<SearchSortItem>,
+    setItems: (keys: Array<string>) => void,
+    selectedItems: ?Array<string>,
```
gsidebo: can you include a comment saying that these props are passed down by searchkit? if you have any ideas for a brief comment describing what these props are used for, that would help too. `setItems`, for example, is a vague name

static/js/components/search/CustomSortingSelect.js
```diff
@@ -1,58 +1,85 @@
...
+  getSelectedItem = (keys: [string, string]) => {
+    const { selectedItems, items } = this.props;
+    if (!selectedItems) {
+      return '';
+    }
+    return items.find(item => selectedItems[0] === item.key && keys.includes(item.key));
+  };
+
+  selectedClass = (keys: [string, string]) => {
+    let selectedItem = this.getSelectedItem(keys);
```
gsidebo: no need to instantiate the variable here

static/js/components/search/CustomSortingSelect.js
```diff
@@ -1,58 +1,85 @@
...
+        <Cell
+          col={4}
+          onClick={() => this.toggleSort(locationKeys)}
+          className={`residence ${this.selectedClass(locationKeys)}`}
+        >
+          Residence {this.sortDirection(locationKeys)}
+        </Cell>
+        <Cell
+          col={3}
+          onClick={() => this.toggleSort(gradeKeys)}
```
gsidebo: you can create separate `toggleSort` functions using ramda instead of creating these anonymous functions on the fly

```javascript
toggleNameSort = R.partial(toggleSort, [nameKeys]);
toggleLocationSort = R.partial(toggleSort, [locationKeys]);
toggleGradeSort = R.partial(toggleSort, [gradeKeys]);
```

you can also add an event argument to the definition of `toggleSort` and use `R.curry` instead of `R.partial`. i'm cool with either.

static/scss/search-page.scss
```diff
@@ -314,6 +314,31 @@
...
+        padding: 20px 0;
+        text-transform: uppercase;
+        font-size: 12px;
+        color: $font-gray;
+
+        .selected {
+          color: black;
+        }
+
+        >div {
```
gsidebo: effectively this just makes the whole row show a pointer cursor. if we want that, would it be better to put the style definition on `.sorting-row`? if not, i think maybe selecting for `.mdl-cell` would be clearer and more helpful here rather than `>div`. i had to go into chrome dev tools to figure out what mdl renders as a div in order to figure out the reasoning behind this selector. other developers will have to do that too

noisecapella:
Comments addressed

PR #2681, opened 2017-02-23
Use babel-plugin-istanbul to fix coverage

### LINKED ISSUE #2638: JS code coverage doesn't match up perfectly
We should investigate to see what the root cause is, so we can have more confidence in our JS code coverage numbers

noisecapella:
Fixes #2638 

#### What's this PR do?
Uses babel-plugin-istanbul to provide coverage for transformed code.

Run `docker-compose run watch npm run coverage` and look at the results in `coverage/`. The covered lines should match up correctly

#### Any background context?
We could use sourcemaps to fix the stack traces when errors occur but this slows the tests down significantly, so I left it out. We still have correct stack traces anyway when running without coverage.


js_test.sh
```diff
@@ -3,17 +3,17 @@ export TMP_FILE=$(mktemp)
 
 if [[ ! -z "$COVERAGE" ]]
 then
-    export CMD="node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha --"
+    export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=html mocha "
 elif [[ ! -z "$CODECOV" ]]
 then
-    export CMD="node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha -- --report lcovonly -R spec"
```
singingwolfboy: Now that we're not using the CLI that ships with the `istanbul` package, I think we can remove `istanbul` from the dependency list in `package.json`. [`nyc` doesn't depend on it.](https://github.com/istanbuljs/nyc/blob/master/package.json#L74-L102)

js_test.sh
```diff
@@ -3,17 +3,17 @@ export TMP_FILE=$(mktemp)
 
 if [[ ! -z "$COVERAGE" ]]
 then
-    export CMD="node ./node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha --"
+    export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=html mocha "
```
singingwolfboy: Why `--reporter=html`?

js_test.sh
noisecapella: Sure

js_test.sh
noisecapella: This was producing HTML before, but with nyc I needed to make it explicit

Issue #2682, opened 2017-02-23
Layout issues on learner search page


roberthouse54:
This was fixed in the past, and something changed to make it come back. There is too much space above the buttons at the top of the learner search. I will wait until the new sorting by column headers is merged, then fix this along with other small style tweaks on this page.

Some other small layout/style tweaks I will address in this:

- vertically center the Company facet selector
- lighten font weight for name
- fix spacing around column headers
- make avatars smaller
- tighten row spacing
- remove the "Program Avg grade" copy in table cells.

![image](https://cloud.githubusercontent.com/assets/20047260/23264629/040fb784-f9b0-11e6-8bcd-da246cd67429.png)

pdpinch:
@roberthouse54 was this fixed by #2688?

roberthouse54:
Yes, sorry I forgot to close it.

PR #2688, opened 2017-02-24
Style changes to the Learner Search page

roberthouse54:
#2682 

#### What's this PR do?
This fixes a small layout problem on the Learner search page, and makes several small style changes.

Make sure the page looks ok, nothing is broken, css looks correct.

Here are the changes:
- vertically center the Company facet selector text
- lighten font weight for user name
- fix spacing around column headers
- make search box and text smaller
- make user avatars smaller
- tighten row spacing
- remove the "Program Avg grade" copy in table cells.
- changed the column widths
- changed the font of the facet headers

image


singingwolfboy:
The alignment of the column headers and column data is incorrect. Here's a screenshot highlighting what I mean:
incorrect alignment

Is this something we can fix?

roberthouse54:
I honestly cannot figure out what is happening there. @singingwolfboy Can you see why those are not aligning?

PR #2691, opened 2017-02-24
Fixed search filtering involving query parameters

### LINKED ISSUE #2685: Program filter is ignored for certain programs, unless there is text in the textbox
For micromasters-rc you can see this on Digital Learning. This is also happening in production for SCM

noisecapella:
Fixes #2685 

#### What's this PR do?
Fixes filtering of elasticsearch queries to deal with some quirks in searchkit. Basically:
 - The `update_from_dict` is moved to happen first so that server side filtering will overwrite it if necessary
 - The root level filter element should actually be post_filter, so one is copied over the other. They seem to have equivalent functionality AFAIK. Searchkit is not using `post_filter` in `0.10.0` but it will switch in the next release to `post_filter` so the code should handle both cases.

Have two programs with users and only have a staff role in one of them. Log in as the staff user. You should see all users when you view the learners page for the program you are staff for, and no users or facets for the other program.

After this, try typing in a common letter in the text box. It should match some user in the program you are staff on, but it should not match anything on the other program.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -83,10 +83,11 @@ def create_search_obj(user, search_param_dict=None, filter_on_email_optin=False)
     """
     search_obj = Search(index=settings.ELASTICSEARCH_INDEX, doc_type=DOC_TYPES)
     # Update from search params first so our server-side filtering will overwrite it if necessary
-    search_obj.update_from_dict(search_param_dict)
-    # Early versions of searchkit use filter which isn't in the DSL but it acts equivalently to post_filter
-    if 'filter' in search_param_dict:
-        search_obj = search_obj.post_filter(search_param_dict['filter'])
+    if search_param_dict is not None:
+        search_obj.update_from_dict(search_param_dict)
+        # Early versions of searchkit use filter which isn't in the DSL but it acts equivalently to post_filter
+        if 'filter' in search_param_dict:
+            search_obj = search_obj.post_filter(search_param_dict['filter'])
 
     # Limit results to one of the programs the user is staff on
     search_obj = search_obj.filter(create_program_limit_query(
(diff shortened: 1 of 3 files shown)


search/api.py
```diff
@@ -69,29 +76,33 @@ def create_search_obj(user, search_param_dict=None, filter_on_email_optin=False)
...
         Search: elasticsearch_dsl Search object
     """
     search_obj = Search(index=settings.ELASTICSEARCH_INDEX, doc_type=DOC_TYPES)
-    # the following filter should come first because the sequence matters in applying them
+    # Update from search params first so our server-side filtering will overwrite it if necessary
+    if search_param_dict is not None:
+        search_obj.update_from_dict(search_param_dict)
+        # Early versions of searchkit use filter which isn't in the DSL but it acts equivalently to post_filter
+        if 'filter' in search_param_dict:
+            search_obj = search_obj.post_filter(search_param_dict['filter'])
```
giocalitri: why not just replacing the key in `search_param_dict` and then just load the query?

search/api.py
```diff
@@ -69,29 +76,33 @@ def create_search_obj(user, search_param_dict=None, filter_on_email_optin=False)
...
-    ))
-    if search_param_dict is not None:
-        search_param_dict['size'] = settings.ELASTICSEARCH_DEFAULT_PAGE_SIZE
-    else:
-        search_param_dict = {'size': settings.ELASTICSEARCH_DEFAULT_PAGE_SIZE}
-
-    search_obj.update_from_dict(search_param_dict)
+    search_obj = search_obj.filter(
+        Q('term', **{'profile.filled_out': True})
+    )
```
giocalitri: why did you change this?

search/api.py
noisecapella: I figured it was cleaner to do it using the search object than to edit the dictionary. I also wanted to leave `search_param_dict` unmodified since it's being passed as an input value

search/api.py
noisecapella: It didn't seem necessary to put the term filter inside a `bool` and `must` when there's only one expression

giocalitri: ok

giocalitri:
works as expected 👍

PR #2692, opened 2017-02-24
refactor dashboard reducer to support multiple dashboards

### LINKED ISSUE #2672: refactor dashboard reducers to handle multiple users
following up on #2531 (which adds the capability to the backend to fetch the dashboard for other users). this issue would cover refactoring the reducer code on the frontend to deal with this properly.

aliceriot:

closes #2672 

#### What's this PR do?

This refactors the dashboard reducer to handle multiple users, like so:

```js
dashboard = {
  first_user_username: DashboardForFirstUser,
  second_user_username: DashboardForSecondUser,
};
```

I had to fix a bunch of tests and make some changes to the container components which use the dashboard as well.

Do a functional test of the dashboard! Things should work as normal.

### DIFF
--- a/static/js/reducers/dashboard.js
+++ b/static/js/reducers/dashboard.js
@@ -1,5 +1,6 @@
 // @flow
 import _ from 'lodash';
+import R from 'ramda';
 
 import {
   REQUEST_DASHBOARD,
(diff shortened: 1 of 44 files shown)


static/js/reducers/dashboard.js
```diff
@@ -13,39 +14,48 @@ import {
...
-const INITIAL_DASHBOARD_STATE: DashboardState = {
+export const INITIAL_DASHBOARD_STATE: DashboardState = {
   programs: []
 };
 
-export const dashboard = (state: DashboardState = INITIAL_DASHBOARD_STATE, action: Action) => {
+const INITIAL_DASHBOARDS_STATE: DashboardsState = {};
+
+const updateDashboardState = (state, username, update) => (
+  _.merge({}, state, {[username]: update })
```
noisecapella: Why are we using `_.merge` here? It would merge dashboard states together, but it seems like we just want to replace the old one with the new one if we have a newer version of a dashboard

aliceriot: Basically I implemented it this way because there are situations where the old state looks something like this:

```js
[username]: {
  programs: [],
}
```

and we want to, for instance, update the state to have a `fetchProcessing` key on it, but we want to keep around the `programs` array. I hadn't considered that this might lead to some strange stuff with the dashboard being merged together.

In particular if you do something like (just tested this out in the node repl):

```js
let obj = { username: { programs: [ {foo: 'bar' }], otherProp: 'potato' } };
_.merge({}, obj, {username: { programs: [] } }) 
// { username: { programs:[ {foo:"bar"} ], otherProp:"potato" } }
```

the old 'programs' value isn't overwritten by the new empty array.

I currently use `R.dissoc` to delete any user specific state when we handle the `CLEAR_DASHBOARD` action, I could do the same thing for some of the other actions too.

aliceriot: How do you think we should handle it?

noisecapella: Hmm... maybe something like this would work:

    {...obj, {username: { ...obj.username, ...newFields } } }

Does `...` work for undefined values?

pdpinch:
Maybe an irrelevant question but, what's the meaning of `first` and `second` here?

```
dashboard = {
  first_user_username: DashboardForFirstUser,
  second_user_username: DashboardForSecondUser,
};
```

aliceriot:
@pdpinch just two different test usernames, could be `a_user`, `b_user` just as easily.

static/js/reducers/dashboard.js
aliceriot: I believe using the spread operator with an undefined value will give a `TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined`

aliceriot: I think I'll just add a call to `R.dissoc` where it makes sense to do so.

pdpinch:
so the current user can request dashboards for multiple other users?

aliceriot:
yes

static/js/reducers/dashboard.js
noisecapella: Sounds good

aliceriot: Just pushed the change, lmk what you think!

noisecapella:
Looks good :+1:

pdpinch:
Sorry, I'm still confused. Does this one reducer now address both the Dashboard (i.e. `/dashboard`) and the staff view of the Learner Profile (e.g. `/learner/{username}`) -- is that why it needs to work for multiple users?

aliceriot:
yea, exactly

Issue #2196, opened 2016-12-22
Allow users to send email to a single user from the user chip, or the profile page.


roberthouse54:
This issue involves allowing users (students and admin) to send messages to other single users via a link from the user chip and the profile page. In both cases we will open a "Send a Message" modal. 

Since this will be the 3rd e-mail UI added to the react app, some refactoring may be desirable -- see #2468 

*** acceptance criteria
- [ ] Add direct email link to the profile page (will replace the user's displayed email address)
- [ ] Restrict this feature to staff and paid learners. ~~For learners who haven't paid, we will display an upsell message~~ _(this upsell message will be implemented in a separate issue)_

![image](https://cloud.githubusercontent.com/assets/20047260/21431928/5b1816ac-c837-11e6-9d5b-059c112a2fdd.png)

![image](https://cloud.githubusercontent.com/assets/20047260/22706559/bfb1f756-ed3d-11e6-8fab-b666bbdcd893.png)

![image](https://cloud.githubusercontent.com/assets/20047260/22706196/5a2b7836-ed3c-11e6-855b-8686811a6832.png)

![image](https://cloud.githubusercontent.com/assets/20047260/22706439/3e4c9130-ed3d-11e6-9677-bb13effcf284.png)

gsidebo:
@pdpinch @roberthouse54 A few important design questions here:

1. Do we currently have some way for learners to actually find other learners? I don't see one
1. The design for the user profile page looks different from what we actually have implemented. Is there supposed to be some design work done before this?
1. One thing we talked about as a benefit of implementing one-way learner-to-learner email in general was to avoid exposing learners' emails to each other. Is that something we're still interested in? If so, we should probably get rid of the email address display when we add the link.

Basically I see 2 places that we can actually implement this right now: (1) on user profile pages, which only staff can reasonably find at the moment, and (2) the user chip on the search results page, which is also staff-only. Is that correct? If so, we should reflect that in the issue description

roberthouse54:
1. We do not currently have this ability for learners to find other learners, but it has been our plan to do it.
2. As to differences in the profile---- do not make changes to the profile that are not required for this feature.
3. Yes, I see your point. Let's not display the email.

pdpinch:
This is blocked by #2641

gsidebo:
[...]

PR #2693, opened 2017-02-27
Added staff-to-learner email with link in learner chip

### LINKED ISSUE #2641: Enable staff direct email to learners via the learner chip
_Split from https://github.com/mitodl/micromasters/issues/2196_

We want to enable direct one-on-one communication with learners. Since designs are still pending and there are some decisions that might need to be made about learner-to-learner email behavior, this issue will cover the work to enable staff-to-learner communication. Most of this functionality should be directly reusable for learner-to-learner messaging.

Acceptance criteria:

1. "Send a Message" icon & link added to the learner chip ~~and the learner profile page (this link will replace the user's displayed email address on the profile page)~~ (this will be implemented in #2196)
1. Icon & link launch an email dialog to send a message
1. API endpoint added to send a direct email message, ensuring that the sender has staff permission
1. Sender's email address is the 'from:' address in the sent email

gsidebo:
Closes #2641

#### What's this PR do?
Implements the API endpoint to send a staff-to-learner email, and adds the link on the learner chip to launch the email dialog

From the /learners page, hover over a learner's name, click the email link, send an email, and repeat for a couple of other learners. Make sure `MAILGUN_RECIPIENT_OVERRIDE` is set to your email address

LearnerResult.js and LearnerChip.js


static/js/components/LearnerChip.js
```diff
@@ -18,12 +19,16 @@ const LearnerChip = ({ profile }: {profile: Profile}): React$Element<*> => (
         { mstr(getEmployer(profile)) }
       </span>
       <a href={`/learner/${profile.username}`} className="mm-minor-action">
-        <Icon name="person" />
-        <span>View profile</span>
+        <Icon name="person"/>
+        <span>View Profile</span>
+      </a>
+      <a onClick={openLearnerEmailComposer} className="mm-minor-action">
```
singingwolfboy: This should be a `<button>` element, for accessibility reasons. See #1222.

static/scss/user-chip.scss
```diff
@@ -32,6 +32,7 @@
 
     a:hover {
       text-decoration: none !important;
+      cursor: pointer;
```
singingwolfboy: If this is only because of the `<a onClick>` element referenced above, then after changing it to a `<button>` element, this line should be removed. (Or is there some other reason why this was added?)

gsidebo:
@roberthouse54 here are some screenshots. let me know what you think:

ss 2017-02-28 at 11 04 04

ss 2017-02-28 at 11 04 34

ss 2017-02-28 at 11 04 42

static/js/components/search/LearnerResult_test.js
```diff
@@ -137,6 +143,21 @@ describe('LearnerResult', () => {
...
+    let openLearnerEmailComposerStub = helper.sandbox.stub();
+
+    let result = renderLearnerResult({openLearnerEmailComposer: openLearnerEmailComposerStub});
+    return helper.listenForActions([SET_LEARNER_CHIP_VISIBILITY], () => {
+      result.find(".learner-name").props().onMouseEnter();
+    }).then(() => {
+      // Execute the openLearnerEmailComposer function passed down to the LearnerChip
+      result.find("LearnerChip").props().openLearnerEmailComposer();
+      assert.isTrue(openLearnerEmailComposerStub.called);
+      assert.isTrue(openLearnerEmailComposerStub.calledWith(USER_PROFILE_RESPONSE));
```
singingwolfboy: I recommend using `sinon.assert.called()` and `sinon.assert.calledWith()` instead of `assert.isTrue()` for these cases. You get a much clearer failure message if the assertion fails. http://sinonjs.org/releases/v1.17.7/assertions/

static/scss/shared-styles.scss
```diff
@@ -0,0 +1,13 @@
+img.rounded-profile-image {
+  &.small {
+    height: 43px;
+    border-radius: 50px;
+  }
+
+  &.medium {
+    height: auto !important;
```
singingwolfboy: Why do you need `!important` here?

mail/views.py
```diff
@@ -23,11 +25,45 @@
...
+    queryset = Profile.objects.all()
+
+    def post(self, request, *args, **kargs):  # pylint: disable=unused-argument
+        """
+        POST method handler
+        """
+        sender_user = request.user
+        recipient_user = self.get_object().user
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
```
singingwolfboy: Please add a test to cover an invalid serialization error.

micromasters/urls_test.py
```diff
@@ -20,7 +20,7 @@ def test_urls(self):
         assert reverse('checkout') == '/api/v0/checkout/'
         assert reverse('user_program_enrollments') == '/api/v0/enrolledprograms/'
         assert reverse('user_course_enrollments') == '/api/v0/course_enrollments/'
-        assert reverse('search_result_mail_api') == '/api/v0/mail/'
+        assert reverse('search_result_mail_api') == '/api/v0/mail/search/'
```
singingwolfboy: It probably also makes sense to add a test for `learner_mail_api`

static/js/actions/email.js
```diff
@@ -31,6 +31,23 @@ export const sendEmailSuccess = createAction(SEND_EMAIL_SUCCESS);
 export const SEND_EMAIL_FAILURE = 'SEND_EMAIL_FAILURE';
 export const sendEmailFailure = createAction(SEND_EMAIL_FAILURE);
 
+export function sendEmail(
+  emailType: string,
```
singingwolfboy: Is `emailType` an arbitrary string, or does it have to be one of a short list of strings? If the latter, please specify that list in the Flow type.

static/js/actions/email.js
```diff
@@ -31,6 +31,23 @@ export const sendEmailSuccess = createAction(SEND_EMAIL_SUCCESS);
...
+  sendFunctionParams: Array<*>
+): Dispatcher<EmailSendResponse> {
+  return (dispatch: Dispatch) => {
+    dispatch(initiateSendEmail(emailType));
+    return sendFunction.apply(null, sendFunctionParams).
+      then(response => {
+        dispatch(sendEmailSuccess(emailType));
+        return Promise.resolve(response);
+      }).catch(error => {
+        dispatch(sendEmailFailure({type: emailType, error: error}));
```
singingwolfboy: Should this return `Promise.reject(error)`? And is there any UI to notify the user in case the email fails to send? A toast message would suffice.

static/js/components/LearnerChip_test.js
```diff
@@ -9,40 +10,49 @@ import { getPreferredName } from '../util/util';
...
+    assert.equal(url, `/learner/${profileClone.username}`);
+  });
+
+  it('should provide a link to email the learner', () => {
+    let sandbox = sinon.sandbox.create();
+    let openLearnerEmailComposer = sandbox.stub();
+    let chip = renderChip(profileClone, openLearnerEmailComposer);
+    let emailLink = chip.find('button').at(0);
+    emailLink.simulate('click');
+    assert.equal(openLearnerEmailComposer.callCount, 1);
```
singingwolfboy: It would be better to use [`sinon.assert.calledOnce(openLearnerEmailComposer)`](http://sinonjs.org/releases/v1.17.7/assertions/#sinonassertcalledoncespy). It's more immediately clear, and Sinon provides a useful failure message if the assertion fails.

static/js/components/search/LearnerResult_test.js
```diff
@@ -137,6 +143,21 @@ describe('LearnerResult', () => {
...
+  it('should pass down a function that can open a dialog to email a learner', () => {
+    helper.store.dispatch(setLearnerChipVisibility(USER_PROFILE_RESPONSE.username));
+    let openLearnerEmailComposerStub = helper.sandbox.stub();
+
+    let result = renderLearnerResult({openLearnerEmailComposer: openLearnerEmailComposerStub});
+    return helper.listenForActions([SET_LEARNER_CHIP_VISIBILITY], () => {
+      result.find(".learner-name").props().onMouseEnter();
+    }).then(() => {
+      // Execute the openLearnerEmailComposer function passed down to the LearnerChip
+      result.find("LearnerChip").props().openLearnerEmailComposer();
```
singingwolfboy: Can you stick in an assertion, to be sure that we find the LearnerChip? Something like this:
```js
const chip = result.find("LearnerChip");
assert.lengthOf(chip, 1);
chip.props().openLearnerEmailCompose();
```

The way this is currently written, if the LearnerChip is not rendered, you'll get a misleading error message about calling `.props()` on nothing. This will make that failure case more clear.

static/js/flow/emailTypes.js
```diff
@@ -31,5 +29,5 @@ export type EmailConfig = {
   title: string,
   renderSubheading: (subheading: string) => React$Element<*>,
   emailOpenParams: (args: any) => Object,
-  emailSendAction: (emailState: EmailState) => Dispatcher<EmailSendResponse>
+  emailSendParams: (emailState: EmailState) => Array<any>
```
singingwolfboy: We should avoid using the `any` type as much as possible. What type or types should this be?

static/js/lib/api_test.js
```diff
@@ -216,6 +218,20 @@ describe('api', function() {
         });
       });
 
+      it('returns expected values when a POST to send a learner email succeeds', () => {
+        fetchJSONStub.returns(Promise.resolve(MAIL_RESPONSE));
+        return sendLearnerMail('subject', 'body', learnerStudentId).then(mailResp => {
+          assert.ok(fetchJSONStub.calledWith(`/api/v0/mail/learner/${learnerStudentId}/`, {
```
singingwolfboy: It would be better to use [`sinon.assert.calledWith()`](http://sinonjs.org/releases/v1.17.7/assertions/#sinonassertcalledwithspy-arg1-arg2-).

static/js/util/util.js
```diff
@@ -517,6 +518,20 @@ export function getDisplayName(WrappedComponent: ReactClass<*>) {
   return WrappedComponent.displayName || WrappedComponent.name || 'Component';
 }
 
+export const wrapWithProps = (addedProps: Object, WrappedComponent: ReactClass<*>) => {
```
singingwolfboy: Nitpick: can you add Flow type information that this function returns a `ReactClass`?

mail/views_test.py
```diff
@@ -199,6 +199,53 @@ def test_course_team_email_unpaid(self):
...
+
+    @patch('mail.views.MailgunClient')
+    def test_send_learner_email_view(self, mock_mailgun_client):
+        """
+        Test that learner emails are correctly sent through the view
+        """
+        self.client.force_login(self.staff_user)
+        mock_mailgun_client.send_individual_email.return_value = Mock(
+            spec=Response,
+            status_code=status.HTTP_200_OK,
```
singingwolfboy: Can you add another test simulating a failure from Mailgun?

singingwolfboy:
There's a usability problem with this feature, although fixing it might be outside the scope of this pull request. If you open the "Send a Message" dialog, type in a message, close the dialog without sending the message, and reopen it, the message you've typed in is gone. These sorts of dialogs are easy to close accidentally, just by clicking outside of the dialog. I could also imagine an instructor writing a message to a student, closing the dialog to check the student's grade in the program, and then reopening the dialog expecting the message to be saved.

Can you save the contents of the message when the dialog is closed, and repopulate the dialog with that message when it's reopened? The saved contents would only get cleared after the email is successfully sent, which would help deal with Mailgun failures. If Mailgun is experiencing a temporary failure, the instructor doesn't have to type out the whole message again -- just reopen the dialog and click send again!

singingwolfboy:
Also, I don't know if we care about viewing this dialog of mobile, but the layout is wrapping when it doesn't need to.
screen shot 2017-03-01 at 10 08 35 am

mail/views_test.py
```diff
@@ -58,7 +58,7 @@ def setUpTestData(cls):
         }
 
     def setUp(self):
-        super(MailViewsTests, self).setUp()
+        super(SearchResultMailViewsTests, self).setUp()
```
noisecapella: Not reviewing, but you can make this just `super().setUp()` in Python 3

static/scss/shared-styles.scss
gsidebo: this was copied from the original style. not sure i want to find out why it's there as a part of this PR

gsidebo:
@singingwolfboy regarding the clearing of the dialog when you cancel out of it, this has been a 'feature' as long as we've had the email dialogs. i agree that the input should be saved. i'll open a separate issue for it

static/scss/shared-styles.scss
singingwolfboy: Gotcha. @roberthouse54, can you take a look at this? I've created #2732 to address this and other issues.

static/js/actions/email.js
gsidebo: nothing in the UI. this will have to be a separate issue

static/js/flow/emailTypes.js
gsidebo: it can literally be anything. for one implementation, it's a complex ES query object. for another implementation, it's an integer ID

static/js/actions/email.js
gsidebo: it can theoretically be any string. `static/js/components/email/constants.js` has a short list of strings that we can use for known email types, but this argument to `sendEmail` just depends on the keys passed into the `withEmailDialog` higher-order component (see: `static/js/containers/DashboardPage.js:732`). those keys can be anything

mail/views_test.py
gsidebo: we don't currently have tests that simulate a mailgun failure for any of the other email views. it will involve patching `requests.post` or something like that. i made a quick attempt at it and failed. it's non-trivial and i would very strongly prefer creating a separate issue for that work. if you have any objections let me know

singingwolfboy: You don't need to patch the `requests` module, just do the same mocking that you're doing here, but have it return 400 Bad Request status code. It's worth verifying that, if Mailgun returns a non-successful HTTP response, then our code should also return a non-successful HTTP response.

gsidebo:
@singingwolfboy should be all set

gsidebo:
@singingwolfboy these merge conflicts are making me very very sad. i don't want to push, but the sooner you can review this, the better.

singingwolfboy:
The functionality looked good when I tested it earlier, and the code problems are resolved. However, you're still missing the following things that I pointed out earlier:

* A test that mocks `MailgunClient.send_individual_email()` to return an error response, and verifies that our code also returns an error response
* Some kind of UI to report a Mailgun error. Currently, this produces silent failures, which are the worst kind of failures. As I said, a simple toast notification would suffice -- we can improve it later, if necessary.

I'm out for the rest of the week, back on Tuesday. I certainly don't mind having someone else take over the review for this, if you need it done before then.

gsidebo:
@singingwolfboy https://github.com/mitodl/micromasters/issues/2749 for adding tests that simulate a Mailgun failure (which we do not have for any of the other views), and https://github.com/mitodl/micromasters/issues/2748 for adding a front-end notification if an email fails. I'll see who else can take over this review tomorrow

noisecapella:
Can you rebase please?

noisecapella:
I see the email link on the learner chip, should there also be one on the learner profile page?

mail/views.py
```diff
@@ -31,6 +32,41 @@
...
+class LearnerMailView(GenericAPIView):
+    """
+    View class that handles HTTP requests to learner mail API
+    """
+    serializer_class = GenericMailSerializer
+    authentication_classes = (
+        authentication.SessionAuthentication,
+        authentication.TokenAuthentication,
+    )
+    permission_classes = (permissions.IsAuthenticated, UserCanMessageLearnersPermission, )
```
noisecapella: We need to also verify that the staff is staff of a program that the user is enrolled in. We don't have an extra permission in the search view because we add a filter on the programs the staff is enrolled in via Elasticsearch, but here we should check this in a permission class

mail/views_test.py
```diff
@@ -252,6 +252,72 @@ def test_course_team_email_unpaid(self):
         assert resp.status_code == status.HTTP_403_FORBIDDEN
 
 
+class LearnerMailViewTests(APITestCase, MockedESTestCase):
```
noisecapella: Can you add a test where a user who is not staff attempts to use this API to send an email?

noisecapella:
Functionality looks great

static/js/components/LearnerSearch.js
```diff
@@ -126,6 +129,14 @@ export default class LearnerSearch extends SearchkitComponent {
   countryNameTranslations: Object = makeCountryNameTranslations();
   degreeTranslations: Object = makeDegreeTranslations();
 
+  constructor(props: Object) {
+    super(props);
+    this.WrappedLearnerResult = wrapWithProps(
+      {openLearnerEmailComposer: this.props.openLearnerEmailComposer},
```
noisecapella: I don't think it's a good idea to set this in the constructor. What if `openLearnerEmailComposer` changes later?

static/js/components/email/lib.js
```diff
@@ -25,28 +31,65 @@ export const COURSE_TEAM_EMAIL_CONFIG: EmailConfig = {
...
+  getEmailSendFunction: () => sendSearchResultMail,
+
+  emailSendParams: (emailState) => ([
+    emailState.inputs.subject || '',
+    emailState.inputs.body || '',
+    emailState.searchkit.buildQuery().query
+  ])
+};
+
+export const LEARNER_EMAIL_CONFIG: EmailConfig = {
```
noisecapella: Can you add tests which cover these functions, either via unit tests and mocks or with an integration test?

gsidebo:
@noisecapella the profile page link will be implemented as part of #2196. Sorry, I had it in mind that I would just implement the learner chip link in this issue and the profile page link in that other issue, but forgot to update the descriptions. They are updated now

static/js/components/LearnerSearch.js
gsidebo: I did this because there was a visually-noticeable performance penalty if the component wrapping is performed in, say, the `render` method. It's also important to note that this method can't change. It's handed down from the email HOC, which is only executed once on the container component, and the function is not part of the state, so it can't change via an action/reducer.

I agree that this is kind of a code smell and it isn't by-the-book React, but I don't quite know how else to handle this without repeatedly re-wrapping this component unnecessarily.

gsidebo:
@noisecapella that should do it

static/js/components/LearnerSearch.js
noisecapella: It would be nice to investigate what the performance penalty was, but feel free to leave as is for this PR

Issue #2510, opened 2017-02-06
Automatically send email in the future, as learners meet criteria


pdpinch:
Staff need to send some emails based on the learner meeting criteria (e.g. having passed two courses, or a new user who employer makes them eligible for a discount). These learners can be identified and messaged via the learner search, but it's extra work and error prone for staff to have to repeatedly check for new learners who meet old criteria. 

Instead, we'll give staff the option to save a query and a message, and have it sent automatically whenever a new learner meets the criteria. 

After discussing options with the team, it seems like using the Percolate function of elastic search is the preferred way to handle this. See this blog post for more information: https://www.elastic.co/blog/when-and-how-to-percolate-1

The UI for this is a simple checkbox in the message pop-up dialog, when it's generated with a search query. 

![future_send_this_email 2x](https://cloud.githubusercontent.com/assets/430126/22658436/0ba24ea8-ec68-11e6-8c2c-b020998b56c0.jpg)

Implementation:
 - [x] Create the model to store percolator queries #2701 
 - [x] On the backend update the mail API to store the value of the query if the checkbox is checked #2704
 - [x] update the redux infrastructure to pass the checkbox value to the REST API #2706
 - [x] add the checkbox in the UI #2707
 - [x] add code/models to match all new user documents and keep track of emails sent #2705

pdpinch:
@noisecapella could you update this ticket with a brief overview of your plan? I see you opened a number of new issues. #2706 in particular I don't quite understand.

noisecapella:
I updated the description

noisecapella:
Do we need to support any kind of templating for the subject and body?

noisecapella:
Also, should we send the email after they complete the profile signup or just when the new user gets created? I'm thinking waiting until after the profile is filled out would give us a better ability to filter users

pdpinch:
I like waiting until the profile is completed. In particular, we want to be able to filter on program. 

Regarding templates, yes, we'd would like to support templating, but I'm not sure exactly how to do this.

noisecapella:
It looks like mailgun supports this in their API: https://documentation.mailgun.com/user_manual.html#batch-sending

I'm going to leave it alone for these issues to keep the scope of what I'm doing narrow, but it shouldn't be hard to add this functionality later

pdpinch:
@noisecapella is this done now that #2707 has been merged?

noisecapella:
Yes

PR #2701, opened 2017-02-27
Added PercolateQuery model

noisecapella:
Part of #2510 

#### What's this PR do?
Adds `PercolateQuery` model which adds and removes itself from Elasticsearch when its contents change.

Create a `PercolateQuery` locally with `{"query": {}}` as `query` (or something else that has a valid syntax). Go to `http://localhost:9100/micromasters/.percolator/{your_id}` and verify that the entry has the same query that you entered.


search/signals.py
```diff
@@ -22,37 +28,50 @@
...
-def handle_delete_employment(sender, instance, **kwargs):  # pylint: disable=unused-argument
+def handle_delete_employment(sender, instance, **kwargs):
     """Update index when Employment model instance is deleted."""
     index_users.delay([instance.profile.user])
+
+
+@receiver(post_save, sender=PercolateQuery, dispatch_uid="percolate_query_save")
+def handle_update_percolate(sender, instance, **kwargs):
+    """When a new query is created or a query is updated, update Elasticsearch too"""
+    index_percolate_queries.delay([instance])
```
noisecapella: Ah, I didn't think about the memory issues. I'll pass the id and refactor the other functions to do the same

search/signals.py
```diff
@@ -22,37 +28,50 @@
...
+
+
+@receiver(post_delete, sender=PercolateQuery, dispatch_uid="percolate_query_delete")
+def handle_delete_percolate(sender, instance, **kwargs):
+    """When a query is deleted, make sure we also delete it on Elasticsearch"""
+    delete_percolate_query.delay(instance)
```
noisecapella: The instance is gone anyway at this point, it just  but I'll change this to pass an id too

rhysyngsun:
Assuming tests/travis pass

Issue #1922, opened 2016-12-01
Performance: use Cloudfront with S3 assets


noisecapella:
Currently this is our profile photos and CMS images

pdpinch:
This should just require some configuration of S3 and cloudfront, possibly including some nginx configuration.

PR #2703, opened 2017-02-27
Load SCSS-referenced URLs via Cloudfront on production

### LINKED ISSUE #2462: Not all static assets are served from Cloudfront
Static assets called from [here](https://github.com/mitodl/micromasters/blob/73926efeb12dd7033d8f8e07f43cb7275885872f/static/scss/homepage.scss) are still being served through micromasters as opposed to Cloudfront. This can be verified and tested in Chrome Developer Tools under Sources. 
screen shot 2017-01-31 at 10 33 18 am

singingwolfboy:
Fixes #2462

#### What's this PR do?
Rewrites URL references in our CSS files to point to Cloudfront on production.

The easy way: wait until this is on the release candidate server, and then verify that the hero image on the landing page (`/static/images/lp_hero.jpg`) is loaded via Cloudfront.

The hard way: first, checkout this branch locally and run `node node_modules/webpack/bin/webpack.js --config webpack.config.prod.js --bail` to compile all the assets via webpack. The compiled versions will end up in the `static/bundles` directory. Open the `style-[hash].js` file, search for `lp_hero.jpg`, and verify that the URL will be _locally_ resolved (and does _not_ reference Cloudfront). This is important to verify, so that local development continues to work without relying on Cloudfront.

Then, set the `CLOUDFRONT_DIST` environment variable to some value, and run the same command again. Open the new `style-[hash].js` file, search for `lp_hero.jpg`, and verify that the URL will be resolved from the Cloudfront distribution you set. For example, if you set `CLOUDFRONT_DIST=abc`, the file should contain `https://abc.cloudfront.net/static/images/lp_hero.jpg`.

#### Additional notes
With this change, we can also use the `resolve()` function in CSS to load files from arbitrary locations, like we do with templates. We can also use `width()`, `height()`, and `size()` functions to get information about images. See the [PostCSS-Assets](https://github.com/borodean/postcss-assets) project for more information.

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,7 +1,20 @@
+const autoprefixer = require('autoprefixer');
+const assets = require('postcss-assets');
+
+const CF_DIST = process.env.CLOUDFRONT_DIST;
+let baseUrl = '/';
+if (CF_DIST) {
+  baseUrl = `https://${CF_DIST}.cloudfront.net/`;
+}
+
 module.exports = {
   plugins: [
-    require('autoprefixer')({
+    autoprefixer({
       browsers: ['> 1%']
+    }),
+    assets({
+      baseUrl: baseUrl,
+      loadPaths: ['static']
     })
   ]
 }
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "node-sass": "^4.3.0",
     "nyc": "^10.1.2",
     "object.entries": "^1.0.4",
+    "postcss-assets": "^4.1.0",
     "postcss-loader": "^1.2.2",
     "ramda": "^0.23.0",
     "raven-js": "^3.8.1",
--- a/static/scss/homepage.scss
+++ b/static/scss/homepage.scss
@@ -32,7 +32,7 @@ body.app-media {
       .brand-logo {
         float: left;
         margin-right: 15px;
-        background: url(/static/images/mit_logo.svg) no-repeat;
+        background: resolve('images/mit_logo.svg') no-repeat;
         height: 36px;
         width: 51px;
         overflow: visible;
@@ -41,7 +41,7 @@ body.app-media {
       }
 
       .brand-logo:hover {
-        background: url(/static/images/mit_logo_on.svg) no-repeat;
+        background: resolve('images/mit_logo_on.svg') no-repeat;
         height: 36px;
         width: 51px;
         overflow: visible;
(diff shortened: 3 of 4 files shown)


postcss.config.js
```diff
@@ -1,7 +1,20 @@
+const autoprefixer = require('autoprefixer');
+const assets = require('postcss-assets');
+
+const CF_DIST = process.env.CLOUDFRONT_DIST;
```
noisecapella: I'm not sure this would work since this retrieves the environment variable at compile time but we may change it without rebuilding the static assets. I believe heroku will just restart Django if we set a new environment variable. @blarghmatey @shaidar can you confirm?

singingwolfboy: That's a good thought that I hadn't considered, but even if you're correct, this pull request is still a step in the right direction. Maybe we can file a follow-up issue for this, so that it doesn't block this PR. (I'd still be curious to hear from @blarghmatey or @shaidar, though.)

noisecapella: If we can't use environment variables at build time then we can't use postcss to solve the problem, so I'm not sure we should merge this at all. Does postcss have a way to update configuration at runtime?

blarghmatey: That is correct, but to @singingwolfboy's point, I'm not sure of an alternative approach that would allow for this to work. FWIW it is unlikely that we would create a new distribution in between deployments (or at all) on a given environment. The alternative would be to rely on posting these assets to S3 and relying on the S3->Cloudfront setup that I'm working in in #1922

noisecapella: We are likely to turn CloudFront on and off again however. Why not just move the relevant CSS to a Django template?

aliceriot: I think that would be a good idea. URLs in CSS is kind of messy in general IMO.

I think in general we should move sometime soon towards extracting our CSS into separate files, instead of leaning on the Webpack loader, in production (using the extract text plugin or similar). Moving these URLs out of our `scss` code would be a step in that direction.

rhysyngsun: Throwing my 2 cents in here: could always whitelabel CF with a domain we control rather that using the direct CF hostname, that way if the delivery mechanism changes we can transparently change it at the DNS level or take the time to recompile this.

pdpinch:
Looks like this needs to be rebased

singingwolfboy:
Closing in favor of #2734.

Issue #2530, opened 2017-02-08
'Number of courses complete' for staff view of learner info


aliceriot:
part of #2480

Issue #2456, opened 2017-01-30
Refactor profile code


aliceriot:
Given the email features we're adding soon it would be good to refactor `ProfileFormContainer` and `ProfileFormFields` as HOC's ([higher-order components](https://facebook.github.io/react/docs/higher-order-components.html)), composable functions which wrap components with extra functionality. This should let us reduce some code-reuse and more easily compose functionality on the same component.

aliceriot:
(not a high priority)

aliceriot:
#2106 is related to this

pdpinch:
Did the HOC approach to the email component work as expected? It seemed like it caused some confusion for Amir.

aliceriot:
It works very well and pretty massively reduces the amount of code duplication. It's just a matter of learning the (pretty simple) pattern and then I think it's clear.

aliceriot:
this work is already done for `ProfileFormContainer` btw.

PR #2708, opened 2017-02-27
Refactor ProfileFormContainer into a HOC

aliceriot:

part of #2456

came up as part of #2530

#### What's this PR do?

This refactors `ProfileFormContainer` to be a HOC instead of a class (which we sub-class elsewhere). I decided to go ahead and do this because, as I was working on #2530, I noticed that the `ProfileFormContainer` subclassing made the sub-classes very hard to change and hard to reason about. We had some very confusing indirection, including a method on the parent class which injected props relevant to the profile into the children of the subclass...yeah.

Anyway, this should make things much easier to read and should also clear the way for making the changes I'll need to make for #2530.

We just need to make sure that there are no regressions relating to profile code. I think it would be good to go through the profile flow and the `/learner` page, trying to edit things and so on. Make sure that validation, redirect behavior, etc is all as it should be.

### DIFF
--- a/static/js/containers/LearnerPage.js
+++ b/static/js/containers/LearnerPage.js
@@ -3,21 +3,27 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import Loader from '../components/Loader';
+import R from 'ramda';
 
 import { FETCH_PROCESSING } from '../actions';
 import { clearProfile } from '../actions/profile';
-import ProfileFormContainer from './ProfileFormContainer';
+import {
+  profileFormContainer,
+  mapStateToProfileProps,
+  childrenWithProps,
+} from './ProfileFormContainer';
 import ErrorMessage from '../components/ErrorMessage';
+import type { ProfileContainerProps } from './ProfileFormContainer';
 
-class LearnerPage extends ProfileFormContainer {
+class LearnerPage extends React.Component<*, ProfileContainerProps, *> {
   componentDidMount() {
-    const { params: { username } } = this.props;
-    this.fetchProfile(username);
+    const { params: { username }, fetchProfile } = this.props;
+    fetchProfile(username);
   }
 
   componentDidUpdate() {
-    const { params: { username } } = this.props;
-    this.fetchProfile(username);
+    const { params: { username }, fetchProfile } = this.props;
+    fetchProfile(username);
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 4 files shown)


static/js/containers/SettingsPage.js
```diff
@@ -3,15 +3,16 @@
...
 import { startProfileEdit } from '../actions/profile';
 import { FETCH_PROCESSING } from '../actions/index';
-import ProfileFormContainer from './ProfileFormContainer';
+import { profileFormContainer, mapStateToProfileProps } from './ProfileFormContainer';
 import PrivacyForm from '../components/PrivacyForm';
 import ProfileProgressControls from '../components/ProfileProgressControls';
 import { privacyValidation } from '../lib/validation/profile';
 
-class SettingsPage extends ProfileFormContainer {
+class SettingsPage extends React.Component {
```
noisecapella: Can you add a props declaration here since `ProfileFormContainer` is no longer doing it?

static/js/containers/ProfilePage.js
```diff
@@ -7,32 +7,38 @@ import R from 'ramda';
...
+import { profileFormContainer, mapStateToProfileProps } from './ProfileFormContainer';
 import { validateProfileComplete } from '../lib/validation/profile';
 import type { Profile, ProfileGetResult } from '../flow/profileTypes';
 import {
   makeProfileProgressDisplay,
   currentOrFirstIncompleteStep,
 } from '../util/util';
 
-class ProfilePage extends ProfileFormContainer {
+class ProfilePage extends React.Component {
```
noisecapella: Can you add a `props` declaration here too?

static/js/containers/LearnerPage.js
```diff
@@ -3,21 +3,22 @@
...
 
 import { FETCH_PROCESSING } from '../actions';
 import { clearProfile } from '../actions/profile';
-import ProfileFormContainer from './ProfileFormContainer';
+import { profileFormContainer, mapStateToProfileProps } from './ProfileFormContainer';
 import ErrorMessage from '../components/ErrorMessage';
 
-class LearnerPage extends ProfileFormContainer {
+class LearnerPage extends React.Component {
   componentDidMount() {
```
noisecapella: Can you define `props` here too?

aliceriot:
@noisecapella I updated it to add Flow annotations to those containers.

PR #2723, opened 2017-02-28
Fixed ExamProfile lookup query (#2716)

### LINKED ISSUE #2716: Exams: profile result processing should look up user by student_id
We sync the `Profile.student_id` so that's what we get back in results. In the code this was indicated to be the same as `Profile.id`, but in actual production data this differs for early users and can't be depended on anyway since it's a different column.

rhysyngsun:
Fixes #2716 

#### What's this PR do?
We send data to Pearson using the `Profile.student_id`, but we were looking it up by `Profile.id`. Locally this was fine, but in production some early users have those values differ.

Set `FEATURE_EXAM_SYNC=True` in `.env` and run this in a shell:

```python
from exams import tasks

user.profile.student_id = 1000
user.profile.save()

# fudge the records to sync

tasks.export_exam_profiles()
```

Then run:

`docker-compose run web ./manage.py simulate_sftp_responses --poll`

And in the shell again:

```python
tasks.batch_process_pearson_zip_files()
ep.refresh_from_db()
print(ep.status)
```

Should print the success status.

### DIFF
--- a/exams/pearson/download_test.py
+++ b/exams/pearson/download_test.py
@@ -278,7 +278,9 @@ class VCDCDownloadTest(MockedESTestCase):
         cls.now = datetime.now(pytz.utc)
         cls.processor = download.ArchivedResponseProcessor(sftp)
         with mute_signals(post_save):
-            cls.success_profiles = ExamProfileFactory.create_batch(2)
+            cls.success_profiles = ExamProfileFactory.create_batch(2) + [
+                ExamProfileFactory.create(profile__id=999, profile__student_id=1000),  # disjoint id and student_id
+            ]
             cls.failure_profiles = ExamProfileFactory.create_batch(2)
 
         cls.success_results = [VCDCResult(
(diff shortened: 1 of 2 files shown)


exams/pearson/download_test.py
```diff
@@ -278,7 +278,9 @@ def setUpTestData(cls):
         cls.now = datetime.now(pytz.utc)
         cls.processor = download.ArchivedResponseProcessor(sftp)
         with mute_signals(post_save):
-            cls.success_profiles = ExamProfileFactory.create_batch(2)
+            cls.success_profiles = ExamProfileFactory.create_batch(2) + [
+                ExamProfileFactory.create(profile__student_id=1000),  # disjoint id and student_id
```
gsidebo: kind of a nitpick, but shouldn't you set `profile__id` _and_ `profile__student_id` if you want them to be explicitly disjoint?

PR #2724, opened 2017-02-28
Use searchkit from props instead of storing it in redux

### LINKED ISSUE #2527: Come up with better solution for the searchkit/search result email bug
Our fix for the critical search result email bug (#2518) is functional, but there are a couple issues with the fix: (1) it puts a complex object into the Redux store, [which is strongly advised against](http://redux.js.org/docs/faq/OrganizingState.html#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state), and (2) we still don't have complete confidence that our new method of getting the 'active' query from searchkit is correct in all cases. To be more specific regarding (2), the new method we're using does not work as expected when it is run upon the opening of the dialog. It _does_ work when the send button is clicked. 

Essentially we are still open to future bugs until we understand more about the `searchkit` object and make use of the Redux store in a more idiomatic way

noisecapella:
Fixes #2527 

#### What's this PR do?
This passes in searchkit as a prop in the email state so that we don't have to pass it to redux for the dialog to see it. This also refactors the searchkit manager code to live in a higher order class.

Go to the learners page and send an email. Everything should work as it did before.

### DIFF
--- a/static/js/components/email/hoc.js
+++ b/static/js/components/email/hoc.js
@@ -1,3 +1,4 @@
+// @flow
 import React from 'react';
 import R from 'ramda';
 
--- a/static/js/components/search/WithSearchkitManager.js
+++ b/static/js/components/search/WithSearchkitManager.js
@@ -1,0 +1,37 @@
+// @flow
+/* global SETTINGS: false */
+import React from 'react';
+import { SearchkitManager, SearchkitProvider } from 'searchkit';
+
+import { getDisplayName } from '../../util/util';
+import { getCookie } from '../../lib/api';
+
+const withSearchkitManager = (WrappedComponent: ReactClass<*>) => {
+  class WithSearchkitManager extends React.Component {
+    searchkit: SearchkitManager;
+
+    constructor(props: Object) {
+      super(props);
+
+      this.searchkit = new SearchkitManager(SETTINGS.search_url, {
+        httpHeaders: {
+          'X-CSRFToken': getCookie('csrftoken')
+        }
+      });
+    }
+
+    render() {
+      return <SearchkitProvider searchkit={this.searchkit}>
+        <WrappedComponent
+          {...this.props}
+          searchkit={this.searchkit}
+        />
+      </SearchkitProvider>;
+    }
+  }
+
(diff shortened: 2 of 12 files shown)


aliceriot:
functionality looks good! reading code.

static/js/components/search/WithSearchkitManager.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+/* global SETTINGS: false */
+import React from 'react';
+import R from 'ramda';
+import { SearchkitManager, SearchkitProvider } from 'searchkit';
+
+import { getDisplayName } from '../../util/util';
+import { getCookie } from '../../lib/api';
+
+const withSearchkitManager = R.curry(
```
aliceriot: Do you need `R.curry` here? It won't do anything for a unary function.

static/js/components/email/hoc.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
aliceriot: 💯

static/js/components/search/WithSearchkitManager.js
noisecapella: No, I just copied and pasted that part. I'll remove it

static/js/components/search/WithSearchkitManager.js
```diff
@@ -1,40 +1,37 @@
...
 import { SearchkitManager, SearchkitProvider } from 'searchkit';
 
 import { getDisplayName } from '../../util/util';
 import { getCookie } from '../../lib/api';
 
-const withSearchkitManager = R.curry(
-  WrappedComponent => {
-    class WithSearchkitManager extends React.Component {
-      searchkit: SearchkitManager;
+const withSearchkitManager = () => WrappedComponent => {
```
aliceriot: why is the `withSearchkitManager` function returning a function?

noisecapella: For consistency with `withEmailDialog`, but I agree it doesn't make much sense. I'll remove it

noisecapella:
Comments addressed, and I added some integration tests for the email dialogs

aliceriot:
cool, taking a look!

aliceriot:
still 👍  from me!

Issue #2705, opened 2017-02-27
When a new user is created, send an email using the stored queries in Elasticsearch


noisecapella:
Part of #2510 

- We should add a new model to keep track of users we have already sent emails to, for a given email campaign. 
- We need to handle the signal to trigger the sending of the email. @pdpinch Should wait until they fill out the profile?

PR #2727, opened 2017-02-28
Add UI to send automatic emails for learner search

### LINKED ISSUE #2707: Add automatic email checkbox
Part of #2510 

Add a checkbox which will cause the email data to be stored to be sent later for every new user

noisecapella:
Fixes #2707 

#### What's this PR do?
Adds the checkbox for sending emails automatically for the learner search page

On the learner search page click the checkbox and send an email. Look in your network tab and verify that `send_automatic_emails: true` was sent. Unclick the checkbox, send another email and verify that `send_automatic_emails: false` was sent.

Also open the other two email dialogs and verify that there is no checkbox at all for these two. One dialog is activated by clicking the 'send a message' link under the user chip, and the other one is the 'contact course team' link on the dashboard.

### DIFF
--- a/static/js/components/email/EmailCompositionDialog.js
+++ b/static/js/components/email/EmailCompositionDialog.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
 import Dialog from 'material-ui/Dialog';
+import Checkbox from 'material-ui/Checkbox';
 
 import { FETCH_PROCESSING } from '../../actions';
 import { dialogActions } from '../inputs/util';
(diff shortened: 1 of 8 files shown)


aliceriot:
functionality looks good! looking at code

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -46,13 +47,29 @@ export default class EmailCompositionDialog extends React.Component {
...
+        value={true}
+        onCheck={e => {
+          updateEmailFieldEdit('sendAutomaticEmails', {
+            target: {
+              value: e.target.checked
+            }
+          });
+        }}
+      />;
+    }
```
aliceriot: could we move this into a helper method? then in the JSX below we could do

```js
{ supportsAutomaticEmails ? this.automaticCheckbox() : null }
```

or something similar.

static/js/flow/emailTypes.js
```diff
@@ -11,12 +11,13 @@ export type EmailInputs = {
...
-  validationErrors: EmailValidationErrors,
-  sendError:        EmailSendError,
-  fetchStatus?:     ?string,
+  inputs:                  EmailInputs,
+  subheading:              ?string,
+  params:                  Object,
+  validationErrors:        EmailValidationErrors,
+  sendError:               EmailSendError,
+  fetchStatus?:            ?string,
+  supportsAutomaticEmails?: boolean,
```
aliceriot: can you fix the alignment of these props? you can use the `column` utility to automatically align them (`column -t` does a table layout).

aliceriot:
I think that the font weight on the text is a little heavy relative to the rest of the dialog:

too_bold

wondering what @roberthouse54 thinks

roberthouse54:
Looks OK to me. It should be font weight 400, default text color.

noisecapella:
Comments addressed. After #2705 is merged we should do another functional test, just to make sure the front and back pieces work together

aliceriot:
Ok! lmk when you want me to take a look again.

PR #2728, opened 2017-02-28
Update mail API to support automatic emails

### LINKED ISSUE #2704: Create API for automatic email queries, and a new model for email data
Part of #2510 

We need a REST API to accept email data and a percolate query, and a model to store the email data in. The query will be created in `PercolateQuery` and there will be a foreign key from the mail data model

noisecapella:
Fixes #2704 

#### What's this PR do?
If the automatic email checkbox is checked and the value is sent to the mail REST API, it will store the email data in the database to make it available to send repeatedly in the future.

You could update the mail API function in `lib/api.js` to send `send_automatic_emails: true` along with the rest of the message, then send an email, then look at the `AutomaticEmail` class to assert that the message content was saved.


rhysyngsun:
looks good

PR #2734, opened 2017-03-01
Load CSS URLs through Django template

### LINKED ISSUE #2462: Not all static assets are served from Cloudfront

singingwolfboy:
Fixes #2462

#### What's this PR do?
Moves our URL references out of the Sass files and into a file that is processed through the Django Templates system. This allows us to dynamically rewrite this file to point to whatever CDN we're using, if any.

Development: run the project locally, and ensure that the page still loads normally. All assets should continue to have local references, rather than pointing to CloudFront.

Production: set `CLOUDFRONT_DIST=d3o95baofem9lo` in your `.env` file and run the project. The page should appear the same as in development, but if you inspect the page, background images should be pointing to CloudFront. In particular, check the MIT logo at the top left of the page, the hero image on the home page, and the search spyglass icon to the left of the search input box.

This pull request was made in response to #2703, where we determined that we can't load the CloudFront URL during build time, only at run time.

### DIFF
--- a/ui/templates/background-images.css
+++ b/ui/templates/background-images.css
@@ -1,0 +1,22 @@
+{% load static %}
+.banner-wrapper {
+  background-image: url({% static "images/lp_hero.jpg" %});
+}
+.navbar .navbar-header .brand-logo {
+  background-image: url({% static "images/mit_logo.svg" %});
+}
+.navbar .navbar-header .brand-logo:hover {
+  background-image: url({% static "images/mit_logo_on.svg" %});
+}
+.navbar .navbar-header .micromasters-logo {
+  background-image: url({% static "images/micromasters_logo.svg" %});
+}
+.navbar .navbar-header .micromasters-logo:hover {
+  background-image: url({% static "images/micromasters_logo_on.svg" %});
+}
+.how-it-works-section .reasons-images-wrap::after {
+  content: url({% static "images/hiw-arrow2.png" %});
+}
+.sk-search-box form .sk-search-box__icon:before {
+  background-image: url({% static "images/search-icon.svg" %});
+}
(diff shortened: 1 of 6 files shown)


ui/templates/background-images.css
```diff
@@ -0,0 +1,22 @@
...
+  background-image: url({% static "images/micromasters_logo.svg" %});
+}
+.navbar .navbar-header .micromasters-logo:hover {
+  background-image: url({% static "images/micromasters_logo_on.svg" %});
+}
+.how-it-works-section .reasons-images-wrap::after {
+  content: url({% static "images/hiw-arrow2.png" %});
+}
+.sk-search-box form .sk-search-box__icon:before {
+  background-image: url({% static "images/search-icon.svg" %});
```
noisecapella: I'm unable to see this on the learner search page. Maybe the other background CSS rule is interfering here

singingwolfboy: Thanks for catching that! It should be fixed now.

ui/templates/base.html
```diff
@@ -7,6 +7,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,400i,500,700" />
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
+    <link rel="stylesheet" href="{% url 'background-images-css' %}" />
```
singingwolfboy: It's optional, but couldn't hurt. I've added it.

PR #2738, opened 2017-03-01
Fix exam card name display

### LINKED ISSUE #2735: Exam card: use first name and last name, not preferred name
When the user's name and address is displayed on the exam card, use their first and last name (romanized, when it exists), not their preferred name. 

The name is displayed here so the user can confirm their "official" name and address as it is on their ID card. We shouldn't be using preferred or nicknames here. 

screenshot 2017-03-01 13 52 00

aliceriot:

closes #2735

#### What's this PR do?

We don't want the preferred name in this situation, we want the full name.

Oh gosh. Get the final exam card to appear, and then confirm that you can see your first and last names, and not your preferred name.

You can use the `mock_api.patch` here: https://github.com/aliceriot/mm_patches

This will set things up so that the 'Last Program' program is set up for showing the exam card.

Then you also need to enable the exam feature flag. You can do this by setting this environment variable:

```
MIDDLEWARE_FEATURE_FLAG_QS_PREFIX=MM
```

and then visiting `/dashboard/?MM_FEATURE_EXAMS=1`

that should do it?

### DIFF
--- a/static/js/components/dashboard/FinalExamCard.js
+++ b/static/js/components/dashboard/FinalExamCard.js
@@ -18,7 +18,7 @@ import {
   PEARSON_PROFILE_SCHEDULABLE
 } from '../../constants';
 import { FETCH_PROCESSING } from '../../actions';
-import { getPreferredName, getLocation } from '../../util/util';
+import { getFullName, getLocation } from '../../util/util';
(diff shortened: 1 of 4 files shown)


aliceriot:
haha, indeed! idk what's going on with it.

Issue #2736, opened 2017-03-01
Dashboard API: handle multiple past run in "can-upgrade" status


giocalitri:
Currently only the latest taken course run can be in `can-upgrade` status, while previous ones can be only in `CHECK_IF_PASSED`.
We need to handle the previous case.

giocalitri:
closing because the remaining part will be handled by #2755

PR #2751, opened 2017-03-02
Added handling of can-upgrade status for past courses

giocalitri:
part of #2736 

#### What's this PR do?
Fixes 2 problems:

1. If an user has a frozen final grade for a course run BUT she did not pay for the course and the upgrade deadline is still in the future, the course appears as upgradable in the dashboard but the final grade is returned as well.
2. If the above mentioned course run (with exactly the same characteristics) is an old one and the user is currently enrolled in a newer run, the old run is returned with status `can-upgrade` as an old run.
**NOTE: this currently does not work on the dashboard UI but only in the dashboard REST API**

1. for the first part:
In a non financial aid program (for simplicity) take a course with multiple runs, then take the oldest of the runs. Set the upgrade deadline in the future. Then, using the `grade.factories.FinalGradeFactory` create  a final grade for your user in the course run, paying attention to set `passed=True`, `status='complete'` and  `course_run_paid_on_edx=False`.
Now you should see on the dashboard the Pay button together with the final grade number.
2. for the second part
Given the previous setup, take the newest course run in the same course as point 1 and create an verified cached enrollment for your user in it. Create also a cached current grade for the same course run. Remember to set the `dashboard.UserCacheRefreshTime` for your user in the future for `enrollment` and `current_grade` so the cache will not be updated.
If now you go to `/api/v0/dashboard/<your_username>/` you should see 2 course runs for the course above picked: the first should have status `currently-enrolled` and the second `can-upgrade`. The second should also have a final grade (it should have the same data of the first part).

`dashboard/api.py`

The change to the `FormatRunTest ` tests is quite extensive only because I split one single huge test in multiple ones: Feel free to review only `test_format_run_with_can_upgrade_no_frozen_grade` and `test_format_run_with_can_upgrade_and_frozen_grade`

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -278,6 +278,8 @@ def get_info_for_course(course, mmtrack):
             else:
                 # any other status means that the student never passed the course run
                 _add_run(run_status.course_run, mmtrack, CourseStatus.NOT_PASSED)
+        elif run_status.status == CourseRunStatus.CAN_UPGRADE:
+            _add_run(run_status.course_run, mmtrack, CourseStatus.CAN_UPGRADE)
 
     return course_data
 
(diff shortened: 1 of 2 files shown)


rhysyngsun:

PR #2754, opened 2017-03-02
Removed error message if there are no enrollments

### LINKED ISSUE #2027: Error message shown occasionally during page load for /dashboard and /learners
I think I filed an issue for this before and fixed it, maybe #1717? Anyway I'm still seeing this behavior. I'm not able to reproduce it consistently

noisecapella:
Fixes #2027 

#### What's this PR do?
Removes the error messages if there is no currently selected enrollment. This can happen:
 - if there are no programs the user is enrolled in
 - when the user first loads the app, before we fetch from the programs API
 - if we fail to fetch programs

The first case should not happen because the user must enroll in a program to complete their profile. The second case happens from time to time but is not an error, so we shouldn't show anything special in this case. For the third case we should show an error message. This PR fixes handling of the second case and leaves the handling of the third case as is.

Go to the dashboard. Clear `localStorage.redux` and refresh the page. You should not see any temporary error message. Repeat for `/learners`

### DIFF
--- a/static/js/components/ErrorMessage_test.js
+++ b/static/js/components/ErrorMessage_test.js
@@ -144,21 +144,21 @@ describe("ErrorMessage", () => {
         });
       });
 
-      it('shows an error if there are no programs', () => {
+      it('shows nothing if there are no programs', () => {
         helper.dashboardStub.returns(Promise.resolve([]));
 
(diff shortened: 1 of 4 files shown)


aliceriot:
this indeed fixes the error!

PR #2759, opened 2017-03-02
Modified management commands for grades

giocalitri:
N/A

#### What's this PR do?
Improves the management commands (and adds another one) to better deal with grades freezing.

1. for the `freeze_final_grades` try to freeze a course run with no "freeze grade": you should see a message saying that such date is missing
2. for the `freeze_final_grades` try to freeze a course run with "freeze grade" in the future: you should see a message saying that the grades cannot be frozen
3. for the `freeze_final_grades` try to freeze a course run that has already a "complete" entry in `CourseRunGradingStatus`: you should see again a message saying that the course is already complete.
4. for the `complete_corse_run_freeze`, try to run it for a course with no entry in `CourseRunGradingStatus`: you should see the course set to complete there (you can use the admin view)
5. for the `complete_corse_run_freeze`, try to run it for a course with an entry in `CourseRunGradingStatus` set to pending: you should see the course set to complete there (you can use the admin view)

### DIFF
--- a/grades/management/commands/freeze_final_grades.py
+++ b/grades/management/commands/freeze_final_grades.py
@@ -1,9 +1,11 @@
 """
 Freezes final grades for a course
 """
+from django.core.exceptions import ImproperlyConfigured
 from django.core.management import BaseCommand, CommandError
 
 from courses.models import CourseRun
+from grades.models import CourseRunGradingStatus
 from grades.tasks import freeze_course_run_final_grades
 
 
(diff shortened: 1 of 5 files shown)


grades/management/commands/complete_corse_run_freeze.py
```diff
@@ -0,0 +1,61 @@
+"""
```
gsidebo: typo in the file name (should be `complete_course_run_freeze.py`)

grades/management/commands/freeze_final_grades.py
```diff
@@ -22,6 +24,20 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+            raise CommandError('Course Run for course_id "{}" is missing the freeze date'.format(course_id))
+        if not can_freeze:
+            raise CommandError('Course Run for course_id "{}" cannot be frozen yet'.format(course_id))
+        if CourseRunGradingStatus.is_complete(run):
+            self.stdout.write(
+                self.style.SUCCESS(
+                    'Final grades for course "{0}" are already complete'.format(course_id)
+                )
+            )
+            return
```
gsidebo: i think it makes more sense to get rid of this `return` and just put the code block below into an `else` block, but that's a stylistic preference

giocalitri:
I changed the variable name and addressed the other comment (but not the `return`.
I also fixed the issue you saw with the `None` value.

giocalitri:
I added a check, but slightly different from what you proposed.

PR #2760, opened 2017-03-02
Fixed dashboard UI to correctly display upgradable past course runs

### LINKED ISSUE #2755: Fix dashboard UI to correctly display upgradable past course runs
It is possible for a user to be in a state where there is a currently enrolled/enrollable course run for a particular course _and_ a past course run that (a) has a final grade, (b) is past its `end_date`, and (c) has a future `upgrade_deadline`. In this case, the dashboard UI looks like this: 

ss 2017-03-02 at 13 48 41

We need to display some more meaningful information for these past-but-graded-and-upgradable course runs. The key file that needs to be changed is `static/js/components/dashboard/CourseSubRow.js`.

gsidebo:
Fixes #2755 

#### What's this PR do?
Fixes the way that course runs are displayed when they are (a) past their end date, (b) passed, and (c) still upgradable (ie: their `upgrade_deadline` is in the future). Users will be able to pay for these courses from the dashboard UI.

For a course with multiple course runs:

1. Set a course run to current and enrolled or enrollable
1. Set another course run to have an `end_date` in the past and a future `upgrade_deadline`
1. Set a `FinalGrade` with a status of 'completed' for the course run in no. 2
1. Load the dashboard and inspect the UI. Click the pay button to make sure it works

CourseSubRow.js

@roberthouse54 spotted some layout tweaks that he wanted when I showed him the dashboard changes, so this PR includes those as well

ss 2017-03-02 at 17 00 58

### DIFF
--- a/static/js/components/OrderSummary.js
+++ b/static/js/components/OrderSummary.js
@@ -75,7 +75,7 @@ class OrderSummary extends React.Component {
     return (
       <div>
         <Card shadow={1}>
-          <p className="intro-text">You are about to enroll in <b>{ course.title }</b></p>
+          <p className="intro-text">You are about to pay for <b>{ course.title }</b></p>
           <div className="wrapper-box">
             <Grid className="summary-box">
               {this.showAmount('Cost of course', this.getCoursePrice())}
--- a/static/js/components/dashboard/CourseAction.js
+++ b/static/js/components/dashboard/CourseAction.js
@@ -116,6 +116,8 @@ export default class CourseAction extends React.Component {
 
   renderTextDescription = this.renderDescription('description', null);
 
+  renderActionBtnDescription = this.renderDescription('course-action-btn-footer', null);
+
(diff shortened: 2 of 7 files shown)


gsidebo:
@pdpinch @roberthouse54 here's what a user sees if they click the "Pay Now" button for one of these past course runs:

ss 2017-03-02 at 17 52 47

The language 'You are about to enroll' doesn't fit here. The user has a final grade, so obviously they enrolled in the past. Would it make sense to change the language to 'You are about to pay for'? This would apply to all cases where a user gets to this page. Not sure how important it is to let the user know explicitly that they are enrolling and paying vs. just paying.

pdpinch:
Go ahead and make that change @gsidebo. For these DEDP courses, the association of payment and enrollment isn't that tight anyway (most users enroll far ahead of paying)

giocalitri:
works as expected
👍 after tests pass

PR #2767, opened 2017-03-03
Configured JS test script to allow for specific test cases to be run

gsidebo:
No ticket

#### What's this PR do?
Changes the `./js_test.sh` script to let us run test cases that match specific test descriptions (ie: the `it`/`describe` strings). If you only want to run one or several tests in a JS test file, this allows you to do that easily

I used this command to test it: `./js_test.sh static/js/components/dashboard/CourseAction_test.js 'shows a message'`. You can test a few other files/descriptions

### DIFF
--- a/js_test.sh
+++ b/js_test.sh
@@ -3,17 +3,33 @@ export TMP_FILE=$(mktemp)
 
 if [[ ! -z "$COVERAGE" ]]
 then
-    export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=html mocha "
+    export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=html mocha"
 elif [[ ! -z "$CODECOV" ]]
 then
-    export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=lcovonly -R spec mocha "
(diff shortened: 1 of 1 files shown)


aliceriot:
why not just `it.only` ?

gsidebo:
@aliceriot you mean using `it.only` instead of this script to run a specific test case? a few reasons. 

1. with `it.only`, you're editing a version controlled file with a change that you do not want to be pushed, and it would be easy to miss. bottom line, i just don't want to have to edit an actual test file to tell our test runner to run a specific test (or group of tests)
1. you can use that description to match multiple tests (eg: every test in some `describe` block)

aliceriot:
fair enough! I just wanted to make sure you knew about it :)

it can also be used on a `describe` block fyi, and we have an eslint rule checking for `.only`.

gsidebo:
@noisecapella can you make sense of the codecov report here? the only change in this PR is to a .sh file

aliceriot:
builds are failing but not on JavaScript so I guess that's ok? yikes.

this works as advertised, including in Docker (i.e. on a linux dev setup), and it doesn't seem to interfere with any of the ways we already run the JavaScript tests.

PR #2770, opened 2017-03-03
Fix postal code validation

### LINKED ISSUE #2765: Exams: additional postal code validation
We need to add the following validation to postal code in the UI:

- If country is Canada, postal code should be a 6-character alphanumeric string (e.g. `A6B9U4`)
- If country is US, postal code can be 5 digits (`02145`) or 9 digits separated by a hyphen (e.g. `02145-4587`)

aliceriot:

fixes #2765 

#### What's this PR do?

This ensures that, for CA and US, the postal code entered is valid.

Go to `/profile` or `/learner`. Make your country the US. Add an invalid postal code. You should see an error message. Fix it, it should go away. Same for Canada.

US codes look like `XXXXX(-XXXX)`, where `X` is in `[0-9]` and the group wrapped in `()` is optional.

Canadian codes look like `YYYYYY`, where `Y` is in `[A-Za-z0-9]`.

The validation behavior should be unchanged for other countries.

### DIFF
--- a/static/js/flow/profileTypes.js
+++ b/static/js/flow/profileTypes.js
@@ -96,6 +96,7 @@ export type Profile = {
   romanized_first_name:        ?string,
   romanized_last_name:         ?string,
   student_id:                  number,
+  postal_code:                 string,
 };
 
 export type Profiles = {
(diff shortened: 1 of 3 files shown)


rhysyngsun:
Code looks solid, manually testing now.

PR #2771, opened 2017-03-03
Use the romanized names on the exam card

### LINKED ISSUE #2735: Exam card: use first name and last name, not preferred name

aliceriot:
closes #2735

#### What's this PR do?

This changes things so that the final exam card will show the romanized version of the user's name, if present, and else will show the first and last names.

Set a first and last name on your profile and visit the dashboard. You should see those values in the final exam card.

Then add non-roman characters (e.g. 🎉) to your first and last names, so that you'll have to fill in the romanized fields. Add new romanized versions and visit the dashboard - you should now see these in the final exam card.


gsidebo:
Code looks great. Trying functionality

aliceriot:
Yep, that's unrelated to this change.

PR #2775, opened 2017-03-03
Refactor MailgunClient for better error handling

### LINKED ISSUE #2749: Add flags to mailgun methods to handle failures using exceptions
This issue has two parts:
- Add a flag to raise an exception for non-2xx status codes for responses from Mailgun.
- As @singingwolfboy pointed out in [this PR](https://github.com/mitodl/micromasters/pull/2693), we don't have any tests in the mail views that simulate a failure response from Mailgun. We should do that.

noisecapella:
Fixes #2749 

#### What's this PR do?
 - adds tests for error handling for all MailgunClient functions
 - adds a `raise_for_status` flag for a few of the functions, which defaults to **True**. This will cause an exception to be raised if the Mailgun response is not in the 2xx range.
 - Changes `send_batch` to raise an exception if any of the batch calls raised an exception. There is also `raise_for_status` for this function which defaults to true. The exception is `SendBatchException` which contains a list of the failed email addresses so that we can make use of it later.

Send an email in all the ways that are available in the UI

### DIFF
--- a/mail/api.py
+++ b/mail/api.py
@@ -8,7 +8,9 @@ import requests
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from rest_framework.status import HTTP_200_OK
+from rest_framework import status
+
+from mail.exceptions import MailException
 from mail.models import FinancialAidEmailAudit
 
 
@@ -78,7 +80,9 @@ class MailgunClient:
         return body, recipients
 
     @classmethod
-    def send_bcc(cls, subject, body, recipients, sender_name=None):
+    def send_bcc(  # pylint: disable=too-many-arguments
+            cls, subject, body, recipients, sender_name=None, raise_on_error=True,
+    ):
         """
         Sends a text email to a BCC'ed list of recipients.
 
@@ -87,6 +91,7 @@ class MailgunClient:
             body (str): Text email body
             recipients (list): A list of recipient emails
             sender_name (str): Sender name
+            raise_on_error (bool): If true and response is not a 2xx status code, raise an exception
         Returns:
             requests.Response: HTTP response from Mailgun
         """
@@ -97,7 +102,18 @@ class MailgunClient:
             subject=subject,
             text=body
         )
-        return cls._mailgun_request(requests.post, 'messages', params, sender_name=sender_name)
+        response = cls._mailgun_request(
+            requests.post,
+            'messages',
+            params,
+            sender_name=sender_name,
+        )
+        if raise_on_error and not 200 <= response.status_code < 300:
+            raise MailException("Expected 2xx status code from Mailgun but got {status}: {content}".format(
+                status=response.status_code,
+                content=response.content,
+            ))
+        return response
 
     @classmethod
     def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-arguments
@@ -114,7 +130,11 @@ class MailgunClient:
             chunk_size (int): The maximum amount of emails to be sent at the same time
 
         Returns:
-            list: List of requests.Response HTTP response from Mailgun
+            list:
+                List of (recipients, requests.Response, exception) where
+                recipients is a list of recipient emails,
+                Response is an HTTP response from Mailgun, if there is one,
+                exception is an exception which occurred when attempting to mail, if there is one
         """
         body, recipients = cls._recipient_override(body, recipients)
         responses = []
@@ -130,21 +150,33 @@ class MailgunClient:
             params['recipient-variables'] = json.dumps({email: {} for email in chunk})
             if sender_address:
                 params['from'] = sender_address
-            responses.append(
-                cls._mailgun_request(
+
+            try:
+                response = cls._mailgun_request(
                     requests.post,
                     'messages',
                     params,
-                    sender_name=sender_name
+                    sender_name=sender_name,
                 )
-            )
+                if response.status_code == status.HTTP_401_UNAUTHORIZED:
+                    message = "Mailgun API keys not properly configured."
+                    log.error(message)
+                    raise ImproperlyConfigured(message)
+
+                responses.append(
+                    (chunk, response, None)
+                )
+            except Exception as exception:  # pylint: disable=broad-except
+                responses.append(
+                    (chunk, None, exception)
+                )
             chunk = list(islice(recipients, chunk_size))
 
         return responses
 
     @classmethod
     def send_individual_email(cls, subject, body, recipient,  # pylint: disable=too-many-arguments
-                              sender_address=None, sender_name=None):
+                              sender_address=None, sender_name=None, raise_on_error=True):
         """
         Sends a text email to a single recipient.
 
@@ -154,16 +186,27 @@ class MailgunClient:
             recipient (str): email recipient
             sender_address (str): Sender email address
             sender_name (str): Sender name
+            raise_on_error (bool): If true and a non-zero response was received,
 
         Returns:
             requests.Response: response from Mailgun
         """
         # Since .send_batch() returns a list, we need to return the first in the list
         responses = cls.send_batch(subject, body, [recipient], sender_address=sender_address, sender_name=sender_name)
-        return responses[0]
+        _, response, exception = responses[0]
+        if exception is not None:
+            raise exception
+        if raise_on_error and not 200 <= response.status_code < 300:
+            raise MailException("Expected 2xx status code from Mailgun but got {status}: {content}".format(
+                status=response.status_code,
+                content=response.content,
+            ))
+        return response
 
     @classmethod
-    def send_financial_aid_email(cls, acting_user, financial_aid, subject, body):
+    def send_financial_aid_email(  # pylint: disable=too-many-arguments
+            cls, acting_user, financial_aid, subject, body, raise_on_error=True,
+    ):
         """
         Sends a text email to a single recipient, specifically as part of the financial aid workflow. This bundles
         saving an audit trail for emails sent (to be implemented).
@@ -173,13 +216,14 @@ class MailgunClient:
             financial_aid (FinancialAid): the FinancialAid object this pertains to (recipient is pulled from here)
             subject (str): email subject
             body (str): email body
+            raise_on_error (bool): If true and we received a non 2xx status code from Mailgun, raise an exception
         Returns:
             requests.Response: response from Mailgun
         """
         from_address = cls.default_params()['from']
         to_address = financial_aid.user.email
-        response = cls.send_individual_email(subject, body, to_address)
-        if response.status_code == HTTP_200_OK:
+        response = cls.send_individual_email(subject, body, to_address, raise_on_error=raise_on_error)
+        if response.status_code == status.HTTP_200_OK:
             FinancialAidEmailAudit.objects.create(
                 acting_user=acting_user,
                 financial_aid=financial_aid,
@@ -191,7 +235,9 @@ class MailgunClient:
         return response
 
     @classmethod
-    def send_course_team_email(cls, user, course, subject, body):
+    def send_course_team_email(  # pylint: disable=too-many-arguments
+            cls, user, course, subject, body, raise_on_error=True,
+    ):
         """
        Sends a text email from a user to a course team.
 
@@ -200,6 +246,7 @@ class MailgunClient:
             course (courses.models.Course): A Course
             subject (str): Email subject
             body (str): Email body
+            raise_on_error (bool): If true and we received a non 2xx status code from Mailgun, raise an exception
         Returns:
             requests.Response: HTTP Response from Mailgun
        """
@@ -213,6 +260,7 @@ class MailgunClient:
             body,
             course.contact_email,
             sender_address=user.email,
-            sender_name=user.profile.display_name
+            sender_name=user.profile.display_name,
+            raise_on_error=raise_on_error,
         )
         return response
--- a/mail/api.py
+++ b/mail/api.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from rest_framework import status
 
+from mail.exceptions import SendBatchException
 from mail.models import FinancialAidEmailAudit
 
 
@@ -34,7 +35,9 @@ class MailgunClient:
         return {'from': settings.EMAIL_SUPPORT}
 
     @classmethod
-    def _mailgun_request(cls, request_func, endpoint, params, sender_name=None):
+    def _mailgun_request(  # pylint: disable=too-many-arguments
+            cls, request_func, endpoint, params, sender_name=None, raise_for_status=True
+    ):
         """
         Sends a request to the Mailgun API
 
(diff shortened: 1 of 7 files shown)


singingwolfboy:
The Python `requests` library has a built-in method on the Response class called [`raise_for_status()`](http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status), which will raise an exception if the response status code is not in the 2XX success range. I recommend calling that method, and possibly renaming the `raise_on_error` method parameter to `raise_for_status` as well.

noisecapella:
@singingwolfboy Will do

mail/views.py
```diff
@@ -63,14 +63,42 @@ def post(self, request, *args, **kargs):  # pylint: disable=unused-argument
...
+    """
+    Helper function to figure out a status code to return. In summary, 200 unless any error, then 500.
+    The user can inspect the contents for more info.
+    """
+    for _, response, exception in responses:
+        if exception is not None:
+            return status.HTTP_500_INTERNAL_SERVER_ERROR
+        if response.status_code != status.HTTP_200_OK:
+            return status.HTTP_500_INTERNAL_SERVER_ERROR
+    return status.HTTP_200_OK
```
singingwolfboy: This seems cleaner:

```python
has_error = any(
    exception is not None or
    response.status_code != status.HTTP_500_INTERNAL_SERVER_ERROR
    for _, response, exception in responses
)
if has_error:
    return status.HTTP_500_INTERNAL_SERVER_ERROR
else:
    return status.HTTP_200_OK
```

mail/api.py
```diff
@@ -114,9 +126,14 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
             chunk_size (int): The maximum amount of emails to be sent at the same time
 
         Returns:
-            list: List of requests.Response HTTP response from Mailgun
+            list:
+                List of (recipients, requests.Response, exception) where
+                recipients is a list of recipient emails,
+                Response is an HTTP response from Mailgun, if there is one,
+                exception is an exception which occurred when attempting to mail, if there is one
```
singingwolfboy: Why change the return values here? I think it's fine to have this function continue to return a list of HTTP responses, since any caller that wants to know if there were exceptions can simply iterate over those responses to see if there were any failures. It's no different from checking the return values for exception objects.

mail/api_test.py
```diff
@@ -262,16 +381,38 @@ def test_financial_aid_email(self, mock_post):
...
+
+        response = MailgunClient.send_financial_aid_email(
+            self.staff_user_profile.user,
+            self.financial_aid,
+            'email subject',
+            'email body',
+            raise_for_status=raise_for_status,
+        )
+
+        assert response.raise_for_status.called is raise_for_status
```
singingwolfboy: Another non-meaningful test.

mail/api_test.py
```diff
@@ -347,3 +494,27 @@ def test_send_to_course_team_without_email(self, mock_post):
...
+        course_with_email = CourseFactory.create(title='course with email', contact_email='course@example.com')
+
+        response = MailgunClient.send_course_team_email(
+            self.user,
+            course_with_email,
+            'email subject',
+            'email body',
+            raise_for_status=raise_for_status,
+        )
+        assert response.raise_for_status.called is raise_for_status
```
singingwolfboy: Another non-meaningful test.

mail/api.py
```diff
@@ -130,21 +147,38 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
...
                     requests.post,
                     'messages',
                     params,
-                    sender_name=sender_name
+                    sender_name=sender_name,
+                )
+                if response.status_code == status.HTTP_401_UNAUTHORIZED:
+                    message = "Mailgun API keys not properly configured."
+                    log.error(message)
+                    raise ImproperlyConfigured(message)
```
singingwolfboy: I don't think this works the way you expect. If you raise this exception here, it will be caught by the `except Exception as exception:` line a few lines later. Then the code will stuff this `ImproperlyConfigured` exception into the list of responses, and merrily continue iterating over chunks. It would be better to have the `ImproperlyConfigured` exception actually raise, instead of being caught, so that this code doesn't try to contact Mailgun more than once. If the API keys aren't configured, contacting Mailgun will never succeed!

mail/api.py
```diff
@@ -173,13 +215,14 @@ def send_financial_aid_email(cls, acting_user, financial_aid, subject, body):
...
+            raise_for_status (bool): If true and we received a non 2xx status code from Mailgun, raise an exception
         Returns:
             requests.Response: response from Mailgun
         """
         from_address = cls.default_params()['from']
         to_address = financial_aid.user.email
-        response = cls.send_individual_email(subject, body, to_address)
-        if response.status_code == HTTP_200_OK:
+        response = cls.send_individual_email(subject, body, to_address, raise_for_status=raise_for_status)
+        if response.status_code == status.HTTP_200_OK:
```
singingwolfboy: Why even check the status code here? `raise_for_status()` will ensure that either the response was successful, or an exception will be raised so that this code won't even execute. (Is the caller prepared for the possibility that this function will raise an exception?)

mail/api.py
```diff
@@ -154,16 +188,24 @@ def send_individual_email(cls, subject, body, recipient,  # pylint: disable=too-
...
 
         Returns:
             requests.Response: response from Mailgun
         """
         # Since .send_batch() returns a list, we need to return the first in the list
         responses = cls.send_batch(subject, body, [recipient], sender_address=sender_address, sender_name=sender_name)
-        return responses[0]
+        _, response, exception = responses[0]
+        if exception is not None:
+            raise exception
```
singingwolfboy: I'm confused. So if an exception is raised in `send_batch()`, the exception is caught by the `except Exception` line. But then the caught exception is returned, so that it can be re-raised again here? This is going to make debugging these exceptions a nightmare. It would be better to just allow exceptions to raise, so that the stack trace doesn't get mucked up.

mail/views.py
```diff
@@ -104,20 +132,17 @@ def post(self, request, *args, **kargs):  # pylint: disable=unused-argument
...
         return Response(
-            status=status.HTTP_200_OK,
+            status=_make_batch_status(mailgun_responses),
             data={
-                "batch_{}".format(batch_num): {
-                    "status_code": resp.status_code,
-                    "data": generate_mailgun_response_json(resp)
-                } for batch_num, resp in enumerate(mailgun_responses)
+                "batch_{}".format(batch_num): _make_batch_response_dict(resp, exception)
+                for batch_num, (_, resp, exception) in enumerate(mailgun_responses)
```
singingwolfboy: Why is `data` a dictionary, rather than a list? Especially since the keys of the dictionary are just strings for the form "batch_{num}" -- seems to me like it really wants to be a list. Why not simply return `data=mailgun_responses`, and not process it at all?

mail/views_test.py
```diff
@@ -38,6 +41,24 @@ def json(*args, **kwargs):  # pylint:disable=unused-argument, missing-docstring
...
+    def _is_response(_status_code):
+        """Helper function to figure out whether to generate a response or an exception"""
+        return isinstance(_status_code, int)
+    return [
+        (
+            ['recipient@example.com'],
+            Mock(spec=Response, status_code=status_code, json=mocked_json())
+            if _is_response(status_code) else None,
+            None if _is_response(status_code) else HTTPError(),
+        ) for status_code in status_codes
```
singingwolfboy: There is _way_ too much logic going on here, and I can't follow it. What are you trying to achieve with this function? Can you make the code clearer? Readability counts.

mail/api_test.py
```diff
@@ -26,7 +29,11 @@
 
 
 @ddt
-@patch('requests.post')
+@patch('requests.post', autospec=True, return_value=Mock(
```
noisecapella: Sure, but I'll leave it alone for now since it's out of the scope of this PR

mail/api_test.py
```diff
@@ -101,6 +109,27 @@ def test_send_bcc(self, sender_name, mock_post):
...
+        )
+
+        response = MailgunClient.send_bcc(
+            'email subject',
+            'email body',
+            ['a@example.com', 'b@example.com'],
+            raise_for_status=raise_for_status,
+        )
+
+        assert response.raise_for_status.called is raise_for_status
```
noisecapella: It's testing that `response.raise_for_status` is called when it should be, which in turn raises an exception. I can change the test description but I think this test is meaningful as is. It says that if `send_bcc` is called and it gets a 400 response, it will call `raise_for_status` and pass along the 400 status code.

mail/utils.py
```diff
@@ -79,13 +75,10 @@ def generate_mailgun_response_json(response):
 
     Args:
         response (requests.Response): response object
+        exception (Exception): If not None, there was an exception when trying to contact Mailgun
```
noisecapella: I'll remove it

mail/views.py
noisecapella: I wanted to keep the response data the same as it was before. I don't believe we actually use it right now.

mail/views.py
noisecapella: Sure

mail/api.py
noisecapella: In a future PR I need to know which email addresses were processed correctly. Also, returning just the response won't work because there may be errors when we attempt to contact the mailgun API, which will cause an error before we can get a response for it.

mail/api.py
noisecapella: You're right, I'll fix this and add a test for it

mail/api.py
noisecapella: In Python 3 doesn't `raise exception_variable` raise an exception which keeps the stack trace of the original exception? Maybe I should double check that

noisecapella: Yes, python 2:

    >>> try:
    ...     try:
    ...         1/0
    ...     except Exception as e:
    ...         raise e
    ... except:
    ...     raise
    ... 
      File "<stdin>", line 5, in <module>
    ZeroDivisionError: integer division or modulo by zero
    >>> 

Python 3:

    >>> try:
    ...     try:
    ...         1/0
    ...     except Exception as e:
    ...         raise e
    ... except:
    ...     raise
    ... 
      File "<stdin>", line 5, in <module>
      File "<stdin>", line 3, in <module>
    ZeroDivisionError: division by zero

mail/api.py
noisecapella: `raise_for_status()` is only called if the `raise_for_status` argument is true

mail/views_test.py
noisecapella: The function takes a list of either status codes or exceptions and creates mock data based on that. I'll clean up this function

mail/views.py
singingwolfboy: If we don't actually use it right now, that's all the more reason to change it now, _before_ other code starts depending on it.

mail/api.py
singingwolfboy: There's a difference between `raise exception` and `raise`. You're using the former, which raises a new exception with a new traceback. The latter re-raises an existing exception in a way that preserves its original traceback. You can also use the `from` clause for exception chaining, which may be what you want here.

But even though we _can_ preserve tracebacks for easier debugging, I'm not convinced that this is the right path to take in the first place. It would be better to call `send_batch()` with `raise_for_status=False`, so that exceptions aren't raised at all. Then this code could call `raise_for_status()` on the returned response object, and raise an exception there if necessary. That way, you don't have to bounce exceptions around the way you're doing.

Here's the official documentation for the `raise` statement: https://docs.python.org/3/reference/simple_stmts.html#grammar-token-raise_stmt

mail/views.py
noisecapella: Sure, I'll remove it then

mail/api.py
singingwolfboy: OK, that's true. But you should still check [`response.ok`](http://docs.python-requests.org/en/master/api/#requests.Response.ok) instead of comparing the status code with 200. There are other successful status codes beyond 200.

mail/api.py
noisecapella: There's no `raise_for_status` for `send_batch`. The `exception` here would come from cases where we were unable to contact Mailgun at all so we won't have a response (or maybe `ImproperlyConfigured` for the 401 case)

noisecapella: It seems like `raise exception` will raise that particular exception:

    >>> try:
    ...     raise KeyError()
    ... except KeyError as e:
    ...     raise e
    ... 
      File "<stdin>", line 4, in <module>
      File "<stdin>", line 2, in <module>
    KeyError

I think `raise exception from other` is only applicable if you wanted to raise a different exception and include the other's stack trace

mail/api.py

noisecapella:
I think that's all comments

mail/api.py
singingwolfboy: The Unix philosophy states that no feedback means that everything was successful. As a result, I would suggest that this function have the following return values:

* Every request to Mailgun was successful: return a list of response objects, just as this function used to do. The caller can assume that every email address in the recipients list was processed correctly.
* At least one request to Mailgun was unsuccessful: raise an exception. Before the exception is raised, however, construct the exception object and attach the list of response objects to it, as well as the list of email addresses that were processed correctly. The caller can catch this exception if this information is necessary.

I'm very, very wary about returning error values instead of raising exceptions -- you're expecting the caller to check for the error value, and it's easy to forget to do so. If we change the code to work the way I've described, there's no chance for a failure to pass unnoticed. The caller has to intentionally handle the exception, or else let it bubble up.

noisecapella: That handles the exception case but the caller will still have to do the work of checking the status codes of each response. The way the function is now, the user will need to check the statuses anyway. If they're using `send_batch` and not `send_individual_email` or one of the other related single email methods they should be prepared to deal with some percent of errors in the output anyway.

singingwolfboy: > That handles the exception case but the caller will still have to do the work of checking the status codes of each response.

What do you mean? With the behavior I'm describing, if the function returns without raising an exception, then that means that every request to Mailgun is successful. That means that the function also checks the status code of each response, to determine whether every request was successful.

> The way the function is now, the user will need to check the statuses anyway.

Yes, that's precisely what concerns me.

> If they're using `send_batch` and not `send_individual_email` or one of the other related single email methods they should be prepared to deal with some percent of errors in the output anyway.

I disagree. If the caller is prepared to deal with errors, then the caller will use a `try`/`except` block to catch the exception that may arise. If the caller is _not_ prepared to deal with errors, then the caller will _not_ use a `try`/`except` block, and the exception will bubble up so that a higher-level caller can handle it as necessary.

noisecapella: Ok, I'll raise an exception as you described

noisecapella:
Do you have any other comments beside the `send_batch` exception?

singingwolfboy:
I expect that changing the return values in the way that I've described will have significant effects on other code that calls that function, which will need to be re-reviewed. Besides that, though, the rest seems alright. Could you make an issue to come back to the automated tests, and implement a library like [Betamax](https://github.com/sigmavirus24/betamax) to make them more closely match the way `requests` is used in production?

I'm out of town for the rest of the week, so you may want to get someone else to finish this review.

giocalitri:
I have just one question: what's the use case for having `raise_for_status=False`?

noisecapella:
@giocalitri There isn't one right now. Should I remove it? I added it mostly as a just in case measure

giocalitri:
I think it is fine: 👍

Issue #2120, opened 2016-12-14
Allow users to expand/hide search facets by clicking facet title


gsidebo:
@roberthouse54 mentioned that search facet titles should be clickable to expand/hide the contents of each facet. Right now, only the arrows are clickable

![facet_title](https://cloud.githubusercontent.com/assets/14932219/21192838/e1bbe8e6-c1f7-11e6-91ea-f622cb24c71e.png)

PR #2777, opened 2017-03-06
Allowed users to expand/hide search facets by clicking facet title

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2120

#### What's this PR do?
- Made facets title clickable, user can expand/hide search facets by clicking facet title

- Go to learners page and click the title of any facet.

@pdpinch 
screen shot 2017-03-06 at 6 28 34 pm

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -200,22 +200,24 @@ export default class LearnerSearch extends SearchkitComponent {
       <Card className="fullwidth" shadow={1}>
         <FilterVisibilityToggle
           {...this.props}
+          title="Course"
           filterName="courses"
         >
           <HierarchicalRefinementFilter
             field={"program.enrollments"}
-            title="Course"
+            title=""
             id="courses"
           />
         </FilterVisibilityToggle>
         <FilterVisibilityToggle
           {...this.props}
           filterName="semester"
+          title="Semester"
         >
           <PatchedMenuFilter
             field="program.semester_enrollments.semester"
             fieldOptions={{ type: 'nested', options: {path: 'program.semester_enrollments'} }}
-            title="Semester"
+            title=""
             id="semester"
             bucketsTransform={sortSemesterBuckets}
           />
@@ -223,6 +225,7 @@ export default class LearnerSearch extends SearchkitComponent {
         <FilterVisibilityToggle
           {...this.props}
           filterName="num-courses-passed"
+          title="# of Courses Passed"
         >
           <RangeFilter
             field="program.num_courses_passed"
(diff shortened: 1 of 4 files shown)


aliceriot:
works great! looking at code now.

static/js/components/LearnerSearch.js
```diff
@@ -200,58 +200,63 @@ export default class LearnerSearch extends SearchkitComponent {
       <Card className="fullwidth" shadow={1}>
         <FilterVisibilityToggle
           {...this.props}
+          title="Course"
           filterName="courses"
         >
           <HierarchicalRefinementFilter
             field={"program.enrollments"}
-            title="Course"
+            title=""
```
aliceriot: can we just omit the `title` prop from these filters, or do we have to pass in the empty string?

static/js/components/search/FilterVisibilityToggle.js
```diff
@@ -13,6 +13,7 @@ export const FILTER_ID_ADJUST = {
 
 export default class FilterVisibilityToggle extends SearchkitComponent {
   props: {
+    title:             string,
```
aliceriot: can you fix the alignment for this prop? (with the indentation of the other props, just for readability)

static/js/components/search/FilterVisibilityToggle.js
```diff
@@ -71,6 +72,20 @@ export default class FilterVisibilityToggle extends SearchkitComponent {
...
+    }
+    const { title } = this.props;
+    return (
+      <div
+        className="sk-hierarchical-refinement-list__header"
+        onClick={this.toggleFilterVisibility}>
+        {title}
+      </div>
+    );
+  };
```
aliceriot: why not roll this into the `openStateIcon` method above? these two methods basically do the same thing (check `this.isInResults` and if so return some UI, else null).

You could write a new method like this:

```js
renderFilterTitle = (children) => {
  if (!this.isInResults(children.props.id)) {
    return null;
  }
  const { title } = this.props;
  return <div className="title-row" onClick={this.toggleFilterVisibility}>
    <Icon
      name="arrow_drop_down"
      className={this.openClass()}
    />
    <div className="sk-hierarchical-refinement-list__header">       <div className="sk-hierarchical-refinement-list__header">
      { title } 
    </div>
  </div>;
}
```

something along those lines! then we're just making the check once, only setting one `onClick` handler, etc.
             
what do you think?

static/js/components/LearnerSearch.js
amir-qayyum-khan: they are compulsory by these components

aliceriot: gotcha, that's too bad. ah well.

amir-qayyum-khan:
@aliceriot any update?

noisecapella:
@amir-qayyum-khan She's out today

amir-qayyum-khan:
ok @noisecapella

Issue #2746, opened 2017-03-01
Indexing is too slow


noisecapella:
We need to find a way to speed up indexing:
 - reduce the amount of data indexed
 - only reindex documents that have changed
 - other elasticsearch optimizations that are available
 - other elasticsearch APIs

We should do some discovery work to see whether the slowdown is in Python, in Elasticsearch, or in the communication between the two

noisecapella:
Removing `MMTrack` entirely from indexing results in a 80% speedup of the running time of indexing, due to the n+1 database queries. We can cache some information that's not specific to the user and that results in a 25% speedup, though I'm not sure if it's worth making `MMTrack` more complicated

PR #2782, opened 2017-03-06
Sent emails when new user fills out profile and their profile matches query

### LINKED ISSUE #2705: When a new user is created, send an email using the stored queries in Elasticsearch

noisecapella:
Fixes #2705

#### What's this PR do?
When a new user fills out their profile and one of the percolated queries matches their information, they should get an email based on the contents of the relevant `AutomaticEmail` model. There should also be one `SentAutomaticEmail` created for each user for each `AutomaticEmail` after the email is successfully sent.

Edit `static/js/lib/api.js` to add `send_automatic_emails: true` to the request parameters for the search API. Then go to `/learners` and do a search. Send an email. You should see a new `AutomaticEmail` model and a new `SentAutomaticEmail` entry for each user, for that `AutomaticEmail` model, and you should receive an email with a list of addresses matching the profiles in your search where `email_optin` was true.

Next, choose one of those profiles. Delete the `SentAutomaticEmail` entry. Update its employment information somehow (either log in as the user, or modify its model directly). After the update goes though you should receive a new email and there should be a new entry for `SentAutomaticEmail` replacing the one you deleted earlier.

Next, modify the education for the profile. Make sure that you did not receive an email in response to this change.


search/indexing_api.py
```diff
@@ -109,12 +110,16 @@ def index_program_enrolled_users(program_enrollments, chunk_size=100):
...
     """
-    return _index_chunks(
+    num_indexed = _index_chunks(
         (serialize_program_enrolled_user(program_enrollment) for program_enrollment in program_enrollments),
         USER_DOC_TYPE,
         chunk_size=chunk_size,
     )
 
+    for program_enrollment in program_enrollments:
+        send_automatic_emails(program_enrollment)
```
gsidebo: this seems like a _really_ good candidate for an asynchronous task. we already have an issue open about our poor indexing performance. this adds quite a lot of work to the indexing of every single user - a `percolate` call to ES, a query to `PercolateQuery`, a query to `AutomaticEmail`, potentially a Mailgun call, etc.

mail/api.py
```diff
@@ -280,3 +292,80 @@ def send_course_team_email(  # pylint: disable=too-many-arguments
...
+                )
+                SentAutomaticEmail.objects.create(
+                    user=user,
+                    automatic_email=automatic_email,
+                )
+            except:  # pylint: disable=bare-except
+                log.exception("Error sending mailgun mail for automatic email %s", automatic_email)
+        except IntegrityError:
+            # Probable race condition, skip this user since we already sent them an email
+            pass
```
gsidebo: You can combine these try/catches into one block, right?

mail/views.py
```diff
@@ -108,22 +110,37 @@ def post(self, request, *args, **kargs):  # pylint: disable=unused-argument
...
+            )
+
+            if automatic_email:
+                mark_emails_as_sent(automatic_email, emails)
+        except SendBatchException as send_batch_exception:
+            if automatic_email:
+                failure_emails = set()
+                for recipients, _ in send_batch_exception.exception_pairs:
+                    failure_emails |= set(recipients)
+                success_emails = set(emails).difference(failure_emails)
```
gsidebo: can you change `SendBatchException` so that you just need to refer to `send_batch_exception.failed_recipient_emails` here instead of doing all of this calculation in the `except` block? could be a property or function. this would be a lot more readable

```python
except SendBatchException as send_batch_exception:
    if automatic_email:
        success_emails = set(emails).difference(send_batch_exception.failed_recipient_emails)
        mark_emails_as_sent(automatic_email, success_emails)
```

mail/api.py
```diff
@@ -280,3 +292,80 @@ def send_course_team_email(  # pylint: disable=too-many-arguments
...
+    Send all automatic emails which match the search criteria for a program enrollment
+
+    Args:
+        program_enrollment (ProgramEnrollment): A ProgramEnrollment
+    """
+    percolate_queries = search_percolate_queries(program_enrollment.id)
+    automatic_emails = AutomaticEmail.objects.filter(
+        query__in=percolate_queries,
+        enabled=True,
+    ).exclude(sentautomaticemail__user__programenrollment=program_enrollment)
```
noisecapella: I don't think there's a use case for a user receiving two similar emails from two different programs, though if there was we should change `SentAutomaticEmail` to connect with `ProgramEnrollment` instead. For right now there will not be an `AutomaticEmail` which matches more than one program because the learner search filters on a single program. What do you think @pdpinch?

mail/api.py
noisecapella: Sure, I'll move the log.exception down here

search/indexing_api.py
noisecapella: All indexing is already done in asynchronous tasks. The above `_index_chunks` call can take a while, it's basically what `recreate_index` does

mail/views.py
noisecapella: Sure

search/indexing_api.py
gsidebo: If we still care about indexing performance (and I believe we do https://github.com/mitodl/micromasters/issues/2746), I think we should figure out a different approach for this. I think I'm mostly hung up on the fact that this indexing method (`index_program_enrolled_users`) is invoked from multiple sources, including `recreate_index`. It seems like it's just being added here for convenience. From what I can tell, this works for our purposes because `dashboard.signals.handle_create_programenrollment` calls `index_program_enrolled_users`, so the effect is that a new program enrollment is created for a user, and the email is sent (if appropriate). On the other hand I don't think it makes any sense for us to attempt to send automatic emails as part of a complete recreation of the index, but that's what will happen. 

I think it will make more sense to trigger the automatic email send only when a new program enrollment is created. `dashboard.signals.handle_create_programenrollment` calls `search.tasks.index_program_enrolled_users` right now, but maybe it makes more sense to create a new task called `index_new_program_enrolled_user`, which would call `search.indexing_api.index_program_enrolled_users` _and then_ attempt to send the automatic email via some other function. Does that make sense?

noisecapella: The intention is to have this be flexible enough to trigger on any change which happens in the future where the user becomes part of a search query, not just for new program enrollments. I'll refactor this so that the email send happens in tasks triggered by signals and not during `recreate_index`

mail/api_test.py
```diff
@@ -576,3 +589,146 @@ def test_send_to_course_team_error(self, raise_for_status, mock_post):
...
+            )
+        cls.automatic_email = AutomaticEmail.objects.create(
+            query=cls.percolate_query,
+            enabled=True,
+            email_subject='subject',
+            email_body='body',
+            sender_name='Sender name',
+        )
+        cls.automatic_email_disabled = AutomaticEmail.objects.create(
+            query=None,
```
gsidebo: Did you mean to set `cls.percolate_query_disabled` as the `query` here? If so, you might want to take a second look at the test logic since it's working without the intended conditions

mail/api_test.py
```diff
@@ -576,3 +589,146 @@ def test_send_to_course_team_error(self, raise_for_status, mock_post):
...
+                query="new query",
+                original_query="original query",
+            )
+        new_automatic = AutomaticEmail.objects.create(
+            enabled=True,
+            email_subject='new subject',
+            email_body='new body',
+            sender_name='new Sender name',
+            query=new_percolate,
+        )
```
gsidebo: Once factories are created, please replace these too. Might just be handy to do a full codebase search for `AutomaticEmail.objects.create` and `PercolateQuery.objects.create`

mail/api_test.py
```diff
@@ -576,3 +589,146 @@ def test_send_to_course_team_error(self, raise_for_status, mock_post):
...
+            count += 1
+            if count == 1:
+                raise KeyError()
+
+        with patch('mail.api.search_percolate_queries', autospec=True, return_value=[
+            self.percolate_query, new_automatic.query
+        ]) as mock_search_queries, patch(
+            'mail.api.MailgunClient', send_individual_email=_send_individual_email
+        ) as mock_mailgun:
+            mock_mailgun.send_individual_email = _send_individual_email
```
gsidebo: 2 things:

1. Is it necessary to set `send_individual_email` on the mock object in the `patch()` _and_ in the body of the `with` block?
1. I think you can accomplish this error-first-then-succeed behavior using `side_effect`: https://docs.python.org/3/library/unittest.mock.html#quick-guide. If you give `side_effect` a list, it returns successive elements of that list. You might be able to do something like:

```python
patch(
    'mail.api.MailgunClient', 
    send_individual_email=Mock(side_effect=[Exception(), None])
)
```

mail/views_test.py
```diff
@@ -140,43 +146,106 @@ def test_automatic_email(self):
         email_results = ['a@example.com', 'b@example.com']
         request_data = self.request_data.copy()
         request_data['send_automatic_emails'] = True
-
-        assert AutomaticEmail.objects.count() == 0
-        search_dict = create_search_obj(
+        for email in email_results:
```
gsidebo: Can you split these two automatic email test cases into a separate test class? You're performing the same ~20 lines of data creation for both of the cases

mail/views_test.py
```diff
@@ -140,43 +146,106 @@ def test_automatic_email(self):
...
+        ]
+
+        with patch(
+            'mail.views.get_all_query_matching_emails', autospec=True, return_value=email_results
+        ), patch(
+            'mail.views.MailgunClient'
+        ) as mock_mailgun_client, patch(
+            'mail.views.add_automatic_email', autospec=True, return_value=automatic_email,
+        ) as mock_add_automatic_email:
+            mock_mailgun_client.send_batch.side_effect = SendBatchException(exception_pairs)
```
gsidebo: You're not able to set this side_effect in the actual `patch()` call? Also, what exactly are we testing for with the `assert send_batch_exception` line below?

mail/api.py
```diff
@@ -280,3 +291,76 @@ def send_course_team_email(  # pylint: disable=too-many-arguments
...
+                automatic_email.email_body,
+                user.email,
+                sender_name=automatic_email.sender_name,
+            )
+            SentAutomaticEmail.objects.create(
+                user=user,
+                automatic_email=automatic_email,
+            )
+        except:  # pylint: disable=bare-except
+            log.exception("Error sending mailgun mail for automatic email %s", automatic_email)
```
noisecapella: I'll add an except clause for `IntegrityError` with a different log message

mail/api_test.py
noisecapella: It looks like this is an oversight, I'll add `percolate_query_disabled`. The test logic shouldn't change because the significant piece is `enabled=False` here. Without `enabled=True` the other business logic shouldn't look at the query at all.

mail/api_test.py
```diff
@@ -576,3 +589,146 @@ def test_send_to_course_team_error(self, raise_for_status, mock_post):
...
+            email_body='body',
+            sender_name='Sender name',
+        )
+        cls.automatic_email_disabled = AutomaticEmail.objects.create(
+            query=None,
+            enabled=False,
+            email_subject='other subject',
+            email_body='other body',
+            sender_name='other Sender name',
+        )
```
noisecapella: Sure

mail/api_test.py
noisecapella: Sure

mail/views_test.py
noisecapella: Sure

mail/views_test.py
noisecapella: The `assert send_batch_exception` line is a typo, it should be comparing the `exception_pairs` to make sure they're the same. I'll move the side_effect to the patch()

noisecapella:
Ok, I think that's everything

noisecapella:
Any other comments on this @gsidebo?

gsidebo:
Looking now

mail/factories.py
```diff
@@ -0,0 +1,22 @@
...
+from search.factories import PercolateQueryFactory
+
+
+class AutomaticEmailFactory(DjangoModelFactory):
+    """Factory for AutomaticEmail"""
+    query = SubFactory(PercolateQueryFactory)
+    enabled = Faker('boolean')
+    email_subject = FuzzyText()
+    email_body = FuzzyText()
+    sender_name = FuzzyText()
```
gsidebo: There's some faker class/method for generating names. If it's easy enough to find that and implement here, go for it. Not very important, obviously

mail/models.py
```diff
@@ -34,3 +34,20 @@ class AutomaticEmail(TimestampedModel):
...
         """String representation of AutomaticEmail"""
         return "AutomaticEmail sender={}, subject={}".format(self.sender_name, self.email_subject)
+
+
+class SentAutomaticEmail(TimestampedModel):
+    """
+    Keeps track of automatic emails which were sent to particular users
+    """
+    user = models.ForeignKey(User, null=False, on_delete=models.CASCADE)
+    automatic_email = models.ForeignKey(AutomaticEmail, null=False, on_delete=models.CASCADE)
```
gsidebo: Just want to make sure we really want these `on_delete=models.CASCADE` properties. This just means that when a `User` is deleted, we want to delete this too, and not the other way around, right? Part of the reason I ask is that this property is absent in other places where we might expect it, but the deletion behavior seems to work anyway. `CourseRun` for example. The `course` property is defined as `models.ForeignKey(Course, null=True)`. Are we just setting this explicitly here because it's a Django best practice?

noisecapella: It will be required to make this explicit for Django 2.0: https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey

Regarding `CASCADE` vs `SET_NULL`, I figured there wasn't much point keeping `SentAutomaticEmail` here if one of the two fields were null

mail/factories.py
noisecapella: Sure

gsidebo:
Just the two comments in there which I will leave to your discretion. This looks great 👍

Issue #2276, opened 2017-01-09
user has 2 python social auth objects


pdpinch:
Not sure how this happened, but now it seems to generate an error with every batch update.

```
MultipleObjectsReturned: get() returned more than one UserSocialAuth -- it returned 2!
```

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/201751696/

```
MultipleObjectsReturned: get() returned more than one UserSocialAuth -- it returned 2!
  File "dashboard/tasks.py", line 96, in batch_update_user_data_subtasks
    UserSocialAuth.objects.get(user=user)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 389, in get
    (self.model._meta.object_name, num)

user "fgrodriguez74" does not have python social auth object
```

pdpinch:
@giocalitri can we just modify the code to use the first social auth?

giocalitri:
in production we have only one case like this:
```
micromasters=> select user_id, count(user_id) from social_auth_usersocialauth group by user_id having count(user_id) > 1;
 user_id | count
---------+-------
    6497 |     2
(1 row)
```

the problem I can see is that the 2 entries are actually associated with 2 different user ids on edX
```
In [1]: from django.contrib.auth.models import User
In [2]: user = User.objects.get(username='fgrodriguez74')
In [3]: aa = list(user.social_auth.all())
In [4]: aa[0].uid
Out[4]: 'fgrodriguez74'
In [5]: aa[1].uid
Out[5]: 'fgrodriguez'
```

Now considering that the user id created locally is `fgrodriguez74`, I assume that the right social auth is the one with the same user id and I have no idea how the association of a different user id has been possible.

Now given that the profile is not filled out and there are no financial aid applications, I suggest to either delete the user completely or we could email the user asking which one is the username on edx and delete the other one.

The solution to use just the first social auth could not work if that one is the wrong one.

giocalitri:
Follow up question: can an user change username on edX?

pdpinch:
Here's another case: https://micromasters.mit.edu/admin/social_auth/usersocialauth/?q=pogorsky and https://sentry.io/mit-office-of-digital-learning/micromasters/issues/201751696/events/4986340904/

I have to agree, it does look like they changed their username, but the edX account settings page says you can't do that: 

[...]

PR #2786, opened 2017-03-06
Force logout before login after a 400/401 error from rest API

### LINKED ISSUE #2776: End the current session before forcing the user to log in again
We have had two cases occur now where a user has associated two edX social auth accounts with one MicroMasters account (see #2276). All our interfaces assume a one-to-one relationship. 

We think this happens because a user is logged into MicroMasters with one account and then, for whatever reason, needs to log into edX again (for example, their token was invalidated). If the user has a current micromasters session, and logs into edX with another account, that 2nd account will be attached to their existing MicroMasters account. 

A simple way to avoid this is to make sure the user is logged out of MicroMasters whenever we reauthenticate  into edX. If we recognize the account, the user will log in as normal. If we don't, we'll create a new account. 

To avoid situations like #2276, we should be sure to log the user out of their current session before asking them to log in again with edX.

giocalitri:
fixes #2776 

#### What's this PR do?
Forces a logout before the login after a 400 or 401 error from a rest api call.

1.  you need 2 users on edx and one of the 2 should not be a micromasters user.
2. login with the first user on micromasters and visit the dashboard
3. open another tab on the admin of micromasters and open the user social object of the logged in user
4. in another tab of the same browser logout from edx and login with the second user
5. on the micromasters admin, remove the content of the `extra_data` and save 
6. reload the dashboard 

what you should see is the login of the second user.

this is a fix for the issue observed in #2276

### DIFF
--- a/static/js/lib/api.js
+++ b/static/js/lib/api.js
@@ -104,7 +104,8 @@ const _fetchJSONWithCSRF = (input: string, init: Object = {}, loginOnError: bool
     // to force a new login
     if (loginOnError === true && (response.status === 400 || response.status === 401)) {
       const relativePath = window.location.pathname + window.location.search;
-      window.location = `/login/edxorg/?next=${encodeURIComponent(relativePath)}`;
+      const loginRedirect = `/login/edxorg/?next=${encodeURIComponent(relativePath)}`;
+      window.location = `/logout?next=${encodeURIComponent(loginRedirect)}`;
     }
 
     // For non 2xx status codes reject the promise adding the status code
--- a/static/js/lib/api_test.js
+++ b/static/js/lib/api_test.js
@@ -668,7 +668,8 @@ describe('api', function() {
           });
 
           return assert.isRejected(fetchJSONWithCSRF('/url', {}, true)).then(() => {
-            assert.include(window.location.toString(), '/login/edxorg/');
+            const redirectUrl = `/logout?next=${encodeURIComponent('/login/edxorg/')}`;
+            assert.include(window.location.toString(), redirectUrl);
           });
         });
       }


static/js/lib/api.js
```diff
@@ -104,7 +104,8 @@ const _fetchJSONWithCSRF = (input: string, init: Object = {}, loginOnError: bool
     // to force a new login
     if (loginOnError === true && (response.status === 400 || response.status === 401)) {
       const relativePath = window.location.pathname + window.location.search;
-      window.location = `/login/edxorg/?next=${encodeURIComponent(relativePath)}`;
+      let loginRedirect = `/login/edxorg/?next=${encodeURIComponent(relativePath)}`;
```
singingwolfboy: Use `const` rather than `let` where possible.

static/js/lib/api_test.js
```diff
@@ -668,7 +668,8 @@ describe('api', function() {
           });
 
           return assert.isRejected(fetchJSONWithCSRF('/url', {}, true)).then(() => {
-            assert.include(window.location.toString(), '/login/edxorg/');
+            let redirectUrl = `/logout?next=${encodeURIComponent('/login/edxorg/')}`;
```
singingwolfboy: Use `const` rather than `let` where possible.

singingwolfboy:
On step 6 of your testing instructions I'm getting a 500 server error when my server requests the dashboard API.

singingwolfboy:
When I walked through the testing steps with @giocalitri, we discovered that the steps he had documented above were wrong. In step 5, instead of deleting the entire `extra_data` dict, I had to reduce the `expires_in` down to a small value, and change `refresh_token` and `access_token`. Once I did that, I saw the expected behavior: the browser redirected, and I was prompted to log in again.

giocalitri:
done

PR #2789, opened 2017-03-06
Moved Elasticsearch connection management to own module

noisecapella:
None

#### What's this PR do?
Moves Elasticsearch connection logic to its own module, which will prevent circular imports in a future PR

Learner search page should still work, `./manage.py recreate_index` should work fine


giocalitri:
👍

PR #2793, opened 2017-03-06
Patch search.tasks instead of search.indexing_api

noisecapella:
Part of #2705

#### What's this PR do?
Patches search.tasks instead of search.indexing_api because patches should patch the module where it's invoked rather than the one where it's defined. This worked fine before because `search.indexing_api` functions usually just called other functions, but in a future PR this will not work because the percolate indexing work uses a function in `search.api`

This should only affect tests so no manual review is necessary

### DIFF
--- a/search/base.py
+++ b/search/base.py
@@ -45,6 +45,8 @@ class MockedESTestCase(TestCase):
     def setUpClass(cls):
         cls.patchers = []
         for name, val in tasks.__dict__.items():
+            # This looks for functions starting with _ because those are the functions which are imported
+            # from indexing_api. The _ lets it prevent name collisions.
             if callable(val) and name.startswith("_"):
                 cls.patchers.append(patch('search.tasks.{0}'.format(name), autospec=True))
         for patcher in cls.patchers:
--- a/search/base.py
+++ b/search/base.py
@@ -7,7 +7,7 @@ from django.test import (
     TestCase,
 )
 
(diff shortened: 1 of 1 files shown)


search/base.py
```diff
@@ -44,9 +44,9 @@ class MockedESTestCase(TestCase):
     @classmethod
     def setUpClass(cls):
         cls.patchers = []
-        for name, val in indexing_api.__dict__.items():
-            if callable(val):
-                cls.patchers.append(patch('search.indexing_api.{0}'.format(name), autospec=True))
+        for name, val in tasks.__dict__.items():
+            if callable(val) and name.startswith("_"):
```
singingwolfboy: Why did you add this condition?

noisecapella: I'm not sure it's necessary, I'll see if I can remove it.

singingwolfboy: I actually think that patching _fewer_ functions is better, if we can get away with it, so I think it's better to have the condition there. I just want to know why it was added, and what changed (if anything).

noisecapella: In `search.tasks` we are basically just importing functions from `search.indexing_api` with the same name. They are prefixed with `_` to prevent a name collision

singingwolfboy: Can you add a comment in the code to make that clearer?

noisecapella: Sure, just did

Issue #2796, opened 2017-03-07
Can't create ExamAuthorization for user because of final grade error


pdpinch:
When trying to run the management command `./manage.py authorization_user_exam --username=Rephaelh` I get the following output:

```
Either exam_module is not set for course id="8" or user="Rephaelh" has no enrollment(s)
Either exam_module is not set for course id="9" or user="Rephaelh" has no enrollment(s)
Either exam_module is not set for course id="11" or user="Rephaelh" has no enrollment(s)
Traceback (most recent call last):
  File "./manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/app/exams/management/commands/authorization_user_exam.py", line 50, in handle
    username=username
  File "/app/exams/utils.py", line 180, in bulk_authorize_for_exam
    authorize_for_exam_given_program(program, username)
  File "/app/exams/utils.py", line 212, in authorize_for_exam_given_program
    authorize_for_latest_passed_course(mmtrack, course_id)
  File "/app/exams/utils.py", line 242, in authorize_for_latest_passed_course
    mmtrack.has_passed_course(edx_course_key) and
  File "/app/dashboard/utils.py", line 259, in has_passed_course
    final_grade = self.extract_final_grade(course_id)
  File "/app/dashboard/utils.py", line 216, in extract_final_grade
    status=FinalGradeStatus.COMPLETE,
  File "/app/.heroku/python/lib/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/app/.heroku/python/lib/python3.5/site-packages/django/db/models/query.py", line 385, in get
    self.model._meta.object_name
grades.models.DoesNotExist: FinalGrade matching query does not exist.
```

rhysyngsun:
[...]

PR #2797, opened 2017-03-07
Fixed exam authorization issues when final grade status is pending

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2796

#### What's this PR do?
Fixed an exception issue which appears when user dont have complete final grade object

- run  command `authorization_user_exam` for a user who has final grade status pending, and see logged exception.

@pdpinch

### DIFF
--- a/exams/utils.py
+++ b/exams/utils.py
@@ -18,6 +18,7 @@ from exams.models import (
     ExamProfile,
     ExamAuthorization
 )
+from grades.models import FinalGrade
 from seed_data.utils import add_year
 
 
@@ -88,6 +89,29 @@ def _has_paid_for_exam(mmtrack, edx_course_key):
     return suppress_payment_for_exam or mmtrack.has_paid(edx_course_key)
 
 
+def _has_passed_course(mmtrack, edx_course_key):
+    """
+    Returns True if user has passed the course
+
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        edx_course_key (str): An edX course key (CourseRun.edx_course_key)
+
+    Returns:
+        bool: whether user has passed the course or not.
+    """
+    has_passed_course = False
+    try:
+        has_passed_course = mmtrack.has_passed_course(edx_course_key)
+    except FinalGrade.DoesNotExist:
+        log.exception(
+            '[Exam authorization] User: %s do not have complete FinalGrade for course %s',
+            mmtrack.user.username,
+            edx_course_key
+        )
+    return has_passed_course
+
+
 def authorize_for_exam(mmtrack, course_run):
     """
     Authorize user for exam if he has paid for course and passed course.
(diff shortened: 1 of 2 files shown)


exams/utils.py
```diff
@@ -88,6 +89,29 @@ def _has_paid_for_exam(mmtrack, edx_course_key):
     return suppress_payment_for_exam or mmtrack.has_paid(edx_course_key)
 
 
+def _has_passed_course(mmtrack, edx_course_key):
```
pdpinch: Why do we need yet another method for checking if a user has passed?

exams/utils.py
```diff
@@ -238,11 +262,7 @@ def authorize_for_latest_passed_course(mmtrack, course_id):
     for enrollment in enrollments_qset:
         # only latest passed course_run per course allowed
         edx_course_key = enrollment.course_run.edx_course_key
-        has_paid_and_passed = (
-            mmtrack.has_passed_course(edx_course_key) and
-            mmtrack.has_paid(edx_course_key)
-        )
-        if has_paid_and_passed:
+        if _has_passed_course(mmtrack, edx_course_key) and _has_paid_for_exam(mmtrack, edx_course_key):
```
pdpinch: does this honor the `SUPPRESS_PAYMENT_FOR_EXAM` feature flag?

pdpinch:
@giocalitri can you review?

@rhysyngsun can we add this to the release to get it out today?

exams/utils.py
amir-qayyum-khan: @pdpinch 
- yes `SUPPRESS_PAYMENT_FOR_EXAM` flag is only to disable payment check.
- user must pass the course to get exam authorization happen.

exams/utils.py
amir-qayyum-khan: this is class method, here we suppress payment check if `SUPPRESS_PAYMENT_FOR_EXAM=True`

giocalitri: just to be precise, this is not a `classmethod` this is a function at the module level and I do not understand why we need it: how do you suppress payment?

if the purpose is just to wrap it in a try except, you can wrap the call to `extract_final_grade` inside `has_passed_course` and fix the tests around. I did not do it because I thought it was ok to raise, but it looks like it is better to do it.

exams/utils.py
giocalitri: why there is `_has_paid_for_exam` call here?

amir-qayyum-khan: @giocalitri updated my answers above.

amir-qayyum-khan: yes you are right we dont need it here

amir-qayyum-khan: `_has_passed_course` wraps code in try/except, i am also logging exception

amir-qayyum-khan:
Done @giocalitri

amir-qayyum-khan:
@giocalitri @rhysyngsun will continue it

PR #2802, opened 2017-03-07
Reintroduce new course enrollment UX

### LINKED ISSUE #2467: New Enroll Now UX

### LINKED ISSUE #2743: Displaying different prices in the Courses card and Enroll Now UX
On my dashboard, I am seeing that my "Personal Course Price is $150 USD per course". When I click Enroll Now, in the modal the Pay Now button shows the price as $1000. Also, clicking Pay Now goes to the Order Summary page with a price of $1000.

image

singingwolfboy:
Fixes #2467 (again) and fixes #2743.

#### What's this PR do?
Reintroduces the changes from #2519, with a bugfix so that it intelligently picks the correct price from the price map, instead of a random one.

Same procedure as #2519. To test the bugfix, set up two different programs with two different prices, and verify that the correct price is displayed for each program.


static/js/components/dashboard/CourseListCard.js
```diff
@@ -42,17 +42,30 @@ export default class CourseListCard extends React.Component {
     setEnrollCourseDialogVisibility: (bool: boolean) => void,
   };
 
+  sortedCourseRuns(): Array<CourseRun> {
```
noisecapella: Can you make this a utility function to make it easier to test?

static/js/components/dashboard/CourseListCard.js
```diff
@@ -11,54 +11,196 @@ import type {
...
+      let price;
+      for (const courseRun of this.sortedCourseRuns()) {
+        price = prices.get(courseRun.id);
+        if (price) {
+          break;
+        }
+      }
+
+      if (!price) {
+        throw "missing price"; // this should never happen, it's just to make flow happy
```
noisecapella: Can you return `null` here instead? If there are no course runs available we should not show the price message at all

singingwolfboy: Could that ever actually happen? I'm throwing an error rather than returning null, so that I can define the flow type of this function to _always_ return a React element. If it's possible to return a null, then that's fine, but I'll need to change the flow type.

static/js/components/dashboard/CourseListCard.js
singingwolfboy: Sure, I pulled it out and wrote automated tests in the most recent commit.

static/js/components/dashboard/CourseListCard.js
noisecapella: It could happen if there are no runs in the course. Can you change the flow return type?

PR #2729, opened 2017-03-01
Search: Included username and e-mail address in name search

amir-qayyum-khan:
#### What are the relevant tickets?
fixes https://github.com/mitodl/micromasters/issues/2668

#### What's this PR do?
Added email and username as search able fields.

#### How should this be manually tested?
**Run:** `docker-compose run web ./manage.py recreate_index`
GO to `http://192.168.99.100:8079/learners/` and in search box add email or user name

@pdpinch 
#### Screenshots (if appropriate)
screen shot 2017-03-01 at 1 29 46 pm

screen shot 2017-03-01 at 1 28 41 pm

screen shot 2017-03-01 at 1 29 37 pm

PR #2810, opened 2017-03-07
Revert username email search

amir-qayyum-khan:
revert https://github.com/mitodl/micromasters/pull/2729

#### What's this PR do?
- revert search by username and email from rc

@pdpinch @rhysyngsun


amir-qayyum-khan:
no need of it, the issue was we missed reindexing on rc

Issue #1445, opened 2016-10-18
Enroll and pay later:  Better to not reload the whole dashboard page


Ferdi:
Can we make this so only the relevant part of the page gets updated (e.g. course status) without full reload of the page ?

https://share.viewedit.com/z4ReSsmdB9z22LicyaFz7a

pdpinch:
There may be some code that can be used from the order callback process.

pdpinch:
@roberthouse54 what should be the UX here? Do we put up a spinner over the link, or over the whole card, or something else?

roberthouse54:
Here is what I propose we do. When the user clicks Enroll and Pay Later, we show the spinner as we do now, then we put an overlay over the whole card, with a checkmark, and the message, "You are now auditing this course <br>But you still need to pay to get credit." After a few seconds, this fades and we see the Auditing status.

![image](https://cloud.githubusercontent.com/assets/20047260/21231347/f09e52cc-c2b5-11e6-94e5-7aea94a592b0.png)

Ferdi:
@roberthouse54  This looks good.  One question re:

>  we show the spinner as we do now

Do we currently show a spinner ?

roberthouse54:
Yeah, George just recently added spinners for pretty much any time a button is submitted. Not sure if those changes are all merged or not

noisecapella:
They should be all done now

Ferdi:
@noisecapella this looks like a but, only the one you click should spin 

https://share.viewedit.com/unSaaZAt6fkYVA7u9YLZsz

Ferdi:
that was in micromasters-ci

noisecapella:
I noticed it too, I filed an issue a few days ago: https://github.com/mitodl/micromasters/issues/2140

Ferdi:
great thx

pdpinch:
@noisecapella what's the status of this? I think the specific issue of the enroll now & pay later button has been addressed.

noisecapella:
This is not yet done. The enroll button shows a spinner during API activity but afterwards the entire dashboard is still updated at once. The fix should be simple but the work just hasn't been done yet

rhysyngsun:
Putting this back into Ready until #2519 is merged since that's touching code this will need to change too.

PR #2821, opened 2017-03-08
Enroll and pay later: Load dashboard page without reloading

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/1445

#### What's this PR do?
- Show inline success message when user press enroll and pay later button,
- It also loads dashboard page without reloading whole page

- Open `http://192.168.99.100:8079/dashboard/` and press enroll and pay later.

screen shot 2017-03-13 at 3 57 59 pm

### DIFF
--- a/static/js/actions/course_enrollments.js
+++ b/static/js/actions/course_enrollments.js
@@ -7,6 +7,7 @@ import { fetchDashboard } from './dashboard';
 import { fetchCoursePrices } from './course_prices';
 import type { Dispatcher } from '../flow/reduxTypes';
 import * as api from '../lib/api';
+import { showEnrollPayLaterSuccess } from './ui';
 
 export const REQUEST_ADD_COURSE_ENROLLMENT = 'REQUEST_ADD_COURSE_ENROLLMENT';
 export const requestAddCourseEnrollment = createAction(REQUEST_ADD_COURSE_ENROLLMENT);
@@ -22,14 +23,28 @@ export const addCourseEnrollment = (courseId: string): Dispatcher<*> => {
     dispatch(requestAddCourseEnrollment(courseId));
     return api.addCourseEnrollment(courseId).
       then(() => {
-        dispatch(receiveAddCourseEnrollmentSuccess());
-        dispatch(fetchDashboard(SETTINGS.user.username));
-        dispatch(fetchCoursePrices(SETTINGS.user.username));
+        dispatch(receiveAddCourseEnrollmentSuccess(courseId));
+        dispatch(fetchDashboard(SETTINGS.user.username, true));
+        dispatch(fetchCoursePrices(SETTINGS.user.username, true));
+        dispatch(showEnrollPayLaterSuccessMessage(courseId));
         return Promise.resolve();
       }).
       catch(() => {
         dispatch(receiveAddCourseEnrollmentFailure());
         return Promise.reject();
       });
+  };
+};
+
+export const showEnrollPayLaterSuccessMessage = (courseId: string): Dispatcher<*> => {
+  return (dispatch: Dispatch) => {
+    dispatch(showEnrollPayLaterSuccess(courseId));
+    let promise = new Promise(function(resolve) {
+      window.setTimeout(function() {
+        dispatch(showEnrollPayLaterSuccess(null));
+        resolve();
+      }, 9000);
+    });
+    return promise;
   };
 };
--- a/static/js/components/dashboard/CourseListCard_test.js
+++ b/static/js/components/dashboard/CourseListCard_test.js
@@ -12,6 +12,7 @@ import CourseListCard from './CourseListCard';
 import CourseRow from './CourseRow';
 import { DASHBOARD_RESPONSE, COURSE_PRICES_RESPONSE } from '../../test_constants';
 import { INITIAL_EMAIL_STATE } from '../../reducers/email';
+import { INITIAL_UI_STATE } from '../../reducers/ui';
 import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import getMuiTheme from 'material-ui/styles/getMuiTheme';
 import IntegrationTestHelper from '../../util/integration_test_helper';
@@ -58,6 +59,7 @@ describe('CourseListCard', () => {
             coursePrice={coursePrice}
             addCourseEnrollment={() => Promise.resolve()}
             prices={prices}
+            ui={INITIAL_UI_STATE}
             openCourseContactDialog={() => undefined}
             closeEmailDialog={() => undefined}
             updateEmailEdit={() => undefined}
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -262,14 +262,14 @@ class DashboardPage extends React.Component {
 
   fetchDashboard() {
     const { dashboard, dispatch } = this.props;
-    if (dashboard.fetchStatus === undefined && dashboard.noSpinner === false) {
+    if (dashboard.fetchStatus === undefined) {
       dispatch(fetchDashboard(SETTINGS.user.username));
     }
   }
 
   fetchCoursePrices() {
     const { prices, dispatch } = this.props;
-    if (prices.fetchStatus === undefined && prices.noSpinner === false) {
+    if (prices.fetchStatus === undefined) {
       dispatch(fetchCoursePrices(SETTINGS.user.username));
     }
   }
@@ -716,7 +716,10 @@ class DashboardPage extends React.Component {
       dashboard,
       prices
     } = this.props;
-    const loaded = R.none(isProcessing, [dashboard.fetchStatus, prices.fetchStatus]);
+    const loaded = R.or(
+      R.none(isProcessing, [dashboard.fetchStatus, prices.fetchStatus]),
+      R.or(dashboard.noSpinner, prices.noSpinner)
+    );
     const fetchStarted = !_.isNil(prices.fetchStatus) && !_.isNil(dashboard.fetchStatus);
 
     const errorMessage = this.renderErrorMessage();
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -262,14 +262,14 @@ class DashboardPage extends React.Component {
 
   fetchDashboard() {
     const { dashboard, dispatch } = this.props;
-    if (dashboard.fetchStatus === undefined) {
+    if (dashboard.fetchStatus === undefined && dashboard.noSpinner === false) {
       dispatch(fetchDashboard(SETTINGS.user.username));
     }
   }
 
   fetchCoursePrices() {
     const { prices, dispatch } = this.props;
-    if (prices.fetchStatus === undefined) {
+    if (prices.fetchStatus === undefined && prices.noSpinner === false) {
       dispatch(fetchCoursePrices(SETTINGS.user.username));
     }
   }
@@ -694,6 +694,7 @@ class DashboardPage extends React.Component {
               coursePrice={coursePrice}
               prices={calculatedPrices}
               key={program.id}
+              ui={ui}
               openFinancialAidCalculator={this.openFinancialAidCalculator}
               addCourseEnrollment={this.addCourseEnrollment}
               openCourseContactDialog={this.openCourseContactDialog}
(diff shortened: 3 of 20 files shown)


amir-qayyum-khan:
@noisecapella can you verify that i am on right track?

static/js/components/dashboard/CourseListCard_test.js
```diff
@@ -58,6 +59,7 @@ describe('CourseListCard', () => {
             coursePrice={coursePrice}
             addCourseEnrollment={() => Promise.resolve()}
             prices={prices}
+            ui={{ showEnrollPayLaterSuccess: false }}
```
amir-qayyum-khan: @noisecapella i am getting flow warnings here, it is excepting all from `UIState`;

amir-qayyum-khan: fixed

static/js/reducers/dashboard.js
```diff
@@ -36,7 +36,11 @@ export const dashboard = (state: DashboardsState = INITIAL_DASHBOARDS_STATE, act
   case REQUEST_DASHBOARD: // eslint-disable-line no-case-declarations
     let newBaseState = R.dissoc(username, state);
     if (action.payload === true) {
-      return updateDashboardState(newBaseState, username, INITIAL_DASHBOARD_STATE);
+      return updateDashboardState(
+        newBaseState,
+        username,
+        _.merge({}, INITIAL_DASHBOARD_STATE, { fetchStatus: null })
```
aliceriot: what's the reason for this change? the `newBaseState` shouldn't have anything saved in it under the `username` and `INITIAL_DASHBOARD_STATE` doesn't have a `fetchStatus` key on it.

static/js/actions/course_enrollments.js
```diff
@@ -22,9 +23,10 @@ export const addCourseEnrollment = (courseId: string): Dispatcher<*> => {
     dispatch(requestAddCourseEnrollment(courseId));
     return api.addCourseEnrollment(courseId).
       then(() => {
-        dispatch(receiveAddCourseEnrollmentSuccess());
-        dispatch(fetchDashboard(SETTINGS.user.username));
-        dispatch(fetchCoursePrices());
+        dispatch(receiveAddCourseEnrollmentSuccess(courseId));
+        dispatch(fetchDashboard(SETTINGS.user.username, true));
+        dispatch(fetchCoursePrices(true));
+        dispatch(showEnrollPayLaterSuccessMessage());
```
aliceriot: probably you want the `.then` callback here to return this call (`dispatch(showEnrollPayLaterSuccessMessage());`)

static/js/actions/course_enrollments.js
```diff
@@ -33,3 +35,16 @@ export const addCourseEnrollment = (courseId: string): Dispatcher<*> => {
...
+    dispatch(showEnrollPayLaterSuccess(true));
+    let promise = new Promise(function(resolve) {
+      window.setTimeout(function() {
+        dispatch(showEnrollPayLaterSuccess(false));
+        resolve('done!');
+      }, 9000);
+    });
+    return promise;
+  };
+};
```
aliceriot: this function would probably be better located in an actions file

aliceriot:
@amir-qayyum-khan I left a few comments for some small things I noticed. I think @noisecapella is much more familiar with the code at hand, though, so probably is a better person for that!

noisecapella:
@amir-qayyum-khan It looks like you are on the right track though I haven't done a functionality test. Can you add unit tests?

amir-qayyum-khan:
ok @noisecapella

amir-qayyum-khan:
Done @noisecapella, please review.

static/js/reducers/ui_test.js
```diff
@@ -218,6 +219,14 @@ describe('ui reducers', () => {
     });
   });
 
+  describe('show enroll pay later success alert', () => {
+    it('should let you set the pay later success alert', () => {
+      assertReducerResultState(
+        showEnrollPayLaterSuccess, ui => ui.showEnrollPayLaterSuccess, {}
```
amir-qayyum-khan: @noisecapella this is failing on CI, but it is working fine on my machine.
```AssertionError: expected {} to deeply equal { course_id: 'course-v1:48' }
      + expected - actual
      -{}
      +{
      +  "course_id": "course-v1:48"
      +}
      ```

File: static/js/actions/course_enrollments.js (hunk 4 above)
Comment (amir-qayyum-khan) [inline static/js/actions/course_enrollments.js]: it is in actions actions/course_enrollments.js,

by the way done with addressing ur comments

**Comment by noisecapella** [comment]
It looks like there's actually two dashboard refreshes happening, when the user clicks the enroll button:

screenshot from 2017-03-15 15-32-51

This is why the spinner is still showing up even though `noSpinner` is true for one of the actions

File: static/js/actions/course_enrollments.js [hunk 6]
```diff
@@ -33,3 +36,16 @@ export const addCourseEnrollment = (courseId: string): Dispatcher<*> => {
       });
   };
 };
+
+export const showEnrollPayLaterSuccessMessage = (courseId: string): Dispatcher<*> => {
+  return (dispatch: Dispatch) => {
+    dispatch(showEnrollPayLaterSuccess({course_id: courseId}));
```
noisecapella: If the only thing we care about is the course key, can we just pass `courseId` here by itself instead of within an object?

static/js/actions/course_enrollments.js
```diff
@@ -33,3 +36,16 @@ export const addCourseEnrollment = (courseId: string): Dispatcher<*> => {
...
+    let promise = new Promise(function(resolve) {
+      window.setTimeout(function() {
+        dispatch(showEnrollPayLaterSuccess(INITIAL_ENROLL_PAY_LATER_STATE));
```
noisecapella: Can you add a test for this code?

static/js/reducers/index.js
```diff
@@ -211,10 +211,14 @@ const INITIAL_COURSE_PRICES_STATE: CoursePricesState = {
 export const prices = (state: CoursePricesState = INITIAL_COURSE_PRICES_STATE, action: Action<any, null>) => {
   switch (action.type) {
   case REQUEST_COURSE_PRICES:
-    return {
-      ...state,
-      fetchStatus: FETCH_PROCESSING
-    };
+    if (action.payload === true) {
+      return INITIAL_COURSE_PRICES_STATE;
```
noisecapella: Can you add a test for this?

amir-qayyum-khan:
@noisecapella added tests, i need your help in investigation why spinner is still running.
can you give me headsup

noisecapella:
I investigated and it looks like this is happening because the `noSpinner` option clears `fetchStatus`, then `DashboardPage` interprets that to mean that the dashboard page fetch has not happened yet. I think the best way to fix this is to change the reducer to add `noSpinner` explicitly to the state if it's `true` and leave `fetchStatus` alone. Then `DashboardPage` will need to look at both `fetchStatus` and `noSpinner` in order to determine whether to show the spinner. Does that make sense?

amir-qayyum-khan:
Fixed @noisecapella

static/js/reducers/dashboard.js
```diff
@@ -34,7 +35,11 @@ export const dashboard = (state: DashboardsState = INITIAL_DASHBOARDS_STATE, act
   case REQUEST_DASHBOARD: // eslint-disable-line no-case-declarations
     let newBaseState = R.dissoc(username, state);
     if (action.payload === true) {
-      return updateStateByUsername(newBaseState, username, INITIAL_DASHBOARD_STATE);
+      return updateStateByUsername(
+        newBaseState,
+        username,
+        _.merge({}, INITIAL_DASHBOARD_STATE, { noSpinner: true })
```
noisecapella: We need to also set `fetchStatus` here. Instead of having the `if (action.payload === true)` we could just have

    {
      fetchStatus: FETCH_PROCESSING,
      noSpinner: action.payload,
    }

amir-qayyum-khan:
Done @noisecapella

static/js/reducers/dashboard.js
amir-qayyum-khan: if i do this spinner start spinning

static/js/reducers/dashboard.js
```diff
@@ -34,7 +35,13 @@ export const dashboard = (state: DashboardsState = INITIAL_DASHBOARDS_STATE, act
   case REQUEST_DASHBOARD: // eslint-disable-line no-case-declarations
     let newBaseState = R.dissoc(username, state);
     if (action.payload === true) {
-      return updateStateByUsername(newBaseState, username, INITIAL_DASHBOARD_STATE);
+      return updateStateByUsername(
+        newBaseState,
+        username,
+        _.merge({}, INITIAL_DASHBOARD_STATE, {
+          noSpinner: true
```
noisecapella: We should have `fetchStatus: FETCH_PROCESSING` here too, and because the two branches of this if statement are similar, you can just set `noSpinner: action.payload`.

Regarding your other comment on the dashboard spinner still spinning, we should change that code to also check `noSpinner`

static/js/containers/DashboardPage.js
```diff
@@ -262,14 +262,14 @@ class DashboardPage extends React.Component {
 
   fetchDashboard() {
     const { dashboard, dispatch } = this.props;
-    if (dashboard.fetchStatus === undefined) {
+    if (dashboard.fetchStatus === undefined && dashboard.noSpinner === false) {
```
noisecapella: If you set `fetchStatus` in all cases as described in my previous comment, you shouldn't need to check `noSpinner` at all here or below.

static/js/reducers/course_prices.js
```diff
@@ -21,22 +21,36 @@ import type {
...
-      R.dissoc(username, state),
-      username,
-      _.merge({}, INITIAL_COURSE_PRICES_STATE, { fetchStatus: FETCH_PROCESSING })
-    );
+    if (action.payload === true) {
+      return updateStateByUsername(
+        R.dissoc(username, state),
+        username,
+        _.merge({}, INITIAL_COURSE_PRICES_STATE, {
+          noSpinner: true
```
noisecapella: Please set `fetchStatus` here too and merge the two if branches like I suggested in the dashboard reducer

static/js/containers/DashboardPage.js
```diff
@@ -694,6 +694,7 @@ class DashboardPage extends React.Component {
               coursePrice={coursePrice}
               prices={calculatedPrices}
               key={program.id}
+              ui={ui}
               openFinancialAidCalculator={this.openFinancialAidCalculator}
               addCourseEnrollment={this.addCourseEnrollment}
               openCourseContactDialog={this.openCourseContactDialog}
```
noisecapella: Look below in `render()`. You should be checking `noSpinner` there in addition to `fetchStatus`

amir-qayyum-khan:
done @noisecapella

static/js/reducers/course_prices.js
```diff
@@ -21,22 +21,37 @@ import type {
...
 export const prices = (state: CoursePriceReducerState = {}, action: Action<any, string>) => {
   const { meta: username } = action;
   switch (action.type) {
   case REQUEST_COURSE_PRICES:
-    return updateStateByUsername(
-      R.dissoc(username, state),
-      username,
-      _.merge({}, INITIAL_COURSE_PRICES_STATE, { fetchStatus: FETCH_PROCESSING })
-    );
+    if (action.payload === true) {
```
noisecapella: Please remove this if statement. Since the only difference between the two cases is the `noSpinner`, you can have this instead:

    return updateStateByUsername(
      R.dissoc(username, state),
      username,
      _.merge({}, INITIAL_COURSE_PRICES_STATE, {
         noSpinner: action.payload,
        fetchStatus: FETCH_PROCESSING
      })
    );

Please do the same for the dashboard reducer

amir-qayyum-khan: done @noisecapella

noisecapella:
@amir-qayyum-khan The spinner is gone but it seems like the dashboard page temporarily disappears while the loading happens. I think we need to update the reducer to preserve the old data in the case that `noSpinner` is true. Can you make the change to the prices and dashboard reducers?

amir-qayyum-khan:
fixed @noisecapella , please take a look

PR #2822, opened 2017-03-08
Fix exam auth eligibility date handlingi (#2814)

### LINKED ISSUE #2814: Exams: dashboard ui doesn't handle authorizations created today
Because we store `date_first_eligible` and `date_last_eligible` as `date` this logic should be `<=` instead of `<`: https://github.com/mitodl/micromasters/blob/master/dashboard/utils.py#L422-L423

rhysyngsun:
Fixes #2814 

#### What's this PR do?
Fixes the dashboard UI so that exam authorizations that were created today and are schedule-able show the button to the user.

Create an `ExamProfile` and `ExamAuthorization` record for your user, both with `success` status. Test with setting `date_last_eligible` to today and then test `date_first_eligible` to the same.

### DIFF
--- a/dashboard/utils_test.py
+++ b/dashboard/utils_test.py
@@ -31,7 +31,7 @@ from micromasters.utils import load_json_from_file
 from search.base import MockedESTestCase
 
 
-# pylint: disable=too-many-arguments
+# pylint: disable=too-many-arguments, too-many-lines
 
 
 @ddt.ddt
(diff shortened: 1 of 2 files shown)


dashboard/utils_test.py
```diff
@@ -974,3 +974,35 @@ def test_get_pearson_exam_status(self, profile_status, expected_status, set_seri
...
+        mmtrack = MMTrack(
+            user=self.user,
+            program=self.program,
+            edx_user_data=self.cached_edx_user_data
+        )
+
+        # should be considered schedulable if past <= datetime.now() <= future
+        for now_value in [past, now, future]:
+            mmtrack.now = now_value
+            assert mmtrack.get_pearson_exam_status() == ExamProfile.PROFILE_SCHEDULABLE
```
noisecapella: Do we have a test somewhere for dates before and after these dates?

rhysyngsun: Good point

rhysyngsun: @noisecapella added

PR #2824, opened 2017-03-08
Freeze grade modified to be race condition safe

### LINKED ISSUE #2819: ValidationError: {'__all__': ['Final grade with this User and Course run already exists.']}
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/228372009/

```
ValidationError: {'__all__': ['Final grade with this User and Course run already exists.']}
(17 additional frame(s) were not displayed)
...
  File "dashboard/api.py", line 225, in get_info_for_course
    for course_run in course.courserun_set.all()]
  File "dashboard/api.py", line 225, in <listcomp>
    for course_run in course.courserun_set.all()]
  File "dashboard/api.py", line 321, in get_status_for_courserun
    api.freeze_user_final_grade(mmtrack.user, course_run, raise_on_exception=True)
  File "grades/api.py", line 186, in freeze_user_final_grade
    course_run_paid_on_edx=final_grade.payed_on_edx
  File "grades/models.py", line 74, in save
    self.full_clean()

ValidationError: {'__all__': ['Final grade with this User and Course run already exists.']}
```

giocalitri:
fixes #2819  (hopefully)

#### What's this PR do?
modifies the function `grades.api.freeze_user_final_grade` to be race condition safe

I have not fully understood how users get in this situation: my guess is that the dashboard rest api is slow enough that they reload the page and trigger another request, but I have no proves for this.

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -177,7 +177,9 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
                     course_run.edx_course_key
                 )
             ) from ex
-    return FinalGrade.objects.create(
+    # the final grade at this point should not exists, but putting a `get_or_create`
+    # should solve the problem when the function is called synchronously from the dashboard REST API multiple times
+    final_grade_obj, _ = FinalGrade.objects.get_or_create(
         user=user,
         course_run=course_run,
         grade=final_grade.grade,
@@ -185,3 +187,4 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
         status=FinalGradeStatus.COMPLETE,
         course_run_paid_on_edx=final_grade.payed_on_edx
     )
+    return final_grade_obj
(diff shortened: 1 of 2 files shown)


grades/api.py
```diff
@@ -177,11 +177,14 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
                     course_run.edx_course_key
                 )
             ) from ex
-    return FinalGrade.objects.create(
+    # the final grade at this point should not exists, but putting a `get_or_create`
+    # should solve the problem when the function is called synchronously from the dashboard REST API multiple times
+    final_grade_obj, _ = FinalGrade.objects.get_or_create(
```
noisecapella: I'm not sure if we should be looking up the `FinalGrade` with all of these parameters. The unique constraint is on `user` and `course_run`, so if the problem was that the record already existed in the database, we will still get the unique constraint error. Maybe we want an exception in that case however? I'm not sure

giocalitri: I think that if 2 (assumed concurrent) requests to freeze the grade give different results, we have a much bigger problem and I probably would like to see it.

PR #2826, opened 2017-03-08
Change date format to be globally accessible

### LINKED ISSUE #2820: Change Date Formatting to be Globally Accessible
In both the app and the marketing site, we are formatting dates as "mm/dd/yyyy". This is not globally accessible, as in come countries this date would be read as "dd/mm/yyyy". 

We should change all instances where we show dates to something that is more globally understood. The easiest way to do this is to write out the month. I suggest that we can abbreviate the months as follows:

Jan, Feb, Mar, Apr, May, June, July, Aug, Sep, Oct, Nov, Dec

And we will format the date as follows:

Sep 10, 2017  --- with no decimal at the end of the abbreviated month.

singingwolfboy:
Fixes #2820

#### What's this PR do?
Changes the dashboard format string for formatting dates. Does _not_ change how months are formatted, e.g. for education and employment.

Verify that dates are formatted in the format described in #2820.

date format

### DIFF
--- a/static/js/components/dashboard/FinancialAidCard_test.js
+++ b/static/js/components/dashboard/FinancialAidCard_test.js
@@ -122,7 +122,7 @@ describe("FinancialAidCard", () => {
       program.financial_aid_user_info.has_user_applied = false;
       let wrapper = renderCard({ program });
       wrapper.find('.full-price').simulate('click');
-      sinon.assert.calledWith(setSkipDialogStub, true);
+      assert.ok(setSkipDialogStub.calledWith(true), 'Dialog should get opened');
     });
   });
 
@@ -169,7 +169,7 @@ describe("FinancialAidCard", () => {
         let wrapper = renderCard({ program, setDocsInstructionsVisibility });
         let link = wrapper.find('.financial-aid-box').find('a');
         link.simulate('click');
-        sinon.assert.called(setDocsInstructionsVisibility);
+        assert.ok(setDocsInstructionsVisibility.called, 'should have called onClick handler');
       });
 
       it(`instruction for status ${FA_STATUS_PENDING_DOCS}`, () => {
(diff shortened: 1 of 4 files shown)


roberthouse54:
Can we make the same change on the Marketing page for Programs. Also the home page (if any). I know there are dates on the Program page.

roberthouse54:
There are dates in the Courses box on the Program page: (see below)
dates

pdpinch:
@roberthouse54 I think we should address the program page in a separate PR, since it's completely different code.

static/js/components/dashboard/FinancialAidCard_test.js
```diff
@@ -122,7 +122,7 @@ describe("FinancialAidCard", () => {
       program.financial_aid_user_info.has_user_applied = false;
       let wrapper = renderCard({ program });
       wrapper.find('.full-price').simulate('click');
-      assert.ok(setSkipDialogStub.calledWith(true), 'Dialog should get opened');
+      sinon.assert.calledWith(setSkipDialogStub, true);
```
aliceriot: I think it's nice to have the `assert.ok` message - if you break a test it gives you quick feedback about what didn't happen.

aliceriot: could we leave the `assert.ok` calls for stubs as-is?

singingwolfboy: Are you talking about the customizable message? [Sinon's assertions](http://sinonjs.org/releases/v1.17.7/assertions/) provide error messages that are also very helpful, although perhaps a customized error message should be _more_ helpful. I'd be happy to revert the changes where I removed a customized error message, but Sinon's default error messages are going to be more helpful and informative than the default `assert.ok` error message.

Is that OK? Revert the changes where I'm removing a customized error message, but keep the changes where there was no customized error message to begin with?

aliceriot: Yes, that's what I was suggesting.

singingwolfboy: Done!

singingwolfboy:
@pdpinch @roberthouse54: yes, I already made that change in this pull request. 😸 
changed date formatting on program page

roberthouse54:
Cool. Thanks.

PR #2827, opened 2017-03-08
Fixed various profile field validations for exams (#2804)

### LINKED ISSUE #2804: Exams: First name and last name are too long
We fall back to `first_name` and `last_name` if `romanized_first_name` and/or `romanized_last_name` are not set. We forgot to validate that `first_name` and `last_name` meet the length requirements.

If the `first_name` is > 30 chars and/or if `last_name` is > 50 chars, prompt for the romanized fields.

Related to #2803 since some warnings indicate this issue.

rhysyngsun:
I addressed a few other small issues I noticed while in the code, this is noted in section 2 below.

Fixes #2804 

#### What's this PR do?

* Added length validation to all name fields
* Added CP-1252 validation where it was missing (city).
* Normalized first name -> given name, last name -> family name language

Go to the profile page and enter first names that are longer than 30 characters and long names that are longer than 50 characters and verify you see the error message. Also verify you can't enter non-latin characters in the city since I fixed that too while I was in there.

selection_039

validating


static/js/lib/validation/profile.js
```diff
@@ -105,13 +108,26 @@ export const personalValidation = (profile: Profile) => {
...
+    else if (profile.romanized_first_name.length > 30) {
+      errors.romanized_first_name = "Given name must be less than 30 characters";
     }
 
     if (!profile.romanized_last_name || !CP1252_REGEX.test(profile.romanized_last_name)) {
-      errors.romanized_last_name = "Latin last name is required";
+      errors.romanized_last_name = "Latin family name is required";
+    }
+    else if (profile.romanized_last_name.length > 50) {
+      errors.romanized_last_name = "Family name must be less than 50 characters";
```
aliceriot: what do you think about abstracting the `CP1252` and length checks? Something like:

```js
const checkCP1252 = (profile, key, label) => (
  CP1252_REGEX.test(_.get(profile, key)) ? { [key]: `${label} must be in Latin characters` : null
);

const lengthCheck = (profile, key, requiredLength, label) => (
  _.get(profile, key).length > requiredLength ? { [key]: `${label} must be less than ${requireLength} characters` } : null
);
```

Just so we reduce a little bit of code duplication. Up to you!

aliceriot:
functionality looks great!

PR #2836, opened 2017-03-09
navbar changes

### LINKED ISSUE #2564: Add Navigation - Sidebar

aliceriot:

closes #2564 

#### What's this PR do?

This puts the navbar on all desktop-width pages, and changes around which links show up for staff, which for learners, etc. Check the issue for details on the changes there.

go to any page. you should see a little hamburger in the upper left, which should open the nav menu. there's a little chevron icon which, when clicked, should close the menu again. clicking outside the menu should also close it.

If you're a staff user you should see the following:

asdfasdfasdf

if you're not you should see:

fasfffjfj

to simulate 'learner' status you can set `SETTINGS.roles` to `[]`.


noisecapella:
When I hover over the chevron it looks a little strange:
screenshot from 2017-03-09 13-20-58

aliceriot:
oh oops I meant to fix that, sorry

aliceriot:
@noisecapella just pushed a little styling change that should fix that!

noisecapella:
It looks like the selenium test is failing due to the user menu being removed. Can you update the test to use the new slider link instead?

Also can you force push to redeploy to heroku? Yarn had a timeout

aliceriot:
I always forget about the selenium tests! looking into it now

aliceriot:
ok I think I fixed it!

PR #2839, opened 2017-03-09
Load the edX logo from CloudFront on the program page

singingwolfboy:
Part of #2462.

#### What's this PR do?
I found an image reference in our HTML template that wasn't being loaded from CloudFront. This pull request changes that template to load the image from CloudFront. There should be no user-visible impact to this pull request.

View the program page, and inspect the edX logo on the right sidebar, below the "More Info" box. It should load from CloudFront on the RC server, and on production.

### DIFF
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -130,7 +130,7 @@
           <p>
             Courses delivered on
           </p>
-          <img src="/static/images/edx_logo.png" alt="edX" />
+          <img src="{% static 'images/edx_logo.png' %}" alt="edX" />
         </div>
       </div>
     </div>


cms/templates/cms/program_page.html
```diff
@@ -130,7 +130,7 @@ <h3 class="title">
           <p>
             Courses delivered on
           </p>
-          <img src="/static/images/edx_logo.png" alt="edX" />
+          <img src="{% static 'images/edx_logo.png' %}" alt="edX" />
```
amir-qayyum-khan: i suppose this is the image

screen shot 2017-03-15 at 4 07 35 pm

amir-qayyum-khan:
how can i test it, there is not heroku build for this PR and on local machine it is
`edX`

amir-qayyum-khan:
please add the image in program description, it will help to identify the location of change

PR #2845, opened 2017-03-10
Use temporary index during recreate_index

### LINKED ISSUE #2550: Replace index atomically during recreate_index
Deleting the index means the index is down during the time that the index is being recreated. If we are careful to only do incremental changes and not delete users or `ProgramEnrollment`s, we can skip deleting the index without problems.

noisecapella:
Fixes #2550

#### What's this PR do?
Does the reindex in a separate Elasticsearch index, then deletes the old one, recreates it with a mapping and replaces it with the temp index's documents. There is still a tiny amount of time where the old index will go away and come back (less than a second for 1400 documents on my computer), and there is still a risk of stale data if documents are modified during the reindex.

This PR also updates the indexing to check if the temp index exists, then update it if it does. This is to prevent stale data in the temp index from overwriting the current one.

Run a reindex and go to `/learners`. The list of learners should not change during the reindex.

### DIFF
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -125,27 +125,30 @@ def _delete_item(document_id, doc_type, index):
         pass
 
 
-def index_program_enrolled_users(program_enrollments, index=None, chunk_size=100):
+def index_program_enrolled_users(program_enrollments, indices=None, chunk_size=100):
     """
     Bulk index an iterable of ProgramEnrollments
 
     Args:
         program_enrollments (iterable of ProgramEnrollment): An iterable of program enrollments
-        index (str): An Elasticsearch index. If None, uses Django Elasticsearch index setting
+        indices (list of str): A list of Elasticsearch indices. If None, uses Django Elasticsearch index setting
         chunk_size (int): The number of items per chunk to index
 
     Returns:
         int: Number of indexed items
     """
-    if index is None:
-        index = settings.ELASTICSEARCH_INDEX
+    if indices is None:
+        indices = get_active_indices()
 
-    return _index_chunks(
-        (serialize_program_enrolled_user(program_enrollment) for program_enrollment in program_enrollments),
-        USER_DOC_TYPE,
-        index,
-        chunk_size=chunk_size,
-    )
+    for index in indices:
+        count = _index_chunks(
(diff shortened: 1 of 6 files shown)


noisecapella:
I'll change this to use an alias as we discussed in chat so that there's no downtime

search/indexing_api.py
```diff
@@ -370,34 +369,60 @@ def recreate_index():
...
+        conn.indices.delete(temp_alias)
+
+    # Point temp_alias toward new backing index
+    conn.indices.put_alias(index=new_backing_index, name=temp_alias)
+
+    # Do the indexing on the temp index
     start = datetime.now(pytz.UTC)
     log.info("Indexing %d program enrollments...", ProgramEnrollment.objects.count())
-    index_program_enrolled_users(ProgramEnrollment.objects.iterator(), [temp_index])
+    index_program_enrolled_users(ProgramEnrollment.objects.iterator(), [new_backing_index])
```
rhysyngsun: I would suggest wrapping the indexing calls in a `try..finally` and clean up the temp index/alias in the finally block.

search/indexing_api.py
```diff
@@ -370,34 +369,60 @@ def recreate_index():
...
+        index=settings.ELASTICSEARCH_INDEX,
+        uuid=uuid4().hex,
+    )
+    clear_index(new_backing_index)
+
+    # Clear away temp alias so we can reuse it
+    temp_alias = get_temp_alias()
+    if conn.indices.exists_alias(name=temp_alias):
+        # Deletes both alias and backing indexes
+        conn.indices.delete(temp_alias)
```
rhysyngsun: This looks like a call to delete an index but an alias name is being passed here, so I don't think it actually would do anything in practice.

search/indexing_api.py
noisecapella: Sure

search/indexing_api.py
noisecapella: The index deletion API works with aliases. When the index is deleted using the alias like this the alias is deleted too.

rhysyngsun: ok, makes sense

Issue #2512, opened 2017-02-06
Learners Search for Learners


roberthouse54:
As a learner, I would like to find out more about other learners in my program and contact them individually.

The learner can browse or search for other learners. The UI is the same as the instructor search except that it doesn't include all of the facets, nor the ability to send bulk e-mail. This results shown needs to respect the learner's setting for whether their profile is shared with other learners. If they are set to not be shown, they will not appear in the list. Learner search will be available to all users. The results will list only users who have paid or enrolled in at least one course.

Users should be able to view other users' profiles. This page should look the same as the profile when viewed by faculty, except we will not display the user's email address, and we will display the user's current location. At the top of the profile will be a link to "Send a Message". (see #2196)

### Acceptance Criteria:

- [ ] no ability to send bulk e-mail
- [ ] results are limited to users who have opted to share their profile with other learners in their program, or with the world
- [ ] results are limited to users who have pair or enrolled in at least one course in their program(s)
- [ ] the search interface is available to all users
- [ ] all facets listed below

### Related Issues:

- #2196 : Learners should be able to e-mail other learners. The ability to send emails will be behind a paywall.
- As a follow up to this issue, we will add a card to the dashboard called "Learners in this Program". This card will link to the Learners Search page. This is a issue #2516
- We need a way for users to navigate to this page. See: #2564 

We will display the following search facets (we will not include facets for grade or payment status):
Course
Semester
Current Residence
Country of Birth
Company
~~Number of courses passed~~

![learner_search_for_learners 2x](https://cloud.githubusercontent.com/assets/20047260/22667009/dba93ace-ec88-11e6-9e73-067a562715f1.png)

![image](https://cloud.githubusercontent.com/assets/20047260/22707246/2e034ba4-ed40-11e6-94c6-6b4d5ad4be7b.png)

pdpinch:
I added some acceptance criteria, which are just a list version of Rob's description.

rhysyngsun:
Moved this back to ready because I'm taking care of #2690 as a prerequisite.

rhysyngsun:
@pdpinch @roberthouse54 where were we expecting this to live url-wise? /learners is where the current staff search lives.

pdpinch:
How much trouble do we get into if this is at the same URL?

rhysyngsun:
@pdpinch @roberthouse54 I'm nearing finalization of my PR for this and just noticed that the mockup and the list of facets we want to show differ. Which is the correct list?

[...]

PR #2846, opened 2017-03-10
Enabled learner-to-learner emails

### LINKED ISSUE #2196: Allow users to send email to a single user from the user chip, or the profile page.

gsidebo:
Closes #2196

#### What's this PR do?

- Adds the back-end functionality that will allow a paid learner to email another learner
- Replaces the email display on the profile page with an email link, which for now will only be seen by staff users

With a staff user, visit the profile of a user that has an email address and `Profile.email_optin=True`. Click the email link and make sure you can send a message. You can also use the learner mail view to send a learner-to-learner message since the link isn't viewable for other learners (yet). I'll post a code snippet to try out that email in the comments

`mail/permissions.py` and `static/js/components/LearnerInfoCard.js`

The email link is being hidden from other learners for now. We need to wait on the learner's search for learners issue (https://github.com/mitodl/micromasters/issues/2512) before we can show it. I will create a new issue to cover this work and update this description when that's done

ss 2017-03-10 at 12 59 35

### DIFF
--- a/mail/permissions.py
+++ b/mail/permissions.py
@@ -5,6 +5,7 @@ Permission classes for mail views
 from rolepermissions.verifications import has_permission
 from rest_framework.permissions import BasePermission
 
+from courses.models import CourseRun
 from roles.roles import Permissions, Staff, Instructor
 from dashboard.models import ProgramEnrollment
 from dashboard.utils import MMTrack
@@ -38,15 +39,43 @@ class UserCanMessageSpecificLearnerPermission(BasePermission):
         Returns:
             boolean
         """
-        sender_user = request.user
-        if not has_permission(sender_user, Permissions.CAN_MESSAGE_LEARNERS):
+        if not obj.email_optin:
             return False
-        # Check if the sender is staff/instructor in a program that the recipient
-        # is enrolled in.
-        return sender_user.role_set.filter(
-            role__in=(Staff.ROLE_ID, Instructor.ROLE_ID),
-            program__programenrollment__user=obj.user,
-        ).exists()
+
+        sender_user = request.user
+        recipient_enrolled_program_ids = obj.user.programenrollment_set.values_list('program', flat=True)
+
+        # If the sender is a staff/instructor in any of the recipients enrolled programs, the
+        # sender has permission
+        if sender_user.role_set.filter(
+                role__in=[Staff.ROLE_ID, Instructor.ROLE_ID],
+                program__id__in=recipient_enrolled_program_ids
+        ).exists():
+            return True
+
+        # If the sender has paid for any course run in any of the recipient's enrolled programs, the
+        # sender has permission
+        matching_program_enrollments = (
+            sender_user.programenrollment_set
+            .filter(program__id__in=recipient_enrolled_program_ids)
+            .select_related('program').all()
+        )
+        edx_user_data = CachedEdxUserData(sender_user)
+        for program_enrollment in matching_program_enrollments:
+            mmtrack = MMTrack(
+                sender_user,
+                program_enrollment.program,
+                edx_user_data
+            )
+            course_run_keys = (
+                CourseRun.objects
+                .filter(course__program=program_enrollment.program)
+                .values_list('edx_course_key', flat=True)
+            )
+            if any(mmtrack.has_paid(course_run_key) for course_run_key in course_run_keys):
+                return True
+
+        return False
 
 
 class UserCanMessageCourseTeamPermission(BasePermission):
--- a/static/js/components/LearnerChip_test.js
+++ b/static/js/components/LearnerChip_test.js
@@ -10,7 +10,7 @@ import { getPreferredName } from '../util/util';
 import ProfileImage from '../containers/ProfileImage';
 
 describe('LearnerChip', () => {
-  let profileClone;
+  let profileClone, sandbox;
 
   const renderChip = (profile, openLearnerEmailComposer) => (
     shallow(<LearnerChip profile={profile} openLearnerEmailComposer={openLearnerEmailComposer} />)
@@ -18,6 +18,11 @@ describe('LearnerChip', () => {
 
   beforeEach(() => {
     profileClone = R.clone(USER_PROFILE_RESPONSE);
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
   });
 
   it('should show the preferred name', () => {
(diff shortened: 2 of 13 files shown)


gsidebo:
Here's how you can test learner-to-learner emails:

```python
from grades.models import FinalGrader
from grades.constants import FinalGradeStatus
from seed_data.lib import CachedEnrollmentHandle
from django.test.client import Client

def set_user_as_paid_in_program(user, program):
    program_course_run = CourseRun.objects.filter(course__program=program).order_by('course__title', '-start_date').first()
    CachedEnrollmentHandler(user).set_or_create(program_course_run, verified=True)
    final_grade = FinalGrade.objects.create(
        user=user,
        course_run=program_course_run,
        status=FinalGradeStatus.COMPLETE,
        course_run_paid_on_edx=True,
        grade=0.75,
        passed=True
    )

sender_user = User.objects.get(username=YOUR_SENDER_USERNAME)
recipient_user = User.objects.get(username=YOUR_RECIPIENT_USERNAME)
client = Client()
client.force_login(sender_user)
url = reverse('learner_mail_api', kwargs={'student_id': recipient_user.profile.student_id})

# Should result in 403 because sender has no paid course in a program that the 
# recipient is enrolled in
client.post(url, data={'email_subject': 'the subject', 'email_body': 'the body'}, format='json')

# Should result in 200
set_user_as_paid_in_program(sender_user, recipient_user.programenrollment_set.first().program)
client.post(url, data={'email_subject': 'the subject', 'email_body': 'the body'}, format='json')
```

noisecapella:
Should the recipient's `account_privacy` value affect the ability of learners to message them?

gsidebo:
@noisecapella I think the `email_optin` flag is sufficient to determine whether or not a user can be messaged. I guess we can theoretically have a user that has `account_privacy=False` and `email_optin=True`. That's a little odd, but that still means they chose the option to allow emails to be sent to them, so I would say that messages should be allowed

static/js/components/LearnerChip_test.js
```diff
@@ -45,12 +45,21 @@ describe('LearnerChip', () => {
   it('should provide a link to email the learner', () => {
     let sandbox = sinon.sandbox.create();
```
noisecapella: Can you remove this too?

static/js/components/LearnerInfoCard.js
```diff
@@ -18,50 +21,83 @@ export default class LearnerInfoCard extends React.Component {
...
   };
 
-  email = (email: string): React$Element<*> => (
-    <span className="profile-email">{email}</span>
-  );
+  isOwnProfilePage: boolean;
+
+  constructor(props: Object) {
+    super(props);
+    this.isOwnProfilePage = isProfileOfLoggedinUser(props.profile);
```
noisecapella: We shouldn't be holding onto props like this. Can we make a function to calculate this value instead?

static/js/components/LearnerInfoCard_test.js
```diff
@@ -97,14 +98,72 @@ describe('LearnerInfoCard', () => {
...
+        profile: {
+          ...USER_PROFILE_RESPONSE,
+          email: 'learner@example.com',
+          email_optin: false
+        }
+      });
+      assert.lengthOf(card.find(".icon-button-link"), 0);
+    });
+
+    it('should not be shown if the user is has no email address', () => {
```
noisecapella: Can you remove the `is` in this line?

mail/permissions.py
```diff
@@ -38,15 +39,43 @@ def has_object_permission(self, request, view, obj):
...
+        # If the sender is a staff/instructor in any of the recipients enrolled programs, the
+        # sender has permission
+        if sender_user.role_set.filter(
+                role__in=[Staff.ROLE_ID, Instructor.ROLE_ID],
+                program__id__in=recipient_enrolled_program_ids
+        ).exists():
+            return True
+
+        # If the sender has paid for any course run in any of the recipient's enrolled programs, the
+        # sender has permission
```
noisecapella: Nevermind this one, I commented on it earlier

mail/permissions.py
```diff
@@ -38,15 +39,43 @@ def has_object_permission(self, request, view, obj):
         Returns:
             boolean
         """
-        sender_user = request.user
-        if not has_permission(sender_user, Permissions.CAN_MESSAGE_LEARNERS):
+        if not obj.email_optin:
```
gsidebo: The options for email optin on the form are `"Faculty, staff, and other learners can send me emails"` and `"I don't want to receive any emails"`

static/js/components/LearnerInfoCard.js
```diff
@@ -18,50 +22,85 @@ export default class LearnerInfoCard extends React.Component {
...
-    <span className="profile-email">{email}</span>
-  );
+  ownProfilePage: boolean;
+
+  isOwnProfilePage = (): boolean => {
+    if (R.isNil(this.ownProfilePage)) {
+      this.ownProfilePage = isProfileOfLoggedinUser(this.props.profile);
+    }
+    return this.ownProfilePage;
+  };
```
gsidebo: @noisecapella i'm kinda in between on this. not the cleanest pattern, but it avoids the unnecessary repeated calculation of the same conditions

noisecapella: The repeated calculations are necessary though. React components should always assume their props can change at any moment. This calculation is not time consuming, it's just comparing the username of the profile and for the `SETTINGS.user`

gsidebo: OK, just fixed this to use repeated calculations. I know there has to be a solution that cleanly eliminates the need for recalculating this value (and others I've seen around the codebase), but I don't really know enough about the React lifecycle yet, and as you said it's very cheap in this case

noisecapella: I think in general if you wanted to avoid extra computationally expensive calculation you would store the calculated value in state and look it up via some key in the props. There's also `shouldComponentUpdate` which can be used to skip rerenders if it's obvious they're not needed (for example if your props were entirely immutable and comparing the references would be enough to ensure equality)

PR #2858, opened 2017-03-13
only show the user a link to the financial aid review page if they have the permission

### LINKED ISSUE #2852: fix financial aid review thing
The financial aid review link the sidebar nav should only link to programs with Financial Aid. If the user does not have the staff role on a program that offers financial aid, we shouldn't display the link. 

Since there is only one program with financial aid for the forseeable future, we don't need an interface yet for the user to choose which program they want to see.

aliceriot:

closes #2852

#### What's this PR do?

makes it so we only show the financial aid review page link to users with the appropriate permission.

make sure you have permissions for financial aid on a program. go anywhere in the site, and a link to the corresponding financial aid review page should show up.

get rid of that permission (or sign into a different account) and you shouldn't see the link any longer.

### DIFF
--- a/static/js/lib/roles_test.js
+++ b/static/js/lib/roles_test.js
@@ -4,6 +4,8 @@ import { assert } from 'chai';
 import {
   hasAnyStaffRole,
   hasStaffForProgram,
+  hasEditAbility,
+  firstFinancialAidProgram,
 } from './roles';
 
 describe('roles library', () => {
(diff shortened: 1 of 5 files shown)


static/js/lib/roles_test.js
```diff
@@ -42,4 +46,30 @@ describe('roles library', () => {
...
   });
+
+  describe('hasEditAbility', () => {
+    it('should return false if user does not have the permission', () => {
+      roles.forEach(role => {
+        assert.isFalse(hasEditAbility(role));
+      });
+    });
+
+    it('should return true if user has permission on any program', () => {
```
noisecapella: Can you add a test that a different permission won't work for `hasEditAbility`?

aliceriot: sure!

PR #2859, opened 2017-03-13
Added validation for invalid name chars (#2837)

### LINKED ISSUE #2837: Exams: Don't allow certain characters in first name / last name
These characters cause issues w/ Pearson. Short-term fix is to disallow them, we may revisit. Characters are: `>`,  `,` and `"`.

rhysyngsun:
Fixes #2837 

#### What's this PR do?
Adds validation to ensure `<`, `,` and `"` aren't allowed in names.

Try inputting the above chars into the romanized / non-romanized name fields

`static/js/lib/validation/profile.js`

### DIFF
--- a/static/js/constants.js
+++ b/static/js/constants.js
@@ -12,6 +12,8 @@ export const DASHBOARD_FORMAT = 'MMM D, Y';
 export const DASHBOARD_MONTH_FORMAT = 'MM[/]YYYY';
 
 export const CP1252_REGEX = /^[\u0020-\u00FF]*$/;
+// greater-than, comma, quotation-mark
+export const INVALID_NAME_CHARS_REGEX = /[\u003E\u002C\u0022]/;
 
 // NOTE: this is in order of attainment
 export const EDUCATION_LEVELS = [
--- a/static/js/lib/validation/profile.js
+++ b/static/js/lib/validation/profile.js
@@ -21,7 +21,8 @@ import {
   PERSONAL_STEP,
   EDUCATION_STEP,
   EMPLOYMENT_STEP,
-  CP1252_REGEX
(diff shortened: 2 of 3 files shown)


static/js/constants.js
```diff
@@ -12,6 +12,8 @@ export const DASHBOARD_FORMAT = 'MMM D, Y';
 export const DASHBOARD_MONTH_FORMAT = 'MM[/]YYYY';
 
 export const CP1252_REGEX = /^[\u0020-\u00FF]*$/;
+// greater-than, comma, quotation-mark
```
amir-qayyum-khan: can we use `closing angle bracket` instead of `greater-than`. Because this is not related to programming,
 it is a text.

static/js/constants.js
```diff
@@ -12,6 +12,8 @@ export const DASHBOARD_FORMAT = 'MMM D, Y';
...
+export const INVALID_NAME_CHAR_REGEX = /[\u003E\u002C\u0022]/;
```
amir-qayyum-khan: what about  `opening angular bracket <`. It allowed me to use it.

screen shot 2017-03-15 at 3 10 56 pm

amir-qayyum-khan: why not `INVALID_NAME_CHARS_REGEX` because it is list if chars.

rhysyngsun: only the characters listed in the original ticket are a concern

static/js/constants.js
rhysyngsun: It's a code comment which is programmer-facing so I'm inclined to leave it as-is because it's more succinct.

amir-qayyum-khan:
👍

PR #2860, opened 2017-03-13
Blarghmatey/circleci

blarghmatey:
mitodl/saltops#246

#### What's this PR do?
Adds configuration for running our tests on CircleCI using their 2.0 beta system.

Review the build log at https://circleci.com/gh/mitodl/micromasters/134

.circleci/config.yml

We have been having issues with slow builds on Travis and want to move to a system that allows us to pay for more workers.

Of note in this configuration is that they do have a docker executor, but there are some limitations in terms of how tests get executed vs our current approach to running the tests. It may be worth considering an update to our compiled containers to support running with the docker executor because it can significantly reduce the amount of overhead for test startup. Documentation can be found here: https://circleci.com/docs/2.0/executor-types/ and here: https://circleci.com/docs/2.0/building-docker-images/

(Optional)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,31 +1,31 @@
 version: 2
 jobs:
   build:
-    working_directory: /src
-    docker:
-      - image: mitodl/mm_web_travis
-        environment:
-          DEBUG: 'False'
-          COVERAGE_DIR: htmlcov
-          PORT: 8079
-          NODE_ENV: 'production'
-          DATABASE_URL: postgres://postgres@localhost:5432/postgres
-          MICROMASTERS_SECURE_SSL_REDIRECT: 'False'
-          MICROMASTERS_DB_DISABLE_SSL: 'True'
-          ELASTICSEARCH_URL: localhost:9200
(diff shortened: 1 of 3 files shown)


aliceriot:
does circleCI have any support for parallel jobs? running all the tests in order is going to be a pretty lengthy build - currently they run in three separate jobs.

aliceriot:
do you have an idea what kind of changes we would need to make in order to use their built-in docker support? I read through the documentation quickly and I didn't see anything that would prevent our current setup from working, but I did not read it super deeply so I'm not sure if I missed anything.

blarghmatey:
Unfortunately they don't have any means of parallelizing the build steps. In terms of the docker support, the issue is that, while you can specify multiple containers to be run, all of the build steps only happen in the first listed container. CircleCI also doesn't have any support for specifying different build configurations within a single application.

I've started trying again to get CodeShip to work but that is also proving to be quite frustrating.

blarghmatey:
There doesn't seem to be much interest in CircleCI so I am going to close this.

Issue #2854, opened 2017-03-13
Exams: schedule button cut off when course title too long


pdpinch:
The super-long title seems to be expanding the list container, which squeezes the button enough that it loses the last word in "Schedule an Exam"

<img width="639" alt="screenshot 2017-03-13 11 11 09" src="https://cloud.githubusercontent.com/assets/430126/23867132/5247a214-07f1-11e7-9997-09520c185328.png">

PR #2863, opened 2017-03-13
Fixes layout issue with schedule an exam button

roberthouse54:
#2854

#### What's this PR do?
This fixes a small issue with the layout of the button in the Pearson Exam card. The button label was getting cut off when there were very long course names. This adds a min-width on a button.

Get into the state where you can schedule a Pearson exam. Make the course name different character lengths. Make sure the button text always is readable as "SCHEDULE AN EXAM".

This also fixes a couple very minor margin issues with the list item.

**Before**
image

**After**
image


aliceriot:
haha 'schedule an'

facepalm

aliceriot:
layout looks great!

static/scss/_variables.scss
```diff
@@ -37,7 +37,7 @@ $select-menu-bg: rgba(0,0,0,.1);
 $darkgrey: #696969;
 $medlightgrey: #8d8d8d;
 $medgrey: #d2d2d2;
-$lightgrey: #eee;
+$lightgrey: #f1f1f1;
```
aliceriot: these are just the same color, right?

PR #2864, opened 2017-03-13
Updated required yarn version in readme file

amir-qayyum-khan:
#### What's this PR do?
Update yarn version that needed for MM development

@pdpinch @aliceriot


aliceriot:
ah, forgot about that! for future reference, the best place to check for the current version of yarn is in the `engines` field in `package.json`.

Issue #2181, opened 2016-12-21
Degrade gracefully when edX is down


noisecapella:
Currently if edX is down and the user needs to refresh some data from the cache, we show an error on the whole screen. Instead, we should log the error and show the old data from the cache, if there is any. I'm not sure what to do about the case where a user first logs in and has no cached data, and edX is also down all of a sudden after being available for login.

noisecapella:
![screenshot from 2016-12-21 10-37-03](https://cloud.githubusercontent.com/assets/863262/21395159/739a5858-c769-11e6-924e-64319ddd4591.png)

pdpinch:
Would this address https://sentry.io/mit-office-of-digital-learning/micromasters/issues/193330670/

noisecapella:
Yes, at least from the user experience perspective. We should still log these events in case they're useful, so we know if we are causing too much API activity with edX

giocalitri:
I think this is a super set of some other issues already filed.

In general I think that this is easily accomplishable from a technical point of view (we already have the cached data) but we need a bit of design from @roberthouse54 to solve the problem of how to tell the user that what she is seeing might not be up to date.

roberthouse54:
How about this:

"Sorry, the edX server seems to be down, so some of your information may not be up to date. Please refresh the browser, or check back later. If the error persists, contact support@example.com for help."

![image](https://cloud.githubusercontent.com/assets/20047260/21647847/e1253d60-d269-11e6-8e4a-073c723dd7a0.png)

Ferdi:
Maybe "edx.org is not responding as expected"  or something like that. Because user might see this error and edx.org load fines for them

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/193925129/

roberthouse54:
OK, let's change the error text to say: "Sorry, edx.org is not responding as expected, so some of the information on this page may not be up to date. Please refresh the browser, or check back later. If the error persists, contact support@example.com for help."

PR #2872, opened 2017-03-15
Modified Dashboard REST API to include edx data freshness status

giocalitri:
first part of #2181 

#### What's this PR do?
Changes tha data structure of the dashboard rest api and the JS code that uses it.
It also adds an extra key to the dashboard rest api response to specify if the edX cached data is fresh.
The dashboard REST API should not fail even if the backend edX instance is not available.
This is necessary to implement the message in the dashboard to specify if the cache has not been refreshed.

The dashboard should work as usual.
if you shut down the edX backend instance, the dashboard should still load.

`dashboard/views.py`
`dashboard/api.py`
`static/js/reducers/ashboard.js`

Most of the changes are in the JS tests because of the change in the structure of the response JSON coming from the REST API

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -9,6 +9,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db import transaction
 import pytz
 
+from backends.exceptions import InvalidCredentialStored
 from courses.models import Program
 from dashboard.api_edx_cache import CachedEdxDataApi, CachedEdxUserData
 from dashboard.utils import MMTrack
@@ -127,12 +128,21 @@ def get_user_program_info(user, edx_client):
     # update cache
     # NOTE: this part can be moved to an asynchronous task
     if edx_client is not None:
-        for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
-            CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
+        try:
+            for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
+                CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
+        except InvalidCredentialStored:
+            # this needs to raise in order to force the user re-login
+            raise
+        except:  # pylint: disable=bare-except
+            log.exception('Impossible to refresh edX cache')
 
     edx_user_data = CachedEdxUserData(user)
 
-    response_data = []
+    response_data = {
+        "programs": [],
+        "is_edx_data_fresh": CachedEdxDataApi.are_all_caches_fresh(user)
+    }
     all_programs = (
         Program.objects.filter(live=True, programenrollment__user=user).prefetch_related('course_set__courserun_set')
     )
@@ -142,7 +152,7 @@ def get_user_program_info(user, edx_client):
             program,
             edx_user_data
         )
-        response_data.append(get_info_for_program(mmtrack_info))
+        response_data['programs'].append(get_info_for_program(mmtrack_info))
     return response_data
 
 
--- a/dashboard/api_edx_cache_test.py
+++ b/dashboard/api_edx_cache_test.py
@@ -240,6 +240,27 @@ class CachedEdxDataApiTests(MockedESTestCase):
         for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
             assert CachedEdxDataApi.is_cache_fresh(self.user, cache_type) is False
 
+    @ddt.data('certificate', 'enrollment', 'current_grade')
+    def test_are_all_caches_fresh(self, cache_type):
+        """Test for are_all_caches_fresh"""
+        assert UserCacheRefreshTime.objects.filter(user=self.user).exists() is False
+        assert CachedEdxDataApi.are_all_caches_fresh(self.user) is False
+        now = datetime.now(tz=pytz.UTC)
+        yesterday = now - timedelta(days=1)
+        user_cache = UserCacheRefreshTimeFactory.create(
+            user=self.user,
+            enrollment=now,
+            certificate=now,
+            current_grade=now,
+        )
+        assert CachedEdxDataApi.are_all_caches_fresh(self.user) is True
+        setattr(user_cache, cache_type, yesterday)
+        user_cache.save()
+        assert CachedEdxDataApi.are_all_caches_fresh(self.user) is False
+        setattr(user_cache, cache_type, now)
+        user_cache.save()
+        assert CachedEdxDataApi.are_all_caches_fresh(self.user) is True
+
     @patch('search.tasks.index_users', autospec=True)
     def test_update_cached_enrollment(self, mocked_index):
         """Test for update_cached_enrollment"""
--- a/dashboard/api_test.py
+++ b/dashboard/api_test.py
@@ -13,9 +13,9 @@ import ddt
 import pytz
 from django.core.exceptions import ImproperlyConfigured
 from django.test import override_settings
+from rest_framework import status as http_status
 
-from exams.models import ExamProfile
-
+from backends.exceptions import InvalidCredentialStored
 from courses.factories import (
     CourseFactory,
     CourseRunFactory,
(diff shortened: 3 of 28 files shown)


static/js/flow/dashboardTypes.js
```diff
@@ -3,11 +3,15 @@ import Decimal from 'decimal.js-light';
 import type { Program } from './programTypes';
 
 // likely to change in very near future
-export type Dashboard = Array<Program>;
+export type Dashboard = {
+  programs:   Array<Program>,
+  is_edx_data_fresh: boolean
```
aliceriot: can we use camelCase here?

static/js/flow/dashboardTypes.js
```diff
@@ -3,11 +3,15 @@ import Decimal from 'decimal.js-light';
...
+};
 
 export type DashboardState = {
-  programs:     Dashboard,
-  fetchStatus?: string,
+  programs:       Array<Program>,
+  isEdxDataFresh: boolean,
+  fetchStatus?:   string,
```
aliceriot: you can express this state using the intersection operator:

```js
export type DashboardState = Dashboard & {
  fetchStatus?: string;
};
```

just removes a tiny amount of duplication.

static/js/reducers/dashboard_test.js
```diff
@@ -139,7 +141,8 @@ describe('dashboard reducers', () => {
           [_username]: {
             fetchStatus: FETCH_FAILURE,
             errorInfo: 'err',
-            programs: []
+            programs: [],
+            "isEdxDataFresh": true
```
aliceriot: you don't need quotes here

aliceriot:
I'm happy to look at the JS side of things, but I'm probably not so qualified to review the Python stuff!

static/js/components/dashboard/CourseRow_test.js
```diff
@@ -60,8 +60,8 @@ describe('CourseRow', () => {
   };
 
   it('forwards the appropriate props', () => {
-    const programs = makeDashboard();
-    const course = programs[0].courses[0];
+    const dashboard = makeDashboard();
+    const course = dashboard.programs[0].courses[0];
```
aliceriot: here you should be able to just do

```js
const { programs } = makeDashboard();
```

then line 64 can stay as it was

giocalitri:
the python part is basically a very small change: I can explain to you what it does ant if you still think it is better that someone else looks into it, we can ask for a second review.

static/js/flow/dashboardTypes.js
giocalitri: I think this is the type of the object coming from the backend and that is not camelCase

static/js/flow/dashboardTypes.js
giocalitri: I suppose if I do not make the previous one camelCase, I cannot do this, right?

static/js/reducers/dashboard_test.js
giocalitri: ok

static/js/components/dashboard/CourseRow_test.js
giocalitri: ok

static/js/flow/dashboardTypes.js
aliceriot: Oh yes, you're right - I forgot that we had changed this around relatively recently. Carry on!

static/js/flow/dashboardTypes.js
aliceriot: yes, that's right

dashboard/api_edx_cache_test.py
```diff
@@ -240,6 +240,23 @@ def test_is_cache_fresh(self):
         for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
             assert CachedEdxDataApi.is_cache_fresh(self.user, cache_type) is False
 
+    def test_are_all_caches_fresh(self):
```
noisecapella: Can we use ddt here to test each of the three cache types?

dashboard/api_test.py
```diff
@@ -1254,15 +1254,19 @@ def test_format(self, mock_cache_refresh):
         for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
             mock_cache_refresh.assert_any_call(self.user, self.edx_client, cache_type)
 
-        assert len(result) == 2
+        assert isinstance(result, dict)
```
noisecapella: Feel free to leave as is but I'm not sure what the value is for this and the next lines since we would error on `assert result['is_edx_data_fresh'] is False` anyway

dashboard/views.py
```diff
@@ -61,6 +61,8 @@ def get(self, request, username, *args, **kargs):  # pylint: disable=unused-argu
                     status=exc.http_status_code,
                     data={'error': str(exc)}
                 )
+            except Exception:  # pylint: disable=broad-except
```
noisecapella: Can we use `except:` here too? Also wouldn't we want this to be re-raised? What kind of exceptions are we expecting here?

dashboard/api_test.py
```diff
@@ -1328,15 +1332,35 @@ def test_past_course_runs(self):
...
+        assert len(result['programs'][0]['courses']) > 0
+        assert len(result['programs'][0]['courses'][0]['runs']) == 2
+        assert all(
+            [run['status'] == api.CourseStatus.NOT_PASSED for run in result['programs'][0]['courses'][0]['runs']]
+        )
+
+    @patch('dashboard.api_edx_cache.CachedEdxDataApi.update_cache_if_expired', new_callable=MagicMock)
+    def test_exception_in_refresh_cache_1(self, mock_cache_refresh):
+        """Test in case the backend refresh cache raises a InvalidCredentialStored exception"""
+        mock_cache_refresh.side_effect = InvalidCredentialStored('error', 400)
```
noisecapella: Can you replace `400` with one of the HTTP constants?

dashboard/api_test.py
```diff
@@ -1328,15 +1332,35 @@ def test_past_course_runs(self):
...
-        assert len(result[0]['courses'][0]['runs']) == 2
-        assert all([run['status'] == api.CourseStatus.NOT_PASSED for run in result[0]['courses'][0]['runs']])
+        assert len(result['programs']) > 0
+        assert len(result['programs'][0]['courses']) > 0
+        assert len(result['programs'][0]['courses'][0]['runs']) == 2
+        assert all(
+            [run['status'] == api.CourseStatus.NOT_PASSED for run in result['programs'][0]['courses'][0]['runs']]
+        )
+
+    @patch('dashboard.api_edx_cache.CachedEdxDataApi.update_cache_if_expired', new_callable=MagicMock)
```
noisecapella: Is `autospec=True` useful here?

giocalitri: you cannot use `new_callable` and `autospec=True` at the same time

dashboard/api_edx_cache_test.py
giocalitri: ok

dashboard/api_test.py
giocalitri: I usually try to check the structure of the dictionary in the tests so if it fails because it is different I know why and I have not infer from the error.
I have no strong opinions about this either.

dashboard/api.py
```diff
@@ -127,12 +128,21 @@ def get_user_program_info(user, edx_client):
...
     if edx_client is not None:
-        for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
-            CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
+        try:
+            for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
+                CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
+        except InvalidCredentialStored:
+            # this needs to raise in order to force the user re-login
+            raise
+        except Exception:  # pylint: disable=broad-except
```
giocalitri: ok

dashboard/api_test.py

dashboard/views.py
giocalitri: In case edX is truly down, the python social auth is rising a custom exception (something related to not being able to establish a connection).
In any case if the refresh of the token fails for any reason that is not a 400/401 we need to not raise if we want to keep serving the dashboard.

giocalitri:

noisecapella:
Code looks fine to me, I haven't done a functionality test though since I assume @aliceriot is doing one

aliceriot:
functionality looks good! It looks like there are a few python linting errors though.

giocalitri:
@aliceriot it is passing now

aliceriot:
then it's a 👍 from me! @noisecapella do you think this is good to go as well?

noisecapella:
Yes :shipit:

PR #2873, opened 2017-03-15
CMS: Added Semester Start Dates

### LINKED ISSUE #2808: Add upcoming semester start dates to program marketing pages
Marketing and DEDP have requested that we add information about start dates for upcoming semesters. We will show start dates for upcoming semesters in the right column in a box called Future Course Dates. This will require a change to wagtail, adding a section where users can enter upcoming semester start dates. Semesters will be labelled Spring, Summer, and Fall. (eg. "Summer 2017").

If no semester start dates are entered in wagtail, do not show this box on the marketing site Program page.

image

image

annagav:
Fix #2808 

#### What's this PR do?
Allow staff to add future semester dates to program page.

Login as staff to /cms/. Start editing a program page, add few semester dates, click 'publish' and check the program page to see the changes.

(Optional)

The ui design in wagtail is a bit different. Originally I was trying to implement a read only field for semester title inside `FieldRowPanel`, but was not able to easily nest it in `InlinePanel` which allows the creation of a “cluster” of related objects to ProgramPage model.
Rob has approved this change.

screen shot 2017-03-15 at 1 09 07 pm

screen shot 2017-03-15 at 1 13 14 pm

(Optional)


cms/models.py
```diff
@@ -257,6 +258,27 @@ class InfoLinks(Orderable):
     ]
 
 
+class SemesterDate(Orderable):
+    """
+    Dates for future start dates of semesters
+    """
+    program_page = ParentalKey(ProgramPage, related_name='semester_dates')
+    semester = models.CharField(
```
singingwolfboy: I'd suggest calling this `semester_name`, to make it clearer that it's a name.

cms/models.py
```diff
@@ -257,6 +258,27 @@ class InfoLinks(Orderable):
...
+        max_length=50,
+        help_text='Name for the semester, example: Fall or Fall 2017'
```
singingwolfboy: Nitpick: can we make this help text always refer to a future date? I'd suggest doing the following:

```python
from datetime import datetime
now = datetime.now()

class SemesterDate(Orderable):
    ...
    semester_name = models.CharField(
        max_length=50,
        help_text='Name for the semester. For example: "Fall" or "Fall {year}"'.format(
            year=now.year + 1
        )
    )
```

singingwolfboy:
Question: right now, this pull request requires that the instructor set a defined start date for a future semester. What if the instructor doesn't know the defined date? For example, maybe the course is planned to start in Fall 2018, but the exact start date hasn't been determined? Perhaps the `start_date` field should be optional, so that the instructor can simply put in the name.

@pdpinch, @roberthouse54, any thoughts on this?

roberthouse54:
My inclination is to not make the date optional, because the entire thing is optional. If you do not know a date, you just do not enter it. If no dates are entered, the box is not supposed to show up at all. 

If someone strongly disagrees, we can discuss.

singingwolfboy:
👍 after tests pass.

annagav:
@singingwolfboy, I addressed your comments.

PR #2880, opened 2017-03-15
Fixed lifecycle handling of DashboardPage to reload cleared items

### LINKED ISSUE #2878: Error when clicking back button on order summary page
screenshot from 2017-03-15 14-32-48

noisecapella:
Fixes #2878 

#### What's this PR do?
Fixes lifecycle handling in DashboardPage so the back button will work as expected

Click 'Pay now' and it should direct you to the order summary page. Press the back button and you should be back at the dashboard without an error.

### DIFF
--- a/static/js/dashboard_routes.js
+++ b/static/js/dashboard_routes.js
@@ -13,6 +13,7 @@ import Learner from './components/Learner';
 import OrderSummaryPage from './containers/OrderSummaryPage';
 
 const errorLoading = error => {
+  console.trace(error);
   throw new Error(`Dynamic page loading failed: ${error}`);
 };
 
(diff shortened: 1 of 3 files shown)


static/js/dashboard_routes.js
```diff
@@ -13,6 +13,7 @@ import Learner from './components/Learner';
 import OrderSummaryPage from './containers/OrderSummaryPage';
 
 const errorLoading = error => {
+  console.trace(error);
```
noisecapella: This is intentional, it shows a better stack trace of the error in Chrome

aliceriot: 💯

PR #2882, opened 2017-03-15
Fixed alter_data commands to work with FA programs and added states

### LINKED ISSUE #2121: Expand 'alter data' commands to work with non-FA programs
Our 'alter data' commands (seed_data/management/commands/alter_data.py) give us the ability to make fine-grained changes in a user's enrollments/grades. This helps us manually test the app, especially in non-local environments. To ease the development effort, these commands were designed to work only with non-FA programs. At some point we should change this to support FA programs. 

This might also be a good opportunity to centralize the logic for creating enrollment/grade data for FA and non-FA programs. That logic is currently repeated several times over in our tests.

gsidebo:
Closes #2121 

#### What's this PR do?

- Makes `alter_data` management commands work with FA-enabled programs
- Uses `FinalGrade` for final grade information rather than cached edX data
- Adds a command to set a course run to paid-but-not-enrolled, and another command to clear all dashboard data (final grade data, cached edX data, payment data) for a course/course run
- Some general code cleanup

Run some of these commands with both FA and non-FA programs. Example commands can be found by running `./manage.py alter_data examples`, and valid parameters can be found by running `./manage.py alter_data --help`

`alter_data.py`


micromasters/utils.py
```diff
@@ -118,6 +118,16 @@ def first_matching_item(iterable, predicate):
     return next(filter(predicate, iterable), None)
 
 
+def remove_falsey_values(iterable):
```
gsidebo: It's just a clearly explained shorthand. `filter(None, iterable)` does the same thing, but I would argue it's not a commonly known trick that `filter` can accept `None` instead of a function as its first arg. `remove_falsey_values` is pretty explicit

seed_data/lib.py
```diff
@@ -284,27 +266,45 @@ def ensure_cached_data_freshness(user):
...
+    """
+    program = None
+    course_run_params = {}
+    if course:
+        program = course.program
+        course_run_params['course'] = course
+    elif course_run:
+        program = course_run.course.program
+        course_run_params['id'] = course_run.id
+    if program.financial_aid_availability:
```
gsidebo: This is intentional. It should be a no-op for FA programs

gsidebo:
@singingwolfboy Just pushed some fixes. I'll look into the ORM thing since it would be pretty useful to know in the future, but in this context (tiny amount of possible records in a non-critical part of the codebase) I would say we can accept it as-is if I don't find any good alternative

seed_data/management/commands/seed_db.py
```diff
@@ -161,10 +151,14 @@ def deserialize_grade_data(user, social_username, course_runs, grade_data_list):
             lambda cr: cr.edx_course_key == edx_course_key
         )
         grade = Decimal(grade_data['grade'])
-        if course_run.course.program.financial_aid_availability:
-            cur_grade_handler.set_or_create(course_run, grade=grade)
-        else:
-            cert_handler.set_or_create(course_run, grade=grade)
+        FinalGrade.objects.create(
```
singingwolfboy: Should this also be `update_or_create()`?

gsidebo: fixed

seed_data/lib.py
singingwolfboy: Could you mention that in the function docstring? In my opinion, it's sufficiently unusual that it should be called out.

gsidebo: done

PR #2886, opened 2017-03-16
show average grade on the staff view of learner page

### LINKED ISSUE #2853: Display average program grade in staff view of learner's profile page
Part of #2480 

When a staff member views the learner profile of a learner in their program, they should be able to view the learner's average grade in the program -- this should be the same value as the one they can view in the learner search. 

enhanced_profile_faculty_view4 2x

aliceriot:

closes #2853 

#### What's this PR do?

If the user has a program average grade we'll show it on the staff view of the learner page.

Well, if you already have a program average grade it should show up on `/learner`. If you don't you should see `--` instead.

if you don't have a program average grade you can use my API mocking patch, here: https://github.com/aliceriot/mm_patches/blob/master/mock_api.patch

to get a dashboard (for 'Last Program') which includes the `grade_average` property. Or I suppose you could edit the relevant line in `dashboard/api.py` to just return a number.

### DIFF
--- a/static/js/components/StaffLearnerInfoCard.js
+++ b/static/js/components/StaffLearnerInfoCard.js
@@ -5,11 +5,12 @@ import R from 'ramda';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 
 import { circularProgressWidget } from './ProgressWidget';
-import { programCourseInfo } from '../util/util';
+import { programCourseInfo, classify } from '../util/util';
 import type { Program } from '../flow/programTypes';
 import CourseDescription from '../components/dashboard/CourseDescription';
 import CourseGrade from '../components/dashboard/CourseGrade';
 import { STATUS_OFFERED } from '../constants';
+import { S, getm } from '../lib/sanctuary';
 
 type StaffLearnerCardProps = {
   program: Program,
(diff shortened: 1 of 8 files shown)


static/js/components/StaffLearnerInfoCard.js
```diff
@@ -56,6 +57,16 @@ const displayCourseRuns = R.compose(
   renderCourseRuns, R.flatten, R.map(formatCourseRuns), R.prop('courses')
 );
 
+// getProgramProp :: String -> Program -> Either String Number
+const getProgramProp = (prop, program) => (
```
singingwolfboy: Nitpick: it would be nice if you could add Flow annotations here.

static/js/components/StaffLearnerInfoCard.js
```diff
@@ -56,6 +57,16 @@ const displayCourseRuns = R.compose(
...
+  S.maybeToEither('--', getm(prop, program))
+);
+
+// formatCourseGrade :: Program -> String
+const formatCourseGrade = program => (
```
singingwolfboy: Nitpick: it would be nice if you could add Flow annotations here.

static/js/lib/sanctuary.js
```diff
@@ -37,3 +39,8 @@ export const guard = (func: Function) => (...args: any) => {
     return S.Just(func(...args));
   }
 };
+
+// getm :: String -> Object -> Maybe a
+export const getm = R.curry((prop, obj) => (
```
singingwolfboy: Nitpick: it would be nice if you could add Flow annotations here.

Issue #2732, opened 2017-03-01
Remove "!important" CSS rules as much as possible


singingwolfboy:
It's possible to use `!important` on a CSS rule to override all other rules. This is **bad practice**, and should be avoided because it makes debugging CSS more difficult. [(see MDN reference)](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#The_!important_exception) It also makes it more difficult for user-specified styles to override website-specified styles, which can sometimes be important for accessibility purposes.

We should avoid `!important` whenever possible in our CSS. When it is not possible to avoid `!important`, we should add a comment justifying its use.

pdpinch:
I don't disagree with the sentiment, but the way this is phrased this is an issue that can never be closed. 

Do we have a lot of these? If its a manageable number, this issue should be to remove them all. If there are too many, we should break them off into bite-sized pieces. 

Should this just be limited to the issue in https://github.com/mitodl/micromasters/pull/2693#discussion_r103716869 and the code it was copied from?

singingwolfboy:
```
$ grep "\!important" static/scss/*.scss static/scss/public_style/*.css | wc -l
164
```

It's a lot -- more than I realized. Bite-sized pieces is probably a good idea.

PR #2892, opened 2017-03-16
Make Material Design Lite configurable

singingwolfboy:
This is a utility that future improvements can build upon. It is related to #2732, but doesn't have its own ticket.

#### What's this PR do?
[Material Design Lite](https://github.com/google/material-design-lite) is designed to be very configurable and flexible, by exposing many, well-organized [Sass variables that can be overridden](https://github.com/google/material-design-lite/blob/mdl-1.x/src/_variables.scss). Previously, we were importing a CSS file for MDL, which had already passed through a Sass processor. This pull request changes how we load MDL, so that we can override the variables and configure it to look and work the way we want, _without_ resorting to `!important`.

This pull request overrides one variable (`$color-primary`), just to verify that it does work properly. However, this override is currently hidden by the CSS cascade, so there should be no user-visible changes in this pull request.

On the dashboard, open the course enrollment dialog. Inspect one of the buttons in that dialog. You should see that the `.mdl-button.mdl-button--colored` CSS rule sets the `color` to `rgb(27,122,227)`, which matches the value set in the override file in this pull request. If you want, you can change the `$color-primary` variable to another color, let Sass recompile, and then inspect the element again. The CSS rule should match whatever color you define.

material-design-lite-override

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "jsdom-global": "^2.1.1",
     "keycode": "2.1.1",
     "lodash": "^4.17.4",
+    "material-design-lite": "^1.3.0",
     "material-ui": "^0.16.3",
     "minimist": "^1.2.0",
     "mocha": "^3.2.0",
--- a/static/js/entry/style.js
+++ b/static/js/entry/style.js
@@ -1,9 +1,11 @@
 __webpack_public_path__ = `${SETTINGS.public_path}`;  // eslint-disable-line no-undef, camelcase
+
+// material design lite (MDL)
+import '../../scss/material-design-lite.scss';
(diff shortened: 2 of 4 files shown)


noisecapella:
The reason we're using that MDL CSS file is because they applied a patch which is recommended with use of the library in their README: https://github.com/react-mdl/react-mdl/blob/master/extra/layout-patch.diff

singingwolfboy:
Ah, I wasn't aware of that. Thanks for pointing it out! In that case, I think it would be better to use the React-MDL CSS file, even though it's less configurable. I'll close this pull request.

PR #2901, opened 2017-03-17
Improve disabled UI buttons

### LINKED ISSUE #2900: Disabled buttons should have a consistent UI

singingwolfboy:
Fixes #2900 

#### What's this PR do?
Improves consistency and accessibility of disabled buttons in the application. Now all disabled buttons should have `$medgrey` background color, white text, and a "not allowed" cursor on hover.

Go to the dashboard for a financial aid program, before you've calculated your financial aid. Make sure that the disabled "pay now" buttons are consistent, as well as the disabled "Apply for Masters" button on the right side of the page. Check the screenshots to see what I mean.

screen shot 2017-03-17 at 12 49 32 pm
screen shot 2017-03-17 at 12 49 41 pm
screen shot 2017-03-17 at 12 49 47 pm

### DIFF
--- a/static/js/components/ProgressWidget.js
+++ b/static/js/components/ProgressWidget.js
@@ -73,7 +73,7 @@ export default class ProgressWidget extends React.Component {
         <p className="heading-paragraph">
           On completion, you can apply for
           the master’s degree program</p>
-           <Button className="progress-button disabled">
+           <Button className="progress-button" disabled>
              Apply for Master’s
            </Button>
       </Card>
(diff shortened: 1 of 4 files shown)


roberthouse54:
I think I forgot to mention this, but can we lose the drop shadow on disabled buttons? Also make sure that there is no change on hover.

singingwolfboy:
@roberthouse54 Done.

roberthouse54:
I notice in the _button.scss that there is another &[disabled] rule under .dashboard-button, that keeps the same background color but adjusts the alpha to .5. I think we can get rid of this also, right?

noisecapella:
Can you use `Fixes #2900` so that github will automatically close that issue when this merges?

PR #2903, opened 2017-03-17
Create thumbnails in Profile.save

### LINKED ISSUE #2750: Move profile image handling code out of view/serializer
Our code to create small and medium thumbnails for a user's profile image exists in `profiles.serializers.ProfileSerializer#update`. This makes it very difficult to programatically set a user's profile image. It should not be necessary for an HTTP request with a newly-uploaded image to be sent in order to create these thumbnails. We should be able to generate and save these thumbnails any time a user's profile image is saved in the `Profile` model itself.

noisecapella:
Fixes #2750 

#### What's this PR do?
Creates thumbnails in `save()` instead of in serializer. This will allow other code to update the image and create the related thumbnails automatically even when they don't use the serializer.

Upload a new image and verify that the image of your avatar is the new image and a medium sized version of it

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -2,12 +2,16 @@
 Models for user profile
 """
 import re
+from uuid import uuid4
 
 from django.contrib.auth.models import User
 from django.contrib.postgres.fields import JSONField
 from django.db import models, transaction
 
 from profiles.util import (
+    IMAGE_SMALL_MAX_DIMENSION,
+    IMAGE_MEDIUM_MAX_DIMENSION,
+    make_thumbnail,
     profile_image_upload_uri,
     profile_image_upload_uri_small,
     profile_image_upload_uri_medium,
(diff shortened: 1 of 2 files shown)


profiles/models.py
```diff
@@ -176,9 +180,16 @@ class Profile(models.Model):
     updated_on = models.DateTimeField(blank=True, null=True, auto_now=True)
 
     @transaction.atomic
-    def save(self, *args, **kwargs):
-        """Set the student_id number to the PK number"""
-        # first save the profile
+    def save(self, *args, update_thumbnails=False, **kwargs):
```
gsidebo: If I have a nitpick, I would rather the `update_thumbnails` kwarg be named `update_image` since we want to update the thumbnails any time the actual profile image is updated. Updating the thumbnails is an implementation detail that is handled by the profile `save` method. You can take or leave this one

noisecapella:
Ok, that should do it

PR #2906, opened 2017-03-17
Fix refresh loop on learner page

### LINKED ISSUE #2905: Dashboard API refresh loop on /learner page
It looks like this is caused by `FETCH_FAILURE` not being handled properly. Should have a fix PR momentarily

### LINKED ISSUE #2907: learners page keeps trying getting dashboard rest api even in case of 500 errors
this is the root cause of the number of [this exception](https://sentry.io/mit-office-of-digital-learning/micromasters/issues/236670197/) and t[his other exception](https://sentry.io/mit-office-of-digital-learning/micromasters/issues/235789042/).

the learner page, if receives a 500 error from the dashboard rest api, retries to request the same resource in a loop instead of just failing.

noisecapella:
Fixes #2905 
Fixes #2907

#### What's this PR do?
Fixes handling of `FETCH_FAILURE` to prevent a refresh loop on dashboard error

On master, in `lib/api.js` have `getDashboard` return `Promise.reject()` then load the learner page. It should cause the loop. Switch to this branch and repeat the change to `lib/api.js` and the failure should only happen once.

### DIFF
--- a/static/js/containers/LearnerPage.js
+++ b/static/js/containers/LearnerPage.js
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 import Loader from '../components/Loader';
 import R from 'ramda';
 
-import { FETCH_PROCESSING, FETCH_SUCCESS } from '../actions';
+import { FETCH_PROCESSING, FETCH_SUCCESS, FETCH_FAILURE } from '../actions';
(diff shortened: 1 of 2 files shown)


aliceriot:
oh jeez, I goofed on this one.

functionality looks great! looking at code real quick.

Issue #2890, opened 2017-03-16
Remove feature flag code related to the final grade algorithm


gsidebo:
_Part of #1856_ 

Final grades are now calculated via the new `FinalGrade` model. We initially rolled this out with a feature flag. As a result, there is a lot of Python logic that pivots around the value of a feature flag (`FINAL_GRADE_ALGORITHM`). Now that the final grade algorithm is settled and has been demonstrated to work, we can remove all of the logic that checks for that feature flag value and all of the logic associated with `v0` of the algorithm

PR #2909, opened 2017-03-17
Cleaned up MMTrack final grade code

gsidebo:
No ticket. This is being done as part of the larger effort to clean up the dashboard API and MMTrack

#### What's this PR do?

Generally refactors the final grade-related code in `MMTrack`. More specifically:

- Does away with the exception-raising behavior related to fetching a final grade for a user's course run.
- Refactors several methods to use single database queries where they previously looped through course run id's and executed a query for each one.
- Fixes inconsistent function naming.

`alter_data` can be used to set course runs to be passed/failed, and can then be confirmed on the dashboard page

`dashboard/utils.py`

While working on #2890, I noticed a few code smells in `dashboard.utils.MMTrack` related to final grades. Some of the issues:

- `extract_final_grade` was using Django's `get`, which throws `DoesNotExist` exceptions, and there was a lot of code for catching those exceptions. This didn't seem like the best approach since a course run not having an associated `FinalGrade` is not exceptional/unexpected. The abundance of code blocks that included `except FinalGrade.DoesNotExist: continue` and the existence of `MMTrack.has_passed_course_for_exam` was evidence of this. Getting rid of the exception-raising behavior eliminated a decent amount of code and is more readable (IMO)
- There were several functions that looped through a list of course run keys and queried `FinalGrade` for each one. The same logic was easily accomplished by defining a simple model manager and using single queries in each case. That got rid of a good amount of code and is obviously more efficient
- Function names/docs were interchangeably referring to grades as 'final' grades and 'frozen' grades. I changed instances of 'frozen' to 'final'

This should also help us move some functionality outside of `MMTrack` where it's appropriate. `mail/permissions.py`, for example, has a couple of permissions where `MMTrack` needs to be  instantiated just to find out if a user has paid for a course run in a given program. Instantiating `MMTrack` does a fair amount of work that isn't necessary for that logic. The `FinalGradeQuerySet` will allow us to create API methods outside off `MMTrack` which can be used in those permission classes (and elsewhere) without the extra overhead

### DIFF
--- a/grades/models.py
+++ b/grades/models.py
@@ -26,10 +26,45 @@ class CourseRunGradingAlreadyCompleteError(Exception):
     """
 
 
+class FinalGradeQuerySet(models.QuerySet):
+    """
+    QuerySet class defining common query parameters for FinalGrade
+    """
+    def passed(self):
+        """
+        Returns a queryset with a filter that will only fetch passed course runs
+        """
+        return self.filter(passed=True)
+
+    def paid_on_edx(self):
+        """
+        Returns a queryset with a filter that will only fetch course runs that are paid on edX.
+        Note that a course run can still be paid via an Order if this value is False for a given user.
+        """
+        return self.filter(course_run_paid_on_edx=True)
+
+    def for_course_run_key(self, edx_course_key):
+        """
+        Returns a queryset with a filter that will only fetch course runs that match a course key
+        """
+        return self.filter(course_run__edx_course_key=edx_course_key)
+
+    def for_course_run_keys(self, edx_course_keys):
+        """
+        Returns a queryset with a filter that will only fetch course runs that match one of many course keys
+        """
+        return self.filter(course_run__edx_course_key__in=edx_course_keys)
+
+
 class FinalGrade(TimestampedModel, AuditableModel):
     """
     Model to store edx final grades
     """
+    # Set a custom manager for this model. This lets us do a couple useful things:
+    # (1) Apply commonly-needed filters by a short, easily-recognized name, and (2) compose a query piece-by-piece.
+    # Docs: https://docs.djangoproject.com/en/1.10/topics/db/managers/#creating-a-manager-with-queryset-methods
+    objects = FinalGradeQuerySet.as_manager()
+
     user = models.ForeignKey(User, null=False)
     course_run = models.ForeignKey(CourseRun, null=False)
     grade = models.FloatField(
@@ -54,6 +89,11 @@ class FinalGrade(TimestampedModel, AuditableModel):
 
     def to_dict(self):
         return serialize_model_object(self)
+
+    @property
+    def grade_percent(self):
+        """Returns the grade field value as a number out of 100 (or None if the value is None)"""
+        return self.grade * 100 if self.grade else None
 
     @classmethod
     def get_frozen_users(cls, course_run):
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -306,9 +306,9 @@ def get_status_for_courserun(course_run, mmtrack):
         CourseRunUserStatus: an object representing the run status for the user
     """
 
-    if mmtrack.has_paid_frozen_grade(course_run.edx_course_key):
+    if mmtrack.has_paid_final_grade(course_run.edx_course_key):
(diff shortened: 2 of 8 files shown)


giocalitri:
My first comment is that the way `extract_final_grade` has been written is intentional.
There is an actual meaning in the fact that the final grade does not exist and it is usually associated with an erroneous state (the final grade should exist and it does not for some reason).
The fact that a lot of code has been written around that without taking in account that that function can raise an exception is the actual problem. 

Now I am fine with reimplementing this in a way that optimizes the queries, but I keep believing that `extract_final_grade` should raise in case there is no frozen final grade.

giocalitri:
Another comment is about the naming policy. 
Given that we will have 3 different "final grades", I think that if we rename the methods in MMTrack, we should do in a way that makes clear that those are the edx final grades (vs the pearson final grades vs the micromaster final grade).

gsidebo:
@giocalitri 

Re: `extract_final_grade` and exceptions - I only saw one specific case where a user not having a final grade for a course run was truly erroneous/exceptional, and that was in `get_status_for_courserun` ([diff line](https://github.com/mitodl/micromasters/pull/2909/files#diff-1bb85575dd5188300f5033a86bd7d1c2R356)). Basically all other code that checks for or fetches a final grade assumes that it might not exist for a given user/course run. I added a new method called `get_required_final_grade` ([diff line](https://github.com/mitodl/micromasters/pull/2909/files#diff-bf545a2544a83f1acac6cd6d14942b19R214)) to be used in cases where the final grade is expected. Let me know if that addresses your concerns.

Re: naming policy - I think as long as the model is still called `FinalGrade`, I think that the functions in `MMTrack` that perform logic on that model data should reflect it. Maybe we won't end up renaming the model, but even so, the concepts of the Pearson final grade and the 'micromasters final grade' don't exist yet (as far as I know), and the renaming of these methods should happen when those concepts are introduced

gsidebo:
@giocalitri any more comments on this?

giocalitri:
@gsidebo I thought you were non working until Monday, so I was going to look it again today.
Sorry, I will re-review as soon as I'll be in the office.

gsidebo:
There's no rush. I'm on vacation but had a few free minutes this morning.

grades/models.py
```diff
@@ -22,6 +22,23 @@ class CourseRunGradingAlreadyCompleteError(Exception):
...
+        return self.filter(passed=True)
+
+    def paid(self):  # pylint: disable=missing-docstring
+        return self.filter(course_run_paid_on_edx=True)
+
+    def for_course_run_key(self, edx_course_key):  # pylint: disable=missing-docstring
+        return self.filter(course_run__edx_course_key=edx_course_key)
+
+    def for_course_run_keys(self, edx_course_keys):  # pylint: disable=missing-docstring
+        return self.filter(course_run__edx_course_key__in=edx_course_keys)
```
giocalitri: I still would like to see docstrings for each method

grades/models.py
```diff
@@ -22,6 +22,23 @@ class CourseRunGradingAlreadyCompleteError(Exception):
...
 
+class FinalGradeQuerySet(models.QuerySet):
+    """
+    QuerySet class defining common query parameters for FinalGrade
+    """
+    def passed(self):  # pylint: disable=missing-docstring
+        return self.filter(passed=True)
+
+    def paid(self):  # pylint: disable=missing-docstring
+        return self.filter(course_run_paid_on_edx=True)
```
giocalitri: this is conceptually wrong: if `course_run_paid_on_edx` is `False` it does not mean that it has not been paid.

grades/models.py
```diff
@@ -41,6 +58,8 @@ class FinalGrade(TimestampedModel, AuditableModel):
     )
     course_run_paid_on_edx = models.BooleanField(default=False)
 
+    objects = FinalGradeQuerySet.as_manager()
```
giocalitri: this should be at the very beginning of the class definition with a pretty detailed explanation of what it is doing, given that this is the very first time that we use a custom manager

gsidebo:
@giocalitri all comments addressed

giocalitri:
I think this is good to go 👍

PR #2910, opened 2017-03-17
Generate robotic avatars

### LINKED ISSUE #915: Update realistic search data to include fake avatars

noisecapella:
Fixes #915

#### What's this PR do?
Adds a management command `attach_avatars` with argument `--name-prefix` to attach robot avatars to profiles.

Look at the learner search. Choose a username prefix and run the management command. Adjust the search or refresh the page and note that the user avatar is now a robot

This uses the robohash library to generate the avatars based on the username, it doesn't actually download it from https://robohash.com

screenshot from 2017-03-17 16-37-02


noisecapella:
@gsidebo Profiles already have signals which handle reindexing and the signals aren't disabled for this management command. You shouldn't need to run recreate_index. Was celery up and running when you ran the command? I had an issue where it was down and the indexing tasks were being queued to be run later

The other two saves are on the image themselves and update the file data rather than the row in the database

seed_data/management/commands/attach_avatars.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    help = "Attach avatars to profiles"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--name-prefix",
+            dest="name_prefix",
+            help="username prefix to user",
+            required=True,
+        )
```
noisecapella: Sure

PR #2931, opened 2017-03-22
Fixed bug with gdm_grade_task_fail_bug management command

### LINKED ISSUE #2930: check_final_grade_freeze_status management command is not working
it raises an exception when the task is submitted

giocalitri:
fixes #2930

#### What's this PR do?
fixes a bug with the `gdm_grade_task_fail_bug` management command.

Submit a freeze task for a course run using the `freeze_final_grades`, then check the status with `gdm_grade_task_fail_bug`: it should not raise any more.

### DIFF
--- a/grades/management/commands/check_final_grade_freeze_status.py
+++ b/grades/management/commands/check_final_grade_freeze_status.py
@@ -37,8 +37,9 @@ class Command(BaseCommand):
                 )
             )
         elif CourseRunGradingStatus.is_pending(run):
-            group_results_id = CACHE_ID_BASE_STR.format(edx_course_key)
-            if cache_redis.get(group_results_id):
+            cache_id = CACHE_ID_BASE_STR.format(edx_course_key)
+            group_results_id = cache_redis.get(cache_id)
+            if group_results_id is not None:
                 results = GroupResult.restore(group_results_id)
                 if not results.ready():
                     self.stdout.write(
--- a/grades/management/commands/complete_course_run_freeze.py
+++ b/grades/management/commands/complete_course_run_freeze.py
@@ -47,8 +47,9 @@ class Command(BaseCommand):
             return
 
         # check if there are tasks running
-        group_results_id = CACHE_ID_BASE_STR.format(edx_course_key)
-        if cache_redis.get(group_results_id):
+        cache_id = CACHE_ID_BASE_STR.format(edx_course_key)
+        group_results_id = cache_redis.get(cache_id)
+        if group_results_id is not None:
             results = GroupResult.restore(group_results_id)
             if results and not results.ready():
                 self.stdout.write(


grades/management/commands/check_final_grade_freeze_status.py
```diff
@@ -38,8 +38,9 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
             )
         elif CourseRunGradingStatus.is_pending(run):
             group_results_id = CACHE_ID_BASE_STR.format(edx_course_key)
-            if cache_redis.get(group_results_id):
-                results = GroupResult.restore(group_results_id)
+            celery_results_id = cache_redis.get(group_results_id)
+            if celery_results_id:
```
noisecapella: Should this be `celery_results_id is not None`?

giocalitri: the check should should handle also an empty string

noisecapella:
There are three places which use `CACHE_ID_BASE_STR`:

freeze_course_run_final_grades in grades/tasks.py:

    # cache id string for this task
    cache_id = CACHE_ID_BASE_STR.format(course_run.edx_course_key)

    # try to get the result id from a previous iteration of this task for this course run
    group_results_id = cache_redis.get(cache_id)

    # if the id is not none, it means that this task already run before for this course run
    # so we need to check if its subtasks have finished
    if group_results_id is not None:
        # delete the entry from the cache (if needed it will be added again later)
        cache_redis.delete(cache_id)
        # extract the results from the id
        results = GroupResult.restore(group_results_id)

check_final_grade_freeze_status.py:

            group_results_id = CACHE_ID_BASE_STR.format(edx_course_key)
            celery_results_id = cache_redis.get(group_results_id)
            if celery_results_id:
                results = GroupResult.restore(celery_results_id)

complete_course_run_freeze.py:

        # check if there are tasks running
        group_results_id = CACHE_ID_BASE_STR.format(edx_course_key)
        if cache_redis.get(group_results_id):
            results = GroupResult.restore(group_results_id)
            if results and not results.ready():

Can we make the variable names consistent for these three cases, and fix the use of group results for the last one?

grades/management/commands/check_final_grade_freeze_status.py
noisecapella: In `freeze_course_run_final_grades` it checks against `None` explicitly, so we should remove that check if that one needs to check against an empty string too

giocalitri: actually reading the code of `freeze_course_run_final_grades` I realized that it cannot be an empt string, so in that case it is better if it fails.

giocalitri:
done

noisecapella:
Looks good, but I'm going to wait to approve to merge my PR first so I can put it in the release candidate

PR #2937, opened 2017-03-22
Added model and admin for Proctorate Exam Grades

### LINKED ISSUE #2790: exams: create models for storing exam results
Part of #2752 

We need a model for storing exam results. 

The results will be displayed to users in their dashboard, and also used to calculate their final grade for the course. 

### acceptance criteria
- [ ] audit tables
- [ ] simple admin interface that allows for looking up exam grades, and allows editing

giocalitri:
closes #2790 

#### What's this PR do?
creates an audited model to storeProctorate Exam Grades

there is no need: it is only a model

NOTE: the size of the `CharField`s is correspondent to what Pearson defined in their document.


grades/admin.py
```diff
@@ -34,17 +34,55 @@ class FinalGradeAuditAdmin(admin.ModelAdmin):
     def has_add_permission(self, *args, **kwargs):  # pylint: disable=unused-argument
         return False
 
+    def has_change_permission(self, *args, **kwargs):  # pylint: disable=unused-argument
```
noisecapella: We left this out originally because it causes the admin to disappear entirely. Instead we set `readonly_fields` to be all of the fields

grades/models.py
```diff
@@ -146,3 +150,59 @@ def create_pending(cls, course_run):
...
+    course = models.ForeignKey(Course, null=False)
+
+    # fields from proctorate exam results
+    passing_score = models.FloatField()
+    score = models.FloatField()
+    grade = models.TextField()
+    client_authorization_id = models.TextField()
+
+    # dump of all the remaining fields coming from the proctorate exam results
+    row_data = JSONField(null=None)
```
noisecapella: Can you change to `null=False` here?

grades/models.py
```diff
@@ -146,3 +150,59 @@ def create_pending(cls, course_run):
...
+
+    # fields coming from unknown source but about proctorate exam
+    # we need the following to compute the percentage grade
+    maximum_exam_score = models.FloatField()
+
+    # custom fields based on proctorate exam results
+    passed = models.BooleanField(null=False)
```
noisecapella: This is redundant since `BooleanField` doesn't allow `null=True`, they require `NullBooleanField` instead for that: https://docs.djangoproject.com/en/1.10/ref/models/fields/#null

grades/models.py
```diff
@@ -146,3 +150,59 @@ def create_pending(cls, course_run):
...
+    """
+    proctored_exam_grade = models.ForeignKey(ProctoredExamGrade, null=True, on_delete=models.SET_NULL)
+
+    @classmethod
+    def get_related_field_name(cls):
+        return 'proctored_exam_grade'
+
+    def __str__(self):
+        return 'Proctored Exam Grade audit for user "{user}", course "{course_title}"'.format(
+            user=self.final_grade.user,
```
noisecapella: `self.final_grade` doesn't exist for this model

giocalitri:
done

PR #2938, opened 2017-03-22
Fix parameter in attach_avatars

noisecapella:
None

#### What's this PR do?
Fixes the parameter used for `attach_avatars` and adds a test for the command

Run `./manage.py attach_avatars --username-prefix fake` to use robotic avatars on your fake users


rhysyngsun:
:+1: assuming tests pass

Issue #2742, opened 2017-03-01
Name search in learners search does not work with full name


noisecapella:
It does a prefix match on first, last, or preferred name (and soon the username and email), but it does not do full name searches. Not sure if this is a bug or feature request

pdpinch:
For example, if you search for the fake user "Xavier Addy" -- i.e. `/learners/?q=Xavier%20Addy` -- you will get no results. But if you search for "Xavier" or "Addy" you will find him.

singingwolfboy:
We can reconfigure the way Searchkit queries the backend to make it return results that are a perfect name match. For example, for a user "John Smith", we could modify the query parameters so that Elasticsearch knows to return that user for the search string "John Smith". However, we can't currently do a prefix match for both first name and last name: the search string "John Smi" would not return the user.

To support prefix matches for both first name and last name, we'll need to add a `full_name` field to the Elasticsearch index. We could do that with an [ingest pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/ingest.html), but we need to upgrade to Elasticsearch 5 before we can use that feature. (#2867).

We might be able to do add a `full_name` field to the Elasticsearch index by changing how Python serializes Profile objects before sending them off to Elasticsearch.

singingwolfboy:
Correction: we can't make Searchkit send queries that will _simultaneously_ return a perfect name match _and_ a prefix match. In order to do that, we still need a `full_name` field.

PR #2940, opened 2017-03-23
Added full name search support

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/2742

#### What's this PR do?
- Allows user to search learner on full name, first name, last name, email, username.

- run `docker-compose run web ./manage.py recreate_index`
- Go to `http://192.168.99.100:8079/learners/?q=full name` and search full name etc

@pdpinch 

screen shot 2017-03-24 at 12 37 12 pm


noisecapella:
In your screenshot there is a second result that doesn't match the full name. I think if a user puts in their full name it should only show cases where that name or a prefix matches. Can you add a separately indexed full_name field and add it to the prefix search?

amir-qayyum-khan:
@pdpinch wants to handle this scenario without reindexing. This this the best solution i found without reindexing, 

if user enters first or last name, username, email it will show exact matches, but if input contain spaces it will go for all records which contain substring of input (splited by space) 

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#type-cross-fields

pdpinch:
I'm actually fine with a looser match, since the full name we get via ZenDesk may not be a strict match for any "full name" we construct for the index. 

I also want to avoid the elasticsearch upgrade that DB was suggesting that as necessary.

noisecapella:
With the change in this PR the names only match when the entire first or last name is inserted:

screenshot from 2017-03-23 13-13-41

noisecapella:
@pdpinch @singingwolfboy's suggestion wouldn't be necessary in any case. He was referring to the ability for ES to do data transformations within ES, but we can do the same transformations in Python since we control that code, so we don't need to upgrade ES

noisecapella:
Looks good to me. @pdpinch are you ok with having the reindex?

pdpinch:
Yeah, this is fine. Thanks for checking.

PR #2941, opened 2017-03-23
redux api wrapper

aliceriot:
(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/static/js/constants.js
+++ b/static/js/constants.js
@@ -125,3 +125,7 @@ export const PEARSON_PROFILE_SUCCESS = 'success';
 export const PEARSON_PROFILE_IN_PROGRESS = 'in-progress';
 export const PEARSON_PROFILE_INVALID = 'invalid';
 export const PEARSON_PROFILE_SCHEDULABLE = 'schedulable';
+
+export const GET = 'GET';
+export const PATCH = 'PATCH';
+export const POST = 'POST';
--- a/static/js/flow/restTypes.js
+++ b/static/js/flow/restTypes.js
@@ -1,0 +1,21 @@
+// @flow
+
+export type RestState = {
+  data?: any,
+  error?: any,
+  processing: boolean,
+  loaded: boolean,
+  fetchStatus?: string,
+};
+
+export type Endpoint = {
+  name: string,
+  url: string,
+  makeOptions: (...args: any) => Object,
+  extraActions?: Object,
+  getPrefix?: string,
+  postPrefix?: string,
+  getFunc?: Function,
+  postFunc?: Function,
+  verbs: Array<string>,
+};
--- a/static/js/lib/redux_rest.js
+++ b/static/js/lib/redux_rest.js
@@ -1,0 +1,173 @@
+// @flow
+import { createAction } from 'redux-actions';
+import type { Dispatch } from 'redux';
+import R from 'ramda';
+
+import type { Action, ActionType, Dispatcher } from '../flow/reduxTypes';
+import type { Endpoint, RestState } from '../flow/restTypes';
+import { fetchJSONWithCSRF } from './api';
+import {
+  FETCH_PROCESSING,
+  FETCH_SUCCESS,
+  FETCH_FAILURE,
+} from '../actions';
+import { GET, PATCH, POST } from '../constants';
+import { couponEndpoint } from '../reducers/coupons';
+
+const actionize = R.compose(R.toUpper, R.join("_"));
+
+export const requestActionType = (...xs: string[]) => `REQUEST_${actionize(xs)}`;
+
+export const successActionType = (...xs: string[]) => `RECEIVE_${actionize(xs)}_SUCCESS`;
+
+export const failureActionType = (...xs: string[]) => `RECEIVE_${actionize(xs)}_FAILURE`;
+
+export const clearActionType = (...xs: string[]) => `CLEAR_${actionize(xs)}`;
+
+export const INITIAL_STATE: RestState = {
+  loaded: false,
+  processing: false,
+};
+
+const defaultRESTPrefixes = { GET, PATCH, POST };
+
+const getPrefixForEndpoint = (verb, endpoint) => (
+  R.propOr(defaultRESTPrefixes[verb], `${R.toLower(verb)}Prefix`, endpoint)
+);
+
+export function makeFetchFunc(endpoint: Endpoint): (...args: any) => Promise<*> {
+  return (...args) => fetchJSONWithCSRF(endpoint.url, endpoint.makeOptions(...args));
+}
+
+type DerivedAction = {
+  action:       (...params: any) => Dispatcher<*>,
+  requestType:  ActionType,
+  successType:  ActionType,
+  failureType:  ActionType,
+};
+
+// we can derive an action from an Endpoint object and an HTTP verb 
+// this returns an object containing the derived REST action (which
+// is an async action suitable for use with redux-thunk), and all the
+// derived action types.
+export const deriveAction = (endpoint: Endpoint, verb: string): DerivedAction => {
+  const prefix = getPrefixForEndpoint(verb, endpoint);
+
+  const requestType = requestActionType(prefix, endpoint.name);
+  const successType = successActionType(prefix, endpoint.name);
+  const failureType = failureActionType(prefix, endpoint.name);
+
+  const requestAction = createAction(requestType);
+  const successAction = createAction(successType);
+  const failureAction = createAction(failureType);
+
+  const fetchFunc = R.propOr(
+    makeFetchFunc(endpoint),
+    `${R.toLower(verb)}Func`,
+    endpoint
+  );
+
+  return {
+    action: (...params) => {
+      return (dispatch: Dispatch): Promise<*> => {
+        dispatch(requestAction());
+        return fetchFunc(...params).then(data => {
+          dispatch(successAction(data));
+          return Promise.resolve(data);
+        }).catch(error => {
+          dispatch(failureAction(error));
+          return Promise.reject(error);
+        });
+      };
+    },
+    requestType, successType, failureType
+  };
+};
+
+// here we loop through the HTTP verbs that an Endpoint implements, and
+// for each one we derive an action. The derived action types are stuck on
+// the function object, as a convenience.
+export const deriveActions = (endpoint: Endpoint) => {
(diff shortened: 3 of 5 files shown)


example.js
```diff
@@ -0,0 +1,32 @@
...
+
+// just usage examples
+
+import { rest } from '../lib/redux_api';
+
+// issue a get request
+const { user: { username }} = SETTINGS;
+rest.dashboard.get(username);
+
+rest.profile.get(username).then(profile => {
```
noisecapella: It looks like this allows multiple verbs per reducer. What about two uses of the same verb? I don't think we have any APIs that use that right now, but theoretically we could implement `/api/v0/programs/`  and `/api/v0/programs/1/` which would both use GET and should put their information in the same reducer. But maybe that's YAGNI for now

static/js/lib/redux_api.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+const defaultRESTPrefixes = { GET, PATCH, POST };
+
+const getPrefixForEndpoint = (verb, endpoint) => (
+  R.propOr(`${R.toLower(verb)}Prefix`, defaultRESTPrefixes[verb], endpoint)
+);
+
+export function makeFetchFunc(endpoint: Endpoint): (...args: any) => Promise<*> {
+  return (...args) => fetchJSONWithCSRF(endpoint.url, endpoint.makeOptions(...args));
```
noisecapella: How should we handle the `loginOnError` parameter for `fetchJSONWithCSRF`? It's sort of arbitrarily used right now, maybe it should be the default behavior instead for the GET requests

static/js/lib/redux_api.js
```diff
@@ -0,0 +1,165 @@
...
+  return {
+    action: (...params) => {
+      return (dispatch: Dispatch): Promise<*> => {
+        dispatch(requestAction());
+        return fetchFunc(...params).then(data => {
+          dispatch(successAction(data));
+          return Promise.resolve(data);
+        }).catch(error => {
+          dispatch(failureAction(error));
+          return Promise.reject();
```
noisecapella: Might want to pass along the error here just in case

static/js/lib/redux_api.js
```diff
@@ -0,0 +1,165 @@
...
+  const successType = requestActionType(prefix, endpoint.name);
+  const failureType = failureActionType(prefix, endpoint.name);
+  const clearType = clearActionType(prefix, endpoint.name);
+
+  const requestAction = createAction(requestType);
+  const successAction = createAction(successType);
+  const failureAction = createAction(failureType);
+
+  const fetchFunc = R.propOr(
+    `${verb}Func`,
```
noisecapella: Is this meant to allow the user to override with their own fetch function?

static/js/lib/redux_api.js
noisecapella: Also it doesn't look like `makeOptions` exists in the coupon endpoint below, is that intentional?

example.js
aliceriot: Ah, I didn't consider that. Ideally we'd support that use-case from the beginning instead of needing to graft it on later, but I'm not sure exactly what the best way to do that would be.

We could maybe call `/api/v0/programs/` the `list` or `index` action, and reserve `get` for a detail view (i.e. for `/api/v0/programs/1`), hrrmmm

that would obviously get more complicated if the index view supported other verbs besides `GET`...

yeah, I'm leaning towards not handling that case for now, since we don't need it right now, and it wouldn't be too difficult to add it in later.

static/js/lib/redux_api.js
aliceriot: exactly, for some things I think that may be easier for some of our APIs. The simple cases (where we just tossing JSON around) we should be able to use the `makeFetchFunc` method, but there are a few cases where I think it would be much simpler to allow overriding. The cases in particular I'm thinking of are the course price API (where we do an `integer -> Decimal` conversion in the `.then` off the fetch Promise) and the profile image (where we make a `FormData` object, append the image blob to that, and then put the `FormData` in the body of a `PATCH` request). Both of those cases seem like edge casey things where it would be overly complicated to try to get the `makeFetchFunc` function to handle them - but it should be fine for our more straightforward APIs (like the `getPrograms` API function should no longer be necessary, for instance).

static/js/lib/redux_api.js
aliceriot: oh yea, good catch :)

static/js/lib/redux_api.js
aliceriot: see below about `makeOptions`

for the `loginOnError` parameter I'm not sure, we definitely want it on the majority of APIs but a few (like the `Profile` API and stuff like that) are publicly accessible (providing that a user's permissions are configured correctly). so yea, maybe it makes more sense to make `true` the default value, and override it on the `Endpoint` object when it should be `false`.

static/js/lib/redux_api.js
noisecapella: It would also be nice for cases where we want to allow quick and dirty mocking of data for development

aliceriot: oh yeah that too!

like

```js
const mockedEndpoint = {
  getFunc: () => Promise.resolve(DASHBOARD_RESPONSE),
};
```

aliceriot:
@noisecapella OK! I think this is ready for an actual review. I have decided to hold off on actually changing over any of the APIs to use this (just so the PR doesn't get too big) but I did try switching over the coupon API to it and things were working great (I'm looking forward to deleting a lot of code!).

static/js/lib/redux_api.js
```diff
@@ -0,0 +1,166 @@
...
+};
+
+const defaultRESTPrefixes = { GET, PATCH, POST };
+
+const getPrefixForEndpoint = (verb, endpoint) => (
+  R.propOr(defaultRESTPrefixes[verb], `${R.toLower(verb)}Prefix`, endpoint)
+);
+
+export function makeFetchFunc(endpoint: Endpoint): (...args: any) => Promise<*> {
+  return (...args) => fetchJSONWithCSRF(endpoint.url, endpoint.makeOptions(...args));
```
noisecapella: Are you delaying the `loginOnError` argument for a future PR?

static/js/lib/redux_api_test.js
```diff
@@ -0,0 +1,341 @@
...
+      it('should no-op if reducer for an action type is not present', () => {
+        let reducer = deriveReducers(endpoint, actions);
+
+        let initialState = { initial: 'State' };
+        let result = reducer(initialState, { type: '❤❤❤❤❤', payload: '<3<3<3<3<3'});
+        assert.equal(result, initialState);
+      });
+    });
+
+    describe.skip('exported reducers', () => { // eslint-disable-line mocha/no-skipped-tests
```
noisecapella: Is this only being skipped for this PR?

static/js/lib/redux_api.js
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// we can derive an action from an Endpoint object and an HTTP verb 
+// this returns an object containing the derived REST action (which
+// is an async action suitable for use with redux-thunk), and all the
+// derived action types.
+export const deriveAction = (endpoint: Endpoint, verb: string): Object => {
```
noisecapella: Is it worth defining the return type for this function?

static/js/lib/redux_api.js
```diff
@@ -0,0 +1,166 @@
...
+      processing: false,
+    }),
+    [action.failureType]: (state: Object, action: Action<any, any>) => ({
+      ...state,
+      [fetchStatus]: FETCH_FAILURE,
+      error: action.payload,
+      loaded: true,
+      processing: false,
+    }),
+  };
```
noisecapella: Are we handling the clear action elsewhere?

aliceriot: Yes, before I was defining two redundant handlers (one for each HTTP verb) which didn't make much sense to me. I refactored it a little bit to define one clear action for the whole reducer. So now the action type and action creator are defined in `deriveActions` instead of `deriveAction` (the latter being the per-verb derivation function) and the function to handle the clear action in the reducer is now defined in `deriveReducers` instead of `deriveReducer` (which, again, is the per-verb function).

The relevant code is on line 147, adding the handler once for the whole reducer. It also allows the creator of the `Endpoint` object to override the default `INITIAL_STATE` if that is necessary (by putting an `initialState` object on the `Endpoint`).

static/js/lib/redux_api_test.js
aliceriot: Yes :) I just wanted to not have to re-write the tests when I do the follow-up PR to switch over the coupons reducer.

static/js/lib/redux_api.js
aliceriot: Oh right I forgot about that, do you think we should just go with what we discussed above?

static/js/lib/redux_api.js
aliceriot: Yes, probably! I'll go ahead and do that.

aliceriot:
One higher-level question: does the module name (`redux_api`) make sense to you? I was also considering calling it `redux_rest` or something similar.

noisecapella:
@aliceriot Either works for me, your choice

aliceriot:
ok! I think I'm going to change it to `redux_rest`.

static/js/lib/redux_api.js
noisecapella: Sure, works for me

aliceriot:
alright, I think comments are all addressed!

static/js/lib/api.js
```diff
@@ -92,7 +92,7 @@ import { fetchWithCSRF } from './api';
  *  - non 2xx status codes will reject the promise returned
  *  - response JSON is returned in place of response
  */
-const _fetchJSONWithCSRF = (input: string, init: Object = {}, loginOnError: boolean = false): Promise<*> => {
+const _fetchJSONWithCSRF = (input: string, init: Object = {}, loginOnError: boolean = true): Promise<*> => {
```
noisecapella: I think we should only default to true for GET requests. 400 is returned whenever there is a validation error for POST requests. Can we leave it as `false` then deal with it in the next PR?

aliceriot: sure

PR #2946, opened 2017-03-24
Cleaned up more MMTrack code

### LINKED ISSUE #2929: Further MMTrack refactoring
`dashboard.utils.MMTrack` exposes a lot of functionality that we initially used for the dashboard API only, but is now used elsewhere in the codebase, eg: `search/indexing_api.py`, `mail/permissions.py`. It's doing some work that isn't necessary for the needs of the codebase outside of the dashboard API. Some examples:

- The `pearson_exam_status` is being calculated as part of the object init. This is only needed in the dashboard API, so any other clients are running unnecessary extra queries against `CourseRun` and `ExamProfile`. Even in the dashboard API there's no reason for this value to be set in the object init (`get_pearson_exam_status()` can be called directly in the only place where it's currently needed), so this can be fixed very simply.
- Everything from line 75 to 93, as well as the `_get_program_fa_prices` function, is only being used to serialize financial aid information for the dashboard API. In addition to the code bloat, it involves queries to `FinancialAid` and `TierProgram`. It should be moved out of the class (probably to a serializer).
- _(optional/less important)_ `mail/permissions.py` needs to know if a user has paid for any course run in a program. It's currently making use of `MMTrack.has_paid`. `MMTrack` requires cached certificates and current grades to be fetched, but determining whether a user has paid for a set of course runs would only require cached enrollment data. That function could be broken out of the class in a way that both the permission classes and `MMTrack` could use it without the unnecessary work. Perhaps more importantly, the usage of `MMTrack` in `dashboard.serializers.UserProgramSearchSerializer` doesn't seem to need cached certificates and current grades either.

gsidebo:
Closes #2929
~~⚠️ Rebased on #2909 ⚠️~~

#### What's this PR do?
Cleans up more `dashboard.utils.MMTrack` code, building off of the work in #2909. The details about the fixes are in the #2929 issue description.

Test that changes to user course data result in the expected changes to the dashboard output. The `alter_data` management command can be used to this end (`manage.py alter_data examples`).

`dashboard/utils.py`

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -216,7 +216,7 @@ class MMTrack:
         Returns:
             bool: whether a frozen final grade exists
         """
-        return self.final_grade_qset.paid_on_edx().for_course_run_key(course_id).exists()
+        return self.final_grade_qset.paid_on_edx().for_course_run_key(edx_course_key).exists()
 
     def has_passed_course(self, edx_course_key):
         """
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -70,7 +70,6 @@ class MMTrack:
                 ).values_list("edx_course_key", "course__id")
             )
             self.course_ids = self.edx_key_course_map.keys()
-            self.pearson_exam_status = self.get_pearson_exam_status()
 
             if self.financial_aid_available:
                 self.paid_course_ids = set(Line.objects.filter(
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -11,7 +11,7 @@ from pytz import utc
 
 from courses.models import CourseRun
 from dashboard.api_edx_cache import CachedEdxUserData
-from ecommerce.models import Line
+from ecommerce.models import Order, Line
 from grades.constants import FinalGradeStatus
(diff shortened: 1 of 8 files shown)


dashboard/utils.py
```diff
@@ -69,231 +59,212 @@ def __init__(self, user, program, edx_user_data):
...
             )
-            self.course_ids = self.edx_key_course_map.keys()
-            self.pearson_exam_status = self.get_pearson_exam_status()
+            self.edx_course_keys = set(self.edx_key_course_map.keys())
 
             if self.financial_aid_available:
-                self.paid_course_ids = set(Line.objects.filter(
-                    Q(order__status='fulfilled') & Q(course_key__in=self.course_ids) & Q(order__user=user)
+                self.paid_course_keys = set(Line.objects.filter(
+                    order__status='fulfilled', course_key__in=self.edx_course_keys, order__user=user
```
rhysyngsun: While you're touching this, could you change `'fulfilled'` to `Order.FULFILLED`?

gsidebo: done, and the branch is rebased

PR #2956, opened 2017-03-24
Added program.price, removed CoursePrice

### LINKED ISSUE #1144: Move course price from CoursePrice into Program

noisecapella:
Fixes #1144

#### What's this PR do?
Removes `CoursePrice` and moves its price into a new field `Program.price`. If there isn't the same price for each `CourseRun` the migration should fail, leaving the original data intact.

Run the migrations, then run them in reverse order. Checkout master and everything should still work like before. Also run `validate_db` after running migrations to confirm that that still works.


courses/migrations/0023_add_price.py
```diff
@@ -0,0 +1,77 @@
...
+    CourseRun = apps.get_model('courses', 'CourseRun')
+
+    program_prices = {}
+    # Note: we are not filtering on live here because all Programs will require prices and liveness won't matter
+
+    # Validate that there is only one price per program and each run had a price. Else, fail the migration
+    for run in CourseRun.objects.all():
+        program = run.course.program
+        # There must be one and only one price per course run
+        price = run.courseprice_set.get(is_valid=True).price
```
noisecapella: Sure

dashboard/utils.py
```diff
@@ -102,7 +102,7 @@ def _get_program_fa_prices(self):
         """
         Returns the financial aid possible cost range.
         """
-        course_max_price = self.program.get_course_price()
+        course_max_price = self.program.price
```
gsidebo: @noisecapella i just merged a PR that moves this code. on master this is now in `financialaid.serializers.FinancialAidDashboardSerializer#get_program_price_range`

noisecapella:
@rhysyngsun Any other comments?

pdpinch:
I need to run the validate management command on rc and prod so we know if the migration is going to work...

PR #2960, opened 2017-03-27
Add refresh_index to fix race condition with percolate

### LINKED ISSUE #2959: Race condition when mailing automatic emails for new users
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/235964047/activity/

noisecapella:
Fixes #2959 

#### What's this PR do?
Adds a couple `refresh_index` commands to flush updates to the index before we run percolate. This should make sure the program enrollment exists by the time we run percolate on it.

I don't think it's easy to reproduce. We will just need to keep an eye on sentry.


giocalitri:
👍

Issue #2952, opened 2017-03-24
Mail: radio buttons for choosing one-time / email campaign 


pdpinch:
Part of #2812

Replace the checkbox in the mail composition dialog with a radio button with two, more explicit options. Display instructional text when "email campaign" is chosen. 

![scheduled_email 2x](https://cloud.githubusercontent.com/assets/20047260/24111439/c41c0d50-0d6c-11e7-8bfd-da02ccc346a9.jpg)

amir-qayyum-khan:
@pdpinch @rhysyngsun what are these widgets
- Start sending 
- Stop sending

are the drop down, date picker or timer picker or text fields?

roberthouse54:
The "Start" and "Stop" fields are intended to be date pickers. But this is not intended to be included as part of this issue. ( @pdpinch correct me if I am wrong).

pdpinch:
That's right. The start/stop sending dates will be a separate issue, which I haven't written yet.

PR #2961, opened 2017-03-27
Added UI for email composition type

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2952

#### What's this PR do?
- adds radio buttons to select email composition type.

go to learners page and email these learners.

@pdpinch 
#### Screenshots (if appropriate):
screen shot 2017-03-30 at 8 06 56 pm

screen shot 2017-03-30 at 8 07 01 pm

### DIFF
--- a/static/js/components/email/AutomaticEmailOptions.js
+++ b/static/js/components/email/AutomaticEmailOptions.js
@@ -1,0 +1,51 @@
+// @flow
+import React from 'react';
+import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
+
+import {
+  ONE_TIME_EMAIL,
+  EMAIL_CAMPAIGN
+} from './constants';
+
+export default class EmailCompositionType extends React.Component {
+  props: {
+    automaticEmailType:    ?string,
+    setAutomaticEmailType: (b: string) => void,
+  };
+
+  handleRadioClick = (event: Event, value: string): void => {
+    const { setAutomaticEmailType } = this.props;
+    setAutomaticEmailType(value);
+  }
+
+  renderEmailCampaign = (): React$Element<*> => (
+    <div className="email-campaign-content">
+      This email will be sent now and in the future whenever users meet the criteria.
+    </div>
+  );
+
+  render() {
+    const { automaticEmailType } = this.props;
+
+    return (
+      <div className="email-type">
+        <RadioButtonGroup
+          className="type-radio-group"
+          name="email-composition-type"
+          valueSelected={automaticEmailType}
+          onChange={this.handleRadioClick}
+        >
+          <RadioButton
+            value={ONE_TIME_EMAIL}
+            label="Send a one-time email"
+            className="one-time-email" />
+          <RadioButton
+            value={EMAIL_CAMPAIGN}
+            label="Create an Email Campaign"
+            className="email-campaign" />
+        </RadioButtonGroup>
+        { automaticEmailType === EMAIL_CAMPAIGN ? this.renderEmailCampaign() : null }
+      </div>
+    );
+  }
+}
--- a/static/js/components/email/AutomaticEmailOptions.js
+++ b/static/js/components/email/AutomaticEmailOptions.js
@@ -26,18 +26,18 @@ export default class AutomaticEmailOptions extends React.Component {
       <div className="email-type">
         <RadioButtonGroup
           className="type-radio-group"
-          name="email-composition-type"
+          name="email-options"
           valueSelected={sendAutomaticEmails}
           onChange={this.handleRadioClick}
         >
           <RadioButton
             value={false}
             label="Send a one-time email"
-            className="one-time-email" />
+          />
           <RadioButton
             value={true}
             label="Create an Email Campaign"
-            className="email-campaign" />
+          />
         </RadioButtonGroup>
         { sendAutomaticEmails ? this.renderEmailCampaign() : null }
       </div>
--- a/static/js/components/email/AutomaticEmailOptions.js
+++ b/static/js/components/email/AutomaticEmailOptions.js
@@ -1,22 +1,16 @@
 // @flow
 import React from 'react';
-import R from 'ramda';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
-
-import {
-  ONE_TIME_EMAIL,
-  EMAIL_CAMPAIGN
-} from './constants';
 
 export default class AutomaticEmailOptions extends React.Component {
   props: {
-    sendAutomaticEmails?:  boolean,
-    setAutomaticEmailType: (b: boolean) => void,
+    sendAutomaticEmails?:   boolean,
+    setSendAutomaticEmails: (b: boolean) => void,
   };
 
-  handleRadioClick = (event: Event, value: string): void => {
-    const { setAutomaticEmailType } = this.props;
-    setAutomaticEmailType(R.equals(value, EMAIL_CAMPAIGN) ? true : false);
+  handleRadioClick = (event: Event, value: boolean): void => {
+    const { setSendAutomaticEmails } = this.props;
+    setSendAutomaticEmails(value);
   }
 
   renderEmailCampaign = (): React$Element<*> => (
@@ -24,10 +18,6 @@ export default class AutomaticEmailOptions extends React.Component {
       This email will be sent now and in the future whenever users meet the criteria.
     </div>
   );
-
-  getEmailType = (sendAutomaticEmails?: boolean): string => (
-    sendAutomaticEmails ? EMAIL_CAMPAIGN : ONE_TIME_EMAIL
-  )
 
   render() {
     const { sendAutomaticEmails } = this.props;
@@ -37,19 +27,19 @@ export default class AutomaticEmailOptions extends React.Component {
         <RadioButtonGroup
           className="type-radio-group"
           name="email-composition-type"
-          valueSelected={this.getEmailType(sendAutomaticEmails)}
+          valueSelected={sendAutomaticEmails}
           onChange={this.handleRadioClick}
         >
           <RadioButton
-            value={ONE_TIME_EMAIL}
+            value={false}
             label="Send a one-time email"
             className="one-time-email" />
           <RadioButton
-            value={EMAIL_CAMPAIGN}
+            value={true}
             label="Create an Email Campaign"
             className="email-campaign" />
         </RadioButtonGroup>
-        { this.getEmailType(sendAutomaticEmails) === EMAIL_CAMPAIGN ? this.renderEmailCampaign() : null }
+        { sendAutomaticEmails ? this.renderEmailCampaign() : null }
       </div>
     );
   }
--- a/static/js/components/email/AutomaticEmailOptions.js
+++ b/static/js/components/email/AutomaticEmailOptions.js
@@ -1,5 +1,6 @@
 // @flow
 import React from 'react';
+import R from 'ramda';
 import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
 
 import {
@@ -7,19 +8,15 @@ import {
   EMAIL_CAMPAIGN
 } from './constants';
 
-export default class EmailCompositionType extends React.Component {
+export default class AutomaticEmailOptions extends React.Component {
   props: {
-    automaticEmailType:    ?string,
-    setAutomaticEmailType: (b: string) => void,
+    sendAutomaticEmails?:  boolean,
+    setAutomaticEmailType: (b: boolean) => void,
   };
 
   handleRadioClick = (event: Event, value: string): void => {
     const { setAutomaticEmailType } = this.props;
-    setAutomaticEmailType('sendAutomaticEmails', {
-      target: {
-        value: value === EMAIL_CAMPAIGN ? true : false
-      }
-    });
+    setAutomaticEmailType(R.equals(value, EMAIL_CAMPAIGN) ? true : false);
   }
 
   renderEmailCampaign = (): React$Element<*> => (
@@ -28,15 +25,19 @@ export default class EmailCompositionType extends React.Component {
     </div>
   );
 
+  getEmailType = (sendAutomaticEmails?: boolean): string => (
+    sendAutomaticEmails ? EMAIL_CAMPAIGN : ONE_TIME_EMAIL
+  )
+
   render() {
-    const { automaticEmailType } = this.props;
+    const { sendAutomaticEmails } = this.props;
 
     return (
       <div className="email-type">
         <RadioButtonGroup
           className="type-radio-group"
           name="email-composition-type"
-          valueSelected={automaticEmailType}
+          valueSelected={this.getEmailType(sendAutomaticEmails)}
           onChange={this.handleRadioClick}
         >
           <RadioButton
@@ -48,7 +49,7 @@ export default class EmailCompositionType extends React.Component {
             label="Create an Email Campaign"
             className="email-campaign" />
         </RadioButtonGroup>
-        { automaticEmailType === EMAIL_CAMPAIGN ? this.renderEmailCampaign() : null }
+        { this.getEmailType(sendAutomaticEmails) === EMAIL_CAMPAIGN ? this.renderEmailCampaign() : null }
       </div>
     );
   }
--- a/static/js/components/email/AutomaticEmailOptions.js
+++ b/static/js/components/email/AutomaticEmailOptions.js
@@ -15,7 +15,11 @@ export default class EmailCompositionType extends React.Component {
 
   handleRadioClick = (event: Event, value: string): void => {
     const { setAutomaticEmailType } = this.props;
-    setAutomaticEmailType(value);
+    setAutomaticEmailType('sendAutomaticEmails', {
+      target: {
+        value: value === EMAIL_CAMPAIGN ? true : false
+      }
+    });
   }
 
   renderEmailCampaign = (): React$Element<*> => (
--- a/static/js/components/email/AutomaticEmailOptions_test.js
+++ b/static/js/components/email/AutomaticEmailOptions_test.js
@@ -1,0 +1,64 @@
+// @flow
+import React from 'react';
+import sinon from 'sinon';
+import { mount } from 'enzyme';
+import { assert } from 'chai';
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
+import getMuiTheme from 'material-ui/styles/getMuiTheme';
+
+import {
+  ONE_TIME_EMAIL,
+  EMAIL_CAMPAIGN
+} from './constants';
+import AutomaticEmailOptions from './AutomaticEmailOptions';
+
+describe('AutomaticEmailOptions', () => {
+  let sandbox, setEmailCompositionTypeStub;
+
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+    setEmailCompositionTypeStub = sandbox.stub();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  const renderComponent = (automaticEmailType: string = ONE_TIME_EMAIL) => (
+    mount(
+      <MuiThemeProvider muiTheme={getMuiTheme()}>
+        <AutomaticEmailOptions
+          setAutomaticEmailType={setEmailCompositionTypeStub}
+          automaticEmailType={automaticEmailType}
+        />
+      </MuiThemeProvider>
+    )
+  );
+
+  it('div renders', () => {
+    let wrapper = renderComponent(ONE_TIME_EMAIL);
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
+  });
+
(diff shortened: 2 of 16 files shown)


aliceriot:
We use the `EmailCompositionDialog` for other things besides the learner search emails, so we will need to only show this conditionally. In particular, this is what shows up if you click the 'contact course team' link:

fjfjfjfjfjfjf

the UI obviously shouldn't be there, because the user can't create campaigns.

aliceriot:
In particular, you're going to have to add a flag to the `EmailConfig` type which captures this configuration information.

The best way would be a field on the config object that holds an `extraEmailUI` react component, that way we don't have to have any implementation-specific details on the `EmailConfig` (compare with like a `showSearchUI` boolean flag for instance - yucky).

roberthouse54:
The results field display should look like it does in the mock.

amir-qayyum-khan:
Done @aliceriot

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -16,7 +17,9 @@ export default class EmailCompositionDialog extends React.Component {
     subheadingRenderer?:        (activeEmail: EmailState) => React$Element<*>,
     closeAndClearEmailComposer: () => void,
     closeEmailComposerAndSend:  () => void,
-    updateEmailFieldEdit:       () => void
+    updateEmailFieldEdit:       () => void,
+    extraUI?:                   (a: string, b: (c: string) => void) => React$Element<*>,
```
aliceriot: was there a difficulty with just passing a `React$Element` here for `extraUI`?

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -93,6 +108,7 @@ export default class EmailCompositionDialog extends React.Component {
       onRequestClose={closeAndClearEmailComposer}
     >
       <div className="email-composition-contents">
+        { this.renderExtraUI() }
```
aliceriot: I don't think you'll need the helper function, you could just do something like:

```js
{ R.isNil(extraUI) ? null : <extraUI {...this.props} /> }
```

static/js/components/email/EmailCompositionType.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import React from 'react';
+import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
+
+import {
+  ONE_TIME_EMAIL,
+  EMAIL_CAMPAIGN
+} from './constants';
+
+export default class EmailCompositionType extends React.Component {
```
aliceriot: We'll need to add more UI for rendering other options for these emails, so I think it would make sense to call this something else, maybe `AutomaticEmailOptions` or similar.

static/js/components/email/EmailCompositionType.js
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  render() {
+    const { emailCompositionType } = this.props;
+    let radioIconStyle = {'marginRight': '8px'};
+    let emailCampaign = null;
+
+    if (emailCompositionType === EMAIL_CAMPAIGN) {
+      emailCampaign = this.renderEmailCampaign();
+    }
```
aliceriot: can you inline this control flow in the JSX below with a ternary?

static/js/components/email/EmailCompositionType.js
```diff
@@ -0,0 +1,59 @@
...
+
+  renderEmailCampaign = (): React$Element<*> => (
+    <div className="email-campaign-content">
+      This email will be sent now and in the future whenever users meet the criteria.
+    </div>
+  );
+
+  render() {
+    const { emailCompositionType } = this.props;
+    let radioIconStyle = {'marginRight': '8px'};
```
aliceriot: is it tricky to style the radio buttons with sass? I'd like to stay away from inline styles wherever we can.

static/js/components/email/EmailCompositionType.js
```diff
@@ -0,0 +1,59 @@
...
+      emailCampaign = this.renderEmailCampaign();
+    }
+
+    return (
+      <div className="email-type">
+        <RadioButtonGroup
+          className="type-radio-group"
+          name="email-composition-type"
+          valueSelected={emailCompositionType}
+          onChange={(event, value)=> this.handleRadioClick(value)}
```
aliceriot: could you make `this.handleRadioClick` that `event, value` as it's arguments? then we can avoid creating this inline function.

static/js/components/email/lib.js
```diff
@@ -55,7 +56,13 @@ export const SEARCH_RESULT_EMAIL_CONFIG: EmailConfig = {
...
     emailState.searchkit.buildQuery().query,
     emailState.inputs.sendAutomaticEmails || false,
-  ])
+  ]),
+
+  extraUI: (emailCompositionType: string, setEmailCompositionType: (b: string) => void) => (
+    <EmailCompositionType
+      emailCompositionType={emailCompositionType}
+      setEmailCompositionType={setEmailCompositionType} />
+  )
```
aliceriot: I think it would be better to avoid the function wrapper, as I said above - we'll need to add more props to this, and if they're all being passed in via this function wrapper it could get pretty messy.

static/js/components/email/EmailCompositionDialog.js
amir-qayyum-khan: what do u mean by that?

Like this `extraUI?: React$Element<*>`

aliceriot: You're passing in a function here for the `extraUI` prop. The function takes a few arguments, and basically just wraps a react component, passing the props down to the react component. Can you just pass the react component instead, and leave passing props to the component to the `EmailCompositionDialog`? Just wondering if there's a technical reason why that wouldn't work, because if we can I think it's cleaner (the function wrapper is needless indirection).

static/js/components/email/lib.js
amir-qayyum-khan: so what is good here?

aliceriot:
@amir-qayyum-khan I just realized there's a functionality problem. If I go to `/learners` and send an email (with the 'automatic email' option selected) it doesn't actually create an `AutomaticEmail` on the backend. So something isn't wired up correctly on the frontend.

On `master` how this works is that the 'automatic' checkbox sets a `sendAutomaticEmails` boolean which lives in the email inputs - this whole `inputs` object is serialized and passed to the backend when we make the API call to create and send the email.

I think rather than adding the new `AUTOMATIC_EMAIL_TYPE` action you wrote we should just re-use the original pattern. You could do this by adding a new function on `EmailCompositionDialog` which uses the `updateEmailFieldEdit` function to set `'sendAutomaticEmails'` - you can take a look at what is currently on `master` in the `automaticCheckbox` method to get a good idea for how to go about that.

Sorry for not catching that earlier!

aliceriot:
@pdpinch 

Instead of adding a new email type field we should just re-use the existing boolean that the checkbox uses. This should also involve removing the existing checkbox UI. If we do this we won't need to make any backend changes in order to retain functionality and get the UI changes we want.

amir-qayyum-khan:
Done @aliceriot

static/js/components/email/AutomaticEmailOptions.js
```diff
@@ -0,0 +1,55 @@
+// @flow
+import React from 'react';
+import { RadioButton, RadioButtonGroup } from 'material-ui/RadioButton';
+
+import {
+  ONE_TIME_EMAIL,
+  EMAIL_CAMPAIGN
+} from './constants';
+
+export default class EmailCompositionType extends React.Component {
```
aliceriot: this class should be called `AutomaticEmailOptions`, going by the file name

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -43,32 +44,22 @@ export default class EmailCompositionDialog extends React.Component {
...
-            value: e.target.checked
-          }
-        });
-      }}
-    />;
-  };
+  renderAutomaticEmailSettings =
+    (sendAutomaticEmails: boolean, updateEmailFieldEdit: () => void): React$Element<*> => (
+    <AutomaticEmailOptions
+      automaticEmailType={sendAutomaticEmails ? EMAIL_CAMPAIGN : ONE_TIME_EMAIL}
```
aliceriot: Why not just pass in a `sendAutomaticEmails` prop here? I don't see that the `automaticEmailType` is useful, we aren't likely to have any more than two different types (and if we do in the future we could add it in at that point). For now I think it would be better to just use a boolean throughout to represent the choice the user is making.

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -43,32 +44,22 @@ export default class EmailCompositionDialog extends React.Component {
...
+      setAutomaticEmailType={updateEmailFieldEdit} />
```
aliceriot: I think it would be best here to do something like what we had before, i.e. just pass down an `onChange` function that can be used in the `AutomaticEmailOptions` component without too much fuss.

aliceriot: I think ideally the function we pass down here would just take a boolean and then call `updateEmailFieldEdit` correctly, then the `handleRadioClick` function in `AutomaticEmailOptions` would just be a slim wrapper around that.

static/js/components/email/AutomaticEmailOptions.js
```diff
@@ -0,0 +1,55 @@
...
+  props: {
+    automaticEmailType:    ?string,
```
aliceriot: as I mentioned below, I think here we should just pass down `sendAutomaticEmails: boolean`

static/scss/email-composition-dialog.scss
```diff
@@ -1,67 +1,133 @@
...
+.email-composition-dialog-wrapper {
+  overflow: auto;
 
-      .subheading-to {
-        padding: 0 10px;
+  .email-composition-dialog {
+    padding-top: 0 !important;
 
+    > div {
+      &:first-child {
```
aliceriot: I think you can just do `> div:first-child` here

aliceriot: the selector below is `> div:first-child > div`, which is sort of abstract - I would suggest trying to figure out a way to refactor that to use a simple class, which would be more readable in the future.

static/scss/email-composition-dialog.scss
```diff
@@ -1,67 +1,133 @@
...
+  .one-time-email {
+    display: flex !important;
+    flex-direction: row;
+    flex-grow: 1;
+    width: 55% !important;
 
-      &.email-subject {
-        margin-bottom: 5px;
+    > div {
+      > div {
```
aliceriot: same comment here and below, I think `> div > div` is a bit of a strange selector to encounter, a simple class or id would be more readable.

aliceriot: (provided it makes sense and is reasonably straightforward to do that)

amir-qayyum-khan:
Done @aliceriot

static/js/components/email/AutomaticEmailOptions.js
```diff
@@ -28,15 +25,19 @@ export default class EmailCompositionType extends React.Component {
...
-    const { automaticEmailType } = this.props;
+    const { sendAutomaticEmails } = this.props;
 
     return (
       <div className="email-type">
         <RadioButtonGroup
           className="type-radio-group"
           name="email-composition-type"
-          valueSelected={automaticEmailType}
+          valueSelected={this.getEmailType(sendAutomaticEmails)}
```
aliceriot: I would suggest that here we just use `sendAutomaticEmails` for `valueSelected`, and we use `true` and `false` for the value of the two `RadioButtons` - sorry if I wasn't clear enough, but this is what I was suggesting before.

aliceriot: get rid of `EMAIL_CAMPAIGN` and `ONE_TIME_EMAIL` completely, in other words, is my suggestion

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -44,11 +43,19 @@ export default class EmailCompositionDialog extends React.Component {
     }
   };
 
-  renderAutomaticEmailSettings =
-    (sendAutomaticEmails: boolean, updateEmailFieldEdit: () => void): React$Element<*> => (
+  setAutomaticEmailType = (sendAutomaticEmails: boolean): void => {
```
aliceriot: I would suggest we call this `setSendAutomaticEmails` instead

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -43,32 +43,30 @@ export default class EmailCompositionDialog extends React.Component {
...
-            value: e.target.checked
-          }
-        });
-      }}
-    />;
-  };
+  renderAutomaticEmailSettings = (sendAutomaticEmails: boolean): React$Element<*> => (
+    <AutomaticEmailOptions
+      sendAutomaticEmails={sendAutomaticEmails}
+      setAutomaticEmailType={this.setAutomaticEmailType} />
```
aliceriot: style nit: can you put the closing JSX bracket on it's own line? like

```js
<MultiLineComponent
  foo='foo'
  bar='bar'
/>
```

instead of

```js
<MultiLineComponent
  foo='foo'
  bar='bar' />
```

amir-qayyum-khan:
Done @aliceriot

static/scss/email-composition-dialog.scss
```diff
@@ -1,67 +1,125 @@
-.email-composition-dialog {
-  .email-composition-contents {
-    display: flex;
-    flex-direction: column;
+.email-composition-dialog-wrapper {
```
aliceriot: is the stuff here copied from your other PR? it would be good to just keep it to what we need to style the stuff in this PR I think

aliceriot: (if it is, that is)

static/js/components/email/AutomaticEmailOptions.js
```diff
@@ -0,0 +1,46 @@
...
+        <RadioButtonGroup
+          className="type-radio-group"
+          name="email-composition-type"
+          valueSelected={sendAutomaticEmails}
+          onChange={this.handleRadioClick}
+        >
+          <RadioButton
+            value={false}
+            label="Send a one-time email"
+            className="one-time-email" />
```
aliceriot: I pointed this out below in the scss file, but I don't think we need classes on these `RadioButton` elements

static/js/components/email/AutomaticEmailOptions_test.js
```diff
@@ -0,0 +1,60 @@
...
+  );
+
+  it('div renders', () => {
+    let wrapper = renderComponent();
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
+  });
+
+  it('div renders for type: one time email', () => {
+    let wrapper = renderComponent();
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
```
aliceriot: this assertion is in the first test, so it's redundant here.

static/js/components/email/AutomaticEmailOptions_test.js
```diff
@@ -0,0 +1,60 @@
...
+    // test setEmailCompositionType is called when one time email selected
+    let radioOneTime = wrapper.find(".one-time-email input");
+    radioOneTime.simulate('change');
+    assert.isTrue(setEmailCompositionTypeStub.called, "called set email composition type handler");
+  });
+
+  it('div renders for type: email campaign', () => {
+    let wrapper = renderComponent(true);
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
```
aliceriot: again, this is redundant

static/js/components/email/AutomaticEmailOptions_test.js
```diff
@@ -0,0 +1,60 @@
...
+  it('div renders for type: one time email', () => {
+    let wrapper = renderComponent();
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
+    // test setEmailCompositionType is called when one time email selected
+    let radioOneTime = wrapper.find(".one-time-email input");
+    radioOneTime.simulate('change');
+    assert.isTrue(setEmailCompositionTypeStub.called, "called set email composition type handler");
+  });
+
+  it('div renders for type: email campaign', () => {
```
aliceriot: this should be renamed to reflect what it's test (that if we have `sendAutomaticEmails == true` we see the message)

static/js/components/email/AutomaticEmailOptions_test.js
```diff
@@ -0,0 +1,60 @@
...
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
+    assert.include(
+      wrapper.text(),
+      'This email will be sent now and in the future whenever users meet the criteria.'
+    );
+    // test setEmailCompositionType is called when campaign selected
+    let radioCampaign = wrapper.find(".email-campaign input");
+    radioCampaign.simulate('change');
+    assert.isTrue(setEmailCompositionTypeStub.called, "called set email composition type handler");
+  });
```
aliceriot: I don't think we need to test twice that the `onChange` callback fires

static/js/components/email/AutomaticEmailOptions_test.js
```diff
@@ -0,0 +1,60 @@
...
+      </MuiThemeProvider>
+    )
+  );
+
+  it('div renders', () => {
+    let wrapper = renderComponent();
+    assert.equal(wrapper.find(".type-radio-group").children().length, 2);
+  });
+
+  it('div renders for type: one time email', () => {
```
aliceriot: could you rename this test to just be about firing the `onChange` handler? also generally we try to name tests something like `it('should render a div')`.

static/js/components/email/EmailCompositionDialog_test.js
```diff
@@ -64,6 +65,35 @@ describe('EmailCompositionDialog', () => {
     assert.equal(document.querySelector('h3').textContent, 'Test Email Dialog');
   });
 
+  it('renders radio button for email type config', () => {
```
aliceriot: could we name this test more explicitly? something like `renders a radio button for setting whether an email is automatic or not` or similar.

aliceriot: it would be good to also check that checking the checkbox actually updates the value, either in this test or in another one

static/js/components/email/EmailCompositionDialog_test.js
```diff
@@ -64,6 +65,35 @@ describe('EmailCompositionDialog', () => {
...
+          updateEmailFieldEdit={() => (updateStub)}
+          closeAndClearEmailComposer={closeStub}
+          closeEmailComposerAndSend={sendStub}
+          dialogVisibility={true}
+          activeEmail={emailState}
+          title={TEST_EMAIL_CONFIG.title}
+          subheadingRenderer={TEST_EMAIL_CONFIG.renderSubheading}
+        />
+      </MuiThemeProvider>
+    );
```
aliceriot: can you use the `renderDialog` function here?

static/scss/email-composition-dialog.scss
aliceriot: oh sorry, I got confused! It looks like this is copied to your other PR too, I wasn't sure which one I was looking at >_<

amir-qayyum-khan: it was main pr, 
without this setting, when user select the automatic email, ui will be damage .

static/scss/email-composition-dialog.scss
```diff
@@ -1,67 +1,125 @@
...
+    > div > div {
+      margin-right: 6px !important;
     }
   }
+
+  .email-campaign-content {
+    margin: 10px 5px 0;
+    font-size: 14px;
+  }
 }
```
amir-qayyum-khan: i am using these classes in tests

static/js/components/email/AutomaticEmailOptions.js
amir-qayyum-khan: i am using them in test file

static/scss/email-composition-dialog.scss
aliceriot: the `find` method provided by Enzyme will return all items that match, so you shouldn't need the classnames to get the `input` elements.

static/scss/email-composition-dialog.scss
aliceriot: gotcha

amir-qayyum-khan:
Done @aliceriot

static/scss/email-composition-dialog.scss
```diff
@@ -1,67 +1,105 @@
...
+  margin: 0 0 10px;
+  background-color: $email-composition-type;
+  padding: 15px;
+  border-radius: 8px;
 
-    .email-automatic {
-      padding: 20px 0;
+  .type-radio-group {
+    display: flex;
+    flex-direction: row;
```
aliceriot: real quick change: I think we want to set `overflow: hidden` on this class, overwise the width of the radio options causes a scrollbar to appear.
overflow-hidden

PR #2980, opened 2017-03-30
Do percolate on document instead of document id

### LINKED ISSUE #2959: Race condition when mailing automatic emails for new users

noisecapella:
Fixes #2959 

#### What's this PR do?
Changes percolate to use the serialized document instead of a document id. This should allow the percolation to work even when the original document has not been fully indexed in Elasticsearch yet.

None, the bug cannot be easily reproduced.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -4,6 +4,7 @@ Functions for executing ES searches
 from django.conf import settings
 from elasticsearch_dsl import Search, Q
 
+from dashboard.models import ProgramEnrollment
 from roles.api import get_advance_searchable_programs
 from search.connection import (
     get_default_alias,
@@ -12,7 +13,11 @@ from search.connection import (
     USER_DOC_TYPE,
 )
 from search.models import PercolateQuery
-from search.exceptions import NoProgramAccessException
+from search.exceptions import (
+    NoProgramAccessException,
+    PercolateException,
+)
+from search.indexing_api import serialize_program_enrolled_user
 
 DEFAULT_ES_LOOP_PAGE_SIZE = 100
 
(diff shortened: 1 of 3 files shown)


search/api.py
```diff
@@ -162,7 +167,15 @@ def search_percolate_queries(program_enrollment_id):
         django.db.models.query.QuerySet: A QuerySet of PercolateQuery matching the percolate results
     """
     conn = get_conn()
-    result = conn.percolate(get_default_alias(), USER_DOC_TYPE, id=program_enrollment_id)
+    enrollment = ProgramEnrollment.objects.get(id=program_enrollment_id)
+    doc = serialize_program_enrolled_user(enrollment)
```
rhysyngsun: This concerns me slightly because we have identified that we spend a lot of time indexing in the serialization of the enrollment and adding another call to `serialize_program_enrolled_user` here increases the overhead of the tasks `index_program_enrolled_users` and `index_users` in `search/tasks.py`.  I worry that this change will effectively double the runtime of those tasks and also increase load on the DB. because we're hitting it twice for the same data for the same enrollment.

noisecapella: Those tasks already have a lot of overhead since they are triggered when a single user's state changes. I'm not sure the extra `serialize_program_enrolled_user` call will change much. It won't slow down `recreate_index` because we don't trigger automatic emails for that case.

If it did become a bottleneck we could refactor `index_users` and `index_program_enrolled_users` to take the serialized dicts instead of the `ProgramEnrollment` objects, but I'm not sure what affect that would have on readability

rhysyngsun: Ok, given that the code looks good, we'll just need to keep an eye our for an changes to those task usages.

Issue #2718, opened 2017-02-28
Refactor tasks to pass database ids instead of object instances


noisecapella:
After talking with @rhysyngsun there is the potential for serialization and memory use problems when passing instances to celery workers. https://github.com/mitodl/micromasters/pull/2701/files#r103470608

We need to do this in two steps to preserve the ability to deal with old Celery tasks which are still in the queue:
 - refactor task to take an id or an object instance, and pass only ids
 - After all old Celery tasks are completed remove the code that handles the object instance

noisecapella:
We need to use [transaction.on_commit](https://docs.djangoproject.com/en/1.10/topics/db/transactions/#performing-actions-after-commit) inside the  signals to ensure that the id points to a valid row in the database

PR #2984, opened 2017-03-31
Pass object ids to Celery tasks instead of objects

noisecapella:
Part of #2718 

#### What's this PR do?
Refactors tasks to use object ids instead of objects, to avoid possible race conditions and serialization problems, especially when deleting objects. 

Update your profile then go to the learners page to verify that the update indexed properly.

The next PR will remove the deprecated logic which is only needed for Celery tasks which are still in the queue during the deploy.


grades/tasks.py
```diff
@@ -93,7 +96,7 @@ def freeze_course_run_final_grades(course_run):
 
     # create a group of subtasks to be run in parallel
     job = group(
-        freeze_users_final_grade_async.s(list_users, course_run) for list_users in chunks(users_qset)
+        freeze_users_final_grade_async.s(list_users, course_run.id) for list_users in chunks(users_qset)
```
noisecapella: Yes, I'll make this change

grades/tasks.py
```diff
@@ -34,20 +35,22 @@ def find_course_runs_and_freeze_grades():
...
     Async task manager to freeze all the users' final grade in a course run
 
     Args:
-        course_run (CourseRun): a course run model object
+        course_run_id (int): a course run model object
 
     Returns:
         None
     """
+    course_run_id = lookup_id(course_run_id)
```
noisecapella: Sure

grades/tasks.py
```diff
@@ -34,20 +35,22 @@ def find_course_runs_and_freeze_grades():
...
 
 @async.task
-def freeze_course_run_final_grades(course_run):
+def freeze_course_run_final_grades(course_run_id):
     """
     Async task manager to freeze all the users' final grade in a course run
 
     Args:
-        course_run (CourseRun): a course run model object
+        course_run_id (int): a course run model object
```
noisecapella: Sure

search/tasks.py
```diff
@@ -19,25 +19,41 @@
...
     Remove program-enrolled user from index
 
     Args:
-        user (User): A program-enrolled user to remove from index
+        program_enrollment_id (int): A ProgramEnrollment to remove from the index
     """
-    _remove_program_enrolled_user(user)
+    # Deprecation warning!
+    program_enrollment_id = lookup_id(program_enrollment_id)
+    _remove_program_enrolled_user(program_enrollment_id)
```
noisecapella: I'm only using it when the tasks are being dispatched, so in the calling functions to here.

dashboard/signals.py
```diff
@@ -13,12 +14,12 @@ def handle_create_programenrollment(sender, instance, created, **kwargs):  # pyl
     """
     When a ProgramEnrollment model is created/updated, update index.
     """
-    index_program_enrolled_users.delay([instance])
+    transaction.on_commit(lambda: index_program_enrolled_users.delay([instance.id]))
```
noisecapella: I didn't either, I just happened upon it because the tests run in transactions so the tests were failing when looking up by id numbers

noisecapella:
Ok, I think that's everything

noisecapella:
@giocalitri I'm also going to set `CELERY_TASK_SERIALIZER` to `json` in this PR so that by the time we merge the other PR all new tasks are JSON and we can set `CELERY_ACCEPT_CONTENT`

noisecapella:
@giocalitri Any other comments?

giocalitri:
Do we need to update also the tasks in the dashboard app?

giocalitri:
Apart from the dashboard tasks the rest of the code appears to be fine

giocalitri:
functionality is fine.

noisecapella:
The dashboard tasks already appear to use user ids so no update is necessary there

giocalitri:
you are right: too tired last night 👍

Issue #2953, opened 2017-03-24
Mail: display search criteria in bulk mail dialog


pdpinch:
Part of #2812 

As a reminder to the user, display the current search criteria in the mail dialog box in the recipients field.

![scheduled_email 2x](https://cloud.githubusercontent.com/assets/20047260/24111439/c41c0d50-0d6c-11e7-8bfd-da02ccc346a9.jpg)

PR #2992, opened 2017-04-03
Mail: displayed search filters as recipients

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2953

#### What's this PR do?
- Added filers selected on learners search to email composer as recipients

- Go to `/learners/` select filters and open `Email These Learners`.

screen shot 2017-04-03 at 10 16 05 pm

### DIFF
--- a/static/js/components/email/lib.js
+++ b/static/js/components/email/lib.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
+import _ from 'lodash';
 
 import {
   sendCourseTeamMail,
@@ -46,6 +47,7 @@ export const SEARCH_RESULT_EMAIL_CONFIG: EmailConfig = {
   emailOpenParams: (searchkit: Object) => ({
     subheading: `${searchkit.getHitsCount() || 0} recipients selected`,
     supportsAutomaticEmails: true,
+    filters: searchkit.query.getSelectedFilters()
   }),
 
   getEmailSendFunction: () => sendSearchResultMail,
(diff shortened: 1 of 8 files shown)


aliceriot:
functionality generally looks good!

one small styling regression. here's what the subject line looks like on `master`:

fjfjfjf

and on your branch:

fjfjfjfj

we lost a little vertical padding, so the text is smooshed at the top.

amir-qayyum-khan:
fixed @aliceriot 

screen shot 2017-04-04 at 8 04 00 pm

amir-qayyum-khan:
@aliceriot please review

static/js/components/email/lib.js
```diff
@@ -55,7 +57,30 @@ export const SEARCH_RESULT_EMAIL_CONFIG: EmailConfig = {
...
-  ])
+  ]),
+
+  renderRecipients: (filters?: Array<any>) => {
+    if (!filters || filters.length <= 0) {
+      return null;
+    }
+    const renderFilterOptions = _.map(filters, filter => (
+      <div className="sk-selected-filters-option sk-selected-filters__item" key={filter.id}>
+        <div className="sk-selected-filters-option__name">
```
aliceriot: are we using these searchkit classes here to re-use some of their styling?

static/js/components/email/lib.js
```diff
@@ -55,7 +57,30 @@ export const SEARCH_RESULT_EMAIL_CONFIG: EmailConfig = {
...
     emailState.searchkit.buildQuery().query,
     emailState.inputs.sendAutomaticEmails || false,
-  ])
+  ]),
+
+  renderRecipients: (filters?: Array<any>) => {
+    if (!filters || filters.length <= 0) {
+      return null;
+    }
+    const renderFilterOptions = _.map(filters, filter => (
```
aliceriot: It would be good to avoid creating this function at runtime - an easy way would be to use `R.map` instead, then it could be declared, for instance, at the top of this module, like:

```js
const renderFilterOptions = R.map(filter => (
  <div className="sk-selected-filters-option sk-selected-filters__item" key={filter.id}>
    <div className="sk-selected-filters-option__name">
      {filter.name}: {filter.value}
    </div>
  </div>
));
```

then `renderFilterOptions` will be a function that takes an array of filters and returns an array of React components, and we can avoid re-creating that function every time we call `renderRecipients`.

static/js/flow/emailTypes.js
```diff
@@ -19,6 +19,7 @@ export type EmailState = {
   sendError:                EmailSendError,
   fetchStatus?:             ?string,
   supportsAutomaticEmails?: boolean,
+  filters:                  ?Array<any>,
```
aliceriot: would it be doable to make a `Filter` type for flow, so we could say `?Array<Filter>` here instead of `?Array<any>`?

static/js/components/email/lib.js
amir-qayyum-khan: yes

amir-qayyum-khan:
Done @aliceriot

PR #2994, opened 2017-04-03
Exams: result/request file audit trail

### LINKED ISSUE #2895: Exams: keep an audit trail of files we send to Pearson
It's come up a couple of times that we need to see what exactly we sent to the SFTP server. Probably just store them in an S3 bucket with a TTL of a month or so.

### LINKED ISSUE #2896: Exams: keep an audit trail of files we receive from Pearson SFTP
We might want to reprocess these in the future or check them to verify we processed them correctly. Probably just store the zip files we process in an S3 bucket. They should already be named with a date in them so we can just keep the filename. For now, we want to keep these around indefinitely.

Acceptance criteria:
- [ ] files are encrypted before sending to S3 using GPG (application only gets the public key)
- [ ] files are encrypted at rest in S3
- [ ] files are named / pathed with the date on which we received them

rhysyngsun:
Fixes #2896 and fixes #2895 

#### What's this PR do?
This adds "auditing" (encrypting and storing of files to S3) for files we send and receive for exams.

###### Set up your .env
- First, you need to generate a GPG key `gpg --gen-key`. Follow the prompts.
- Get the key fingerprint `gpg --list-keys --fingerprint`, put this in `.env` as `EXAMS_AUDIT_ENCRYPTION_FINGERPRINT`
- Export the gpg key `gpg --armor --export <EMAIL> | base64 -w 0` (use email you entered in step 1), put this as `EXAMS_AUDIT_ENCRYPTION_PUBLIC_KEY` 
- Requisition yourself values for `EXAMS_AUDIT_S3_BUCKET`, `EXAMS_AUDIT_AWS_ACCESS_KEY_ID`, and `EXAMS_AUDIT_AWS_SECRET_ACCESS_KEY` from me

###### Test it in shell
Easiest way to test this is in shell, because otherwise you have to simulate the entire request/response flow locally which is a pain.

- First, write a test file to your project root. Can have whatever contents.
- In `./manage.py shell`:

```python
from exams.pearson.audit import ExamDataAuditor
from django.test import override_settings
auditor.configure() # should return without error
with override_settings(EXAMS_AUDIT_ENABLED=True):
     auditor.audit_request('/src/my_test_file.txt') # replace with your unique filename
```
- At this point, the file should be on the RC S3 bucket. To verify:
```python
b = auditor.store.get_bucket()
keys = b.get_all_keys(max_keys=100)
keys
```
- You should see a key with your file printed under the `exam_audits/request/` path.
- To verify contents:
```python
with open('/src/encrypted_file.gpg', 'wb') as f:
   key = keys[0] # whatever index it appears in
   key.get_contents_to_file(f)
```
- You should now have `encrypted_file.gpg` in your project root. `cat` to verify it isn't plaintext.
- Decrypt it to be sure it was encrypted correctly: `gpg --output decrypted.txt --decrypt encrypted.gpg`
- Clean up your all these files once you've verified, sorry for the mess.

exams/pearson/audit.py

This provides a trail of what happened historically in case questions come up regarding what we sent or what we received.

https://giphy.com/gifs/JTSs32Pa9iXYc/


exams/pearson/audit.py
```diff
@@ -0,0 +1,265 @@
...
+            )
+
+        if self.fingerprint in self.get_fingerprints():
+            return
+
+        result = self.gpg.import_keys(self.public_key)
+
+        counts = dict(result.counts)
+
+        if counts['not_imported'] != 0:
```
singingwolfboy: I don't understand what's going on here. What is the `result` variable that is returned from `gpg.import_keys()`? Why are you transforming `result.counts` into a dictionary? Why do you say that it's an error if `counts['not_imported']` isn't zero? If there were an error, shouldn't the `gpg.import_keys()` function raise an exception?

exams/pearson/audit.py
```diff
@@ -0,0 +1,265 @@
...
+            raise ImproperlyConfigured('Error importing GPG keys for exam auditing: {}'.format(result.summary()))
+
+        if self.fingerprint not in result.fingerprints:
```
singingwolfboy: Surely the GPG module would have some built-in method of verifying fingerprints, doesn't it? It would be better to do the standard thing here.

exams/pearson/audit.py
```diff
@@ -0,0 +1,265 @@
...
+                self.fingerprint,
+                armor=False,
+                symmetric=False,
+                output=encrypted_filename
+            )
+
+        try:
+            yield encrypted_filename
+        finally:
+            self._cleanup(encrypted_filename)
```
singingwolfboy: If you want to make sure that `_cleanup()` is called to clean up any encrypted file that might have been generated, wouldn't it be better to put the `gpg.encrypt()` call inside the `try` block? The way this code is currently written, there could be an exception in that call after the file is already written, and `_cleanup()` wouldn't get called.

exams/pearson/audit_test.py
```diff
@@ -0,0 +1,254 @@
...
+    with override_settings(**settings):
+        yield settings
+
+
+@pytest.fixture
+def auditor():
+    """
+    ExamDataAuditor that cleans up the GPG keys after itself
+    """
+    import gnupg
```
singingwolfboy: Shouldn't this import go at the top of this file?

exams/pearson/audit.py
```diff
@@ -0,0 +1,265 @@
...
+SSE_ENCRYPTION_ALGORITHM = 'AES256'
+REQUIRED_SETTINGS = [
+    'EXAMS_AUDIT_ENCRYPTION_FINGERPRINT',
+    'EXAMS_AUDIT_ENCRYPTION_PUBLIC_KEY',
+]
+
+log = logging.getLogger(__name__)
+
+
+class S3AuditStorage:
```
rhysyngsun: I wanted to have the option of swapping storage mechanisms down the road, a class-based design makes this easier.

exams/pearson/audit.py
```diff
@@ -0,0 +1,265 @@
...
+        s3_key.set_contents_from_filename(
+            filename,
+            headers={
+                'x-amz-server-side-encryption': SSE_ENCRYPTION_ALGORITHM,
+            }
+        )
+        return s3_key.key
+
+
+class ExamDataAuditor:
```
rhysyngsun: Using a class was intentional, although you're pointing out one oversight I forgot to address - I want to only configure this class and cache that state in memory because we could be potentially auditing a lot of files in one run and I'd prefer not to keep performing IO to inspect state that shouldn't change under normal circumstances. I need to store that state somewhere so a class is best for that.

exams/pearson/audit.py
rhysyngsun: That method doesn't raise an error (it probably does raise low-level IO errors), instead it returns a [`gnupg._parsers.ImportResult`](https://github.com/isislovecruft/python-gnupg/blob/master/gnupg/_parsers.py#L1087). I can't speak to the author's design decisions, since it's calling the `gpg` CLI, they probably though it best to return a parsed response of the command output rather than erroring.

I honestly can't defend the cast of `results.count` (an `OrderedDict`) to a `dict`, so that can be removed.

exams/pearson/audit.py
rhysyngsun: I'm not sure I understand (such a method doesn't exist in the lib anyway)  - here I'm verifying the fingerprint matches the fingerprint of the key I just imported based on the result. If those don't match, it means we have a mismatch on the configured fingerprint and the key.

exams/pearson/audit.py
rhysyngsun: Good catch, must've messed that up in my many refactors.

exams/pearson/audit.py
```diff
@@ -0,0 +1,265 @@
...
+            ImproperlyConfigured
+                if a required setting is not set
+        Returns:
+            boto.s3.connection.S3Connection
+                a connection to S3
+        """
+        if any([key is None for key in [
+                settings.EXAMS_AUDIT_AWS_ACCESS_KEY_ID,
+                settings.EXAMS_AUDIT_AWS_SECRET_ACCESS_KEY,
+        ]]):
```
rhysyngsun: I made the same change for line 157.

exams/pearson/audit.py
singingwolfboy: Ah, OK. If the library doesn't have anything built-in to do this sort of thing, then this is fine.

gsidebo:
Functionality looks great

exams/pearson/audit_test.py
```diff
@@ -0,0 +1,281 @@
...
+    mocker.spy(auditor.gpg, 'import_keys')
+    auditor.configure()
+    assert auditor.gpg.import_keys.call_count == 0
+    assert auditor.configured is True
+
+
+def test_exam_data_auditor_enabled(auditor, missing_settings, mocker):
+    """Test that audit() respected the enabled flag"""
+    mocker.patch.object(auditor, 'configure')
+    mocker.patch.object(auditor, 'upload_encrypted_file')
```
gsidebo: can you use `patch.multiple` instead of 2 `patch.object` calls here? http://www.voidspace.org.uk/python/mock/patch.html#patch-multiple

exams/pearson/audit.py
```diff
@@ -0,0 +1,272 @@
...
+                the path where the file was stored
+        """
+        if not settings.EXAMS_AUDIT_ENABLED:
+            return None
+
+        self.configure()
+
+        return self.upload_encrypted_file(filename, file_type)
+
+    def audit_response(self, filename):
```
gsidebo: nit: can you call these methods `audit_response_file` and `audit_request_file` respectively?

exams/pearson/audit_test.py
```diff
@@ -0,0 +1,281 @@
...
+    mock = mocker.patch.object(auditor.gpg, 'import_keys')
+    mock.return_value.counts = {
+        'not_imported': 1,
+    }
+    with pytest.raises(ImproperlyConfigured):
+        auditor.configure()
+    assert auditor.configured is False
+
+
+def test_exam_data_s3_store_upload(s3_store, mocker):
```
gsidebo: I think this test case is confusing me the most out of all of them, so I think it should have a docstring 😄 . Also, it seems like this is a straightforward test of `exams.pearson.audit.S3AuditStorage#upload` and not a test for any `ExamDataAuditor` functionality. If that's true, can you move this to be nearer to the other `S3AuditStorage`-specific test cases?

exams/pearson/audit.py
```diff
@@ -0,0 +1,272 @@
...
+            raise ImproperlyConfigured(
+                "One of EXAMS_AUDIT_ENCRYPTION_PUBLIC_KEY or "
+                "EXAMS_AUDIT_ENCRYPTION_FINGERPRINT is required but not set"
+            )
+
+        if self.fingerprint in self.get_fingerprints():
+            self.configured = True
+            return
+
+        print(self.public_key)
```
rhysyngsun: Remove the `print` statement, silly.

exams/pearson/audit_test.py
```diff
@@ -0,0 +1,281 @@
...
+    assert mock_key.set_contents_from_filename.call_args == call(
+        'filename',
+        headers={
+            'x-amz-server-side-encryption': 'AES256'
+        }
+    )
+
+
+def test_exam_data_auditor_audit(auditor):
+    """Test that the auditor encrypts and uploads the file"""
```
gsidebo: can you add notes in the docstring or inline comments about the fact that (a) `.audit()` results in an upload, and (b) this is testing that the file is removed after uploading?

exams/pearson/audit_test.py
```diff
@@ -0,0 +1,281 @@
...
+    with tempfile.NamedTemporaryFile() as audit_file:
+        expected_encrypted_filename = '{}.gpg'.format(audit_file.name)
+        expected_encrypted_keypath = 'keypath'  # computed inside _upload_to_s3
+
+        def side_effect(encrypted_filename, file_type):
```
gsidebo: can you rename to `upload_side_effect`?

conftest.py
```diff
@@ -0,0 +1,15 @@
+"""Configure pytest"""
```
gsidebo: it looks like this fixture is only needed in the `exams` app. if that's right, can you move this `conftest.py` file to that directory?

exams/tasks.py
```diff
@@ -129,6 +139,15 @@ def export_exam_authorizations(self):
...
         try:
+            audit.ExamDataAuditor().audit_request(tsv.name)
+        except ImproperlyConfigured:
+            log.exception('Exam auditing improperly configured')
+            return
+        except:  # pylint: disable=bare-except
+            log.exception('Unexpected error auditing EAD file')
+            return
+
+        try:
```
gsidebo: lots of repetition in this file, but it looks like that existed before this PR and could/should be addressed separately

exams/pearson/audit.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Args:
+            filename (str):
+                filename of the encrypted file
+            file_type (str):
+                type of the encrypted file
+
+        Returns:
+            boto.s3.key.Key:
+                the key to store the file in
```
rhysyngsun: I've seen both styles (and looking in search/indexing_api.py I see some stlye mixing, which is even worse), perhaps we can collectively decide on which to standardize?

exams/pearson/audit.py
```diff
@@ -0,0 +1,272 @@
...
+                    source_file,
+                    self.fingerprint,
+                    armor=False,
+                    symmetric=False,
+                    output=encrypted_filename
+                )
+
+            return self.store.upload(encrypted_filename, file_type)
+        finally:
+            self._cleanup(encrypted_filename)
```
rhysyngsun: I don't know if it's a recommended pattern per se, but it's mentioned in the standard docs: https://docs.python.org/3/reference/compound_stmts.html#the-try-statement
 
> When a return, break or continue statement is executed in the try suite of a try...finally statement, the finally clause is also executed ‘on the way out.’ A continue statement is illegal in the finally clause. (The reason is a problem with the current implementation — this restriction may be lifted in the future).

rhysyngsun:
@gsidebo feedback addressed

PR #2997, opened 2017-04-04
Fixed unexpected course enrollment counts/results in learners search

### LINKED ISSUE #2942: Search: Semester facet is not restricted to just selected course
Consider a search query like https://micromasters.mit.edu/learners/?courses[0][0]=Microeconomics&courses[1][0]=Paid&semester[0]=2017%20-%20Spring&sort=grade-high-low -- the selected facets are microeconomics, paid and Spring 2017. The user assumes that the result set will be filtered down to learners who have paid for a spring 2017 offering of microeconomics. 

However, as currently implemented, this includes users like https://micromasters.mit.edu/learner/NoraWU who are not enrolled in the spring 2017 run _of microeconomics_. This learner is in the results set because they a) have enrolled in a run of microeconomics and b) have enrolled in a spring 2017 course run (it happens to be Global Poverty).

This is not intuitive for our staff users. They expect that filtering on microeconomics and spring 2017 will return results filtered specifically on microeconomics spring 2017, and not just the union microeconomics and spring 2017. 

For simplicity, I don't think this requires a change to the UI.

gsidebo:
Fixes #2942

#### What's this PR do?
Fixes the bug we had where search results were being filtered in ways that our users weren't expecting. Filters on our course enrollments were being treated as 'OR' filters rather than 'AND', and the number of results being reported was a count of the nested documents that matched rather than the number of users.

I'll add some detail here later, but in general, you want to get users into a state where there is a venn-diagram-like overlap between course enrollments based on course title, course payment statuses, and enrolled semesters.

Probably dashboard/serializers.py and static/js/components/search/NestedAggregatingMenuFilter.js

1. The per-course final grade histogram was removed and will be reimplemented in another issue.
1. The course payment status was moved to its own facet

ss 2017-04-04 at 09 11 01

### DIFF
--- a/dashboard/serializers.py
+++ b/dashboard/serializers.py
@@ -10,70 +10,66 @@ class UserProgramSearchSerializer:
     """
     Provides functions for serializing a ProgramEnrollment for the ES index
     """
+    PAID_STATUS = 'Paid'
+    UNPAID_STATUS = 'Auditing'
+
     @classmethod
     def serialize_enrollments(cls, mmtrack):
         """
-        Serializes a user's enrollment data for search results
+        Serializes a user's enrollment data for search results in such a way that enrollments
+        in multiple runs of a single course will result in just one serialization.
 
         Args:
             mmtrack (MMTrack): An MMTrack object
         Returns:
-            list: Serialized courses
+            list: Serialized course enrollments
         """
-        enrollment_status_map = {}
-        course_runs = mmtrack.get_all_enrolled_course_runs()
-        for course_run in course_runs:
-            course_title = course_run.course.title
-            is_verified = mmtrack.is_enrolled_mmtrack(course_run.edx_course_key)
-            # If any course run for this course was verified/paid, maintain the verified status
-            enrollment_status_map[course_title] = enrollment_status_map.get(course_title) or is_verified
-        serialized_enrollments = []
-        for course_title, is_verified in enrollment_status_map.items():
-            enrollment_status = 'Paid' if is_verified else 'Auditing'
-            serialized_enrollments.extend([
-                {'level': 1, 'value': course_title, 'ancestors': []},
-                {'level': 2, 'value': enrollment_status, 'ancestors': [course_title]}
-            ])
-        return serialized_enrollments
-
-    @classmethod
-    def serialize_final_grades(cls, mmtrack):
-        """
-        Serializes a user's final grades
-
-        Args:
-            mmtrack (MMTrack): An MMTrack object
-        Returns:
-            list: Serialized final grades
-        """
-
-        serialized_final_grades = []
-        final_grades = mmtrack.get_all_final_grades()
-        for final_grade in final_grades.values():
-            serialized_final_grades.append({
-                'title': final_grade.course_run.course.title,
-                'grade': final_grade.grade_percent
-            })
-        return serialized_final_grades
-
-    @classmethod
-    def serialize_semester_enrollments(cls, mmtrack):
-        """
-        Serializes the semesters that a user is enrolled in
-
-        Args:
-            mmtrack (MMTrack): An MMTrack object
-        Returns:
-            list: Serialized semester enrollments
-        """
-        semester_values = set()
+        serialized_enrollments_map = {}
         for course_run in mmtrack.get_all_enrolled_course_runs():
-            year_season_tuple = get_year_season_from_course_run(course_run)
-            if year_season_tuple:
-                semester_values.add(
-                    '{} - {}'.format(year_season_tuple[0], year_season_tuple[1])
-                )
-        return [{'semester': semester_value} for semester_value in semester_values]
+            course_title = course_run.course.title
+            # If any course run for this course was verified/paid, count it as verified
+            if course_title not in serialized_enrollments_map or \
+                    serialized_enrollments_map[course_title]['payment_status'] == cls.UNPAID_STATUS:
+                serialized_enrollments_map[course_title] = cls.serialize_enrollment(mmtrack, course_run)
+        return list(serialized_enrollments_map.values())
+
+    @classmethod
+    def serialize_enrollment(cls, mmtrack, course_run):
+        """
+        Serializes information about a user's enrollment in a course run
+
+        Args:
+            mmtrack (MMTrack): An MMTrack object
+            course_run (CourseRun): An enrolled CourseRun
+        Returns:
+            dict: Serialized course enrollment
+        """
+        course_title = course_run.course.title
+        has_paid = mmtrack.has_paid(course_run.edx_course_key)
+        payment_status = cls.PAID_STATUS if has_paid else cls.UNPAID_STATUS
+        final_grade = mmtrack.get_final_grade_percent(course_run.edx_course_key)
+        semester = cls.serialize_semester(course_run)
+        return {
+            'final_grade': final_grade,
+            'semester': semester,
+            'course_title': course_title,
+            'payment_status': payment_status,
+        }
+
+    @classmethod
+    def serialize_semester(cls, course_run):
+        """
+        Serializes the semester in which a user has been enrolled
+
+        Args:
+            course_run (CourseRun): An enrolled CourseRun
+        Returns:
+            str: Serialized semester enrollment or None
+        """
+        year_season_tuple = get_year_season_from_course_run(course_run)
+        if year_season_tuple:
+            return '{} - {}'.format(year_season_tuple[0], year_season_tuple[1])
+        return None
 
     @classmethod
     def serialize(cls, program_enrollment):
@@ -87,8 +83,6 @@ class UserProgramSearchSerializer:
         return {
             'id': program.id,
             'enrollments': cls.serialize_enrollments(mmtrack),
-            'final_grades': cls.serialize_final_grades(mmtrack),
-            'semester_enrollments': cls.serialize_semester_enrollments(mmtrack),
             'grade_average': mmtrack.calculate_final_grade_average(),
             'is_learner': is_learner(user, program),
             'num_courses_passed': mmtrack.count_courses_passed(),
--- a/dashboard/serializers_test.py
+++ b/dashboard/serializers_test.py
@@ -71,13 +71,6 @@ class UserProgramSearchSerializerTests(MockedESTestCase):
             factory_kwargs['data'] = data
         return [CachedEnrollmentFactory.create(course_run=course_run, **factory_kwargs) for course_run in course_runs]
 
-    @staticmethod
-    def sort_final_grades_values(final_grades):
-        """
-        Returns a sorted list of serialized final grades
-        """
-        return sorted(final_grades, key=lambda k: k['grade'])
-
     @classmethod
     def setUpTestData(cls):
         with mute_signals(post_save):
@@ -125,8 +118,6 @@ class UserProgramSearchSerializerTests(MockedESTestCase):
         non_fa_cached_edx_data = CachedEdxUserData(cls.user, program=program)
         non_fa_mmtrack = MMTrack(cls.user, program, non_fa_cached_edx_data)
         cls.serialized_enrollments = UserProgramSearchSerializer.serialize_enrollments(non_fa_mmtrack)
-        cls.semester_enrollments = UserProgramSearchSerializer.serialize_semester_enrollments(non_fa_mmtrack)
-        cls.final_grades = UserProgramSearchSerializer.serialize_final_grades(non_fa_mmtrack)
         cls.program_enrollment = ProgramEnrollment.objects.create(user=cls.user, program=program)
         # create a financial aid program
         cls.fa_program, _ = create_program()
@@ -158,8 +149,6 @@ class UserProgramSearchSerializerTests(MockedESTestCase):
         cls.fa_serialized_enrollments = (
             UserProgramSearchSerializer.serialize_enrollments(fa_mmtrack)
         )
-        cls.fa_semester_enrollments = UserProgramSearchSerializer.serialize_semester_enrollments(fa_mmtrack)
-        cls.fa_final_grades = UserProgramSearchSerializer.serialize_final_grades(fa_mmtrack)
 
     def test_full_program_user_serialization(self):
         """
@@ -170,16 +159,12 @@ class UserProgramSearchSerializerTests(MockedESTestCase):
         expected_values = {
             'id': program.id,
             'enrollments': self.serialized_enrollments,
-            'semester_enrollments': self.semester_enrollments,
-            'final_grades': self.sort_final_grades_values(self.final_grades),
             'grade_average': 75,
             'is_learner': True,
             'num_courses_passed': 1,
             'total_courses': 1
         }
         serialized_enrollments = UserProgramSearchSerializer.serialize(self.program_enrollment)
-        serialized_enrollments['final_grades'] = self.sort_final_grades_values(serialized_enrollments['final_grades'])
-
         assert serialized_enrollments == expected_values
 
     def test_full_program_user_serialization_financial_aid(self):
@@ -193,15 +178,12 @@ class UserProgramSearchSerializerTests(MockedESTestCase):
         expected_result = {
             'id': self.fa_program.id,
             'enrollments': self.fa_serialized_enrollments,
-            'semester_enrollments': self.fa_semester_enrollments,
-            'final_grades': self.sort_final_grades_values(self.fa_final_grades),
             'grade_average': 95,
             'is_learner': True,
             'num_courses_passed': 1,
             'total_courses': 2
         }
         serialized_enrollments = UserProgramSearchSerializer.serialize(self.fa_program_enrollment)
-        serialized_enrollments['final_grades'] = self.sort_final_grades_values(serialized_enrollments['final_grades'])
         assert serialized_enrollments == expected_result
 
     def test_full_program_user_serialization_staff(self):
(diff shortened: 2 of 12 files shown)


gsidebo:
A couple python tests are still failing, but this is otherwise ready to review

gsidebo:
The `seed_db` command sets all of the fake users to paid by default. The `unpay` function in this snippet can be used to set a user/course run to audit. The `alter_data` commands _should_ take care of this, but I missed the fact that an enrollment needs to be present for passed/failed courses in order for everything to be indexed properly.

```python
from ecommerce.models import Line
from grades.models import FinalGrade
from seed_data.lib import CachedEnrollmentHandler
from search.indexing_api import index_users

def unpay(username, edx_course_key):
    user = User.objects.get(username=username)
    course_run = CourseRun.objects.get(edx_course_key=edx_course_key)
    FinalGrade.objects.filter(user=user, course_run=course_run).update(course_run_paid_on_edx=False)
    Line.objects.filter(
        order__status='fulfilled', course_key=course_run.edx_course_key, order__user=user
    ).delete()
    CachedEnrollmentHandler(user).set_or_create(course_run, verified=False)
    index_users([user])
    
def get_users_enrolled_in_course_run(edx_course_key):
    course_run = CourseRun.objects.get(edx_course_key=edx_course_key)
    return (
        CachedEnrollment.objects.filter(course_run__edx_course_key=edx_course_key)
        .values_list('user__username', flat=True)
    )
```

To manually test these changes, you can set any number of users in specific course runs to be unpaid and add/remove course enrollment filters ('Course', 'Payment Status', 'Semester') and observe (1) the accuracy/count of the search results and (2) the counts displayed by the search facets, which should reflect the number of users that will be returned if selected.

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+   */
+  buildSharedQuery(query) {
+    query = super.buildSharedQuery(query);
+
+    // Construct an 'AND' query for all filters applied on this nested path
+    let groupedNestedFilter = this.createGroupedNestedFilter(query);
+    if (groupedNestedFilter) {
+      let nestedPath = this.getNestedPath();
+      // Find the element that Searchkit added to query.index.filters for this nested document and remove it
+      _.remove(query.index.filters, (filter) => (
```
amir-qayyum-khan: why you need this?

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
+import PatchedMenuFilter from './PatchedMenuFilter';
```
gsidebo: no, we don't have any tests for searchkit components yet. we probably should, but i don't think we want to get into that as part of this bug fix. @pdpinch you can chime in if you like

static/js/components/search/NestedAggregatingMenuFilter.js
gsidebo: We need to change the values that searchkit sets for `query.index.filters` and `query.index.filtersMap` in order for the filters to work the way we want them to. The comment on this function (`buildSharedQuery`) has a little detail about it

singingwolfboy:
Just found a problem where the "Course", "Payment Status", and "Semester" filters still appear, even if there are no options to filter on.
screenshot demonstrating filter bug

noisecapella:
@gsidebo You wouldn't happen to have a script to get users into a state to test this, would you?

dashboard/serializers.py
```diff
@@ -13,67 +13,60 @@ class UserProgramSearchSerializer:
...
-        serialized_final_grades = []
-        final_grades = mmtrack.get_all_final_grades()
-        for final_grade in final_grades.values():
-            serialized_final_grades.append({
-                'title': final_grade.course_run.course.title,
-                'grade': final_grade.grade_percent
-            })
-        return serialized_final_grades
+        course_title = course_run.course.title
+        is_verified = mmtrack.is_enrolled_mmtrack(course_run.edx_course_key)
```
annagav: Since we are indexing enrollments in courses in cases where users have a final grade but are missing CachedErollment, `is_enrolled_mmtrack` returns false for users that have only final grade. We might want to use 
`is_verified = mmtack.has_paid(course_run.edx_course_key)`.

gsidebo:
@noisecapella along with the code sample i posted above, you can run this to "un-pay" 15 users from 'Digital Learning 200'. you can adjust this as needed. after running this, you'll be able to select that course as a filter on the learners page, then drill down to Paid/Auditing

```python

course_to_unpay = 'Digital Learning 200'
num_users_to_unpay = 15

course = Course.objects.get(title=course_to_unpay)
edx_key = CachedEnrollment.objects.filter(course_run__course=course).first().course_run.edx_course_key
usernames_to_unpay = get_users_enrolled_in_course_run(edx_key)[0:num_users_to_unpay]
for username in usernames_to_unpay:
    unpay(username, edx_key)
```

noisecapella:
Took a little while but I confirmed that functionality works. Looking at code now

noisecapella:
It seems like there are a bunch of extra filters being added:

screenshot from 2017-04-05 16-25-42

I think only the first and the last ones should be necessary, for the program and the semester+course. I haven't read through the source yet so I'm not sure what's causing it. Here's the entire screenshot to show the filters that are applied:

screenshot from 2017-04-05 16-29-03

static/js/components/search/NestedAggregatingMenuFilter.js
noisecapella: I just merged a PR which fixes tests in `LearnerSearchPage_test` and also adds a few tests for the work history and education facets. Can we add a test to make sure the query is being constructed properly?

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+    // Construct an 'AND' query for all filters applied on this nested path
+    let groupedNestedFilter = this.createGroupedNestedFilter(query);
+    if (groupedNestedFilter) {
+      let nestedPath = this.getNestedPath();
+      // Find the element that Searchkit added to query.index.filters for this nested document and remove it
+      _.remove(query.index.filters, (filter) => (
+        _.find(filter, ['nested', 'path']) === nestedPath
+      ));
+      // Add the 'AND' filter that we constructed
+      query.index.filters.push(groupedNestedFilter);
```
noisecapella: I think `query.addFilter(nestedPath, groupedNestedFilter)` will take care of `filters` and `filtersMap`

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+      )(allTermFilters);
+    }
+    return otherTermFilters.length > 0 ? BoolMust(otherTermFilters) : undefined;
+  }
+
+  /**
+   * Gets the appropriate terms bucket for this element's agg query.
+   */
+  getTermsBucket(query) {
+    let otherAppliedFiltersOnPath = this.createFilterForOtherElementsOnPath(query);
```
noisecapella: Why are we adding the filter here and in `createAggFilter`? Wouldn't the filter be added to the aggregations if it exists in `filters` and `filtersMap`?

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * Overrides getItems in MenuFilter
+   * Before the aggregation results are rendered, set the doc_count of each item to be the
+   * "reverse nested" doc_count. This effectively means that we will show how many unique users
+   * match the query against a set of nested elements, as opposed to the total count of nested
+   * elements that match (which could be greater than the number of users).
+   */
+  getItems() {
+    let items = super.getItems();
+    items = _.map(items, (item) => {
```
noisecapella: This will modify each item which is passed in. Can we use a `_.forEach` instead to make the mutation explicit?

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+   */
+  buildOwnQuery(query) {
+    if (!this.loadAggregations) {
+      return query;
+    } else {
+      return query
+        .setAggs(FilterBucket(
+          this.uuid,
+          this.createAggFilter(query),
+          ...this.fieldContext.wrapAggregations(
```
noisecapella: What does `wrapAggregations` do?

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+      // Find the element that Searchkit added to query.index.filters for this nested document and remove it
+      _.remove(query.index.filters, (filter) => (
+        _.find(filter, ['nested', 'path']) === nestedPath
+      ));
+      // Add the 'AND' filter that we constructed
+      query.index.filters.push(groupedNestedFilter);
+      // Add the same 'AND' filter to query.index.filtersMap, with the nested path (not the uuid) as the key
+      query.index.filtersMap[nestedPath] = groupedNestedFilter;
+      // If it exists, delete the key for this specific filter (since all filters on this path are grouped together).
+      delete query.index.filtersMap[this.uuid];
```
noisecapella: I don't think we are allowed to do this, it is an ImmutableQuery. Maybe `query.update(...)` would work with a `$set` command? https://facebook.github.io/react/docs/update.html

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+    query = super.buildSharedQuery(query);
+
+    // Construct an 'AND' query for all filters applied on this nested path
+    let groupedNestedFilter = this.createGroupedNestedFilter(query);
+    if (groupedNestedFilter) {
+      let nestedPath = this.getNestedPath();
+      // Find the element that Searchkit added to query.index.filters for this nested document and remove it
+      _.remove(query.index.filters, (filter) => (
+        _.find(filter, ['nested', 'path']) === nestedPath
+      ));
```
noisecapella: This also mutates the `ImmutableQuery`, can you find an alternative?

static/js/components/search/NestedAggregatingMenuFilter.js
noisecapella: Actually I'm not sure if modifiying the result from `getItems` is a good idea. Can we do something like this instead? `items.map(item => ({ ...item, doc_count: item[REVERSED_NESTED_AGG_KEY] ? item[REVERSED_NESTED_AGG_KEY].doc_count : item.doc_count }))`

gsidebo:
I'm fine with that

gsidebo:
It won't take care of those values in the way we need it to. I'll explain
when I get into the office

gsidebo:
The need for setting filters in this non-standard way is based on the need
to treat the nested filters as an AND rather than an OR. This is another
thing I can explain when I get to the office. Since we set those filter
values in a non-standard way to begin with, we also can't count on the
filters to be correctly applied to the aggregation by default

On Apr 5, 2017 5:57 PM, "George Schneeloch" <notifications@github.com>
wrote:

*@noisecapella* commented on this pull request.
------------------------------

In static/js/components/search/NestedAggregatingMenuFilter.js
<https://github.com/mitodl/micromasters/pull/2997#discussion_r110037809>:

> +          R.not,
+          R.equals(this.key),
+          R.head,
+          R.keys,
+          R.prop('term')
+        )
+      )(allTermFilters);
+    }
+    return otherTermFilters.length > 0 ? BoolMust(otherTermFilters) :
undefined;
+  }
+
+  /**
+   * Gets the appropriate terms bucket for this element's agg query.
+   */
+  getTermsBucket(query) {
+    let otherAppliedFiltersOnPath =
this.createFilterForOtherElementsOnPath(query);

Why are we adding the filter here and in createAggFilter? Wouldn't the
filter be added to the aggregations if it exists in filters and filtersMap?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mitodl/micromasters/pull/2997#pullrequestreview-31180742>,
or mute the thread

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+    return otherTermFilters.length > 0 ? BoolMust(otherTermFilters) : undefined;
+  }
+
+  /**
+   * Gets the appropriate terms bucket for this element's agg query.
+   */
+  getTermsBucket(query) {
+    let otherAppliedFiltersOnPath = this.createFilterForOtherElementsOnPath(query);
+    let termsKey = otherAppliedFiltersOnPath ? INNER_TERMS_AGG_KEY : this.key;
+    let termsBucket = ReverseNestedTermsBucket(termsKey, this.key, _.omitBy({
```
noisecapella: I don't think the `_.omitBy` is necessary here. If a value is undefined the JSON serializer should exclude the key and value from the output

static/js/components/search/NestedAggregatingMenuFilter.js
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * Creates an 'AND' filter for all filters that are applied on this element's nested path, minus the filter for
+   * this specific element (if it exists). If no other filters are applied on this element's nested path, undefined is
+   * returned.
+   *
+   * Example return value: {
+   *   'bool': {
+   *     'must': [
+   *       {'term': {'program.enrollments.course_title': 'Some Course Title'}
```
noisecapella: Can you add another `}` to the end here?

static/js/components/search/NestedAggregatingMenuFilter.js
gsidebo: this was lifted directly from searchkit source ([link](https://github.com/searchkit/searchkit/blob/master/src/core/accessors/FacetAccessor.ts#L186)). i can look into it to determine if it's necessary

gsidebo:
@singingwolfboy @noisecapella just tested that issue with the filter toggles hanging around even when there were no results. i couldn't reproduce it through normal means, but i do remember that i saw it when i switched from a different branch and an incompatible index to this branch. i'll look into it, but note that it isn't actually happening when you select a filter that results in a 0 count for any of the 3 course enrollment filters

gsidebo:
@noisecapella just pushed a few changes. let me know what you think

noisecapella:
I can confirm that the duplicate filters aren't there anymore

noisecapella:
Can you add a test to `LearnerSearchPage_test` which asserts that, if one of the course or semester facets was clicked, the appropriate filter was added in `post_filter`? You may need to modify `ELASTICSEARCH_RESPONSE` to have course infomration in the aggregate or provide your own data

dashboard/serializers.py
```diff
@@ -10,70 +10,66 @@ class UserProgramSearchSerializer:
...
-        serialized_enrollments = []
-        for course_title, is_verified in enrollment_status_map.items():
-            enrollment_status = 'Paid' if is_verified else 'Auditing'
-            serialized_enrollments.extend([
-                {'level': 1, 'value': course_title, 'ancestors': []},
-                {'level': 2, 'value': enrollment_status, 'ancestors': [course_title]}
-            ])
-        return serialized_enrollments
+            # If any course run for this course was verified/paid, count it as verified
+            if course_title not in serialized_enrollments_map or \
```
noisecapella: Can we use the `course_run.id` as the key here instead, since it's guaranteed to be unique and is always present?

noisecapella: Or rather, `course_run.course.id`

gsidebo: if anything we should use `course_run.course.id`. we only want a single enrollment item for multiple runs of the same course. i'll make this change

dashboard/serializers_test.py
```diff
@@ -442,48 +415,32 @@ def get_sorted_semester_values(serialized_program_user):
...
+        num_courses = 5
+        courses = CourseFactory.create_batch(num_courses, program=self.program)
+        course_runs = [CourseRunFactory.create(course=course) for course in courses]
+        for course_run in course_runs:
+            CachedEnrollmentVerifiedFactory.create(user=self.user, course_run=course_run)
+        with patch(
+            'dashboard.serializers.get_year_season_from_course_run', autospec=True, return_value=(2017, 'Spring')
+        ) as get_year_season_patch:
+            serialized_program_user = UserProgramSearchSerializer.serialize(self.program_enrollment)
+        assert len(serialized_program_user['enrollments']) == num_courses
```
noisecapella: Can you add an assertion for the semester value in each enrollment? As it is I can change the `semester` value in `serialized_enrollment` without breaking any tests, since we're only checking the length here and the next test is on `serialize_semester`

dashboard/serializers.py
gsidebo: actually, if we have multiple courses with the same title in the same program, and i change this code to use the course id, those will show up as separate options in the facet with the same title and no indication to the user about the difference between the two. i don't know if it's better for UX to do that or just to consider them as one course for the purposes of search. this is a very very tiny edge case either way...

noisecapella: I see... I'll leave it up to you then

noisecapella:
Looks great :+1:

annagav:
@gsidebo, @noisecapella, does this serialize only the first found paid course run for a course? I am afraid that might not reflect the highest final grade for the course that the learner could have gotten.

noisecapella:
I think you're right, can you create an issue to fix this @annagav?

PR #2999, opened 2017-04-04
Set thumbnails to null if main image is null

### LINKED ISSUE #2998: Thumbnail creation fails when the main profile image is empty
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/246176383/

noisecapella:
Fixes #2998 

#### What's this PR do?
Handles case where main profile image is empty. If it's empty and `update_image=True`, the thumbnails will be empty too

Set a user's profile image to be empty using the shell with `update_image=True`, and verify that the main thumbnails are empty too (`image_small` and `image_medium`).

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -184,12 +184,16 @@ class Profile(models.Model):
     def save(self, *args, update_image=False, **kwargs):
         """Set the student_id number to the PK number and update thumbnails if necessary"""
         if update_image:
-            small_thumbnail = make_thumbnail(self.image.file, IMAGE_SMALL_MAX_DIMENSION)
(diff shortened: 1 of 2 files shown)


giocalitri:
code is looks fine: testing functionality

giocalitri:
actually, do we need a migration?

giocalitri:
Functionality is fine: just the question about the migration, otherwise 👍

noisecapella:
We shouldn't need migrations because there shouldn't be any case where the thumbnails are currently different from the main image. If the user set the main image to something new the thumbnails will already be updated, and if they set it to something blank (they can only do this via REST API right now) then it will error like in the linked sentry issue

PR #3006, opened 2017-04-04
Search: Fix SelectedFilters titles

### LINKED ISSUE #2964: Search: Fix selected filter text on search page
ss 2017-03-27 at 11 50 34

Our selected filter component currently shows the ES field name in the text (eg: `courses:`, `program.semester_enrollments.semester:`). This information is irrelevant to our users, and it's jarring to look at. We should remove that text. 

This most likely happened as a result of upgrading searchkit.

annagav:
Fix #2964

#### What's this PR do?
Change the selected filter tag to display the filter name instead of the `field` prop.
Create SelectedFilter component to override the `SelectedFilters` component.
Go to learners search and select filters. Check the labels:
screen shot 2017-04-04 at 2 27 47 pm

### DIFF
--- a/static/js/components/search/ModifiedSelectedFilter.js
+++ b/static/js/components/search/ModifiedSelectedFilter.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import { SEARCH_FACET_FIELD_LABEL_MAP } from '../../constants';
+import { makeCountryNameTranslations } from '../LearnerSearch';
+
+export default class ModifiedSelectedFilter extends React.Component {
+  props: {
+    labelKey: string,
+    labelValue: string,
+    removeFilters: Function,
+    bemBlocks?: any,
+    filterId: string
+  };
+
+  countryNameTranslations: Object = makeCountryNameTranslations();
+
+  render() {
+    let {
+      labelKey,
+      labelValue,
+      removeFilter,
+      bemBlocks,
+      filterId,
+    } = this.props;
+
+    if (labelKey in SEARCH_FACET_FIELD_LABEL_MAP) {
+      labelKey = SEARCH_FACET_FIELD_LABEL_MAP[labelKey];
+    } else if (labelKey in this.countryNameTranslations) {
+      labelKey = this.countryNameTranslations[labelKey];
+    }
+    if (labelValue in this.countryNameTranslations) {
+      labelValue = this.countryNameTranslations[labelValue];
+    }
+    // This comes from searchkit documentation on "Overriding Selected Filter Component"
+    return (
+      <div className={bemBlocks.option().mix(bemBlocks.container("item")).mix(`selected-filter--${filterId}`)()}>
+        <div className={bemBlocks.option("name")}>{labelKey}: {labelValue}</div>
+        <div className={bemBlocks.option("remove-action")} onClick={removeFilter}>x</div>
+      </div>
+    );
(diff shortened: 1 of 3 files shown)


pdpinch:
This reminds me that it would be really nice if we could display the full name of the country, instead of the abbreviation (i.e. "Canada" instead of "CA"). Let me know if it should be a separate issue.

annagav:
Added new commit to display full names for countries and states:

screen shot 2017-04-04 at 5 23 59 pm

static/js/components/search/ModifiedSelectedFilter.js
```diff
@@ -0,0 +1,32 @@
...
+    } else if (filterName in this.countryNameTranslations) {
+      filterName = this.countryNameTranslations[filterName];
+    }
+    return <div className={bemBlocks.option()
+    .mix(bemBlocks.container("item"))
+    .mix(`selected-filter--${this.props.filterId}`)()}>
+      <div className={this.props.bemBlocks.option("name")}>{filterName}: {filterValue}</div>
+      <div className={this.props.bemBlocks.option("remove-action")} onClick={removeFilter}>x</div>
+    </div>;
+  }
```
gsidebo: I played around with getting rid of the need for custom rendering code, but due to the extensive usage of `this.props` in searchkit I couldn't see a clean way to do it. I did make this `render` method a little more concise in the process, though. What do you think of this?

```javascript
render() {
    let {
      labelKey, 
      labelValue, 
      removeFilter, 
      bemBlocks, 
      filterId
    } = this.props;

    if (labelKey in FILTER_NAMES) {
      labelKey = FILTER_NAMES[labelKey];
    } else if (labelKey in this.countryNameTranslations) {
      labelKey = this.countryNameTranslations[labelKey];
    }
    if (labelValue in this.countryNameTranslations) {
      labelValue = this.countryNameTranslations[labelValue];
    }

    return (
      <div className={
        bemBlocks.option()
        .mix(bemBlocks.container("item"))
        .mix(`selected-filter--${filterId}`)()
      }>
        <div className={bemBlocks.option("name")}>{labelKey}: {labelValue}</div>
        <div className={bemBlocks.option("remove-action")} onClick={removeFilter}>x</div>
      </div>
    );
  }
```

static/js/constants.js
```diff
@@ -24,6 +24,16 @@ export const EDUCATION_LEVELS = [
   {value: DOCTORATE, label: "Doctorate"}
 ];
 
+export const FILTER_NAMES = {
```
gsidebo: this name could be more helpful. `SEARCH_FILTER_ID_LABEL_MAP`, maybe?

annagav: How about this one: `SEARCH_FACET_FIELD_TITLE_MAP`?

gsidebo: maybe `SEARCH_FACET_FIELD_LABEL_MAP` since the value is being used more as a 'label' than a title. either one would work. i'll leave it to you

annagav: That's true. The reason for 'title' is that searchkit uses the title prop of the component for the title of the facet as well as the label for the selected filter. Also, I think it makes it clear that the title of the facet is used to label the selected filter.

gsidebo:
@annagav once that variable name is changed, this is good to go. i'm happy with my suggestion or yours.

PR #3010, opened 2017-04-05
Use reverse nested aggregation for education and fix related tests

### LINKED ISSUE #2963: Search: Degree facet counts users once for each degree
To reproduce:

- As a staff user, go to the search page. Pick a degree type and note the count.
- As a learner user, go to your profile and add a degree of that type.
- As a staff user, note that the count for that degree has gone up by 1 as expected.
- As a learner user, go to your profile and add another degree of that same type.
- As a staff user, note that the count for that degree has gone up by 1 again, which it shouldn't.

This crux of this issue is that the education history is being indexed as nested documents and so the aggregation is counting the number of nested documents (education objects) not the number of top-level objects we index (analogous to users).

### LINKED ISSUE #3009: Count doesn't show up for work history select facet
This is due to a change in the Elasticsearch aggregation key being used. I'm working on it along with #2963

noisecapella:
Fixes #2963 
Fixes #3009 

#### What's this PR do?
Uses a reverse nested aggregation to ensure that the education count is per person, not per education. It also fixes a problem with the searchkit id for the work history facet and fixes the tests to mock the elasticsearch responses properly.

For a user set up two `Employment` objects which are both current for the same company name (non-current employments are filtered out). Using the search box filter the name of the user so that they are the only user in the learner search page. Verify that the work history facet shows the employment for that user with a `(1)` in the name.

Do a similar thing for `Education` but with two `Education` objects for that user who have the same degree type. You should see `1` next to the degree type in the facet.

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -295,11 +295,7 @@ class Education(models.Model):
     school_country = models.TextField()
 
     def __str__(self):
-        degree_title = ''
-        for degree, title in self.DEGREE_CHOICES:
-            if self.degree_name == degree:
-                degree_title = title
-                break
+        degree_title = dict(self.DEGREE_CHOICES).get(self.degree_name, '')
 
         return 'Education for {user}, {degree} at {title} {date}'.format(
             user=self.profile.user.username,
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -271,7 +271,7 @@ class Profile(models.Model):
             name_components.append('({})'.format(self.preferred_name))
         return ' '.join(name_components)
 
-import json.tool
+
 class Education(models.Model):
     """
     A user education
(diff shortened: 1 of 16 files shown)


profiles/models.py
```diff
@@ -288,3 +293,17 @@ class Education(models.Model):
     school_city = models.TextField()
     school_state_or_territory = models.TextField()
     school_country = models.TextField()
+
+    def __str__(self):
+        degree_title = ''
```
rhysyngsun: Suggestion: `degree_title = dict(self.DEGREE_CHOICES).get(self.degree_name, '')` would simplify this.

rhysyngsun:
Functionality looks good.

profiles/models.py
rhysyngsun: Feel free to merge either way you decide to go with this since it's so trivial.

PR #3012, opened 2017-04-05
add automatic email index page

### LINKED ISSUE #3011: add automatic email admin page
this is doing part of #2950, basically implementing the basic page, ability to toggle whether a mail is active, and adding a link to the sidebar nav.

aliceriot:

closes #3011

#### What's this PR do?

This

- adds a new page (at `/automaticemails`) which allows staff users to edit automatic emails that they created
- adds a link to the sidebar nav to this page, which appears if you are staff
- added some redirect protection, so that if you do not have a staff role you will be redirected away from the automatic emails page (even though the API wouldn't let you see anything anyway)

I think that's it!

Confirm the following things:

- a link to the email page shows up in the sidebar if you are staff, and not if you are a learner
- the automatic emails page shows any automatic emails you've created.
- the automatic emails page should show a placeholder message if you haven't yet created any.
- the automatic emails page should show a spinner while the GET request is in-flight
- you should be able to toggle the 'active' state of an automatic email, and it should show a little spinner while that's in-flight (you can use chrome's internet connection throttling feature to make this more obvious)
- if you are not a staff user, when you try to load `/automaticemails` you should be redirected to `/dashboard`	

I think that's it...

### DIFF
--- a/mail/urls.py
+++ b/mail/urls.py
@@ -1,5 +1,6 @@
 """URLs for mail app"""
-from django.conf.urls import url
+from django.conf.urls import url, include
+from rest_framework import routers
 
 from mail.views import (
     LearnerMailView,
@@ -9,11 +10,14 @@ from mail.views import (
     AutomaticEmailView,
 )
 
+router = routers.DefaultRouter()
+router.register(r'automatic_email', AutomaticEmailView, base_name='automatic_email_api')
+
 urlpatterns = [
     url(r'^api/v0/financial_aid_mail/(?P<financial_aid_id>[\d]+)/$', FinancialAidMailView.as_view(),
         name='financial_aid_mail_api'),
     url(r'^api/v0/mail/search/$', SearchResultMailView.as_view(), name='search_result_mail_api'),
     url(r'^api/v0/mail/course/(?P<course_id>[\d]+)/$', CourseTeamMailView.as_view(), name='course_team_mail_api'),
     url(r'^api/v0/mail/learner/(?P<student_id>[\d]+)/$', LearnerMailView.as_view(), name='learner_mail_api'),
-    url(r'^api/v0/mail/automatic_email/$', AutomaticEmailView.as_view(), name='automatic_email_api'),
+    url(r'^api/v0/mail/', include(router.urls)),
 ]
--- a/mail/views.py
+++ b/mail/views.py
@@ -9,7 +9,9 @@ from rest_framework import (
     status,
 )
 from rest_framework.generics import GenericAPIView, ListAPIView
+from rest_framework.mixins import UpdateModelMixin
 from rest_framework.views import APIView
+from rest_framework.viewsets import GenericViewSet
 from rest_framework.response import Response
 
 from courses.models import Course
@@ -74,7 +76,7 @@ class LearnerMailView(GenericAPIView):
         )
 
 
-class AutomaticEmailView(ListAPIView):
+class AutomaticEmailView(ListAPIView, UpdateModelMixin, GenericViewSet):
     """
     View class that deals with listing and editing automatic mails
     """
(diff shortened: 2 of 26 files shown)


mail/urls.py
```diff
@@ -16,4 +16,9 @@
     url(r'^api/v0/mail/course/(?P<course_id>[\d]+)/$', CourseTeamMailView.as_view(), name='course_team_mail_api'),
     url(r'^api/v0/mail/learner/(?P<student_id>[\d]+)/$', LearnerMailView.as_view(), name='learner_mail_api'),
     url(r'^api/v0/mail/automatic_email/$', AutomaticEmailView.as_view(), name='automatic_email_api'),
+    url(
```
rhysyngsun: Having two urls named the same could cause some issues if they're used with `reverse()` at some point. I also think you can update this so that you only need one definition. I tested these changes locally so it should be a simple copy/paste:

```python
from django.conf.urls import url, include  # added include
from rest_framework import routers

router = routers.DefaultRouter()
router.register(r'automatic_email', AutomaticEmailView, base_name='automatic_email_api')

urlpatterns = [
    ...
    url(r'^api/v0/mail/', include(router.urls)),  # replaces lines 18-22
]
```

Then add this to `AutomaticEmailView`:
```python
from rest_framework.viewsets import GenericViewSet  # add import

class ProfileViewSet(RetrieveModelMixin, UpdateModelMixin, GenericViewSet):  # add GenericViewSet
    ...
    lookup_value_regex = '[-\w.]+'  # pylint: disable=anomalous-backslash-in-string

```

rhysyngsun:
Functional testing looked fine aside from the one issue I mentioned above.

static/js/reducers/automatic_emails.js
```diff
@@ -0,0 +1,31 @@
...
+  }),
+  patchSuccessHandler: (payload, oldData) => R.update(
+    R.findIndex(R.propEq('id', payload.id), oldData), payload, oldData
+  ),
+  initialState: { ...INITIAL_STATE, emailsInFlight: new Set()},
+  extraActions: {
+    [TOGGLE_EMAIL_PATCH_IN_FLIGHT]: (state: Object, action: Action<number, void>) => {
+      const emails = new Set(state.emailsInFlight);
+      const { payload: id } = action;
+      emails.has(id) ? emails.clear(id) : emails.add(id);
```
aliceriot: ohhh dear, I should really try these sorts of things with more than one entry in the set! haha

mail/views.py
```diff
@@ -84,11 +85,17 @@ class AutomaticEmailView(ListAPIView):
...
     permission_classes = (permissions.IsAuthenticated, UserCanMessageLearnersPermission, )
     serializer_class = AutomaticEmailSerializer
+    lookup_field = "id"
+    lookup_url_kwarg = "email_id"
 
     def get_queryset(self):
         """Get the queryset which should be serialized"""
         return AutomaticEmail.objects.filter(staff_user=self.request.user)
 
+    def patch(self, request, *args, **kwargs):
```
aliceriot: I believe without it I got an error but I don't remember exactly, I can remove it and see what happens

aliceriot: yea if I comment out that method I get a `405`

mail/urls.py
aliceriot: hrm so taking that approach seems to break the usage of `reverse` that was working before. In one of the views tests I'm just doing:

```py
cls.url = reverse('automatic_email_api')
```

that's now throwing a `django.urls.exceptions.NoReverseMatch`

aliceriot: btw the approach I was using before seemed to work alright with `reverse`, I could pass `kwargs={'email_id': id}` to get a `PATCH` url or just leave it off otherwise.

mail/views.py
aliceriot: oh yes you're right, it works with the URL routing change (I think due to the inclusion of `GenericViewSet`)

mail/urls.py
rhysyngsun: @aliceriot DRF appends suffixes to the urls, depending on which one you need: http://www.django-rest-framework.org/api-guide/routers/#defaultrouter

aliceriot: oh cool, that looks like what I need!

aliceriot:
comments addressed!

PR #3016, opened 2017-04-06
Modified the mmtrack has paid to better handle FA programs

### LINKED ISSUE #2995: Dashboard: course_paid_on_edx should not be equivalent to having paid for a FA course
From [ZenDesk 6882](https://odl.zendesk.com/agent/tickets/6882)

### Steps to Reproduce

1. Create a FA course. 
2. Create a user with a fake enrollment in the FA course
3. Fake a verified certificate for the user in the FA course (optional)
3. Create a final grade for the user with `course_paid_on_edx=True`
(Steps 2-4 are equivalent to running `manage.py alter_date set_to_passed`)

### Expected Behavior:

When viewing the learner's dashboard (or when viewing their profile as staff), the course should not display as paid. The learner still needs to upgrade to the FA course price. 

(optional: When the learner does upgrade, they should get an automatic $50 coupon, since they have a certificate)

### Actual behavior:

When viewing the learner's dashboard, the course displays as paid. The learner can't upgrade.

giocalitri:
fixes #2995 

#### What's this PR do?
Changes the behavior of `mmtrack.has_paid` to handle the FA courses in the right way

Create a final grade with `course_run_paid_on_ed=True` for a course run in a FA program and verify that it is still not paid in the dashboard.
Adding an order and a line for the same course will turn the course to paid.

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -116,32 +116,6 @@ class MMTrack:
         """
         return self.is_enrolled(edx_course_key) and self.has_paid(edx_course_key)
 
-    def has_paid(self, edx_course_key):
-        """
-        Returns true if user paid for a course run.
-
-        Args:
-            edx_course_key (str): an edX course run key
-
-        Returns:
-            bool: whether the user is paid
-        """
-        if self.has_paid_final_grade(edx_course_key):
-            return True
-
-        # financial aid programs need to have an audit enrollment and a paid entry for the course
-        if self.financial_aid_available:
-            return edx_course_key in self.paid_course_keys
-
-        # normal programs need to have a verified enrollment
-        return self.has_verified_enrollment(edx_course_key)
-
-    def has_paid_for_any_in_program(self):
-        """
-        Returns true if a user has paid for any course run in the program
-        """
-        return any(self.has_paid(edx_course_key) for edx_course_key in self.edx_course_keys)
-
     def has_verified_enrollment(self, edx_course_key):
         """
         Returns true if user has a verified enrollment
@@ -210,6 +184,42 @@ class MMTrack:
         """
         return self.final_grade_qset.for_course_run_key(edx_course_key).exists()
 
+    def has_paid(self, edx_course_key):
+        """
+        Returns true if user paid for a course run.
+
+        Args:
+            edx_course_key (str): an edX course run key
+
+        Returns:
+            bool: whether the user is paid
+        """
+        # financial aid programs need to have a paid entry for the course
+        if self.financial_aid_available:
+            return edx_course_key in self.paid_course_keys
+
+        # normal programs need to have paid_on_edx in the final grades or a verified enrollment
+        if self.has_final_grade(edx_course_key):
+            return self.has_final_grade_paid_on_edx(edx_course_key)
+        return self.has_verified_enrollment(edx_course_key)
+
+    def has_paid_for_any_in_program(self):
+        """
+        Returns true if a user has paid for any course run in the program
+        """
+        return any(self.has_paid(edx_course_key) for edx_course_key in self.edx_course_keys)
+
+    def has_final_grade_paid_on_edx(self, edx_course_key):
+        """
+        Checks if there is a a frozen final grade and the user paid for it.
+
+        Args:
+            edx_course_key (str): an edX course run key
+        Returns:
+            bool: whether or not a user has a final grade and has paid
+        """
+        return self.final_grade_qset.paid_on_edx().for_course_run_key(edx_course_key).exists()
+
     def has_paid_final_grade(self, edx_course_key):
         """
         Checks if there is a a frozen final grade and the user paid for it.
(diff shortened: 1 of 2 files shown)


dashboard/utils.py
```diff
@@ -210,6 +184,42 @@ def has_final_grade(self, edx_course_key):
...
+            edx_course_key (str): an edX course run key
+
+        Returns:
+            bool: whether the user is paid
+        """
+        # financial aid programs need to have a paid entry for the course
+        if self.financial_aid_available:
+            return edx_course_key in self.paid_course_keys
+
+        # normal programs need to have paind_on_edx in the final grades or a verified enrollment
```
pdpinch: typo `paid_on_edx`

dashboard/utils.py
```diff
@@ -219,7 +229,7 @@ def has_paid_final_grade(self, edx_course_key):
         Returns:
             bool: whether a frozen final grade exists
         """
-        return self.final_grade_qset.paid_on_edx().for_course_run_key(edx_course_key).exists()
+        return self.has_final_grade(edx_course_key) and self.has_paid(edx_course_key)
```
pdpinch: Why was this change necessary?

gsidebo: I'm guessing it's because a user can still be considered to have a paid final grade if (a) it's a FA program, (b) they have a FinalGrade that wasn't paid on edX, and (c) they _have_ paid via ecommerce. @giocalitri is that right?

giocalitri: @gsidebo what you say is right, but it is more complicated than that:
a FA user must considered as "not paid" even if the "paid on edX" is True.

giocalitri: and the "paid on edX" in this case will be used to assign discount coupons only

pdpinch: Got it. Thanks.

pdpinch:
Functionality looks good to me. 👍

dashboard/utils_test.py
```diff
@@ -478,11 +479,31 @@ def test_has_paid_fa_with_final_grade(self):
         key = self.crun_fa.edx_course_key
         assert mmtrack.has_paid(key) is False
         final_grade = FinalGradeFactory.create(user=self.user, course_run=self.crun_fa, course_run_paid_on_edx=True)
-        assert mmtrack.has_paid(key) is True
+        assert mmtrack.has_paid(key) is False
         final_grade.course_run_paid_on_edx = False
         final_grade.save()
         assert mmtrack.has_paid(key) is False
 
+    def test_has_paid_fa_with_final_grade_course_run_paid_on_mm(self):
```
gsidebo: these test case names are a little long. can you change these to `test_not_paid_fa_with_course_run_paid_on_edx` and `test_has_paid_fa_with_course_run_paid_on_mm`

dashboard/utils.py
```diff
@@ -210,6 +184,42 @@ def has_final_grade(self, edx_course_key):
...
+            return self.has_paid_on_edx_final_grade(edx_course_key)
+        return self.has_verified_enrollment(edx_course_key)
+
+    def has_paid_for_any_in_program(self):
+        """
+        Returns true if a user has paid for any course run in the program
+        """
+        return any(self.has_paid(edx_course_key) for edx_course_key in self.edx_course_keys)
+
+    def has_paid_on_edx_final_grade(self, edx_course_key):
```
gsidebo: i think `has_final_grade_paid_on_edx` would be a better name

dashboard/utils.py
```diff
@@ -210,6 +184,42 @@ def has_final_grade(self, edx_course_key):
...
+        Args:
+            edx_course_key (str): an edX course run key
+        Returns:
+            bool: whether a frozen final grade with the flag paid on edx exists
+        """
+        return self.final_grade_qset.paid_on_edx().for_course_run_key(edx_course_key).exists()
+
     def has_paid_final_grade(self, edx_course_key):
         """
         Checks if there is a a frozen final grade and the user paid for it.
```
gsidebo: not related to this PR, but might as well fix it here: Checks if there is a ~~a~~ frozen final grade and the user paid for it.

dashboard/utils.py
```diff
@@ -219,7 +229,7 @@ def has_paid_final_grade(self, edx_course_key):
         Returns:
             bool: whether a frozen final grade exists
```
gsidebo: not related to this PR. this should be something like: "whether or not a user has a final grade and has paid"

gsidebo:
general comment: in `MMTrack`, we should consider loading the complete set of final grade and overall payment information in `__init__` or as a result of some method call. we already do something like this with `paid_course_keys`. as it is, we execute a bunch of queries we could avoid, and this PR adds some calls to `has_final_grade` that can be avoided. something for the future, perhaps

giocalitri:
>general comment: in MMTrack, we should consider loading the complete set of final grade and overall payment information in __init__ or as a result of some method call. we already do something like this with paid_course_keys. as it is, we execute a bunch of queries we could avoid, and this PR adds some calls to has_final_grade that can be avoided. something for the future, perhaps

I agree, but I also think we should split the MMTrac in multiple classes and somehow compose the aggregate one either via subclassing or using class properties that point at the other smaller class instances

giocalitri:
done

Issue #2241, opened 2017-01-04
Welcome e-mail for new users


pdpinch:
Some programs would like to send out an automatic welcome e-mail to new users, when they enroll. The e-mail would contain information about the program, and a link to a (qualtrics) survey. 

It would be useful to have a mechanism to send out this e-mail to existing users enrolled in the program who haven't been sent one yet. 

### acceptance criteria (MVP)
- [ ] course team gives us a mail template which we can add (via django admin or shell)
   - the template will include a link to a qualtrics survey. The link should include some user-identifiable information, like their edx username 
- [ ] support one query: when learner enrolls in program 

### related issues (out of scope)
- [ ] The course team will need an interface to be able to edit and customize the e-mail. 
- [ ] The course team should have options about what triggers the e-mail. Options include:
   - When the learner pays for (any) course in the program
   - When the learner pays for a specific course

pdpinch:
This is a special case of #2510 and should be implemented in a similar way, using the Percolate feature of ElasticSearch.

pdpinch:
This needs to be re-scoped.

pdpinch:
I emailed Arthur from SCM asking for:

1) Search criteria for welcome mail
2) text for welcome mail

noisecapella:
Do we have either of those yet?

pdpinch:
**Search Criteria:** Anyone who joins the portal as a paid, verified learner in at least one course.

**email template:** 

```
Greetings  {first name},
 
Thank you for verifying for your course in the MITx MicroMasters of Supply Chain Management. Welcome! 
 
We will use micromasters.mit.edu as our primary communication channel with you, though all of your course work will take place at edX.org. This dashboard will help you navigate between courses. It is linked directly to edX and will update automatically with your course progress. It will help us to contact you with vital information about courses and allow you to reach us directly with non-assignment questions or concerns. It will enable us to connect you with other learners in your geographic area or reach you with special opportunities. 
 
Please take about three minutes to complete the survey here.
http://mit.co1.qualtrics.com/jfe/form/SV_9zqju8s4t4VniQd
 
As your course work progresses we hope to learn more about you. Knowing the answers to these questions will help us provide you with the best learning experience possible. On behalf of the entire MicroMasters team we welcome you.
 
Best wishes,
 
Dr. Chris Caplice, Director, SCM MicroMasters Program, Silver Family Research Fellow
Dr. Eva Ponce, Executive Director, SCM MicroMasters Program 
```

[...]

PR #3019, opened 2017-04-06
Add mail_id and template variables to Mailgun functions

### LINKED ISSUE #3035: Tech Debt: local bug in mail/api.py send_batch
I noticed this while reviewing nearby code. It is contingent on `MAILGUN_RECIPIENT_OVERRIDE` being set so this shouldn't affect production.

- Here we're consuming an iterable: `original_recipients_chunk`https://github.com/mitodl/micromasters/blob/617f80c7ccff8c8e6ad57a677c5c4fafec11b513/mail/api.py#L158
- In the case of `MAILGUN_RECIPIENT_OVERRIDE` being set, that value ends up being the original `recipients` generator passed into `send_batch` (see L140 and L114 in the same file).
- This causes the generator to be exhausted when we create the exception on L158 (and could, if the generator consumed enough memory, cause an OOM-kill) and then I would expect that `chunks` then stops iterating since we've mutated the same reference it is mutating.
- Additionally, the further downside is the logged exception wouldn't actually contain the set of emails from the chunk, but rather the remainder of emails from the generator that we hadn't sent or attempted to send yet.

TLDR: the exception logging should probably just unconditionally use `chunk`.

noisecapella:
Part of #2241 
Fixes #3035 

#### What's this PR do?
 - Adds a UUID field to Profile
 - Uses that UUID and `first_name` in the template variables for automatic emails and bulk mail sent out via search.

Send an email with the email campaign option set and `MAILGUN_RECIPIENT_OVERRIDE` set to your email. Verify that you see a UUID and a first name in the body of the email.

Next, change or create a new Profile which matches the criteria for the automatic email. Verify that a new email was sent just for that user, also containing their template variables.


pdpinch:
I think we want to use preferred_name instead in e-mails. @roberthouse54 is that right?

noisecapella:
Sure, are there any variables we need?

pdpinch:
For this case, preferred name and UUID should cover it.

mail/api.py
```diff
@@ -337,3 +332,28 @@ def mark_emails_as_sent(automatic_email, emails):
...
+    Returns:
+        generator of dict:
+            A dictionary of template variables along with email so we can tell who is who
+    """
+    queryset = User.objects.filter(email__in=emails).values(
+        'email',
+        'profile__mail_id',
+        'profile__preferred_name',
+    )
+    return (
```
rhysyngsun: Since you're returning a generator here, I would probably suggest using `iterator()` so that the entire result isn't loaded into memory at start of generator iteration. If only `OrderedDict` generators were a thing, that would be even nicer :smile:

mail/api.py
```diff
@@ -127,19 +112,27 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
...
-                subject=subject,
-                text=body
-            )
-            params['recipient-variables'] = json.dumps({email: {} for email in chunk})
+        for chunk in chunks(recipients, chunk_size=chunk_size):
+            params = {
+                'to': [email for email, _vars in chunk],
+                'subject': subject,
+                'text': body,
+                'recipient-variables': json.dumps({email: _vars or {} for email, _vars in chunk}),
```
rhysyngsun: Suggestion: refactor this into a variable and then just pass `keys()` to `to`.

rhysyngsun:
Moving on to manual testing.

rhysyngsun:
Functionality looks good aside form the issues we filed separate issues for.

mail/api.py
```diff
@@ -111,7 +93,10 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
         Args:
             subject (str): Email subject
             body (str): Text email body
-            recipients (iterable): A list of recipient emails
+            recipients (iterable of (recipient, recipient_vars)):
```

mail/api.py
```diff
@@ -127,19 +112,27 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
                along with recipients we failed to send to.
         """
         original_recipients = recipients
-        body, recipients = cls._recipient_override(body, original_recipients)
+
+        if settings.MAILGUN_RECIPIENT_OVERRIDE is not None:
```
noisecapella: I'm not sure it's necessary, and maybe there's the possibility that we want to set this environment variable on a PR build where `DEBUG` is false

mail/api.py
```diff
@@ -127,19 +112,27 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
...
-        while len(chunk) > 0:
-            params = dict(
-                to=chunk,
-                subject=subject,
-                text=body
-            )
-            params['recipient-variables'] = json.dumps({email: {} for email in chunk})
+        for chunk in chunks(recipients, chunk_size=chunk_size):
+            params = {
+                'to': [email for email, _vars in chunk],
```

mail/api.py

noisecapella:
Changes addressed, I also fixed #3035 as part of this

By the way you may want to send an email without `MAILGUN_RECIPIENT_OVERRIDE` set with certain template variables set just as an integration test. For example `%recipient.preferred_name%` should be replaced with the preferred name of the recipient. We can't do this with `MAILGUN_RECIPIENT_OVERRIDE` since that variable doesn't specify what template variables go with that email.

rhysyngsun:
:+1: looks good

PR #3022, opened 2017-04-07
fix Promise error handling

aliceriot:

none

#### What's this PR do?

I was looking at a JavaScript error in Sentry this morning and I realized (after reading [this issue](https://github.com/facebook/react/issues/7617), among others) that our error handling for `Promises` is done incorrectly, in a way that can swallow errors and lead to some strange behavior.

In short, these two things are not equivalent:

```js
// pattern we have in our code a fair amount right now:
promise.then(success => {
  doSomethingWithSuccess(success);
  return Promise.resolve(success); // maybe, we don't always do this
}).catch(error => {
  ohNoAnError(error); // :(
  return Promise.reject();
};

// correct pattern
promise.then(success => {
  doSomethingWithSuccess(success);
  return Promise.resolve(success); // maybe, we don't always do this
}, error => {
  ohNoAnError(error); // :(
  return Promise.reject();
});
```

basically, the first problem is an issue because `.catch` becomes overly broad. Instead of just handling the `rejected` case for `promise`, it also catches any errors that happen in the `.then` handler. This can, in certain cases, include rendering / UI errors which are not related to the `promise` itself, but related to an error in the code we use to handle or render the data once it comes back. In this case, the UI error will be silently swallowed by the `.catch` (because it's occurring in the `.then` handler) and we won't see it. Then, often, the error will show up somewhat cryptically with a different stack trace.

this is a problem in particular with redux, because our async (redux-thunk) action dispatchers trigger a `render()` call when `dispatch` fires, so that if there is an error anywhere in that `render()` call (say, because of wacky data) it will trace back to the `dispatch()` and be silently swallowed by `.catch`. yucky.

The correct way to use promises is `.then(onResolve, onReject)`. Then our handler functions are strictly confined to dealing with resolution or rejection of the original promise itself. Of course, if want to chain off of this function that is fine and dandy as well, but we need to make sure that there is an `onReject` handler for the initial promise beforehand, otherwise the error can be propagated somewhat strangely.

Anyway, I believe this is the root cause of a React error we're seeing rarely but consistently (`Cannot read property 'getHostNode' of null`).

two other React issues relating to this:

https://github.com/facebook/react/issues/8267

https://github.com/facebook/react/issues/4199

tests should pass, app should behave normally, there should be no regressions.

### DIFF
--- a/static/js/actions/programs.js
+++ b/static/js/actions/programs.js
@@ -11,10 +11,7 @@ import { fetchDashboard } from './dashboard';
 import { fetchCoursePrices } from './course_prices';
 import { setToastMessage, setEnrollProgramDialogVisibility } from '../actions/ui';
 import type { Dispatcher } from '../flow/reduxTypes';
-import type {
-  AvailableProgram,
-  AvailablePrograms,
-} from '../flow/enrollmentTypes';
+import type { AvailableProgram } from '../flow/enrollmentTypes';
 import * as api from '../lib/api';
 
 export const SET_CURRENT_PROGRAM_ENROLLMENT = 'SET_CURRENT_PROGRAM_ENROLLMENT';
@@ -29,12 +26,13 @@ export const receiveGetProgramEnrollmentsSuccess = createAction(RECEIVE_GET_PROG
 export const RECEIVE_GET_PROGRAM_ENROLLMENTS_FAILURE = 'RECEIVE_GET_PROGRAM_ENROLLMENTS_FAILURE';
 export const receiveGetProgramEnrollmentsFailure = createAction(RECEIVE_GET_PROGRAM_ENROLLMENTS_FAILURE);
 
-export function fetchProgramEnrollments(): Dispatcher<AvailablePrograms> {
+export function fetchProgramEnrollments(): Dispatcher<void> {
   return (dispatch: Dispatch) => {
     dispatch(requestGetProgramEnrollments());
     return api.getPrograms().
-      then(enrollments => dispatch(receiveGetProgramEnrollmentsSuccess(enrollments))).
-      catch(error => {
+      then(enrollments => {
+        dispatch(receiveGetProgramEnrollmentsSuccess(enrollments));
+      }, error => {
         dispatch(receiveGetProgramEnrollmentsFailure(error));
         // the exception is assumed handled and will not be propagated
       });
@@ -50,7 +48,7 @@ export const receiveAddProgramEnrollmentSuccess = createAction(RECEIVE_ADD_PROGR
 export const RECEIVE_ADD_PROGRAM_ENROLLMENT_FAILURE = 'RECEIVE_ADD_PROGRAM_ENROLLMENT_FAILURE';
 export const receiveAddProgramEnrollmentFailure = createAction(RECEIVE_ADD_PROGRAM_ENROLLMENT_FAILURE);
 
-export const addProgramEnrollment = (programId: number): Dispatcher<AvailableProgram> => {
+export const addProgramEnrollment = (programId: number): Dispatcher<?AvailableProgram> => {
   return (dispatch: Dispatch) => {
     dispatch(requestAddProgramEnrollment(programId));
     return api.addProgramEnrollment(programId).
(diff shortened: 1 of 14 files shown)


noisecapella:
If you have a promise chain like this:

    // correct pattern
    fetchSomething().then(() => transformIt()).then(success => {
      doSomethingWithSuccess(success);
      return Promise.resolve(success); // maybe, we don't always do this
    }, error => {
      ohNoAnError(error); // :(
      return Promise.reject();
    });

We would be handling errors from either `transformIt` or `fetchSomething`, right? And if we added the `catch` instead of using `onReject` then we would also handle errors happening from `doSomethingWithSuccess`? Just making sure

aliceriot:
Yes, exactly. And, in particular, in our action dispatchers, which look like this:

```js
export function fetchProgramEnrollments(): Dispatcher<AvailablePrograms> {
  return (dispatch: Dispatch) => {
    dispatch(requestGetProgramEnrollments());
    return api.getPrograms().
      then(enrollments => dispatch(receiveGetProgramEnrollmentsSuccess(enrollments))).
      catch(error => {
        dispatch(receiveGetProgramEnrollmentsFailure(error));
        // the exception is assumed handled and will not be propagated
      });
  };
}
```

the `.catch` handler will swallow any errors which happen as a result of the `dispatch` call. This might seem ok but it is not - `dispatch` is from `react-redux`, and a call to `dispatch` will trigger a re-render of the UI. Any UI errors that happen as a result of that re-render will therefore be in the same call stack, and so will happen in the `.then` handler, and will then be silently swallowed by the `.catch` handler. Not what we want!

[This comment](https://github.com/facebook/react/issues/7617#issuecomment-247710003) probably explains the problem better than I will be able to.

noisecapella:
In api.js:125 there's a `catch` we should probably move into the preceeding `then`

static/js/actions/programs.js
```diff
@@ -50,7 +48,7 @@ export const receiveAddProgramEnrollmentSuccess = createAction(RECEIVE_ADD_PROGR
 export const RECEIVE_ADD_PROGRAM_ENROLLMENT_FAILURE = 'RECEIVE_ADD_PROGRAM_ENROLLMENT_FAILURE';
 export const receiveAddProgramEnrollmentFailure = createAction(RECEIVE_ADD_PROGRAM_ENROLLMENT_FAILURE);
 
-export const addProgramEnrollment = (programId: number): Dispatcher<AvailableProgram> => {
+export const addProgramEnrollment = (programId: number): Dispatcher<?AvailableProgram> => {
```
noisecapella: Why do we need to mark this with `?`?

aliceriot: because the `onResolved` handler doesn't actually return a `Promise<AvailableProgram>`, it just returns `undefined`.

aliceriot: before I think Flow wasn't able to accurately type-check things.

aliceriot:
ok! I re-wrote `fetchJSONWithCSRF` to have some cleaner handling of the different cases we need to deal with, and to just have one `.then` call off of `response.text()`.

static/js/lib/api.js
```diff
@@ -92,12 +93,17 @@ import { fetchWithCSRF } from './api';
  *  - non 2xx status codes will reject the promise returned
```
noisecapella: Can you move this comment down to `_fetchJSONWithCSRF`?

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/249296145/

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/228916007/

static/js/lib/api.js
noisecapella: And also add a comment explaining about the conversion from maybe to promise in case someone isn't familiar with sanctuary

static/js/lib/api.js
```diff
@@ -84,6 +85,15 @@ const _fetchWithCSRF = (path: string, init: Object = {}): Promise<*> => {
 export { _fetchWithCSRF as fetchWithCSRF };
 import { fetchWithCSRF } from './api';
 
+const resolveEither = S.either(
```
noisecapella: Can you add a comment here in case someone is unfamiliar with sanctuary?

PR #3032, opened 2017-04-09
Pdpinch/change order courserun

### LINKED ISSUE #2654: Allow an admin to shift a purchase from one course run to another
We're having a growing number of users make purchases for one course run and then request that the purchase be transferred to another run. 

I've been handling these ad hoc via the django shell, but the number is getting to be great enough that we should make this easier. 

### options:

1. Allow lines to be changed in the django admin. This would be easier for me than using the shell and less prone to error. 
2. do 1) above and give limited admin access to Vinnie (customer support)
3. Add some web-based ui to the financial aid review form 

### nice to have

- It would be nice to generate an automatic e-mail to the learner documenting his updated course purchases (the initial purchase generates an e-mail from CyberSource, but this update wouldn't touch CyberSource)

### other considerations

- since the changes the timing of an order fulfillment, it has an impact on the financial team (different course runs have their royalties distributed differently; the fiscal year a purchase is recognized is significant)

pdpinch:

fixes #2654

#### What's this PR do?

Makes `line.course_key` editable in the admin. Any changes create a corresponding `OrderAudit`

Open up the admin at `/admin/ecommerce/line/` and try changing a course_key. Check to see that an OrderAudit is created in `/admin/ecommerce/orderaudit/`

Note that there isn't any validation applied to the course key. It's up to the admin to make sure that a valid course key is used. 

Learners asking to defer their payments to another semester is our 2nd most common support issue (after financial aid resets).

It's the only issue that currently requires an admin (i.e me) to change data via the django shell. I'd feel much more comfortable doing this via the admin. 

image001

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -25,13 +25,20 @@ class LineAdmin(admin.ModelAdmin):
     """Admin for Line"""
     model = Line
 
-    readonly_fields = [name for name in get_field_names(Line) if name != 'edx_course_key']
+    readonly_fields = [name for name in get_field_names(Line) if name != 'course_key']
 
     def has_add_permission(self, request):
         return False
 
     def has_delete_permission(self, request, obj=None):
         return False
+
+    def save_model(self, request, obj, form, change):
+        """
+        Saves object and logs change to object
+        """
+        super(LineAdmin, self).save_model(request, obj, form, change)
+        obj.order.save_and_log(request.user)
 
 
 class OrderAdmin(admin.ModelAdmin):
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -25,7 +25,7 @@ class LineAdmin(admin.ModelAdmin):
     """Admin for Line"""
     model = Line
 
-    readonly_fields = get_field_names(Line)
+    readonly_fields = [name for name in get_field_names(Line) if name != 'edx_course_key']
 
     def has_add_permission(self, request):
         return False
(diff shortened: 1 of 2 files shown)


ecommerce/admin.py
```diff
@@ -25,14 +25,21 @@ class LineAdmin(admin.ModelAdmin):
     """Admin for Line"""
     model = Line
 
-    readonly_fields = get_field_names(Line)
+    readonly_fields = [name for name in get_field_names(Line) if name != 'course_key']
```
amir-qayyum-khan: so all fields other then `course_key` are read only?

pdpinch: That's correct. We use the same pattern elsewhere in this admin where only one field is editable.

ecommerce/admin_test.py
```diff
@@ -31,3 +31,14 @@ def test_save_and_log_model(self):
             change=Mock()
         )
         assert OrderAudit.objects.count() == 1
+
+        line = LineFactory.create()
```
amir-qayyum-khan: @pdpinch i think here we should separate test, that should create a Line with a dummy course key, then when should assert `OrderAudit` change. Afterword we should update course key in same Line and see the behaviour in  `OrderAudit`

noisecapella:
Hmm, it looks like PostGIS isn't installed in our databases by default. Can you switch back to using the `DecimalField` like before? We can do the migration to use PostGIS at some later date

amir-qayyum-khan:
@noisecapella  wrong PR :)

PR #3041, opened 2017-04-10
Exams: process exam results

### LINKED ISSUE #2791: exams: process exam results files from Pearson
Exam results from Pearson arrive in files from sftp (first expected on 3/21/17). These files need to be parsed and the data saved, to the model created in #2790 

### acceptance criteria
- [x] save the file to long-term location, like an encrypted S3 bucket (#2896)
- [ ] log and raise and errors (parsing errors or errors found in the file)

rhysyngsun:
Based off of #2994 
----
Best way to see the true diff of this PR is [here](https://github.com/mitodl/micromasters/compare/nl/2896-result-file-audit-trail...nl/2791-process-exam-results)

Fixes #2791 

#### What's this PR do?
Adds support to process exam results

The easiest way to test this is to acquire a production result file (ping me for this) and modify your `ExamAuthorization` record for your user to match the result record in that file. Then run:
```python
from exams.pearson.download import ArchivedResponseProcessor
ArchivedResponseProcessor().process_zip('result.zip')
```
The method should return `(True, [])`.

Verify that this creates a `grades.models.ProctoredExamGrade` record with data matching what was in the exam file within the zip. Also verify that your `ExamAuthorization.exam_taken == True`.

exams/pearson/download.py (`process_exam_file()`)

Exams


gsidebo:
Here's what I ended up running based on your code sample:

```python
from exams.pearson import sftp
from exams.pearson.download import ArchivedResponseProcessor
with sftp.get_connection() as sftp_connection:
    res = ArchivedResponseProcessor(sftp_connection).process_zip('decrypted.zip')
print(res)
```

The value of `res` was `True`, not `(True, [])`. Is that OK?

gsidebo:
Functionality looks great (pending your feedback on the comment above). Looking at code

exams/pearson/constants.py
```diff
@@ -6,6 +6,23 @@
...
+
+# these are files we intentionally skip, but treat as if we processed them
+PEARSON_INTENDED_SKIP_FILE_TYPES = (
+    PEARSON_FILE_TYPE_CAND,
+    PEARSON_FILE_TYPE_SURV,
+    PEARSON_FILE_TYPE_CMNT,
+    PEARSON_FILE_TYPE_SECT,
+    PEARSON_FILE_TYPE_ITEM,
+    PEARSON_FILE_TYPE_RESP,
+)
```
gsidebo: Would be nice to namespace these and make a namedtuple (et al) called `PEARSON_FILE_TYPES`. We'd then get/set these values like `PEARSON_FILE_TYPES.CAND`, `PEARSON_FILE_TYPES.SURV` etc., but that can be done in a different PR

exams/pearson/download.py
```diff
@@ -50,6 +56,37 @@ def locally_extracted(zip_file, member):
...
+        ' student_id=`{student_id}`' if client_candidate_id else '',
+        ' username=`{username}`' if username else '',
+        ' error=`{error}`' if error else '',
+    ]).format(
+        authorization_id=client_authorization_id,
+        student_id=client_candidate_id,
+        username=username,
+        error=error,
+        **kwargs
+    )
```
gsidebo: Here's a possibly cleaner and probably-not-working way to rewrite this function:
 
```python
value_output_dict = dict(
    authorization_id=client_authorization_id,
    student_id=client_candidate_id,
    username=username,
    error=error
)
value_output = ' '.join(['{}={}'.format(k, v) for k, v in value_output_dict.items() if v])
formatted_message = ' '.join([message, value_output]).format(**kwargs)
log.error(formatted_message)
return formatted_message
```

exams/pearson/download_test.py
```diff
@@ -342,32 +346,29 @@ def test_process_result_vcdc(self):
...
-                ]
-            )
-            for profile in self.success_profiles:
-                profile.refresh_from_db()
-                assert profile.status == ExamProfile.PROFILE_SUCCESS
+            result, errors = self.processor.process_vcdc_file("/tmp/file.ext")
+
+        assert result is True
+        assert all(error.startswith('ExamProfile sync failed:') for error in errors)
+        assert 'error=`Bad address`' in errors[1]
```
gsidebo: Is this right? We're expecting backticks in the error message?

exams/pearson/constants.py
rhysyngsun: I'm not familiar with how you'd use a namedtuple in that way, would an `enum` be more appropriate here?

exams/pearson/download.py
```diff
@@ -233,38 +273,90 @@ def process_eac_file(self, extracted_file):  # pylint: disable=no-self-use
...
+                messages.append(format_and_log_error(
+                    'Unable to find a matching ExamAuthorization record:',
+                    client_candidate_id=result.client_candidate_id,
+                    client_authorization_id=result.client_authorization_id,
+                ))
+                continue
+
+            if result.grade.lower() not in EXAM_GRADES:
+                messages.append(format_and_log_error(
+                    'Unexpected grade value: grade=`{}`',
```
rhysyngsun: Good catch, must've missed that in my refactor of the logging code.

exams/pearson/download_test.py
rhysyngsun: Yes, that's the way they were originally formatted, I can change that if you think we should since I'm touching that code in this PR anyway.

rhysyngsun:
In response to [your comment](https://github.com/mitodl/micromasters/pull/3041#issuecomment-294223828) above, yes, that's the correct return type, my mistake.

exams/pearson/utils.py
```diff
@@ -33,17 +26,56 @@ def is_zip_file(filename):
     return bool(ZIP_FILE_RE.match(filename))
 
 
-def get_file_type(filename):
+def parse_or(parser, default=None):
```
rhysyngsun: Done

exams/pearson/utils.py
```diff
@@ -33,17 +26,56 @@ def is_zip_file(filename):
...
+    """
+    Generate a function that safely parses a value or returns the default
+
+    Args:
+        parser (callable): callable to parse the value
+        default (Any): default value if parser fails, default is None
+
+    Returns:
+        callable: function that safely wraps the inner parser
```
rhysyngsun: Done

exams/pearson/download.py
rhysyngsun: Actually ended up getting rid of this altogether (see my simplified rewrite of `format_and_log_error`).

exams/pearson/download.py
rhysyngsun: I refactored this to be even simpler, it now just formats all `kwargs` as `{}={}` which makes it much easier to use.

exams/pearson/constants.py
gsidebo: Not `namedtuple`, sorry (although it _could_ be used). I meant `SimpleNamespace` ([doc](https://docs.python.org/3/library/types.html#types.SimpleNamespace)). It would look something like this:

```python
PEARSON_FILE_TYPES = SimpleNamespace(
    EAC='eac',
    VCDC='vcdc',
    # ...
)

# Usage:
from exams.pearson.constants import PEARSON_FILE_TYPES
def is_eac_file_type(file_type):
    if filetype == PEARSON_FILE_TYPES.EAC:
        return True
```

gsidebo:
Looks great! 👍

Issue #3037, opened 2017-04-10
Fix selected search filter label regression


gsidebo:
Selected search filter labels were fixed in issue #3006, but the structure of the index was being changed in #2942 at the same time. As a result, #2942 was merged without fixing the selected search filter labels that had regressed. It should be pretty straightforward to fix these again. If there's a way for us to either test for these labels or to solve this problem in a general way that doesn't require us to change a bunch of constants, we really should do that. As it is, it is incredibly easy for this to regress again

![ss 2017-04-10 at 13 43 03](https://cloud.githubusercontent.com/assets/14932219/24875060/d227066e-1df4-11e7-8555-7ba44654c378.png)

amir-qayyum-khan:
it is working fine on MM local instance and master branch. My super admin and staff access at rc is revoked, could not able to see behaviour there.

<img width="954" alt="screen shot 2017-04-11 at 1 10 36 pm" src="https://cloud.githubusercontent.com/assets/10431250/24899100/8ccd17e2-1eb8-11e7-938f-e725035b0fd7.png">

amir-qayyum-khan:
@pdpinch 
what i did is build, migrate and reindex latest master and this was working fine

PR #3042, opened 2017-04-11
Fixed selected search filter label regression

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3037

#### What's this PR do?
Fixed a regression which is cased by https://github.com/mitodl/micromasters/issues/2942. So that selected search filter display fine.

- go to learners search page and select semester from facets.

@pdpinch 

screen shot 2017-04-11 at 5 21 45 pm

### DIFF
--- a/static/js/components/search/ModifiedSelectedFilter.js
+++ b/static/js/components/search/ModifiedSelectedFilter.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import R from 'ramda';
 import { SEARCH_FACET_FIELD_LABEL_MAP } from '../../constants';
 import { makeCountryNameTranslations } from '../LearnerSearch';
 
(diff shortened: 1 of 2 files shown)


static/js/components/search/ModifiedSelectedFilter.js
```diff
@@ -21,8 +22,9 @@ export default class ModifiedSelectedFilter extends React.Component {
       bemBlocks,
       filterId,
     } = this.props;
-
-    if (labelKey in SEARCH_FACET_FIELD_LABEL_MAP) {
+    if (R.isEmpty(labelKey)) {
```
aliceriot: in what case will this be `""`? just trying to understand the change.

amir-qayyum-khan: for the average grade it has empty key on local machine

aliceriot: ok

PR #3053, opened 2017-04-11
Fix race condition resulting in multiple emails sent

### LINKED ISSUE #3036: Mail: receiving multiple percolate emails when only updating profile once
Locally I'm randomly receiving multiple copies of percolate-triggered emails. It ends up resulting in the below stacktrace. I discussed this with @noisecapella and suggested these changes as options to reduce the duplicates:

#### To have a "at least once" guarantee of mail sending:

- Create `SentAutomaticEmail` with status = pending
- Create transaction
- Use `select_for_update()` to acquire row-level lock on `SentAutomaticEmail` created above
- Send email to Mailgun
- Update `SentAutomaticEmail` status = sent
- Commit transaction

#### To have a "at most once" guarantee of mail sending:

- Create `SentAutomaticEmail` with status = pending
- Create transaction
- Use `select_for_update()` to acquire row-level lock on `SentAutomaticEmail` created above
- Update `SentAutomaticEmail` status = sent
- Commit transaction
- Send email to Mailgun

cc @pdpinch, probably a product decision which type of guarantee of send we want.

```
celery_1    | [2017-04-10 16:04:05,150: ERROR/Worker-6] IntegrityError: SentAutomaticEmail was likely already created
celery_1    | Traceback (most recent call last):
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
celery_1    |     return self.cursor.execute(sql, params)
celery_1    | psycopg2.IntegrityError: duplicate key value violates unique constraint "mail_sentautomaticemail_user_id_2b0d2dde_uniq"
celery_1    | DETAIL:  Key (user_id, automatic_email_id)=(242, 45) already exists.
celery_1    |
celery_1    |
celery_1    | The above exception was the direct cause of the following exception:
celery_1    |
celery_1    | Traceback (most recent call last):
celery_1    |   File "/src/mail/api.py", line 288, in send_automatic_emails
celery_1    |     log.exception("Error sending mailgun mail for automatic email %s", automatic_email)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
celery_1    |     return getattr(self.get_queryset(), name)(*args, **kwargs)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/query.py", line 399, in create
celery_1    |     obj.save(force_insert=True, using=self.db)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/base.py", line 796, in save
celery_1    |     force_update=force_update, update_fields=update_fields)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/base.py", line 824, in save_base
celery_1    |     updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/base.py", line 908, in _save_table
celery_1    |     result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/base.py", line 947, in _do_insert
celery_1    |     using=using, raw=raw)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
celery_1    |     return getattr(self.get_queryset(), name)(*args, **kwargs)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/query.py", line 1045, in _insert
celery_1    |     return query.get_compiler(using=using).execute_sql(return_id)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/models/sql/compiler.py", line 1054, in execute_sql
celery_1    |     cursor.execute(sql, params)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/backends/utils.py", line 79, in execute
celery_1    |     return super(CursorDebugWrapper, self).execute(sql, params)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/raven/contrib/django/client.py", line 114, in execute
celery_1    |     return real_execute(self, sql, params)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
celery_1    |     return self.cursor.execute(sql, params)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/utils.py", line 94, in __exit__
celery_1    |     six.reraise(dj_exc_type, dj_exc_value, traceback)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/utils/six.py", line 685, in reraise
celery_1    |     raise value.with_traceback(tb)
celery_1    |   File "/usr/local/lib/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
celery_1    |     return self.cursor.execute(sql, params)
celery_1    | django.db.utils.IntegrityError: duplicate key value violates unique constraint "mail_sentautomaticemail_user_id_2b0d2dde_uniq"
celery_1    | DETAIL:  Key (user_id, automatic_email_id)=(242, 45) already exists.
celery_1    |
```

Update
---
On further testing, this appears to caused by one `index_users` tasks being triggered for each `Profile`, `Employment`, and `Education` record a user has (see the `post_save` signals in search/signals.py) because we `PATCH` the entire profile structure including the nested `Education` and `Employment` records when we update any of that data and DRF correctly re-saves the data (often with no change) and triggers the `post_save` signal for each record, causing a race condition. After discussing with George, we both agreed we shouldn't change the signals, but instead go ahead with one of the options I described above.

noisecapella:
Fixes #3036

#### What's this PR do?
Uses select_for_update to lock the `SentAutomaticEmail` row until after the email was sent. It also refactors `mark_emails_as_sent` to be a context manager used wherever we need to mark automatic emails as sent

Add a few `Education` and `Employment` entries to your profile and save. As a separate user create an `AutomaticEmail` for your user using the learner search page interface. Switch back to your original user, update your profile and you should see only one email sent to that user's email.


mail/api.py
```diff
@@ -322,18 +321,26 @@ def add_automatic_email(original_search, email_subject, email_body, sender_name,
...
+            sentautomaticemail__automatic_email=automatic_email,
+        ).values_list('id', flat=True)
+
+        for user_id in user_ids:
+            SentAutomaticEmail.objects.create(
+                user_id=user_id,
+                automatic_email=automatic_email,
+                status=SentAutomaticEmail.PENDING,
+            )
+        SentAutomaticEmail.objects.filter(user_id__in=user_ids).select_for_update()
```
rhysyngsun: Here you're locking all rows in `SentAutomaticEmail` for the list of users, I think you only want to lock the ones that  are in pending status for the particular `AutomaticEmail` you're trying to send here.

mail/api.py
```diff
@@ -322,18 +321,26 @@ def add_automatic_email(original_search, email_subject, email_body, sender_name,
...
+        yield
+        SentAutomaticEmail.objects.filter(user_id__in=user_ids).update(status=SentAutomaticEmail.SENT)
```
rhysyngsun: Should be updating the same set of `SentAutomaticEmail`shere that you selected for update above.

mail/api.py
```diff
@@ -322,18 +321,26 @@ def add_automatic_email(original_search, email_subject, email_body, sender_name,
...
-            user=user,
-            automatic_email=automatic_email,
-        )
+    with transaction.atomic():
+        user_ids = User.objects.filter(email__in=emails).exclude(
+            sentautomaticemail__automatic_email=automatic_email,
+        ).values_list('id', flat=True)
+
+        for user_id in user_ids:
+            SentAutomaticEmail.objects.create(
```
rhysyngsun: I think you should move this out of the transaction and switch back to `get_or_create` because a) you only want to send emails that you have an atomic guarantee you were the only one who did so. The best way I know to do that is to create them outside the transaction so they get committed and then only `select_for_update` and yield the emails for those which you actually created and then pass that into `send_batch`. Otherwise, you could yield down below and then the entire transaction fails to commit because you couldn't create one record or some other race condition.

noisecapella:
Comments addressed

mail/api.py
```diff
@@ -272,17 +273,18 @@ def send_automatic_emails(program_enrollment):
...
-                sender_name=automatic_email.sender_name,
-            )
-            SentAutomaticEmail.objects.create(
-                user=user,
-                automatic_email=automatic_email,
-            )
+            with mark_emails_as_sent(automatic_email, [user.email]) as user_ids:
+                # user_ids should just contain user.id except when we already sent the user the email
+                # in a separate process
+                for recipient_email in User.objects.filter(id__in=user_ids).values_list('email', flat=True):
```

Issue #3057, opened 2017-04-12
refactor API frontend code to use redux-rest


aliceriot:
We now have the redux-rest abstraction which we can use to eliminate a lot of boilerplate around hitting some of our APIs. We should do the work to switch them all (or as many as is practical) over to it.

aliceriot:
^^ just opened a PR to do the course enrollments reducer

aliceriot:
looks like I did the course price API too!

PR #3058, opened 2017-04-12
refactor course enrollments 

aliceriot:

part of #3057 

#### What's this PR do?

This refactors the course enrollments front-end API code to use the redux-rest abstraction we wrote recently.

Make sure you can add a course enrollment, and also that the failure case is handled well.


static/js/containers/DashboardPage.js
```diff
@@ -429,7 +430,11 @@ class DashboardPage extends React.Component {
 
   addCourseEnrollment = (courseId: string): Promise<*> => {
     const { dispatch } = this.props;
-    return dispatch(addCourseEnrollment(courseId)).catch(() => {
+    return dispatch(actions.courseEnrollments.post(courseId)).then(courseId => {
```
rhysyngsun: There's a bug here, the argument `course_id` for the function passed to `then` is the same as the argument to `addCourseEnrollment` so it ends up passing a course object (instead of id) to `showEnrollPayLaterSuccessMessage` and the message doesn't show up (but doesn't error because the comparison is valid).

aliceriot: I didn't quite realize what that did! I think it should be fixed now.

PR #3061, opened 2017-04-12
Preserve search URL on reload

### LINKED ISSUE #3003: Search: Preserve search URL after refresh
After a successful refresh the query parameters are removed from the URL. We should change this behavior to preserve them so the user can do multiple refreshes

noisecapella:
Fixes #3003

#### What's this PR do?
Changes `ProgramFilter` to use a `StatefulAccessor` which lets us propagate changes in the current enrollment to the state of the accessor when they happen. This lets us remove the workaround which calls `searchkit.resetState()`  which caused the existing query parameters to be removed.

Go to `/learners/?q=xyz` and verify that you see a text box with `xyz` in it. Click a facet then refresh the browser. It should remain selected after refresh. Switch programs and you should see the list of people change accordingly.

This changes the behavior where we clear the search when switching between programs. I'm not sure what the ideal is here so I just left it alone.

### DIFF
--- a/static/js/components/ProgramFilter.js
+++ b/static/js/components/ProgramFilter.js
@@ -16,7 +16,7 @@ class ProgramFilterAccessor extends StatefulAccessor {
     this.state = new State();
   }
 
-  buildOwnQuery(query) {
+  buildOwnQuery(query: Object) {
     let programId = this.state.getValue();
     if (_.isNil(programId)) {
       return query;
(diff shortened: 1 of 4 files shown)


static/js/components/ProgramFilter.js
```diff
@@ -1,35 +1,68 @@
...
 
+class ProgramFilterAccessor extends StatefulAccessor {
+  constructor() {
+    super();
+
+    this.state = new State();
+  }
+
+  buildOwnQuery(query: Object) {
+    let programId = this.state.getValue();
```
rhysyngsun: Nit: could be `const`.

rhysyngsun:
Manual testing mostly looked good, although I noticed one behavior that seemed out-of-place when switching from one program to another having selected a program-specific facet (i.e. Course):

selection_053

noisecapella:
@rhysyngsun Yeah, maybe I should just clear all facets when switching between programs just in case

noisecapella:
It looks like the selenium test caught a race condition which occurred when the currently selected program wasn't in localStorage yet. I think I fixed it in the latest commit

rhysyngsun:
I agree, clearing the facets on program switch is probably best, at least for now since it can lead to some funky state.

noisecapella:
Ok, I think I've addressed everything and added some selenium tests to cover these cases

PR #3067, opened 2017-04-14
Restored final grade histogram for selected courses in learner search

### LINKED ISSUE #2939: Search: course final grades facet filters is not restricted to just selected course
We need to change the indexing to filter the final grades per program

Similar to #2942

### LINKED ISSUE #3070: Refactor custom nested searchkit accessors
We implemented custom solutions for filtering on course enrollments (#2942) and for filtering on a histogram of final grades for a course (#2939). The implementations are in `static/js/components/search/NestedAggregatingMenuFilter.js` and `static/js/components/search/FinalGradeRangeFilter.js` respectively. Both filters are essentially providing us a way to apply 'AND' filters on nested documents, and as such they share a lot of functionality. Unfortunately this shared functionality resulted in a lot of code repetition, which is there because there's a somewhat complicated inheritance pattern. The accessors share a lot of implementation details, but they also need to subclass from Searchkit classes in order to get some base functionality & properties. This issue covers the work to get rid of that repetition and compose these classes in a more sensible way

gsidebo:
Closes #2939 
Closes #3070 

#### What's this PR do?
Restores the final grade histogram for selected courses in learner search, and makes it work with the selected course.

Filter by a course in the learner search page, move the histogram bounds to filter by grade ranges, remove the course filter to make sure that the histogram also disappears, and test the histogram's interoperability with other available filters.

`static/js/components/search/NestedAggregatingMenuFilter.js` , `static/js/components/search/FinalGradeRangeFilter.js`, `static/js/components/search/util.js`

~~There's a lot of repeated code between the two files I mentioned above. I did my best to keep the implementation differences separate and obvious. The repetition is there because there's a somewhat complicated inheritance pattern. The accessors need to subclass from Searchkit classes in order to get some base functionality & properties, and they also share a lot of the same custom implementation that we've created. I'll open an issue to address this repetition *(issue added: #3070)*~~ The work described here is already merged into this PR

ss 2017-04-14 at 09 53 30


gsidebo:
FYI: There are 3 JS tests failing and I'm having trouble figuring out the cause. I think it has something to do with our static test data. Working on that now

annagav:
Functionality looks great!
 The only thing I found was the courses facet disappearing when no results caused by the range in the final grade, but the selected course title stays, so that helps.
screen shot 2017-04-19 at 2 53 31 pm

gsidebo:
@annagav good catch. definitely a bug. i'll look into it

gsidebo:
@annagav Fixed. The visibility toggle stays around when an empty final grade range is chosen. Still looks just a little wonky with an empty result set (still no doc count displayed in the facet) but it's definitely more consistent with the visibility toggle still there

ss 2017-04-19 at 16 05 09

annagav:
It works now for the final grade, but not for 'Average Grade in Program'

screen shot 2017-04-19 at 5 18 21 pm

gsidebo:
@annagav Looks like these histogram UI issues exist on master. I added an issue to fix them https://github.com/mitodl/micromasters/issues/3095

giocalitri:
@gsidebo do you need to rebase master for this PR?

gsidebo:
@giocalitri just rebased

Issue #3071, opened 2017-04-14
Fixed recipients representation on search email


amir-qayyum-khan:
### See this
<img width="1061" alt="screen shot 2017-04-15 at 12 54 35 am" src="https://cloud.githubusercontent.com/assets/10431250/25054368/29b6aee6-2176-11e7-95aa-a27b611cef4d.png">

### and this (fixed this)
<img width="786" alt="screen shot 2017-04-15 at 12 54 43 am" src="https://cloud.githubusercontent.com/assets/10431250/25054378/3eea3d96-2176-11e7-81dc-9c083c5fc426.png">

PR #3074, opened 2017-04-17
[Regression] Fixed recipient keys on email composition dialog

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3071

#### What's this PR do?
Fixes regression of #2942, key if recipients object.  

select filter and open email dialog.

@pdpinch 
screen shot 2017-04-17 at 6 39 09 pm

screen shot 2017-04-17 at 6 39 13 pm

### DIFF
--- a/static/js/components/email/EmailCompositionDialog_test.js
+++ b/static/js/components/email/EmailCompositionDialog_test.js
@@ -172,6 +172,10 @@ describe('EmailCompositionDialog', () => {
         id: '2',
         name: "ES",
         value: "foo"
+      },{
+        id: '3',
+        name: "profile.birth_country",
+        value: "ES"
       }]
     }, { renderRecipients: SEARCH_RESULT_EMAIL_CONFIG.renderRecipients });
 
@@ -182,6 +186,10 @@ describe('EmailCompositionDialog', () => {
     assert.include(
       getDialog().querySelector('.sk-selected-filters').textContent,
       "Spain: foo"
+    );
+    assert.include(
+      getDialog().querySelector('.sk-selected-filters').textContent,
+      "Country of Birth: Spain"
     );
   });
 });
--- a/static/js/components/email/EmailCompositionDialog_test.js
+++ b/static/js/components/email/EmailCompositionDialog_test.js
@@ -168,12 +168,20 @@ describe('EmailCompositionDialog', () => {
         id: '1',
         name: "program.enrollments.semester",
         value: "2015"
+      }, {
+        id: '2',
+        name: "ES",
+        value: "foo"
       }]
     }, { renderRecipients: SEARCH_RESULT_EMAIL_CONFIG.renderRecipients });
 
     assert.include(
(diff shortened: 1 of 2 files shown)


static/js/components/email/lib.js
```diff
@@ -12,17 +12,31 @@ import type { Profile } from '../../flow/profileTypes';
...
-      {filter.name}: {filter.value}
+const countryNameTranslations: Object = makeCountryNameTranslations();
+const renderFilterOptions = R.map(filter => {
+  let labelKey;
+  if (R.isEmpty(filter.name)) {
+    labelKey = SEARCH_FACET_FIELD_LABEL_MAP[filter.id];
+  } else if (filter.name in SEARCH_FACET_FIELD_LABEL_MAP) {
+    labelKey = SEARCH_FACET_FIELD_LABEL_MAP[filter.name];
+  } else if (filter.name in countryNameTranslations) {
+    labelKey = countryNameTranslations[filter.name];
```
singingwolfboy: Could you add a test for this line? I see that there's already an automated test that turns `program.enrollments.semester` into `Semester`, which is great. But it would be good to also verify that `ES` gets turned into `Spain`, for example.

amir-qayyum-khan:
Done @singingwolfboy

static/js/components/email/EmailCompositionDialog_test.js
```diff
@@ -166,14 +166,22 @@ describe('EmailCompositionDialog', () => {
...
 
     assert.include(
-      getDialog().querySelector('.sk-selected-filters-option__name').textContent,
-      "key: test"
+      getDialog().querySelector('.sk-selected-filters').textContent,
+      "Semester: 2015"
+    );
+    assert.include(
+      getDialog().querySelector('.sk-selected-filters').textContent,
+      "Spain: foo"
```
singingwolfboy: This doesn't seem like a valid test case. A valid test case might be `profile.birth_country: ES` converted into `Country of Birth: Spain`. In fact, I just manually tested that, and it doesn't seem like your pull request correctly does that conversion.

amir-qayyum-khan:
Done @singingwolfboy

static/js/components/email/EmailCompositionDialog_test.js
singingwolfboy: If this isn't a valid test case, can you remove it?

amir-qayyum-khan: this logic is also present in `static/js/components/search/ModifiedSelectedFilter.js`, 
which is code to use in displaying facets.

singingwolfboy: Ah, thanks for pointing that out. In that case, it _is_ a valid test case.

amir-qayyum-khan: these recipients are same as filters display on learners search page

Issue #2155, opened 2016-12-19
Search: decrease the default number of facet hits for "Current Residence"


pdpinch:
Similar in goal to #2038, but as mentioned on that issue, the "Current Residence" facet uses a hierarchical component that doesn't support limiting the number of hits out of the box. 

We want to try setting a max height in the CSS and therby force a scollbar.

PR #3076, opened 2017-04-18
Added max height and scroll to Current residence

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2155

#### What's this PR do?
changes in css to add scroll and max height to current residence.

Go to learners page and see facet country resident, you must have more the 7 countries there.

@pdpinch 
screen shot 2017-04-18 at 10 23 17 pm

### DIFF
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -59,18 +59,6 @@
     .sk-hierarchical-menu-list {
       padding-bottom: 10px;
       position: relative;
-
-      &:after {
-        content: '';
-        width: 0;
-        height: 0;
-        border-left: 7px solid transparent;
-        border-right: 7px solid transparent;
-        border-top: 7px solid $font-gray;
-        position: absolute;
-        right: 10px;
-        bottom: -3px;
-      }
     }
 
     // Search facet headers
@@ -139,8 +127,12 @@
         justify-content: space-between;
       }
 
-      .sk-hierarchical-menu-list__hierarchical-options > div {
-        align-items: stretch;
+      .sk-hierarchical-menu-list__hierarchical-options {
+        margin-right: 10px;
+
+        > div {
+          align-items: stretch;
+        }
       }
     }
 
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -64,11 +64,11 @@
         content: '';
         width: 0;
         height: 0;
-        border-left: 6px solid transparent;
-        border-right: 6px solid transparent;
-        border-top: 6px solid $font-gray;
+        border-left: 7px solid transparent;
+        border-right: 7px solid transparent;
+        border-top: 7px solid $font-gray;
         position: absolute;
-        right: 8px;
+        right: 10px;
         bottom: -3px;
       }
     }
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -58,7 +58,7 @@
 
     .sk-hierarchical-menu-list {
       max-height: 150px;
-      overflow: auto;
+      overflow-y: scroll;
     }
 
     // Search facet headers
--- a/static/scss/search-page.scss
+++ b/static/scss/search-page.scss
@@ -57,8 +57,20 @@
     }
 
     .sk-hierarchical-menu-list {
-      max-height: 150px;
-      overflow-y: scroll;
+      padding-bottom: 10px;
+      position: relative;
+
+      &:after {
+        content: '';
+        width: 0;
+        height: 0;
+        border-left: 6px solid transparent;
+        border-right: 6px solid transparent;
+        border-top: 6px solid $font-gray;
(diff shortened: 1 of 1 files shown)


pdpinch:
I don't see the scroll bar in your screenshot? Where does it display?

amir-qayyum-khan:
my osx settings hides scroll bar after use

static/scss/search-page.scss
```diff
@@ -56,6 +56,11 @@
       padding: 0 10px 0 23px;
     }
 
+    .sk-hierarchical-menu-list {
+      max-height: 150px;
+      overflow: auto;
```
aliceriot: just one comment: it might be better to use `overflow: scroll;` to just be a little more intentional about the behavior we want.

static/scss/search-page.scss
```diff
@@ -56,6 +56,11 @@
       padding: 0 10px 0 23px;
     }
 
+    .sk-hierarchical-menu-list {
+      max-height: 150px;
+      overflow-y: scroll;
```
singingwolfboy: This still doesn't cause the scrollbars to show up all the time, as far as my testing shows.

amir-qayyum-khan:
Done @aliceriot @pdpinch @singingwolfboy 

screen shot 2017-04-18 at 10 23 17 pm

static/scss/search-page.scss
aliceriot: Yeah, this just sets how an overflow should be handled. Scrollbar display is left up to the browser. most these days do not display a scrollbar all the time, and I think all browsers on macOS will use a system scrollbar, which by default is not persistant. getting a persistant, cross-platform scrollbar unfortunately requires using a JS reimplementation of the scrolling already implemented by the browser, which is both yucky and unfortunate.

amir-qayyum-khan: you are right @aliceriot

pdpinch:
Can we solve it this way?

http://stackoverflow.com/questions/2295766/how-to-create-a-scrollable-div-tag-vertically/2297269#2297269

it seems to default to being "on" on my Mac (Firefox and Chrome. Haven't tried Safari yet).

aliceriot:
See my comment above, but no, that won't work on all platforms. My guess is that if your scrollbars are persistent you have your computer configured that way (macOS has a setting to not hide system scrollbars, or at least it did the last time I used a mac regularly). Here's a SO answer that explains the situation in a little more depth: https://stackoverflow.com/questions/18317634/force-visible-scrollbar-in-firefox-on-mac-os-x

amir-qayyum-khan:
this is mac setting we are talking about.

screen shot 2017-04-18 at 11 16 16 pm

pdpinch:
Thanks. Looks like I have mine set to `Always`. I hardly ever change defaults -- except for maybe that one.

amir-qayyum-khan:
@pdpinch @aliceriot @singingwolfboy Added an indicator for scroll, please give me feedback.

screen shot 2017-04-20 at 1 10 21 pm

pdpinch:
@roberthouse54 what is UX best practice here, when the user's system hides scroll bars. is there a standard way to hint to the user that the area is scrollable? 

Amir added a little down arrow, but I don't know if that will be generally recognized.

roberthouse54:
An arrow is not best practice. My feeling is that we should do nothing. Users on computers that hide scrollbars will be used to this behavior. I have see this build, and I think it is basically fine. We should lose the arrow, since it does not convey that there is a scrollbar.

One thing I would like to see changed is that the scrollbar seems to cover the numbers in the right column, although this looks different than the screenshot Amir posted above. I don't know if this is easily fixable. (See below)

image

pdpinch:
Ok. I agree. Let's remove the arrow. 

@roberthouse54 what browser were you using when you saw the overlap? @amir-qayyum-khan is there a way to fix that?

amir-qayyum-khan:
Done @roberthouse54

roberthouse54:
@pdpinch I was using Chrome Version 57.0.2987.133 (64-bit) on Mac OX 10.11.6.

pdpinch:
@singingwolfboy can you re-review this? We've decided the absence of scroll bars on (some) Macs is a user preference that we shouldn't try to interfere with.

singingwolfboy:
This works correctly, but adjusting the numbers to account for the scrollbar means that they are now misaligned compared to all the other numbers. Here's a screenshot to demonstrate:

screen shot demonstrating offset numbers

At this point, I think it's still worth merging this PR, in spite of the misalignment. Design is a series of compromises, and I think this is a compromise we can live with. Maybe a future pull request can move _all_ the numbers to the left, so that any of them can use a scrollbar. @pdpinch, what do you think?

pdpinch:
I agree with DB. Let's open another issue for the alignment issue and get this one merged!

Issue #449, opened 2016-06-01
Add page titles


gsidebo:
All of the Profile pages have the title of "MIT Micromasters". As a result, a user's browser history (as revealed by right-clicking the back button) might look something like this:

![ss 2016-05-31 at 11 50 03 am](https://cloud.githubusercontent.com/assets/14932219/15726256/db55e494-281d-11e6-92d3-2b340f2894cb.png)

For this and other reasons, it would be wise to add some helpful titles to the Profile pages.

pdpinch:
First draft of page titles

| Path                   | Title                                                                        |   |
|------------------------|------------------------------------------------------------------------------|---|
| /                      | MIT MicroMasters:  An online MIT credential available to anyone in the world |   |
| /:program_slug         | [Program Title] \| MIT MicroMasers                                            |   |
| /terms_of_service/     | Terms of Service \| MIT MicroMasters                                          |   |
| /profile/personal/     | Personal Profile \| MIT MicroMasters                                          |   |
| /profile/education/    | Education Profile \| MIT MicroMasters                                         |   |
| /profile/professional/ | Professional Profile \| MIT MicroMasters                                      |   |
| /profile/privacy/      | Privacy \| MIT MicroMasters                                                   |   |
| /dashboard/            | Learner Dashboard \| MIT MicroMasters                                         |   |
| /users/:username       | [Preferred Name] [Last Name] \| MIT MicroMasters                              |   |
|                        |                                                                              |   |

gsidebo:
@pdpinch i imagine we want these with `'s` at the end?

noisecapella:
We're now doing just `MicroMasters` without the apostrophe: #557

roberthouse54:
First draft of page titles

| Path | Title |  |
| --- | --- | --- |
| / | MIT MicroMasters:  An online MIT credential available to anyone in the world |  |
| /:program_slug | [Program Title] | MIT MicroMasers |  |
| /terms_of_service/ | Terms of Service | MIT MicroMasters |  |
| /profile/personal/ | Personal Profile | MIT MicroMasters |  |
| /profile/education/ | Education Profile | MIT MicroMasters |  |
| /profile/professional/ | Professional Profile | MIT MicroMasters |  |
| /profile/privacy/ | Privacy | MIT MicroMasters |  |
| /dashboard/ | Dashboard | MIT MicroMasters |  |
| /users/:username | [First Name] [Last Name] | MIT MicroMasters |  |
|  |  |  |

PR #3081, opened 2017-04-19
Added page titles all over the MM app

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/449

#### What's this PR do?
- Adds page title all over the application.

Go to each page in app and see the tab title.

@pdpinch 
#### screenshot
screen shot 2017-04-19 at 3 46 12 pm

screen shot 2017-04-19 at 3 43 52 pm

screen shot 2017-04-19 at 3 44 11 pm

screen shot 2017-04-19 at 3 44 46 pm

screen shot 2017-04-19 at 3 45 33 pm

### DIFF
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -6,7 +6,7 @@
 {% wagtailuserbar %}
 {% block body_class %}template-programpage{% endblock %}
 
-{% block title %}MIT MicroMasters: An online MIT credential available to anyone in the world{% endblock %}
+{% block title %}MITx MicroMasters: Bringing MIT to you{% endblock %}
 {% block description %}
   MicroMasters is a professional and academic credential for online learners from anywhere in the world. Learners who pass an integrated set of MITx graduate-level courses on edX.org, and one or more proctored exams, will earn a MicroMasters credential from MITx, and can then apply for an accelerated, on campus, master’s degree program at MIT or other top universities.
 {% endblock %}
--- a/cms/templates/cms/home_page.html
+++ b/cms/templates/cms/home_page.html
@@ -6,7 +6,7 @@
 {% wagtailuserbar %}
 {% block body_class %}template-programpage{% endblock %}
 
-{% block title %}MITx MicroMasters: Bringing MIT to you{% endblock %}
+{% block title %}MIT MicroMasters: An online MIT credential available to anyone in the world{% endblock %}
 {% block description %}
   MicroMasters is a professional and academic credential for online learners from anywhere in the world. Learners who pass an integrated set of MITx graduate-level courses on edX.org, and one or more proctored exams, will earn a MicroMasters credential from MITx, and can then apply for an accelerated, on campus, master’s degree program at MIT or other top universities.
 {% endblock %}
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -4,7 +4,7 @@
 {% load wagtailcore_tags %}
 {% load wagtailimages_tags %}
 
-{% block title %}{{ page.title }}{% endblock %}
+{% block title %}{{ page.title }} MicroMasters{% endblock %}
 {% block description %}{% if page.program.description %}{{ page.program.description }}{% endif %}{% endblock %}
 {% block keywords %}micromasters, MIT, {{ page.program.title }}, online master's, online courses, MOOC{% endblock %}
 
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -4,7 +4,7 @@
 {% load wagtailcore_tags %}
 {% load wagtailimages_tags %}
 
-{% block title %}{{ page.title }} MicroMasters{% endblock %}
+{% block title %}{{ page.title }}{% endblock %}
 {% block description %}{% if page.program.description %}{{ page.program.description }}{% endif %}{% endblock %}
 {% block keywords %}micromasters, MIT, {{ page.program.title }}, online master's, online courses, MOOC{% endblock %}
 
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "react-bootstrap": "^0.30.8",
     "react-cropper": "^0.10.1",
     "react-datepicker": "^0.41.1",
+    "react-document-title": "^2.0.3",
     "react-dom": "^15.4.0",
     "react-dropzone": "^3.6.0",
     "react-ga": "^2.1.2",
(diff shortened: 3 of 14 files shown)


cms/templates/cms/home_page.html
```diff
@@ -6,7 +6,7 @@
 {% wagtailuserbar %}
 {% block body_class %}template-programpage{% endblock %}
 
-{% block title %}MITx MicroMasters: Bringing MIT to you{% endblock %}
+{% block title %}MIT MicroMasters: An online MIT credential available to anyone in the world{% endblock %}
```
pdpinch: Revert this change. The home page title is good as is.

cms/templates/cms/program_page.html
```diff
@@ -4,7 +4,7 @@
 {% load wagtailcore_tags %}
 {% load wagtailimages_tags %}
 
-{% block title %}{{ page.title }} MicroMasters{% endblock %}
+{% block title %}{{ page.title }}{% endblock %}
```
pdpinch: This is also OK as is. It's useful to append 'MicroMasters' to the program name.

static/js/containers/LearnerPage.js
```diff
@@ -39,6 +39,7 @@ class LearnerPage extends React.Component<*, LearnerPageProps, *> {
     const { params: { username }, fetchProfile } = this.props;
     fetchProfile(username);
     this.fetchDashboard();
+    document.title = "Learner's Profile";
```
pdpinch: Is it possible to include the Learner's Full name here? i.e. `[Preferred Name] [Last Name]'s MicroMasters Profile`

static/js/containers/ProfilePage.js
```diff
@@ -24,6 +24,7 @@ class ProfilePage extends React.Component<*, ProfileContainerProps, *> {
   componentDidMount() {
     const { scrollToError } = this.props;
     scrollToError.call(this);
+    document.title = "Personal Profile";
```
pdpinch: How is this different from LearnerPage.js above?

ui/templates/terms_of_service.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% load i18n static %}
 
-{% block title %}{% trans "MITx MicroMasters: Terms of Service" %}{% endblock %}
+{% block title %}{% trans "Terms of Service" %}{% endblock %}
```
pdpinch: Let's leave this as is.

static/js/containers/LearnerPage.js
amir-qayyum-khan: can i use `micromasters/static/js/util/util.js` line  ``getUserDisplayName``

static/js/containers/ProfilePage.js
aliceriot: this is the container component for all of the profile signup flow pages

amir-qayyum-khan: this is title for 3 profile tabs, personal, eduction, work, each tab has its own title

amir-qayyum-khan:
Done @aliceriot

static/js/containers/LearnerPage.js
```diff
@@ -95,11 +97,16 @@ class LearnerPage extends React.Component<*, LearnerPageProps, *> {
...
       };
       toRender = childrenWithProps(children, props);
+      name = `${profile.profile.preferred_name} ${profile.profile.last_name}'s`;
     }
     const { errorInfo } = profile;
-    return <Loader loaded={loaded}>
-      {errorInfo && loaded ? <ErrorMessage errorInfo={errorInfo} /> : toRender }
-    </Loader>;
+    return (
+      <DocumentTitle title={`${name} MicroMasters Profile`}>
```
aliceriot: If I do a full refresh on this page (`/learner`) I see `undefined undefined MicroMasters Profile` for a second while the request for the profile data is in-flight.

I think we should take a different approach - maybe we should have a little helper function that formats the title, something like:

```js
getDocumentTitle = () => {
  const {
    params: { username },
    profiles,
  } = this.props;
  
  if ( profiles[username] !== undefined ) {
    let first = R.propOr('', 'preferred_name', profiles[username]);
    let last = R.propOr('', 'last_name', profiles[username]);
    return `${first} ${last} MicroMasters Profile`.trim();
  }
  return 'MicroMasters Profile';
}
```

probably there's a more elegant way to express that but you get the idea. We don't ever want to stick `undefined` into a template, because `${undefined}` evaluates to the string `"undefined"`.

package.json
```diff
@@ -79,6 +79,7 @@
     "react-bootstrap": "^0.30.8",
     "react-cropper": "^0.10.1",
     "react-datepicker": "^0.41.1",
+    "react-document-title": "^2.0.3",
```
aliceriot: oh, looks like you just did! carry on.

amir-qayyum-khan: Done @aliceriot

pdpinch:
We're missing the site name from all these titles. I'm going to update the table on the issue and hopefully make it clearer.

pdpinch:
Ok, I updated the description with how the page titles should look. Basically, any of the ones that we're changing in this PR should end with a pipe and the site same, i.e. `| MITx MicroMasters` -- just for branding. 

@amir-qayyum-khan if you have any questions about the list on the issue, please ask.

amir-qayyum-khan:
Done @pdpinch @aliceriot

static/js/containers/LearnerPage.js
```diff
@@ -72,6 +73,20 @@ class LearnerPage extends React.Component<*, LearnerPageProps, *> {
...
+  getDocumentTitle = () => {
+    const {
+      params: { username },
+      profiles,
+    } = this.props;
+
+    if (profiles[username] !== undefined) {
+      let first = R.propOr('', 'preferred_name', profiles[username]);
+      let last = R.propOr('', 'last_name', profiles[username]);
+      return `${first} ${last} | MicroMasters Profile`.trim();
```
aliceriot: did you test this to see if it works? it does not, because `profiles[username]` is an object with a key called `profile` that points to the actual profile. You'll need to do this instead:

```js
R.pathOr('', ['profile', 'preferred_name'], profiles[username]);
```

maybe a more elegant approach overall would be to do this:

```js
getDocumentTitle = () => {
  const {
    params: { username },
    profiles,
  } = this.props;
  let profilePath = [username, 'profile'];

  let first = R.pathOr('', profilePath.concat('preferred_name'), profiles);
  let last = R.pathOr('', profilePath.concat('last_name'), profiles);

  return `${first} ${last} | MITx MicroMasters Profile`
    .trim()
    .replace(/^\|\s/, '');
};
```

the current solution can, if `first` and `last` end up being `""`, render the page title as `| MicroMasters Profile`.

amir-qayyum-khan:
Done @aliceriot

Issue #2432, opened 2017-01-25
JavaScript error with date picker in FinancialAidCard


pdpinch:
This doesn't happen too frequently, but this feature isn't used that frequently. The issue seems to be limited to Chrome

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/190476376/

```
TypeError: Cannot read property 'format' of null
  at onChange (./static/js/components/dashboard/FinancialAidCard.js:94:57)
...
(16 additional frame(s) were not displayed)
```

singingwolfboy:
Looks like the problematic code is here:
```jsx
<DatePicker
  selected={moment(documentSentDate)}
  onChange={(obj) => setDocumentSentDate(obj.format(ISO_8601_FORMAT))}
/>
```

I'm guessing that if the user clicks "submit" while the datepicker is blank, then the `obj` parameter will be `null`. The proper solution here is to validate that the datepicker has a value before submitting. However, this means we need an error state for this form, so that we can indicate to the user that the value is required. This might require a bit of design work.

amir-qayyum-khan:
#### To produce the issues:
On master branch, open browser console and remove date from FA calculator. You will see exception.

<img width="843" alt="screen shot 2017-04-19 at 4 54 27 pm" src="https://cloud.githubusercontent.com/assets/10431250/25178858/fcbbd750-2520-11e7-8fff-21614eecbbce.png">

PR #3082, opened 2017-04-19
Fixed issue where date change is empty

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2432

#### What's this PR do?
It fixes the issue on dashboard, which appears when you clear a date field using keyboard from FA calculator, when state is pending-doc. This is because date object is null at that time.

#### Note:
We dont need validation here because date widget is sticky, when u remove date and focus out the text field it automatically picks the current date.

- Go To dashboard, 
- Land to FA calculator with state pending doc. 
- Try to clear date using back space/delete
- See the console, there must be no error
- Focus out the text field, you will see the current date on it

@pdpinch 
screen shot 2017-04-19 at 4 54 27 pm

### DIFF
--- a/static/js/components/dashboard/FinancialAidCard.js
+++ b/static/js/components/dashboard/FinancialAidCard.js
@@ -57,9 +57,15 @@ export default class FinancialAidCard extends React.Component {
     updateDocumentSentDate(financialAidId, documents.documentSentDate);
   };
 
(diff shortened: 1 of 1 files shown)


aliceriot:
works as advertised! taking a look at the code now.

Issue #3025, opened 2017-04-07
Exams: update all Pearson profiles


pdpinch:
Since Pearson staff have been tweaking some profiles, we can't be confident in the state of all our profiles in their system. To ensure everything is synchronized, we should force an update of all profiles.

rhysyngsun:
All we need to do to trigger this is set `ExamProfile.status = ExamProfile.PROFILE_PENDING` on all profiles and save via `save()` so that `updated_on` gets updated too.

rhysyngsun:
Propose we run this code:

```python
from exams.models import ExamProfile

for exam_profile in ExamProfile.objects.iterator():
    exam_profile.status = ExamProfile.PROFILE_PENDING
    exam_profile.save()
```

rhysyngsun:
Apparently I was incorrect, we *are* using `updated_on` from `Profile`, so we need to address that first.

PR #3083, opened 2017-04-19
Added timestamp fields to ExamProfile

rhysyngsun:
#3025 

#### What's this PR do?
This adds `updated_on` and `created_on` fields to `ExamProfile`. We need this because we previously sent the value of `Profile.updated_on` when syncing `ExamProfile` records and we need to be able to update that without affecting the user's `Profile`.

- Create an `ExamProfile` for your user
- Run the command `./manage.py resync_exam_profiles --ids=678,567,567`, which shouldn't match any records
- Run the command `./manage.py resync_exam_profiles --ids=?`, passing your `ExamProfile.id` which shouldn't match the one record
- Run the command `./manage.py resync_exam_profiles`, which should match all records

See exams/models.py, but bulk of changes are to exams/pearson/writers.py


annagav:
Would it make sense to put all three operations into one migration file?

rhysyngsun:
Yeah, that's a good idea.

rhysyngsun:
@annagav at most, I'd be able to get it down to 2 migrations, because of the transactions wrapping the migration I can't update the database and change the schema in 1 transaction.

rhysyngsun:
@annagav could you give this another quick look? I separated the latter two migrations to another branch to avoid the data race condition, so the only migration being applied here is to add nullable versions of `created_on` and `updated_on`. The changes I made are the inverse of this diff: https://github.com/mitodl/micromasters/compare/nl/3025-updatable-exam-profiles...nl/3025-exam-profile-timestamp-values

rhysyngsun:
@annagav sorry for all the churn on this. Last change I promise. PR is now just the first migration and the command.

annagav:
works great! 👍

PR #3088, opened 2017-04-19
add basic HTML email capability

### LINKED ISSUE #2948: Mail: allow simple HTML
Staff would like to include hyperlinks in the mail sent through the portal (in particular, through the search interface).

See [ZenDesk #6790](https://odl.zendesk.com/agent/tickets/6790)

> Is there a way to add a hyperlink to text so that I could write "MITx MicroMasters Portal" and link the homepage? If not, not a problem just wanted to check!

Related issue #1938

aliceriot:

closes #2948 

#### What's this PR do?

This adds support for basic HTML rich text to our emails. For rich-text editing we're using [react-draft-wysiwyg](https://github.com/jpuri/react-draft-wysiwyg), which is a slim wrapper around Facebook's [draft-js](https://github.com/facebook/draft-js). It's pretty slick!
 
Here are things to test:

- check that you can use HTML to compose an email, and that the HTML is displayed accurately in the email you receive
- check that a reasonable text fallback is generated. it should basically have all the text in the email, and if there are any anchor tags in the HTML it should replace those anchor tags with their `href` attributes.
- when you're composing an email you should get a validation message if you add a link to the email whose `href` attribute does not start with `http|https` (this relates to a restriction that draft-js enforces - if you try to instantiate an `EditorState` object, which is uses to manage its internal state, with an HTML string containing anchor tags which don't meet this requirement it will just turn those anchor tags into plain text - not what we want).
- if you save an HTML `AutomaticEmail` (from `/learners`) you should be able to edit the HTML of that `AutomaticEmail` on the automatic email admin page without any problems.

that's all I can think of specifically...

email_editing


rhysyngsun:
Code looks good and general functionality looks good too. I only have a few usability nits:

- The editor is rendering in our site font, which (*super super* IMO) felt a little odd and made selecting text feel tricky. 
- I think the editor could use a little horizontal padding so the cursor isn't so close to the left edge.
- Maybe some kind of "clear formatting" button would be good if draft-js provides that, I know I personally use it in my own email editor often when I get things really messed up rather than just throw all the text out.

Might be worth getting this in front of Rob as well because none of the above occurred to me until I actually started playing with it.

aliceriot:
Sure, thanks for taking a look! @roberthouse54 what do you think? I can add more screenshots if that would help, or you can take a look at the PR build (https://micromasters-ci-pr-3088.herokuapp.com/) if you want to play around with it.

For clearing the formatting, I don't think that draft-js includes anything like that, but it wouldn't be terribly difficult to implement, we'd just need to do something on the frontend more-or-less like what we do to generate the text fallback on the backend. If we want to implement that it might be better as something to take care of in a second pass, so as not to make this changeset too large.

mail/api.py
```diff
@@ -133,10 +135,18 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
             chunk_dict = {email: context for email, context in chunk}
             emails = list(chunk_dict.keys())
 
+            # parse our HTML body in order to generate a plain-text fallback
+            # the only thing we need to do manually is ensure that we keep the
+            # href for any URLs in the text
+            soup = BeautifulSoup(body, 'html5lib')
+            for link in soup.find_all('a'):
+                link.replace_with(link.attrs['href'])
```
aliceriot: I just realized there's no reason to recalculate this every time we loop through the recipients, so I'm going to move this outside of the loop.

rhysyngsun:
The demo for `react-draft-wysiwyg` you're using has a button that strips the formatting: https://jpuri.github.io/react-draft-wysiwyg/#/

aliceriot:
Oh so it does! haha. I'll add it in.

mail/api.py
```diff
@@ -136,7 +145,8 @@ def send_batch(cls, subject, body, recipients,  # pylint: disable=too-many-argum
             params = {
                 'to': emails,
                 'subject': subject,
-                'text': body,
+                'html': body,
+                'text': soup.get_text().strip(),
```
rhysyngsun: Might also be worth moving generation of the `text` value outside the loop as well so we only incur that cost once too. I think we parameterize the templates using Mailgun vars so that should be ok.

aliceriot: sure, I'll move that too

roberthouse54:
I think we should add "font style" (ie. Normal, H1, H2, etc), Font-size, and font name. We can remove strikethrough, and the two indentation buttons. Also, add padding to the text box itself.

aliceriot:
here's what it looks like with those changes:

fewer_buttons

I just added `10px` to the left of the edit box. I think the button row looks a little busy though, here's what it looks like without the border around it:

no-border

@roberthouse54 what do you think?

roberthouse54:
This looks good, except that the buttons should not have the link color. Can you give them the same styling as this: https://jpuri.github.io/react-draft-wysiwyg/#/

aliceriot:
yeah, I'm not sure why they're that color, I'll look into it!

aliceriot:
updated:

updated_dialog

rhysyngsun:
Also one of the Travis tests failed.

aliceriot:
Yeah, I'm not sure what's up with the `font-family` and `font-size` options, it's being set correctly  in the HTML (obviously) but for some reason it's stripped out by the HTML export function. I suspect I didn't read the docs closely enough...

pdpinch:
Fonts are nice, but not super critical. If it's easier to drop them from the WYSIWYG tool than to figure out how to preserve the formatting, I'd be fine with that.

aliceriot:
Yeah, I've dug into it a bit and it's definitely going to be easier to just axe font-size and font-family. There will still be h1-h6 for formatting different size headers and so on.

aliceriot:
If we care a lot about fonts we could come back to add that in later, but doing so will (as far as I can tell) probably involve either forking or submitting a PR to the library I'm using here for converting the draft-js `ContentState` to raw HTML (the library is [this one](https://github.com/sstur/draft-js-export-html)).

PR #3092, opened 2017-04-20
Pin selenium container to a non-broken version

rhysyngsun:
N/A, blocking deployments because failing Travis.

#### What's this PR do?
Fixes selenium tests that are failing Travis tests because new tags for the container got pushed.

If Travis passes, we're good.

:notes: Cause if you want it to pass you better put a pin on it. :notes:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,7 +118,7 @@ services:
       - grid
 
   grid:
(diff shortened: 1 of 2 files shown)


giocalitri:
👍

Issue #3060, opened 2017-04-12
Loader for learner page


noisecapella:
We should add the `Loader` with the spinner to the learner search page, similar to how we currently have it for `DashboardPage`

pdpinch:
Is this actually in progress?

PR #3101, opened 2017-04-24
Added loader to learners search page

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3060

#### What's this PR do?
- Added a spinner on the learners search table.

- go to search table and try to refresh.

@pdpinch

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -20,6 +20,7 @@ import iso3166 from 'iso-3166-2';
 import R from 'ramda';
 import _ from 'lodash';
 
+import Loader from './Loader';
 import ProgramFilter from './ProgramFilter';
 import LearnerResult from './search/LearnerResult';
 import CountryRefinementOption from './search/CountryRefinementOption';
@@ -132,6 +133,16 @@ export default class LearnerSearch extends SearchkitComponent {
       {openLearnerEmailComposer: this.props.openLearnerEmailComposer},
       LearnerResult
     );
+    this.forceSearch = true;
+  }
+
+  loaded = (): boolean => {
+    if (this.forceSearch) {
+      this.forceSearch = false;
+      return true;
+    } else {
+      return !this.isInitialLoading();
+    }
   }
 
   getNumberOfCoursesInProgram = (): number => {
@@ -329,27 +340,29 @@ export default class LearnerSearch extends SearchkitComponent {
     const { currentProgramEnrollment } = this.props;
 
     return (
-      <div className="learners-search">
-        <ProgramFilter
-          currentProgramEnrollment={currentProgramEnrollment}
-        />
-        <Grid className="search-grid">
-          <Cell col={3} className="search-sidebar">
-            { this.renderFacets(currentProgramEnrollment) }
-          </Cell>
-          <Cell col={9}>
-            <Card className="fullwidth results-padding" shadow={1}>
-              { this.renderSearchHeader() }
-              <Hits
-                className="learner-results"
-                hitsPerPage={SETTINGS.es_page_size}
-                itemComponent={this.WrappedLearnerResult}
-              />
-              <CustomNoHits />
-            </Card>
-          </Cell>
-        </Grid>
-      </div>
+      <Loader loaded={this.loaded()}>
+        <div className="learners-search">
+          <ProgramFilter
+            currentProgramEnrollment={currentProgramEnrollment}
+          />
+          <Grid className="search-grid">
+            <Cell col={3} className="search-sidebar">
+              { this.renderFacets(currentProgramEnrollment) }
+            </Cell>
+            <Cell col={9}>
+              <Card className="fullwidth results-padding" shadow={1}>
+                { this.renderSearchHeader() }
+                <Hits
+                  className="learner-results"
+                  hitsPerPage={SETTINGS.es_page_size}
+                  itemComponent={this.WrappedLearnerResult}
+                />
+                <CustomNoHits />
+              </Card>
+            </Cell>
+          </Grid>
+        </div>
+      </Loader>
     );
   }
 }
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -133,16 +133,6 @@ export default class LearnerSearch extends SearchkitComponent {
       {openLearnerEmailComposer: this.props.openLearnerEmailComposer},
       LearnerResult
     );
-    this.forceSearch = true;
-  }
-
-  loaded = (): boolean => {
-    if (this.forceSearch) {
-      this.forceSearch = false;
-      return true;
-    } else {
-      return !this.isInitialLoading();
-    }
   }
 
   getNumberOfCoursesInProgram = (): number => {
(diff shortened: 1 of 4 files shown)


aliceriot:
functionality looks good! looking at code now.

static/js/containers/LearnerSearchPage.js
```diff
@@ -41,21 +49,32 @@ class LearnerSearchPage extends React.Component {
...
+    const {
+      currentProgramEnrollment,
+      openEmailComposer,
+      programs,
+      profiles
+    } = this.props;
+    const loaded = R.none(isProcessing, [
+      R.pathOr(null, [SETTINGS.user.username, 'getStatus'], profiles),
+      R.pathOr(null, ['getStatus'], programs)
+    ]);
```
aliceriot: why are we checking for `profiles` and `programs` here? it seems like that's probably not the data we're waiting on here - maybe instead we should show the spinner until we first get a search result, for example.

static/js/components/LearnerSearch.js
```diff
@@ -128,6 +129,16 @@ export default class LearnerSearch extends SearchkitComponent {
       {openLearnerEmailComposer: this.props.openLearnerEmailComposer},
       LearnerResult
     );
+    this.forceSearch = true;
+  }
+
+  loaded = (): boolean => {
+    if (this.forceSearch) {
```
amir-qayyum-khan: I notice that when the components first render then the hit search url, 
Without rendering for first time search wont initiate.

isInitialLoading() method from searchkit get update when search is complete for first time

amir-qayyum-khan:
done @aliceriot

aliceriot:
sorry for not getting to this sooner, it just slipped by me. Taking a look now!

aliceriot:
functionality looks great! reading the code now.

static/js/components/LearnerSearch.js
aliceriot: I'm not sure I exactly understand what's going on here - it looks like we'll only have `this.forceSearch === true` on the first render, after which we'll set it to false (on the line below).

Why is it that we can't just call `this.isInitialLoading`, and why, in the case where we're doing our first render, we would want to return `true` here automatically (on line 138)?

aliceriot:
It looks like you have a few failing JavaScript tests as well

static/js/components/LearnerSearch.js
amir-qayyum-khan: because the component below initiate the search it self, if we dont render them search query wont call

aliceriot: ok, in that case I have two suggestions - the first is that the `forceSearch` variable be re-named to something more descriptive (perhaps `renderOnce`?), and the second is that the `loaded` method be renamed or separated from the behavior that sets the boolean value on `this`. `loaded` as a function name implies purity to me (just returning a value) and as it is written it has a side effect (it mutates the state that lives on the component instance, by setting `this.forceSearch`). 

it might be good also to leave a comment that explains what is going on and why the method is necessary, just so it's clear to others reading the code later.

aliceriot: seems like we could also just inline a call like

```js
<Loader loaded={this.forceSearch || !this.isInitialLoading}>
```

in the JSX, and the method could just do

```js
trackRenderedOnce = () => {
  this.forceSearch = false;
}
```

I think you could then just have `trackRenderedOnce` be called in `componentDidMount`, but I'm not sure we have a guarantee that `componentDidMount` is called after exactly one call to `render`, @noisecapella do you know if that would work?

amir-qayyum-khan:
Done @aliceriot

aliceriot:
@amir-qayyum-khan it looks like some tests are still failing here

amir-qayyum-khan:
yes selenium test is failing

amir-qayyum-khan:
@giocalitri @noisecapella @aliceriot the urls is persevered in browser but it is not working in selenium test, it seem it is redirecting , how can i debug this?

noisecapella:
@amir-qayyum-khan If you clear `localStorage` first by entering `localStorage.redux = null` in the console, then refresh the screen, the query string is removed from the URL. So the test is correct here. I'll look at your PR and see if there's anything which would obviously cause a problem

selenium_tests/basic_test.py
```diff
@@ -143,7 +143,7 @@ def test_query_string_preserved(self):
         """The querystring should not be affected"""
         self.learner_setup()
         self.get("{}/learners/?q=xyz".format(self.live_server_url))
-        self.wait().until(lambda driver: self.num_elements_on_page('.learner-result', driver=driver) == 0)
+        self.wait().until(lambda driver: driver.find_element_by_class_name('learner-result'))
```
noisecapella: We shouldn't change this. It asserts that there are no `learner-result` divs on the page, which would be true if the filter for `xyz` was still present on the page

noisecapella:
One thing you can consider is changing `Loader` so that `children` is always rendered and use `display` to make it visible or not. I made a similar change on my local branch and it seemed to fix the problem, and you no longer need `forceSearch` either

amir-qayyum-khan:
Done @noisecapella

static/js/components/LearnerSearch.js
```diff
@@ -330,25 +331,30 @@ export default class LearnerSearch extends SearchkitComponent {
...
-              <Hits
-                className="learner-results"
-                hitsPerPage={SETTINGS.es_page_size}
-                itemComponent={this.WrappedLearnerResult}
-              />
-              <CustomNoHits />
-            </Card>
-          </Cell>
-        </Grid>
+        <div className="loader" style={{'display': this.isInitialLoading() ? "block" : "none"}}>
```
noisecapella: Can you make this change in the `Loader` component instead of inlining it here?

amir-qayyum-khan:
Done @noisecapella

static/js/components/Loader.js
```diff
@@ -6,22 +6,28 @@ export default class Loader extends React.Component {
...
+  );
 
-    return <div>
-      {content}
-    </div>;
+  render() {
+    const { loaded, children, shouldRenderAll = false } = this.props;
+    return (
+      <div>
+        { shouldRenderAll ? this.renderAll(loaded, children) : this.renderLoadedOrSpinner(loaded, children) }
```
noisecapella: Why is `shouldRenderAll` necessary here? Is there a change of behavior if we use `renderAll` for all cases?

amir-qayyum-khan: yes, more the 15 tests fail in system

amir-qayyum-khan:
@aliceriot @noisecapella

aliceriot:
functionality is looking great! reading the code now

static/js/components/Loader.js
aliceriot: I'm guessing any tests where we assert that the children aren't on the DOM will fail, because they're still being rendered (albeit invisibly)

static/js/components/Loader.js
```diff
@@ -6,22 +6,28 @@ export default class Loader extends React.Component {
...
+      </div>
+    );
```
aliceriot: could you add some tests for the changes you've made to `<Loader />` ?

amir-qayyum-khan:
Done @aliceriot

Issue #2947, opened 2017-03-24
404 page as a logged in learner viewing a public_to_mm profile


rhysyngsun:
- Log in as a learner (no staff permissions)
- View another learner's profile that has `account_privacy` set to `Profile.PUBLIC_TO_MM`

Expected result: user can see the profile
Actual result: 404 page

amir-qayyum-khan:
@pdpinch 
This is because your user is not a `Verified micromaster user` and we dont show `PUBLIC_TO_MM` profiles to users who are not verified. please see https://github.com/mitodl/micromasters/blob/master/profiles/permissions.py#L64

#### And setting i am talking about is
<img width="622" alt="screen shot 2017-04-24 at 5 33 59 pm" src="https://cloud.githubusercontent.com/assets/10431250/25337242/8a2e2146-2914-11e7-8e69-723e86429d67.png">

amir-qayyum-khan:
fixed by #3102

PR #3102, opened 2017-04-24
Shown public_to_mm profiles when requesting user is enrolled in one of the programs where profile user is enrolled

amir-qayyum-khan:
https://github.com/mitodl/micromasters/issues/2947#issuecomment-296650753

#### What's this PR do?
- Now non staff users can view  public_to_mm profiles, when they are enrolled in one of the programs where profile user is enrolled.
- Staff can already view profiles.

- Create a learner and try to view profile of other learner in same program.
- Create a learner and try to view profile of other learner in other program.

@pdpinch

### DIFF
--- a/profiles/permissions_test.py
+++ b/profiles/permissions_test.py
@@ -204,19 +204,50 @@ class CanSeeIfNotPrivateTests(MockedESTestCase):
         perm = CanSeeIfNotPrivate()
         with mute_signals(post_save):
             ProfileFactory.create(user=self.profile_user, account_privacy=Profile.PUBLIC_TO_MM)
+            program = ProgramFactory.create()
+            ProgramEnrollment.objects.create(
+                program=program,
+                user=self.profile_user,
+            )
 
         with mute_signals(post_save):
-            verified_user = UserFactory.create()
-            username = "{}_edx".format(verified_user.username)
-            verified_user.social_auth.create(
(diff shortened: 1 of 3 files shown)


profiles/permissions_test.py
```diff
@@ -204,20 +204,51 @@ def test_can_view_if_verified_mm_user(self):
...
+                user=self.profile_user,
+            )
 
         with mute_signals(post_save):
-            verified_user = UserFactory.create()
-            username = "{}_edx".format(verified_user.username)
-            verified_user.social_auth.create(
+            user = UserFactory.create()
+            username = "{}_edx".format(user.username)
+            user.social_auth.create(
```
rhysyngsun: I'd suggest writing a factory for the social auth creation here. I see very similar code repeated else where and think it's worth writing that now so we can eliminate a bunch of that boilerplate. That factory would probably live best in the `micromasters` module.

rhysyngsun:
Functionality looks good.

profiles/permissions_test.py
amir-qayyum-khan: This is out of scope, i think we should handle this in a separate task.
please create an issue, i am happy to take that task

rhysyngsun: Ok, fair enough

Issue #2792, opened 2017-03-06
Dashboard: Exam grades and final grades UI


pdpinch:
The goal of this ticket is to a) add a section for "Grades" to each course that is in progress, and to show a "Passed" status at the course level, whenever the course has been passed.

**Details**

![image](https://cloud.githubusercontent.com/assets/20047260/25346695/562fd3cc-28e6-11e7-9e47-f2c58ce79222.png)

- Once a course is (or has been) enrolled, it is considered to be in progress until it is complete and "Passed." For In Progress or passed courses, grades are displayed directly beneath the course name. For programs that have proctored exams at the course level, we will display 3 grades: a) edX Course, b) Exam, c) Final Grade. For courses that do not require an exam, we will only display one grade.
- For edX grade, and exam grade, display the highest completed grade (if the course or exam has been taken multiple times). This includes "non passing" grades.
- The Final Grade is a weighted average of the edX grade, and the Exam grade. Weighting to be configured on a per-program basis.
- Clicking on grades will link to a "details view" that shows All Completed grades, but that is not included in this ticket. That will be done in #3049 . For this ticket, grades are not links.
- Clicking a "?" icon will display a tooltip help text, that will explain how the grading works (copy TBD).
![image](https://cloud.githubusercontent.com/assets/20047260/25346724/6e40bcf6-28e6-11e7-899a-0f11a3f2ca54.png)

Tooltip for edX Course: This is your grade in the edX course. If you complete the course more than once, it is your best grade that counts toward the MicroMasters.

Tooltip for Exam: This is your grade in the Final Exam. If you complete the exam more than once, it is your best grade that counts toward the MicroMasters.

Tooltip for Final Grade: Your Final Grade is a weighted average of your highest edX Course grade (40% weighted) and your highest Exam grade (60% weighted). 

- Once the course has been passed (at the course level), we will display a Check icon, and the word passed.
![image](https://cloud.githubusercontent.com/assets/20047260/25346753/8750a44a-28e6-11e7-8d65-89370cd5a3d4.png)

### acceptance criteria:

- [x] display exam grade
- [x] display final grade, which is a combination of exam grade and online course grade
- [ ] use feature flag so we can test this before turning it on

pdpinch:
@roberthouse54 can you add a screenshot when you think its ready?

PR #3109, opened 2017-04-25
implement simple exam grade display

aliceriot:

a temporary shim for part of #2792

#### What's this PR do?

this just implements a super basic display of the highest exam grade (per-course) on the dashboard.

Add an exam grade to a course you're enrolled in, and you should see it on the dashboard. Add two exams, and you should see the highest grade.

weewewewewe

### DIFF
--- a/static/js/components/dashboard/CourseGrade.js
+++ b/static/js/components/dashboard/CourseGrade.js
@@ -2,14 +2,42 @@
 // @flow
 import React from 'react';
 import _ from 'lodash';
+import R from 'ramda';
 
-import type { CourseRun } from '../../flow/programTypes';
+import type { CourseRun, Course } from '../../flow/programTypes';
 import { formatGrade } from './util';
 import { EDX_LINK_BASE } from '../../constants';
+import { S, getm } from '../../lib/sanctuary';
+
+const findLargestGrade = R.compose(
+  R.prop('percentage_grade'),
+  R.reduce(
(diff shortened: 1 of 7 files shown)


pdpinch:
@aliceriot I'm still wondering if we can display the exam grade "above" all the course run information -- to more closely associate it with the course instead of the course run. 

Barring that, it seems like it should at least go above/before the course grade.

aliceriot:
ok, switching the order would be pretty trivial, putting it above everything else less so. @roberthouse54 what do you think?

aliceriot:
here's what it looks like with the order flipped:

order_changed

if we think this is good for now I'd love to merge this so I can get started on the full set of changes to the dashboard.

pdpinch:
Please don't merge this until DEDP has signed off on the final exam grades.

rhysyngsun:
Code looks good and functionality generally works, but I don't see the flipped order of exam / final grades.

aliceriot:
oh yeah I don't think I pushed that commit yet. I think it's looking like we wont need to merge this PR anyhow, I've been working on the full dashboard redesign and that may be done before we get the go-ahead from DEDP I think.

aliceriot:
@rhysyngsun the updated version was just pushed, would you mind taking a look over this again? I think we wanted to go ahead with merging it.

rhysyngsun:
:+1: assuming tests pass

Issue #3085, opened 2017-04-19
Exams: add ExamRun model to track when exams occur


rhysyngsun:
Adding this model should solve a few problems:

- We currently re-create `ExamAuthorization`s ad infinitum even if there are not actually any exams scheduled. This makes deleting the pilot period ones for #3027 difficult/impossible.
- We scope `ExamSeriesCode` to program/course currently. That structure needs to change for #3077 anyway and these values could, in theory, be scoped only to a single `ExamRun` so we should just put in on this record to make our lives easier anyway.
- This causes us to have to do odd things like set `FEATURE_EXAMS_CARD_VISIBLE=False` to hide the card when there aren't any exams to take (this should obviate the need for this flag entirely so we could then do #3002)
- We "unique" the authorizations on user + course + a rough bound of eligibility dates. If we added a foreign key to this new table we could do away with that in favor of a real unique constraint.

So I propose we do the following (each phase represent a deployment so as to avoid errors in prod):

### ~~Phase 1 - #3112~~
- [x] Add `ExamRun` model with these fields: `date_first_eligible`, `date_last_eligible`, `exam_series_code`
- [x] Admin page for `ExamRun`
- [x] Add nullable foreign key to `ExamRun` from `ExamAuthorization`
- [x] Make `date_first_eligible` and `date_last_eligible` on `ExamAuthorization` nullable
- [x] Update [this code](https://github.com/mitodl/micromasters/blob/master/exams/utils.py#L106-L126) to `get_or_create` on `user` + `course` + `exam_run` and only for the first `ExamRun` in the future for that course.

### ~~Phase 2 - #3122~~

- [x] Create a dummy `ExamRun` record for each course that was in the pilot and assign the pilot authorizations to these. We should give them an `ExamSeriesCode` of `"PILOT"` or similar so they're clearly marked. (Addresses #3027)
- [x] Change `ExamAuthorization.exam_run` to be non-nullable
- [x] Update the `ExamAuthorization` writer to use the fields from `ExamRun`

### ~~Phase 3 - #3123~~
- [x] Remove these fields from `ExamAuthorization`:  `date_first_eligible`, `date_last_eligible`
- [x] Remove these fields from `Course`:  `exam_module`
- [x] Remove these fields from `Program`:  `exam_series_code`

rhysyngsun:
[...]

PR #3112, opened 2017-04-25
ExamRuns: model and authorization logic

rhysyngsun:
Part of #3085 

#### What's this PR do?
Adds the ExamRun model and adjusts existing exam logic accordingly.

The new model has timestamps for when scheduling starts and ends as well as being the new home for the authorization eligibility dates because we are going to now only authorize for dates ranging over the actual run of a set of exams (e.g. May exams) rather than just give a very broad 1 year eligibility.

There was some cleanup done here of older code used during the pilot, sorry for the big diff but it was complicating test cases quite a bit. The bulk of these changes removed the command `authorization_user_command` and reimplemented similar logic in a cron task because we needed to automatically generate authorizations on the first day of exam scheduling.

##### Prereq:
- Set `FEATURE_EXAMS_CARD_ENABLED=True` set in `.env`

###### No Exams
- With a new user, enroll in that program
  - check the dashboard, you should not see the exam card.

###### Current Exam
- In django admin, create a new `ExamRun` for a finanical aid course with `date_first_schedulable` in the past (or now) and `date_last_schedulable` in the future
  - Refresh the dashboard, ensure you see the exam card now, telling you the course has exams and w'ell let you know when ready
  - Run the task `exams.tasks.authorize_exam_runs()`
  - Verify no `ExamAuthorization` or `ExamProfile` records exist for your user (note: if the hourly cron happens to run, you will see these appear for test data users, that is expected)
- Enroll in the course you created the run for.
  - Run the task `exams.tasks.authorize_exam_runs()`
  - Verify an `ExamProfile` got created, but not an `ExamAuthorization`
- Create a passing `FinalGrade` record for your user
  - Run the task `exams.tasks.authorize_exam_runs()`
  - Verify you now have a `ExamAuthorization` record.

###### Noncurrent Exams
- In django admin, do the following (combining these to save time):
  - Delete your `ExamRun` from the previous scenario
  - Create an `ExamRun` for a finanical aid course with `date_first_schedulable` in the future and `date_last_schedulable` in the future
  - Create an `ExamRun` for a finanical aid course with `date_first_schedulable` in the past and `date_last_schedulable` in the past
  - Refresh the dashboard, ensure you see the exam card now, telling you the course has exams and w'ell let you know when ready
  - Run the task `exams.tasks.authorize_exam_runs()`
  - Verify no `ExamAuthorization` or `ExamProfile` record exist for your user (note: if the hourly cron happens to run, you will see these appear for test data users, that is expected)
- Enroll in the course you created the run for.
  - Run the task `exams.tasks.authorize_exam_runs()`
  - Verify no `ExamAuthorization` or `ExamProfile` records exist for your user
- Create a passing `FinalGrade` record for your user
  - Run the task `exams.tasks.authorize_exam_runs()`
  - Verify no `ExamAuthorization` or `ExamProfile` records exist for your user

exams/api.py, exams/models.py

Exam Run


noisecapella:
I updated the `Set FEATURE_EXAM_CARD_ENABLE=True set in .env` text to `FEATURE_EXAMS_CARD_ENABLED` FYI

courses/models.py
```diff
@@ -273,12 +272,7 @@ def has_exam(self):
         """
         Check if the user is authorized for an exam for a course run
         """
-        return (
-            self.edx_course_key and
-            self.course and
-            not is_blank(self.course.exam_module) and
-            not is_blank(self.course.program.exam_series_code)
-        )
+        return self.course.exam_runs.filter(date_last_eligible__gt=datetime.now(pytz.utc)).exists()
```
rhysyngsun: `exam_series_code` isn't nullible for the `ExamRun`, so that shouldn't be an issue. Not sure what you been by the coruse key, because we don't handle that with exams.

exams/api.py
```diff
@@ -32,3 +57,115 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+            user=mmtrack.user.username,
+            course_id=course_run.edx_course_key
+        )
+        raise ExamAuthorizationException(errors_message)
+
+    ExamAuthorization.objects.get_or_create(
+        user=mmtrack.user,
+        course=course_run.course,
+        exam_run=exam_run,
+        defaults={  # write these out even though they're deprecated as putting nulls in here forfeits a db rollback
```
noisecapella: What's the problem with nulls there?

exams/api.py
```diff
@@ -32,3 +57,115 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+    log.info(
+        '[Exam authorization] user "%s" is authorized for the exam for course id "%s"',
+        mmtrack.user.username,
+        course_run.edx_course_key
+    )
+
+
+def authorize_for_latest_passed_course(mmtrack, exam_run):
+    """
+    Authorize user for exam on given course. Using latest passed run.
```
noisecapella: This says 'latest passed run' but the queryset looks like it just gets the latest run. Is it assumed to have been passed if the `CachedEnrollment` exists?

exams/api.py
```diff
@@ -32,3 +57,115 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+    ).order_by('-course_run__end_date')
+
+    if not enrollments_qset.exists():
+        return
+
+    for enrollment in enrollments_qset:
+        try:
+            authorize_for_exam_run(mmtrack, enrollment.course_run, exam_run)
+        except ExamAuthorizationException:
+            log.debug(
```
noisecapella: Why not `log.exception` here? Is this a common case?

exams/api.py
```diff
@@ -32,3 +57,115 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
         str(client_candidate_id),
     ]).encode('iso-8859-1')
     return hashlib.sha256(data).hexdigest()
+
+
+def authorize_for_exam_run(mmtrack, course_run, exam_run):
```
noisecapella: Does `exam_run.course` need to be the same as `course_run.course`? If so, can we just obtain `exam_run` from `course_run`?

exams/factories.py
```diff
@@ -34,6 +35,48 @@ class Meta:
...
+    class Meta:
+        model = ExamRun
+
+    class Params:
+        scheduling_past = factory.Trait(
+            date_first_schedulable=factory.LazyFunction(
+                lambda: FAKE.date_time_this_year(before_now=True, after_now=False, tzinfo=pytz.utc)
+            ),
+            date_last_schedulable=factory.LazyFunction(
+                lambda: FAKE.date_time_this_year(before_now=True, after_now=False, tzinfo=pytz.utc)
```
noisecapella: It seems like there's no guarantee that `date_last_schedulable` comes after `date_first_schedulable`. Can you change this to use a `LazyAttribute` similar to below?

exams/models.py
```diff
@@ -52,6 +114,18 @@ class ExamAuthorization(TimestampedModel):
 
     Exam Authorization record are sent to Pearson authorizing an individual for a specific exam.
     This model exists so we can track state on our side and corresponds to a row in a Pearson EAD file.
+
+    ExamProfile.status is a state machine following this flow:
```
noisecapella: Should this be `ExamAuthorization.status`?

exams/tasks.py
```diff
@@ -188,3 +193,31 @@ def batch_process_pearson_zip_files():
         log.exception('PEARSON_EXAMS_SYNC enabled, but not configured correctly')
     except RetryableSFTPException:
         log.exception('Retryable error during SFTP operation')
+
+
+@async.task
+def update_exam_run(exam_run_id):
```
noisecapella: Can you add this argument to the docstring?

exams/api.py
rhysyngsun: These two columns were previously not-null, this PR makes them nullable, however, were we to decide to roll back the release, the db rollback would fail if we ever inserted a record with null in these columns, this code goes away in the second PR and the column goes away entirely in the third PR, soI don't think it's worth a lot of thinking about.

exams/api.py
rhysyngsun: Yes, we could hit this a lot if they failed the course at least once recently. Honestly, I'd just have this be a `pass`, but a debug statement is probably useful if we need it.

exams/api.py
rhysyngsun: This walks the `CachedEnrollments` backwards chronologically and bails as soon as it finds one it was able to successfully authorize for. If you have any suggestions for how to clarify that, let me know.

exams/api.py
rhysyngsun: There could be multiple future `ExamRun`s per course, so we want to explicitly pass it here.

courses/models.py
rhysyngsun: Generally, learners need to register 24 hours before the last exam, so `gt` is right here I think. It's somewhat rough timing anyway. @pdpinch does that make sense?

exams/api.py
```diff
@@ -32,3 +57,115 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+
+    # If user has not paid for course then we dont need to process authorization
+    if not mmtrack.has_paid(course_run.edx_course_key):
+        errors_message = MESSAGE_NOT_ELIGIBLE_TEMPLATE.format(
+            user=mmtrack.user.username,
+            course_id=course_run.edx_course_key
+        )
+        raise ExamAuthorizationException(errors_message)
+
+    # if user paid for a course then create his exam profile if it is not creaated yet
```
noisecapella: `creaated` should be `created`

exams/api.py
noisecapella: Can you add a check that `exam_run.course == course_run.course`?

exams/api.py
noisecapella: Sounds good

exams/api.py
noisecapella: Can you add this description to the docstring?

exams/factories.py
```diff
@@ -34,6 +35,48 @@ class Meta:
...
+    class Meta:
+        model = ExamRun
+
+    class Params:
+        scheduling_past = factory.Trait(
+            date_first_schedulable=factory.LazyFunction(
+                lambda: FAKE.date_time_this_year(before_now=True, after_now=False, tzinfo=pytz.utc)
+            ),
+            date_last_schedulable=factory.LazyAttribute(
+                lambda exam_run: exam_run.date_first_schedulable + timedelta(days=10)
```
noisecapella: It might be safer to have this use `date_time_this_year` and have `date_first_schedulable` be that date minus 10 days

exams/factories.py
```diff
@@ -34,6 +35,48 @@ class Meta:
...
+    date_first_schedulable = factory.LazyFunction(
+        lambda: FAKE.date_time_this_year(before_now=True, after_now=False, tzinfo=pytz.utc)
+    )
+    date_last_schedulable = factory.LazyFunction(
+        lambda: FAKE.date_time_this_year(before_now=False, after_now=True, tzinfo=pytz.utc)
+    )
+    date_first_eligible = factory.LazyFunction(
+        lambda: FAKE.date_time_this_year(before_now=False, after_now=True, tzinfo=pytz.utc).date()
+    )
+    date_last_eligible = factory.LazyAttribute(
```
noisecapella: Does it matter if this is sometimes in the past and sometimes in the future?

noisecapella:
Functionality looks good, looking at tests now

exams/factories.py
rhysyngsun: I'm not quite sure I follow how that could happen since it's based on `date_first_eligible` which is only `after_now=True`, that should be ok, correct?

exams/factories.py
rhysyngsun: Good catch, I think Travis just caught this by chance on one of the other PRs.

exams/factories.py
noisecapella: You're right, I thought it was in the past

exams/tasks.py
```diff
@@ -188,3 +193,34 @@ def batch_process_pearson_zip_files():
...
+    try:
+        exam_run = ExamRun.objects.get(id=exam_run_id)
+    except ExamRun.DoesNotExist:
+        return
+
+    api.update_authorizations_for_exam_run(exam_run)
+
+
+@async.task
+def authorize_exam_runs():
```
noisecapella: Can you add a test for this function?

exams/api_test.py
```diff
@@ -38,3 +75,184 @@ def test_that_no_client_code_raises(self):
         """Verifies that if we don't set the passphrase we raise an exception"""
         with self.assertRaises(ImproperlyConfigured):
             sso_digest(123, 1486069731, 1800)
+
+
+class ExamAuthorizationApiTests(TestCase):
```
noisecapella: Should we use `MockedESTestCase` here?

exams/api.py
noisecapella: Nevermind, looks like you already have one

exams/api.py
```diff
@@ -32,3 +57,119 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+    """
+    Authorize user for exam if he has paid for course and passed course.
+
+    Args:
+        mmtrack (dashboard.utils.MMTrack): a instance of all user information about a program.
+        course_run (courses.models.CourseRun): A CourseRun object.
+        exam_run (exams.models.ExamRun): the ExamRun we're authorizing for
+    """
+    if course_run.course != exam_run.course:
+        raise ExamAuthorizationException(
```
noisecapella: Can you add a test for this case?

exams/api.py
rhysyngsun: Added it between your reviews ;-)

exams/api.py
```diff
@@ -32,3 +57,119 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+    for program_enrollment in ProgramEnrollment.objects.filter(
+            program=exam_run.course.program
+    ).iterator():
+        mmtrack = get_mmtrack(
+            program_enrollment.user,
+            program_enrollment.program
+        )
+
+        try:
+            authorize_for_latest_passed_course(mmtrack, exam_run)
```
noisecapella: Can you add a test which asserts that if the first call fails, the second call is still attempted?

exams/api.py
```diff
@@ -32,3 +57,119 @@ def sso_digest(client_candidate_id, timestamp, session_timeout):
...
+        except ExamAuthorizationException:
+            log.debug('User "%s" not authorized for exam', program_enrollment.user)
+
+
+def update_authorizations_for_exam_run(exam_run):
```
noisecapella: Can you add a test for this function? We should also check that it doesn't affect other exam runs and that it doesn't affect ExamRuns with `exam_taken=True`

exams/api.py
rhysyngsun: Did you mean to leave this comment in `authorize_for_latest_passed_course` above?

rhysyngsun: Addressed this assuming you meant the other method.

noisecapella: I think we should test both methods since they both have a similar loop with try block pattern

rhysyngsun: Actually, the `try` block here is superfluous. `authorize_for_latest_passed_course` already handles the exception. I'll still add a test.

rhysyngsun:
@noisecapella I've addressed all your feedback

exams/api_test.py
```diff
@@ -38,3 +80,280 @@ def test_that_no_client_code_raises(self):
...
+    def test_update_authorizations_for_exam_run(self):
+        """Tests update_authorizations_for_exam_run()"""
+        exam_run = ExamRunFactory.create()
+        taken_auth = ExamAuthorizationFactory.create(
+            exam_run=exam_run,
+            exam_taken=True,
+            status=ExamAuthorization.STATUS_SUCCESS,
+            operation=ExamAuthorization.OPERATION_ADD,
+        )
+        print(taken_auth)
```
noisecapella: Stray print statement

rhysyngsun: :scream:

Issue #2949, opened 2017-03-24
Mail: Support Mailgun macros in e-mail template


pdpinch:
Part of #2812

When staff users compose an email in the search interface, they would like to be able to use macros for things like the user's preferred name and (possibly) the name of the program. 

### Acceptance criteria
- [ ] Use basic macro tokens added for the [welcome email](https://github.com/mitodl/micromasters/issues/2241#issuecomment-294159020)
   - `%recipient.mail_id%` is the UUID
   - `%recipient.email%` is the user's email
   - `%recipient.preferred_name%` is the user's preferred name)
- [ ] Add macro hints to the mail UI
- [ ] (nice to have) clicking the macro hint in the UI adds it to the message body at the cursor location

pdpinch:
This should be addressed by the WYSIWYG editor implemented to address #2948

pdpinch:
@annagav I updated the description with the tokens that [George added earlier](https://github.com/mitodl/micromasters/issues/2241#issuecomment-294159020): 
   - `%recipient.mail_id%` is the UUID
   - `%recipient.email%` is the user's email
   - `%recipient.preferred_name%` is the user's preferred name)

PR #3115, opened 2017-04-26
Mail: Filter recipient variables

annagav:
Related to #2949

#### What's this PR do?
Convert users inserted variables like `[PreferredName]` to mailgun recipient variables `%recipient.preferred_name%`

Send an email to your user (with your email address). Use [PreferredName] and [Email] in the header or body of the email. Make sure your variable `MAILGUN_RECIPIENT_OVERRIDE` is not set.

### DIFF
--- a/mail/utils.py
+++ b/mail/utils.py
@@ -21,6 +21,11 @@ from financialaid.constants import (
 
 log = logging.getLogger(__name__)
 
+RECIPIENT_VARIABLE_NAMES = {
+    'PreferredName': 'preferred_name',
+    'Email': 'email',
+}
+
 
 def generate_financial_aid_email(financial_aid):
     """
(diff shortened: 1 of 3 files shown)


mail/utils.py
```diff
@@ -85,3 +85,21 @@ def generate_mailgun_response_json(response):
...
+
+def filter_recipient_variables(text):
+    """
+    Filter out recipient variables, like [PreferredName], and substitute it with %recipient.preferred_name%
+    Args:
+        text (string): subject or body of the email
+    Returns:
+        string: with replaced correct recipient variables
+    """
+    substitute_values = {
```
rhysyngsun: This could be a module-level constant.

mail/utils.py
```diff
@@ -85,3 +85,21 @@ def generate_mailgun_response_json(response):
             "message": response.reason
         }
     return response_json
+
+
+def filter_recipient_variables(text):
```
rhysyngsun: Could you write a test or two for this, I think it's technically covered by other tests, but I don't think the expected behavior is being asserted.

rhysyngsun:
Functionality is good.

mail/utils.py
annagav: I added a test

Issue #2474, opened 2017-02-01
Jumbled financial aid dialog on mobile device


noisecapella:
![screenshot_20170201-120421](https://cloud.githubusercontent.com/assets/863262/22517812/6613cc5c-e878-11e6-9520-95f1c90f49d7.png)

roberthouse54:
A few points on this layout for mobile:

- Decrease the passing in the modal
- Decrease the font size and line height of the Checkbox text. Also, remove the left/right padding of the checkbox element.
- The income field, and the currency input field should both be full width, one on top of the other
- Skip this and pay full price should be on a separate line ABOVE the buttons.
- If the content of the modal is too much to fit vertically, the modal should scroll.

pdpinch:
Similar things can happen in the viewport is very short on a desktop browser.

noisecapella:
If I enter text in the income text field and then clear it, the dialog moves such that I'm able to proceed:

![screenshot_20170317-144608](https://cloud.githubusercontent.com/assets/863262/24058424/6c68f5f2-0b21-11e7-89ec-def091c5bf9e.png)
![screenshot_20170317-144616](https://cloud.githubusercontent.com/assets/863262/24058425/6c69051a-0b21-11e7-9fc2-98ae7dc8afa3.png)
![screenshot_20170317-144622](https://cloud.githubusercontent.com/assets/863262/24058426/6c693904-0b21-11e7-9137-3522e40ec2ff.png)

roberthouse54:
I think this still makes sense to proceed as I said above. The main issue is that we need to make sure that content in the modal can scroll and not cover the buttons. But the other issues I mentioned should also be fixed.

PR #3116, opened 2017-04-27
Fixed mobile view of FA calculator

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/2474

#### What's this PR do?
Fixes the layout of  FA calculator.

Open FA calculator on mobile.

@pdpinch 
screen shot 2017-04-27 at 4 31 19 pm
screen shot 2017-04-27 at 4 31 24 pm
screen shot 2017-04-27 at 4 35 39 pm

### DIFF
--- a/static/scss/calculator.scss
+++ b/static/scss/calculator.scss
@@ -13,17 +13,43 @@
   }
 
   h3 {
-    border-bottom: none !important;
-    padding: 40px 40px 20px !important;
+    border-bottom: none;
+
+    @include breakpoint(phone) {
+      padding: 20px 20px 10px;
+    }
+
+    @include breakpoint(desktop) {
+      padding: 40px 40px 20px;
+    }
   }
 
   .financial-aid-calculator-body {
-    padding: 0 40px 20px !important;
-    overflow-y: visible !important;
+    @include breakpoint(desktop) {
+      padding: 0 40px 20px;
+      overflow-y: visible;
+    }
+
+    @include breakpoint(phone) {
+      overflow: scroll;
+      padding: 0 20px 10px;
+
+      .checkbox {
+        .mdl-checkbox__label {
+          font-size: 14px;
+        }
+      }
+    }
   }
 
   .salary-input, .salary-input > div {
-    display:  flex;
+    @include breakpoint(phone) {
+      display: block;
+    }
+
+    @include breakpoint(desktop) {
+      display: flex;
+    }
   }
 
   .api-error {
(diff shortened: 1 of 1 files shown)


aliceriot:
functionality looks great! I tested it out on master, it really was a mess before (yikes). looking at code now.

aliceriot:
one thing I just noticed: if I leave the 'I testify ...' box unchecked and click 'submit' the validation error that pops up is mushed up with the text:
overlap

static/scss/calculator.scss
```diff
@@ -81,3 +81,56 @@
...
+    }
+
+    h3 {
+      border-bottom: none !important;
+      padding: 20px 20px 10px !important;
+    }
+
+    .financial-aid-calculator-body {
+      overflow: scroll !important;
+      padding: 0 20px 10px !important;
```
aliceriot: there's a lot of `!important` here, is there any way we can avoid these? I know some things are styles set by mdl which we have to override, but it would be good to avoid `!important` as much as possible.

static/scss/calculator.scss
```diff
@@ -81,3 +81,56 @@
     }
   }
 }
+
+@include breakpoint(phone) {
+  .financial-aid-calculator, .confirm-dialog {
```
aliceriot: can you nest the mobile-specific code within the block for these selectors that we already have above?

aliceriot: one of the nice things (I think!) about the `breakpoint` mixin is it allows you to easily use breakpoints all over the place, for instance, my preferred way to use them is to set styles on specific things, so I might have something like

```scss
.my-component-class {
  .buttons {
    .first-button {
      color: blue,
      font-size: 12px;
      
      @include breakpoint(phone) {
        margin: 12px;
      }
      @include breakpoint(desktop) {
        margin: 24px;
      }
    }
  }
}
```

instead of having a big block for mobile code which declares all of the same selectors. does that make sense? IMO that approach makes the CSS easier to change, for instance if the class `.first-button` gets refactored to `.save-button` we only need to change it in one place.

amir-qayyum-khan:
Done @aliceriot

roberthouse54:
I looked at this with Alice and she is going to have you make a couple more small tweaks.

static/scss/calculator.scss
```diff
@@ -13,17 +13,43 @@
...
+      overflow-y: visible;
+    }
+
+    @include breakpoint(phone) {
+      overflow: scroll;
+      padding: 0 20px 10px;
+
+      .checkbox {
+        .mdl-checkbox__label {
+          font-size: 14px;
```
aliceriot: here can we do `font-size: 13px;` and `line-height: 1.2em;`?

static/scss/calculator.scss
```diff
@@ -65,19 +106,35 @@
...
+    }
+
+    @include breakpoint(phone) {
+      padding: 0 10px 10px 20px;
+      display: block;
+      text-align: left;
+    }
 
     .buttons {
       display: flex;
```
aliceriot: can you add `justify-content: flex-end;` here?

amir-qayyum-khan:
Done @aliceriot

aliceriot:
I'm still seeing the validation error overlap:

iphone 6:
still_overlap

iphone 5:
still_overlap_5

amir-qayyum-khan:
fixed @aliceriot

Issue #3111, opened 2017-04-25
email dialog display bug in FireFox


pdpinch:
This may be moot after #3088 is merged, but the e-mail dialog looks pretty ugly in Firefox 52.0.2:

![screenshot-micromasters-rc herokuapp com-2017-04-25-15-14-03](https://cloud.githubusercontent.com/assets/430126/25403333/05504ca4-29ca-11e7-9876-bb53f9dc1d7b.png)

Chrome and Safari both look fine.

PR #3127, opened 2017-05-02
Firefox fixed email type radios

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3111

#### What's this PR do?
fixes the style of radio buttons on email dialog search.

Go to learners search page as staff and op email dialog

screen shot 2017-05-03 at 12 48 49 am

### DIFF
--- a/static/scss/email-composition-dialog.scss
+++ b/static/scss/email-composition-dialog.scss
@@ -126,13 +126,13 @@
       font-size: 15px;
     }
 
-    > div:first-child {
-      max-width: 35%;
-    }
-  }
+    > div {
+      display: flex !important;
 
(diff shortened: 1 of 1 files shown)


static/scss/email-composition-dialog.scss
```diff
@@ -131,6 +131,10 @@
     }
   }
 
+  .send-one-time-email {
```
noisecapella: Can you add this as a rule under `.type-radio-group` above, and also apply it to all child divs instead of just `send-one-time-email`?

amir-qayyum-khan:
Done @noisecapella

Issue #3105, opened 2017-04-24
For non-FA courses, do not display Order Summary after clicking Pay Now button


roberthouse54:
This issue addresses a problem that we have in SCM, where the final exam (which is configured as an edX course) has a different price than other courses. The proposed solution is to limit the places where we display the cost of courses in the UI, so that the final source-of-truth for course pricing is edX. 

Solution:

1) For non-FA programs, we will skip the order summary page that is shown after the user clicks "Pay Now", and instead link directly to the edX site where users can complete the enrollment.

2) For all programs, remove the course price displayed on the Pay Now button in the modal. (see below)

![image](https://cloud.githubusercontent.com/assets/20047260/25391940/620346ec-29a5-11e7-8736-21fd07cb8706.png)

This should (mostly) solve the issue of displaying the course cost, since the modal and the order summary page are the 2 main places where we display the course price that may be leading to user confusion.

amir-qayyum-khan:
@roberthouse54 @pdpinch on edX where should we link user

> For non-fa instead link directly to the edX site where users can complete the enrollment.

pdpinch:
There should be existing code for this, because there is a link from the order summary view. I'm pretty sure it's the `enrollment_url` on the course run, possibly with a failover to a constructed URL based on the course key. 

If you can't find the existing code, let me know.

pdpinch:
Ugh. `enrollment_url` was the wrong answer. 

I think the existing code called a checkout function. We should have stuck with that. 

See https://github.com/mitodl/micromasters/blob/e5f6866b3879f04e3e5711605dac4d2ff08671fa/static/js/components/OrderSummary.js#L100

PR #3135, opened 2017-05-04
Bypassed order summary for non FA courses and redirect users to edX course enrollment page

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3105

#### What's this PR do?
Redirects user to course enrollment page instead of progress summary for Non financial aid programs only.

- Set course enrollment url from admin panel to non FA course run.
- Go to non FA program on dashboard
- Click `pay now` button.
- You will be redirect to course enrollment page. 

@pdpinch

### DIFF
--- a/static/js/components/CourseEnrollmentDialog_test.js
+++ b/static/js/components/CourseEnrollmentDialog_test.js
@@ -13,13 +13,14 @@ import { makeCourse, makeRun } from '../factories/dashboard';
 import { getEl } from '../util/test_utils';
 
 describe("CourseEnrollmentDialog", () => {
-  let sandbox, setVisibilityStub, addCourseEnrollmentStub, routerPushStub;
+  let sandbox, setVisibilityStub, addCourseEnrollmentStub, routerPushStub, windowOpenStub;
 
   beforeEach(() => {
     sandbox = sinon.sandbox.create();
     setVisibilityStub = sandbox.spy();
     addCourseEnrollmentStub = sandbox.spy();
     routerPushStub = sandbox.spy();
+    windowOpenStub = sandbox.stub(window, 'open');
   });
 
   afterEach(() => {
@@ -31,6 +32,7 @@ describe("CourseEnrollmentDialog", () => {
     courseRun = makeRun(1),
     course = makeCourse(1),
     open = true,
+    financialAidAvailability = true
   ) => {
     mount(
       <MuiThemeProvider muiTheme={getMuiTheme()}>
@@ -41,6 +43,7 @@ describe("CourseEnrollmentDialog", () => {
           price={price}
           setVisibility={setVisibilityStub}
           addCourseEnrollment={addCourseEnrollmentStub}
+          financialAidAvailability={financialAidAvailability}
         />
       </MuiThemeProvider>,
       {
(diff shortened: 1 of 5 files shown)


aliceriot:
my 'analog learning' program looks like this:

fjfjfjfffff

So I have two courses I'm already enrolled in (already showing 'pay now') and one I'm not yet enrolled in. This change works for the not-yet-enrolled case but the courses I'm already auditing aren't covered (i.e. if I click the 'pay now' button for analog learning 200 or 300 I get redirected to the order summary page, instead of opening the enrollment url).

Is this change supposed to covered paying for courses the user is currently auditing as well? It seems like it should to me.

amir-qayyum-khan:
Done @aliceriot

aliceriot:
functionality is fixed! reading code now.

static/js/containers/DashboardPage.js
```diff
@@ -550,6 +550,10 @@ class DashboardPage extends React.Component {
     );
   }
 
+  openLinkInNewTab = (url: ?string): void => {
+    window.open(url, '_blank');
+  }
```
aliceriot: It seems unnecessary to create this method on `DashboardPage` and pass it down through the component tree - I think it would be better either to directly call `window.open` where we need it or make this a small util function. When testing it I think we should be able to use sinon to stub out `window.open` for testing purposes.

static/js/components/dashboard/CourseAction_test.js
```diff
@@ -41,6 +42,11 @@ describe('CourseAction', () => {
   let setEnrollCourseDialogVisibilityStub;
   let openFinancialAidCalculatorStub;
   let routerPushStub;
+  let openLinkInNewTabStub;
+  const enrollmentUrlData = [
+    { url: undefined, expected: false },
+    { url: "http://test.com", expected: true }
+  ];
```
aliceriot: it looks like this array is only being used lower in the one test - can you move the declaration of it down where it is used, for readability?

static/js/components/CourseEnrollmentDialog_test.js
```diff
@@ -13,13 +13,17 @@ import { makeCourse, makeRun } from '../factories/dashboard';
 import { getEl } from '../util/test_utils';
 
 describe("CourseEnrollmentDialog", () => {
-  let sandbox, setVisibilityStub, addCourseEnrollmentStub, routerPushStub;
-
+  let sandbox, setVisibilityStub, addCourseEnrollmentStub, routerPushStub, openLinkInNewTabStub;
+  const enrollmentUrlData = [
+    { url: undefined, expected: false },
+    { url: "http://test.com", expected: true }
+  ];
```
aliceriot: same comment here (about moving next to the relevant test)

static/js/containers/DashboardPage.js
amir-qayyum-khan: it is to add tests.

aliceriot: did you try directly stubbing out `window.open` ?

amir-qayyum-khan: i think it is not possible, i am asserting the method in dashboard is called with correct params

aliceriot: I just tried in the node repl and I was able to do it, so there shouldn't be any problem doing it in our tests. You just do something like this in the `beforeEach` method for the relevant tests:

```js
let windowOpenStub = sandbox.stub(window, 'open');
```

then you can do the test and check that `windowOpenStub` is called with the arguments you expect.

amir-qayyum-khan:
Done with changes @aliceriot

PR #3143, opened 2017-05-04
Fixed missing module column in exam auth export (#3142)

### LINKED ISSUE #3142: Add Module column back to exam authorization export
Apparently we need this, even though we don't.

rhysyngsun:
Fixes #3142 

#### What's this PR do?
Adds this column back in but with an fixed empty string value

Tests should pass, this is simple enough change.

### DIFF
--- a/exams/pearson/writers.py
+++ b/exams/pearson/writers.py
@@ -272,6 +272,7 @@ class EADWriter(BaseTSVWriter):
             ('ClientAuthorizationID', 'id'),
             ('ClientCandidateID', 'user.profile.student_id'),
             ('ExamSeriesCode', 'exam_run.exam_series_code'),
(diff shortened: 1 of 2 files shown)


exams/pearson/writers.py
```diff
@@ -272,6 +272,7 @@ def __init__(self):
             ('ClientAuthorizationID', 'id'),
             ('ClientCandidateID', 'user.profile.student_id'),
             ('ExamSeriesCode', 'exam_run.exam_series_code'),
+            ('Modules', lambda _: ''),
```
noisecapella: Does Pearson complain if `Modules` is missing?

rhysyngsun: Yeah the processing errors on it.

PR #3145, opened 2017-05-05
Mail: Add Recipient Variables Toolbar

### LINKED ISSUE #2949: Mail: Support Mailgun macros in e-mail template

annagav:
Fix #2949 

#### What's this PR do?
Adding a toolbar below the email editor that inserts recipient variables into the email content.

As staff, open the email composition dialog and play with the toolbar.

screen shot 2017-05-05 at 12 20 18 am


noisecapella:
Can you add a test that a button click inserts the text in the editor state?

noisecapella:
Functionality looks good, it looks like this needs a mail ID button too

Also, I noticed that the markup is always added to the body of the email even if I'm focused on the title. I'm not sure if that matters though

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -78,6 +79,9 @@ export default class EmailCompositionDialog extends React.Component {
   constructor (props: EmailDialogProps) {
     super(props);
     this.state = editorStateFromProps(props);
+    if (R.isNil(this.state.editorState)) {
```
noisecapella: @aliceriot Do you have a suggestion how to update `editorStateFromProps` to set this? I see the `S.map` but I'm not sure how you convert a nothing into a different value

static/js/components/email/RecipientVariableButton.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
noisecapella: Can you add a `// @flow` before this line?

static/scss/email-composition-dialog.scss
```diff
@@ -165,3 +165,20 @@
     }
   }
 }
+
+.toolbar-below {
+  padding: 6px 5px 0;
+  display: -webkit-box;
```
noisecapella: The autoprefixer should take care of this and the `-ms-flexbox` below. You just need `display: flex`

noisecapella:
Can you also fix the lint and flow errors?

static/js/components/email/EmailCompositionDialog.js
aliceriot: What are we trying to accomplish? I think if `this.state.editorState` is an empty object the `<Editor />` component will instantiate it's own state, which will then be set as the state the first time the `onEditorStateChange` function fires.

aliceriot: so unless I'm missing something we shouldn't need to manually set `this.state.editorState` to be `EditorState.createEmpty()`

noisecapella: She needs to take care of the case where the first action the user does is to click one of the variable buttons, before `Editor` triggers `onEditorStateChange`

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -158,6 +205,7 @@ export default class EmailCompositionDialog extends React.Component {
       renderRecipients,
     } = this.props;
     const { editorState } = this.state;
+    let customButtons = R.map(this.makeCustomToolbarButton, ['PreferredName', 'Email']);
```
aliceriot: here you could declare the `makeCustomToolbarButton` as an anonymous function wrapped with `R.map` above, which would avoid needing to declare this here (you could just inline the function call in the JSX below).

static/js/components/email/EmailCompositionDialog.js
aliceriot: Oh gotcha, that makes sense.

aliceriot: In that case, the last function in the `editorStateFromProps` composition needs to be modified. Right now it looks like this:

```js
S.maybe({}, R.objOf('editorState'))
```

The `S.maybe` function has a the type signature `maybe :: b -> (a -> b) -> Maybe a -> b` (or something along those lines). Basically, the arguments are `maybe(defaultValue, onJustFunc, Maybe)`. If the `Maybe` is `Nothing`, the function will return `defaultValue`, if the `Maybe` is `Just`, it will return `onJustFunc(value)`, where `value` is the value wrapped up in the `Just`.

So right now if the function in question receives a `Nothing` it will return `{}`, and if it receives a `Just(x)`, it will return `R.objOf('editorState', x)`, which returns `{editorState: x}`. So I think all we would need to change is the default value, which should go from being `{}` to being `{editorState: EditorState.createEmpty()}`. So that would look like:

```js
maybe({ editorState: EditorState.createEmpty() }, R.objOf('editorState'))
```

 Because `EditorState` is an Immutable value we shouldn't need to worry about creating it every time the `editorStateFromProps` function is called (I think).

aliceriot: @annagav does that all make sense? lmk if there are any questions!

annagav:
@noisecapella , @aliceriot I think I addressed all your comments.

noisecapella:
@annagav It looks like the buttons are inside the text area on Firefox and IE, and they're also not clickable:
screenshot from 2017-05-09 10-48-36

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -86,6 +87,42 @@ export default class EmailCompositionDialog extends React.Component {
     }
   }
 
+  insertRecipientVariable = (variableName: string) => {
+
+    const { editorState } = this.state;
+    if(editorState !== null && editorState !== undefined) {
```
noisecapella: I don't think this should ever be null since `editorState` is defined in the constructor

static/js/components/email/EmailCompositionDialog.js
```diff
@@ -86,6 +87,42 @@ export default class EmailCompositionDialog extends React.Component {
     }
```
noisecapella: Can you remove this function? `this.state.editorState` shouldn't ever be null or undefined here so we shouldn't need to do this since `this.state` is set in the constructor and `editorStateFromProps` always returns sets some value for `editorState`

static/js/components/email/EmailCompositionDialog.js
annagav: I added this to avoid flow errors.

aliceriot: Probably you should update the type for `this.state` then, from `editorState: ?Object` to `editorState: Object`. `?Object` is no longer correct, since we can't really get into a state where it is `null` or `undefined`

noisecapella:
Firefox and IE functionality works fine, just waiting on the flow change @aliceriot suggested

PR #3147, opened 2017-05-05
Fixed an issue with old ExamAuthorizations updating (#3146)

### LINKED ISSUE #3146: When ExamRuns update, don't updated ExamAuthorizations if the run is not schedulable
This isn't necessary and probably leaves us open to edge case bugs.

rhysyngsun:
Fixes #3146 

#### What's this PR do?
Adds a check not to update `ExamAuthorizations` if the `ExamRun` is not currently schedulable.

If the tests pass, that's sufficient.

### DIFF
--- a/exams/api_test.py
+++ b/exams/api_test.py
@@ -3,6 +3,7 @@ Tests for exams API
 """
 from unittest.mock import patch
 
+import ddt
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models.signals import post_save
 from django.test import (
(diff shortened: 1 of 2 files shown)


exams/api_test.py
```diff
@@ -257,6 +257,78 @@ def test_update_authorizations_for_exam_run(self):
             assert auth.status == ExamAuthorization.STATUS_PENDING
             assert auth.operation == ExamAuthorization.OPERATION_UPDATE
 
+    def test_update_authorizations_for_past_exam_run(self):
```
noisecapella: Instead of adding these two tests can you parameterize the previous test to handle the three cases?

rhysyngsun: They're slightly different, but I'll see what I can do..

rhysyngsun:
@noisecapella addressed your feedback

PR #3165, opened 2017-05-09
CMS: Link ProgramCourse to Course

### LINKED ISSUE #3107: Program Page: Use Course Descriptions from the CMS
We have a section in the CMS for course teams to enter course descriptions, but they're not actually used anywhere. 

In the course card on the program page, we display course descriptions pulled from the courses model. This doesn't make any sense because course teams don't have access to change those fields. 

So, we should update the Program Page template to pull course descriptions from the CMS instead of from the course model. 

### acceptance criteria:

- [ ] migrate existing course descriptions from course model to CMS
- [ ] update Program Page template to pull from CMS
- [ ] update admin for course model to indicate that the description is deprecated

### future issues:

- [ ] remove description from course model

annagav:
Fix #3107

#### What's this PR do?
Show the description of the course as it was entered in wagtail.

In cms, make sure you have a ProgramPage with some ProgramCourses, where the title and description do not correspond to the actual Course in `courses`.  Run the migrations. Notice in wagtail that your program courses for that program now correspond to actual courses in that program. Now when you change the description of a course in wagtail, it should be visible in the popover when you click on the course on the program page.


cms/migrations/0028_delete_program_courses.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.5 on 2017-05-08 19:06
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def remove_program_courses(apps, schema_editor):
+    ProgramCourse = apps.get_model('cms', 'ProgramCourse')
+    ProgramCourse.objects.all().delete()
```
rhysyngsun: I'm not sure we need to delete these ahead of time because the added `course` column is nullable. I think you could just delete them at the start of 0030 instead and save a migration.

cms/migrations/0030_populate_program_courses.py
```diff
@@ -0,0 +1,39 @@
...
+            ProgramCourse.objects.create(
+                course=course,
+                title=course.title,
+                description=course.description,
+                program_page=program_page
+            )
+
+
+def reverse_populate_program_courses(apps, schema_editor):
+    ProgramCourse = apps.get_model('cms', 'ProgramCourse')
```
rhysyngsun: We might actually want to do a `RunPython.noop` for the reverse migration here, my thinking is if we reverted this migration we would end up in a state with no `ProgramCourse`s.

annagav:
@rhysyngsun, I think we don't want them to be able to change the course title, only description.

PR #3178, opened 2017-05-12
On pay now redirected users to checkout page instead of order summary for non FA programs

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3105

#### What's this PR do?
- Redirect user to checkout instead of progress summary for `NON FA` programs. 

- Click `pay now` buttons for no fa course

@pdpinch

### DIFF
--- a/static/js/components/dashboard/CourseRow.js
+++ b/static/js/components/dashboard/CourseRow.js
@@ -37,7 +37,7 @@ export default class CourseRow extends React.Component {
     setEnrollSelectedCourseRun: (r: CourseRun) => void,
     setEnrollCourseDialogVisibility: (b: boolean) => void,
     ui: UIState,
-    checkout: Function,
+    checkout: (s: string) => void,
   };
 
   shouldDisplayGradeColumn = (run: CourseRun): boolean => (
--- a/static/js/components/dashboard/CourseRow.js
+++ b/static/js/components/dashboard/CourseRow.js
@@ -37,6 +37,7 @@ export default class CourseRow extends React.Component {
     setEnrollSelectedCourseRun: (r: CourseRun) => void,
     setEnrollCourseDialogVisibility: (b: boolean) => void,
     ui: UIState,
+    checkout: Function,
   };
 
   shouldDisplayGradeColumn = (run: CourseRun): boolean => (
@@ -83,6 +84,7 @@ export default class CourseRow extends React.Component {
       openCourseContactDialog,
       setEnrollSelectedCourseRun,
       setEnrollCourseDialogVisibility,
+      checkout
     } = this.props;
 
     let lastColumnSize, lastColumnClass;
@@ -121,6 +123,7 @@ export default class CourseRow extends React.Component {
           now={now}
           prices={prices}
           hasFinancialAid={hasFinancialAid}
+          checkout={checkout}
           financialAid={financialAid}
           openFinancialAidCalculator={openFinancialAidCalculator}
           addCourseEnrollment={addCourseEnrollment}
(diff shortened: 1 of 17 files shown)


noisecapella:
It looks like the price is still part of the modal. The linked issue says it should be removed
screenshot from 2017-05-12 10-55-13

static/js/containers/DashboardPage.js
```diff
@@ -550,6 +552,11 @@ class DashboardPage extends React.Component {
     );
   }
 
+  dispatchCheckout = (courseId: string) => {
+    const { dispatch } = this.props;
```
noisecapella: Can you add a test which would cover this function?

static/js/components/dashboard/CourseRow.js
```diff
@@ -37,6 +37,7 @@ export default class CourseRow extends React.Component {
     setEnrollSelectedCourseRun: (r: CourseRun) => void,
     setEnrollCourseDialogVisibility: (b: boolean) => void,
     ui: UIState,
+    checkout: Function,
```
noisecapella: Can you make this type more specific?

amir-qayyum-khan:
Done @noisecapella

static/js/containers/DashboardPage_test.js
```diff
@@ -619,4 +621,40 @@ describe('DashboardPage', () => {
       });
     });
   });
+
+  describe("checkout", () => {
+    let dashboardResponse;
+    beforeEach(() => {
+      // Limit the dashboard response to 1 program
+      dashboardResponse = {"programs": [R.clone(DASHBOARD_RESPONSE.programs[0])]};
```
noisecapella: Can you use `makeProgram()` instead? This should also let you use `program.courses[0]` for the course in the test below.

static/js/containers/DashboardPage_test.js
```diff
@@ -619,4 +621,40 @@ describe('DashboardPage', () => {
...
+    });
+
+    it('redirects to edX when the checkout API tells us to', () => {
+      let promise = Promise.resolve(EDX_CHECKOUT_RESPONSE);
+      let checkoutStub = helper.sandbox.stub(storeActions, 'checkout').returns(() => promise);
+      let course = makeCourse();
+      // Set all course runs to unpaid
+      course.runs = R.chain(R.set(R.lensProp('has_paid'), false), course.runs);
```
noisecapella: `course` can be modified directly so you shouldn't need the lens or setter methods here, and it looks like you only need to modify one run in the course. Can you do something like: `course.runs[0] = { ...course.runs[0], status: STATUS_CAN_UPGRADE, course_end_date: moment().add(-1, 'months'), ` and so on

static/js/containers/LearnerSearchPage_test.js
```diff
@@ -230,7 +230,7 @@ describe('LearnerSearchPage', function () {
         query: 'xyz',
         type: 'phrase_prefix',
       });
-      assert.equal(window.location.toString(), "http://fake/?q=xyz");
+      assert.equal(window.location.toString(), "http://edx.org/?q=xyz");
```
noisecapella: Why is this change necessary? If the problem is that `window.location` from one test is affecting this one, we should solve the problem in `babelhook.js` or `global_init.js` instead, where we handle `window.location` mocking

static/js/containers/DashboardPage_test.js
amir-qayyum-khan: dashboard is not rendering with makeProgram(), i dont know what is missing

static/js/containers/LearnerSearchPage_test.js
amir-qayyum-khan: any thing in ur mind

amir-qayyum-khan:
Done @noisecapella

Issue #3187, opened 2017-05-15
Search: Country of Residence selected values disappear when there are no search results


rhysyngsun:
Select a combination of CoR and other filters to get 0 results, observe that the selected values disappear.

Workaround: click "X" on the "Selected Filters" item under the search box to remove the filter

PR #3188, opened 2017-05-15
Ensure facet titles for selected filters stay visible

### LINKED ISSUE #3095: Fix histogram and range facets in UI
We have a few UI issues with the histogram and range facets on the learner's search page

**1) When a range is selected on a histogram that yields no results, and a course filter is applied, the selected course filter remains in the facet** 

ss 2017-04-20 at 14 04 59

~~**2) The '# of courses passed' selected filter title is missing**~~

ss 2017-04-20 at 14 05 46

**3) When a range is applied for '# of courses passed', and any course filter is also applied, the '# of courses passed' facet disappears while the filter remains**

ss 2017-04-20 at 14 05 54

rhysyngsun:
Fixes #3095 

#### What's this PR do?
Adds logic to still show titles if the filter is selected when there are 0 results.

- Go to learner search
- Selected any combination of filters to get 0 results (hint: you can select filters that narrow your search down to 1 and then search for a nonmatching name to test all of them).
- Verify the facet titles remain

Known issue: #3187

### DIFF
--- a/static/js/components/search/EnabledSelectionRangeFilter.js
+++ b/static/js/components/search/EnabledSelectionRangeFilter.js
@@ -1,0 +1,20 @@
+import { RangeFilter, RangeAccessor } from "searchkit";
+
+export class EnabledSelectionRangeAccessor extends RangeAccessor {
+  isDisabled() {
+    return !this.state.hasValue() && super.isDisabled();
+  }
+}
+
+export default class EnabledSelectionRangeFilter extends RangeFilter {
+  defineAccessor() {
+    const {
+      id, title, min, max, field, fieldOptions,
+      interval, showHistogram
+    } = this.props;
+    return new EnabledSelectionRangeAccessor(id, {
+      id, min, max, title, field,
+      interval, loadHistogram: showHistogram, fieldOptions
+    });
+  }
+}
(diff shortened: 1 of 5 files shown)


noisecapella:
It looks like final grade disappears completely:
screenshot from 2017-05-16 10-56-20

screenshot from 2017-05-16 10-55-59

static/js/components/search/FilterVisibilityToggle.js
```diff
@@ -64,15 +64,19 @@ export default class FilterVisibilityToggle extends SearchkitComponent {
     return false;
   };
 
+  isFilterSelected = (id: string): boolean => {
```
noisecapella: Can you change `stayVisibleIfEmpty` to make use of this function since it looks like they overlap?

rhysyngsun: I think I already did that, unless I'm missing something.

noisecapella: Yes, I missed that change

rhysyngsun:
@noisecapella addressed your feedback

static/js/components/search/EnabledSelectionRangeFilter.js
```diff
@@ -0,0 +1,20 @@
+import { RangeFilter, RangeAccessor } from "searchkit";
```
noisecapella: Can you add `// @flow` here?

PR #3189, opened 2017-05-15
Fixed # of Courses Passed facet disappearing (#3095)

### LINKED ISSUE #3095: Fix histogram and range facets in UI

rhysyngsun:
Fixes #3095

#### What's this PR do?
Adds the number of courses to the `/programs` api so that the learner search can set a proper maximum range value, rather than depending on the seemingly flaky value in the first hit (which we could be nonexistent if zero results).

- Adjust the bounds of the '# of courses passed' range filter (I set it to 2-3)
- Select any one of the course title filters (I chose 'Digital Learning 100')
- Verify the range slider for '# of courses passed' is still present

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -135,7 +135,8 @@ export default class LearnerSearch extends SearchkitComponent {
   }
 
   getNumberOfCoursesInProgram = (): number => {
-    return R.pathOr(0, ['hits', 'hits', 0, '_source', 'program', 'total_courses'], this.getResults());
+    const { currentProgramEnrollment } = this.props;
+    return R.pathOr(0, ['total_courses'], currentProgramEnrollment);
   };
 
   renderSearchHeader = (): React$Element<*>|null => {
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -11,6 +11,7 @@ class ProgramSerializer(serializers.ModelSerializer):
     """Serializer for Program objects"""
     programpage_url = serializers.SerializerMethodField()
     enrolled = serializers.SerializerMethodField()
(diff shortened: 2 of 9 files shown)


noisecapella:
It looks like `currentProgramEnrollment` will keep the old data even if the programs are refreshed. Is there a way to force a refresh of `localStorage` for this PR?

rhysyngsun:
I'd have to add some code that detects the structural difference between what's in `localStorage` vs. what came back from the API, but it's doable, if ugly.

noisecapella:
I think it's necessary because most people will only be looking at one program so they won't have the opportunity to refresh their localStorage. We could just check if `total_courses` exists, set it to null, then proceed as before

static/js/components/LearnerSearch.js
```diff
@@ -135,7 +135,8 @@ export default class LearnerSearch extends SearchkitComponent {
   }
 
   getNumberOfCoursesInProgram = (): number => {
-    return R.pathOr(0, ['hits', 'hits', 0, '_source', 'program', 'total_courses'], this.getResults());
+    const { currentProgramEnrollment } = this.props;
```
noisecapella: Can you add a test that the facet range has the right value?

PR #3191, opened 2017-05-16
Update edX cache only for active users

### LINKED ISSUE #2099: Don't attempt to update edX cache for users who are inactive
We get the following error every time we run a batch update:

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/193071074/

```
DoesNotExist: UserSocialAuth matching query does not exist.
  File "dashboard/tasks.py", line 111, in batch_update_user_data_subtasks
    UserSocialAuth.objects.get(user=user)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 385, in get
    self.model._meta.object_name
```

The user in question does not have active set to `true` so I think we can just skip any updates when `user.active = false`

annagav:
Fix #2099

#### What's this PR do?
Exclude all inactive users from cached users.

Set your user to `is_active=False` and make sure that user is not being cached anymore.

### DIFF
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -57,12 +57,13 @@ def batch_update_user_data():
         users_expired_cache = UserCacheRefreshTime.objects.filter(
             Q(enrollment__lt=refresh_time_limit) |
             Q(certificate__lt=refresh_time_limit) |
-            Q(current_grade__lt=refresh_time_limit)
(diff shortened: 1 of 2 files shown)


noisecapella:
Please add a test for this

PR #3195, opened 2017-05-16
Added wait function

noisecapella:
None

#### What's this PR do?
Added a `wait` function which wraps around `setTimeout` and returns a promise which resolves when the time has elapsed.

Go to `/dashboard/?status=receipt&course_key=missing` and look at your network tab. Every three seconds there should be a dashboard refresh

### DIFF
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -554,3 +554,11 @@ export const mapObj = R.curry((fn, obj) => R.compose(
   R.map(fn),
   R.toPairs
 )(obj));
+
+export const wait = (millis: number): Promise<void> => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve();
+    }, millis);
+  });
+};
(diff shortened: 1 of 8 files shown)


aliceriot:
functionality looks good! reading code now.

static/js/util/util.js
```diff
@@ -554,3 +554,11 @@ export const mapObj = R.curry((fn, obj) => R.compose(
   R.map(fn),
   R.toPairs
 )(obj));
+
+export const wait = (millis: number): Promise<void> => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve();
+    }, millis);
```
aliceriot: since we're not passing any value to `resolve` here (for the `.then` handler to grab later) could we just do

```js
  setTimeout(resolve, millis);
});
```

aliceriot: could even one-line it with (seemingly?) no ill effects:

```js
export const wait = (millis: number) => (
  new Promise(resolve => setTimeout(resolve, millis))
);
```

noisecapella: Sounds good

PR #3199, opened 2017-05-16
Hid final exam grades that aren't available yet (#3194)

### LINKED ISSUE #3194: Exams: hide proctored exam grades until they are approved
We need to hide these grades in a way that we can process them, but not show them to learners. I propose:

- Add `ExamRun.date_grades_available` when will represent the date after which exam grades will be shown.
- Add a `ForeignKey` from `ProctoredExamGrade` to `ExamRun`
- Filter dashboard exam grades to ones which have `ExamRun.date_grades_available__lte=now`.

rhysyngsun:
Fixes #3194

#### What's this PR do?
Adds fields to `ExamRun` and `ProctoredExamGrade` so we can filter out grades that have been processed but aren't available yet.

- Before running the migrations ensure you user has a `ProctoredExamGrade` for an `ExamRun` in the past that has `exam_series_code='PILOT'`. 
- After the migrations the grade should have `exam_run` populated and the run should have `date_grades_available` populated from the value of `date_last_eligible`.
- Ensure you now can still see the grade on the dashboard.
- Change `date_grades_available` to the future and verify you no longer see the grade.

Since we're constructing a new FK, I didn't set `ProctoredExamGrade.exam_run` `null=False` because I would rather not the migration fail in case we encounter something unexpected. We can reconcile those with issue later if they come up.


exams/models.py
```diff
@@ -41,6 +43,23 @@ def get_currently_schedulable(cls, course):
             date_last_schedulable__gte=now,
         )
 
+    @classmethod
+    def get_grades_available_exams(cls, course):
```
gsidebo: suggested name change: `get_all_with_grades_available`

exams/models.py
```diff
@@ -21,6 +21,8 @@ class ExamRun(TimestampedModel):
     date_first_eligible = models.DateField()
     date_last_eligible = models.DateField()
 
+    date_grades_available = models.DateTimeField()
+
```
gsidebo: any reason for the whitespaces above & below the column definition here?

rhysyngsun: Personal style, I like to visually separate related fields. Should I change?

gsidebo: ehhh... i won't be a stickler right now since the pattern is repeated elsewhere, but i'd prefer no grouping by whitespace for a bunch of small reasons (people having different ideas of 'related', etc). no need to change this now

rhysyngsun:
Addressed the func name suggestion.

grades/models.py
```diff
@@ -223,7 +225,8 @@ def for_user_course(cls, user, course):
         """
         Returns a queryset of the exam result for an user in a course of a program
         """
-        return cls.objects.filter(user=user, course=course)
+        grades_available_runs = ExamRun.get_all_with_available_grades(course)
+        return cls.objects.filter(user=user, course=course, exam_run__in=grades_available_runs)
```
gsidebo: The two separate queries here (in `ProctoredExamGrade`) and in `ExamRun` cause the resulting DB query to have a subquery, ie: `select * from grades_proctoredexamgrade where course_id in (select course_id from exam_run ... )`. Not the end of the world obviously, but it kinda seems like that logic is only relevant in this model (this is the only client of `ExamRun.get_all_with_available_grades`, and I would argue that it's likely to stay that way). If you perform the query for `date_grades_available__lte=now` in this method, we avoid the nested query and we keep the logic more proximal. What do you think?

rhysyngsun: That's fine, I agree with that. Postgres probably optimizes the subquery into a join anyway.

PR #3203, opened 2017-05-17
email composition dialog CSS fixes

### LINKED ISSUE #3197: Reset the CSS on the New Email form
The rich text editor on the email form does not embed styles in the output html, but instead just sets the basic html elements (H1, H2, P, etc). However, in the actual form, the text that is set to H1, H2 etc gets styles from Micromasters css. This means that the email as viewed by the sender will look different than the email viewed by the recipient. Also, because of various css anomalies, the h2 header is smaller than the h4 header etc.

To fix this issue, we should apply a css rest to the email composition form. This will allow the email to look as similar as possible to both sender and receiver (allowing for the usual variations in how different mail clients display text etc).

For example...

Email as viewed by sender:
image

 Same email as received:
image

aliceriot:

closes #3197

#### What's this PR do?

This is just some CSS to make the heading tags display normally-ish inside of the email composition dialog.

Make sure everything looks alright!

### DIFF
--- a/static/scss/email-composition-dialog.scss
+++ b/static/scss/email-composition-dialog.scss
@@ -40,9 +40,58 @@
           height: auto;
           border: 1px solid $border-color;
           border-radius: 2px;
+          max-height: 50vh;
 
           .public-DraftEditor-content {
             padding-left: 10px;
+            color: black;
+            font-weight: normal;
+            font-family: arial;
+
+            // this is an attempt at a hacky default-ish display for hX tags
+            h1 {
+              font-size: 2em;
+              margin-top: 0.67em;
+              margin-bottom: 0.67em;
+            }
+
+            h2 {
+              font-size: 1.5em;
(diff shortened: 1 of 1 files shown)


static/scss/email-composition-dialog.scss
```diff
@@ -40,9 +40,78 @@
...
+              font-size: .67em;
+              margin-top: 2.33em;
+              margin-bottom: 2.33em;
+              margin-left: 0;
+              margin-right: 0;
+              font-weight: bold;
+            }
+
+            h1, h2, h3, h4, h5, h6 {
+              padding: 0;
```
rhysyngsun: Could you refactor `margin-left`, `margin-right`, `display`, and `font-weight` from the individual style defs above into this one since they're all the same?

aliceriot: oh yeah, that makes sense, haha

Issue #3125, opened 2017-05-02
Coupons without a code aren't showing up on order summary page


noisecapella:
If the coupon code is blank (which may happen for automatic coupons) we should still show the discount in the order summary page

PR #3210, opened 2017-05-18
Shown Coupons without code on order summary page

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3125

#### What's this PR do?
Allow coupon without code display in order summary

Go to order summary on a course which has a coupon without code

@pdpinch

### DIFF
--- a/static/js/components/OrderSummary.js
+++ b/static/js/components/OrderSummary.js
@@ -70,9 +70,9 @@ class OrderSummary extends React.Component {
   render() {
     let { course, courseRun, checkout, checkoutStatus, couponCode, discount} = this.props;
     let discountInfo;
-    const message = couponCode ? `Discount from coupon ${couponCode}` : 'Discount from coupon';
 
     if (discount) {
+      const message = couponCode ? `Discount from coupon ${couponCode}` : 'Discount from coupon';
       discountInfo = [
         this.showAmount(message, this.getDiscountAmount()),
         <Cell col={10} tablet={6} phone={4} className="division-line" key="division"/>,
--- a/static/js/components/OrderSummary.js
+++ b/static/js/components/OrderSummary.js
@@ -68,15 +68,18 @@ class OrderSummary extends React.Component {
   }
 
   render() {
-    let { course, courseRun, checkout, checkoutStatus, couponCode} = this.props;
+    let { course, courseRun, checkout, checkoutStatus, couponCode, discount} = this.props;
     let discountInfo;
-    if (couponCode) {
+    const message = couponCode ? `Discount from coupon ${couponCode}` : 'Discount from coupon';
+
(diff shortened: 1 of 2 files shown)


static/js/components/OrderSummary.js
```diff
@@ -68,15 +68,18 @@ class OrderSummary extends React.Component {
   }
 
   render() {
-    let { course, courseRun, checkout, checkoutStatus, couponCode} = this.props;
+    let { course, courseRun, checkout, checkoutStatus, couponCode, discount} = this.props;
     let discountInfo;
-    if (couponCode) {
+    const message = couponCode ? `Discount from coupon ${couponCode}` : 'Discount from coupon';
```
gsidebo: This line should be moved inside the `if (discount) {` block since it's only used there

static/js/components/OrderSummary_test.js
```diff
@@ -64,27 +64,32 @@ describe('OrderSummary', () => {
...
-    ));
-    let firstRun = course.runs[0];
-    const wrapper = renderOrderSummary({
-      courseRun: firstRun,
-      course: course,
-      discount: COURSE_PRICES_RESPONSE[1].price,
-      finalPrice: 0,
-      couponCode: '561KH'
+  for (let code of ['561KH', null, '']) {
+    it(`shows discount calculation if user has a coupon with code ${code}`, () => {
```
gsidebo: I would recommend changing the loop and the description of the test cases to something like this:

```javascript
[
  ['561KH', 'non-blank coupon code'],
  [null, 'null coupon code'],
  ['', 'blank coupon code']
].forEach(([code, codeDescription]) => {
  it(`shows discount calculation if user has a coupon with ${codeDescription}`, () => {
    // ...    
  });
});
```

amir-qayyum-khan:
Done @gsidebo

static/js/components/OrderSummary_test.js
```diff
@@ -64,26 +64,35 @@ describe('OrderSummary', () => {
...
-      course: course,
-      discount: COURSE_PRICES_RESPONSE[1].price,
-      finalPrice: 0,
-      couponCode: '561KH'
+  [
+    ['561KH', 'non-blank coupon code'],
+    [null, 'null coupon code'],
+    ['', 'blank coupon code']
+  ].forEach(([code, codeDescription]) => {
+    it(`shows discount calculation if user has a coupon with code ${codeDescription}`, () => {
```
gsidebo: one last nitpick: please change this to `shows discount calculation if user has a coupon with ${codeDescription}`. the way it is now, an example description would be "shows discount calculation if user has a coupon with code blank coupon code". we want that to be "shows discount calculation if user has a coupon with ~~code~~ blank coupon code"

PR #3215, opened 2017-05-18
Fixed bug involving exam no-shows

### LINKED ISSUE #3084: Exams: handle empty fields for no-shows
It appears some fields can be empty (and we try to parse as a date/bool/int) if the user was a no-show. We should not fail or alert in these cases because there's nothing to do about it:

Example error (exact values redacted):
```
Unable to parse row '{'ClientCandidateID': '1234567890', 'Incorrect': '', 'PassingScore': '', 'NoShow': 'true', 'Form': '', 'ExamSeriesCode': 'MicroMasters', 'TCID': '12345', 'Unscored': '', 'Grade': '', 'TimeUsed': '', 'Score': '', 'Skipped': '', 'ExamName': '', 'Voucher': '', 'Attempt': '', 'ExamLanguage': 'ENU', 'ClientAuthorizationID': '123', 'ExamDate': '03/1/2017 12:30:00', 'NDARefused': '', 'RegistrationID': '12345', 'Correct': '', 'ExamRevision': ''}'
```

Issue here is the empty values. This should be as simple as littering `EXAMReader` with `parse_or_default` wherever we parse to a non-string type to not encounter the parsing error.

We should not record a `ProctoredExamGrade` if `NoShow is True` but we should still set `ExamAuthorization.exam_taken=True` and we should probably add another field `ExamAuthorization.no_show` so we know why there is no `ProctoredExamGrade`, which defaults to `False` and set to `True` in these cases.

gsidebo:
Fixes #3084 

#### What's this PR do?
Fixes bug involving exam no-shows

Using the example no-show pearson file, the following code block should yield a result with no error messages

```python
FILE = 'noshow.dat'
from exams.pearson.readers import EXAMReader
from exams.pearson.download import ArchivedResponseProcessor
from exams.pearson import sftp
process_result = None
with sftp.get_connection() as sftp_connection, open(FILE, 'r') as f:
    processor = ArchivedResponseProcessor(sftp_connection)
    process_result = processor.process_exam_file(f)
process_result
```


exams/pearson/factories.py
```diff
@@ -49,7 +49,20 @@ class Meta:  # pylint: disable=missing-docstring
     class Params:  # pylint: disable=missing-docstring
         passed = factory.Trait(
             grade='pass',
+            no_show=False,
         )
         failed = factory.Trait(
             grade='fail',
+            no_show=False,
```
rhysyngsun: I don't think you need to specify these as `False` as that's implicit.

rhysyngsun:
Also, needs a rebase because your migration version # collided with one already in master.

exams/pearson/factories.py
rhysyngsun: ignore

PR #3229, opened 2017-05-23
Add function to calculate current time in UTC

### LINKED ISSUE #3205: Standardize date creation around UTC now
We have a ton of calls to `datetime.datetime.now` with the UTC timezone as an arg or `tz` kwarg. This is clearly the standard way to get the current datetime in the codebase, so we should have a way to calculate 'now' without passing a specific parameter to a specific function. This will help us simplify our code and avoid non-standard usage. The simplest approach would probably be creating a util function and replacing all existing `now` calls with that, but there might be some library function that already exists

noisecapella:
Fixes #3205 

#### What's this PR do?
Adds a utility function to calculate the current time in UTC, and fixes a couple places where we weren't specifying a timezone

Tests should pass, nothing should break

### DIFF
--- a/financialaid/migrations/0003_added_timestamps.py
+++ b/financialaid/migrations/0003_added_timestamps.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals
 
 import datetime
 from django.db import migrations, models
-
+import pytz
 
 class Migration(migrations.Migration):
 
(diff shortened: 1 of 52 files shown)


giocalitri:
I found other places where UTC is not present... not sure they are relevant
```
courses/utils_test.py
29:        fall_2016_dt = datetime(2016, 10, 1)

dashboard/serializers_test.py
445:        valid_semester_course_run = CourseRunFactory.build(start_date=datetime(2017, 1, 1))

exams/migrations/0005_populate_timestamps.py
16:        exam_profile.created_on = datetime(2017, 3, 1)  # ExamProfile records created for the beta

exams/migrations/0008_populate_exam_run.py
9:PILOT_SCHEDULE_START_DATE = datetime(2017, 3, 6)
10:PILOT_SCHEDULE_END_DATE = datetime(2017, 3, 30)
11:PILOT_ELIGIBILITY_START_DATE = datetime(2017, 3, 18)
12:PILOT_ELIGIBILITY_END_DATE = datetime(2017, 3, 31)

financialaid/views_test.py
720:            "date_documents_sent": datetime.datetime(2016, 9, 25).strftime("%Y-%m-%d")

```

plus I am not sure why this one is in a localized TZ

```
seed_data/utils.py
68:def localized_datetime(dt, tz='US/Eastern'):
```

financialaid/migrations/0003_added_timestamps.py
```diff
@@ -4,7 +4,7 @@
 
 import datetime
 from django.db import migrations, models
-
+import pytz
```
noisecapella: This was necessary because it was causing warnings in the selenium tests, which explicitly run all migrations to populate an initial database

micromasters/utils_test.py
```diff
@@ -294,3 +295,10 @@ def test_dict_with_keys(self):
             'a': 1,
         }
         assert dict_with_keys(source_dict, ['a', 'b']) == source_dict
+
+
+def test_now_in_utc():
+    """now_in_utc() should get the current time"""
```
noisecapella: Sure

noisecapella:
@giocalitri I added the timezones to the places you specified. I left the localized timezone alone since I didn't want to make any changes in behavior in this PR

financialaid/migrations/0003_added_timestamps.py
giocalitri: Yes, I understand, but what about adding an extra migration instead of modifying this one?

noisecapella: It wouldn't help because the warning would occur on the first migration, even if the time zone is added on the second migration

giocalitri: Understood.
BUT, do we need the extra migration anyway to convert the existing data?

noisecapella: I don't think so, all `DateTimeField`s will have a timezone set since `USE_TZ = True` in the settings, even if they were set with naive `datetime` object. For example:

    In [8]: n = datetime.now()
    
    In [9]: n
    Out[9]: datetime.datetime(2017, 5, 25, 14, 17, 24, 487031)
    
    In [10]: n.tzinfo
    
    In [11]: p.updated_on = n
    
    In [12]: p.save()
    
    In [13]: p.refresh_from_db()
    
    In [14]: p.updated_on
    Out[14]: datetime.datetime(2017, 5, 25, 14, 17, 34, 931526, tzinfo=<UTC>)

giocalitri:
Ok, this is good to go  👍

Issue #3202, opened 2017-05-17
Show selected recipients when editing an email campaign


roberthouse54:
When users create a new email message or campaign, they see the recipients as filters. However, when users are editing an email campaign, the recipients are not shown. 

- We should show the recipients when editing email campaigns. The recipients are not editable... this is view only.
- We should also change the name of that modal to "Edit Email Campaign" (currently it is "Edit Message"

![image](https://cloud.githubusercontent.com/assets/20047260/26160955/8f3b977a-3af0-11e7-96f9-1f74d6fb9a90.png)

amir-qayyum-khan:
@pdpinch @noisecapella @giocalitri @aliceriot 

Currently there is no way to save recipients on servers side what approach should i use

1) should i save comma separated recipients on model?
or any thing else is in ur mind?

aliceriot:
I'm not sure the exact approach, but the query that is used to generate the automatic email is saved as a percolate query, right? perhaps we can use that saved percolate query to generate the labels and so on that we would want to display?

PR #3238, opened 2017-05-25
Show recipients on email edit box

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3202

#### What's this PR do?
Adds filters/recipients to automatic email page

(Required)
- on learners search page select as much filters u can
- create an automatic email from learner search page.
- open the automatic email page and press edit 
- you will see the filters 

screen shot 2017-05-31 at 12 10 51 am

### DIFF
--- a/static/js/components/email/lib.js
+++ b/static/js/components/email/lib.js
@@ -148,7 +148,7 @@ export const convertEmailEdit = mapObj(([k, v]) => (
   [k.match(/^subject$|^body$/) ? `email_${k}` : k.replace(/^email_/, ""), v]
 ));
 
-const findTerms = (tree) => {
+export const findTerms = (tree) => {
   if (tree.hasOwnProperty("term") && !tree.term.hasOwnProperty("program.id")) {
     return [tree.term];
   }
@@ -159,12 +159,11 @@ const findTerms = (tree) => {
       .map(obj => findTerms(obj))
     );
   }
-  return undefined;
+  return [];
 };
 
 export const getFilters  = (root: Object) => {
-  let terms = R.filter(obj => obj !== undefined, findTerms(root));
-
+  let terms = findTerms(root);
   return _.map(terms, (term: Object) => ({
     'id': Object.keys(term)[0],
     'name': Object.keys(term)[0],
--- a/static/js/components/email/lib.js
+++ b/static/js/components/email/lib.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 import R from 'ramda';
+import _ from 'lodash';
 
 import {
   sendCourseTeamMail,
@@ -10,7 +11,7 @@ import {
 import { makeProfileImageUrl, mapObj } from '../../util/util';
 import type { Profile } from '../../flow/profileTypes';
 import type { Course } from '../../flow/programTypes';
-import type { EmailConfig, EmailState, Filter } from '../../flow/emailTypes';
+import type { AutomaticEmail, EmailConfig, EmailState, Filter } from '../../flow/emailTypes';
 import { actions } from '../../lib/redux_rest.js';
 import { SEARCH_FACET_FIELD_LABEL_MAP } from '../../constants';
 import { makeCountryNameTranslations } from '../LearnerSearch';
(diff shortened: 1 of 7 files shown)


aliceriot:
I don't see the recipients:

no-recipients

amir-qayyum-khan:
please pull the code, i missed one line

aliceriot:
looks great! reading code

static/js/components/email/lib.js
```diff
@@ -147,9 +148,57 @@ export const convertEmailEdit = mapObj(([k, v]) => (
   [k.match(/^subject$|^body$/) ? `email_${k}` : k.replace(/^email_/, ""), v]
 ));
 
+const findTerms = (tree, terms = []) => {
+  if (tree.hasOwnProperty("term") && !tree.term.hasOwnProperty("program.id")) {
+    return terms.push(tree.term);
+  }
+  if (R.any(_.isObject, Object.values(tree))) {
+    return terms.concat(
```
aliceriot: here we could probably just do `[].concat(`

static/js/components/email/lib.js
```diff
@@ -147,9 +148,57 @@ export const convertEmailEdit = mapObj(([k, v]) => (
...
+      Object.values(tree)
+      .filter(_.isObject)
+      .map(obj => findTerms(obj))
+    );
+  }
```
aliceriot: (to avoid the implicit `undefined` return)

amir-qayyum-khan:
Done @aliceriot

amir-qayyum-khan:
Done @aliceriot

static/js/components/email/lib_test.js
```diff
@@ -260,6 +261,43 @@ describe('Specific email config', () => {
...
+          assert.deepEqual(filters, [{
+            "id": "program.enrollments.payment_status",
+            "name": "program.enrollments.payment_status",
+            "value": "Paid"
+          }]);
+        });
+      });
+
+      describe('findTerms', () => {
+        it('should return filters', () => {
```
aliceriot: this is the same test name you have above

static/js/components/email/lib_test.js
```diff
@@ -260,6 +261,43 @@ describe('Specific email config', () => {
...
+                    "program.id":1
+                  }
+                }
+              ]
+            }
+          }`;
+          let filters = findTerms(JSON.parse(query));
+          assert.equal(filters.length, 0);
+        });
+      });
```
aliceriot: I would suggest that you try to come up with tests that more methodically cover all the possible cases for what the input to `findTerms` could look like and what the output should be - just to start it should cover these at a minimum:

```js
it('should return an empty list if passed an empty object', ...);

it('should return an empty list if passed an object without any "term" key on it', ...);

it('should return an empty list if passed an object without any nested "term" keys', ...);

it('should return a list of filters defined at the top-level', ...);

it('should return a list of filters which are nested in the object', ...);

it('should return a list of filters which are nested at different levels', ...);
```

there are probably a few more key cases to test, but yeah, I think if we write some good unit tests for the function we can have a lot of confidence that it's implemented correctly.

amir-qayyum-khan: Done @aliceriot

static/js/components/email/lib_test.js
```diff
@@ -273,16 +273,153 @@ describe('Specific email config', () => {
...
-        it('should return filters', () => {
-          let filters = findTerms(JSON.parse(queryFilters));
-          assert.equal(filters.length, 1);
-          assert.deepEqual(filters, [{
-            "program.enrollments.payment_status": "Paid"
-          }]);
+      describe('findFilters', () => {
+        it('should return an empty list if passed an empty object', () => {
+          let filters = findFilters({});
+          assert.equal(filters.length, 0);
```
aliceriot: can also do `assert.deepEqual(filters, []);` which is slightly more readable IMO

static/js/components/email/lib_test.js
```diff
@@ -273,16 +273,153 @@ describe('Specific email config', () => {
...
+                {
+                  "term": {
+                    "profile.country": "US"
+                  }
+                }
+              ]
+            }
+          }`;
+          let filters = findFilters(JSON.parse(query));
+          assert.equal(filters.length, 3);
```
aliceriot: instead of just checking the length here, could you use `assert.deepEqual` to make sure the results are exactly what we expect?

aliceriot: (and in the next few tests as well)

amir-qayyum-khan:
Done @aliceriot

PR #3240, opened 2017-05-25
Mail search now skips users without a profile

### LINKED ISSUE #3239: Missing profiles break mail on the search page
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/279915830/

noisecapella:
Fixes #3239 

#### What's this PR do?
Changes the mail search filter to skip users where `Profile` is `None`. For the linked sentry issue there were two users who had two `User` objects each and only one had a `Profile`. I'm not sure what caused this but it seems like only the one with a `Profile` is valid.

For some fake user with a valid profile, create another `User` object with the same `email`. On master you should get a 500 error when attempting to send email, and the error should not be there in this branch.

### DIFF
--- a/mail/api.py
+++ b/mail/api.py
@@ -22,6 +22,7 @@ from mail.models import (
 )
 from mail.utils import filter_recipient_variables
 from micromasters.utils import chunks
+from profiles.models import Profile
 from search.api import (
     adjust_search_for_percolator,
     search_percolate_queries,
@@ -377,15 +378,15 @@ def get_mail_vars(emails):
         generator of dict:
             A dictionary of template variables which includes email so we can tell who is who
     """
-    queryset = User.objects.filter(email__in=emails).values(
-        'email',
-        'profile__mail_id',
-        'profile__preferred_name',
+    queryset = Profile.objects.filter(user__email__in=emails).values(
+        'user__email',
+        'mail_id',
+        'preferred_name',
     ).iterator()
     return (
         {
-            'email': values['email'],
-            'mail_id': values['profile__mail_id'].hex,
-            'preferred_name': values['profile__preferred_name'],
+            'email': values['user__email'],
+            'mail_id': values['mail_id'].hex,
+            'preferred_name': values['preferred_name'],
         } for values in queryset
     )
--- a/mail/api_test.py
+++ b/mail/api_test.py
@@ -684,3 +684,12 @@ class RecipientVariablesTests(MockedESTestCase):
     def test_missing_email(self):
         """get_mail_vars should skip missing emails without erroring"""
         assert list(get_mail_vars(['missing@email.com'])) == []
+
+    def test_missing_profile(self):
+        """get_mail_vars should skip User objects without a Profile"""
+        with mute_signals(post_save):
+            profile = ProfileFactory.create()
+        user = profile.user
+        profile.delete()
+
+        assert list(get_mail_vars([user.email])) == []


mail/api.py
```diff
@@ -377,15 +378,15 @@ def get_mail_vars(emails):
...
             A dictionary of template variables which includes email so we can tell who is who
     """
-    queryset = User.objects.filter(email__in=emails).values(
-        'email',
-        'profile__mail_id',
-        'profile__preferred_name',
+    queryset = Profile.objects.filter(user__email__in=emails).values(
+        'user__email',
+        'mail_id',
+        'preferred_name',
```
giocalitri: This a pretty large query considering that it needs to join all the profiles with the users.
What about querying the users with the emails and then filtering the profiles by those users?

noisecapella: Wouldn't that just do the same join in Python? What's the advantage there?

giocalitri: yes, but you will not do the join on the table on the entire set of users: I am not sure it is really an issue here, actually

noisecapella: The query is not substantially different from before. This is the old query:

    SELECT
      "auth_user"."email",
      "profiles_profile"."mail_id",
      "profiles_profile"."preferred_name"
    FROM "auth_user"
      LEFT OUTER JOIN "profiles_profile" ON ("auth_user"."id" = "profiles_profile"."user_id")
    WHERE "auth_user"."email" IN (c, a, b)

The new query:

    SELECT
      "auth_user"."email",
      "profiles_profile"."mail_id",
      "profiles_profile"."preferred_name"
    FROM "profiles_profile"
      INNER JOIN "auth_user" ON ("profiles_profile"."user_id" = "auth_user"."id")
    WHERE "auth_user"."email" IN (c, a, b)

The `LEFT OUTER JOIN` was changed to an `INNER JOIN` which prevents the null values from the profile table from being included. The query plans are also not substantially different:

    postgres=# explain SELECT "auth_user"."email", "profiles_profile"."mail_id", "profiles_profile"."preferred_name" FROM "auth_user" LEFT OUTER JOIN "profiles_profile" ON ("auth_user"."id" = "profiles_profile"."user_id") WHERE "auth_user"."email" IN ('c', 'a', 'b');
                                      QUERY PLAN                                   
    -------------------------------------------------------------------------------
     Hash Left Join  (cost=10.45..21.30 rows=3 width=564)
       Hash Cond: (auth_user.id = profiles_profile.user_id)
       ->  Seq Scan on auth_user  (cost=0.00..10.82 rows=3 width=520)
             Filter: ((email)::text = ANY ('{c,a,b}'::text[]))
       ->  Hash  (cost=10.20..10.20 rows=20 width=52)
             ->  Seq Scan on profiles_profile  (cost=0.00..10.20 rows=20 width=52)
    (6 rows)
    
    postgres=# explain SELECT "auth_user"."email", "profiles_profile"."mail_id", "profiles_profile"."preferred_name" FROM "profiles_profile" INNER JOIN "auth_user" ON ("profiles_profile"."user_id" = "auth_user"."id") WHERE "auth_user"."email" IN ('c', 'a', 'b');
                                      QUERY PLAN                                   
    -------------------------------------------------------------------------------
     Hash Join  (cost=10.45..21.30 rows=1 width=564)
       Hash Cond: (auth_user.id = profiles_profile.user_id)
       ->  Seq Scan on auth_user  (cost=0.00..10.82 rows=3 width=520)
             Filter: ((email)::text = ANY ('{c,a,b}'::text[]))
       ->  Hash  (cost=10.20..10.20 rows=20 width=52)
             ->  Seq Scan on profiles_profile  (cost=0.00..10.20 rows=20 width=52)
    (6 rows)

giocalitri: ok, this is enough for me

giocalitri:
I'll manually test this now

giocalitri:
works as expected

giocalitri:
👍

PR #3247, opened 2017-05-26
Handle reuse_db option for selenium tests

### LINKED ISSUE #3030: Use keepdb flag to speed up selenium tests
#2674 will speed up the time between selenium tests, but there is still the initial 30 second time to initialize the database before the first test. We should check for `keepdb` and use the existing test database if it's set. This will make running selenium tests much faster on dev environments.

noisecapella:
Fixes #3030 

#### What's this PR do?
Adds support for `keepdb` option to provide quicker running of local selenium tests.

Remove `--reuse-db` from `pytest.ini` and run the selenium tests. Then add it back and run the selenium tests again. The second time should be considerably faster, roughly 20 seconds.

There are also other minor changes meant to fix flaky tests

### DIFF
--- a/selenium_tests/base.py
+++ b/selenium_tests/base.py
@@ -21,7 +21,12 @@ from django.core.management import call_command
 from django.db import connection
 from django.db.models.signals import post_save
 from factory.django import mute_signals
-from selenium.common.exceptions import StaleElementReferenceException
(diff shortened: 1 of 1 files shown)


mbertrand:
Looks good to me! 👍

PR #3250, opened 2017-05-26
Add script to use webpack dev server to serve javascript bundles

### LINKED ISSUE #3232: Investigate running selenium tests against webpack dev server in dev environment
This would be useful in letting the user skip the webpack step locally and iterate more quickly when working with JS code

noisecapella:
Fixes #3232 

#### What's this PR do?
Adds a new script `scripts/test/run_selenium_tests_dev.sh` which runs the selenium tests using an existing webpack dev server instance. The user is required to set an environment variable `WEBPACK_DEV_SERVER_HOST` to tell selenium where to find the webpack dev server. The restriction that `DEBUG` is true is removed because tests will always have `DEBUG == False` and there is already `USE_WEBPACK_DEV_SERVER`

Set the environment variable `USE_WEBPACK_DEV_SERVER` to the IP address of the dev server (don't include the port number). This should generally be the same as whatever IP address you use to access your local dev environment (I think). Then run `./scripts/tests/run_selenium_tests_dev.sh`. All tests should pass.

Go to `CourseAction` and remove or modify the `enroll-button` CSS class from the `SpinnerButton`. Run the tests again and the `test_zero_price_purchase` test should fail.


aliceriot:
I can't seem to get the tests to pass. I did this:

```
export USE_WEBPACK_DEV_SERVER=0.0.0.0
./scripts/test/run_selenium_tests_dev.sh
```

I get a failed test that looks like this:
test_zero_price_purchase

so it looks like the selenium thingy isn't able to get the static asserts (I can also see a CONNECTION REFUSED error in the test output)

aliceriot:
oh, haha, that explains it!

aliceriot:
functionality looks good!

scripts/test/run_selenium_tests_dev.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Please start the webpack dev server before running this script."
+    exit 1
+fi
+
+# TODO: can we detect this somehow?
+if [[ -z "$WEBPACK_DEV_SERVER_HOST" ]]
+then
+    echo "Please set WEBPACK_DEV_SERVER_HOST to the IP address of your webpack dev server, omitting the port number."
+    exit 1
+fi
```
aliceriot: it would be cool if we could call `docker inspect`, parse the JSON, and figure out the IP automatically - do you think that would be too complicated? I imagine it would be tricky to deal with the Mac environment as well.

noisecapella: Yeah, it doesn't look trivial to set up for OS X. I'd like to leave automatically detecting the IP address for another PR

aliceriot: sounds good

PR #3255, opened 2017-05-31
rewrite API fetch wrappers to use async/await syntax

aliceriot:

no ticket, I was just messing around with this and I think it's a bit cleaner.

#### What's this PR do?

this rewrites `_fetchJSONWithCSRF` and `_fetchWithCSRF` in `lib/api.js` to be functions declared with the `async` keyword. this lets us, inside the function, write asynchronous code using promises in a straightforward, synchronous-looking syntax (via the `await` keyword).

Basically, instead of something like:

```js
fetch(url, options).then(response {
 // response handling code
})
```

we can do something like:

```js
let response = await fetch(url, options);
// response handling code
```

which is nice!

the async/await stuff will all get transpiled away, so there should be no change in functionality. all tests should pass without modification.

### DIFF
--- a/static/js/lib/api.js
+++ b/static/js/lib/api.js
@@ -67,18 +67,14 @@ const formatRequest = R.compose(
 
 const formatJSONRequest = R.compose(formatRequest, jsonHeaders);
 
-const _fetchWithCSRF = (path: string, init: Object = {}): Promise<*> => {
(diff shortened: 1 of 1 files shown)


mbertrand:
Looks good to me! :thumbsup:

Issue #2384, opened 2017-01-19
Replace city/state/country fields with geosuggest field for education & employment


singingwolfboy:
Follow-up from #2135. We implemented a geosuggest component with Google Places API integration in #2167, but removed the UI in #2305. It would be nice to use this component in places where we really only care about city, state, and country, such as the education form and employment form. This change was reverted for the personal form because, in order to register users for a Pearson exam, we need to know their address all the way down to street address and apartment number. However, we don't need this precision for education and employment.

Here's a mockup of what this might look like:
<img width="748" alt="school typeahead" src="https://cloud.githubusercontent.com/assets/132355/22112459/ffa54f96-de30-11e6-860a-d59800f3ff7d.png">

singingwolfboy:
Even though this feature is an enhancement, I added the "tech debt" label because it _also_ resolves the debt of having unused code in our codebase.

pdpinch:
We tried to make this work with our existing data model, but Google's API and our database of states/territories weren't compatible.

PR #3270, opened 2017-06-02
Use Geosuggest React component for Profile education and employment forms

mbertrand:
https://github.com/mitodl/micromasters/issues/2384

#### What's this PR do?
Replaces the city, state, and country input fields for the profile education and employment dialogs with a Geosuggest component input field.

Add and edit employment and education entries on your profile.  Start typing in a city (like 'Boston'), you should see some suggestions start to appear below the input field. Choose one and save.  Then go back to edit the entry and make sure that the value you selected is there.

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "fetch-mock": "^5.11.0",
     "file-loader": "^0.11.1",
     "flow-bin": "^0.43.1",
+    "fuse-js-latest": "^2.6.2",
     "history": "^4.6.0",
     "iflow-lodash": "^1.1.16",
     "iflow-react-router": "^1.1.16",
--- a/static/js/lib/location.js
+++ b/static/js/lib/location.js
@@ -1,6 +1,63 @@
 import R from 'ramda';
 import iso3166 from 'iso-3166-2';
+import Fuse from 'fuse-js-latest';
 
-export const codeToCountryName = (code: string) =>  R.pathOr(
-  "", ['name'], iso3166.country(R.defaultTo("", code))
+const notAvailable = "Not Available";
+
+/**
+ * Gets the name of a country given its code
+ * @param code Country code
+ */
+export const codeToCountryName =  R.compose(
+  R.pathOr("", ['name']),
+  iso3166.country,
+  R.defaultTo("")
 );
+
+/**
+ * Gets the code for a country given its name
+ * @param name Country name
+ */
+export const nameToCountryCode = R.compose(
+  R.pathOr("", ['code']),
+  iso3166.country,
+  R.defaultTo("")
+);
+
+/**
+ * Gets the name of a state/territory given its ISO3166 code (as in 'US-MA')
+ * @param code ISO3166 code ('US-MA', 'US-CO', etc).
+ */
+export const codeToStateName = R.compose(
+  R.pathOr(notAvailable, ['name']),
+  iso3166.subdivision,
+  R.defaultTo("")
+);
+
+/**
+ * Get a list of subregion objects for a country, each with a 'code' and 'name' property.
+ * @param country Country code
+ */
+export const getSubcodes = (country:string) => (
+  R.map(function(key) {
+    return {"code": key, "name": iso3166.data[country]["sub"][key]["name"]};
+  }, Object.keys(iso3166.data[country]["sub"]))
+);
+
+/**
+ * Gets the ISO3166 code for a state/territory given a country and state name.
+ * If no exact match is found for a state name, try a fuzzy match.
+ * Return a notAvailable value if no match can be found either way.
+ * @param country Country code
+ * @param name State/territory name
+ */
+export const nameToStateCode = (country: string, name: string) =>  {
+  let statecode = iso3166.subdivision(R.defaultTo("", country), R.defaultTo("", name));
+  if (!R.isEmpty(statecode)) {
+    return statecode['code'];
+  } else {
+    const fuse = new Fuse(getSubcodes(country), {keys: ["name"], threshold:0.4, tokenize: true});
+    const fuzzymatches = fuse.search(R.defaultTo("", name));
+    return R.isEmpty(fuzzymatches) ? notAvailable : fuzzymatches[0]["code"];
+  }
+};
--- a/static/js/lib/validation/profile.js
+++ b/static/js/lib/validation/profile.js
@@ -237,6 +237,13 @@ const nestedValidator = R.curry((key: string, findErrors: Function, profile: Pro
   return { [key]: errors };
 });
 
+const mergeListOfArgs = R.compose(R.mergeAll, Array);
+
+const extraErrorCheck = R.curry((key, msg, predicate, entry, errors) => (
+  predicate(entry) ? R.merge(errors, { [key]: msg }) : errors
+));
+
+
 /*
  * Education Validation
  */
@@ -245,10 +252,7 @@ const educationMessages: ErrorMessages = {
   'graduation_date': 'Please enter a valid graduation date',
   'field_of_study': 'Field of study is required',
   'online_degree': 'Online Degree is required',
-  'school_name': 'School name is required',
-  'school_city': 'City is required',
-  'school_state_or_territory': 'State is required',
-  'school_country': 'Country is required'
+  'school_name': 'School name is required'
 };
 
 const isHighSchool: (e: EducationEntry) => boolean = R.compose(
@@ -263,8 +267,16 @@ const educationKeys: (e: EducationEntry) => string[] = R.ifElse(
   isHighSchool, R.compose(excludeFieldOfStudy, R.keys), R.keys
 );
 
-const educationErrors: (es: EducationEntry[]) => ValidationErrors[] = R.map(entry => (
-  findErrors(entry, educationKeys(entry), educationMessages)
+const schoolLocationIsValid = extraErrorCheck('location', 'Location must contain city, state, country', entry => (
+  isNilOrEmptyString(entry.school_city) ||
+  isNilOrEmptyString(entry.school_state_or_territory) ||
+  isNilOrEmptyString(entry.school_country)
+));
+
+const additionalSchoolValidation = R.converge(mergeListOfArgs, [schoolLocationIsValid]);
+
+const educationErrors: (xs: EducationEntry[]) => ValidationErrors[] = R.map(entry => (
+  additionalSchoolValidation(entry, findErrors(entry, educationKeys(entry), educationMessages))
 ));
 
 export const educationValidation = nestedValidator('education', educationErrors);
@@ -277,9 +289,6 @@ const workMessages: ErrorMessages = {
   'industry': 'Industry is required',
   'company_name': 'Name of Employer is required',
   'start_date': 'Please enter a valid start date',
-  'city': 'City is required',
-  'country': 'Country is required',
-  'state_or_territory': 'State or Territory is required',
 };
 
 // functions to perform extra checks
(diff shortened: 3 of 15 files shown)


aliceriot:
With already existing entries it does a slightly strange thing:

sccjjcjcjc

It looks like it's taken the first alphabetical country code (AD: Andorra) and the first state code in Andorra (AD-05) since I didn't already have values for those saved. Would it be straightforward to use a blank value until the user edits the field? If it's non-trivial I think this is OK, but optimally we wouldn't assume they live in Andorra! haha.

mbertrand:
@aliceriot thanks for catching that, I think it should be doable to use a blank value until edited, will work on it.  Andorra does look like a nice place to live though :)

aliceriot:
another functionality thing: if I go to add another education entry to my profile there is a placeholder value (Anytown, Massachusetts, USA). If I don't edit the field and try to save it doesn't let me (there is a validation error, visible in the console in the screenshot) but the error isn't surfaced to the user in the UI, so the user would have to guess why they aren't allowed to save the entry:

validationstuff

aliceriot:
Hmm, so I wasn't able to repro what I was seeing before exactly. There is a slight bit of strangeness relating to the fact that we previously stored country and state/territory codes, but are now storing the country name and then the territory name - is it possible to configure the geosuggest component to keep the old behavior? I'm thinking there may be places where we assume the data will be normalized codes but I'm not exactly sure.

@noisecapella do you think there would be any issues with elasticsearch?

noisecapella:
It looks like the exam code uses the country codes so I'm not sure we can make that change. Can we translate from country name to code in the JS?

mbertrand:
@aliceriot I pushed some changes that should fix the lack of a validation message.  Will look into configuring the component to use state/country codes instead of full names.

aliceriot:
there's a function called `codeToCountryName` in `static/js/lib/location` which may be helpful!

mbertrand:
@aliceriot @noisecapella I modified the geosuggest code to save state & country values as codes ('US-MA', 'US') as they were before, but display their full names to the user, using the iso-3166-2 javascript library and some helper functions in [../lib/location.js](https://github.com/mitodl/micromasters/blob/1c6f43cab943c031603dba3ffc92deb11cc38047/static/js/lib/location.js)

aliceriot:
cool, I'll take a look now!

aliceriot:
it seems like maybe some locations don't have states? not sure exactly what's up, but I can't set Kandahar, Afghanistan as my location:

somethings_dont_have_states

I just typed Kandahar in and selected the first entry

aliceriot:
It looks like the functionality is working other than that though!

aliceriot:
hm I found another strange thing. I typed in `Paris` and the box offered the suggestion `Paris, Île-de-France, France`. I selected that and saved the entry, and then went to edit it. When editing, I typed in `Tokyo` and selected the first option (`Tokyo, Japan`). Somehow that resulted in this:
uhhhh_what

aliceriot:
the suggestions are also cut off:

z-index-or-something

maybe we can do some creative z-index fiddling to get them to show up at the bottom? I've had some trouble in the past with menu things in the dialog boxes, atm I don't remember *exactly* how I fixed it though.

mbertrand:
@aliceriot, @noisecapella  Ugh, it looks like the iso3166 library does not have a corresponding value for every state/territory name returned by Google Maps API, 'Kandahar' being one of them:

```
subdivision('AF', 'Kandahar') -> Object{}
```
but:
```
subdivision('AF', 'Kandahār') -> Object {type: "Province", name: "Kandahār", countryName: "Afghanistan", countryCode: "AF", code: "AF-KAN"…}
```

aliceriot:
I'm guesing the accent in Île-de-France is causing the same issue

aliceriot:
Maybe there's a way we can have google maps deal with converting to the country code, so we don't have to worry about it? I don't know if there's some setting we can pass to the react-geosuggest component to have it spit it country codes instead of names.

mbertrand:
@aliceriot @noisecapella Google returns only 2 attributes for each element of a suggestion: short_name and long_name.  For country, short_name seems to be synonymous with the ISO3166 code.  However, for state/territory, short_name sometimes is equivalent to the last half of the ISO3166 code and sometimes not, for example:

```
Object {long_name: "Massachusetts", short_name: "MA", types: Array(2)} //iso3166 is 'US-MA'
Object {long_name: "Kandahar", short_name: "Kandahar", types: Array(2)} //iso3166 is 'AF-KAN'
```

That led me to use the iso-3166-2 javascript library, but unfortunately it doesn't handle alternate name spellings (like Kandahar vs Qandahar vs Kandahār).  So unless the state/territory name in the Google Maps response is an exact match with the [data used by the iso-3166-2 library](https://github.com/olahol/iso-3166-2.js/blob/master/data.csv), no state code will be returned.

I might have to find a complete list of all possible state/territory names with their corresponding iso3166 codes.

aliceriot:
the fuzzy search sounds like it might be the best idea going, provided we only use that as a fallback in the case where the returned value isn't an ISO3166 code and isn't a direct match to any of the data in the iso-3166-2 library.

mbertrand:
The state/territory code should now be determined based on an exact name match if possible, and if that fails then by a fuzzy match, and if that fails then a default of 'Not Available'.  Added some more tests too.  Also took a shot at modifying the css so that the auto-suggestions are not hidden (by making the suggestions appear above rather than below the field).

aliceriot:
the functionality looks good! In particular all of the entries that tripped it up before seem to be handled now. Do you have an example of a location which results in 'Not available'? I'd like to just confirm that works ok, if possible.

static/js/lib/validation/profile.js
```diff
@@ -263,8 +267,18 @@ const educationKeys: (e: EducationEntry) => string[] = R.ifElse(
...
+  isNilOrEmptyString(entry.school_country)
+));
+
+const additionalSchoolValidation = R.converge(mergeListOfArgs, [
+  schoolLocationIsValid
+]);
+
+const educationErrors: (xs: EducationEntry[]) => ValidationErrors[] = R.map(entry => (
+  additionalSchoolValidation(entry, findErrors(entry, educationKeys(entry), educationMessages))
 ));
```
aliceriot: could we just define `educationErrors` here using `R.converge`? Something like

```js
const educationErrors = R.converge(
  mergeListOfArgs,
  [
    schoolLocationIsValid,
    R.map(entry => findErrors(entry, educationKeys(entry), educationMessages))
  ]
);
```

might be slightly simpler that way, although the `schoolLocationIsValid` function would have to just return an empty object when there were no errors.

static/js/util/profile_edit.js
```diff
@@ -7,8 +7,8 @@ import Checkbox from 'material-ui/Checkbox';
 import Geosuggest from 'react-geosuggest';
 import R from 'ramda';
 import ReactTelInput from 'react-telephone-input';
-
 import DateField from '../components/inputs/DateField';
```
aliceriot: we normally leave one blank line between outside dependencies and our own modules in the imports at the top of a JS file (sorry, this stuff isn't written down anywhere)

mbertrand:
For 'Not Available', try typing "Sana'a" and then pick "Sana'a, Capital Municipality, Yemen" from the suggestions.

aliceriot:
works as advertised!

static/js/util/profile_edit.js
```diff
@@ -299,9 +317,25 @@ export function boundGeosuggest(
...
   ]);
 
+
+  //Retrieve an profile's property value given a property path.
+  const getPath = (addressKey) => (R.path(addressKey, profile));
+
+  // Get all the values of a location (city, state, country) as an array.
+  // Convert state and country elements to ISO3166 codes
+  const getLocationElements = () => {
+    let elements = R.map(getPath, R.values(addressMapping));
```
aliceriot: we could replace `getPath` here with this:

```js
R.path(R.__, profile)
```

which is just a little more compact (`R.__` is like an argument placeholder, so that function already has `profile` bound as its second argument and is waiting for its first)

static/js/lib/location.js
```diff
@@ -1,6 +1,50 @@
...
   "", ['name'], iso3166.country(R.defaultTo("", code))
 );
+
+/**
+ * Gets the code for a country given its name
+ * @param name Country name
+ */
+export const nameToCountryCode = (name: string) =>  R.pathOr(
+  "", ['code'], iso3166.country(R.defaultTo("", name))
+);
```
aliceriot: this could be written point-free if you want:

```js
R.compose(
  R.pathOr("", ['code']),
  iso3166.country,
  R.defaultTo("")
)

File: static/js/lib/location.js [hunk 5]
```diff
@@ -1,6 +1,50 @@
...
 
+const notAvailable = "Not Available";
+
+/**
+ * Gets the name of a country given its code
+ * @param code Country code
+ */
 export const codeToCountryName = (code: string) =>  R.pathOr(
   "", ['name'], iso3166.country(R.defaultTo("", code))
 );
```
aliceriot: so could this one! haha

package.json
```diff
@@ -53,6 +53,7 @@
     "fetch-mock": "^5.11.0",
     "file-loader": "^0.11.1",
     "flow-bin": "^0.43.1",
+    "fuse-js-latest": "^2.6.2",
```
aliceriot: I don't see changes in `yarn.lock`, did you run `yarn add fuse-js-latest`?

static/js/data/region_codes.js
```diff
@@ -0,0 +1,5822 @@
+export default {
```
mbertrand: Yes, but transformed from csv to json, and grouped by country, so that fusejs can do a fuzzy search on the state/territory names for a country

static/js/lib/location.js
mbertrand: will do!

static/js/lib/location.js
```diff
@@ -1,6 +1,50 @@
...
+ * Return a notAvailable value if no match can be found either way.
+ * @param country Country code or nameempty
+ * @param name State/territory name
+ */
+export const nameToStateCode = (country: string, name: string) =>  {
+  let statecode = iso3166.subdivision(R.defaultTo("", country), R.defaultTo("", name));
+  if (!R.isEmpty(statecode)) {
+    return statecode['code'];
+  } else {
+    let fuse = new Fuse(isocodes[R.defaultTo("", country)]["regions"], {
```
mbertrand: I'll try that out and see

aliceriot:
hm, I can't seem to reply to those comments for some reason, but yeah we could def talk about that tomorrow!

It looks like the yarn change is sorted too. I'll take a look over the code again!

static/js/lib/location.js
```diff
@@ -1,6 +1,56 @@
...
+ * @param name State/territory name
+ */
+export const nameToStateCode = (country: string, name: string) =>  {
+  let statecode = iso3166.subdivision(R.defaultTo("", country), R.defaultTo("", name));
+  if (!R.isEmpty(statecode)) {
+    return statecode['code'];
+  } else {
+    let subcodes = R.map(function(key) {
+      return {"code": key, "name": iso3166.data[country]["sub"][key]["name"]};
+    }, Object.keys(iso3166.data[country]["sub"]));
```
aliceriot: it looks like this `subcodes` value only depends on the `country` - what do you link about pulling it out into a little helper function (`getSubcodes` or something) that takes a `country` and returns the subcodes? just for a little readability win.

static/js/lib/location.js
```diff
@@ -46,10 +46,11 @@ export const nameToStateCode = (country: string, name: string) =>  {
...
   } else {
-    let subcodes = R.map(function(key) {
-      return {"code": key, "name": iso3166.data[country]["sub"][key]["name"]};
-    }, Object.keys(iso3166.data[country]["sub"]));
-    let fuse = new Fuse(subcodes, {keys: ["name"], threshold:0.4, tokenize: true});
+    const getSubcodes = (key: string) => (
+      {"code": key, "name": iso3166.data[country]["sub"][key]["name"]}
+    );
+    const subcodes = R.map(getSubcodes, Object.keys(iso3166.data[country]["sub"]));
+    const fuse = new Fuse(subcodes, {keys: ["name"], threshold:0.4, tokenize: true});
```
aliceriot: sorry, I wasn't clear! I was proposing moving it out of the `nameToStateCode` function entirely, something like:

```js
const getSubCodes = country => R.map(
  key => ({ code: key, name: iso3166.data[country].sub.key["name"] }),
  Object.keys(iso3166.data[country].sub)
);
```

Then when you're creating your `Fuse` object you could just call `getSubcodes(country)` inline.

static/js/lib/location.js
mbertrand: Sure, good idea, done

static/js/lib/validation/profile.js
aliceriot: let's just not worry about this :)

mbertrand:
@aliceriot Thanks for the review!

PR #3278, opened 2017-06-06
Use database templates for faster database restore during selenium tests

### LINKED ISSUE #3273: An errored selenium test leaves database in an improper state
This is easily worked around by removing `--reuse-db` from `pytest.ini` and running the tests again, but we should change the error handling to ensure that the original test database is restored no matter what error happens

noisecapella:
Fixes #3273 

#### What's this PR do?
 - Uses database templates for faster database restore. It previously dumped the database to SQL and reimported the SQL. This makes the `snapshot_dashboard_states` management command go roughly twice as fast.
 - Refactors the selenium tests to allow separate test cases
 - Adds a new function `setUpTestData` named after a similar function in `TestCase` which lets the user do some of the database setup per test case instead of per test
 - Fixes error handling for selenium test to leave the test database in a proper state on failure or exceptions
 - Adds `--suffix` to `snapshot_dashboard_states` to let the user generate screenshots for only a certain scenario
 - Adds `--list-scenarios` to `snapshot_dashboard_states` to list the scenarios which would be generated
 - I also made minor tweaks to the script which runs selenium tests to allow the use of `WEBPACK_DEV_SERVER_HOST`  in `.env`

Run the `snapshot_dashboard_states` management command and verify that it takes less than three minutes, roughly.


noisecapella:
I'm also working on a couple problems @rhysyngsun pointed out:
 - `WEBPACK_DEV_SERVER_HOST` is also used in the script to start the webpack dev server and it will fail to start the server if this is defined. I'll look for a workaround here
- `self.selenium` might be `None` when `tearDown` executes

rhysyngsun:
Functionality works as expected.

micromasters/utils.py
```diff
@@ -230,3 +230,20 @@ def now_in_utc():
...
+
+    Args:
+        env_string (str): Output from running env in a shell, converted to UTF-8
+
+    Returns:
+        dict: A dict of key value environment variable pairs (these are bytes)
+    """
+    lines = env_string.split("\n")
+    # Remove empty lines
+    lines = [line for line in lines if line]
```
rhysyngsun: Nit: could just inline the `if line` into the generator on the next line.

noisecapella:
OS X stuff works so I'm going to merge this when tests pass

PR #3279, opened 2017-06-06
Added check for exam attempts

### LINKED ISSUE #2286: Exams: restrict learners to 2 exam attempts

rhysyngsun:
Fixes #2286 

#### What's this PR do?
Adds a check to not generate more authorizations if a user has reached their limit of exam attempts for how many times they've paid.

The tests added should pass, otherwise you can test by:

For an FA course:
- Create an `ExamRun` for the course with all dates in the past.
- Create two `ExamRunAuthorization`s for the past `ExamRun`
- Create a second `ExamRun` for the course with `date_first_eligible` in the past and `date_last_eligible` in the future.
- Enroll in the current course run for that course
- Create a passing `FinalGrade` for the current course run of that course
- Verify you have no `ExamAuthorization` record for that course
- Pay for the course
- Verify do not have a third `ExamAuthorization` record for that course

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -206,6 +206,19 @@ class MMTrack:
             return self.has_final_grade_paid_on_edx(edx_course_key)
         return self.has_verified_enrollment(edx_course_key)
 
+    def get_course_paid_count(self, course_key):
+        """
+        Gets the count of paid course runs for the course
+
+        Args:
+            edx_course_key (str): an edX course run key
+        Returns:
+            int: count of paid course runs
+        """
+        return Line.objects.filter(
+            order__status=Order.FULFILLED, order__user=self.user, course_key=course_key
+        ).values('order_id').distinct().count()
+
     def has_paid_for_any_in_program(self):
         """
         Returns true if a user has paid for any course run in the program
--- a/exams/api_test.py
+++ b/exams/api_test.py
@@ -27,6 +27,7 @@ from exams.api import (
     sso_digest,
     MESSAGE_NOT_ELIGIBLE_TEMPLATE,
     MESSAGE_NOT_PASSED_OR_EXIST_TEMPLATE,
+    ATTEMPTS_PER_PAID_RUN,
 )
 from exams.exceptions import ExamAuthorizationException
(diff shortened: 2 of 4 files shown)


pdpinch:
Just to be clear, this PR only prevents creation of authorizations after the 2nd exam attempt. It doesn't include any UI that communicates this to the learner, right?

rhysyngsun:
@pdpinch to be precise, it prevents authorizations after N * 2 attempts where N = # of paid course runs for that course. No UI.

exams/api_test.py
```diff
@@ -130,6 +130,37 @@ def test_exam_authorization(self):
...
+        """
+        test exam_authorization when user passed and paid, but used all their attempts
+        """
+        create_order(self.user, self.course_run)
+        mmtrack = get_mmtrack(self.user, self.program)
+        self.assertTrue(mmtrack.has_paid(self.course_run.edx_course_key))
+        self.assertTrue(mmtrack.has_passed_course(self.course_run.edx_course_key))
+        old_run = ExamRunFactory.create(course=self.course_run.course)
+        ExamAuthorizationFactory.create_batch(
+            2,
```
mbertrand: Maybe replace hard-coded 2 in tests with 'ATTEMPTS_PER_PAID_RUN', just in case it changes in the future?

dashboard/utils.py
```diff
@@ -206,6 +206,19 @@ def has_paid(self, edx_course_key):
             return self.has_final_grade_paid_on_edx(edx_course_key)
         return self.has_verified_enrollment(edx_course_key)
 
+    def get_course_paid_count(self, course_key):
+        """
+        Gets the count of paid course runs for the course
```
rhysyngsun: Yes this is the standard style, we roughly follow this: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

mbertrand:
Looks great, 👍

Issue #3263, opened 2017-06-01
Incorrect coupon message for course-level coupon


noisecapella:
The program-level message should show that all courses cost $1000 each, but it's using the $50 of the fixed price coupon, which is incorrect since that coupon is on the course level

![screenshot from 2017-06-01 11-47-20](https://cloud.githubusercontent.com/assets/863262/26688251/22afd8ba-46c0-11e7-8ebf-60a08f5e4104.png)

amir-qayyum-khan:
Not able to produce

noisecapella:
Can you show what your `Coupon` object looks like? Is there a `UserCoupon` object connecting that coupon to your logged in user?

amir-qayyum-khan:
<img width="1066" alt="screen shot 2017-06-06 at 11 18 56 pm" src="https://user-images.githubusercontent.com/10431250/26844983-9e62e252-4b0e-11e7-9e62-32926d845a04.png">

<img width="899" alt="screen shot 2017-06-06 at 11 21 20 pm" src="https://user-images.githubusercontent.com/10431250/26845085-e69d1196-4b0e-11e7-9472-a159fbbfca95.png">

#### couse id == object id == 1, am I right. i picked from url.
<img width="917" alt="screen shot 2017-06-06 at 11 22 10 pm" src="https://user-images.githubusercontent.com/10431250/26845188-1dd98a5e-4b0f-11e7-91f3-1b63b837f78f.png">

amir-qayyum-khan:
produced issue, The issue was coupon type. I made it `standard` and issue produced

noisecapella:
It looks like this didn't fix all problems, see #3342 for example. Reopening just to make sure we don't release without fixing this first.

PR #3281, opened 2017-06-07
Fixed message for course-level coupon

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3263

#### What's this PR do?
- calculates regular prices for course and show it in coupon message

create coupon and bind it with user, then see the message on dashboard

@pdpinch 
screen shot 2017-06-07 at 5 16 35 pm

### DIFF
--- a/static/js/components/dashboard/CourseListCard.js
+++ b/static/js/components/dashboard/CourseListCard.js
@@ -3,6 +3,7 @@ import React from 'react';
 import moment from 'moment';
 import R from 'ramda';
 import { Card, CardTitle } from 'react-mdl/lib/Card';
+import Decimal from 'decimal.js-light';
 
 import type { Program, Course } from '../../flow/programTypes';
 import type {
@@ -47,6 +48,26 @@ export default class CourseListCard extends React.Component {
     checkout:                        (s: string) => void,
   };
 
+  getRegularPrice = (program: Program, prices: CalculatedPrices): Decimal|null => {
+    let price;
+    let list = [];
+
+    for (const courseRun of sortedCourseRuns(program)) {
+      price = prices.get(courseRun.id);
+      if (price) {
+        list.push(price);
+      }
+    }
+    // retruning max price, assuming that max price is regular price otherwise discounted.
+    // what happen when user has coupans for all course in program for COUPON_CONTENT_TYPE_COURSE.
+    // should we show max price here?
+    if (list.length > 0) {
+      return R.sort((left, right) => right - left, list)[0];
+    }
+
+    return null;
+  }
+
   renderFinancialAidPriceMessage(): ?React$Element<*> {
     const {
       program,
@@ -56,13 +77,7 @@ export default class CourseListCard extends React.Component {
     const finAidStatus = program.financial_aid_user_info.application_status;
 
     if (FA_TERMINAL_STATUSES.includes(finAidStatus)) {
-      let price;
-      for (const courseRun of sortedCourseRuns(program)) {
-        price = prices.get(courseRun.id);
-        if (price) {
-          break;
-        }
-      }
+      let price = this.getRegularPrice(program, prices);
 
       if (!price) {
         return null;
@@ -145,13 +160,7 @@ export default class CourseListCard extends React.Component {
       return this.renderCouponPriceMessage();
     }
 
-    let price;
-    for (const courseRun of sortedCourseRuns(program)) {
-      price = prices.get(courseRun.id);
-      if (price) {
-        break;
-      }
-    }
+    let price = this.getRegularPrice(program, prices);
     if (!price) {
       return null;
     }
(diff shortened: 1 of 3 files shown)


static/js/components/dashboard/CourseListCard.js
```diff
@@ -45,6 +46,26 @@ export default class CourseListCard extends React.Component {
...
+    let list = [];
+
+    for (const courseRun of sortedCourseRuns(program)) {
+      price = prices.get(courseRun.id);
+      if (price) {
+        list.push(price);
+      }
+    }
+    // retruning max price, assuming that max price is regular price otherwise discounted.
+    // what happen when user has coupans for all course in program for COUPON_CONTENT_TYPE_COURSE.
```
amir-qayyum-khan: @noisecapella

noisecapella: The way that coupons are structured currently, there should never be more than one coupon per program, even if they are for a course.

static/js/components/dashboard/CourseListCard_test.js
```diff
@@ -144,6 +144,37 @@ describe('CourseListCard', () => {
       );
     });
 
+    it('should show regular course price on coupon message', () => {
+      const coupon = makeCourseCoupon(program["courses"][0].id, program.id);
```
noisecapella: Can you change the syntax here to `program.courses[0].id`?

static/js/factories/dashboard.js
```diff
@@ -129,6 +129,16 @@ export const makeCoupon = (program: Program): Coupon => ({
   object_id: program.id,
 });
 
+export const makeCourseCoupon = (courseId: number, programId: number): Coupon => ({
```
noisecapella: Can you change `courseId` and `programId` to `course` and `program` for consistency with `makeCoupon`?

amir-qayyum-khan:
> This doesn't fix the problem at all. CourseListCard has a prop prices which has the coupon information already calculated.

@noisecapella  i dont want to change course price there, if discounted price of a course is $50 then let it be.
Here i change the display for coupon message to show regular price instead of first run price

noisecapella:
If you run the `./manage.py snapshot_dashboard_states` command there are some screenshots which illustrate the problem. Here are the results from when I ran the command on your branch: 

[dashboard.tar.gz](https://github.com/mitodl/micromasters/files/1061256/dashboard.tar.gz)

The files with `coupon` in the name relate to various coupon states. I'm changing the filenames in a future PR to be more descriptive but there are two boolean values and the first one is `True` if the coupon is a program coupon and `False` if the coupon is a course coupon. (This is more clear if you look at the source code to the management command.) It shows that the problem still exists in this PR

amir-qayyum-khan:
i am not able to understand it @noisecapella can you guid me what is that problem here in my branch that we want to solve?

pdpinch:
It looks to me like the screenshot `dashboard_state_032_coupon_percent-discount_False_True.png` is incorrect. 

image

It's a little hard to tell for me though because the text file `dashboard_state_032_coupon_percent-discount_False_True.txt` doesn't include the output of the coupons and course_prices APIs. Is that hard to add?

noisecapella:
@pdpinch No, I can look into that

amir-qayyum-khan:
Done @noisecapella 

dashboard_state_026_coupon_fixed-price_course_not-free

dashboard_state_028_coupon_fixed-discount_course_not-free

dashboard_state_030_coupon_percent-discount_course_not-free

dashboard_state_032_coupon_percent-discount_course_free

static/js/components/dashboard/CourseListCard.js
```diff
@@ -47,22 +48,21 @@ export default class CourseListCard extends React.Component {
     checkout:                        (s: string) => void,
   };
 
+  getRegularPrice = (coursePrice?: CoursePrice): Decimal|null => {
+    // assuming all courses have same price
+    return coursePrice ? coursePrice.price : null;
```
amir-qayyum-khan: https://github.com/mitodl/micromasters/blob/master/static/js/containers/DashboardPage.js#L676

noisecapella:
It looks like the functionality is fixed now, looking at code and tests...

static/js/components/dashboard/CourseListCard.js
```diff
@@ -47,22 +48,20 @@ export default class CourseListCard extends React.Component {
...
 
     if (FA_TERMINAL_STATUSES.includes(finAidStatus)) {
-      let price;
-      for (const courseRun of sortedCourseRuns(program)) {
-        price = prices.get(courseRun.id);
-        if (price) {
-          break;
-        }
-      }
+      let price = this.getRegularPrice(coursePrice);
```
noisecapella: Can you replace `this.getRegularPrice(coursePrice)` with `R.pathOr(null, ['price'], coursePrice)` here and below?

PR #3283, opened 2017-06-07
Use is_passing in MMTrack

### LINKED ISSUE #2539: modify MMTrack to use the passed flag returned from edx-api-client
part of #2538 
depends on mitodl/edx-api-client#30

Once we've added a `passed` static method to Certificates in edx-api-client, we need to modify MMTrack to use it instead of the hard-coded passed statuses introduced in #2537.

annagav:
Fix #2539

#### What's this PR do?
Change MMTrack `has_passing_certificate` to use `is_passing` property of the certificate coming from edx.

should be covered in dashboard/utils_test.py

### DIFF
--- a/dashboard/utils_test.py
+++ b/dashboard/utils_test.py
@@ -557,14 +557,12 @@ class MMTrackTest(MockedESTestCase):
         assert mmtrack.has_paid_for_any_in_program() is True
 
     @ddt.data(
-        ("verified", "downloadable", True, True),
-        ("audit", "downloadable", False, False),
-        ("verified", "generating", True, True),
-        ("verified", "notpassing", False, False),
-        ("verified", "unverified", False, False),
+        ("verified", True, True),
+        ("audit", False, False),
+        ("verified", False, False),
     )
     @ddt.unpack
-    def test_has_passing_certificate(self, certificate_type, status, is_passing, expected_result):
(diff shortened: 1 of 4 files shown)


dashboard/utils_test.py
```diff
@@ -557,14 +557,14 @@ def test_has_paid_for_any_in_program(self):
         assert mmtrack.has_paid_for_any_in_program() is True
 
     @ddt.data(
```
rhysyngsun: I'm not 100% sure, but reading the code I think these test cases could be simplified now since `has_passing_certificate is contingent only on values of `certificate_type` and `is_passing`, so I think there's only 3 cases that are still worth testing:

- "audit"
- "verified" + `is_passing=True`
- "verified" + `is_passing=False`

annagav: right, the status is not used anymore...

dashboard/utils_test.py
```diff
@@ -573,7 +571,6 @@ def test_has_passing_certificate(self, certificate_type, status, is_passing, exp
             "username": "staff",
             "course_id": course_key,
             "certificate_type": certificate_type,
-            "status": status,
```
rhysyngsun: `"status"` should still probably be present, but it can be a fixed value.

annagav: ok, I thought it wouldn't make any difference...

PR #3286, opened 2017-06-07
Use test index for snapshot_dashboard_states

### LINKED ISSUE #3282: Snapshot management command clears the Elasticsearch index
This isn't a huge problem, the developer can just run `recreate_index` to get it back

noisecapella:
Fixes #3282 

#### What's this PR do?
It looks like the environment variables are ignored in the management command because `settings.py` was already imported before these variables were set. Instead `override_settings` is used to explicitly change the Django settings.

Run `./manage.py snapshot_dashboard_states --suffix contact_course` and verify that there are results in `/learners`.

### DIFF
--- a/selenium_tests/management/commands/snapshot_dashboard_states.py
+++ b/selenium_tests/management/commands/snapshot_dashboard_states.py
@@ -8,6 +8,7 @@ from django.core.management import (
     BaseCommand,
     call_command,
 )
+from django.test import override_settings
 import pytest
 
 from courses.factories import CourseRunFactory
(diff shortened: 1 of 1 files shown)


selenium_tests/management/commands/snapshot_dashboard_states.py
```diff
@@ -224,10 +225,6 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         os.environ['DJANGO_LIVE_TEST_SERVER_ADDRESS'] = '0.0.0.0:8286'
-        os.environ['ELASTICSEARCH_INDEX'] = 'testindex'
```
noisecapella: None of these were actually being used but the default values for the other three were sufficient. `DJANGO_LIVE_TEST_SERVER_ADDRESS` is special since it's used explicitly as an environment variable and not as a means to set a setting property.

Issue #3254, opened 2017-05-31
Mail dialog rich text editor jumbles up letters


noisecapella:
To reproduce, open the mail compose dialog and type 'hello world' in the body textbox. This is what I see on Chrome:
![screenshot from 2017-05-31 11-04-35](https://cloud.githubusercontent.com/assets/863262/26638761/f563215c-45f0-11e7-87f3-aa85762b7a0b.png)

On Firefox:
![screenshot from 2017-05-31 11-05-33](https://cloud.githubusercontent.com/assets/863262/26638826/1d05ec94-45f1-11e7-848a-0185dbb03ef9.png)

I feel like there's some initial state that's not being set properly

PR #3290, opened 2017-06-08
Fixed mail dialog rich text editor jumbling up letters

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3254

#### What's this PR do?
On chrome/firefox OSX, Email composer rich text editor was reversing first 2 characters, this occur on first time after opening email composer. After that it worked fine. 
I fixed that behaviour

go to email composer and type text like `hello world`

@pdpinch 

screen shot 2017-06-08 at 4 14 13 pm

### DIFF
--- a/static/js/components/email/EmailCompositionDialog.js
+++ b/static/js/components/email/EmailCompositionDialog.js
@@ -165,8 +165,8 @@ export default class EmailCompositionDialog extends React.Component {
 
   onEditorStateChange = (editorState: EditorState) => {
     const { updateEmailBody } = this.props;
-    this.setState({ editorState });
     updateEmailBody(editorState);
+    this.setState({ editorState });
   };
 
   clearEditorState = () => {


static/js/components/email/EmailCompositionDialog.js
```diff
@@ -165,8 +165,8 @@ export default class EmailCompositionDialog extends React.Component {
 
   onEditorStateChange = (editorState: EditorState) => {
     const { updateEmailBody } = this.props;
-    this.setState({ editorState });
     updateEmailBody(editorState);
+    this.setState({ editorState });
```
aliceriot: good catch! I'm guessing we were getting a race condition with `componentWillReceiveProps`, where the `updateEmailBody` call was causing a prop change and then we were setting the state twice. it's nice when there's a simple solution to things :smile:

amir-qayyum-khan: thank :)

PR #3298, opened 2017-06-09
Fixed encoding issue for binary audit files

### LINKED ISSUE #3296: Exam response zip files are corrupted
Root cause appears to be an encoding issue. We need to be sure auditing files are encrypted correctly so they can be decrypted later. We missed this in our testing because we tested with text files which were not susceptible to corruption from the encoding passes.

rhysyngsun:
Fixes #3296 

#### What's this PR do?
Monkey patches `gnupg` so that it doesn't utf-8 encode the binary data. Note there aren't any tests around this because I couldn't find a way to reliably execute `gpg --decrypt` directly under python, even using `gnupg`.

- Create a zip file in your project root: `echo "test" > test.txt && zip archive.zip test.txt`
- In a shell:
```python
from exams.pearson.audit import ExamDataAuditor
auditor = ExamDataAuditor()
auditor.configure() # this will likely raise an ImproperlyConfigred error, ignore it
auditor.encrypt('archive.zip', 'archive.zip.gpg`)
```
- Back in bash shell: `gpg --output archive2.zip --decrypt archive.zip.gpg`
- Verify the file is a valid zip archive: `zip -l archive2.zip`
- Unzip and verify the integrity of the text file contents

### DIFF
--- a/exams/pearson/audit.py
+++ b/exams/pearson/audit.py
@@ -11,6 +11,7 @@ from boto.s3 import (
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 import gnupg
+import gnupg._util
 
 from micromasters import utils
 
@@ -21,6 +22,10 @@ REQUIRED_SETTINGS = [
 ]
 
 log = logging.getLogger(__name__)
+
+# monkey patch gnupg so it doesn't utf-8 encode our files
+# see: https://github.com/isislovecruft/python-gnupg/issues/169
+gnupg._util.binary = lambda data: data  # pylint: disable=protected-access
 
 
 class S3AuditStorage:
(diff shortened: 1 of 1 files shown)


exams/pearson/audit.py
```diff
@@ -182,6 +187,25 @@ def configure(self):
 
         self.configured = True
 
+    def encrypt(self, filename, encrypted_filename):
+        """
+        Encrypts the local file with GPG
+
+        Args:
+            filename: absolute path to the local file
```
noisecapella: Can you add types to the arguments here?

noisecapella:
I'm getting this error when reproducing:

    In [4]: auditor.encrypt('archive.zip', 'archive.zip.gpg')
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-4-8e310c20bca5> in <module>()
    ----> 1 auditor.encrypt('archive.zip', 'archive.zip.gpg')
    
    /src/exams/pearson/audit.py in encrypt(self, filename, encrypted_filename)
        201             self.gpg.encrypt(
        202                 source_file,
    --> 203                 self.fingerprint,
        204                 armor=False,
        205                 symmetric=False,
    
    /src/exams/pearson/audit.py in fingerprint(self)
        130     def fingerprint(self):
        131         """The fingerprint for the public GPG key"""
    --> 132         return re.sub(r'\s+', '', settings.EXAMS_AUDIT_ENCRYPTION_FINGERPRINT)
        133 
        134     @property
    
    /usr/local/lib/python3.6/re.py in sub(pattern, repl, string, count, flags)
        189     a callable, it's passed the match object and must return
        190     a replacement string to be used."""
    --> 191     return _compile(pattern, flags).sub(repl, string, count)
        192 
        193 def subn(pattern, repl, string, count=0, flags=0):
    
    TypeError: expected string or bytes-like object

noisecapella:
I'm guessing the workaround would neutralize gnupg's ability to encode anything. Just confirming, we're only ever going to encode binary data?

rhysyngsun:
Sorry, you need to generate a new gpg key:

###### Set up your .env
- First, you need to generate a GPG key `gpg --gen-key`. Follow the prompts.
- Get the key fingerprint `gpg --list-keys --fingerprint`, put this in `.env` as `EXAMS_AUDIT_ENCRYPTION_FINGERPRINT`
- Export the gpg key `gpg --armor --export <EMAIL> | base64 -w 0` (use email you entered in step 1), put this as `EXAMS_AUDIT_ENCRYPTION_PUBLIC_KEY` 
- Requisition yourself values for `EXAMS_AUDIT_S3_BUCKET`, `EXAMS_AUDIT_AWS_ACCESS_KEY_ID`, and `EXAMS_AUDIT_AWS_SECRET_ACCESS_KEY` from me

rhysyngsun:
We will be encrypting text files too, those tested fine with the code for me as well.

exams/pearson/audit_test.py
```diff
@@ -288,3 +289,32 @@ def test_exam_data_auditor_audit_response_file(auditor, mocker):
...
+
+        # use gpg directly since we can't test the lib with the lib
+        result = subprocess.run([
+            'gpg',
+            '--output',
+            decrypted_filename,
+            '--decrypt',
+            encrypted_filename,
+        ], shell=True)
+        print(result)
```
noisecapella: Stray print statement

noisecapella:
Functionality works for me, feel free to merge when the test is fixed

exams/pearson/audit_test.py
rhysyngsun: That test accidentally got committed. Removed the whole block.

PR #3304, opened 2017-06-13
Use UserInfo in edx_api to get user data

### LINKED ISSUE #1375: The edxorg backend should use edx-api-client library and not do direct calls
We should implement a client in `edx-api-client` for the user profile infos to handle 
`/api/mobile/v0.5/my_user_info`
and than remove the direct calls in the edxorg backend.

annagav:
Fix #1375

#### What's this PR do?
Uses `EdxApi` to get user data.

Should be able to login successfully.

### DIFF
--- a/backends/edxorg.py
+++ b/backends/edxorg.py
@@ -3,8 +3,8 @@ EdX.org backend for Python Social Auth
 """
 from urllib.parse import urljoin
 
-from edx_api.client import EdxApi
 from django.conf import settings
+from edx_api.client import EdxApi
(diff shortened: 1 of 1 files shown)


backends/edxorg.py
```diff
@@ -3,6 +3,7 @@
 """
 from urllib.parse import urljoin
 
+from edx_api.client import EdxApi
```
noisecapella: Can you move this under `from django.` so they're in alphabetical order?

PR #3319, opened 2017-06-20
Fixed bug with weird grades coming from edx

### LINKED ISSUE #3316: Handle certificate API responses with a grade of ''
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/298481159/

```
ValueError: could not convert string to float: 
(25 additional frame(s) were not displayed)
...
  File "dashboard/api.py", line 184, in get_info_for_program
    get_info_for_course(course, mmtrack)
  File "dashboard/api.py", line 234, in get_info_for_course
    for course_run in course.courserun_set.all()]
  File "dashboard/api.py", line 234, in <listcomp>
    for course_run in course.courserun_set.all()]
  File "dashboard/api.py", line 329, in get_status_for_courserun
    api.freeze_user_final_grade(mmtrack.user, course_run, raise_on_exception=True)
  File "grades/api.py", line 188, in freeze_user_final_grade
    course_run_paid_on_edx=final_grade.payed_on_edx

ValueError: could not convert string to float: 
```

giocalitri:
fixes #3316

#### What's this PR do?
defaults the grade value to 0 (zero) in case the value coming from edx is a weird one (like an empty string)

Talk to me

### DIFF
--- a/grades/api_test.py
+++ b/grades/api_test.py
@@ -213,6 +213,40 @@ class GradeAPITests(MockedESTestCase):
         assert grade.grade == self.certificates.get(course_key).data.get('grade')
         assert grade.payed_on_edx is payed_edx_result
 
+    @ddt.data('', None, ' ', 'foo', )
+    def test_compute_grade_odd_grade(self, odd_value):
+        """
+        Tests in case the grade coming from edX is not a number
+        """
+        # test for grade computed on current grades
+        test_values = (
+            (self.run_fa, api._compute_grade_for_fa, ),
+            (self.run_no_fa, api._compute_grade_for_non_fa, ),
+        )
+        for course_run, grade_func in test_values:
+            course_key = course_run.edx_course_key
+            current_grade = self.current_grades[course_key]
+            current_grade.data['percent'] = odd_value
+            current_grade.save()
+            user_edx_data = CachedEdxUserData(self.user)
+            run_data = user_edx_data.get_run_data(course_key)
+            grade = grade_func(run_data)
+            assert grade.grade == 0.0
+        # test for grade computed on certificates
+        test_values = (
+            (self.run_fa_with_cert, api._compute_grade_for_fa, ),
+            (self.run_no_fa_with_cert, api._compute_grade_for_non_fa, )
+        )
+        for course_run, grade_func in test_values:
+            course_key = course_run.edx_course_key
+            cert = self.certificates[course_key]
+            cert.data['grade'] = odd_value
+            cert.save()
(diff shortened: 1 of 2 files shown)


grades/api_test.py
```diff
@@ -213,6 +236,29 @@ def test_compute_grade_for_non_fa_certs(self, certificate_type, status, grade_re
...
+        assert grade1.grade == 0.0
+
+        course_key2 = self.run_no_fa_with_cert.edx_course_key
+        cert2 = self.certificates[course_key2]
+        cert2.data['grade'] = odd_value
+        cert2.save()
+        user_edx_data2 = CachedEdxUserData(self.user)
+        run_data2 = user_edx_data2.get_run_data(course_key2)
+        grade2 = api._compute_grade_for_non_fa(run_data2)
+        assert grade2.grade == 0.0
```
gsidebo: I understand the repetition in `grades/api.py`, but here it seems less necessary. I would rather we loop over the FA/non-FA runs than copy the test case just to change a couple values.

```python
@ddt.data('', None, ' ', 'foo', )
def test_compute_grade_odd_current_grade(self, odd_value):
    """
    Tests for _compute_grade_for_non_fa function in case the grade coming from edX is not a number
    """
    for course_run in [self.run_no_fa, self.run_fa]:
        course_key1 = course_run.edx_course_key
        current_grade1 = self.current_grades[course_key1]
        current_grade1.data['percent'] = odd_value
        current_grade1.save()
        user_edx_data1 = CachedEdxUserData(self.user)
        run_data1 = user_edx_data1.get_run_data(course_key1)
        compute_grade_func = api._get_compute_func(course_run)
        grade1 = compute_grade_func(run_data1)
        assert grade1.grade == 0.0
```

You could create a similar test case for the runs with certificates (`test_compute_grade_odd_cert_grade`, looping over `self.run_no_fa_with_cert`, `self.run_fa_with_cert`). Let me know what you think

gsidebo:
Functionality looks great

grades/api_test.py
giocalitri: If at the end I still have 2 different tests, I prefer having them focused on the tested function.

gsidebo: it would be 2 different tests, but each test would be shorter by about half, and we wouldn't be repeating the same logic. if you prefer, it would also be possible to limit it to one test case and loop over tuples:

```python
for course_run, course_run_with_cert in [(self.run_no_fa, self.run_no_fa_with_cert), (self.run_fa, self.run_fa_with_cert)]:
```

gsidebo:
Looks good 👍

Issue #3318, opened 2017-06-20
Investigate running travis builds on container infrastructure


noisecapella:
We would need to change our testing infrastructure to not use travis, but I think we don't strictly need it for most of our tests

PR #3320, opened 2017-06-20
Use container build in Travis for JS tests

noisecapella:
Part of #3318 

#### What's this PR do?
Runs the javascript tests and linting without docker which allows Travis to use container based builds, which should make them faster to run.

Look at the travis output and confirm that sudo is not used for the Javascript tests


aliceriot:
I think we should be able to delete the travis-watch dockerfiles as well!

aliceriot:
oh, maybe we'll still need them for the selenium tests actually

noisecapella:
We could move the JS static bundle generation into Travis and then copy it into the docker containers we use for selenium tests

noisecapella:
Closing this, it's much slower than just using docker

Issue #2627, opened 2017-02-15
Refactor dialog actions and reducers


gsidebo:
We currently have a huge amount of code dedicated to the relatively simple act of opening and closing dialogs. For every dialog in Micromasters, we define an action to set the dialog's visibility to `true`, and a reducer to change the state accordingly. PR #2555 introduces the `SHOW_DIALOG` and `HIDE_DIALOG` actions, which can show and hide any number of dialogs without the boilerplate code. We should change our existing dialog actions/reducers to follow this new pattern. As an example, the invocation of the `setLearnerPageDialogVisibility` action would change from `setLearnerPageDialogVisibility()` to `showDialog(LEARNER_PAGE_DIALOG)`.

We also have 3 actions which appear to be dead code: `UPDATE_DIALOG_TEXT`, `UPDATE_DIALOG_TITLE`, and `SET_DIALOG_VISIBILITY` in `static/js/actions/ui.js`.

aliceriot:
Yes, those are dead, I think I added those originally for the signup dialog and they weren't taken out.

PR #3325, opened 2017-06-21
move some dialog stuff over to the `dialogVisibility` system

aliceriot:

part of #2627 

#### What's this PR do?

This moves a few dialogs over to the much more concise `showDialog` / `hideDialog` system. In particular:

- delete some old, unused code (`setDialogTitle` and so on)
- move the profile image, financial aid calculator, and confirm income dialogs over
- delete all the old, now unnecessary code!

The three things I mentioned above should all work 100% the same as they did before, so testing would just involve confirming that there hasn't been a change in functionality.

Also, the tests should be passing and all that


static/js/containers/ProfileImage.js
```diff
@@ -70,16 +70,25 @@ class ProfileImage extends React.Component {
     });
   };
 
+  setDialogVisibility = (b: boolean) => {
```
noisecapella: Can you rename `b` to `visibility` to be more explicit here?

PR #3334, opened 2017-06-22
Added selenium test for financial aid review page

### LINKED ISSUE #2675: Add selenium test for financial aid review page

noisecapella:
Fixes #2675 

#### What's this PR do?
Adds a selenium test for the financial aid review page

N/A

### DIFF
--- a/selenium_tests/basic_test.py
+++ b/selenium_tests/basic_test.py
@@ -6,6 +6,12 @@ from selenium.webdriver.common.keys import Keys
 
 from courses.factories import ProgramFactory
 from dashboard.models import ProgramEnrollment
+from financialaid.constants import FinancialAidStatus
+from financialaid.factories import FinancialAidFactory
+from financialaid.models import (
+    FinancialAid,
+    TierProgram,
+)
 from roles.models import (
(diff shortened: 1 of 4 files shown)


selenium_tests/basic_test.py
```diff
@@ -173,3 +179,47 @@ def test_switch_program(self):
...
+    def setUp(self):
+        super().setUp()
+        self.login_via_admin(self.user)
+
+    def test_approve_docs(self):
+        """Test the financial aid review page"""
+        self.get("/financial_aid/review/{}/{}".format(self.program.id, FinancialAidStatus.DOCS_SENT))
+        self.wait().click(
+            # Mark as received
+            lambda driver: driver.find_element_by_css_selector("tr td:nth-child(7) button")
```
gsidebo: unless there's a really compelling reason not to do it, i think we should add a css class or a data- attribute to this button and select for that here rather than going for a child at some numeric position

noisecapella: Makes sense to me. I was trying to avoid touching the implementation when writing this test but this seems like a tiny change

PR #3335, opened 2017-06-23
Learners in Program Card

### LINKED ISSUE #2516: Learners in this Program card on Dashboard
This is a simple card on the dashboard that shows a grid of user photos, with a View All link that goes to the Learner Search page.

This design is dependent on the Learners Search for Learners issue #2512.

### acceptance criteria

- [ ] This should be behind a feature flag for now

learners_in_program_card 2x

annagav:
Fix #2516

#### What's this PR do?
Add Learners in The Program card to the dashboard.

Make sure there are at least 9 or 10 learners in the program you are enrolled, that have profile images and their profiles are not private. Reload the page several times to see if the selected 8 profiles change randomly.


aliceriot:
hey anna, is this ready for review?

aliceriot:
functionality looks good!

static/js/containers/test_util.js
```diff
@@ -26,7 +26,7 @@ export const SUCCESS_ACTIONS: Array<ActionType> = [
   RECEIVE_GET_USER_PROFILE_SUCCESS,
 ];
 
-export const DASHBOARD_SUCCESS_ACTIONS = SUCCESS_ACTIONS.concat([
+export const SUBSET_DASHBOARD_SUCCESS_ACTIONS = SUCCESS_ACTIONS.concat([
```
aliceriot: `SUBSET` is a little less than clear, could we call this `DASHBOARD_SUCCESS_NO_LEARNERS_ACTIONS` or something similar?

static/js/containers/test_util.js
```diff
@@ -35,6 +35,12 @@ export const DASHBOARD_SUCCESS_ACTIONS = SUCCESS_ACTIONS.concat([
   RECEIVE_FETCH_COUPONS_SUCCESS,
 ]);
 
+
```
aliceriot: extra blank line

static/js/util/integration_test_helper.js
```diff
@@ -42,7 +43,7 @@ export default class IntegrationTestHelper {
     this.sandbox = sinon.sandbox.create();
     this.store = configureMainTestStore((...args) => {
       // uncomment to listen on dispatched actions
-      // console.log(args);
+      // console.log(args[args.length-1]);
```
aliceriot: typo? :)

static/scss/layout.scss
```diff
@@ -37,6 +37,7 @@
 @import "final-exam-card";
 @import "staff-learner-info-card";
 @import "automatic-emails";
+@import "learners-card";
```
aliceriot: could you put the learners-card scss file in the dashboard subdirectory? we have so much styling for the dashboard, I think it's a little cleaner to keep it all in one directory. there's a file in there (`static/scss/dashboard/dashboard.scss`) which imports all the other dashboard files as well.

static/js/containers/DashboardPage.js
```diff
@@ -686,6 +702,11 @@ class DashboardPage extends React.Component {
       />;
     }
 
+    let learnersInProgramCard;
+    if (R.pathOr(false, [program.id, 'data', 'learners'],programLearners)) {
+      learnersInProgramCard = <LearnersInProgramCard programLearners={programLearners[program.id].data}/>;
+    }
```
aliceriot: what do you think about moving this out to a helper function? It could take `program` or `programID` and `programLearners` and return the `<LearnersInProgramCard />` with the right props passed in, so that we could just call that helper function in the JSX below, on like 759.

static/js/containers/DashboardPage.js
```diff
@@ -150,8 +152,12 @@ class DashboardPage extends React.Component {
 
   componentWillUnmount() {
     const { dispatch } = this.props;
+    let program = this.getCurrentlyEnrolledProgram();
     dispatch(clearDashboard(SETTINGS.user.username));
     dispatch(actions.prices.clear(SETTINGS.user.username));
+    if (program !== undefined) {
+      dispatch(actions.programLearners.clear(program.id));
+    }
```
aliceriot: this will only clear for the currently enrolled program, but we could have fetch program info for, say, two programs (if the user has switched programs using the program switching widget). how important do you think it is to clear all program? something like 

```js
R.keys(programLearners.forEach(id => (
  dispatch(actions.programLearners.clear(id))
))
```

I'm not sure we need to worry about it, but idk.

static/js/components/LearnersInProgramCard.js
```diff
@@ -0,0 +1,29 @@
...
+  };
+
+  render() {
+    const { programLearners } = this.props;
+    const learnersList = programLearners.learners.map(learner => (
+        <img key={learner.username}
+          src={learner.image_small}
+          className='learner-image'
+        />
+    ));
```
aliceriot: what do you think about moving this to a helper function? something like

```js
const renderLearners = R.map(learner => (
  <img
    key={learner.username} 
    src={learner.image_small}
    className="learner-image"
  />
))
```

courses/views.py
```diff
@@ -38,6 +41,44 @@ class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = ProgramSerializer
 
 
+class ProgramLearnersView(APIView):
```
aliceriot: could you use the `ListAPIView` class from django rest framework here? I was able to use it when I made the `AutomaticEmailView`, and it sort of did some stuff automatically for me. I'm not so sure of the API stuff so I don't know if it would work, but it might make it easier!

courses/views.py
```diff
@@ -38,6 +41,44 @@ class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
...
+        learners_result = {
+            'learners_count': queryset.count(),
+            'learners': ProfileImageSerializer(queryset[:8], many=True).data
+        }
+        return Response(
+            status=status.HTTP_200_OK,
+            data=learners_result
+        )
+
+
```
aliceriot: are there tests for the python changes?

pdpinch:
Is this behind a feature flag?

courses/views.py
annagav: I am modifying the returned list slightly, making it a dictionary. Not sure how I would use `ListAPIView`

static/js/containers/DashboardPage.js
annagav: not sure if I did what you wanted https://github.com/mitodl/micromasters/pull/3335/commits/c33019da2bb10078b93ab2e856b53ec5fac6df0e

courses/views.py
aliceriot: oh, I don't really know, it just helped me out in the past so I thought I would mention it!

annagav:
@aliceriot, I think this is ready for another review.

courses/views.py
```diff
@@ -38,6 +41,46 @@ class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
...
+            program_id=program_id
+        ).values_list('user', flat=True)
+
+        queryset = Profile.objects.exclude(
+            image_small__exact=''
+        ).filter(user__in=users).exclude(
+            account_privacy='private'
+        ).exclude(
+            user=request.user
+        ).order_by('?')
```
aliceriot: I don't super understand how querysets work, but will this run two separate queries, one to get the user list (`users` on line 62 above) and then another here to get the profiles? is there any way to do it in one?

static/js/containers/DashboardPage.js
```diff
@@ -649,6 +664,15 @@ class DashboardPage extends React.Component {
     return null;
   };
 
+  renderLearnersInProgramCard(programID: number) {
+    const { programLearners } = this.props;
+    let learnersInProgramCard;
+    if(R.pathOr(false, [programID, 'data', 'learners'],programLearners)) {
```
aliceriot: should we still display the card if the result of the `pathOr` call is `[]`? If you do `if ([]) { foo() }` then `foo` will be executed.

static/js/containers/DashboardPage_test.js
```diff
@@ -666,9 +674,14 @@ describe('DashboardPage', () => {
...
       let coursePrices = makeCoursePrices(dashboardResponse);
       let availablePrograms = makeAvailablePrograms(dashboardResponse);
+      let programLearners = makeProgramLearners();
       helper.dashboardStub.returns(Promise.resolve(dashboardResponse));
       helper.programsGetStub.returns(Promise.resolve(availablePrograms));
       helper.coursePricesStub.returns(Promise.resolve(coursePrices));
+      helper.programLearnersStub = helper.fetchJSONWithCSRFStub.withArgs(
+        `/api/v0/programlearners/${program.id}/`
+      );
+      helper.programLearnersStub.returns(Promise.resolve(programLearners));
```
aliceriot: it would probably be good to have at least one test making sure that the card doesn't show up in the API doesn't return anything

static/js/lib/api.js
```diff
@@ -99,6 +99,10 @@ export async function getPrograms(): Promise<AvailablePrograms> {
   }
 }
 
+export function getProgramLearners(programId: number): Promise<ProgramLearners> {
+  return fetchJSONWithCSRF(`/api/v0/programlearners/${programId}/`, {}, true);
```
aliceriot: the `fetchJSONWithCSRF` function no longer takes the `loginOnError` argument that it used to. if you want to have the `loginOnError` functionality (redirecting after a 401) you'll need to do something like what we do in the `getPrograms` function right above.

static/js/reducers/program_learners.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const programLearnersEndpoint: Endpoint = {
+  name: 'programLearners',
+  namespaceOnUsername: true,
+  checkNoSpinner: true,
+  verbs: [GET],
+  getFunc: getProgramLearners,
+  initialState: {},
+  usernameInitialState: INITIAL_PROGRAM_LEARNERS_STATE,
+  fetchFunc: fetchJSONWithCSRF,
```
aliceriot: oh, you shouldn't need to specify the `fetchFunc` here since you're specifying the `getFunc`. A different way to do it is to specify `getURL` and let the redux-hammock code figure out how to go about calling `fetch` with that URL, but if you specify the `getFunc` `redux-hammock` will just use that.

courses/views.py
noisecapella: Just happened to see this, but `values_list` won't evaluate the query. Instead the query is used as a subquery in the `filter` clause above

aliceriot: oh nifty! TIL

noisecapella: Yeah, if you see `values_list` being used there's usually some evaluation happening which is subtle, for example during iteration or `list(...)`: https://docs.djangoproject.com/en/1.11/ref/models/querysets/#when-querysets-are-evaluated

It's also confusing since `repr(...)` will also do it, which means just looking at a queryset in your console will evaluate it

annagav:
@aliceriot, I addressed all the comments.

annagav:
@aliceriot, sorry, I added one more thing. Could you please take a look?

aliceriot:
sure :)

PR #3337, opened 2017-06-23
Added selenium test for program page, refactored ProgramPageFactory

### LINKED ISSUE #3068: Add selenium test for program page
This should catch javascript errors which occur when a user loads the program page

noisecapella:
Fixes #3068 

#### What's this PR do?
Adds a selenium test for the program page, refactors ProgramPageFactory

N/A

### DIFF
--- a/cms/factories.py
+++ b/cms/factories.py
@@ -44,6 +44,7 @@ class ImageFactory(DjangoModelFactory):
 
         return image
 
+
 class ProgramPageFactory(DjangoModelFactory):
     """Factory for ProgramPage"""
     class Meta:
@@ -63,6 +64,7 @@ class ProgramPageFactory(DjangoModelFactory):
         home_page.add_child(instance=page)
         page.save_revision().publish()
         return page
+
 
 class ProgramCourseFactory(DjangoModelFactory):
     """Factory for ProgramCourse"""
--- a/cms/factories.py
+++ b/cms/factories.py
@@ -1,15 +1,21 @@
 """Factories for making test data"""
-from contextlib import contextmanager
-import os.path
-import shutil
-import tempfile
+from tempfile import NamedTemporaryFile
+import uuid
 
 import factory
 from factory.django import DjangoModelFactory
+from factory.fuzzy import FuzzyText
+from robohash import Robohash
 from wagtail.wagtailimages.models import Image
-from willow.image import Image as WillowImage
 
-from cms.models import ProgramPage, ProgramFaculty, ProgramCourse
+from cms.models import (
+    HomePage,
+    InfoLinks,
+    ProgramPage,
+    ProgramFaculty,
+    ProgramCourse,
+    SemesterDate,
+)
 from courses.factories import ProgramFactory, CourseFactory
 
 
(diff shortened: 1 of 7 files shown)


cms/factories.py
```diff
@@ -18,42 +24,46 @@ class ImageFactory(DjangoModelFactory):
...
-
+    description = factory.Faker('text')
     program = factory.SubFactory(ProgramFactory)
+    faculty_description = factory.Faker('paragraph')
+    program_contact_email = factory.Faker('email')
+    title_over_image = factory.Faker('sentence')
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        home_page = HomePage.objects.first()
```
gsidebo: do we not need some logic here to create a `HomePage` if one doesn't exist?

selenium_tests/basic_test.py
```diff
@@ -173,3 +180,23 @@ def test_switch_program(self):
...
+        """Test viewing the program page"""
+        self.login_via_admin(self.user)
+
+        page = ProgramPageFactory.create(program=self.program, title="A Program Title")
+        for _ in range(3):
+            FacultyFactory.create(program_page=page)
+        for _ in range(3):
+            InfoLinksFactory.create(program_page=page)
+        for _ in range(3):
+            SemesterDateFactory.create(program_page=page)
```
gsidebo: i think these 3 loops can be refactored into this:

```python
FacultyFactory.create_batch(3, program_page=page)
InfoLinksFactory.create_batch(3, program_page=page)
SemesterDateFactory.create_batch(3, program_page=page)
```

selenium_tests/basic_test.py
```diff
@@ -173,3 +180,23 @@ def test_switch_program(self):
...
+        for course in self.program.course_set.all():
+            ProgramCourseFactory.create(program_page=page, course=course)
```
gsidebo: this loop can be refactored, but you can take or leave this one:

```python
courses = self.program.course_set.all()
ProgramCourseFactory.create_batch(len(courses), program_page=page, course=factory.Iterator(courses))
```

selenium_tests/basic_test.py
```diff
@@ -173,3 +180,23 @@ def test_switch_program(self):
...
+
+        self.get("/a-program-title/")
```
gsidebo: if we're creating that test data above but not testing for any artifacts from that test data here, why are we creating the test data?

cms/factories.py
noisecapella: No, there is one created in a database migration: `cms/migrations/0004_create_homepage`

selenium_tests/basic_test.py
noisecapella: Sounds good

selenium_tests/basic_test.py
noisecapella: Sure

selenium_tests/basic_test.py
noisecapella: The `get(...)` here includes a check for any Javascript errors so this is checking that there are no errors rendering of the test data in the page. I'll add some checks to make sure the rendered output exists

noisecapella:
Should be ready for review again. I also made a change to `CourseListItemWithPopover` which removes a link element that is superfluous.

selenium_tests/basic_test.py
```diff
@@ -223,3 +234,56 @@ def is_now_pending(driver):  # pylint: disable=unused-argument
...
+
+        semester_elements = self.selenium.find_elements_by_css_selector(".semester-date")
+        assert len(semester_dates) == len(semester_elements)
+        for semester_date, semester_element in zip(semester_dates, semester_elements):
+            assert semester_date.semester_name in semester_element.text
+
+        program_course_elements = self.selenium.find_elements_by_css_selector(".program-course .title")
+        assert len(program_courses) == len(program_course_elements)
+        for program_course, program_course_element in zip(program_courses, program_course_elements):
+            assert program_course.course.title in program_course_element.text
```
gsidebo: you can take or leave this one, but these tests seem a bit over-specific for selenium tests. all of the logic around grabbing specific elements, then grabbing specific elements and attributes within those and sorting them is kind of hard to follow. i think it's perfectly fine just to check that the length of UI elements found is the same as the number of records created for some given model. for example, i think this would be fine to test that faculty objects are being rendered correctly

```python
faculty_elements = self.selenium.find_elements_by_css_selector(".faculty-tile")
assert len(faculty) == len(faculty_elements)
```

i'd strongly prefer this to super-exact and verbose approach, but i'll leave this decision to you. good to go either way

noisecapella: Sure, I'll see if I can shift some of these specific tests to a view test

Issue #3342, opened 2017-06-26
Fixed value coupons show incorrect value in messaging


noisecapella:

noisecapella:
Closed by #3357

Issue #3340, opened 2017-06-26
Prevent coupons from being created for non-financial aid programs


noisecapella:
Just a precaution

PR #3343, opened 2017-06-26
Fix coupon selenium screenshots

noisecapella:
Related to #3340 

#### What's this PR do?
Changes `dashboard_snapshot_states` to create coupons for financial aid programs. Currently this creates coupons for a non-financial aid program, which is an invalid state since there can't be coupons in non-financial aid programs.

Run `./scripts/test/run_snapshot_dashboard_states.sh --match coupon` and look at the screenshots. Some values are incorrect, see #3342 for the issue to fix it.

### DIFF
--- a/selenium_tests/management/commands/snapshot_dashboard_states.py
+++ b/selenium_tests/management/commands/snapshot_dashboard_states.py
@@ -24,10 +24,7 @@ from ecommerce.models import (
 )
 from exams.factories import ExamRunFactory
 from financialaid.factories import FinancialAidFactory
-from financialaid.models import (
-    FinancialAid,
-    FinancialAidStatus,
-)
+from financialaid.models import FinancialAidStatus
 from grades.factories import ProctoredExamGradeFactory
 from seed_data.management.commands.alter_data import EXAMPLE_COMMANDS
 from selenium_tests.base import SeleniumTestsBase
--- a/selenium_tests/management/commands/snapshot_dashboard_states.py
+++ b/selenium_tests/management/commands/snapshot_dashboard_states.py
@@ -24,7 +24,10 @@ from ecommerce.models import (
 )
 from exams.factories import ExamRunFactory
 from financialaid.factories import FinancialAidFactory
-from financialaid.models import FinancialAidStatus
+from financialaid.models import (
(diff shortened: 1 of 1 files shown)


selenium_tests/management/commands/snapshot_dashboard_states.py
```diff
@@ -154,6 +159,16 @@ def with_coupon(self, amount_type, is_program, is_free):
         )
         UserCoupon.objects.create(user=self.user, coupon=coupon)
 
+    def with_coupon_no_financial_aid_application(self):
+        """Coupon without a financial aid application"""
+        self.with_coupon(Coupon.PERCENT_DISCOUNT, True, False)
+        FinancialAid.objects.all().delete()
```
rhysyngsun: This feels like it would be very destructive, is there a way to scope it just to the relevant user?

selenium_tests/management/commands/snapshot_dashboard_states.py
```diff
@@ -154,6 +159,16 @@ def with_coupon(self, amount_type, is_program, is_free):
...
+
+    def with_coupon_pending_financial_aid(self):
+        """Coupon with a non-terminal financial aid application"""
+        self.with_coupon(Coupon.PERCENT_DISCOUNT, True, False)
+        FinancialAid.objects.update(status=FinancialAidStatus.PENDING_MANUAL_APPROVAL)
```
rhysyngsun: Same comment as above.

selenium_tests/management/commands/snapshot_dashboard_states.py
noisecapella: At this point there should only be one user with financial aid but I'll add a filter on that user to make sure

rhysyngsun: Ok, for some reason I thought this was running against the regular db, not the test db, am I mistaken?

noisecapella: Yes, the `snapshot_dashboard_states` uses the test database, and much of the selenium testing infrastructure too

rhysyngsun:
:+1:

PR #3352, opened 2017-06-27
fix automatic email editing

### LINKED ISSUE #3350: automatic email editing is broken
There's a problem on master with editing automatic emails on `/automaticemails`. If you attempt to edit an email and save it this happens:

tasdfasdfasdfasdf

the problem is that the email ID isn't being passed in to the function which figures out the URL to issue the PATCH request for, so we end up PATCH-ing `/api/v0/mail/automatic_email/undefined`, which obviously doesn't correspond to a record we can update.

the bug was introduced in this PR: https://github.com/mitodl/micromasters/pull/3238

### LINKED ISSUE #3351: toggling 'active' state on automatic emails is broken
not sure what's going on, but this behavior is broken on master for some reason.

aliceriot:

closes #3350 
closes #3351 

#### What's this PR do?

this fixes the editing for automatic emails. basically, the ID was accidentally ommitted. that isn't great.

I added a test that should fail on master. But also, you should just confirm yourself that things aren't working on master. go to `/automaticemails` and try to edit one. When you click 'save' you should get an error in the JS console. The branch should restore the functionality.

To test that the test I added fails on master you can do:

```
git cc ap/fix-automatic-emails -- static/js/containers/AutomaticEmailPage_test.js
```

and confirm that there's a failing test. then check out my branch and confirm all tests are passing.

Also, for the 'toggling' functionality, confirm that you can toggle the 'active' state for any automatic email that shows up on the page. If you toggle an email, you should not see any change in 'filters' that show up in the recipients thingy (if you open the email editor dialog again after toggling it).

### DIFF
--- a/static/js/components/email/lib.js
+++ b/static/js/components/email/lib.js
@@ -190,6 +190,7 @@ export const AUTOMATIC_EMAIL_ADMIN_CONFIG: EmailConfig = {
       subject: emailOpenParams.email_subject,
       body: emailOpenParams.email_body,
       sendAutomaticEmails: emailOpenParams.enabled,
+      id: emailOpenParams.id,
     },
     filters: getFilters(emailOpenParams.query.original_query.post_filter)
   }),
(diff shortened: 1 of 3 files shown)


aliceriot:
I just updated the PR with a fix for #3351 which was also broken in the same PR.

mbertrand:
Looks good! 👍

PR #3356, opened 2017-06-27
integrate redux-hammock and delete old stuff

### LINKED ISSUE #3293: repackage redux-rest as a standalone module
We should publish the redux-rest library as a separate module, so that we can use it in all our projects without needing to copy+paste code or anything like that. Perhaps other people will find it helpful too!

The first problem, though, is that we would need to come up with a new name, since [redux-rest](https://github.com/Kvoti/redux-rest) is already taken.

maybe restful-redux? redux-hammock? naptime? chamomile? I could go on with these!

@noisecapella any thoughts on this?

aliceriot:

closes #3293 

#### What's this PR do?

This PR integrates redux-hammock, switching over some of our API functions to depend on it and also the redux-rest stuff.

Basically, we should just be checking that anything API related is still working as it was before.


noisecapella:
Can you rebase please? I think there's some automatic email stuff in this PR which was part of another PR that was merged

aliceriot:
@noisecapella rebased!

static/js/components/email/lib_test.js
```diff
@@ -187,8 +188,9 @@ describe('Specific email config', () => {
     it('shouldnt use the sendMail email action, if the email config specifies differently', () => {
       let automaticEmailState = _.clone(filledOutEmailState);
       automaticEmailState.inputs.id = 1;
-      helper.fetchJSONWithCSRFStub.withArgs('/api/v0/mail/automatic_email/1/').
-        returns(Promise.resolve());
+      fetchMock.mock('/api/v0/mail/automatic_email/1/', () => (
```
noisecapella: Is this change related to this PR? It looks like it came from the automatic emails PR

aliceriot: I had to make this change to get the mocking to work correctly, mocking at the `fetch` level instead of at the `fetchJSONWithCSRF` level.

aliceriot: so yes, it's specific to this pr.

noisecapella:
It looks like the login redirect behavior is broken in this PR. When I edit the dashboard API to return a 401 status code, it causes a refresh loop on master (which is expected) but I see this error message on this branch: 

screenshot from 2017-06-29 14-49-01

static/js/containers/AutomaticEmailPage_test.js
```diff
@@ -93,18 +97,9 @@ describe('AutomaticEmailPage', () => {
...
 
   it('should let you save an email', () => {
-    helper.fetchJSONWithCSRFStub.withArgs(
-      `/api/v0/mail/automatic_email/${GET_AUTOMATIC_EMAILS_RESPONSE[0].id}/`,
-      {
-        method: 'PATCH',
-        body: JSON.stringify({
-          email_subject: GET_AUTOMATIC_EMAILS_RESPONSE[0].email_subject,
-          email_body: GET_AUTOMATIC_EMAILS_RESPONSE[0].email_body,
-          sendAutomaticEmails: true,
```
aliceriot: I'm just using `fetchMock` to mock the correct response, I can add an assertion to check that the request is formed correctly

static/js/lib/api.js
```diff
@@ -154,7 +37,8 @@ export function patchUserProfile(username: string, profile: Profile): Promise<Pr
 }
 
 export function getDashboard(username: string): Promise<Dashboard> {
-  return fetchJSONWithCSRF(`/api/v0/dashboard/${username}/`, {}, true);
+  return fetchJSONWithCSRF(`/api/v0/dashboard/${username}/`, {})
```
aliceriot: sure

static/js/containers/AutomaticEmailPage_test.js
aliceriot: I just added that!

aliceriot:
OK I just pushed some changes that should fix the redirect issue, lmk if it's still a problem.

PR #3345, opened 2017-06-26
Fix course coupon program messages

noisecapella:
#### What are the relevant tickets?
Fixes #3263 

#### What's this PR do?
Fixes coupon course message bugs

#### How should this be manually tested?
Run `./scripts/test/run_snapshot_dashboard_states.sh --match coupon` on `master` and this branch, and compare the output. The program messages should be correct in the branch for this PR

PR #3357, opened 2017-06-28
Revert "Fixed message for course-level coupon (#3281)"

noisecapella:
Reverts #3281 

#### What's this PR do?
Reverts course message fix PR. The fix is incomplete and we need to do a release before #3345 is ready to merge.

Tests should pass


aliceriot:
tests are passing, is there any manual testing I should do here?

noisecapella:
You could run `./scripts/test/run_snapshot_dashboard_states.sh` but some of the coupon states will be broken anyway, so I'm not sure there's anything we should do here

aliceriot:
ok, I'm fine merging then, since this is just a revert.

PR #3361, opened 2017-06-28
Populate exam_run for ProctoredExamGrades

### LINKED ISSUE #3300: ProctoredExamGrade.exam_run is not being set
This is required for grades to show up in the dashboard. We need to set it on the record when we process EXAM files and then write a migration to backpopulate the data. We shouldn't encounter a data race between the code deploy and migration because we're not currently processing grade responses.

- [ ] In `exams.pearson.download.ArchivedResponseProcessor.process_exam_file` ([link](https://github.com/mitodl/micromasters/blob/master/exams/pearson/download.py#L323-L332)) we should set `ProtoredExamGrade.exam_run` to the value of `exam_authorization.exam_run`.
- [ ] Then write a migration to populate the `exam_run` for existing `ProtectedExamGrade` records that don't have a value in that column. The correct `ExamRun` can be found by the query `ExamAuthorization.objects.get(id=peg.client_authorization_id).exam_run`.

annagav:
Fix #3300

#### What's this PR do?
Populate exam_run for ProctoredExamGrades

Run the migration and make sure that `ProctoredExamGrade`s that had null for `exam_run` field, have that field now populated.


annagav:
@rhysyngsun, I addressed all comments.

PR #3363, opened 2017-06-29
fix the email validation

### LINKED ISSUE #3360: Emails with link to email address do not work
If in the learners page I create an email with a link to an email like`mailto:giovanni@example.com` I get an error.

This has been found by Kayli Maffei in production while trying to send an email. 

screen shot 2017-06-28 at 3 05 50 pm

aliceriot:

closes #3360

#### What's this PR do?

this fixes the email validation so the user can save an email with a `mailto` link in it.

If you create an email and add a `mailto` link you should be able to save the email. You should still be unable to save a link which does not begin with `http|https` (e.g. `wikipedia.org/wiki/Potato`).

### DIFF
--- a/static/js/lib/validation/profile.js
+++ b/static/js/lib/validation/profile.js
@@ -361,7 +361,7 @@ let emailMessages: ErrorMessages = {
 
 const emailBodyValid = R.ifElse(
   R.test(/<a.*>.*<\/a>/),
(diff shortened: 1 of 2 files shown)


noisecapella:
Can you update the `All link URLs must start with 'http' or 'https'` validation message in `static/js/lib/validation/profile.js`?

aliceriot:
yeah that makes sense

PR #3370, opened 2017-06-30
Command to diff dashboard_states screenshots

rhysyngsun:
N/A

#### What's this PR do?
Wrote this for myself to diff two branches when reviewing, I think it's generally useful to quickly see the diffs on a large set of dashboard_states.

- Install imagemagick on your host w/ png support (`apt-get install libpng-dev` before installing `imagemagick`)
- Checkout this branch, then create a dummy branch off it to compare against with a small dashboard text change (need to do this so the `--output` arg is supported on the comparison branch)
- `./scripts/test/diff_dashboard_states.sh OTHER_BRANCH_NAME --match coupon`
- `./scripts/test/diff_dashboard_states.sh OTHER_BRANCH_NAME`


scripts/test/diff_dashboard_states.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Requirements:
+# - Imagemagick installed with PNG support locally
```
noisecapella: Can you add a check for the binary `magick` at the beginning?

scripts/test/diff_dashboard_states.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -e -o pipefail
+
+BRANCH1=$(git rev-parse --abbrev-ref HEAD)
+BRANCH2=$1
```
noisecapella: Can you add a check that this is a valid branch name?

scripts/test/diff_dashboard_states.sh
```diff
@@ -0,0 +1,32 @@
...
+
+shift 1
+
+OUT1="output/dashboard_states_A"
```
noisecapella: Instead of `A` and `B` can we name them `current` and `other`, or maybe use the branch names themselves, mangled to remove the `/`?

selenium_tests/management/commands/snapshot_dashboard_states.py
```diff
@@ -299,7 +297,7 @@ def test_dashboard_states(self):
             scenarios_with_numbers = (
                 (num, (run_scenario, name))
                 for (num, (run_scenario, name)) in scenarios_with_numbers
-                if match in self._make_filename(num, name)
+                if match in self._make_filename(num, name, output_directory)
```
noisecapella: We might want to pass an empty string for the output directory here, so that we don't match against the directory name by accident

noisecapella:
It looks like the `magick compare` returns a non-zero error code even when producing a proper output file

rhysyngsun:
@noisecapella feedback addressed

scripts/test/diff_dashboard_states.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+set -e -o pipefail
+
+OUT1="output/dashboard_states_current"
+OUT2="output/dashboard_states_other"
+DIFFOUT="output/dashboard_states_diff"
+
+mkdir -p output
+chmod o+rwx output
```
noisecapella: What's this for?

scripts/test/diff_dashboard_states.sh
```diff
@@ -0,0 +1,50 @@
...
+
+mkdir -p "$DIFFOUT"
+
+set +e # imagemagick returns nonzero codes, so ignore them
+
+for f1 in $OUT1/*.png ; do
+  fname=$(basename $f1)
+  echo "Diffing: $fname"
+  magick compare "$OUT1/$fname" "$OUT2/$fname" "$DIFFOUT/$fname"
+done
```
noisecapella: Can you add an `exit 0` at the end so the script doesn't return the non-zero exit code for `magick`?

scripts/test/diff_dashboard_states.sh
rhysyngsun: If `output/` is created by the command running under the docker container's user (via `os.makedirs("output/dashboard_states_current")`), that directory is owned by that user and the host system user can't `mkdir -p output/dashbaord_states_diff` and vice versa. This was the only way to make that functional (short of docker quackery).

Issue #3365, opened 2017-06-29
Jumbled text when entering text in search mail dialog in chrome


noisecapella:
To reproduce:

 - click 'Email these learners'
 - Click on the subject and type "Hi"
 - Press the tab button and then write "Hello". It will appear as "lloeH"

This issue doesn't appear to happen if you click the body before entering text, just when you tab into it

PR #3372, opened 2017-07-03
Fixed jumbled text when entering text in search mail dialog in chrome

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3365

#### What's this PR do?
Fixed jumbled text when user use tab on email composer

go to email compose and use tab to set subject and body

@pdpinch @aliceriot

### DIFF
--- a/static/js/components/email/EmailCompositionDialog.js
+++ b/static/js/components/email/EmailCompositionDialog.js
@@ -84,8 +84,11 @@ export default class EmailCompositionDialog extends React.Component {
   }
 
   componentWillReceiveProps (nextProps: EmailDialogProps) {
-    if (!this.state.editorState.getCurrentContent().hasText()) {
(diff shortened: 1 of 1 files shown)


aliceriot:
@amir-qayyum-khan we have to use `this.state` to hold the `editorState` object because it's not a plain JS object, it's an [Immutable](https://facebook.github.io/immutable-js/) object, which can't really be stored in the redux store.

noisecapella:
It looks like this fixes the issue on Chrome but I'm seeing a problem on Firefox. Can you investigate further?

pdpinch:
@amir-qayyum-khan what's happening with this PR?

amir-qayyum-khan:
@noisecapella please review thsi

noisecapella:
Confirmed that IE, Firefox and Chrome work fine now

PR #3397, opened 2017-07-12
fix potentially flaky JS thingy in the program factory

aliceriot:

none

#### What's this PR do?

removed the `""` empty string from the array of possible choices for `pearson_exam_status` in the `makeProgram` factory. this could be chosen, rarely, and would cause a test in `StatusMessage_test` to fail.

make sure it makes sense and nothing is failing?


aliceriot:
this is probably not actually the issue, because now I can't reproduce the failure. `¯\_(ツ)_/¯`

rhysyngsun:
I think you can close this because my other PR #3398 fixes the real issue. `""` is in actually a possible value for`pearson_exam_status` so we should be testing it: https://github.com/mitodl/micromasters/blob/master/dashboard/utils.py#L356

aliceriot:
yup, closing this.

PR #3398, opened 2017-07-12
Specify course_end_date for makeRun

rhysyngsun:
N/A

#### What's this PR do?
This defines `course_end_date` for `makeRun` to avoid a race condition with `courseUpcomingOrCurrent` comparing now to now because we were passing `undefined` to `moment()`.

Tests should pass


aliceriot:
confirmed the fix works by running the test 1000 times. on master this resulted in 9 failures, on this branch, 0.

PR #3400, opened 2017-07-12
upgrade react telephone input component

aliceriot:

none

#### What's this PR do?

this updates `react-telephone-input` to the latest version. the most recent version fixes a warning we were seeing in the JS console, because previously it used `React.createClass`, which is a soon-to-be deprecated API. Now it is fixed, so we may upgrade and all that.

The tests should pass without error, and the telephone input component should behave as before.


aliceriot:
I just got a notification about [this issue](https://github.com/mukeshsoni/react-telephone-input/issues/119) closing, which prompted the PR.

PR #3403, opened 2017-07-13
Fix JS race condition in tests

### LINKED ISSUE #3399: Fix intermittent Unmatched GET to /api/v0/dashboard/jane/ warning

noisecapella:
Fixes #3399

#### What's this PR do?
This is caused by a test in `DashboardPage_test.js` which didn't wait for all actions which were dispatched. This caused a fetch to happen after the sandbox was cleared, which meant that `fetchMock` handled it instead without any mock for that endpoint. This:
 - fixes the test to listen for the additional actions
 - Moves `fetchMock.restore()` into `global_init` `afterEach` so we aren't listening for actions in between tests anymore
 - adds commented out code to aid in debugging this issue in the future

N/A

### DIFF
--- a/static/js/containers/AutomaticEmailPage_test.js
+++ b/static/js/containers/AutomaticEmailPage_test.js
@@ -36,7 +36,7 @@ describe('AutomaticEmailPage', () => {
 
   afterEach(() => {
     helper.cleanup();
-  }); 
+  });
 
   const baseActions = [
     REQUEST_GET_USER_PROFILE,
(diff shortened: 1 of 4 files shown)


aliceriot:
do you happen to know which test in `DashboardPage_test.js` was causing the issue? Just want to confirm the fix

noisecapella:
It's the `only if status is ${status}` one

aliceriot:
cool, thx

aliceriot:
confirmed fix works!

PR #3418, opened 2017-07-19
Created management command to make exam grade adjustments

### LINKED ISSUE #3309: adjust exam results
We need to make adjustments to the results of the first round of exams. 

It's possible that we may need to do this every time an exam is offered.

gsidebo:
Closes #3309 

#### What's this PR do?
Creates a management command that can adjust exam grades based on data in a csv

This gist will help: https://gist.github.com/gsidebo/8c3252a4d8f0b9f7bd67ffa3ba2782f6

It contains a script for ensuring that a set of users have exam grades, and an example csv for the grade adjustments. Once you have a set of users with exam grades, download that csv to your project root and change the `proctoredexam_id` column values to match the `ProctoredExamGrade.id` values for your users (the `user_id` column in this csv is ignored - the only column that matters for matching to our records is `proctoredexam_id`). Run the management command with this file (e.g.: `./manage.py adjust_exam_grades_from_csv "./test_grade_adjust.csv"`), then check the exam scores in the database to make sure that the adjustments worked. 

You can also mess with the csv (change scores to be invalid, change id's to something nonexistent, etc) to see how it handles errors

The csv's that we get from DEDP are tab-delimited

### DIFF
--- a/grades/management/commands/adjust_exam_grades_from_csv.py
+++ b/grades/management/commands/adjust_exam_grades_from_csv.py
@@ -1,0 +1,119 @@
+"""
+Freezes final grades for a course
+"""
+import csv
+import argparse
+from collections import namedtuple
+from django.core.management import BaseCommand, CommandError
+
+from grades.models import ProctoredExamGrade
+
+
+class ParsingError(CommandError):
+    """Custom class for parsing exceptions"""
+    pass
+
+
+class GradeRowParser:
+    """Parser for rows of grade adjustment information in a CSV"""
+    RowProps = namedtuple('RowProps', ['exam_grade_id', 'score'])
+    default_col_names = dict(
+        exam_grade_id='proctoredexam_id',
+        score='score',
+    )
+
+    def __init__(self, col_names=None):
+        """
+        Args:
+            col_names (dict): Mapping of RowProps property name to the name of the column in the CSV
+        """
+        col_names = col_names or {}
+        self.col_names = self.RowProps(**{**self.default_col_names, **col_names})
+
+    def parse_and_validate_row(self, row):
+        """Parses a row of grade adjustment info and makes sure it doesn't contain bad data"""
+        try:
+            parsed_row = self.RowProps(
+                exam_grade_id=int(row[self.col_names.exam_grade_id]),
+                score=float(row[self.col_names.score]),
+            )
+        except KeyError as e:
+            raise ParsingError('Row is missing a required column: {}'.format(str(e)))
+        except ValueError as e:
+            raise ParsingError('Row has an invalid value: {}'.format(str(e)))
+
+        if parsed_row.score < 0.0 or parsed_row.score > 100.0:
+            row_identifier = '{}: {}'.format(self.col_names.exam_grade_id, parsed_row.exam_grade_id)
+            raise ParsingError('[{}] "score" value must be between 0 and 100'.format(row_identifier))
+        return parsed_row
+
+    def parse_exam_grade_adjustments(self, csv_reader):
+        """
+        Parses all rows of grade adjustment info from a CSV and yields each ProctoredExamGrade object
+        with its associated grade adjustment row from the CSV
+
+        Args:
+            csv_reader (csv.DictReader): A DictReader instance
+
+        Returns:
+            tuple(ProctoredExamGrade, RowProps):
+                A tuple containing a ProctoredExamGrade and its associated parsed CSV row
+        """
+        parsed_row_dict = {}
+        for row in csv_reader:
+            parsed_row = self.parse_and_validate_row(row)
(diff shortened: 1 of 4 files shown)


pdpinch:
@rhysyngsun can you take a look at this too? Did we decide that this could be a destructive change, or am I missing something?

rhysyngsun:
@pdpinch We definitely don't want this to be destructive. That said I ***think***  we store the original data in the `row_data` blob. We should absolutely verify that's the case in production before proceeding with this code, though, otherwise we face a very rough re-export process of the original grades again.

grades/management/commands/adjust_exam_grades_from_csv.py
```diff
@@ -0,0 +1,98 @@
...
+        csvfile = kwargs.get('csvfile')
+        reader = csv.DictReader(csvfile.read().splitlines(), delimiter='\t')
+        grade_row_parser = GradeRowParser(col_names=col_names)
+
+        total_rows = 0
+        grades_changed = 0
+        grades_unchanged = 0
+        not_found = 0
+        for row in reader:
+            parsed_row = grade_row_parser.parse_and_validate_row(row)
```
rhysyngsun: I would suggest we parse and validate all the rows upfront because if one part of the CSV is invalid it casts doubt on the rest of it.

grades/management/commands/adjust_exam_grades_from_csv.py
```diff
@@ -0,0 +1,98 @@
...
+            try:
+                exam_grade = ProctoredExamGrade.objects.get(id=parsed_row.exam_grade_id)
+                if exam_grade.score != parsed_row.score:
+                    exam_grade.set_score(parsed_row.score)
+                    exam_grade.save_and_log(None)
```
rhysyngsun: I think we had settled on the destructive `set_score` being ok since we log the changes, right?

grades/models.py
```diff
@@ -231,6 +232,15 @@ def for_user_course(cls, user, course):
         now = now_in_utc()
         return cls.objects.filter(user=user, course=course, exam_run__date_grades_available__lte=now)
 
+    def set_score(self, new_score):
+        """
+        Sets a value for the score field, then sets related fields based on that value
+        """
+        self.score = new_score or 0.0
```
rhysyngsun: This doesn't seem right to me. If we get a falsy score we should probably error instead of defaulting to 0.

grades/models_test.py
```diff
@@ -139,3 +142,31 @@ def test_for_user_course(self):
...
+        (0.0, True, EXAM_GRADE_PASS),
+        (-5.0, False, EXAM_GRADE_FAIL),
+    )
+    @ddt.unpack
+    def test_set_score(self, grade_adjust, expected_passed_value, expected_grade_str):
+        """Tests that the set_score helper method sets score-related fields appropriately"""
+        passing_score = 60.0
+        grade = ProctoredExamGradeFactory.build(
+            passing_score=passing_score,
+            score=None,
```
rhysyngsun: Curious why you're passing `None` to these other 3 values since in the real world those fields will have values.

grades/management/commands/adjust_exam_grades_from_csv.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    def __init__(self, col_names=None):
+        col_names = col_names or {}
+        self.col_names = self.RowProps(**{**self.default_col_names, **col_names})
+
+    def parse_and_validate_row(self, row):
+        """Parses a row of grade adjustment info and makes sure it doesn't contain bad data"""
+        try:
+            parsed_row = self.RowProps(
```
rhysyngsun: Could also get a `ValueError` out of this.

rhysyngsun:
Functionality looks good.

grades/management/commands/adjust_exam_grades_from_csv.py
gsidebo: Yup, `save_and_log` is exactly what ensures that the change isn't destructive. A `ProctoredExamGradeAudit` record will be created with the old value

grades/models.py
gsidebo: Yeah, agreed

grades/models_test.py
gsidebo: `set_score` sets all 3 values, so I figured a good way to assert that `set_score` was doing its job was to initialize the object with all of those values initially set to `None`

grades/management/commands/adjust_exam_grades_from_csv.py
gsidebo: I approached this assuming that there could be a typo and that the feedback cycle with Pearson could be unpredictable, so we might want to process these files in a way that works for valid rows and only fails for invalid rows. I take your point, though

Issue #3153, opened 2017-05-05
Clicking the learner link does not clear filters on the learner page


noisecapella:
Probably a minor issue, but if you're already on the learner search page and you click on the link on the upper left, or the link in the panel on the left, it will clear the filters and search text from the URL but it won't affect the learner page itself. We should clear the filters and search text when this happens, either by making them regular links, or by doing an explicit refresh in searchkit

PR #3422, opened 2017-07-20
Cleared filters of learner page on learner page link refresh

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3153

#### What's this PR do?
clears filter when url search is empty.

Go to search page and select filters and click on learners link in menu, all filters will be reaset.

@pdpinch

### DIFF
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -328,13 +328,7 @@ export default class LearnerSearch extends SearchkitComponent {
 
   render () {
     const { currentProgramEnrollment } = this.props;
-    // Remove any filters that are still applied by searchkit, but don't appear in the querystring
-    let hasFiltersOtherThanSelectedProgram = _.get(this, 'searchkit.query.index.filters.length', 0) > 1;
-    if (window.location && R.isEmpty(window.location.search) && hasFiltersOtherThanSelectedProgram) {
-      this.searchkit.getQueryAccessor().keepOnlyQueryState();
-    }
 
-    // Render the searchkit component
     return (
       <Loader loaded={!this.isInitialLoading()} shouldRenderAll={true}>
         <div className="learners-search">
--- a/static/js/components/LearnerSearch.js
+++ b/static/js/components/LearnerSearch.js
@@ -13,7 +13,6 @@ import {
   SearchBox,
   SortingSelector,
   MultiMatchQuery,
-  SearchkitManager,
 } from 'searchkit';
 import Grid, { Cell } from 'react-mdl/lib/Grid';
 import Card from 'react-mdl/lib/Card/Card';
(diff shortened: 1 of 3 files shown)


gsidebo:
I tried this on a local branch and got this to work in our searchkit component. It's basically the same code, just in a place that I think makes a little more sense and without the call to `performSearch`.

I changed the `render` method in `static/js/components/search/WithSearchkitManager.js` to look like this:

```javascript
    render() {
      // Remove any filters that are still applied by searchkit, but don't appear in the querystring
      let hasFiltersOtherThanSelectedProgram = _.get(this, 'searchkit.query.index.filters.length', 0) > 1;
      if (window.location && R.isEmpty(window.location.search) && hasFiltersOtherThanSelectedProgram) {
        this.searchkit.getQueryAccessor().keepOnlyQueryState();
      }
      // Render the searchkit component
      return <SearchkitProvider searchkit={this.searchkit}> \\...
```

You'll need to import ramda and lodash to get that working

amir-qayyum-khan:
Done @gsidebo

static/js/components/LearnerSearch.js
```diff
@@ -328,7 +328,13 @@ export default class LearnerSearch extends SearchkitComponent {
 
   render () {
     const { currentProgramEnrollment } = this.props;
+    // Remove any filters that are still applied by searchkit, but don't appear in the querystring
+    let hasFiltersOtherThanSelectedProgram = _.get(this, 'searchkit.query.index.filters.length', 0) > 1;
+    if (window.location && R.isEmpty(window.location.search) && hasFiltersOtherThanSelectedProgram) {
+      this.searchkit.getQueryAccessor().keepOnlyQueryState();
+    }
 
+    // Render the searchkit component
```
gsidebo: @amir-qayyum-khan - This looks good, but I had a couple reasons for suggesting that we put this code in `static/js/components/search/WithSearchkitManager.js` rather than `static/js/components/LearnerSearch.js`

1. `this.searchkit.getQueryAccessor().keepOnlyQueryState()` changes the state of the searchkit component (see [this function](https://github.com/searchkit/searchkit/blob/f59f28398928c47ef0ec3b25706c224b008766b5/src/core/accessors/BaseQueryAccessor.ts#L15)). In cases where this state change is needed, I believe everything above the `LearnerSearch` component will already be rendered, and the state change will cause a re-render. I *think* that changing the state before `SearchkitProvider` is rendered with the `searchkit` prop will avoid that re-render.
1. To me it just makes a little more sense to make this change in the same place that we instantiate the `SearchkitManager` than to make it in a specific component which happens to be the only component that leverages searchkit.

@aliceriot Do you have any opinions on the 2 points above (particularly the first)?

amir-qayyum-khan:
done @gsidebo

PR #3426, opened 2017-07-21
Refactored various factory classes and usages

gsidebo:
No ticket

#### What's this PR do?
Changes up some factory definitions and usages to follow more intuitive patterns and take better advantage of factory-boy features

No need for manual testing

### DIFF
--- a/dashboard/tasks_test.py
+++ b/dashboard/tasks_test.py
@@ -16,7 +16,6 @@ from dashboard.tasks import (
 )
 from dashboard.factories import UserCacheRefreshTimeFactory
 from micromasters.factories import UserFactory
-from micromasters.utils import now_in_utc
 from search.base import MockedESTestCase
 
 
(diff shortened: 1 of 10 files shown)


dashboard/tasks_test.py
```diff
@@ -60,13 +59,7 @@ def test_batch_update_user_date_for_active_users(self, mocked_subtasks):
...
             user.is_active = False
             user.save()
-        now = now_in_utc()
-        UserCacheRefreshTimeFactory.create(
-            user=self.user2,
-            enrollment=now,
-            certificate=now,
-            current_grade=now,
-        )
+        UserCacheRefreshTimeFactory.create(user=self.user2, unexpired=True)
```
noisecapella: I think this should be `expired=True` since `now` would have become the past in the time that `batch_update_user_data` is called. I'm not sure it matters for this test though

gsidebo: From `dashboard.tasks.batch_update_user_data`:

```python
        refresh_time_limit = now_in_utc() - timedelta(hours=MAX_HRS_MAIN_TASK)

        users_expired_cache = UserCacheRefreshTime.objects.filter(
            Q(enrollment__lt=refresh_time_limit) |
            Q(certificate__lt=refresh_time_limit) |
            Q(current_grade__lt=refresh_time_limit),
            user__is_active=True,
            user__profile__fake_user=False
        ).values_list('user', flat=True).distinct()
```

These refresh times are considered expired if they are older than some number of hours _before_ the current time. If the refresh time is being set to now in our test cases, they're intended to be unexpired.

noisecapella: Oh right, makes sense

PR #3427, opened 2017-07-21
Added factories to produce social auth for Users

### LINKED ISSUE #3166: UserFactory does not create social auth
This means some data (e.g. ProfileSerializer) generates with `None` values.

gsidebo:
Closes #3166 

#### What's this PR do?
Adds factories to produce social auth for Users, and changes a lot of test cases to use those factories

No manual testing needed

`micromasters/factories.py` and `profiles/factories.py`

- Here's some example usage for the new factories:
    ```python
    # Create a user with a social auth provider 'abc'
    SocialUserFactory.create(username='user_0', social_auth__provider='abc')
    # Same as above, but create a Profile instead of a User
    SocialProfileFactory.create(first_name='Joe', user__username='user_0', user__social_auth__provider='abc')
    ```
- There are a bunch of other test cases that manually add social_auth records on their own, but this PR has enough lines as it is

### DIFF
--- a/micromasters/factories.py
+++ b/micromasters/factories.py
@@ -2,9 +2,40 @@
 Factory for Users
 """
 from django.contrib.auth.models import User
-from factory import Sequence
+from factory import (
+    Sequence,
+    SubFactory,
+    LazyAttribute,
+)
 from factory.django import DjangoModelFactory
 from factory.fuzzy import FuzzyText
+from social_django.models import UserSocialAuth
+from backends.edxorg import EdxOrgOAuth2
+from micromasters.utils import pop_matching_keys_from_dict
+
+
+def extract_related_model_kwargs(orig_kwargs, related_model_prop_name):
+    """
+    Extracts a set of factory kwargs that refer to a related model, removes the model prefix from the keys,
+    and returns the dictionary.
+    Ex. usage:
+        user_factory_kwargs = dict(username='user_0', social_auth__provider='edx', social_auth__uid='user_0')
+        extract_related_model_kwargs(user_factory_kwargs, 'social_auth')
+        #> dict(provider='edx', uid='user_0')
+
+    Args:
+        orig_kwargs (dict): A dict of kwargs being passed to a factory generation method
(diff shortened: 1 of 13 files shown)


micromasters/factories.py
```diff
@@ -14,3 +45,27 @@ class UserFactory(DjangoModelFactory):
...
     class Meta:
         model = User
+
+
+class UserSocialAuthFactory(DjangoModelFactory):
+    """Factory for UserSocialAuth"""
+    user = SubFactory(UserFactory)
+    provider = EdxOrgOAuth2.name
+    extra_data = {"access_token": "fooooootoken"}
+    uid = LazyAttribute(lambda social_auth: social_auth.user.username)
```
noisecapella: Can you change this to be different from `user.username`? The `_edx` suffix was originally added to help make it obvious if we leaked the Django username through the API instead of using the edX one

micromasters/factories.py
```diff
@@ -2,9 +2,40 @@
...
+
+    Returns:
+        dict: A dict of kwargs related to a specific model (without the model prefix)
+    """
+    related_model_prefix = '{}__'.format(related_model_prop_name)
+    extracted_kwargs = pop_matching_keys_from_dict(
+        orig_kwargs,
+        lambda key: key.startswith(related_model_prefix)
+    )
+    return {k.replace(related_model_prefix, ''): v for k, v in extracted_kwargs.items()}
```
noisecapella: This might do something unexpected if `related_model_prefix` exists twice in the prop name. Can you change this to `replace(related_model_prefix, '', 1)` and add a test for this edge case?

profiles/permissions_test.py
```diff
@@ -73,23 +74,12 @@ class CanSeeIfNotPrivateTests(MockedESTestCase):
     def setUp(self):
         super(CanSeeIfNotPrivateTests, self).setUp()
         with mute_signals(post_save):
```
noisecapella: Do we still need `mute_signals` here and below?

micromasters/factories.py
gsidebo: good catch 👍

Issue #3299, opened 2017-06-12
2 users missing social auth objects


pdpinch:
MultipleObjectsReturned: get() returned more than one UserSocialAuth -- it returned 2!

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/288659721/

```
MultipleObjectsReturned: get() returned more than one UserSocialAuth -- it returned 2!
  File "dashboard/tasks.py", line 101, in batch_update_user_data_subtasks
    UserSocialAuth.objects.get(user=user)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 389, in get
    (self.model._meta.object_name, num)

user "Juja08" does not have python social auth object
```

pdpinch:
I feel like this has happened before. We weren't able to identify the root cause, so we just repaired the users accounts.

pdpinch:
The last occurrence was in #2276

pdpinch:
@amir-qayyum-khan are you doing anything with this issue? If not, you should return it to ready so someone else can work on it.

amir-qayyum-khan:
```unique_together = ('provider', 'uid')```

https://github.com/python-social-auth/social-app-django/blob/master/social_django/models.py#L75

can i check records for `Juja08`

@pdpinch

pdpinch:
That user has two social auths: same username and provider, but different uid.

IIRC, @giocalitri speculated that that could happen if the user's email address changed at edX. In the past, we just deleted one of the social auths. 

I'm wondering if we could just use the most recent one? Or maybe change the unique_together to be provider and username (I'm a little scared of that).

pdpinch:
@amir-qayyum-khan points out that the 2nd social auth has a different username. I think this suggests that the user 

1. Logged in to MM through edX and created a MM account with 1st social auth
2. Logged out of edx.org, but not MM.
3. While still logged in to MM, somehow logged into edX again, with a different edX account. This creates second social auth

amir-qayyum-khan:
there is issue with https://micromasters.d.mitx.mit.edu/account/settings
@mitodl/devops @blarghmatey  can you check?

pdpinch:
Here are some more detailed steps to reproduce:

1. Log in to the MicroMasters portal with your 1st edX account
2. Logout of edX
3. Log out of the MicroMasters portal 
[...]

PR #3429, opened 2017-07-24
Fixed issue when a user have more the one social auth objects

amir-qayyum-khan:
fixed https://github.com/mitodl/micromasters/issues/3299

#### What's this PR do?
Allow users to have more the one social auth objects, previously application was throwing error when user have more the one social auth objects

- This PR changes the query on social auth table, Instead of looking for one object it filter out all objects of a user and the picks the latest.
- To test it create a duplicate social object you can create it manually, then see exceptions on console
- previously app was throwing exception when it was running `micromasters/dashboard/tasks.py`
- now it wont throw
- you will see dashboard like this
screen shot 2017-07-25 at 7 00 01 pm
- switch to my branch `fix/aq/duplicate_social_auth` and things will be fine.

@pdpinch

### DIFF
--- a/dashboard/api_edx_cache.py
+++ b/dashboard/api_edx_cache.py
@@ -12,7 +12,6 @@ from requests.exceptions import HTTPError
 
 from backends import utils
 from backends.exceptions import InvalidCredentialStored
-from backends.edxorg import EdxOrgOAuth2
 from courses.models import CourseRun
 from dashboard import models
 from micromasters.utils import now_in_utc
--- a/dashboard/api_edx_cache.py
+++ b/dashboard/api_edx_cache.py
@@ -16,7 +16,7 @@ from backends.edxorg import EdxOrgOAuth2
 from courses.models import CourseRun
 from dashboard import models
 from micromasters.utils import now_in_utc
-from profiles.api import get_social_username
+from profiles.api import get_social_username, get_social_auth
 from search import tasks
 
 log = logging.getLogger(__name__)
@@ -342,7 +342,7 @@ class CachedEdxDataApi:
             None
         """
         # get the credentials for the current user for edX
-        user_social = user.social_auth.get(provider=EdxOrgOAuth2.name)
+        user_social = get_social_auth(user)
         utils.refresh_user_token(user_social)
         # create an instance of the client to query edX
         edx_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL)
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -14,7 +14,6 @@ from edx_api.client import EdxApi
 from social_django.models import UserSocialAuth
 
 from backends import utils
-from backends.edxorg import EdxOrgOAuth2
 from dashboard.models import UserCacheRefreshTime
 from dashboard.api_edx_cache import CachedEdxDataApi
 from micromasters.celery import app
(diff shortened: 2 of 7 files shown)


pdpinch:
@amir-qayyum-khan can you add some more information about what this does and how to test it?

amir-qayyum-khan:
added @pdpinch

amir-qayyum-khan:
you can also produce duplication of social auth:

0) Switch to MM master
1) Login to MM
2) logout from edX instance in separate tab
3) on MM tab open this url https://micromasters-ci.herokuapp.com/login/edxorg, this will take you to edX login
4) on edX login screen choose create user and create new user.
5) when you are done it will take you to MM where you will see bug on dashboard
6) switch to my branch and every thing will be fine

dashboard/api_edx_cache.py
```diff
@@ -342,7 +342,7 @@ def update_all_cached_grade_data(cls, user):
             None
         """
         # get the credentials for the current user for edX
-        user_social = user.social_auth.get(provider=EdxOrgOAuth2.name)
+        user_social = user.social_auth.filter(provider=EdxOrgOAuth2.name).latest('id')
```
gsidebo: This ORM query is repeated in many places (and, of course, it was repeated before you made this fix). Ideally we would be able add a method to the model itself to run this query, but since `User` is provided by Django, I think a good alternative option is to add a new API method to `profiles/api.py`. 

```python
def get_social_auth(user):
    return user.social_auth.filter(provider=EdxOrgOAuth2.name).latest('id')
```

You could replace the queries in this PR with calls to that method, and `get_social_username` in the same module could call it directly

noisecapella:
I'm a little skeptical of the idea of just choosing the more recent social auth account. It seems like if the user created a new account on edX, we need to figure out how to get social_auth to create a new `User` to go along with that account. I'm not sure we should leave two `SocialAuth` objects attached to the same `User` and just ignore one.

gsidebo:
I don't love it either. I think it's a problem that a user can log into 2 different edX accounts and they are both associated with the same MM account according to our social auth plugin. That said, I _think_ it's safe to avoid the error behavior with this fix, but we should open up a new issue at the very least. That issue can look into avoiding the association of multiple edX accounts to a single MM account to begin with. If we're not in any kind of rush to fix the current error behavior, I would be fine with holding off on this PR until we figure out the core issue.

As for the "real" fix, I have a feeling that we will want to make some changes to the `SOCIAL_AUTH_PIPELINE` setting used by python-social-auth ([doc](http://python-social-auth.readthedocs.io/en/latest/configuration/django.html#personalized-configuration)), and possibly insert a new custom step in the pipeline (which we already do for different reasons).

noisecapella:
It looks like it's `associate_user` which is the problem. Can we do some investigating on fixing this there? It just seems pointless to change a whole bunch of code for a workaround if pipeline itself can be fixed

pdpinch:
Happy to discuss how to address the root cause, but we already have two accounts in this state that throw errors every six hours. I'd very much like that to stop. 

And addressing the root cause won't get these accounts out of their error state.

amir-qayyum-khan:
@gsidebo @noisecapella we can keep this PR or discard 
if discard then we can write a script to delete user with duplicate account, because i fount that if user has 2 social accounts he wont able to dashboard

gsidebo:
Unless I'm missing something, it seems like we should definitely merge a fix for the immediate problem (errors being thrown every 6 hours, inaccessible dashboard) and create a separate issue for the root cause

gsidebo:
Code looks great. I'll test the functionality once more after standup

PR #3444, opened 2017-07-31
Fix duplicate social auth creation during log in

### LINKED ISSUE #3432: Only allow one edX social auth per account
Each account on MicroMasters should be associated with one and only one account on edX. 

Under certain conditions (see #3299 and #2276) users have been able to create multiple edX social auth associations. This should be prevented.

noisecapella:
Fixes #3432 

#### What's this PR do?
Logs out the user in the social auth `complete` view before they log in. The root cause of #3299 was that the new social auth object was created for a new edX account, but it was attached to the currently logged in user even if that user isn't necessarily related to the edX user.

- First, log out in MM and edX.
- Go to edX and create a user with username `social_auth_1` and email `social_auth_1@example.com`. Log in on MM. Don't worry about filling out the profile.
 - Open the browser console and type `SETTINGS.user.username`. You should see `social_auth_1`.
- Log out on edX (but stay logged in on MM). Create a user with username `social_auth_2` and email `social_auth_2@example.com` on edX. On MM go to `/login/edxorg/`. You should see an empty profile.
 - Open the browser console and type `SETTINGS.user.username`. You should see `social_auth_2`.
 - Open the Django shell. Run these commands:

        def print_auth(user):
            print(user.username, user.social_auth.values_list('uid, 'provider'))
        
        print_auth(User.objects.get(username='social_auth_1'))
        print_auth(User.objects.get(username='social_auth_2'))

For the last two lines you should see:

    social_auth_1 <QuerySet [('social_auth_1', 'edxorg')]>
    social_auth_2 <QuerySet [('social_auth_2', 'edxorg')]>

- Now, go to edX and log out there (stay logged in on MM). Log in as the first user (`social_auth_1@example.com`). On MM go to `/login/edxorg/`.
- In your browser console type `SETTINGS.user.username`. You should see `social_auth_1`.
- In your Django shell run the last two lines again:

        print_auth(User.objects.get(username='social_auth_1'))
        print_auth(User.objects.get(username='social_auth_2'))

You should have this output (same as before):

    social_auth_1 <QuerySet [('social_auth_1', 'edxorg')]>
    social_auth_2 <QuerySet [('social_auth_2', 'edxorg')]>

### DIFF
--- a/backends/urls.py
+++ b/backends/urls.py
@@ -2,12 +2,12 @@
 from django.conf.urls import url
 
 from social_django import views
-from social_django.urls import extra
+from social_django.urls import (
+    app_name,  # pylint: disable=unused-import
+    extra,
+)
 
 from backends.views import complete
-
-
-app_name = 'social'
 
 urlpatterns = [
     # authentication / association
--- a/backends/urls.py
+++ b/backends/urls.py
@@ -9,13 +9,18 @@ from social_django.urls import (
 
 from backends.views import complete
 
+
+# These four endpoints were originally copied from social_django.urls. 'complete' is a modified view which
+# we are substituting to log out the user to workaround social auth creation issues.
 urlpatterns = [
-    # authentication / association
-    url(r'^login/(?P<backend>[^/]+){0}$'.format(extra), views.auth,
-        name='begin'),
     # Override default complete view to force logout before login
     url(r'^complete/(?P<backend>[^/]+)/$', complete,
         name='complete'),
+
+    # These three go to views in social_core
+    # authentication / association
+    url(r'^login/(?P<backend>[^/]+){0}$'.format(extra), views.auth,
+        name='begin'),
     # disconnection
     url(r'^disconnect/(?P<backend>[^/]+){0}$'.format(extra), views.disconnect,
         name='disconnect'),
--- a/backends/urls.py
+++ b/backends/urls.py
@@ -1,0 +1,27 @@
+"""URLs module"""
+from django.conf import settings
+from django.conf.urls import url
+
+from social_core.utils import setting_name
+from social_django import views
+
+from backends.views import complete
+
+
+extra = getattr(settings, setting_name('TRAILING_SLASH'), True) and '/' or ''
+
+app_name = 'social'
+
+urlpatterns = [
+    # authentication / association
+    url(r'^login/(?P<backend>[^/]+){0}$'.format(extra), views.auth,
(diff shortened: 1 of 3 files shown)


backends/urls.py
```diff
@@ -0,0 +1,24 @@
...
+
+from social_django import views
+from social_django.urls import extra
+
+from backends.views import complete
+
+
+app_name = 'social'
+
+urlpatterns = [
```
gsidebo: This isn't too crucial, but I think it would be good to either include a comment(s) indicating which URL rules we're lifting directly from the standard `social_django.urls` module, or import those rules directly and add this custom URL rule on top of them. Could look something like this:

```python
urlpatterns = (
  [pattern for pattern in social_django.urls.urlpatterns if pattern.name != 'complete'] +
  [url(r'^complete/(?P<backend>[^/]+)/$', complete, name='complete')]
)
```

Basically I think it would be helpful to be more explicit about the line between the library code and the custom code

backends/views.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@never_cache
+@csrf_exempt
+@psa('social:complete')
+def complete(request, *args, **kwargs):
+    """Override this method so we can force user to be logged out."""
+    if request.user.is_authenticated():
+        key = "{}_state".format(request.backend.name)
+        backend_state = request.session.get(key)
```
gsidebo: Just wondering: what's the significance of `backend_state` and the key that is being used? Is this something that `social_django` sets at some point?

backends/urls.py
```diff
@@ -0,0 +1,24 @@
+"""URLs module"""
+from django.conf.urls import url
+
+from social_django import views
+from social_django.urls import extra
+
+from backends.views import complete
+
+
+app_name = 'social'
```
gsidebo: Is this needed? Could you just import the `app_name` directly from `social_django.urls`?

noisecapella: Right, I'll do that

backends/urls.py
noisecapella: Sure

backends/views.py
noisecapella: The logout erases all session state but `social_django` still needs the state set in the session in order to complete authentication. At least, it raises an exception if it's not there.

gsidebo: Got it. Can you include a brief comment about that?

noisecapella:
Comments addressed

PR #3445, opened 2017-07-31
Raise an exception if there are two social auth objects

noisecapella:
Related to #3299 

#### What's this PR do?
Changes behavior to raise an exception if there is more than one social auth per user. This should only be merged after #3444 is merged and the existing duplicate social auth accounts are removed.

N/A

### DIFF
--- a/profiles/api_test.py
+++ b/profiles/api_test.py
@@ -3,6 +3,10 @@ Tests for profile functions
 """
 
 from unittest.mock import Mock
+
+from django.core.exceptions import MultipleObjectsReturned
+from testfixtures import LogCapture
+
 from backends.edxorg import EdxOrgOAuth2
 
 from profiles.api import get_social_username, get_social_auth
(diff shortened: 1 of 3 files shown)


pdpinch:
What's blocking this?

noisecapella:
We should first delete the duplicate social auth objects in production, but that's it

profiles/api_test.py
```diff
@@ -53,17 +57,31 @@ def test_one_social(self):
...
-        exists, it returns the most recently created
+        exists, it raises an exception
         """
         assert get_social_auth(self.user) == self.user.social_auth.get(provider=EdxOrgOAuth2.name)
-        new_social_auth = UserSocialAuthFactory.create(user=self.user, uid='other name')
-        assert get_social_auth(self.user) == new_social_auth
+        UserSocialAuthFactory.create(user=self.user, uid='other name')
+        with self.assertRaises(MultipleObjectsReturned) as ex:
+            get_social_auth(self.user)
+        assert ex.exception.args[0] == 'get() returned more than one UserSocialAuth -- it returned 2!'
```
gsidebo: I don't think this assertion is necessary

gsidebo:
Do we have a plan for correcting the production data? Will we just be deleting the most recent social auth for a given user? I don't think this should merge until we're sure

profiles/api_test.py
```diff
@@ -53,17 +57,31 @@ def test_one_social(self):
...
-        get_social_username should return latest social user name if
-        there are two social edX accounts for a user
+        get_social_username should return None if there are two social edX accounts for a user
         """
-        new_social_auth = UserSocialAuthFactory.create(user=self.user, uid='other name')
-        assert get_social_username(self.user) == new_social_auth.uid
+        self.user.social_auth.create(
+            provider=EdxOrgOAuth2.name,
+            uid='other name',
+        )
```
gsidebo: `UserSocialAuthFactory` can/should still be used here (`UserSocialAuthFactory.create(user=self.user, uid='other name')`)

PR #3486, opened 2017-08-15
Let user pay for a course if auditing

### LINKED ISSUE #3455: Learner is not offered to upgrade/pay when auditing course if already paid for previous runs
Case 1: I have passed the course in the past, now I am auditing the course, and I would like to upgrade.

screen shot 2017-08-07 at 10 30 59 am
screen shot 2017-08-07 at 10 31 07 am

Case 2: I have failed the course in the past, now I am auditing the course, and I would like to upgrade.

screen shot 2017-08-07 at 11 20 21 am
screen shot 2017-08-07 at 11 20 31 am

annagav:
Fix  #3455

#### What's this PR do?
Show the same course status message when user is auditing a course, independently of previous enrollments for that course.

Run
```
./manage.py alter_data set_past_run_to_passed --username staff --program-title Analog --course-title 100
./manage.py alter_data set_to_needs_upgrade --username staff --program-title Analog --course-title 100
```
You should be able to enroll in the course again.
screen shot 2017-08-15 at 4 12 40 pm

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages.js
+++ b/static/js/components/dashboard/courses/StatusMessages.js
@@ -116,14 +116,10 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
 
   // handle other 'in-progress' cases
   if (courseUpcomingOrCurrent(firstRun) || firstRun.status === STATUS_CAN_UPGRADE) {
-    if (hasPaidForAnyCourseRun(course)) {
-      messages.push({message: "You are auditing this course."});
-    } else {
-      messages.push({
-        message: "You are auditing. To get credit, you need to pay for the course.",
-        action: courseAction(firstRun, COURSE_ACTION_PAY)
-      });
-    }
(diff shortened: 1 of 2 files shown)


aliceriot:
I had thought that we did want to distinguish between the cases where the user had already enrolled once and paid, and was now re-enrolling, because of some users who take courses many times b/c they are acting as teaching assistants, or something along those lines. That's the reason this was initially implemented the way it is - @roberthouse54 is that no longer a concern? I'm just wondering if we may want to still have two messages, something like:

"You are auditing, you need to pay to get credit" and "You are auditing, and have paid previously. To get credit for the current run of the course you need to pay again"

just to remove more ambiguity, but idk exactly what we want to do.

roberthouse54:
I think we can use the same message for both cases. At this point, the user should have already seen messaging about re-enrolling, before they chose to re-enroll and audit, etc.

PR #3488, opened 2017-08-16
Complete freeze final grades task when cache refresh fails

### LINKED ISSUE #2657: Final Grades: deal with users who do not get a frozen grade because of 401 errors
Users who do not have valid credential for refreshing current grades and certificates cannot get a Frozen Final Grade.
This results in the `CourseRunGradingStatus` never setting the status for the course run to `complete`.
We should decide if skip these users or automatically do something else.

annagav:
Fix #2657

#### What's this PR do?
Counts the number of users where user authentication failed and completes the task in the next run of the task.

Make a past passed course run, set the freezable date

```from courses.models import *
course_run = CourseRun.objects.get(edx_course_key=course_key)

course_run.freeze_grade_date=now_in_utc()-timedelta(days=1)
course_run.save()
```

Try to freeze the grades.

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -6,6 +6,8 @@ from collections import namedtuple
 
 from django.contrib.auth.models import User
 
+from django_redis import get_redis_connection
+
 from dashboard.api_edx_cache import CachedEdxUserData, CachedEdxDataApi
 from dashboard.models import CachedEnrollment
 from grades.exceptions import FreezeGradeFailedException
@@ -165,6 +167,8 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
     try:
         CachedEdxDataApi.update_all_cached_grade_data(user)
     except Exception as ex:  # pylint: disable=broad-except
+        con = get_redis_connection("redis")
+        con.lpush('{}'.format(course_run.edx_course_key), user.id)
         if not raise_on_exception:
             log.exception('Impossible to refresh the edX cache for user "%s"', user.username)
             return
--- a/grades/api.py
+++ b/grades/api.py
@@ -16,6 +16,7 @@ from grades.models import (
     FinalGradeStatus,
 )
 
+CACHE_KEY_FAILED_USERS_BASE_STR = "failed_users_{0}"
 
 log = logging.getLogger(__name__)
 
@@ -168,7 +169,7 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
         CachedEdxDataApi.update_all_cached_grade_data(user)
     except Exception as ex:  # pylint: disable=broad-except
         con = get_redis_connection("redis")
-        con.lpush('{}'.format(course_run.edx_course_key), user.id)
+        con.lpush(CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key), user.id)
         if not raise_on_exception:
             log.exception('Impossible to refresh the edX cache for user "%s"', user.username)
             return
--- a/grades/tasks.py
+++ b/grades/tasks.py
@@ -8,6 +8,7 @@ from celery.result import GroupResult
 from django.contrib.auth.models import User
 from django.core.cache import caches
 from django.db.models import Count
+from django_redis import get_redis_connection
 
 from courses.models import CourseRun
 from grades import api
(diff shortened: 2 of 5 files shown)


grades/api.py
```diff
@@ -165,6 +167,8 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
     try:
         CachedEdxDataApi.update_all_cached_grade_data(user)
     except Exception as ex:  # pylint: disable=broad-except
+        con = get_redis_connection("redis")
+        con.lpush('{}'.format(course_run.edx_course_key), user.id)
```
giocalitri: question: why do you need to format an empty string?

grades/tasks.py
```diff
@@ -83,13 +84,21 @@ def freeze_course_run_final_grades(course_run_id):
     # extract the users to be frozen for this course
     user_ids_qset = api.get_users_without_frozen_final_grade(course_run).values_list('id', flat=True)
 
+    # find number of users for which cache could not be updated
+    con = get_redis_connection("redis")
+    failed_users_count = con.llen('{}'.format(course_run.edx_course_key))
     # if there are no more users to be froze, just complete the task
-    if not user_ids_qset.exists():
+    if not user_ids_qset.exists() or user_ids_qset.count() == failed_users_count:
```
giocalitri: I think you should check the list of IDs, not just the number.

grades/tasks.py
```diff
@@ -83,13 +84,21 @@ def freeze_course_run_final_grades(course_run_id):
...
+        log.info('Completing grading with %d users getting refresh cache errors', failed_users_count)
         CourseRunGradingStatus.set_to_complete(course_run)
         return
 
     # if the task reaches this point, it means there are users still to be processed
 
+    # clear the list for users for whom cache update failed
+    for _ in range(0, failed_users_count):
+        con.lpop('{}'.format(course_run.edx_course_key))
```
giocalitri: question: cannot you just delete the key with `con.del(course_run.edx_course_key)`?

grades/api.py
annagav: Oh, I think I was thinking maybe to prefix it with something. But I don't know if there is any benefit in doing that. So I'll remove it

giocalitri: well... it is actually a good idea to prefix it

annagav:
@giocalitri, addressed all comments

grades/tasks.py
```diff
@@ -83,13 +84,26 @@ def freeze_course_run_final_grades(course_run_id):
...
-    if not user_ids_qset.exists():
+    # find number of users for which cache could not be updated
+    con = get_redis_connection("redis")
+    failed_users_count = con.llen(api.CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key))
+
+    # get the list of users that failed authentication last run of the task
+    failed_users_list = list(map(int, con.lrange(
+        api.CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key), 0, failed_users_count)))
+    users_need_freeze = list(user_ids_qset)
+    users_left = [user_id for user_id in users_need_freeze if user_id not in failed_users_list]
```
giocalitri: you can use sets for this and do something like`set(list1) - set(list2)

grades/tasks.py
```diff
@@ -83,13 +84,26 @@ def freeze_course_run_final_grades(course_run_id):
...
 
-    # if there are no more users to be froze, just complete the task
-    if not user_ids_qset.exists():
+    # find number of users for which cache could not be updated
+    con = get_redis_connection("redis")
+    failed_users_count = con.llen(api.CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key))
+
+    # get the list of users that failed authentication last run of the task
+    failed_users_list = list(map(int, con.lrange(
+        api.CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key), 0, failed_users_count)))
```
giocalitri: nit: please format better this

grades/tasks.py
```diff
@@ -83,13 +84,26 @@ def freeze_course_run_final_grades(course_run_id):
...
+    # if there are no more users to be frozen, just complete the task
+    if not users_left:
+        log.info('Completing grading with %d users getting refresh cache errors', len(failed_users_list))
         CourseRunGradingStatus.set_to_complete(course_run)
         return
 
     # if the task reaches this point, it means there are users still to be processed
 
+    # clear the list for users for whom cache update failed
+    con.delete(api.CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key))
```
giocalitri: nit: Please create a variable with the formatted string to be used in the 3 places in this task.

annagav:
@giocalitri, ready for review

giocalitri:
Can you add a test for the `grades/api.py` to assert that if the refresh of the token fails, the user id will end up in redis?

giocalitri:
The rest of the code is fine: I'm going to test the functionality now

giocalitri:
Functionality looks good: once you add the missing test we can merge

giocalitri:
👍

PR #3490, opened 2017-08-17
Added task to generate course certificates

### LINKED ISSUE #3460: Enable MicroMasters certificate generation
_Part of #2552_

Dynamic HTML certificates were implemented in #3459. We need some method for generating certificates for users that have passed a course run in an FA program (preferably automatic). We have a few options here:

1. Signals - Any time a `FinalGrade` is created, and the course run is in an FA program, create a `MicromastersCourseCertificate` as well
1. Task - Create an async task that checks for `FinalGrade` records with an FA program course run and no associated `MicromastersCourseCertificate`

If it matters to us, we can also create a management command to generate these certificates on demand.

In order to receive a certificate, the learner must:
1. Pass the online course
2. Pass the proctored exam
3. Getting a passing (greater than 50%) final grade, which is a weighted average of 40% of the course grade and 60% of the proctored exam grade.

gsidebo:
Closes #3460 

#### What's this PR do?
Adds a scheduled task to generate course certificates

Certificates should be generated if a user has a passing `FinalGrade`, and either of the following are true:
- There are no exam runs scheduled for the given course
- There are exam runs for the given course, and the user has a passing `ProctoredExamGrade` for one of them

This script should generate 2 different certificates.
```python
import factory
from courses.models import Program
from exams.factories import ExamRunFactory
from grades import tasks
from grades.factories import (
    FinalGradeFactory,
    ProctoredExamGradeFactory,
)
from grades.models import MicromastersCourseCertificate

# Get your user and the financial aid-enabled program
user = User.objects.get(username='YOUR_USERNAME')
fa_program = Program.objects.filter(financial_aid_availability=True).first()
fa_courses = fa_program.course_set.all()
course_without_exam = fa_courses[0]
course_with_exam = fa_courses[1]

# Create passing FinalGrades
FinalGradeFactory.create(user=user, course_run=course_without_exam.courserun_set.first(), passed=True)
FinalGradeFactory.create(user=user, course_run=course_with_exam.courserun_set.first(), passed=True)

# Create an exam run for one of the courses, along with a passing ProctoredExamGrade
exam_run = ExamRunFactory.create(course=course_with_exam)
ProctoredExamGradeFactory.create(
    user=user, 
    course=course_with_exam, 
    exam_run=exam_run,
    passed=True
)

# Run the task
tasks.generate_course_certificates_for_fa_students.delay()

# After the task finished, see the generated certificates
MicromastersCourseCertificate.objects.all()
```

You will also need to have a program page in the CMS and at least one certificate signatory associated with the course (model: `CourseCertificateSignatories`). Navigate to `/cms/`, click "Explorer" and select the home page (this should be the only option that comes up). If you don't have a program page for "Digital Learning", click "Add Child Page" and create the program page for that program. At the bottom of the form to create/edit a program page, you should see "Course Certificate Signatories". For the "Course" option, select the `Course` that your user's `FinalGrade` is associated with. You'll need to add one or more of these to the program page in order to see certificates.

 You can navigate to a rendered certificate by visiting `http://<your_local_mm_url>/certificate/<certificate_hash>`.

`grades/tasks.py`

I also fixed some references to the old model name for the course certificates


amir-qayyum-khan:
url is not working 
http://192.168.99.100:8079/certificate/08899497d6ce05c47b0b317c4e63e548

```In [13]: MicromastersCourseCertificate.objects.all()[0].__dict__
Out[13]: 
{'_state': <django.db.models.base.ModelState at 0x7f769d6552b0>,
 'created_on': datetime.datetime(2017, 8, 21, 11, 40, 24, 132204, tzinfo=<UTC>),
 'final_grade_id': 88,
 'hash': '08899497d6ce05c47b0b317c4e63e548',
 'id': 1,
 'updated_on': datetime.datetime(2017, 8, 21, 11, 40, 24, 132989, tzinfo=<UTC>)}
```

grades/tasks.py
```diff
@@ -22,6 +28,57 @@
...
+    """
+    # Fetch FinalGrades in FA programs with no associated MicromastersCourseCertificate
+    final_grade_values = (
+        FinalGrade.objects
+        .filter(
+            course_run__course__program__financial_aid_availability=True,
+            passed=True,
+        )
+        .annotate(certificate_count=Count('micromasterscoursecertificate'))
+        .filter(certificate_count=0)
```
amir-qayyum-khan: we need test for certificate_count > 0

amir-qayyum-khan: need a test for exam_run_count==0

grades/tasks.py
```diff
@@ -22,6 +28,57 @@
...
+    )
+    # Filter the results for FinalGrades that should have a MicromastersCourseCertificate created.
+    # A FinalGrade should have a certificate if:
+    #   1. No exams were scheduled for the course, or
+    #   2. The course has a scheduled exam and the user has a passing ProctoredExamGrade in the course.
+    final_grades_to_certify = filter(
+        lambda final_grade_dict: (
+            final_grade_dict['exam_run_count'] == 0 or
+            ProctoredExamGrade.objects.filter(
+                user_id=final_grade_dict['user__id'],
```
amir-qayyum-khan: need a test where 
```python
ProctoredExamGrade.objects.filter(
   user_id=final_grade_dict['user__id'],
   course_id=final_grade_dict['course_run__course__id'],
   passed=True
).exists() == False
```

amir-qayyum-khan: when final_grade_dict['exam_run_count'] == 0 what that mean? i am assuming it mean that user never attempted an exam, 
So should we filter him? should it be `final_grade_dict['exam_run_count'] >0` ?

grades/factories.py
```diff
@@ -61,8 +61,8 @@ class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods,
         model = ProctoredExamGrade
 
 
-class MicromastersCertificateFactory(DjangoModelFactory):
-    """Factory for MicromastersCertificate"""
+class MicromastersCourseCertificateFactory(DjangoModelFactory):
```
gsidebo: These certificates and edX certificates serve different purposes. There isn't a scenario where we would want to retrieve a `MicromastersCourseCertificate` or an edX `Certificate` instead

gsidebo:
@rhysyngsun I forgot to ask last week for you to weigh in on this. The high-level logic for generating our own certificate for a user is the following:

1. The user has a passing `FinalGrade` in an FA-enabled program
2. Either one of the following is true:
  - The course in question does not have any exams
  - The course in question has exams, and the user has a passing `ProctoredExamGrade`

My logic to determine if a course does not have exams is basically to check if there are zero `ExamRun` records associated with the `Course`. Does that sound right to you, or is there a flag or something that indicates that the course is 'exam-enabled'?

grades/tasks.py
gsidebo: `final_grade_dict['exam_run_count'] == 0` means that the course has no exam runs, which means that we can generate a certificate based on the `FinalGrade` only. We don't need to check for exam grades since there are no exams

amir-qayyum-khan:
sound good, why link is not working for me, it is giving 404

gsidebo:
@amir-qayyum-khan Sorry! I forgot to include instructions for the CMS part of this. The course needs to have at least one `CourseCertificateSignatories` record for the given `Course`. That can be done in the CMS. If you need it, I can provide more detail shortly...

amir-qayyum-khan:
ok, sure, please provide me detail

rhysyngsun:
@gsidebo, cc @pdpinch 

> My logic to determine if a course does not have exams is basically to check if there are zero ExamRun records associated with the Course. Does that sound right to you, or is there a flag or something that indicates that the course is 'exam-enabled'?

I'm not convinced the presence of those records are a sufficient test. The records are meant to track the runs, not the requirement of the course for learners to take one. We'd need a new flag for that probably, but I understand that does complicate configuration of exams even further.

That said, if we chose to go with this approach we need to watch out for the race condition for new courses where the `ExamRun`s need to get configured before grades are finalized. Right now we'd probably be ok for the one program that is using exams, because all of the courses have had at least one `ExamRun`, but I'm a little concerned the record check might burn us down the road if a new MM program decides to use these exams.

gsidebo:
It would be simple enough to add a boolean flag on the `Course` model to indicate whether or not it is 'exam-enabled', back-fill any existing courses with associated `ExamRun`s to set that value to true, and change this query as part of a separate PR

amir-qayyum-khan:
so what you suggestion @pdpinch @rhysyngsun should i give thumbs up.

@gsidebo please tell me procedure to enable certificates

gsidebo:
@amir-qayyum-khan This comment: https://github.com/mitodl/micromasters/pull/3490#issuecomment-323786366. You just need to follow those steps, then navigate to the `/certificate/<hash>` URL to see your certificate

pdpinch:
Regarding the race condition mentioned in https://github.com/mitodl/micromasters/pull/3490#issuecomment-323794627 -- we have zero plans for new courses with exam at the moment, and the next course of any kind won't be added until about January 2018. 

Please open an issue to add a boolean as suggested, but we don't have to address for some time, if ever.

PR #3509, opened 2017-09-05
Filter on coupon id for automatic emails

### LINKED ISSUE #3508: Automatic coupon is being incorrectly marked as already redeemed
Automatic coupons do not have limits on how many other people can redeem them, but due to an incorrect filter this check wasn't done. This would prevent the coupon from being redeemable by a user if another user had already redeemed that coupon
See also https://odl.zendesk.com/hc/en-us/requests/10358

noisecapella:
Fixes #3508 

#### What's this PR do?
Fixes a filter bug where the user id was compared with the coupon id. This caused some coupons to incorrectly be marked as already redeemed, preventing users from using them.

Talk to me

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -317,7 +317,7 @@ class Coupon(TimestampedModel, AuditableModel):
                 True if the coupon is not automatic and it has already been redeemed by someone else
         """
         return (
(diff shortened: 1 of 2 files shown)


noisecapella:
Just revert the code change and make sure the test fails

Issue #3475, opened 2017-08-15
create channel API


pdpinch:
_part of #3467_

When a staff member creates a channel (based on a profile search query), the following actions should be taken:

- [ ] create channel 
- [ ] create percolate query, similar to #2704/PR #2728 
- [ ] subscribe all users in query result
- [ ] create users in open discussions as necessary

noisecapella:
Closed by #3527

PR #3514, opened 2017-09-07
Add REST API to create channels

noisecapella:
Part of #3475

#### What's this PR do?
Adds a REST API to create channels on the open-discussions server and store an associated percolate query used to subscribe users to that channel. This PR does not yet subscribe those users.

Go to `/api/v0/channels/` and create a channel. Use `{}` for the percolate query. You should see a response with an updated percolate query with the program selector. If you browse the reddit server you should see the new channel.


test_requirements.txt
```diff
@@ -6,8 +6,8 @@ ipdb
 nplusone>=0.8.1
 pdbpp
 pip-tools
-pylint==1.7.1
-pylint-django==0.7.2
+pylint==1.7.2
+ -e git+git://github.com/mitodl/pylint-django.git@2c7136f1cb3ff618aa08189c7fab4f8e08772eae#egg=pylint-django-0.7.2fix
```
noisecapella: This is a forked version from 0.7.2 which cherry picks a bug fix for a rare recursive pylint error

rhysyngsun:

discussions/api.py
```diff
@@ -112,3 +120,42 @@ def update_discussion_user(discussion_user):
...
+    """
+    client = get_staff_client()
+    response = client.channels.create(
+        title=title,
+        name=name,
+        public_description=public_description,
+        channel_type=channel_type,
+    )
+    if not response.ok:
+        raise ChannelCreationException("Error creating channel: {}".format(response.content))
```
rhysyngsun: My only feedback is that this line results in a 500 error from the API (e.g. put a space in the channel name for an invalid channel name), I suppose since this is an administrative API that we can be ok with this because it won't get used frequently, but just wanted to mention since i ran into it accidentally.

noisecapella: Yeah, I didn't want to overcomplicate the validation at this point for a use case that wasn't common

PR #3525, opened 2017-09-15
add package.json script for running JS tests in watch mode

aliceriot:

none

#### What's this PR do?

what it says on the can

you should be able to do `npm run watch` in your docker container. the tests should run in watch mode, and should all be passing. everything else should be unchanged.

### DIFF
--- a/static/js/global_init.js
+++ b/static/js/global_init.js
@@ -79,7 +79,15 @@ process.on('unhandledRejection', reason => { // eslint-disable-line no-unused-va
 
 // fix 'unknown prop' error
 import injectTapEventPlugin from 'react-tap-event-plugin';
-injectTapEventPlugin();
+
+// injectTapEventPlugin should only be called once, and it throws an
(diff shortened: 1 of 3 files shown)


static/js/global_init.js
```diff
@@ -79,7 +79,11 @@ process.on('unhandledRejection', reason => { // eslint-disable-line no-unused-va
 
 // fix 'unknown prop' error
 import injectTapEventPlugin from 'react-tap-event-plugin';
-injectTapEventPlugin();
+
+try {
```
noisecapella: Why was this needed?

aliceriot: it throws an error if it's already been called

aliceriot: (and this file is basically re-required every time mocha goes to run the tests)

noisecapella: Can you add a comment to explain the try block?

aliceriot:
@noisecapella comments addressed

PR #3527, opened 2017-09-15
Add contributors when new channel is created

noisecapella:
Part of #3475

#### What's this PR do?
Adds contributors when a new channel is created, and synchronizes new contributors when the account is created or updated

Creating a channel should populate contributors based on the provided query:
 - Seed your database using seed_db
 - Go to `/learners`, open your network tab and execute some search targeting a handful of people with some substring of their name, for example `fl`. Copy the search query from the request.
 - Go to `/api/v0/channels/` and fill out the form. Paste the query into the search query field. You should get a 201 with information about the new channel
 - On open-discussions go to `/api/v0/channels/<channel_name>/contributors/` and verify that there are the right number of users (all the users in your search plus one for the API user)

When a user updates their profile it should unenroll or enroll them via profile signal:
 - Pick a user who matched the search criteria earlier. Log in as that user on reddit. You can change the password like this:
    
        from r2.models.account import *
        a = Account._by_name("your_username_here")
        a.password = bcrypt_password("your new password")
        a._commit()

You will also need to edit the max length for the input field on the login page to allow the ULID

 - Update the profile so that their name does not fit the search criteria anymore. Verify that the user is now not a contributor and they do not see that channel in their list of subscriptions.
  - Update the profile so that their name is now in the search criteria again. Verify that the user is a member of the channel again.

### DIFF
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -92,22 +92,23 @@ def create_discussion_user(discussion_user):
     profile = discussion_user.user.profile
 
     api = get_staff_client()
-    try:
-        result = api.users.create(
-            name=profile.full_name,
-            image=profile.image.url if profile.image else None,
-            image_small=profile.image_small.url if profile.image_small else None,
-            image_medium=profile.image_medium.url if profile.image_medium else None,
-        )
-        result.raise_for_status()
+    result = api.users.create(
+        name=profile.full_name,
+        image=profile.image.url if profile.image else None,
+        image_small=profile.image_small.url if profile.image_small else None,
+        image_medium=profile.image_medium.url if profile.image_medium else None,
+    )
 
-        discussion_user.username = result.json()['username']
-        discussion_user.last_sync = profile.updated_on
-        discussion_user.save()
+    try:
+        result.raise_for_status()
     except HTTPError as ex:
         raise DiscussionUserSyncException(
             "Error creating discussion user for {}".format(profile.user.username)
         ) from ex
+
+    discussion_user.username = result.json()['username']
+    discussion_user.last_sync = profile.updated_on
+    discussion_user.save()
 
 
 def update_discussion_user(discussion_user):
@@ -126,22 +127,23 @@ def update_discussion_user(discussion_user):
         return
 
     api = get_staff_client()
-    try:
-        result = api.users.update(
-            discussion_user.username,
-            name=profile.full_name,
-            image=profile.image.url if profile.image else None,
-            image_small=profile.image_small.url if profile.image_small else None,
-            image_medium=profile.image_medium.url if profile.image_medium else None,
-        )
-        result.raise_for_status()
+    result = api.users.update(
+        discussion_user.username,
+        name=profile.full_name,
+        image=profile.image.url if profile.image else None,
+        image_small=profile.image_small.url if profile.image_small else None,
+        image_medium=profile.image_medium.url if profile.image_medium else None,
+    )
 
-        discussion_user.last_sync = profile.updated_on
-        discussion_user.save()
+    try:
+        result.raise_for_status()
     except HTTPError as ex:
         raise DiscussionUserSyncException(
             "Error updating discussion user for {}".format(profile.user.username)
         ) from ex
+
+    discussion_user.last_sync = profile.updated_on
+    discussion_user.save()
 
 
 def add_to_channel(channel_name, discussion_username):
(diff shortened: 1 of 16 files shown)


discussions/api.py
```diff
@@ -155,7 +284,48 @@ def add_channel(original_search, title, name, public_description, channel_type):
...
-        return Channel.objects.create(
+        channel = Channel.objects.create(
             query=percolate_query,
             name=name,
         )
+
+    # Do a one time sync of all matching profiles
+    user_ids = search_for_field(updated_search, 'user_id')
+    from discussions import tasks
+    tasks.add_contributors.delay(channel.name, user_ids)
```
rhysyngsun: This is now erroring because `user_ids` is a `set`, which isn't JSON serializable.

noisecapella: Ah right

discussions/api.py
```diff
@@ -105,24 +126,122 @@ def update_discussion_user(discussion_user):
...
+def sync_user_to_channels(user_id):
+    """
+    Make sure the user's profile exists on open-discussions,
+    then update user's subscription and contributor status.
+
+    Args:
+        user_id (int): A user id
+    """
+    if not settings.FEATURES.get('OPEN_DISCUSSIONS_USER_SYNC', False):
+        return
```
giocalitri: nit: maybe we should add some logging here otherwise we will not know that it does not work because the feature is not enabled

discussions/api.py
```diff
@@ -155,7 +284,48 @@ def add_channel(original_search, title, name, public_description, channel_type):
...
         )
-        return Channel.objects.create(
+        channel = Channel.objects.create(
             query=percolate_query,
             name=name,
         )
+
+    # Do a one time sync of all matching profiles
+    user_ids = list(search_for_field(updated_search, 'user_id'))
+    from discussions import tasks
```
giocalitri: is this to avoid circular imports?

discussions/api_test.py
```diff
@@ -138,20 +149,158 @@ def test_update_discussion_user_no_update(mock_staff_client):
...
+    )
+
+    mock_staff_client.channels.add_contributor.assert_called_once_with(channel_name, discussion_username)
+    mock_staff_client.channels.add_subscriber.assert_called_once_with(channel_name, discussion_username)
+
+
+@pytest.mark.parametrize("contributor_status_code,subscriber_status_code", [
+    (200, 200),
+    (404, 404),
+    (409, 404),
```
giocalitri: can we use `rest_framework.status` instead of pure numbers? (same for all the other tests)

discussions/tasks.py
```diff
@@ -44,3 +44,19 @@ def sync_discussion_users():
             api.create_or_update_discussion_user(user_id)
         except DiscussionUserSyncException:
             log.error('Impossible to sync user_id %s to discussions', user_id)
+
+
+@app.task()
+def add_contributors(channel_name, user_ids):
+    """
+    Add users to a open-discussions channel as contributors and subscribers
```
giocalitri: nit: IMHO the name of the task is misleading because it sounds like it takes care only of contributors

discussions/api.py
```diff
@@ -155,7 +284,48 @@ def add_channel(original_search, title, name, public_description, channel_type):
...
+    tasks.add_contributors.delay(channel.name, user_ids)
+    return channel
+
+
+def add_contributors(channel_name, user_ids, retries=3):
+    """
+    Add users to a open-discussions channel as contributors and subscribers
```
giocalitri: Nit: IMHO function name is misleading because it talks only about contributors

profiles/signals.py
```diff
@@ -8,7 +8,7 @@
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
-from .models import Profile
+from profiles.models import Profile
```
giocalitri: +1 for fixing relative imports

discussions/tasks.py
noisecapella: Right, I'll change it to `add_users_to_channel`

discussions/api.py
noisecapella: Yes

discussions/api.py
noisecapella: Sure, I'm also going to move this check into `search/tasks.py` since all other checks for this feature flag occur in various tasks

noisecapella:
Should be ready for review again

giocalitri:
👍 works as expected

Issue #3476, opened 2017-08-15
"Recent Posts" react component 


pdpinch:
See https://invis.io/NB7VM1GSG#/242718039_Dashboard_with_Discussions_card

In the interests of MVP, we can omit some fields/content that appear in the mock-up.

pdpinch:
@aliceriot please let me know if this issue will depend on #3477 and #3478

aliceriot:
It depends how the authentication stuff is set-up so far. I think @rhysyngsun is probably in a better place to answer these questions - I'm not sure, for instance, that we are in a position to directly make an API call from the frontend to open-discussions, or whether it needs to go through MM first, to add profile information and so on.

rhysyngsun:
@pdpinch @aliceriot I'm not sure what #3477 is supposed to be. I thought we had discussed that we didn't need #3478 and would instead go directory to `open-discussions` API from the browser, right?

aliceriot:
Yeah that was my recollection as well

pdpinch:
I updated the description of this with a link to the invision mock-up. 

If you find that any part of the design is slowing you down, let's talk. We don't necessarily need all the elements.

PR #3530, opened 2017-09-18
simple discussions link

aliceriot:

first draft of #3476 

#### What's this PR do?

this adds a very simple card to the dashboard page that links the user over to discussions

if you open up the dashboard you should see something like this:

fjfjfjfffffawefawefd

clicking on the link should open the discussions app in a new tab, and you should be signed in, able to post, comment, etc.

### DIFF
--- a/static/js/components/DiscussionCard.js
+++ b/static/js/components/DiscussionCard.js
@@ -1,0 +1,24 @@
+// @flow
+import React from "react"
+import { Card } from "react-mdl/lib/Card"
+import Icon from "react-mdl/lib/Icon"
+
+import type { Program } from "../flow/programTypes"
+
+type DiscussionCardProps = {
+  program: Program
+}
+
+const DiscussionCard = ({ program }: DiscussionCardProps) => (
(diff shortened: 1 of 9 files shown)


static/js/components/DiscussionCard.js
```diff
@@ -0,0 +1,25 @@
...
+import Icon from 'react-mdl/lib/Icon';
+
+import type { Program } from '../flow/programTypes';
+
+type DiscussionCardProps = {
+  program: Program
+};
+
+const DiscussionCard = ({program}: DiscussionCardProps) =>
+  <Card className="discussion-card" shadow={0}>
```
rhysyngsun: Could you gate this with a feature flag so we can have it hidden until OD goes live?

aliceriot: yeah I think we should do that too, will do

aliceriot:
@rhysyngsun I added a feature flag (`FEATURE_OPEN_DISCUSSIONS_POST_UI`)

PR #3535, opened 2017-09-18
Upgraded yarn and Node to the latest version

giocalitri:
N/A

#### What's this PR do?
Upgrades yarn and node to the latest version

Everything should keep working as usual

NOTE: I already updated the image on docker hub

### DIFF
--- a/Dockerfile-node
+++ b/Dockerfile-node
@@ -1,9 +1,7 @@
-FROM node:6.2
+FROM node:8.5
 LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
 
 RUN apt-get update && apt-get install libelf1
-
-RUN adduser --disabled-password --gecos "" mitodl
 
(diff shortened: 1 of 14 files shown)


aliceriot:
you said in Slack that a few tests aren't passing after the upgrade - in that case, what we should do is change the name of the container that we push to docker hub (with `Dockerfile-travis-watch-build`) to be something like `mitodl/mm_watch_travis_next`, so that we can build a new image using the newer version, and get tests passing on that. then, after it merges, in like a week or two when all existing PRs have been either merged or rebased over the changes, we can rename the image back to `mitodl/mm_watch_travis`.

giocalitri:
I pushed a old version of the container and in the meantime George is taking a look at the tests

aliceriot:
cool

noisecapella:
I pushed a commit to fix a couple race conditions affecting the tests

PR #3528, opened 2017-09-15
Add template for MM Program certificate

annagav:
#### What are the relevant tickets?
Fix #3519

#### What's this PR do?
Add cms model for `ProgramCertificateSignatories`, add a view for program certificate.

#### How should this be manually tested?
create a program certificate 
```from django.contrib.auth.models import User
user = User.objects.filter(username='username')[0]

from courses.models import *
from grades.models import *
from grades.factories import *

program = Program.objects.filter(title='Digital Learning')[0]
for c in program.course_set.all():

    course_run = c.courserun_set.all().first()

    ### Create grade, complete status, and certificate
    final_grade = FinalGradeFactory.create(user=user, course_run=course_run, passed=True, status='complete', grade=0.8)
    CourseRunGradingStatus.objects.create(course_run=course_run, status='complete')
    cert = MicromastersCourseCertificate.objects.create(final_grade=final_grade)
```
Once all the courses in the program have a certificate, the program certificate should get created automatically.
In the cms admin create `ProgramCertificateSignatories` for the program. And then you should be able to view the certificate.

PR #3546, opened 2017-09-21
Show program certificate on dashboard

### LINKED ISSUE #3521: add link on ProgressWidget to program certificate
As a learner, I'd like to see a link to my program certificate on my dashboard. 

Screen mock-up at https://invis.io/NB7VM1GSG#/239889419_Program_Complete1

annagav:
Fix #3521

#### What's this PR do?
If the learner earned a program certificate show it in the ProgressWidget.

Create a program certificate (follow instructions in this PR #3528). You should be able to see in your dashboard.

screen shot 2017-09-21 at 4 21 41 pm


gsidebo:
If a `MicromastersProgramCertificate` exists for a user, but there is no signatory set in the CMS, we are providing the link to the certificate which will then return a 404 error because of the lack of signatories. Definitely need to fix that. If I'm thinking about this correctly, we just need to check if 1+ signatories exist in order to show the certificate link. We might be able to use `annotate` to get the signatory count in a similar way to this function: `grades.tasks.generate_course_certificates_for_fa_students`

static/js/components/ProgressWidget_test.js
```diff
@@ -141,4 +142,21 @@ describe("ProgressWidget", () => {
     )
     assert.equal(wrapper.find(".circular-progress-widget-txt").text(), "3/5")
   })
+  it("progress widget display program certificate", () => {
```
gsidebo: Can you change this text to `"should display program certificate when a certificate link exists"`

PR #3551, opened 2017-09-22
recent posts component

### LINKED ISSUE #3476: "Recent Posts" react component

aliceriot:

closes #3476 

#### What's this PR do?

This improves the existing 'discussions' card to implement the mocks, by hitting the OD backend to fetch the frontpage and then rendering it.

Make sure you have all the stuff set up for OD and MM to talk to each other. You should also have `FEATURE_OPEN_DISCUSSIONS_POST_UI` set to `True`. Then you should see something like this on the dashboard:

obama

If you don't set the `POST_UI` feature flag or you don't set the redirect URL env var you shouldn't see the card. If you do, you should!


pdpinch:
I added the Needs Review label. If that was premature, you can remove it.

static/js/lib/auth.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import "isomorphic-fetch"
+
+export const refreshAuthToken = () => {
```
noisecapella: Can you write tests for these two functions?

aliceriot: :woman_facepalming:

aliceriot: I even had that in my checklist when I was working and just totally forgot, haha

aliceriot:
@noisecapella I added a few tests that should cover the functions in `lib/auth.js`

static/js/lib/auth.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const fetchJSONWithAuthToken = async (url: string) => {
+  let response = await fetch(url, { credentials: "include" })
+
+  if (response.status === 401) {
+    await refreshAuthToken()
+    response = await fetch(url, { credentials: "include" })
+  }
+
+  if (response.status < 200 || response.status > 300) {
```
noisecapella: Can you change the expression to be `>= 300` just to be clear that `300` is unexpected?

aliceriot:
@noisecapella I fixed that check and updated the relevant test

PR #3552, opened 2017-09-25
Remove afterImageUpload callback which was erroring and is unnecessary

### LINKED ISSUE #3544: Bug in image loading when filling out profile
On micromasters-ci

screenshot from 2017-09-20 18-34-29

screenshot from 2017-09-20 18-36-04

However the image does show up in the profile and it gets saved correctly

noisecapella:
Fixes #3544 

#### What's this PR do?
Fixes a console error when uploading the profile image

Go to `/profile` and upload an image. You should see a console error when on master but not on this branch

### DIFF
--- a/static/js/components/PersonalTab.js
+++ b/static/js/components/PersonalTab.js
@@ -18,7 +18,6 @@ import {
 } from "../lib/validation/profile"
 import type {
   Profile,
-  ProfileFetchResponse,
   SaveProfileFunc,
   ValidationErrors,
   UpdateProfileFunc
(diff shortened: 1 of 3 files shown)


aliceriot:
I think I remember why this was necessary (I think re-patching the profile could overwrite the image or something), do you know what changed so that we don't need it anymore?

noisecapella:
I think we ended up not needing validation for the profile image, not 100% sure though

aliceriot:
oh right, that does sound familiar

PR #3553, opened 2017-09-26
Add shell with imported models

noisecapella:
Any opinions about this PR?

None

#### What's this PR do?
Adds a script `repl.py` which runs the Django shell with all models imported.

Run `docker-compose run web ./repl.py` and type `Role`. You should get the name of the model


pdpinch:
Certainly would save me time. 

Although it does increase the size of the gun I can shoot myself in the foot with.

PR #3567, opened 2017-09-27
Overall final grade for course

### LINKED ISSUE #3492: Add overall final grade data to the dashboard API
The concept of an 'overall' final grade (a grade that combines a user's course final grade [`FinalGrade`] and exam grade [`ProctoredExamGrade`] with appropriate weight) is only represented in front end code right now. This was done to speed up implementation. All of our other grading logic is in the back end, and all other grading data can be retrieved via the dashboard API. This split of front end/back end logic is inconvenient, difficult to reason about and potentially dangerous. The logic for calculating overall final grades should be moved to the back end, and overall final grade data should be returned in the dashboard API response.

annagav:
Fix #3492

#### What's this PR do?
Added overall final grade data to the dashboard API.

Set a past course run to passed and create an exam run with a passing exam grade. you should be able to see the final grade for course just as before.

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -418,7 +418,7 @@ class MMTrack:
             )
             return ExamProfile.PROFILE_INVALID
 
-    def get_best_proctorate_exam_grade(self, course):
+    def get_best_proctored_exam_grade(self, course):
 
         """
         Returns the best exam grade
@@ -429,7 +429,9 @@ class MMTrack:
         Returns:
             grades.models.ProctoredExamGrade: the best exam grade
         """
-        return self.get_course_proctorate_exam_results(course).filter(passed=True).order_by('percentage_grade').last()
+        return self.get_course_proctorate_exam_results(course).filter(
+            passed=True
+        ).order_by('-percentage_grade').first()
 
     def get_course_proctorate_exam_results(self, course):
         """
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -418,6 +418,19 @@ class MMTrack:
             )
             return ExamProfile.PROFILE_INVALID
 
+    def get_best_proctorate_exam_grade(self, course):
+
+        """
+        Returns the best exam grade
+
+        Args:
+            course (courses.models.Course): a course
+
+        Returns:
+            grades.models.ProctoredExamGrade: the best exam grade
+        """
+        return self.get_course_proctorate_exam_results(course).filter(passed=True).order_by('percentage_grade').last()
+
     def get_course_proctorate_exam_results(self, course):
         """
         Returns the queryset of the proctorate exams results for the user in a course
--- a/static/js/components/dashboard/courses/Grades.js
+++ b/static/js/components/dashboard/courses/Grades.js
@@ -5,7 +5,7 @@ import Icon from "react-mdl/lib/Icon"
 
 import type { Course } from "../../../flow/programTypes"
 import { formatGrade } from "../util"
-import { S, reduceM } from "../../../lib/sanctuary"
+import { S, reduceM, getm } from "../../../lib/sanctuary"
 import { classify } from "../../../util/util"
 import {
   getLargestExamGrade,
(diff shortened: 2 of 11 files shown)


static/js/components/dashboard/courses/Grades.js
```diff
@@ -57,7 +56,8 @@ const renderFinalGrade = R.ifElse(
   R.compose(
     reduceM("--", renderGrade("Final Grade")),
     S.map(formatGrade),
-    calculateFinalGrade
+    S.filter(R.complement(R.equals(""))),
```
gsidebo: I think this can be replaced with `filterEmpty` in `static/js/lib/grades.js`

gsidebo:
There are a couple things jumping out at me in terms of front-end logic that still does grade calculation that should be provided by the dashboard API.

In `static/js/components/dashboard/courses/Grades.js`, the `renderPassed` function calls `passedCourse` in `static/js/lib/grades.js`, which checks whether a course has an exam or not. This should not be a concern of the front end anymore. The overall final grade returned by the API should be the final word on that.

`static/js/lib/grades.js` also has a function `renderFinalGrade`, which is also called in `static/js/components/dashboard/courses/Grades.js`. That function first checks `hasPearsonExam` to determine what to show in the overall final grade. Same idea here. Unless I'm missing some logic specific to this case, the front end should not be deciding what overall final grade to show. I don't know how many other similar bits of front end functionality are still hanging around, but we should be careful to find them all

dashboard/utils.py
```diff
@@ -418,6 +418,19 @@ def get_pearson_exam_status(self):  # pylint: disable=too-many-return-statements
...
+        """
+        Returns the best exam grade
+
+        Args:
+            course (courses.models.Course): a course
+
+        Returns:
+            grades.models.ProctoredExamGrade: the best exam grade
+        """
+        return self.get_course_proctorate_exam_results(course).filter(passed=True).order_by('percentage_grade').last()
```
gsidebo: Another thing that I discovered recently: We should only be considered `ProctoredExamGrade`s with an associated `ExamRun` that has a `date_grades_available` in the past. We might want to create a `ProctoredExamGradeQuerySet`, a bit like `FinalGradeQuerySet` is for `FinalGrade`. That could implement something like `with_available_grade` to check for that `ExamRun` date, and also some other helpful methods like `passed`.

annagav:
@gsidebo, In `renderFinalGrade`, it decides based on `hasPearsonExam` to display a placeholder or to not display anything for final grade. I think it makes sense, unless we think we would like to display a placeholder for the final grade, independently of 'has_exam'.

static/js/components/dashboard/courses/Grades.js
annagav: @aliceriot recommended to use `R.equals("")` to be more explicit for what we are checking here.

gsidebo:
Given a course with exams, if a user has a course final grade and no exam grade, then we shouldn't return anything for the overall final grade in the API. On the front end, we should just be showing the overall final grade if the API returned something for it. It's part of the effort to get rid of the logic split between the back end and front end

annagav:
How would you know when to show a placeholder `--` for final grade or not to show the grade box at all?

dashboard/utils.py
annagav: This is the queryset function `get_course_proctorate_exam_results`. It calls `for_user_course` class method that filters on `exam_run__date_grades_available__lte=now`.

pdpinch:
@annagav can we incorporate a check on include in this a check on the [examrun.date_grades_available](https://github.com/mitodl/micromasters/blob/master/exams/models.py#L22) in this calculation?

annagav:
@pdpinch, The `get_course_proctorate_exam_results` uses `ProctoredExamGrade.for_user_course` which filters for `exam_run__grades_date_available` to earlier than today.

gsidebo:
@annagav Got it. I was making some different assumptions about the UI. This might be something that we want to talk to @roberthouse54 about. Whatever the actual overall final grade for a course is, I think it makes sense to show that to the user in the same place. With the current setup a user could have 2 courses - one with exams, one without - and their effective overall final grade for each course in two different ways. I think we should have a consistent display for effective final grade whether the course has an exam or not. That can be done in a separate issue though. I'll create it

PR #3580, opened 2017-09-28
Link channels to users and add all staff as moderators of channel

### LINKED ISSUE #3574: When creating a channel, subscribe all staff in the program
As a staff, I'd like to be subscribed to any new channels in my program. I'm already a moderator, but I won't see posts in the new channel unless I'm subscribed. 

Since staff are omitted from the queries that are used to create channels, we just need to subscribe them after everyone in the query results in subscribed.

noisecapella:
Fixes #3574 

#### What's this PR do?
 - Adds many to many relationship between `Channel` to `Program`
 - Users who are staff or instructors of any program linked to the channel are added as moderators when the channel is created
 - When a user's profile is updated and they are moderator of a channel, we do not update their subscription status for that channel

Create a channel and verify that you see the channel in your subscribed channels list afterwards. Update your profile and verify that the channel is still in your subscribed channels.


Ferdi:
Why should we explicitly make a link between Chanel and program? Is it necessary?

noisecapella:
It's so we can add users who are staff of a program as moderators, and maintain that status when the program enrollment information is updated

noisecapella:
@Ferdi After talking with @pdpinch the database model was changed to link a channels with multiple programs

Ferdi:
I can't look at the pr now but it looks to me that we are including MM internals (programs) in open discussion. Another approach is to use MM program staff as one time seed for channel moderators. Later we could create a ui in open discussion for adding moderators. Channel moderators should not be restricted to just staff

noisecapella:
We need some way to keep track of moderators so that they remain subscribed even if their profile doesn't match with the percolate query

discussions/api.py
```diff
@@ -163,51 +165,110 @@ def add_to_channel(channel_name, discussion_username):
...
+    Remove subscriber from a channel
+
+    Args:
+        channel_name (str): An open-discussions channel
+        discussion_username (str): The username used by open-discussions
+    """
+    admin_client = get_staff_client()
+
+    # If the channel is private and the user is not a contributor, their subscription status will always
+    # look like it's false. To work around this we always remove the subscriber first.
```
annagav: I feel like this comment makes more sense in `remove_from_channel`

discussions/api.py
```diff
@@ -221,25 +282,43 @@ def sync_user_to_channels(user_id):
...
+    # Make sure we don't include extra channels beyond what's in all_channel_ids
+    membership_channel_ids = membership_channel_ids.intersection(all_channel_ids)
 
-    all_channel_ids = set(Channel.objects.values_list("id", flat=True))
     membership_channels = Channel.objects.filter(id__in=membership_channel_ids)
     for channel in membership_channels:
-        add_to_channel(channel.name, discussion_user.username)
+        # This is done in this order because a user cannot be added as a subscriber
+        # to a private channel without first being a contributor
+        add_contributor_to_channel(channel.name, discussion_user.username)
```
annagav: why not use `add_to_channel` here?

discussions/api.py
```diff
@@ -221,25 +282,43 @@ def sync_user_to_channels(user_id):
...
+        add_subscriber_to_channel(channel.name, discussion_user.username)
 
     nonmembership_channels = Channel.objects.filter(
         id__in=all_channel_ids.difference(membership_channel_ids)
     )
     for channel in nonmembership_channels:
-        remove_from_channel(channel.name, discussion_user.username)
+        # This is done in this order because a user can't be removed as a subscriber
+        # if they are no longer a contributor
+        remove_subscriber_from_channel(channel.name, discussion_user.username)
```
annagav: and `remove_from_channel` here?

noisecapella:
Comments addressed, looks like I overlooked those other methods

annagav:
Works great!

PR #3605, opened 2017-10-05
Allow empty public_description

annagav:
Depends on https://github.com/mitodl/open-discussions/pull/254

#### What's this PR do?
Make public_description not required

Should be able to create a channel with empty description.


noisecapella:
Can you update a test to omit `public_description` to verify that it's not required?

PR #3609, opened 2017-10-05
Upgrade eslint configuration and fix throw literal warnings

noisecapella:
None

#### What's this PR do?
Upgrades eslint configuration, fixes related warnings

N/A

### DIFF
--- a/static/js/dashboard_routes.js
+++ b/static/js/dashboard_routes.js
@@ -13,7 +13,7 @@ import Learner from './components/Learner';
 import OrderSummaryPage from './containers/OrderSummaryPage';
 
 const errorLoading = error => {
-  console.trace(error);
+  console.log("loading error caused by:", error)
   throw new Error(`Dynamic page loading failed: ${error}`);
 };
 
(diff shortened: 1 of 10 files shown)


static/js/dashboard_routes.js
```diff
@@ -13,7 +13,7 @@ import Learner from './components/Learner';
 import OrderSummaryPage from './containers/OrderSummaryPage';
 
 const errorLoading = error => {
-  console.trace(error);
+  console.log("loading error caused by:", error)
```
amir-qayyum-khan: why console here?

noisecapella: `console.trace` was showing the stack trace of `errorLoading`, which doesn't contain any useful information. Instead we want to see the stack trace of `error` which shows up via `console.log`

amir-qayyum-khan:
@noisecapella just out of curiosity, You choosed `thorw new Error` which logs `error , (custom message)` to be more readable?

noisecapella:
No, in javascript the stack trace of an `Error` object is defined at the point when the `Error` was created, not when it was thrown. In order to get the stack trace of a caught error it needs to be an `Error` object.

PR #3611, opened 2017-10-05
Add payment deadline to course status

### LINKED ISSUE #3516: Display payment deadline on dashboard
The customer support team is getting a lot of questions about the specific day and time when payment is due. It would help them if this information was displayed -- on the dashboard and/or the program page. 

See [ZenDesk #10860](https://odl.zendesk.com/agent/tickets/10860)

annagav:
Fix  #3516

#### What's this PR do?
Display payment deadline for course 

Use this command ```./manage.py alter_data set_past_run_to_passed --username staff --course-title 'Analog Learning 200' --audit```

You should be able to see the payment deadline

screen shot 2017-10-05 at 3 01 16 pm

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages.js
+++ b/static/js/components/dashboard/courses/StatusMessages.js
@@ -95,6 +95,7 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
   const passedExam = hasPassingExamGrade(course)
   const failedExam = hasFailingExamGrade(course)
   const paymentDueDate = moment(firstRun.course_upgrade_deadline)
+  const dueDate = paymentDueDate.format(COURSE_CARD_FORMAT)
   if (firstRun.status === STATUS_PAID_BUT_NOT_ENROLLED) {
     const contactHref = `mailto:${SETTINGS.support_email}`
     return S.Just([
@@ -146,7 +147,7 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
   ) {
     messages.push({
       message:
-        "You are auditing. To get credit, you need to pay for the course.",
+        `You are auditing. To get credit, you need to pay for the course. (Payment due on ${dueDate})`,
       action: courseAction(firstRun, COURSE_ACTION_PAY)
     })
     return S.Just(messages)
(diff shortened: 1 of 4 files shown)


static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -95,6 +95,7 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
   const passedExam = hasPassingExamGrade(course)
   const failedExam = hasFailingExamGrade(course)
   const paymentDueDate = moment(firstRun.course_upgrade_deadline)
+  const dueDate = paymentDueDate.format(COURSE_CARD_FORMAT)
```
amir-qayyum-khan: @annagav do we have always `firstRun.course_upgrade_deadline` set?

pdpinch: We should, but it would be good to fail gracefully if it's missing.

pdpinch:
Are we displaying the date/time in the user's timezone? Can we?

amir-qayyum-khan:
@pdpinch i dont think so, they are according to user's timezone.

pdpinch:
@annagav we need to include the time and the timezone. Since we have users across the globe, they can't tell if the deadline is in their timezone or EST/EDT. 

Ideally, I'd like to display the date & time in the user's local time, based on their profile. But if that's too hard/unreliable, we can display the date and time in UTC.

noisecapella:
@pdpinch All datetime information coming from the REST API is in ISO-8601 format which is explicit about it being UTC. When this is formatted via moment.js (like here) it's shown to the user in the browser's local time zone

roberthouse54:
So we should show the time in the user's local time, and display the timezone abbreviation, which according to Anna requires a new library that we currently don't have.

pdpinch:
Do we have an option to implement this in Python, or are we limited to doing it in JavaScript?

noisecapella:
IMO having the REST API output UTC timestamps is the way to go. moment.js already knows what the local timezone is since it's using it to render the time. @annagav which library would we need?

annagav:
@noisecapella I used moment-timezone to get the abbreviation of the user's timezone.

noisecapella:
Yes that sounds right to me

amir-qayyum-khan:
please fix tests and style issues

amir-qayyum-khan:
Looks good to me

@pdpinch @noisecapella

Issue #3602, opened 2017-10-04
ZenDesk Widget throwing 1000s of TypeErrors: Failed to execute 'getComputedStyle' on 'Window'


pdpinch:
I'm not sure if this is a bug in ZenDesk's JavaScript widget, or if it's an issue in how we're trying to use it. 

TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/361180415/

```
TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
  at callback (./static/js/entry/zendesk_widget.js:105:41)
  at apply (./static/js/entry/zendesk_widget.js:243:9)
```

pdpinch:
If this is a ZenDesk error, then we should find a way to suppress the error so it doesn't go to Sentry. We're severely clogging the sentry logs with this error.

noisecapella:
The stack trace is from our code. I think a CSS selector no longer matches because they adjusted something on their end, and this is causing a null error

pdpinch:
@amir-qayyum-khan this is happening again. I'm hoping maybe it's just some cached javascript code, but I don't know. 

See, for example, https://sentry.io/mit-office-of-digital-learning/micromasters/issues/361180415/events/8370678176/

Issue #3435, opened 2017-07-27
Zendesk errors


noisecapella:
I'm seeing this in Chrome on production:

![screenshot from 2017-07-27 16-39-11](https://user-images.githubusercontent.com/863262/28691262-43d3d52c-72ea-11e7-8579-1391540517f7.png)

PR #3624, opened 2017-10-11
Fixed failed to execute getComputedStyle on Window error on zendesk script

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3602

#### What's this PR do?
it updates css select on `zendesk_widget.js`, it appears that zendesk changed style on their end, to match with it i need to change the selector.

login to your MM and see console (Probably on dashboard)

@pdpinch

### DIFF
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -99,7 +99,7 @@ const zendeskCallbacks = {
 
   launcherLoaded: () => {
     const iframe = document.querySelector("iframe.zEWidget-launcher")
-    const btn = iframe.contentDocument.querySelector(".u-userBackgroundColor")
+    const btn = iframe.contentDocument.querySelector(".u-userLauncherColor")
 
     const regularBackgroundColor = "rgba(0, 0, 0, .14)"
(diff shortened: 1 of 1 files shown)


pdpinch:
do you think this will fix https://github.com/mitodl/micromasters/issues/3435 as well?

amir-qayyum-khan:
No @pdpinch this will not fix #3435 , that error messages are temporary and we are logging them

amir-qayyum-khan:
@pdpinch these are temporary errors. 
Should i remove `console.error(`couldn't find Zendesk iframe: ${name}`) // eslint-disable-line no-console` from [here](https://github.com/mitodl/micromasters/blob/master/static/js/entry/zendesk_widget.js#L215). because app retries by itself

noisecapella:
I think this fix will work until zendesk changes something again and our CSS selector stops working. Can we look through our zendesk code to see where else we may run into this problem?

amir-qayyum-khan:
there are many `document.querySelector` selectors in [code](https://github.com/mitodl/micromasters/blob/master/static/js/entry/zendesk_widget.js) which can create same issue in future.

PR #3636, opened 2017-10-13
Check if document needs updating before reindexing

### LINKED ISSUE #3634: Reduce number of reindexes due to program enrollment changes
There may be some places where we save the profile, program enrollment, or related information when we don't need to.

noisecapella:
Fixes #3634 

#### What's this PR do?
Reduces the number of reindexes and tasks triggered after reindexing

Make sure you have open-discussions set up and you have a channel created. Open your open-discussions logs in a terminal. In micromasters go to your profile and save it without making any changes. You should see only the update of the `api/v0/users/.../` endpoint.

Make a change to your profile and save it. You should see the update to the `api/v0/users/.../` endpoint but also `DELETE` or `POST` to the subscribers and contributors APIs.

### DIFF
--- a/search/tasks.py
+++ b/search/tasks.py
@@ -10,11 +10,11 @@ from dashboard.models import ProgramEnrollment
 from discussions.api import sync_user_to_channels as _sync_user_to_channels
 from mail.api import send_automatic_emails as _send_automatic_emails
 from micromasters.celery import app
+from search.api import document_needs_updating as _document_needs_updating
 from search.indexing_api import (
     get_default_alias,
     index_program_enrolled_users as _index_program_enrolled_users,
     remove_program_enrolled_user as _remove_program_enrolled_user,
-    index_users as _index_users,
     index_percolate_queries as _index_percolate_queries,
     delete_percolate_query as _delete_percolate_query,
     refresh_index as _refresh_index,
@@ -25,12 +25,12 @@ from search.models import PercolateQuery
 log = logging.getLogger(__name__)
 
 
-def post_indexing_handler(program_enrollment_ids):
+def post_indexing_handler(program_enrollments):
     """
     Do the work which happens after a profile is reindexed
 
     Args:
-        program_enrollment_ids (list of int): A list of ProgramEnrollment ids
+        program_enrollments (list of ProgramEnrollment): A list of ProgramEnrollments
     """
     feature_sync_user = settings.FEATURES.get('OPEN_DISCUSSIONS_USER_SYNC', False)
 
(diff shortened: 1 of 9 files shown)


search/tasks.py
```diff
@@ -68,22 +68,32 @@ def index_program_enrolled_users(program_enrollment_ids):
...
     """
-    _index_users(user_ids)
+    enrollments = list(ProgramEnrollment.objects.filter(user__in=user_ids))
 
-    # Send email for profiles that newly fit the search query for an automatic email
-    enrollment_ids = list(ProgramEnrollment.objects.filter(user__in=user_ids).values_list("id", flat=True))
-    post_indexing_handler(enrollment_ids)
+    if check_if_changed:
+        enrollments = [
+            enrollment for enrollment in enrollments if _document_needs_updating(enrollment)
```
rhysyngsun: This seems like it could result in additional overhead in cases where the enrollment/profile *was* updated because we end up querying for and serializing the data twice. Once in `_document_needs_updating` and the other in `_get_program_enrolled_users_documents` which is called by `index_program_enrolled_users`. I suspect you place the check here so we never skip documents when rebuilding the index, but I'm a little concerned we're trading off overhead on one system for overhead on another.

Does it make sense to carry `check_if_changed` all the way to `_get_program_enrolled_users_documents` and simply not yield there instead?

noisecapella: I'm not sure I follow, where is the overhead tradeoff happening?

rhysyngsun: We're hitting reindexing less, but for those things we do index, we're computing the document twice now (including all the DB queries to do so). At worst, we've doubled our overhead on the MM side.

noisecapella: The information on whether or not to reindex needs to exist both in this function (to decide on the post indexing tasks) and in the indexing function. In order to avoid serializing twice we would need to do it here and pass down the dict into the other functions. Also, currently `document_needs_updating` will do a Elasticsearch request for each document, which would change indexing from a one request per chunk operation to one per document.

I think ideally we would have a stream of ES documents and `ProgramEnrollment`s sorted by id, do the comparison and reindex the modified ones but it would require some refactoring, and `index_users` currently only handles one user at a time so I'm not sure it's worth it.

rhysyngsun: Yeah, I'm starting to see your point there. I agree with the current design this is probably the best approach even if not optimal.

rhysyngsun:
In my testing, it appears that just opening up the profile, not changing anything, and then clicking "Save" is still triggering an API update to OD.

PR #3637, opened 2017-10-13
Split CSS into separate file for production

### LINKED ISSUE #3485: Output CSS in independent static file
Currently our CSS is bundled with the JS, but we can add a build step to split this off into a separate file for production builds. This will increase performance a little bit since the CSS can be downloaded and processed independently of the JS

More info: https://github.com/webpack-contrib/extract-text-webpack-plugin

noisecapella:
Fixes #3485 

#### What's this PR do?
This sets up the extract-text-plugin so that production builds will have separate CSS files which will contain all style information. 

View the PR build, log in, and view the dashboard. Nothing should look odd

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "eslint-plugin-mocha": "4.11.0",
     "eslint-plugin-react": "7.1.0",
     "express": "^4.15.3",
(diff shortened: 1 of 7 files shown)


aliceriot:
works on the PR build and works locally! reading over the code now.

PR #3640, opened 2017-10-16
Use HEROKU_APP_NAME as ELASTICSEARCH_INDEX value for PR builds

### LINKED ISSUE #2342: Configure PR builds to create/use separate indices and clean up old indices
_Part of #2270_

All -ci builds currently use the same Elasticsearch index. We need reconfigure them to use separate indices (eg: PR 1234 creates and uses index `micromasters-ci-1234` instead of using `micromasters-ci`). We'll also need to periodically remove indices that aren't being used anymore (ie: PRs that have been closed)

Acceptance criteria:
1. PR builds create and use their own Elasticsearch index
1. Some routine has been created that runs on a schedule and removes indices for PR builds that are no longer open

We can split this into 2 issues if needed

noisecapella:
Fixes #2342

#### What's this PR do?
If the deployment is a PR build, use the PR build app name in place of the given `ELASTICSEARCH_INDEX` value which is just copied from `micromasters-ci`. This should prevent PR builds from overwriting the index of CI

In a Django shell verify that the `ELASTICSEARCH_INDEX` setting matches the name of the PR build.

### DIFF
--- a/app.json
+++ b/app.json
@@ -105,9 +105,11 @@
       "required": true
     },
     "HEROKU_APP_NAME": {
+      "description": "Value provided by Heroku containing the app name (eg micromasters-ci)",
       "required": true
     },
     "HEROKU_PARENT_APP_NAME": {
+      "description": "Value provided by Heroku containing the parent app name (eg micromasters-ci for a PR build)",
       "required": true
     },
     "MAILGUN_KEY": {
--- a/app.json
+++ b/app.json
@@ -104,6 +104,12 @@
       "description": "API key for accessing Google services",
       "required": true
     },
+    "HEROKU_APP_NAME": {
+      "required": true
+    },
+    "HEROKU_PARENT_APP_NAME": {
+      "required": true
+    },
     "MAILGUN_KEY": {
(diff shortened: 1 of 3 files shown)


app.json
```diff
@@ -104,6 +104,12 @@
       "description": "API key for accessing Google services",
       "required": true
     },
+    "HEROKU_APP_NAME": {
+      "required": true
+    },
+    "HEROKU_PARENT_APP_NAME": {
+      "required": true
+    },
```
pdpinch: It's not obvious to me what these settings should contain. Can we add a description? 

Does this mean we'll have to add these settings to all our deployments?

noisecapella: They are special heroku environment variables which are populated by heroku with the name of the app and the name of the app's parent (if it's a PR build). They are marked as required but since they're filled in automatically there shouldn't be any extra work on our part. I'll add a description

pdpinch:
I can confirm that `ELASTICSEARCH_INDEX` is set to `micromasters-ci-pr-3640` on the PR build.

PR #3657, opened 2017-10-20
Log a diff of the ES document and serialized enrollment

### LINKED ISSUE #3654: Log reason for profile reindexes
We should log what the diff was for profile reindexes at debug level. This would help in diagnosing why a profile reindex was triggered

noisecapella:
Fixes #3654 

#### What's this PR do?
Log a diff of the difference between the Elasticsearch document and the serialized enrollment

Make a change to a `Profile` and save it using `with mute_signals(post_save)` to prevent Elasticsearch from updating. In the Django shell run `profile.save()` and verify that you see a logging message with a diff of the field that was changed.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -1,11 +1,15 @@
 """
 Functions for executing ES searches
 """
+import json
+import logging
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Q as Query
 from elasticsearch.exceptions import NotFoundError
 from elasticsearch_dsl import Search, Q
+from jsonpatch import make_patch
 
 from courses.models import Program
 from dashboard.models import ProgramEnrollment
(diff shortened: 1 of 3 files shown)


amir-qayyum-khan:
@noisecapella some tests are failing in `search` app

search/api.py
```diff
@@ -300,4 +307,14 @@ def document_needs_updating(enrollment):
         return True
     serialized_enrollment = serialize_program_enrolled_user(enrollment)
     del serialized_enrollment['_id']
-    return serialized_enrollment != document['_source']
+    source = document['_source']
+
+    if serialized_enrollment != source:
```
rhysyngsun: Should we have this behind a flag so we can only turn it on when we need to debug? I'm not sure there's utility in logging every diff we see.

rhysyngsun:
The functionality works as intended.

PR #3661, opened 2017-10-24
Reduced number of side effects from reindexing

### LINKED ISSUE #3655: Automatic emails may get blocked by open-discussion errors
We should fix the post indexing code to log errors but otherwise complete all post indexing work. Currently an exception will stop all other work for a user

### LINKED ISSUE #3660: Eliminate unnecessary API calls to open-discussions
We should eliminate extraneous calls to open-discussions when the membership hasn't actually changed.

rhysyngsun:
Fixes #3660 

#### What's this PR do?
Adds a new `PercolateQueryMembership` record that tracks when membership in a `PercolateQuery` actually changes. Updates to these records are then shipped to OD via a cron job that runs every minute.

- Create a channel that matches a user name that doesn't exist.

Do the `PercolateQuery` hokey-pokey:
- Login to a non-staff user account and update the profile name to match your query. 
- Verify you see *only* an API request to OD to add that user
- Update some other property on the profile (to trigger reindex)
- Verify you see *no* API requests to OD for you user
- Change your user's name to so that their name doesn't match any more
- Verify you see *only* an API request to OD to remove that user
- Update some other property on the profile (to trigger reindex)
- Verify you see *no* API requests to OD for you user


discussions/api.py
```diff
@@ -267,46 +265,58 @@ def remove_from_channel(channel_name, discussion_username):
...
-    nonmembership_channels = Channel.objects.filter(
-        id__in=all_channel_ids.difference(membership_channel_ids)
-    )
-    for channel in nonmembership_channels:
-        remove_from_channel(channel.name, discussion_user.username)
+            channel = membership.query.channels.first()
+
+            # if we can't look up the program, skip this one
+            if channel is None or channel.id not in program_ids_by_channel_id:
+                continue
```
noisecapella: Can you add a test covering this line?

discussions/api.py
```diff
@@ -267,46 +265,58 @@ def remove_from_channel(channel_name, discussion_username):
...
+                    discussion_user = create_or_update_discussion_user(membership.user_id)
+
+                    if membership.is_member:
+                        add_to_channel(channel.name, discussion_user.username)
+                    else:
+                        remove_from_channel(channel.name, discussion_user.username)
+
+                    membership.needs_update = False
+                    membership.save()
+                except DiscussionUserSyncException:
```
noisecapella: Can you add a test for this case? Just to make sure an exception doesn't propagate

discussions/tasks.py
```diff
@@ -47,28 +47,24 @@ def sync_discussion_users():
...
 @app.task()
-def add_users_to_channel(channel_name, user_ids):
+def add_moderators_to_channel(channel_name):
     """
-    Add users to a open-discussions channel as contributors and subscribers
-
-    Args:
-        channel_name (str): The name of the channel
-        user_ids (list of int): profile ids to sync
+    Add moderators to a open-discussions chane
```
noisecapella: Can you change `chane` to `channel`?

discussions/api.py
```diff
@@ -267,46 +265,58 @@ def remove_from_channel(channel_name, discussion_username):
     remove_contributor_from_channel(channel_name, discussion_username)
 
 
-def sync_user_to_channels(user_id):
+def sync_channel_memberships():
```
noisecapella: Since this runs every minute, are we doing anything to prevent two instances of this task from running concurrently?

search/api.py
```diff
@@ -240,18 +245,31 @@ def search_percolate_queries(program_enrollment_id, source_type):
...
     # We don't need this to search for percolator queries and
     # it causes a dynamic mapping failure so we need to remove it
     del doc['_id']
     result = conn.percolate(get_default_alias(), USER_DOC_TYPE, body={"doc": doc})
     failures = result.get('_shards', {}).get('failures', [])
     if len(failures) > 0:
         raise PercolateException("Failed to percolate: {}".format(failures))
-    result_ids = [row['_id'] for row in result['matches']]
-    return PercolateQuery.objects.filter(id__in=result_ids, source_type=source_type)
+    return [int(row['_id']) for row in result['matches']]
```
noisecapella: Was the `int(...)` here necessary?

search/api.py
```diff
@@ -301,3 +319,85 @@ def document_needs_updating(enrollment):
...
+    """
+    Atomically determine and update memberships
+
+    Args:
+        percolate_query_ids (set of int): a set of PercolateQuery.id
+        membership_ids (list of int): A list of ids for PercolateQueryMemberships to update
+            user (User): A User to check for membership changes
+        force_save (bool): True if membership saves should be force even if no change
+    """
+    if not percolate_query_ids:
```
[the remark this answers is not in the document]
rhysyngsun: Actually, in this case both `percolate_query_ids` and `membership_ids` are empty. Since the method that calls this determines `membership_ids` based off `percolate_query_ids`, if the latter is empty the former always will be too.

discussions/api.py
rhysyngsun: Not right now, we don't have a good locking utility, although code exists in dashboard/tasks.py, that would add time to generalize. An alternative would be to space out the task runs.

discussions/api.py
noisecapella: Can you add a comment describing the race condition?

search/api.py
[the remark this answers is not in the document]
noisecapella: I'm not sure we should trust that that will always be true here. Can you add `and if not membership_ids` here to make sure both are empty?

search/tasks.py
```diff
@@ -41,8 +44,8 @@ def post_indexing_handler(program_enrollments):
     for program_enrollment in program_enrollments:
         _send_automatic_emails(program_enrollment)
 
-        if feature_sync_user:
-            _sync_user_to_channels(program_enrollment.user.id)
+        # only update for discussion queries for now
```
noisecapella: Are we still checking the feature flag somewhere?

rhysyngsun: Yup, it's done at the start of the `discussions.tasks.sync_channel_memberships` task now.

noisecapella:
Can you look into this? https://github.com/mitodl/micromasters/pull/3661#discussion_r147729468

search/api.py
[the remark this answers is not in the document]
rhysyngsun: I think we should probably exit if either are empty, right?

rhysyngsun:
To test the locking behavior, I'd suggest adding `time.sleep(10)` or something like that in the api method to simulate a long-running task. Then call this:

```python
from discussions import tasks

for _ in range(5):
    tasks.sync_channel_memberships.delay()
```

discussions/tasks.py
```diff
@@ -67,4 +97,21 @@ def sync_channel_memberships():
         log.debug('OPEN_DISCUSSIONS_USER_SYNC is set to False (so disabled) in the settings')
         return
 
-    api.sync_channel_memberships()
+    # establish when we'll stop syncing this "batch"
+    end_time = time.time() + SYNC_MEMBERSHIPS_LOCK_TTL_SECONDS
```
noisecapella: Can you change this to `now_in_utc() + timedelta(seconds=SYNC_MEMBERSHIPS_LOCK_TTL_SECONDS)`? And similarly below in `should_sync`?

search/api.py
```diff
@@ -301,3 +319,84 @@ def document_needs_updating(enrollment):
...
+    """
+    Atomically determine and update memberships
+
+    Args:
+        percolate_query_ids (set of int): a set of PercolateQuery.id
+        membership_ids (list of int): A list of ids for PercolateQueryMemberships to update
+        user (User): A User to check for membership changes
+        force_save (bool): True if membership saves should be force even if no change
+    """
+    if not percolate_query_ids or not membership_ids:
```
noisecapella: I think this should be an `and` since this is optimizing for the case that both are empty. If `percolate_query_ids` was empty but not membership ids, it would mean we should clear membership status for those members

discussions/tasks.py
```diff
@@ -98,18 +99,19 @@ def sync_channel_memberships():
...
     lock = _get_sync_memberships_lock()
 
-    def should_sync():
-        """Returns True if we should continue syncing"""
-        return end_time > time.time()  # should only sync if the lock hasn't expired
+    def _get_memberships():
+        """Generator for membership ids to sync"""
+        for membership_id in api.get_membership_ids_needing_sync():
+            # this will stop yielding once our lock has expired
+            if end_time > now_in_utc():
```
noisecapella: Should we have an `else: break` here?

discussions/api.py
```diff
@@ -265,28 +265,32 @@ def remove_from_channel(channel_name, discussion_username):
     remove_contributor_from_channel(channel_name, discussion_username)
 
 
-def sync_channel_memberships(should_sync):
+def get_membership_ids_needing_sync():
+    """
+    Returns a list of memberships that need to be synced
+
+    Returns:
+        list of int: list of database ids for memberships that need to be synced
```
noisecapella: Can you make the return type a QuerySet instead?

discussions/tasks.py
```diff
@@ -98,18 +99,19 @@ def sync_channel_memberships():
...
 
     if lock.acquire():
         try:
-            api.sync_channel_memberships(should_sync)
-        except LockError as ex:
-            log.debug("Unable to get lock for syncing channel memberships: %s", ex)
+            api.sync_channel_memberships(_get_memberships())
         finally:
             try:
                 lock.release()
```
noisecapella: What case would we not own the lock anymore? When it expires?

discussions/tasks.py
rhysyngsun: Yes

discussions/tasks.py
rhysyngsun: Correct, if we happen to run past the expiration of the lock and another process recreates the lock it will raise this error.

rhysyngsun:
@noisecapella all set

search/api.py
```diff
@@ -336,7 +336,9 @@ def update_percolate_memberships(user, source_type):
         user
     )
 
-    _update_memberships([query.id for query in percolate_queries], membership_ids, user)
+    # if there are no percolate queries or memberships then there's nothing to do
+    if percolate_queries and membership_ids:
```
noisecapella: This should be an `or` I think. If only one or the other is false it seems like we want to do the update, right?

rhysyngsun: No, if `membership_ids` is empty, there are no memberships to process, if `percolate_queries` is empty, there's also nothing to do, although arguably, if `percolate_queries` is empty, `membership_ids` will be too as we would have not created any on line 334. So really, in theory, just checking one should be sufficient.

Either that or I could return earlier if `percolate_queries` is empty, but I think the difference between the two approaches is more or less moot.

noisecapella: If it doesn't matter which one we check, can we just check `membership_ids` to make this logic simpler?

discussions/tasks.py
```diff
@@ -47,28 +78,44 @@ def sync_discussion_users():
...
+    lock = _get_sync_memberships_lock()
+
+    def _get_memberships():
+        """Generator for membership ids to sync"""
+        for membership_id in api.get_membership_ids_needing_sync():
+            # this will stop yielding once our lock has expired
+            if end_time > now_in_utc():
+                yield membership_id
+            else:
+                break
```
noisecapella: Can we add a test for this case? It would help to verify that the task will not run too long

PR #3668, opened 2017-10-25
conditionally hide course progress

### LINKED ISSUE #3313: Suppress "in progress" grades for some courses
Some courses, like the SCM Comprehensive Final Exam (CFx) shouldn't display in-progress grades before the grades are finalized. The course team doesn't want to mislead learners while the results of the course are still being interpreted. 

This needs to be addressed before the next CFx is offered, in February 2018.

aliceriot:

closes #3313 

#### What's this PR do?

This adds a boolean column to the `Course` table called `should_display_progress`, which governs whether course runs for that course should report progress before they have a finalized grade or not.

The column has a default value of `True`, so you should see no changes initially after running the migration. Then, if you have an in-progress course which is reporting a 'current_grade' on one of it's course runs, you should stop seeing that information in the UI if you switch `should_display_progress` to be `False`.


pdpinch:
Will this still allow the final grade to be displayed?

aliceriot:
yep

dashboard/api_test.py
```diff
@@ -232,6 +232,34 @@ def test_format_run_with_currently_enrolled(self):
             self.expected_ret_data
         )
 
+    def test_format_run_currently_enrolled_dont_display_progress(self):
```
noisecapella: Can you also add a test for `CAN_UPGRADE`, or make this one parameterized?

dashboard/api_test.py
```diff
@@ -232,6 +232,34 @@ def test_format_run_with_currently_enrolled(self):
...
+        self.expected_ret_data.update({
+            'status': api.CourseStatus.CURRENTLY_ENROLLED
+        })
+        del self.expected_ret_data['final_grade']
+        self.crun.course.should_display_progress = False
+        self.assertEqual(
+            api.format_courserun_for_dashboard(self.crun, api.CourseStatus.CURRENTLY_ENROLLED, self.mmtrack),
+            self.expected_ret_data
+        )
+        self.crun.course.should_display_progress = True
```
noisecapella: Instead of resetting the value manually can you add `self.course.refresh_from_db()` to `setUp` before `create_run` is called?

dashboard/api_test.py
```diff
@@ -232,6 +232,34 @@ def test_format_run_with_currently_enrolled(self):
...
+        """
+        test that we still return a final grade with should_display_progress
+        set to False
+        """
+        self.crun.course.should_display_progress = False
+        self.assertEqual(
+            api.format_courserun_for_dashboard(self.crun, api.CourseStatus.PASSED, self.mmtrack),
+            self.expected_ret_data
+        )
+        self.crun.course.should_display_progress = True
```
noisecapella: Instead of resetting the value manually can you add `self.course.refresh_from_db()` to `setUp` before `create_run` is called?

dashboard/api_test.py
aliceriot: sure

aliceriot:

Issue #3665, opened 2017-10-25
Radio buttons on new user profile pages are misaligned


rhysyngsun:
![selection_093](https://user-images.githubusercontent.com/28598/32008758-0a581c90-b97b-11e7-997e-c15d74ed7905.png)

noisecapella:
I tried to find the commit introducing this via git bisect. It's at least a few months old and before that I run into migration issues I didn't want to deal with locally. I feel like it might be caused by the Chrome update rather than a library change, but I don't know for sure

PR #3669, opened 2017-10-25
Fixes a layout fix with radio buttons on the profile pages in Chrome

roberthouse54:
#3665 

#### What's this PR do?
This fixes a layout bug in Chrome where the radio buttons on the profile education and employment tabs were overlapping with the labels. 

I think this bug may have been introduced with the most recent version of Chrome, so it should be tested there. But if you have a not-up-to-date version of Chrome, test it there first, then update, then test again.

BUG:
image

FIX:
image

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -118,7 +118,7 @@ class EducationForm extends ProfileFormFields {
     const valueSelected = level.value in educationLevelAnswers ? "false" : null
     return (
       <RadioButtonGroup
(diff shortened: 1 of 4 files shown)


gsidebo:
Safari is looking wonky

ss 2017-10-30 at 15 46 59

roberthouse54:
OK, I think I fixed it. I had to use !important. But I was unsure how else to fix the "display: table" inline style that was causing the problem. I think this is some part of the material component css. Why it is putting in inline styles is beyond me. If you can see why Gavin, let me know.

Issue #3659, opened 2017-10-24
Space out requests in batch_update_user_data_subtasks


noisecapella:
We should slow down the number of requests made by this task to reduce the workload on edX

PR #3670, opened 2017-10-25
Refactor batch_update_user_data, fix lock behavior

noisecapella:
Part of #3659 

#### What's this PR do?
Refactors code for the `batch_update_user_data` task. Additionally the locking code should be fixed. Neither change should affect behavior since this task is only run once every 6 hours.

Delete the `UserCacheRefreshTime` object for your user. Run the `batch_update_user_data` task. Verify that the logs say that the lock is released and that there is a new `UserCacheRefreshTime` object for your user.

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -7,11 +7,15 @@ from urllib.parse import urljoin
 import pytz
 
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.core.exceptions import ImproperlyConfigured
 from django.db import transaction
 from django.urls import reverse
+from edx_api.client import EdxApi
+from social_django.models import UserSocialAuth
 
 from backends.exceptions import InvalidCredentialStored
+from backends import utils
 from courses.models import Program
 from dashboard.api_edx_cache import CachedEdxDataApi, CachedEdxUserData
 from dashboard.utils import MMTrack
@@ -20,6 +24,8 @@ from grades import api
 from grades.models import FinalGrade
 from grades.serializers import ProctoredExamGradeSerializer
 from exams.models import ExamAuthorization, ExamRun
+from micromasters.utils import now_in_utc
+from profiles.api import get_social_auth
 
 # maximum number of exam attempts per payment
 ATTEMPTS_PER_PAID_RUN = 2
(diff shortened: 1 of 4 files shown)


dashboard/api.py
```diff
@@ -524,3 +530,74 @@ def get_overall_final_grade_for_course(mmtrack, course):
...
+    """
+    # pylint: disable=bare-except
+    try:
+        user = User.objects.get(pk=user_id)
+    except:
+        log.exception('edX data refresh task: unable to get user "%s"', user_id)
+        return
+
+    try:
+        UserSocialAuth.objects.get(user=user)
```
amir-qayyum-khan: No need of this, we are excluding users without social auth

dashboard/api_test.py
```diff
@@ -1819,3 +1821,206 @@ def test_no_passing_exam(self):
...
+    user = SocialProfileFactory.create().user
+    user.profile.fake_user = False
+    user.profile.save()
+    now = now_in_utc()
+    UserCacheRefreshTimeFactory.create(user=user, enrollment=now, certificate=now, current_grade=now)
+    return user
+
+
+@pytest.fixture
+def calc_users(db):
```
amir-qayyum-khan: can we make this function name more meaningful

dashboard/api_test.py
```diff
@@ -1819,3 +1821,206 @@ def test_no_passing_exam(self):
...
+    needs_update = [_make_fake_real_user() for _ in range(5)]
+    for user in needs_update:
+        user.usercacherefreshtime.delete()
+
+    return needs_update, up_to_date
+
+
+def test_calculate_up_to_date(calc_users):
+    """
+    calculate_users_to_refresh should return a list of user ids for users whose cache has expired or who are not cached
```
amir-qayyum-khan: split this string in two lines

noisecapella:
@rhysyngsun Can you take a look at this too when you get some free time?

dashboard/tasks.py
```diff
@@ -2,36 +2,49 @@
...
+    """
+    # lock will expire if 2 hrs are pass and task is not complete.
+    return cache_redis.add(LOCK_ID, 'true', LOCK_EXPIRE)
+
+
+def _release_lock():
+    """
+    remove lock by deleting key from storage.
+    """
+    return cache_redis.delete(LOCK_ID)
```
rhysyngsun: To have a truly safe distributed lock you need to set the key to a unique value and then only release it if they value matches the value you set it as (so you don't release a lock created by another process). Since we're only holding the lock for 2 hours but only run the task every 6 we should be ok, but it might be worth adding a note of this limitation in case we change how often we run it down the road.

noisecapella: Ok I'll add a comment

noisecapella: Actually I'm thinking we should just get rid of the subtasks altogether and check the lock similar to how it's done for `PercolateQueryMembership`.

noisecapella: Changed my mind again, I'll just add the comment to this PR and then work on the refactor in the next PR, which will also include throttling changes

Issue #3619, opened 2017-10-10
Cybersource Redirect fails on Safari


pdpinch:
See ticket https://odl.zendesk.com/agent/tickets/11443

When a purchase is successfully completed, Cybersource redirects the user back to a URL such as `dashboard/?status=receipt&course_key=course-v1%3AMITx%2B14.740x%2B3T2017`

Safari 10.1.2 claims that the server drops the connection when trying to follow the redirect. The user who submitted the ZenDesk ticket was using Safari 11 and was "stuck indefinitely with a
black background and a white spinning wheel in the middle"

amir-qayyum-khan:
@pdpinch is there Cybersource test account for testing

pdpinch:
There's a list of cards that will work with the test center account (connected to CI and RC) at https://www.cybersource.com/developers/other_resources/quick_references/test_cc_numbers/

Note that for testing purposes, some dollar values will return an error. In particular, if you test with a $1000 purchase you will get an error. You will need to set a personal price or use a coupon in order to get a different total.

amir-qayyum-khan:
@pdpinch @giocalitri getting error, this is very easy to reproduce
Got order confirmation email.

> ERROR 21 [nplusone] notifiers.py:40 - [5b13615ad0eb] - Potential n+1 query detected on `Course.courserun_set`

pdpinch:
Ok, but what does an n+1 error have to do with Safari?

amir-qayyum-khan:
@pdpinch issue can be produce on chrome as well

noisecapella:
The `Potential n+1 query` messages are logged as errors but should not be treated as errors. They are just sources of inefficiency and probably unrelated to this issue

amir-qayyum-khan:
https://micromasters-rc.odl.mit.edu
Returning from cybersource this is what happen on safari

<img width="1243" alt="screen shot 2017-10-24 at 7 09 30 pm" src="https://user-images.githubusercontent.com/10431250/31947996-1ab9575c-b8ef-11e7-9921-02fb46448cd8.png">

PR #3671, opened 2017-10-26
Fixed Cybersource Redirect fails on Safari

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3619

#### What's this PR do?
Configure nginx to add  setting `keepalive_disable safari;` This will disable keep alive connection, helps in redirect. Also add SSL shared memory cache of 10 MB. 

> 1down vote | One of the reasons why it might happen is the way Safari handles some SSL certificated (mainly, LetsEncrypt). The solution is to put the line:

- https://stackoverflow.com/questions/33895463/safari-ajax-request-failed-to-load-resource-the-network-connection-was-lost
- http://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_disable

(Required)

(Optional)

@pdpinch 

screen shot 2017-10-26 at 5 36 25 pm

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - "9100:9200"
 
   nginx:
-    image: nginx
+    image: nginx:1.11
     ports:
       - "8079:8079"
     links:


amir-qayyum-khan:
@blarghmatey can we deploy this to RC and test?

blarghmatey:
@amir-qayyum-khan there's no need to list the keepalive_disable parameter multiple times in the configuration. Placing it in the `http` block will cause it to be inherited by the `server` blocks contained within it.

config/nginx.conf.erb
```diff
@@ -41,23 +41,27 @@ http {
...
         listen <%= ENV["PORT"] %>;
         server_name micromasters.herokuapp.com;
+        keepalive_disable safari;
         return 301 https://micromasters.mit.edu$request_uri;
     }
 
     server {
         listen <%= ENV["PORT"] %>;
         server_name mm.mit.edu;
+        keepalive_disable safari;
```
pdpinch: Do we need this for these names that are just going to redirect anyway?

pdpinch: Oh, I see @blarghmatey said the same thing in https://github.com/mitodl/micromasters/pull/3671#issuecomment-339981509

amir-qayyum-khan:
Guys this is in progress, the solution is not working for me.

pdpinch:
If this approach isn't working, can we close this PR?

amir-qayyum-khan:
this solution is not right

PR #3675, opened 2017-10-27
get final grade from current grades

### LINKED ISSUE #3648: use current_grade.percent  instead of certificate.grade for FinalGrade
We have been informed by edX that the grades that are returned by the certificate API may not be the learner's final grade, because certificates may be generated as soon as the learner passes a course and won't be updated if they improve their grade. 

Instead of getting final grades from the certificate, we should get them from current_grade (i.e. the `course_grade` API). 

A quick fix would be to change https://github.com/mitodl/micromasters/blob/8e85c3a74f833716c64b94252e82a1504234e55f/grades/api.py#L43 and https://github.com/mitodl/micromasters/blob/8e85c3a74f833716c64b94252e82a1504234e55f/grades/api.py#L74 to use `current_grade.percent`

After this is fixed, we need to re-freeze grades for two SCM courses. 

As tech debt, we may also want to remove the certificate.grade entirely, since edX is not treating it as a meaningful grade.

annagav:
Fix #3648

#### What's this PR do?
Use current grades and calculating final grade.

Tests should pass

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -36,21 +36,17 @@ def _compute_grade_for_fa(user_edx_run_data):
         UserFinalGrade: a namedtuple of (float, bool,) representing the final grade
             of the user in the course run and whether she passed it
     """
-    run_passed = None
-    grade = None
     if user_edx_run_data.certificate is not None:
         run_passed = user_edx_run_data.certificate.status == 'downloadable'
-        grade = user_edx_run_data.certificate.grade
         # the following line should be updated when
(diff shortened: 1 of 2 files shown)


pdpinch:
Looks like this failed tests because of the psycopg2 issue that was fixed in #3666? Does it need a rebase?

See https://travis-ci.org/mitodl/micromasters/jobs/293693930#L1518-L1521

Issue #3630, opened 2017-10-12
DoesNotExist: UserSocialAuth matching query does not exist.


pdpinch:
We get multiple `DoesNotExist: UserSocialAuth matching query does not exist.` errors every six hours. Most of these errors are related to superusers on the CI and RC servers, but there is also at least one user in production. 

Since these users don't have an edX SocialAuth, we shouldn't even attempt to update their edX cache. Although I am curious about that one user in production, I don't see a lot of value is raising an exception every six hours for that user. Presumably, if they have a successful log in to micromasters.mit.edu the issue should clear. 

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/193071074/

```
DoesNotExist: UserSocialAuth matching query does not exist.
  File "dashboard/tasks.py", line 101, in batch_update_user_data_subtasks
    UserSocialAuth.objects.get(user=user)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 385, in get
    self.model._meta.object_name

user "nawalyathccdd90e46a0f459a" does not have python social auth object
```

pdpinch:
@noisecapella which PR addresses this?

amir-qayyum-khan:
@pdpinch #3670

amir-qayyum-khan:
https://github.com/mitodl/micromasters/pull/3670 merged

PR #3677, opened 2017-10-30
Filtered users without social auth before subtask processing

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3630

#### What's this PR do?
It filter out users without `UserSocialAuth` object before calling subtask `batch_update_user_data_subtasks`. Which actually update users edX data. This will help to avoid unnecessary attempts to fetch data

- Run tests
- Reduce the time of this [task](https://github.com/mitodl/micromasters/blob/master/micromasters/settings.py#L468) and check for exception. You worn see exception related missing social auth.
 
@pdpinch

### DIFF
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -68,7 +68,9 @@ def batch_update_user_data():
             Q(id__in=users_expired_cache)
         ).filter(is_active=True, profile__fake_user=False).values_list('id', flat=True)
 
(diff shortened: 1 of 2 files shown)


noisecapella:
Sorry that I did not assign myself to that issue but my PR also addresses this: https://github.com/mitodl/micromasters/pull/3670

PR #3683, opened 2017-10-31
CourseCertificates: create if final grade is complete

annagav:
None

#### What's this PR do?
Add a filter to `FinalGrade`s when creating `MicromastersCourseCertificate`. This will prevent the certificates creation before the freeze of final grades.

The test creates a final grade with `passed=True` and `status='pending'` and checks that no course certificate is created for it.

### DIFF
--- a/grades/tasks_pytest_test.py
+++ b/grades/tasks_pytest_test.py
@@ -44,6 +44,7 @@ def test_generate_course_certificates():
     )
     FinalGradeFactory.create(course_run__course=non_fa_course, passed=True)
     FinalGradeFactory.create(course_run__course=course, passed=False)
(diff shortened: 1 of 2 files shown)


grades/tasks_pytest_test.py
```diff
@@ -44,6 +44,7 @@ def test_generate_course_certificates():
     )
     FinalGradeFactory.create(course_run__course=non_fa_course, passed=True)
     FinalGradeFactory.create(course_run__course=course, passed=False)
+    FinalGradeFactory.create(course_run__course=course, passed=True, status='pending')
```
pdpinch: I get why this is the only change necessary, but can you update the comment on #L75-L76 to explain?

PR #3686, opened 2017-10-31
Fixed discussion user sync issue for users with no profile

rhysyngsun:
N/A

#### What's this PR do?
This fixes edge cases where a user has no profile. This would only happen for admin users or simialr edge cases.

Tests should pass

### DIFF
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -275,6 +275,7 @@ def get_membership_ids_needing_sync():
     """
     return PercolateQueryMembership.objects.filter(
         query__source_type=PercolateQuery.DISCUSSION_CHANNEL_TYPE,
+        user__profile__isnull=False,
         needs_update=True
     ).values_list("id", flat=True)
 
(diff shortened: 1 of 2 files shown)


discussions/api.py
```diff
@@ -90,7 +90,10 @@ def create_discussion_user(discussion_user):
     Raises:
         DiscussionUserSyncException: if there was an error syncing the profile
     """
-    profile = discussion_user.user.profile
+    try:
```
noisecapella: It seems strange to me to have the code raise a `DiscussionUserSyncException` if the API call fails but silence this error. It seems like we would want to raise an error since it's an error case, we just don't want it to halt the processing in the task or fill sentry with useless errors. What do you think of filtering out the missing profile and missing discussion username cases in `get_membership_ids` instead?

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/396062807/

discussions/api.py
rhysyngsun: @noisecapella I added code to filter out in `get_membership_ids`, the missing username should no longer be a case now too.

PR #3687, opened 2017-11-01
Fixed an exception, happens when ZenDesk floating widget is not loaded

amir-qayyum-khan:
- fixes https://github.com/mitodl/micromasters/issues/3602#event-1308561463
- https://sentry.io/mit-office-of-digital-learning/micromasters/issues/361180415/events/8370678176/

#### What's this PR do?
It adds a checks/conditions to avoid rendering of zenDesk widget button logic, it happens when button is not loaded in document.

- I was not able to produce the issue, but try to refresh page and look for below exception on console.

```trace
TypeError
Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
```

@pdpinch

### DIFF
--- a/static/js/entry/zendesk_widget.js
+++ b/static/js/entry/zendesk_widget.js
@@ -100,6 +100,10 @@ const zendeskCallbacks = {
 
   launcherLoaded: () => {
     const iframe = document.querySelector("iframe.zEWidget-launcher")
+    if (_.isNull(iframe)) {
+      return
+    }
+
     const btn = iframe.contentDocument.querySelector(".u-userLauncherColor")
     if (_.isNull(btn)) {
       return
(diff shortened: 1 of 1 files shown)


static/js/entry/zendesk_widget.js
```diff
@@ -100,9 +101,16 @@ const zendeskCallbacks = {
   launcherLoaded: () => {
     const iframe = document.querySelector("iframe.zEWidget-launcher")
```
pdpinch: Is it possible the iframe is null too?

amir-qayyum-khan: yes it is possible

amir-qayyum-khan: added a check for it as well

PR #3700, opened 2017-11-02
Count cache update failure for user

### LINKED ISSUE #2132: Search: edX returns 401 error on enrollments API
(this is a search issue because search results won't be accurate without accurate enrollment data)

This is happening in production, but only a few times. I don't think there's any user impact, because the server should simply re-authenticate with the refresh token, but it would be good to confirm and  quiet the error. 

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/193079253/

```
HTTPError: 401 Client Error: UNAUTHORIZED for url: https://courses.edx.org/api/enrollment/v1/enrollment
  File "dashboard/tasks.py", line 137, in batch_update_user_data_subtasks
    CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
  File "dashboard/api_edx_cache.py", line 282, in update_cache_if_expired
    update_func(user, edx_client)
  File "dashboard/api_edx_cache.py", line 165, in update_cached_enrollments
    enrollments = edx_client.enrollments.get_student_enrollments()
  File "edx_api/enrollments/__init__.py", line 36, in get_student_enrollments
    resp.raise_for_status()
  File "requests/models.py", line 862, in raise_for_status
    raise HTTPError(http_error_msg, response=self)

Unable to refresh cache enrollment for student NimishSharma
```

### LINKED ISSUE #2383: Limit the number of times celery will attempt to connect with a bad token
Sometimes the OAuth token we store for users will stop working and edX will return a 401 error when we periodically update a user's enrollmentCache, certificateCache and gradeCache. When the user actually logs in to the portal, this issue is fixed when they reauthenticate. But until that happens, we generate thousands of 401 errors in our logs whenever celery runs the cache updates. See issues like #2101, #2099, #2132). 

To mitigate this, we can keep track of failed attempts, and skip the celery update after the failures have gone over a threshold of 3 attempts. 

We should continue to raise exceptions for the initial failures. And we should log whenever a cache update is skipped because the failure count is over the threshold. We may want to periodically look at this numbers to help understand the root causes of this problem. 

### acceptance criteria
- model for tracking the errors
- increment error count when celery tasks encounter a 401
- reset the error count when celery successfully authenticates
- reset the error count when the user successfully authenticates
- log anytime a task a skipped because the error count is too high
- continue to raise an exception on initial failures (so we are alerted to any new, systemic problems)

annagav:
Fix #2383 
Fix #2132

#### What's this PR do?
Count the number of authentication failures per student, and after the third one do not try to update cache for that user until that user logs in and updates the credentials.

Play around with making your user credentials invalid, and then calling `refresh_user_data`.

### DIFF
--- a/backends/views.py
+++ b/backends/views.py
@@ -4,6 +4,8 @@ from django.views.decorators.cache import never_cache
 from django.views.decorators.csrf import csrf_exempt
 from social_django.views import complete as social_complete
 from social_django.utils import psa
+from django_redis import get_redis_connection
+from dashboard.api import CACHE_KEY_FAILURE_NUMS_BY_USER, FIELD_USER_ID_BASE_STR, CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE
 
 
 @never_cache
@@ -24,4 +26,13 @@ def complete(request, *args, **kwargs):
         request.session[key] = backend_state
 
     # Continue with social_core pipeline
-    return social_complete(request, *args, **kwargs)
+    social_complete_rtn = social_complete(request, *args, **kwargs)
+
+    # Update redis cache if user had invalid credentials
+    con = get_redis_connection("redis")
+    user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
+    if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
+        con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
+        con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
+
+    return social_complete_rtn
--- a/backends/views.py
+++ b/backends/views.py
@@ -29,10 +29,11 @@ def complete(request, *args, **kwargs):
     social_complete_rtn = social_complete(request, *args, **kwargs)
 
     # Update redis cache if user had invalid credentials
-    con = get_redis_connection("redis")
-    user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
-    if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
-        con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
-        con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
+    if request.user.is_authenticated():
+        con = get_redis_connection("redis")
+        user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
+        if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
+            con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
+            con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
 
     return social_complete_rtn
--- a/backends/views.py
+++ b/backends/views.py
@@ -32,8 +32,7 @@ def complete(request, *args, **kwargs):
     if request.user.is_authenticated():
         con = get_redis_connection("redis")
         user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
-        if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
-            con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
-            con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
+        con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
+        con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
 
     return social_complete_rtn
--- a/backends/views.py
+++ b/backends/views.py
@@ -33,6 +33,6 @@ def complete(request, *args, **kwargs):
         con = get_redis_connection("redis")
         user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
         con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
-        con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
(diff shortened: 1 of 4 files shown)


backends/views.py
```diff
@@ -24,4 +26,13 @@ def complete(request, *args, **kwargs):
         request.session[key] = backend_state
```
noisecapella: Can you write a test covering this code?

backends/views.py
```diff
@@ -24,4 +26,13 @@ def complete(request, *args, **kwargs):
...
 
     # Continue with social_core pipeline
-    return social_complete(request, *args, **kwargs)
+    social_complete_rtn = social_complete(request, *args, **kwargs)
```
noisecapella: I think this should be within the `if request.user.is_authenticated():` block, it wouldn't make sense to run it on anonymous users

backends/views.py
```diff
@@ -24,4 +26,13 @@ def complete(request, *args, **kwargs):
...
+
+    # Update redis cache if user had invalid credentials
+    con = get_redis_connection("redis")
+    user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
+    if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
+        con.hdel(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key)
+        con.srem(CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE, user_key)
+
+    return social_complete_rtn
```
noisecapella: Can we call `social_complete(request, *args, **kwargs)` here? If there's no good reason to call it before the redis update vs after, we should just call it here so our code is grouped together

dashboard/api.py
```diff
@@ -592,5 +598,20 @@ def refresh_user_data(user_id):
...
             CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
         except:
+            save_cache_update_failure(user_id)
             log.exception("Unable to refresh cache %s for student %s", cache_type, user.username)
             continue
+
+
+def save_cache_update_failure(user_id):
+    """
+    Store the number of time update cache failed for a user
```
noisecapella: Can you add the argument to the docstring?

dashboard/api.py
```diff
@@ -592,5 +598,20 @@ def refresh_user_data(user_id):
         try:
             CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
         except:
+            save_cache_update_failure(user_id)
```
noisecapella: I'm not sure we want to call this function in both places, if there is an exception raised it would have a different reason for each case. We should handle only the specific error we expect, which is in the above except clause. Can you remove this line?

dashboard/api.py
```diff
@@ -32,6 +33,10 @@
 COURSE_GRADE_WEIGHT = 0.4
 EXAM_GRADE_WEIGHT = 0.6
 
+CACHE_KEY_FAILURE_NUMS_BY_USER = "update_cache_401_failure_numbers"
```
noisecapella: What's the difference between these two keys? It seems like if we store the number of failures we can put the logic to determine if that's too many in the place where we do the update. We don't need an extra field showing whether or not to update the user

noisecapella:
It looks like the PR records when a 401 auth fails but it doesn't prevent future attempts

backends/views.py
annagav: the redis operations should be after we call `social_complete` because the user might be logged out at that point.

backends/views.py
annagav: which code are you referring to here?

dashboard/api.py
annagav: this is  for this error https://sentry.io/mit-office-of-digital-learning/micromasters/issues/228392933/

annagav:
@noisecapella I think I addressed all your comments.

backends/views.py
noisecapella: This line, the social complete view code

dashboard/api.py
noisecapella: Can you add a comment for each of these keys describing what information is stored there?

dashboard/api_test.py
```diff
@@ -1913,6 +1917,18 @@ def test_calculate_expired(users_without_with_cache, enrollment, certificate, cu
     cache.save()
 
     expected = needs_update + [up_to_date[0]] if expired else needs_update
+    assert sorted(api.calculate_users_to_refresh_in_bulk()) == sorted([user.id for user in expected])
+
+
+@patch("dashboard.api.CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE", SET_USER_IDS_NOT_TO_UPDATE)
```
noisecapella: Can we do this with a fixture instead? The fixture should patch the two constants and also clear the values beforehand so that it's always empty when the test runs

dashboard/api.py
```diff
@@ -592,5 +606,23 @@ def refresh_user_data(user_id):
...
+def save_cache_update_failure(user_id):
+    """
+    Store the number of time update cache failed for a user
+
+    Args:
+        user_id (int): The user id
+    """
+    con = get_redis_connection("redis")
+    user_key = FIELD_USER_ID_BASE_STR.format(user_id)
+    if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
```
noisecapella: It looks like this check isn't necessary. If the key doesn't exist `hincrby` assumes a starting value of 0 and adds to that: https://redis.io/commands/hincrby

backends/views.py
```diff
@@ -24,4 +26,14 @@ def complete(request, *args, **kwargs):
...
 
     # Continue with social_core pipeline
-    return social_complete(request, *args, **kwargs)
+    social_complete_rtn = social_complete(request, *args, **kwargs)
+
+    # Update redis cache if user had invalid credentials
+    if request.user.is_authenticated():
+        con = get_redis_connection("redis")
+        user_key = FIELD_USER_ID_BASE_STR.format(request.user.id)
+        if con.hexists(CACHE_KEY_FAILURE_NUMS_BY_USER, user_key):
```
noisecapella: It looks like this check is not necessary. If `hdel` or `srem` are operating on a value which doesn't exist it just returns a 0: https://redis.io/commands/hdel and https://redis.io/commands/srem

noisecapella:
I pushed a fix for the test

PR #3725, opened 2017-12-15
Switched create channel to new description field

### LINKED ISSUE #3715: Switch to open-discussions description field for channel creation
We currently set `public_description`, we should instead use `description` since that's what we display on open-discussions.

rhysyngsun:
Closes #3715 

#### What's this PR do?
Switches to the new field we're using for channel descriptions

Create a new channel using markdown in description, ensure this shows up in the channel sidebar once created.

### DIFF
--- a/static/js/flow/discussionTypes.js
+++ b/static/js/flow/discussionTypes.js
@@ -17,10 +17,10 @@ export type Post = {
 type ChannelType = "private" | "public";
 
 export type ChannelInputs = {
-  name:               string,
-  title:              string,
-  public_description: string,
-  channel_type:       ChannelType,
+  name:         string,
+  title:        string,
+  description:  string,
+  channel_type: ChannelType,
 };
 
 export type Filter = {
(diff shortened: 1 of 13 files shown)


static/js/flow/discussionTypes.js
```diff
@@ -19,7 +19,7 @@ type ChannelType = "private" | "public";
 export type ChannelInputs = {
   name:               string,
   title:              string,
-  public_description: string,
+  description: string,
```
noisecapella: Can you adjust the indentation here and below? I guess the reformatter isn't aware of flow type declarations

rhysyngsun: Yeah, these need to be manually formatted actually cause the formatter actually removes the alignments.

noisecapella: That's fine, let's just leave it then

rhysyngsun: Fixed it anyway since it was small/quick

PR #3756, opened 2018-01-11
Schedule freeze grades task

### LINKED ISSUE #2223: Final Grades: add celerybeat hook to run the freeze automatically via cron
#2126 added the tasks to freeze the final grades.
this issue is to add a hook in celerybeat to https://github.com/mitodl/micromasters/blob/master/grades/tasks.py#L23 to run the freeze every X hours

**IMPORTANT**: this should not be done until we are confident that everything works fine.

annagav:
Fix #2223

#### What's this PR do?
Schedule the freezing of all freezable course runs to run 3 days a week.

N/A

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -497,7 +497,7 @@ CELERY_BEAT_SCHEDULE = {
         'schedule': crontab(minute='*', hour='*')
     },
     'freeze-final-grades-3-days-a-week': {
-        'task': 'grades.find_course_runs_and_freeze_grades',
+        'task': 'grades.tasks.find_course_runs_and_freeze_grades',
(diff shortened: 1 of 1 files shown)


micromasters/settings.py
```diff
@@ -496,6 +496,10 @@
         'task': 'discussions.tasks.sync_channel_memberships',
         'schedule': crontab(minute='*', hour='*')
     },
+    'freeze-final-grades-3-days-a-week': {
+        'task': 'grades.find_course_runs_and_freeze_grades',
```
noisecapella: I think this needs to be `grades.tasks.find_course_runs_and_freeze_grades`

annagav:
@noisecapella, fixed!

Issue #3667, opened 2017-10-25
Add course number to course model 


pdpinch:
As a data consumer, I'd like to be able to see data correlated to course number which is easier to recognize in graphs and charts. 

As an admin doing customer support, I would be able to identify course enrollments and registrations more quickly if I could see the course number instead of reading the course name. 

Note that for SCM we added the course number to the course title, so that it displays on the dashboard. This could be refactored also to avoid putting course numbers into the titles.

amir-qayyum-khan:
@pdpinch what will be format of course number, should it be auto generated?

pdpinch:
Just a string — the course number is similar to the 2nd segment of the course key, e.g. 14.73x or SC0x.

I’ll be setting the values for these manually, at least for the foreseeable future.

pdpinch:
@maxliu I just wanted to let you know that I've updated the course number data in production, so it should now be available to you in BI. Let me know if you have any questions.

PR #3774, opened 2018-01-19
Added course num to course model

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3667

#### What's this PR do?
adds a new column name course_num in course model

login to admin app, go to course model see this field

@pdpinch 
screen shot 2018-01-19 at 6 13 04 pm


pdpinch:
Can we spell out "course_number" instead of abbreviating it to "course_num"

Is there any way we can check the field length with edX? 50 characters seems like plenty, but it would be most useful to have the same definition as edX.

amir-qayyum-khan:
@pdpinch that field is a text no length specified. 
- https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L712
- https://github.com/edx/opaque-keys

PR #3781, opened 2018-01-22
Update update_docker_hub.sh to use a new hash for each image

noisecapella:
Related to https://github.com/mitodl/cookiecutter-djangoapp/pull/120

#### What's this PR do?
Updates `update_docker_hub.sh` to generate a new hash for each updated image

N/A

### DIFF
--- a/travis/update-docker-hub.sh
+++ b/travis/update-docker-hub.sh
@@ -1,8 +1,18 @@
 #!/bin/bash
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


pdpinch:
What's blocking this?

noisecapella:
The `check_pip.sh` script is hard to test because most of the time the travis docker image will get updated during the tests, and then there's no difference so the tests pass. I probably should mark this work in progress instead

noisecapella:
You should be able to run it without interfering with existing docker images

PR #3788, opened 2018-01-23
Create DiscussionUser on attempted navigation to open-discussions

### LINKED ISSUE #3736: UnableToAuthenticateDiscussionUserException: Unable to generate a JWT token for user
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/363706532/

```
UnableToAuthenticateDiscussionUserException: Unable to generate a JWT token for user
(1 additional frame(s) were not displayed)
...
  File "discussions/views.py", line 68, in discussions_redirect
    raise UnableToAuthenticateDiscussionUserException("Unable to generate a JWT token for user")

UnableToAuthenticateDiscussionUserException: Unable to generate a JWT token for user
```

rhysyngsun:
Fixes #3736 

#### What's this PR do?
Whenever a user goes to /discussions, we will now block and force creation of the discussion user. The reason to do this is to address a race condition between user signup and syncing on the new user.This only fixes the user issue, they would more than likely not see any channels, which we may have to fix separately.

To reliably reproduce, delete your discussion user and navigate to /discussions

### DIFF
--- a/discussions/utils.py
+++ b/discussions/utils.py
@@ -3,10 +3,11 @@ from django.conf import settings
 from django.core.urlresolvers import reverse
 from open_discussions_api import utils
 
+from discussions import api
 from discussions.models import DiscussionUser
 
 
-def get_token_for_user(user, auth_url=None, session_url=None):
+def get_token_for_user(user, auth_url=None, session_url=None, force_create=False):
     """
     Generates a token for the given user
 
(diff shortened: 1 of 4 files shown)


aliceriot:
I'm able to delete my `DiscussionUser` record and then immediately navigate to `/discussions` without any problems. It does seem my user is not subscribed to the channels that the old `DiscussionUser` was subscribed to, but I suppose we can tackle that in a follow-up.

aliceriot:
oops ignore that comment just saw you mentioned the channel issue above! haha

PR #3791, opened 2018-01-25
Added CombinedFinalGrade model

### LINKED ISSUE #3790: Persist course "final grade" for courses with proctored exams
For courses with an online edX grade and a face-to-face proctored exam grade, we use a weighted average to give the learner a combined final grade and determine if they passed the course. Currently we calculate this in the front-end in the dashboard, and also in our analytics pipeline. Shortly we will want to use this combined grade in the learner search index as well.  

To avoid inconsistencies and duplicate code, we should centralize the business logic for calculating the combined final grade and persist it in the database. 

### acceptance criteria
- [ ] add a CombinedFinalGrade class to the grades model (suggested name)
- [ ] create/update the CombinedFinalGrade whenever a proctored exam grade is created/updated

### related issues
- [ ] update the search index to use the CombinedFinalGrade for courses that have it
- [ ] update the dashboard API to include the CombinedFinalGrade for course runs that have one
(tech debt)
- [ ] update the dashboard front-end to use the API value for CombinedFinalGrade instead of calculating it on the fly (tech debt)

annagav:
Fix #3790

#### What's this PR do?
Add a CombinedFinalGrade model, add signal to ProctoredExamGrade to create/update the CombinedFinalGrade for user.

Run migrations, tests should pass


rhysyngsun:
@annagav both travis builds failed could you look into those and then ping me when it's all set?

annagav:
@rhysyngsun, I fixed the tests. Thanks!

grades/api.py
```diff
@@ -230,3 +232,33 @@ def generate_program_certificate(user, program):
         user.username,
         program.title
     )
+
+
+def update_combined_final_grade(user, course):
+    """
+    Update or create CombineFinalGrade
```
rhysyngsun: Nit: typo in "Combine[d]FinalGrade"

grades/api.py
```diff
@@ -230,3 +232,33 @@ def generate_program_certificate(user, program):
...
+    Args:
+        user (User): a django User
+        course (Course): a course model object
+
+    Returns:
```
rhysyngsun: You can just omit the `Returns` section here since nothing is returned

grades/api.py
```diff
@@ -230,3 +232,33 @@ def generate_program_certificate(user, program):
         user.username,
         program.title
     )
+
+
+def update_combined_final_grade(user, course):
```
rhysyngsun: Suggestion: add `_or_create` to the function name so the side effects are clearer

grades/api.py
```diff
@@ -230,3 +232,33 @@ def generate_program_certificate(user, program):
...
+    if final_grade is None:
+        log.error('User [%s] does not have a final for course [%s]', user, course)
+        return
+
+    best_exam = mmtrack.get_best_proctored_exam_grade(course)
+    if best_exam is None:
+        log.error('User [%s] does not have a passing exam grade for course [%s]', user, course)
+        return
+
+    calculated_grade = round(final_grade.grade_percent * COURSE_GRADE_WEIGHT + best_exam.score * EXAM_GRADE_WEIGHT, 1)
```
rhysyngsun: It would probably be good to refactor this and the calculation in `dashboard/api.py` into a util function so we don't have to keep those implementations in sync. Although, it also might make sense to just merge your changes from the other PR into this one since it solves the duplicate code issue too.

annagav: So in my other PR I am removing this calculation in `dashboard/api.py`. Do you still want me to refactor this into utils?

grades/management/commands/populate_combined_final_grades.py
```diff
@@ -0,0 +1,29 @@
...
+
+from courses.models import Course
+from grades.api import update_or_create_combined_final_grade
+from grades.models import ProctoredExamGrade
+from micromasters.utils import now_in_utc
+
+
+class Command(BaseCommand):
+    """
+    Finds all passing exam grades for courses that have frozen runs
```
rhysyngsun: Could you update this description and the help to indicate this generates the records in `CombinedFinalGrade`?

dashboard/api.py
```diff
@@ -514,28 +510,20 @@ def get_certificate_url(mmtrack, course):
...
-        return ""
-    if not course.has_exam:
-        return str(round(best_grade.grade_percent))
-
-    best_exam = mmtrack.get_best_proctored_exam_grade(course)
-    if best_exam is None:
-        return ""
-
-    return str(round(best_grade.grade_percent * COURSE_GRADE_WEIGHT + best_exam.score * EXAM_GRADE_WEIGHT))
+    combined_grade = CombinedFinalGrade.objects.filter(user=user, course=course)
```
rhysyngsun: So in testing this, I realized that this is going to cause combined final grades to disappear until we run the populate command. That might be why you split the two PRs in the first place. So I'd suggest that we just feature flag the old code vs. new code so that we continue to use the old code (removed in this diff) until we run the command, then we can set the flag to `True` and switch to looking up `CombinedFinalGrade` and finally follow this up with a tech debt issue to remove the feature flag and old code. Does that make sense?

grades/management/commands/populate_combined_final_grades.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    help = "Finds all users that might need a CombinedFinalGrade"
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        courses = Course.objects.filter(
+            program__live=True,
+            program__financial_aid_availability=True,
+            has_exam=True
```
rhysyngsun: `has_exam` is a `@property` on the `Course` class, so this query ends up failing. We don't have a ton of courses so it may just be simpler to remove this and then filter to `course.has_exam` in code.

annagav:
@rhysyngsun I addressed your comments, thanks!

grades/api.py
```diff
@@ -230,3 +232,31 @@ def generate_program_certificate(user, program):
...
+    """
+    mmtrack = get_mmtrack(user, course.program)
+    final_grade = mmtrack.get_best_final_grade_for_course(course)
+    if final_grade is None:
+        log.error('User [%s] does not have a final for course [%s]', user, course)
+        return
+
+    best_exam = mmtrack.get_best_proctored_exam_grade(course)
+    if best_exam is None:
+        log.error('User [%s] does not have a passing exam grade for course [%s]', user, course)
```
rhysyngsun: Sorry just noticed this, should these be `warn` instead? I don't think we want to get sentry errors every time one of the signals trigger but we decide not to create a `CombinedFinalGrade`

dashboard/api.py
```diff
@@ -516,26 +513,32 @@ def get_certificate_url(mmtrack, course):
 
 def get_overall_final_grade_for_course(mmtrack, course):
     """
-    Calculate overall grade for course
+   Calculate overall grade for course
```
rhysyngsun: Doc comment appears to have shifted one space to the left.

grades/api.py
```diff
@@ -230,3 +232,31 @@ def generate_program_certificate(user, program):
...
+
+def update_or_create_combined_final_grade(user, course):
+    """
+    Update or create CombinedFinalGrade
+
+    Args:
+        user (User): a django User
+        course (Course): a course model object
+    """
+    mmtrack = get_mmtrack(user, course.program)
```
rhysyngsun: Should we also be checking `has_exam` here so we don't bother checking for a course that doesn't have one?

rhysyngsun:
Functionality looks good, so just my last set of comments.

grades/api.py
annagav: Its is called from signals where we save a an exam grade, so the course should have `has_exam` true. And it is called in the management command where we check if the course has exam. Should I check that inside the function anyway?

rhysyngsun: It mostly occurred to me because I saw that it's also called on `post_save` for `FinalGrade` and that we could be executing this against courses in programs other than DEDP. That said, those will never have `ProtoredExamGrade`s created for them, so at worst we're just doing a few unnecessary lookups by not checking for `has_exam`. I don't think there's a functional difference aside from that overhead though so I'm fine either way.

annagav: right, I forgot about the final grade signal

grades/api.py
```diff
@@ -230,3 +232,35 @@ def generate_program_certificate(user, program):
...
+
+def update_or_create_combined_final_grade(user, course):
+    """
+    Update or create CombinedFinalGrade
+
+    Args:
+        user (User): a django User
+        course (Course): a course model object
+    """
+    if course.has_exam:
```
rhysyngsun: Suggest inverting this to a `if not course.has_exam: return` so you don't have to indent the whole method.

annagav:
@rhysyngsun, I think I addressed last set of comments

rhysyngsun:
👍 assuming tests pass

PR #3799, opened 2018-01-31
restrict channel creation to superusers

### LINKED ISSUE #3784: Limit channel creation to django superusers
Until we reach the point where program staff can edit and update channels, it would be simpler to limit channel creation to the ODL engineering team.

aliceriot:

closes #3784 

#### What's this PR do?

This changes the permission for channel creation to just check if the user is a django superuser, instead of looking to see if they have a staff role on the program in question.

I also added a little thing to display the error that is returned by the backend if you try to create a channel but do not have permission.

youdonthaveperms

Few things to check:

- if your django user has `is_superuser` you should be able to create a channel
- if you do not, you should see the above permission error when you try

um, that's it.

### DIFF
--- a/discussions/views_test.py
+++ b/discussions/views_test.py
@@ -115,11 +115,13 @@ def _make_create_channel_input(program_id, description="default description"):
 
 @pytest.mark.parametrize("description", ["public description", ""])
 def test_create_channel(description, mocker, patched_users_api):
-    """Staff can create a channel using the REST API"""
+    """superuser can create a channel using the REST API"""
     client = APIClient()
     role = RoleFactory.create(role=Staff.ROLE_ID)
-    user = role.user
-    client.force_login(user)
+    role.user.is_superuser = True
+    role.user.save()
+
+    client.force_login(role.user)
 
     channel = ChannelFactory.create()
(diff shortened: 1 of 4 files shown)


noisecapella:
Functionality works great, but I'm wondering if we should also remove the button if the user isn't a superuser. Not sure how complicated that gets to implement though. What do you think?

aliceriot:
I'm not too worried about it, given that something like 5 people use this feature currently. do we currently have a way to figure out if a user is a superuser from the frontned?

discussions/views_test.py
```diff
@@ -186,11 +188,23 @@ def test_create_channel_user_without_permission(patched_users_api):
     assert resp.status_code == 403
```
noisecapella: Can you add a check that `add_channel` is also never called, just to verify that the error code and behavior match up?

noisecapella:
I don't think we pass that information to the frontend currently, I'm fine leaving the behavior as is

discussions/views_test.py
aliceriot: sure

aliceriot:
@noisecapella comment addressed!

Issue #3034, opened 2017-04-10
Add test for skip financial aid failure in DashboardPage


noisecapella:
These lines are uncovered:

https://github.com/mitodl/micromasters/pull/3022/files#diff-93b2de43928d583d22fa9a936ba53ad9R411

PR #3805, opened 2018-02-05
fix financial aid skip bug

### LINKED ISSUE #3803: "Skip and Pay Full Price" error
### Steps to Reproduce (on CI and in production)

1. Delete or reset the learner's financial aid object. 
2. Click "Pay Full Price" on the dashboard, and again on the confirmation dialog

### Expected Behavior

- The learner's dashboard is updated with the full price: $1000

### Actual Behavior

- The learner gets a toast message "Failed to Skip Financial Aid"
- However, it appears that it actually succeeded"
  - the API returns a 200 (with no content)
  - the learner's financial aid object has a status of "skipped" 
  - reloading the browser restores the dashboard to the expected state

aliceriot:

closes #3803 

#### What's this PR do?

fixes a bug! the bug was due to our use of the `fetchJSONWithCSRF` function for skipping financial aid. The `fetchJSONWithCSRF` function assumes that the response has a JSON payload, and it tries to parse the payload. If it doesn't parse correctly it throws an error.

The `skipFinancialAid` endpoint doesn't return a JSON payload when it is `PATCH`'ed, so we were trying to parse JSON where there was none to be parsed. Anyway, this meant that basically after the request came back as a 200 but before the 'success' action could be dispatched to the redux store we were throwing an error in the `fetchJSONWithCSRF` function that was causing the failure-case toast message to pop-up.

So, to fix it I just changed the API function to use `fetchWithCSRF` instead, which doesn't assume the body of the request contains anything in particular, and will only reject if HTTP code is an error code. This should fix the error in production. I also did a small bit of cleanup to the relevant function in the `DashboardPage` component.

Delete your financial aid or otherwise do something to get rid of it. Confirm that, on master, if you try to skip financial aid it will a) work successfully and return a 200 and b) show the 'failure' message and fail to go on to refresh the dashboard and course price information.

Then, get rid of your financial aid again, check out this branch, and confirm that my changes fix the issue. So skipping financial aid should work, and after it goes through successfully you should see the dashboard refresh so that the new course price information can be reflected in the UI.

### DIFF
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -465,7 +465,7 @@ class DashboardPage extends React.Component {
     )
   }
 
-  skipFinancialAid = programId => {
+  skipFinancialAid = async programId => {
     const { dispatch, financialAid } = this.props
 
     const program = this.getCurrentlyEnrolledProgram()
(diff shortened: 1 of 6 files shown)


pdpinch:
Does this fix #3034 as well? @noisecapella was looking to see these lines get test coverage:

https://github.com/mitodl/micromasters/pull/3022/files#diff-93b2de43928d583d22fa9a936ba53ad9R411

aliceriot:
No it doesn't fix that now, I can add a test though.

static/js/containers/DashboardPage.js
```diff
@@ -477,20 +477,18 @@ class DashboardPage extends React.Component {
...
-              icon:    TOAST_FAILURE
-            })
-          )
-        }
-      )
+      try {
+        await dispatch(skipFinancialAid(programId))
+        this.setConfirmSkipDialogVisibility(false)
+      } catch (_) {
+        this.setConfirmSkipDialogVisibility(false)
```
mbertrand: Might be good to add a unit test for this block to increase coverage.

aliceriot:
@mbertrand comments addressed!

PR #3809, opened 2018-02-06
count_courses_passed for courses with exams

### LINKED ISSUE #3777: Search: include exam results in Nr of courses passed facet
DEDP Staff users expect the number of courses passed facet to be based on the results of the online course and the proctored exam, not just the online course. 

Note that none of the upcoming micromasters will use a proctored, face-to-face exam similar to DEDP. It's not necessary to generalize this use case. 

### acceptance criteria
- [ ] special case for courses with a proctored exam
- [ ] when updating the search index for number of courses passed, use the weighted average (40%/60%) of the online course grade and the proctored exam grade. An average greater than 50% is passing.  
- [ ] signal to update the search index when the proctored exam grades are updated

### related issues
- [ ] assuming the business logic for calculating the weighted average is moved to the back-end, we should also update the dashboard API and front-end to use the same method. 
- [x] #3790 Persist course "final grade" for courses with proctored exams

annagav:
Fix #3777

#### What's this PR do?
Changes the logic of count_courses_passed in MMTrack to count differently for financial aid programs. If a course has exam then existence of the CombinedFinalGrade counts, if the course doesn't have an exam then we check for a passing FinalGrade.

Set a course run to past_passed, create ExamRun, create CombinedFinalGrade.
Run recreate index.
You learner should appear in the search results when you slide the "# of courses passed" facet.

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -8,7 +8,7 @@ from django.db import transaction
 from django.db.models import Q, Count
 from django.urls import reverse
 
-from courses.models import CourseRun, Course
+from courses.models import CourseRun
 from dashboard.api_edx_cache import CachedEdxUserData
 from ecommerce.models import Order, Line
 from grades.constants import FinalGradeStatus
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -8,7 +8,7 @@ from django.db import transaction
 from django.db.models import Q, Count
 from django.urls import reverse
 
-from courses.models import CourseRun
+from courses.models import CourseRun, Course
 from dashboard.api_edx_cache import CachedEdxUserData
 from ecommerce.models import Order, Line
 from grades.constants import FinalGradeStatus
@@ -16,6 +16,7 @@ from grades.models import (
     FinalGrade,
     ProctoredExamGrade,
     MicromastersProgramCertificate,
+    CombinedFinalGrade
 )
 from exams.models import (
     ExamProfile,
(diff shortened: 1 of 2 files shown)


dashboard/utils.py
```diff
@@ -363,11 +364,21 @@ def count_courses_passed(self):
...
         """
-        return (
-            self.final_grade_qset.for_course_run_keys(self.edx_course_keys).passed()
-            .values_list('course_run__course__id', flat=True)
-            .distinct().count()
-        )
+        if self.program.financial_aid_availability:
+            courses = Course.objects.filter(program=self.program)
+            return sum([
+                CombinedFinalGrade.objects.filter(user=self.user, course=course).exists()
```
rhysyngsun: I think this should be checking if the `CombinedFinalGrade` is considered passed, right? Maybe we need to add a `passed` field to that model like in `FinalGrade` so you can then filter and do a simple `exists()` or even `filter(course__in=[course for course in courses if course.has_exam]).count()`?

If you do the latter, you could probably then refactor the `else` clause here into a separate function that takes a lsit of `edx_course_keys` as a parameter and pass `[course.edx_course_key for course in course if not course.has_exam]` which would cut the query overhead for financial aid programs from N (# of courses) to 2.

annagav: I think the idea for CombinedFinalGrade was that you get one only if you passed the course.

annagav: Meaning CombinedFinalGrade is created only if the learner passed any course_run and has a passing exam grade (>50%)

rhysyngsun: That doesn't appear to be enforced in `update_or_create_combined_final_grade` and I'm not sure we'd want to do that because wouldn't that mean if you didn't pass you never see the grade on your dashboard?

annagav: yeah, that correct. If you don't pass you just see the edx grade and the exam grade but not the combined grade. That is how it was originally implemented, before `CombinedFinalGrade`. I also confirmed that with @pdpinch when working on `CombinedFinalGrade`

annagav: In `update_or_create_combined_final_grade`, `get_best_final_grade_for_course` get only passing final grades and `get_best_proctored_exam_grade` gets only passing exam grades.

rhysyngsun: Ah ok, missed that. Thanks for clarifying.

rhysyngsun: I think it would still be worth trying to cut down on the query count though, right? While it might not be too noticeable on the dashboard it'll probably add up on reindexing everything since the queries are executed serially.

rhysyngsun:
Confirmed that the functionality works as expected.

rhysyngsun:
@annagav looks good

Issue #3783, opened 2018-01-22
When creating discussion user, pass email address to OD


pdpinch:
In order to allow the OD app to send via email, MicroMasters has to pass along the user's email address. 

- [x] Update `open-discussions-client` to the most recent version (note: there were function signature changes on `client.users.create()` and  `client.users.update()` required to support the new fields for this ticket. Need to adjust existing code in MM to account for that
- [x] Pass `email` on both `create` and `update`
- [x] Only pass `email_optin` on `create`

### questions
- [x] How should we handle users with `email_optin = false`? We could pass along that setting, or we could simply not pass an email address to the discussions app

### related issues
- Batch task to update existing user's email addresses
- Allow MicroMasters user to update email address

pdpinch:
This is blocked by updates to Open Discussion API and the Open Discussions client.

rhysyngsun:
I think it would be good to push the email plus an optin flag even if opting out of emails, it could help with reconciling multiple accounts down the road.

PR #3822, opened 2018-02-09
When creating discussion user, added email address to OD

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3783

#### What's this PR do?
passes email address and email optin flag to OD user api

- set `-e git+https://github.com/mitodl/open-discussions-client.git@master#egg=open-discussions-client` in requirments
- You should be able to go to /discussions/ in MM and that view will force-create it
@pdpinch

### DIFF
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -96,12 +96,14 @@ def create_discussion_user(discussion_user):
 
     api = get_staff_client()
     result = api.users.create(
-        name=profile.full_name,
         email=profile.user.email,
-        email_optin=profile.email_optin,
-        image=profile.image.url if profile.image else None,
-        image_small=profile.image_small.url if profile.image_small else None,
-        image_medium=profile.image_medium.url if profile.image_medium else None,
+        profile=dict(
+            name=profile.full_name,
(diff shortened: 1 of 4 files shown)


discussions/api.py
```diff
@@ -97,6 +97,8 @@ def create_discussion_user(discussion_user):
     api = get_staff_client()
     result = api.users.create(
         name=profile.full_name,
+        email=profile.user.email,
+        email_optin=profile.email_optin,
         image=profile.image.url if profile.image else None,
```
rhysyngsun: Sorry I missed this on the ticket, but the profile fields have been refactored in `open-discussion-client` to a single `profile=` arg. See https://github.com/mitodl/open-discussions-client/blob/master/open_discussions_api/users/client_test.py#L26-L35

Please also functionally test this code against open-discussions before putting it back in review.

amir-qayyum-khan: @rhysyngsun can you give me overview how to test it?

rhysyngsun:
Code looks good, testing the functionality now.

PR #3827, opened 2018-02-09
Fix missing field

noisecapella:
None

#### What's this PR do?
Fixes a missing field in the mapping which was preventing reindexing

Remove the field added in this PR from the mapping and rerun the test. The test will fail


blarghmatey:
:+1:

PR #3830, opened 2018-02-12
Disable dynamic mapping

### LINKED ISSUE #3829: Turn off dynamic mapping for Elasticsearch
It appears that the new percolate query index does not work well with dynamic mapping. If a field specified in the query does not exist then the percolate query indexing fails. We don't really have a use case for dynamic mapping anyway, Elasticsearch includes it for easier prototyping. We should just put all fields we care about in the mapping explicitly

noisecapella:
Fixes #3829 

#### What's this PR do?
Turns off dynamic mapping and adds all fields explicitly to the mapping. In Elasticsearch 5 the percolate query requires all fields in the mapping to already exist or else it errors, so we need to add them proactively. We don't really have a use case for dynamic mapping anyway, it is meant as a prototyping feature, so it's better to just turn it off.

Reindexing should work without error, learner search page should work


search/indexing_api.py
```diff
@@ -136,15 +136,26 @@
...
                     'filled_out': BOOL_TYPE,
                     'first_name': FOLDED_SEARCHABLE_KEYWORD_TYPE,
                     'full_name': FOLDED_SEARCHABLE_KEYWORD_TYPE,
                     'gender': KEYWORD_TYPE,
                     'last_name': FOLDED_SEARCHABLE_KEYWORD_TYPE,
+                    'nationality': KEYWORD_TYPE,
                     'preferred_language': KEYWORD_TYPE,
                     'preferred_name': FOLDED_SEARCHABLE_KEYWORD_TYPE,
                     'pretty_printed_student_id': KEYWORD_TYPE,
+                    'image': KEYWORD_TYPE,
```
rhysyngsun: Nit: suggest moving these up to keep with the lexicographical sort we've been using on key names.

rhysyngsun:
looks good assuming tests pass

PR #3841, opened 2018-02-15
Serialize best final grades for search

### LINKED ISSUE #3810: Misleading final grade search results
The way enrollments are serialized [here](https://github.com/mitodl/micromasters/blob/master/dashboard/serializers.py#L17) only the first found paid course run enrollment is serialized and indexed, and that is the course run grade and semester that will show up in the search results.

It looks like for the final grade facet we should pick the best final grade the learner earned for the course.

annagav:
Fix #3810

#### What's this PR do?
Change the serialization for enrollments to include the best passing grade for the course.

Run an alter data command for you learner to set few enrollments to 'passed', you can pass the desired grade as an argument. Then make sure the highest grade appears in the final grade facet.

### DIFF
--- a/dashboard/utils_test.py
+++ b/dashboard/utils_test.py
@@ -7,6 +7,8 @@ from unittest.mock import (
     MagicMock,
     PropertyMock
 )
+
+from django.test import override_settings
 from django.urls import reverse
 
 import pytz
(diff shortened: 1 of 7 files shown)


dashboard/utils_test.py
```diff
@@ -733,6 +735,21 @@ def test_get_best_final_grade_for_course(self):
...
+        """
+        Test for get_overall_final_grade_for_course to return CombinedFinalGrade for course
+        """
+        mmtrack = MMTrack(
+            user=self.user,
+            program=self.program_financial_aid,
+            edx_user_data=self.cached_edx_user_data
+        )
+        finaid_course = self.crun_fa.course
+        assert mmtrack.get_overall_final_grade_for_course(finaid_course) == ""
```
mbertrand: Might be good to parametrize this test (or add more tests) to check for result when: 
  - `Course.has_exam` is True or False
  - `MMTrack.get_best_proctored_exam_grade` is None or between 0-1
  -  Grades for course runs exist or not

annagav: That part that is under `else` we were going to get rid of once CombinedFinalGrade is populated in production. I can remove it in this PR if you prefer

mbertrand: OK, never mind then.

Issue #3746, opened 2018-01-05
Upgrade to Django 2


noisecapella:

noisecapella:
Looks like this is blocked by wagtail. However their master branch now supports Django 2.x so it shouldn't be long before the next official release does

noisecapella:
wagtail 2.0b1 is now out and supports Django 2

PR #3847, opened 2018-02-23
Fix some deprecation warnings

noisecapella:
Part of #3746 

#### What's this PR do?
Fixes some deprecation warnings in preparation for the upgrade to Django 2.0

Nothing should break

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,5 @@
 [tox]
 envlist = py36
-skip_missing_interpreters = True
 skipsdist = True
 
 [testenv]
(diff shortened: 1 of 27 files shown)


tox.ini
```diff
@@ -1,6 +1,5 @@
 [tox]
 envlist = py36
-skip_missing_interpreters = True
```
mbertrand: Just curious why this was removed?

noisecapella: If a user was missing Python 3.6 then the tests would pass anyway. Since we only have the one interpreter we should have it fail if it's missing

PR #3863, opened 2018-02-28
Schedule task to create CombinedFinalGrades

annagav:
None

#### What's this PR do?
Set up hourly task to check if any CombinedFinalGrades need to be created.

Tests should pass

### DIFF
--- a/grades/tasks.py
+++ b/grades/tasks.py
@@ -10,7 +10,7 @@ from django.core.cache import caches
 from django.db.models import Count
 from django_redis import get_redis_connection
 
-from courses.models import CourseRun
+from courses.models import CourseRun, Course
 from grades import api
 from grades.constants import FinalGradeStatus
 from grades.models import (
@@ -18,10 +18,10 @@ from grades.models import (
     ProctoredExamGrade,
     MicromastersCourseCertificate,
     CourseRunGradingStatus,
+    CombinedFinalGrade,
 )
 from micromasters.celery import app
-from micromasters.utils import chunks
-
+from micromasters.utils import chunks, now_in_utc
 
 CACHE_ID_BASE_STR = "freeze_grade_{0}"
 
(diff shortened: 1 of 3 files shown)


grades/tasks.py
```diff
@@ -82,6 +82,28 @@ def generate_course_certificates_for_fa_students():
...
+        program__financial_aid_availability=True
+    )
+    for course in courses:
+        if course.has_frozen_runs() and course.has_exam:
+            exam_grades = ProctoredExamGrade.objects.filter(
+                course=course,
+                passed=True,
+                exam_run__date_grades_available__lte=now_in_utc()
+            )
+            users_with_grade = list(CombinedFinalGrade.objects.filter(course=course).values_list('user', flat=True))
```
noisecapella: Since this is only used for the user id lookup this should probably be a `set` instead of a `list`

grades/tasks_pytest_test.py
```diff
@@ -79,3 +79,36 @@ def test_generate_course_certificates():
...
+    )
+    for exam_grade in exam_grades[:3]:
+        CombinedFinalGrade.objects.create(user=exam_grade.user, course=course_run.course, grade=0.7)
+    # Only 3 users will have combined grades
+    for exam_grade in exam_grades[3:]:
+        FinalGradeFactory.create(user=exam_grade.user, course_run=course_run, passed=True)
+
+    tasks.create_combined_final_grades.delay()
+
+    assert update_mock.call_count == 2
```
noisecapella: Can you also assert the call arguments here?

annagav: is it enough to assert the last call arguments?

noisecapella: We should assert both calls, it looks like the arguments should be the exam user and the course so it should be straightforward. You can use `assert_any_call` which will just make sure one of the calls has the arguments

annagav:
@noisecapella I addressed all your comments.

PR #3879, opened 2018-03-07
Fix travis errors

rhysyngsun:
N/A, see also mitodl/open-discussions#335

#### What's this PR do?
Fixes error output on the celery worker in our travis builds. Example: https://travis-ci.org/mitodl/micromasters/jobs/350367945

You should be able to run celery locally and should see no error output from the celery container in the python job.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -88,7 +88,7 @@ services:
     command: >
       /bin/bash -c '
       sleep 3;
-      celery -A micromasters worker -B -l ${MICROMASTERS_LOG_LEVEL:-INFO}'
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -88,7 +88,7 @@ services:
     command: >
       /bin/bash -c '
       sleep 3;
-      celery -A micromasters worker -B -l ${MICROMASTERS_LOG_LEVEL:-INFO}'
+      celery -A micromasters.celery:app worker -B -l ${MICROMASTERS_LOG_LEVEL:-INFO}'
```
noisecapella: Should we also update `Procfile`?

PR #3880, opened 2018-03-07
fix coupon / personalized price copy

### LINKED ISSUE #3804: Coupons: let user know that coupons will still work when paying full price
As a learner who is part of a bulk purchase, I have a 100% off coupon (for a course or for the program) that I would like to use. I find it confusing that I have to fill out the personal pricing form and/or "skip and pay full price" to apply my 100% off coupon.

I think it would help if we had a brief message on the the personal price form and the "skip and pay full price" confirmation dialog that made it clear that coupons will still be applied. Maybe "(If you are using a coupon, it will be reflected when you checkout.)"

aliceriot:

closes #3804 

#### What's this PR do?

changes a little bit of copy to be in line with the image (mainly trying to clarify how coupons interact with personalized pricing)

Make sure that the relevant UI elements are changed in-line with what is in the attached issue

### DIFF
--- a/static/js/components/CourseEnrollmentDialog.js
+++ b/static/js/components/CourseEnrollmentDialog.js
@@ -100,13 +100,17 @@ export default class CourseEnrollmentDialog extends React.Component {
         </Button>
       )
     } else if (hasUserApplied) {
-      message = (
-        <p>
+      message = [
+        <p key="1">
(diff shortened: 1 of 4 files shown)


pdpinch:
@amir-qayyum-khan can you review this?

aliceriot:
I don't think it's super crucial to add unit tests, it's just a copy change and not a logic change

PR #3545, opened 2017-09-21
Payment for courses not course run

giocalitri:
#### What are the relevant tickets?
#3326 

#### What's this PR do?
Modifies the `has_paid` function of the `MMTrack` to return true if any course run of a course has been paid for.

#### How should this be manually tested?
TBD

PR #3896, opened 2018-03-14
Snapshots: add more failed course states

annagav:
Related to #3545

#### What's this PR do?
Add more states for failed course run state with other course runs offered in the future.

Run ./scripts/test/run_snapshot_dashboard_states.sh --match "paid_failed"

Here are some of the states:
This is a state with course runs in the future and fuzzy enrollment dates (both types exist).
screen shot 2018-03-15 at 3 05 21 pm

With currently enrollable run
screen shot 2018-03-15 at 3 07 24 pm

Course run with fuzzy start date
screen shot 2018-03-15 at 3 00 43 pm

No more course runs
screen shot 2018-03-15 at 3 00 31 pm


noisecapella:
Don't worry about the travis lint error, it should be fixed in master

PR #3898, opened 2018-03-15
add .pytest_cache to gitignore

aliceriot:
I'm not sure if we upgraded a version or what, but I've just noticed this new directory being written.


noisecapella:
Yeah I think this is a version upgrade thing

PR #3900, opened 2018-03-15
fix redirect

### LINKED ISSUE #3899: profile redirect behavior is broken

aliceriot:

closes #3899 

#### What's this PR do?

This fixes a redirect issue we were having, where it was possible for the user to navigate to the dashboard even if their profile wasn't completed yet (we assume in a lot of places that this can't happen, so it breaks stuff if a user does this).

The issue with the existing implementation was that it didn't prevent the user from navigating away from the profile page after the first load. So they could go to the profile, click to open the side nav, and then navigate to the dashboard, and the app wouldn't stop them. 

I changed it so that the check is done on every prop change on the top-level App component. So, if the user is on the profile page and then clicks over to the dashboard, they should be redirected back to the profile page (and they should see a little message).

I also realized that we already perform a redirect on the backend if the profile is not yet done. There was some code on App.js that was dealing with the case where the user visits the dashboard first and then gets redirected to the profile page, but that actually can't happen b/c we already redirect on the dashboard. So, I decided to remove that code, both because it was superfluous and b/c keeping it around complicated things.

If you're on the profile page and have an incomplete, clicking (through the nav menu) over to the dashboard should redirect you back to the profile page, and you should see a little toast that says something about filling out your profile. Besides that, you should be able to interact with the whole app as normal.


noisecapella:
I'm not sure why but I don't see the nag message. I changed my profile to not be filled out, then I went to `/learners`. I was redirected to `/profile/personal` but I don't see any toast

noisecapella:
It works fine if I'm already within the app and I click a link or press the back button, it looks like the only case where I don't is if I go straight to one of those pages from outside the app

aliceriot:
Yeah, that's because of the backend redirect - so if you go to `/dashboard` the backend redirects you, so then from the perspective of the frontend the initial page-load happens on `/profile/personal` or whatever. It's kind of unfortunate :(

noisecapella:
Oh right... I guess we can't really fix that

PR #3901, opened 2018-03-15
add reenroll button

### LINKED ISSUE #3780: Dashboard: re-enroll button
According to @annagav in https://github.com/mitodl/micromasters/issues/3233#issuecomment-342169648, we never implemented the re-enroll button in the dashboard redesign. I think it would still be useful. 

paid

aliceriot:

closes #3780 

#### What's this PR do?

If the user passed a course, we show them a little 'view certificate' link. However, in addition to this we'd like to give them the option to re-enroll in the course. This adds that option.

I discovered that the backend code for the dashboard API was set up to not return future course runs, in the case where the user already has a passing final grade on a past run. So, in order to get around that, I needed to make a small change to the dashboard API code, so that even if the user passes it will still return a future enrollable run (so we have something on the frontend in which we can enroll the user).

Get a course set up with two course runs, one, in the past, where you have a certificate and `STATUS_PASSED` and so on, and one, open for enrollment now, where you do not have an enrollment yet. If you go onto master, you should confirm that the passed run is the only one returned by the API. Then checkout this branch, and confirm that, in addition to seeing the 'view certificate' link, you see a link to 'Re-enroll'. Also confirm that the second course run is now returned in the dashboard API response.

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages.js
+++ b/static/js/components/dashboard/courses/StatusMessages.js
@@ -157,6 +157,19 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
           >
             View Certificate
           </a>
+          {S.maybe(
+            null,
+            () => [
+              <span key="first">{" | "}</span>,
+              <a
+                key="second"
+                onClick={() => setShowExpandedCourseStatus(course.id)}
+              >
+                Re-enroll
+              </a>
+            ],
+            futureEnrollableRun(course)
+          )}
         </div>
       )
     })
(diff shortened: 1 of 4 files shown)


annagav:
@pdpinch, @roberthouse54 Do we want to wait until the grades a frozen and/or certificates created before we show the re-enroll button?

pdpinch:
We shouldn't be telling users "You passed this course." or "You failed this course" unless the grades are final.

annagav:
@pdpinch Looking at dashboard api it looks like we say if a user passed based on a passing FinalGrade independently of the CourseRunGradingStatus.

pdpinch:
That's OK. If the learner has a FinalGrade we consider that final. 

The CourseRunGradingStatus is for internal use, just to determine if we need to continue checking for final grades.

annagav:
@aliceriot, once i created the certificate, this is what I got:

screen shot 2018-03-16 at 12 48 25 pm

annagav:
@pdpinch, sounds good. Could you please take a look at my first question again?

aliceriot:
@annagav try pulling down the latest commits, the thing where the links were squished up should be fixed

static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -157,6 +157,15 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
           >
             View Certificate
           </a>
+          {S.maybe(
+            null,
+            () => (
+              <a onClick={() => setShowExpandedCourseStatus(course.id)}>
+                {" | Re-enroll"}
```
annagav: Could you separate the vertical bar and the link? when I hover over it it looks broken.

annagav:
Nothing happens when I click `re-enroll`

screen shot 2018-03-16 at 3 57 32 pm

aliceriot:
hmmm, it was working before...blech

aliceriot:
@annagav ok comments should be addressed!

Issue #3870, opened 2018-03-02
Only create one course certificate per user per course


pdpinch:
With the release of the most recent proctored exam results, it's come to our attention that we are creating duplicate certificates for some learners. We create certificates when a learner passes an edX online course run and the corresponding proctored exam. 26 learners who have passed a proctored exam have taken passed multiple course runs and now have multiple course certificates. See https://bi.odl.mit.edu/queries/179 for a list

Although it hasn't happened yet, I expect additional certificates will be created if and when a learner passes a proctored exam multiple times. 

Instead, we should limit learners to one certificate per course

### acceptance criteria:
- [ ] Before creating a certificate, check the certificates table for existing course certificates for the user and the course in question
- [ ] if the user already has a course certificate, don't create a new one

### related issues
- For users with multiple certificates, link to the one with the best grade

pdpinch:
You told me in #3872 that we have a reliable way to show only the "best" certificate. Given that, I don't think we need to do anything to clean up the existing duplicates -- at least for now. 

I'm a little nervous about just deleting the duplicates, because I don't know if users have linked to them.

annagav:
Currently the MicromastersCourseCertificate is linked to a final grade. So if user passed a course he would have as many course certificates as passing final grades. Multiple exam runs do not lead to the creation of certificates, only multiple final grade do.
If I limit the creation of the certificates a user likely will have a certificate not for the best final grade. So the way we choose which certificate to show is not going to work anymore.

pdpinch:
@annagav I think this is done now that you fixed the migration. I'm going to close this. 

@maxliu FYI, there shouldn't be duplicate course certificates any more.

PR #3904, opened 2018-03-19
Define a model for unique course certificates

annagav:
Related to #3870

#### What's this PR do?
Create a new model for course certificates where user receives only one certificate per course.

Not yet

I haven't worked on tests yet. Looking for approval of the idea first.
I wasn't sure how to name the model, since MicromasterCourseCertificate already exists. FA* helps to remember that those certificates only exist for FA courses.


gsidebo:
After a quick look through this, my main concern is that there would now be 2 different sources of truth for a user's course certificate. I think what we need to do is change the existing model (`MicromastersCourseCertificate`) to mirror this new model you've created, and add a data migration. I just asked Sar how many records are currently in that table right now in production. If that number isn't too high, I would strongly advocate for that. We created that model with an assumption that is no longer true, so I say we need to fix that model instead of adding a new one. 

Possible approach:

*First PR*
- Add `user` and `course` keys to the `MicromastersCourseCertificate` model, set `final_grade` to nullable, and create a migration
- Add a script element to the forward part of that migration. This script would loop over all `MicromastersCourseCertificate` objects, fetch `user` and `course` from the `final_grade` and fill in those values.
- Change the logic of `generate_course_certificates_for_fa_students` to avoid creating `MicromastersCourseCertificate`s with identical `user`/`course` (there might also be some other logic that depends on `MicromastersCourseCertificate.final_grade` - that should be changed too)

*Second PR* (which wouldn't be urgent
- In the `MicromastersCourseCertificate` model, remove `final_grade` and set `user` and `course` to be unique together, and create the migration
- The backward migration would first add the `final_grade` column back in, then run a script to repopulate the `final_grade` column, then remove the `user`/`course` columns.

What do you think?

gsidebo:
We should close this one, right?

Issue #3607, opened 2017-10-05
IntegrityError: null value in column "country_of_residence" violates not-null constraint


pdpinch:
I don't understand how this happens. I thought that if the user is missing required profile data, they are bounced to the profile form before they would load their dashboard. 

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/312366357/

```
IntegrityError: null value in column "country_of_residence" violates not-null constraint
DETAIL:  Failing row contains (9814, created, 0, 0, INR, null, 2017-10-05 10:17:36.695033+00, 24091, 1, 2017-10-05 10:17:36.706705+00, 2017-10-05 10:17:36.706774+00, null, null, null).

  File "django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "newrelic/hooks/database_dbapi2.py", line 25, in execute
    *args, **kwargs)

IntegrityError: null value in column "country_of_residence" violates not-null constraint
DETAIL:  Failing row contains (9814, created, 0, 0, INR, null, 2017-10-05 10:17:36.695033+00, 24091, 1, 2017-10-05 10:17:36.706705+00, 2017-10-05 10:17:36.706774+00, null, null, null).

(25 additional frame(s) were not displayed)
...
  File "rest_framework/generics.py", line 191, in post
    return self.create(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 21, in create
    self.perform_create(serializer)
  File "rest_framework/mixins.py", line 26, in perform_create
    serializer.save()
  File "financialaid/serializers.py", line 86, in save
    country_of_residence=user.profile.country,
  File "financialaid/models.py", line 94, in save
    super().save(*args, **kwargs)

IntegrityError: null value in column "country_of_residence" violates not-null constraint
DETAIL:  Failing row contains (9814, created, 0, 0, INR, null, 2017-10-05 10:17:36.695033+00, 24091, 1, 2017-10-05 10:17:36.706705+00, 2017-10-05 10:17:36.706774+00, null, null, null).

```

pdpinch:
This also caused issues when we tried to reindex after the elasticsearch 5 upgrade

pdpinch:
https://sentry.io/mit-office-of-digital-learning/micromasters/issues/460624900/

pdpinch:
One way to address this might be to fix #2297

amir-qayyum-khan:
- open admin pages, open profiles
- open dashboard in different browser
[...]

Issue #2297, opened 2017-01-10
Guard against incomplete profiles


rhysyngsun:
We should guard against incomplete profiles causing issues. Most recently this triggered #2273 due to a null `country` in the frontend yet the user had an enrollment. Needs design around how we want to approach fixing this (backend, frontend, both?)

pdpinch:
This might be related to #3483

PR #3907, opened 2018-03-21
Redirect user to profile wizard if residence is missing

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3607

#### What's this PR do?
it fixes `api/v0/financial_aid_request/` api to move user to profile wizard if country is missing

- open a tab and type:
```python
docker-compose run web ./manage.py shell
u = User.objects.get(id=122)
u.profile.country = None
u.profile.save()
u.profile.country == None
```
- with out refreshing dashboard page , submit income

@pdpinch

### DIFF
--- a/static/js/containers/DashboardPage.js
+++ b/static/js/containers/DashboardPage.js
@@ -24,6 +24,7 @@ import {
   fetchDashboard,
   clearDashboard
 } from "../actions/dashboard"
+import { clearProfile } from "../actions/profile"
 import {
   COUPON_CONTENT_TYPE_COURSE,
   COUPON_CONTENT_TYPE_PROGRAM,
@@ -53,6 +54,7 @@ import {
   setShowExpandedCourseStatus
 } from "../actions/ui"
 import { showEnrollPayLaterSuccessMessage } from "../actions/course_enrollments"
+import { clearCalculatorEdit } from "../actions/financial_aid"
 import { findCourseRun } from "../util/util"
 import CourseListCard from "../components/dashboard/CourseListCard"
 import DashboardUserCard from "../components/dashboard/DashboardUserCard"
@@ -88,6 +90,7 @@ import {
   pearsonSSOFailure,
   setPearsonError
 } from "../actions/pearson"
+import { INCOME_DIALOG } from "./FinancialAidCalculator"
 import { processCheckout } from "./OrderSummaryPage"
 import { generateSSOForm } from "../lib/pearson"
 import { getOwnDashboard, getOwnCoursePrices } from "../reducers/util"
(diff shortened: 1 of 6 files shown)


aliceriot:
I'm thinking it might be better to error if the user's profile is incomplete at all, not just on the `country_of_residence` field. What do you think?

amir-qayyum-khan:
Done @aliceriot

amir-qayyum-khan:
Done @aliceriot , assuming these fields are compulsory https://github.com/mitodl/micromasters/issues/2297

static/js/containers/DashboardPage.js
```diff
@@ -294,6 +297,17 @@ class DashboardPage extends React.Component {
     this.fetchCoupons()
     this.handleOrderStatus()
     this.fetchDiscussionsFrontpage()
+    this.fetchUserProfile()
+  }
+
+  fetchUserProfile = () => {
```
aliceriot: we aren't actually fetching the user profile here (it's fetched by a lifecycle method in `App.js`) so I might suggest we call this instead `checkFinancialAidError` or something similar.

static/js/containers/DashboardPage.js
```diff
@@ -294,6 +297,17 @@ class DashboardPage extends React.Component {
...
+    const { financialAid: { fetchError }, dispatch } = this.props
+    if (fetchError && fetchError.message === "Profile is not complete") {
+      dispatch(clearProfile(SETTINGS.user.username))
```
aliceriot: what's the reason for clearing the profile here?

static/js/containers/DashboardPage.js
```diff
@@ -294,6 +297,17 @@ class DashboardPage extends React.Component {
...
+      this.context.router.push(`/profile/`)
+      dispatch(hideDialog(INCOME_DIALOG))
+      dispatch(clearCalculatorEdit())
```
aliceriot: it might be good to put up a toast message here (similar to what we do in `App.js`) that says something like 'hey, fill out your profile'

static/js/containers/DashboardPage.js
amir-qayyum-khan: because if i dont then it will load missing data from cache.
This is for my use case when i delete users profile data from console and do not refresh dashboard and just submit income

aliceriot: Ok I see, how likely do you think it is that a user will end up in the same position?

amir-qayyum-khan: i think that is the case, it is a bug in system which leads use to submit income with completing profile. This check is important

aliceriot: Ok, I tested it out and it seems to work correctly. FYI the most likely thing that allowed a user to end up in this kind of state (a bug in our profile redirect behavior on the frontend) was fixed in #3900, so hopefully we won't run into it again. But I think what you said makes sense, and we should keep the `clearProfile` call here

static/js/containers/DashboardPage.js
aliceriot: oh nvm the toast message in `App.js` seems to firing anyway

PR #3910, opened 2018-03-22
Updating MicromastersCertificateModel to relate to User and Course

annagav:
Related to #3870

#### What's this PR do?
Updating `MicromastersCertificateModel` part 1.
Adding `user` and `course` fields, making `final_grade` nullable.
Populating the new fields.
Changing the logic involving certificates.

To create a certificate for a user:
Run:
`alter_data set_to_passed --username "username" --course-title 'Digital Learning 200'`
Then for your `user`:
```
from courses.models import *
from grades.models import *
from grades.factories import *
course = Course.objects.filter(title='Digital Learning 200')[0]
course_run = course.courserun_set.all().first()
final_grade = FinalGradeFactory.create(user=user, course_run=course_run, passed=True, status='complete', grade=0.8)
CourseRunGradingStatus.objects.create(course_run=course_run, status='complete')
cert = MicromastersCourseCertificate.objects.create(final_grade=final_grade, user=user, course=course)
```
If you are creating second certificate make sure that you pick a different course run for that course.

If you want to see the certificate in the dashboard, also run this:
```
from cms.factories import *
CourseCertificateSignatoriesFactory.create(course=course)
```


grades/tasks.py
```diff
@@ -32,54 +31,38 @@
...
+                passed=True,
+                exam_run__date_grades_available__lte=now_in_utc(),
+                user__in=users_need_cert
+            ).values_list('user', flat=True))
+
+        for user in users_need_cert:
+            MicromastersCourseCertificate.objects.create(
+                user_id=user,
+                course=course
+            )
```
gsidebo: This updated logic doesn't account for users that have `FinalGrade`s in multiple course runs in the same course. When that's the case, the `FinalGrade` query above returns 2 identical user id's, and the for loop tries to create a `MicromastersCourseCertificate` for each of them, resulting in the same hash value and an integrity error. I tinkered with the original query in a django shell to see what could be done in a single query, but the result is probably not readable enough. I do like the idea of iterating through courses though, and I came up with the snippet below which limits the number of queries being executed (no separate query for users that have a cert, no `list()`/`set()` usage which executes the queries right away). I added `related_name='user_course_certificates'` to the `MicromastersCourseCertificate.user` definition just for readability. Let me know what you think.

```python
from django.db.models import Q, Count
#...
users_needing_cert_qset = (
  FinalGrade.objects
  .filter(
    course_run__course=course,
    status=FinalGradeStatus.COMPLETE,
    passed=True
  )
  .annotate(
    certificate_count=Count(
      'user__user_certificates',
      filter=Q(user__user_course_certificates__course=course)
    )
  )
  .filter(certificate_count=0)
  .values_list('user', flat=True)
)
if course.has_exam:
  users_needing_cert_qset = ProctoredExamGrade.objects.filter(
    user__in=users_needing_cert_qset
    #...
  )
for user in users_needing_cert_qset:
  #...
```

grades/api_test.py
```diff
@@ -449,7 +449,7 @@ def test_successful_program_certificate_generation(self):
         )
         CourseRunGradingStatus.objects.create(course_run=self.run_1, status='complete')
         with mute_signals(post_save):
-            MicromastersCourseCertificate.objects.create(final_grade=final_grade)
+            MicromastersCourseCertificate.objects.create(course=self.run_1.course, user=self.user)
```
gsidebo: I think it's probably clearer to go back to storing the `final_grade` variable and to pass the kwargs like this: `(course= final_grade.course_run.course, user=final_grade.user)`

grades/api.py
```diff
@@ -214,17 +214,22 @@ def generate_program_certificate(user, program):
...
+    courses_in_program_ids = set(program.course_set.all().values_list('id', flat=True))
+
+    courses_with_cert = set(
+        MicromastersCourseCertificate.objects.filter(
+            user=user,
+            course_id__in=courses_in_program_ids
+        ).values_list('course__id', flat=True))
+
+    if courses_in_program_ids.difference(courses_with_cert):
+        return
```
gsidebo: Might be missing something, but I think this block could just be expressed like this:

```python
num_courses_with_certs = (
  MicromastersCourseCertificate.objects.filter(
    user=user,
    course__program=program
  )
  .values_list('course', flat=True)
  .distinct().count()
)
if num_courses_with_certs < program.course_set.count():
  return
```

dashboard/api.py
```diff
@@ -503,8 +503,9 @@ def get_certificate_url(mmtrack, course):
     if best_grade:
         course_key = best_grade.course_run.edx_course_key
         if mmtrack.financial_aid_available:
-            if best_grade.has_certificate and course.signatories.exists():
-                url = reverse('certificate', args=[best_grade.certificate.hash])
+            certificate = MicromastersCourseCertificate.objects.filter(user=mmtrack.user, course=course).first()
+            if certificate and course.signatories.exists():
+                url = reverse('certificate', args=[certificate.hash])
```
gsidebo: `MMTrack` implements `get_program_certificate_url` and `program_certificate_qset`. Unless you think that's a bad approach in general, it definitely seems like we should implement `get_course_certificate_url` and `course_certificate_qset`, then call `get_course_certificate_url` from here

grades/tasks_pytest_test.py
```diff
@@ -60,27 +73,29 @@ def test_generate_course_certificates():
...
-        user=factory.Iterator([final_grade.user for final_grade in final_grades_with_failed_exam]),
+        user=factory.Iterator([final_grade.user for final_grade in passed_final_grades_with_exam[4:6]]),
         course=course_with_exams,
         passed=False,
     )
-
+    # course runs need to have CourseRunGradingStatus to get certificates
+    for final_grade in all_final_grades:
+        CourseRunGradingStatus.objects.create(course_run=final_grade.course_run, status='complete')
+    print(all_final_grades)
```
gsidebo: print

grades/tasks_pytest_test.py
```diff
@@ -60,27 +73,29 @@ def test_generate_course_certificates():
...
     tasks.generate_course_certificates_for_fa_students.delay()
 
     # Make sure that certificates were created only for passed and 'complete' status FinalGrades that either
     # had no course exam, or had a passed ProctoredExamGrade.
-    created_certificates = MicromastersCourseCertificate.objects.all()
-    assert len(created_certificates) == 6
-    certificate_grade_ids = set([certificate.final_grade.id for certificate in created_certificates])
+    certificate_user_ids = set(MicromastersCourseCertificate.objects.all().values_list('user', flat=True))
```
gsidebo: No need for `.all()`

grades/tasks_pytest_test.py
```diff
@@ -60,27 +73,29 @@ def test_generate_course_certificates():
...
     ProctoredExamGradeFactory.create_batch(
         2,
-        user=factory.Iterator([final_grade.user for final_grade in final_grades_with_failed_exam]),
+        user=factory.Iterator([final_grade.user for final_grade in passed_final_grades_with_exam[4:6]]),
         course=course_with_exams,
         passed=False,
     )
-
+    # course runs need to have CourseRunGradingStatus to get certificates
+    for final_grade in all_final_grades:
```
gsidebo: This seems to be the only usage of `all_final_grades`. I think it's probably easier and more readable to get rid of that list and instead just query `FinalGrade` here and fetch the associated `course_run`s.

grades/api.py
annagav: I think this is the same amount of db requests. But I like my version because it doesn't require the join of certificates, course, and program tables.

gsidebo: Good point. I do still think the `set` logic isn't necessary. I think this would be a bit clearer

```python
program_course_ids = program.course_set.values_list('id', flat=True)
num_courses_with_cert = (
  MicromastersCourseCertificate.objects.filter(
    user=user,
    course_id__in=program_course_ids
  )
  .values_list('course__id', flat=True)
  .distinct().count()
)
if len(program_course_ids) > num_courses_with_cert:
  return
```

grades/tasks.py
```diff
@@ -32,54 +32,41 @@
...
+        ).values_list('user', flat=True))
+
+        if course.has_exam:
+            # need also to pass exam
+            users_need_cert = list(ProctoredExamGrade.objects.filter(
+                course=course,
+                passed=True,
+                exam_run__date_grades_available__lte=now_in_utc(),
+                user__in=users_need_cert
+            ).values_list('user', flat=True))
```
gsidebo: We should get rid of both of these `list` calls. Calling `list` on a queryset runs the actual query against the db right away, and we don't need to in this case. If you remove the `list` calls, then we're only running a single query even when the course has an exam (`users_need_cert` remains as a queryset, and the `user__in=users_need_cert` clause becomes a nested query, so we save a trip to the database).

grades/api.py
annagav: If I don't use set() in the `program_course_ids`calculation the queryset will be evaluated twice, in getting `num_courses_with_cert` and later on `len(program_course_ids)`

annagav: unless `program_course_ids` qset will get cached in the first evaluation....

gsidebo: Yeah I think you're right, and even if the first query is cached, it's probably better to make it explicit. Either `list` or `set` for the first query will work then.

gsidebo:
Code looks great. Testing functionality one more time

PR #3925, opened 2018-04-03
Updated README to refer to common web app guide where appropriate

gsidebo:
No ticket. Related to https://github.com/mitodl/open-discussions/issues/41

#### What's this PR do?
- Refers to the [ODL Docker/Django/Webpack Web App Guide](https://github.com/mitodl/handbook/blob/master/common-web-app-guide.md) where appropriate
- Removes repeated instructions between this project's README and that guide
- Adds a few new bits of documentation

The only real way to test it would be to create/run MicroMasters from scratch using this guide, but I'll leave it to the reviewer(s) to decide if that's needed.


README.md
```diff
@@ -1,32 +1,26 @@
...
-  - OSX recommended install method: [Installer on Node website](https://nodejs.org/en/download/)
-  - No specific version has been chosen yet.
+# Initial setup
 
-## (OSX only) Getting your machine Docker-ready
+MicroMasters mostly follows the same [initial setup steps outlined in the common ODL web app guide](https://github.com/mitodl/handbook/blob/master/common-web-app-guide.md).
+Run through those steps in order with the following changes:
 
-#### Create your docker container:
+1. After [creating the database models](https://github.com/mitodl/handbook/blob/master/common-web-app-guide.md#3-create-data-structures), you should create the Elasticsearch indices via management command:
```
aliceriot: I don't see a `3-create-data-structures` section in the common guide, did you mean to link to https://github.com/mitodl/handbook/blob/master/common-web-app-guide.md#3-create-database-tables-from-the-django-models ?

README.md
```diff
@@ -1,32 +1,26 @@
...
-    eval "$(docker-machine env)"
+1. **Do not run the `createsuperuser` command.** When you log into this app via edX, a MicroMasters 
+Django user is created for you automatically. See the section below for instructions on how to run edX locally.
 
-## Running edX devstack locally _(optional, but recommended)_
 
-MicroMasters can work with a live instance of edX, but it's recommended that
+### Running edX devstack locally _(optional, but recommended)_
+
+MicroMasters can work with a deployed instance of edX, but it's recommended that
```
aliceriot: I think it would be good here to include information about how to set up MM locally with the edX sandbox, as well as how to run edX locally. I've always used the sandbox and not run into any problems, so for people who aren't working on edX stuff as much it will most likely be an easier option than getting edX running locally.

aliceriot: (although if we don't want to add that now that is fine :smile: )

README.md
```diff
@@ -321,26 +262,49 @@ If a test errors but selenium is still working, it will take
...
     ./scripts/test/run_snapshot_dashboard_states.sh
 
-##### Viewing tests via browser
+#### Validating financial aid info
+
+To validate prices and financial aid discounts for all programs run:
+
+    docker-compose run web ./manage.py validate_db
+    
+Screenshots are saved to the `output` directory.
```
aliceriot: is this line supposed to be part of the section above?

README.md
gsidebo: Right you are. I updated that section title and didn't update the link

README.md
gsidebo: Right you are!

README.md
gsidebo: Yeah I'd say that's a separate issue. There weren't any details about using the sandbox before this PR. I've never actually used the sandbox. If you have a few minutes for a quick write-up and you think it would be easier, I'd be happy to review and include in this PR

aliceriot: eh, it's not so essential to add right now :)

Issue #3406, opened 2017-07-14
missed-deadline state not handled properly in dashboard


noisecapella:
You can reproduce this via `./scripts/test/run_snapshot_dashboard_states.sh --match 008`. The status is `missed-deadline` but the message indicates you can still pay to upgrade:
![screenshot from 2017-07-14 17-44-10](https://user-images.githubusercontent.com/863262/28231965-34f1df6e-68bc-11e7-8137-0ad1b11f1e07.png)

noisecapella:
Looking at the source I think this is only hit if the `course_end_date` is in the future but `upgrade_deadline` is in the past

amir-qayyum-khan:
@noisecapella now 008 is `can-upgrade`. File is dashboard_state_008_set_to_offered.dashboard.json
```json
                    "runs": [
                        {
                            "id": 1266,
                            "course_id": "course-v1:MITx+Analog+Learning+100+Apr_2018",
                            "title": "Analog Learning 100 - April 2018",
                            "status": "can-upgrade",
                            "has_paid": false,
                            "position": 1,
                            "course_start_date": "2018-04-08T12:19:24.537247Z",
                            "course_end_date": "2018-05-08T12:19:24.537247Z",
                            "fuzzy_start_date": null,
                            "enrollment_url": null,
                            "course_upgrade_deadline": "2018-04-23T12:19:24.537247Z",
                            "current_grade": null
                        }
                    ],
```

PR #3937, opened 2018-04-13
Fixed status message of current/future course when status is missed upgrade deadline

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3406

#### What's this PR do?
fixes a special case when there is only one future/current run in the course and status is`missed-deadline`. This PR restrict user to pay and resume course in this state.

- run ./scripts/test/run_snapshot_dashboard_states.sh --match 009
- create a course with one run and make start, due, enrollment dates null and set upgrade deadline to past date and see dashboard there should be no payment button.

@pdpinch 
dashboard_state_009_set_to_offered

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages_test.js
+++ b/static/js/components/dashboard/courses/StatusMessages_test.js
@@ -550,11 +550,27 @@ describe("Course Status Messages", () => {
       )
     })
 
-    it("should nag about missing the payment deadline for future course", () => {
+    it("should nag about missing the payment deadline for future course with one run", () => {
       course.runs = [course.runs[0]]
       course.runs[0].course_start_date = ""
       course.runs[0].course_end_date = ""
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
Done @annagav

static/js/components/dashboard/courses/StatusMessages_test.js
```diff
@@ -565,6 +565,22 @@ describe("Course Status Messages", () => {
       ])
     })
 
+    it("should nag about missing the payment deadline for current course with one run", () => {
+      course.runs = [course.runs[0]]
+      course.runs[0].course_start_date = moment()
```
annagav: You could use `makeRunCurrent` here

amir-qayyum-khan:
Done @annagav

PR #3941, opened 2018-04-17
display courserun dates in admin list view; make them editable

### LINKED ISSUE #3940: make course run dates editable in admin list view
As an admin, it would be convenient if I could view and edit course start and end dates in a grid instead of having to view and edit each course run individually.

pdpinch:

fixes #3940 

#### What's this PR do?

makes the admin list view for courseruns a little more convenient to use:
- displays all course run dates
- makes the date editable in list view
- adds program and course filters to the courserun list

View /admin/courses/courserun/

You may also want to set the course_number of a few courses (seed_db doesn't do this)

We're currently scheduling course runs 2-3 years into the future, with 2-3 runs per year. Since we (still) don't have a reliable source for this data, I have to enter it by hand.

### DIFF
--- a/courses/admin.py
+++ b/courses/admin.py
@@ -42,11 +42,11 @@ class CourseAdmin(admin.ModelAdmin):
 
 class CourseRunAdmin(admin.ModelAdmin):
     """ModelAdmin for Courses"""
-    list_display = ('title', 'course_number', 'edx_course_key', 'enrollment_start','start_date','enrollment_end',
-        'end_date','upgrade_deadline','freeze_grade_date', )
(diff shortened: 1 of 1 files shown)


adorsk:
@pdpinch I think this will fix your travis woes:
1. Change https://github.com/mitodl/micromasters/blob/master/test_requirements.txt#L13 to `-e git+/https://...`

adorsk:
@pdpinch after https://github.com/mitodl/micromasters/pull/3944 merges, just merge master and this should work.

amir-qayyum-khan:
screen shot 2018-04-23 at 9 12 25 pm

PR #3950, opened 2018-04-19
If no courses show I'm insterested button

### LINKED ISSUE #3949: Marketing Page template
Sometimes we need to market a program that doesn't yet have courses available on edX. In that case, there's no purpose to having people sign up / log in with edX. 

We can make a CMS template similar to the current program_page but instead of the sign_up button it has a call to sign up for the mailing list, similar to the mailchimp form on the home page.

annagav:
Fix #3949

#### What's this PR do?
If program does not have courses then show a different button to subscribe for mailing list.

screen shot 2018-04-19 at 4 37 54 pm

### DIFF
--- a/ui/views_test.py
+++ b/ui/views_test.py
@@ -650,6 +650,7 @@ class TestProgramPage(ViewsTests):
 
     def test_login_button(self):
         """Verify that we see a login button"""
+        CourseFactory.create(program=self.program_page.program)
         response = self.client.get(self.program_page.url)
         self.assertContains(response, "Sign Up Now")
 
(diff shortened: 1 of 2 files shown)


noisecapella:
Can you add a test that "I'm interested" is on the page if there are no courses? You can copy the `test_course_listing` test in `ui/views_test.py` and modify it slightly for this purpose

noisecapella:
For the failing test I think you need to add a course to the program to get the right message

ui/views_test.py
```diff
@@ -709,6 +710,18 @@ def test_course_listing(self):
...
+    def test_no_courses(self):
+        """
+        Verify that no courses result in a different button
+        """
+        response = self.client.get(self.program_page.url)
+        js_settings = json.loads(response.context['js_settings_json'])
+        self.assertIn("program", js_settings)
+        self.assertIn("courses", js_settings["program"])
+        self.assertEqual(len(js_settings["program"]["courses"]), 0)
+        print(response)
```
noisecapella: Can you remove this print statement?

Issue #3927, opened 2018-04-04
503 error from API does not result in an error message being shown


noisecapella:
![screenshot from 2018-04-04 13-21-33](https://user-images.githubusercontent.com/863262/38323401-221d4fee-380b-11e8-9337-1bcc96728c4c.png)

We should show an error message instead when the API call fails to complete

PR #3957, opened 2018-04-23
Handle 503 error on dashboard api

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3927

#### What's this PR do?
- When there is 503 error on dashboard api fetchJSONWithCSRF expects json but we have text/html error message. 
- And in this case of error we are returning success promise instead of rejection
- because of that dashboard neither have content nor the error info
- I fixed this by returning `Promise.reject('Unable to process')

- go to nginx.conf and put these lines to create 503 error
```
    location /api/v0/dashboard/ {
        include uwsgi_params;
        if ($request_method = GET ) {
            return 503;
        }
    }
```
- And on dashboard you should see error message
- you may need to restart docker

@pdpinch 

screen shot 2018-04-23 at 8 17 07 pm


static/js/lib/api.js
```diff
@@ -55,13 +50,12 @@ export function patchUserProfile(
...
 
-export async function getDashboard(username: string): Promise<Dashboard> {
+export async function getDashboard(username: string): Promise<Response> {
   try {
     const response = await fetchJSONWithCSRF(`/api/v0/dashboard/${username}/`)
     return response
   } catch (response) {
-    loginOnEdXError(response)
-    return response
+    return loginOnEdXError(response)
```
aliceriot: I think with this function (and the `getPrograms` function below) it might be better to return `Promise.reject(response)` directly, so that we will see a Sentry error which has information about what went wrong (instead of just the generic 'Unable to process'). What do you think?

static/js/lib/api_test.js
```diff
@@ -169,6 +169,16 @@ describe("api", function() {
           })
         })
       }
+
+      it("dashboard api returns error if status = 503", () => {
```
aliceriot: it's really if the error is any code other than 400 or 401, right? might be good to test a few different error codes here, just for completeness' sake.

amir-qayyum-khan:
Done @aliceriot

amir-qayyum-khan:
Done @aliceriot

PR #3977, opened 2018-05-09
Check freeze status show correct enrollment numbers

annagav:
None
There are users that un-enrolled from course run and have CachedEnrollment, but do not have CachedCurrentGrade, we do not want to count them in the process of freezing grades.

#### What's this PR do?
In the status massage reports the number of enrollments intersected with current grades.

Run this command for a frozen run:
`python manage.py check_final_grade_freeze_status <edx_course_key>`


annagav:
@noisecapella, should be fixed now. Thanks!

Issue #3968, opened 2018-04-30
Stop trying to freeze a user's grade when they are no longer enrolled


pdpinch:
Conversation in Slack: https://mitodl.slack.com/archives/G1VK0EDGA/p1525102712000368

Sentry error: InvalidCredentialStored: Received a 401 status code from the OAUTH server

https://sentry.io/mit-office-of-digital-learning/micromasters/issues/256030207/

```
HTTPError: 401 Client Error: Unauthorized for url: https://courses.edx.org/oauth2/access_token/
(1 additional frame(s) were not displayed)
...
  File "social_core/storage.py", line 55, in refresh_token
    response = backend.refresh_token(token, *args, **kwargs)
  File "backends/edxorg.py", line 106, in refresh_token
    response = super(EdxOrgOAuth2, self).refresh_token(token, *args, **kwargs)
  File "social_core/backends/oauth.py", line 431, in refresh_token
    request = self.request(url, **request_args)
  File "social_core/backends/base.py", line 229, in request
    response.raise_for_status()
  File "requests/models.py", line 862, in raise_for_status
    raise HTTPError(http_error_msg, response=self)

InvalidCredentialStored: Received a 401 status code from the OAUTH server
  File "grades/api.py", line 170, in freeze_user_final_grade
    CachedEdxDataApi.update_all_cached_grade_data(user)
  File "dashboard/api_edx_cache.py", line 345, in update_all_cached_grade_data
    utils.refresh_user_token(user_social)
  File "backends/utils.py", line 47, in refresh_user_token
    _send_refresh_request(user_social)
  File "backends/utils.py", line 26, in _send_refresh_request
    http_status_code=exc.response.status_code

Impossible to refresh the edX cache for user "gangakaushik"
```

annagav:
The first time task ran it tried freezing grades for enrollments that were not properly/timely  updated and got this [error](https://sentry.io/mit-office-of-digital-learning/micromasters/issues/416616973/), along with InvalidCredentialStored errors. When the freeze grades task tries to fetch the current grade for users that unenrolled from the course on edx, it throws this error `AttributeErrorgrades.tasks.freeze_users_final_grade_async
error'NoneType' object has no attribute 'percent` and fails to save the results of running the task. So the next time the task is run, the results is null, it thinks it is freezing the grades first time, and therefore tries to update cache for all users missing final grade again.

PR #3980, opened 2018-05-14
Add users missing grades to the cached list

annagav:
Related to #3968

#### What's this PR do?
2 things:
Add info about what course was being processed when 'InvalidCredentialStored' is thrown.
Add users that don't have current grade to the cached users list, then the next time the task is run it will be able to set the status to COMPLETE, if no other errors occur.

Create a CachedEnrollment but no CachedCurrentGrade for a user and try to freeze grades for that course. The second time you run it should say `Final grades for course <edx_course_key> are complete`.

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -172,7 +172,11 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
         con = get_redis_connection("redis")
         con.lpush(CACHE_KEY_FAILED_USERS_BASE_STR.format(course_run.edx_course_key), user.id)
         if not raise_on_exception:
-            log.exception('Impossible to refresh the edX cache for user "%s"', user.username)
+            log.exception(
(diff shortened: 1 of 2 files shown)


grades/api.py
```diff
@@ -172,7 +172,11 @@ def freeze_user_final_grade(user, course_run, raise_on_exception=False):
...
         if not raise_on_exception:
-            log.exception('Impossible to refresh the edX cache for user "%s"', user.username)
+            log.exception(
+                'Impossible to refresh the edX cache for user "%s" in course %s',
+                user.username,
+                course_run.edx_course_key
+            )
             return None
         else:
             raise FreezeGradeFailedException(
```
noisecapella: Can you update the exception message to include the course key like above?

annagav:
@noisecapella, I addressed your comment.

PR #4000, opened 2018-06-01
Fixed regression on semester facet front end side

amir-qayyum-khan:
N/A

#### What's this PR do?
- It fixes a regression where semester facet is not disabling when courses are select and vise versa

- Go to learner search as staff
- Select one or all semesters then see course faces must be disable 
- Select a course then see semester facet must be disbale

@pdpinch @annagav 

screen shot 2018-06-01 at 6 29 25 pm

screen shot 2018-06-01 at 6 29 32 pm


annagav:
Works for me 👍

PR #4010, opened 2018-06-07
Use flag to determine if discussions user should be updated

mbertrand:
https://github.com/mitodl/open-discussions/issues/783 (3rd checkbox)

#### What's this PR do?
If `FEATURE_OPEN_DISCUSSIONS_USER_SYNC` is `True` or `None`, update the discussions profile when the Micromasters profile is updated.  If set to `False`, do not update.

- Set `FEATURE_OPEN_DISCUSSIONS_USER_SYNC=False` in your `.env` file
- Log in as a new user in Micromasters, fill in the required blanks, then go to 'View Discussions' link from the dashboard.
- In a discussions shell:
  ```python
  from profiles.models import Profile
  Profile.objects.first().name
  ```
  - The full name should be the same as the first + last name in Micromasters
- In Micromasters, edit your profile and change your first or last name.
- In a discussions shell:
  ```python
  Profile.objects.first().name
  ```
  - The profile name should be unchanged (original first + last name).
- Set `FEATURE_OPEN_DISCUSSIONS_USER_SYNC=True` in your `.env` file, or remove it altogether.
- Restart Micromasters
- Change your micromasters profile first or last name
- In a discussions shell:
  ```python
  Profile.objects.first().name
  ```
  - The profile name should be changed to match the micromasters profile.

### DIFF
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -78,8 +78,8 @@ def create_or_update_discussion_user(user_id, allow_email_optin=False):
         if discussion_user.username is None:
             create_discussion_user(discussion_user)
         else:
-            update_discussion_user(discussion_user, allow_email_optin=allow_email_optin)
-
+            if settings.FEATURES.get('OPEN_DISCUSSIONS_USER_SYNC', True):
+                update_discussion_user(discussion_user, allow_email_optin=allow_email_optin)
         return discussion_user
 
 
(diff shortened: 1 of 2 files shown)


pdpinch:
@rhysyngsun is this what you had in mind?

rhysyngsun:
looks good to me

PR #4048, opened 2018-06-29
Fix semester user count bug

### LINKED ISSUE #4009: Semester filter count doesn't match the actual results number
screen shot 2018-06-06 at 7 07 31 pm

annagav:
Fix #4009

#### What's this PR do?
make RefinementListFilter use reverse nested aggregation to count users for semester enrollment.

Make a user enrolled in two different course runs in the same semester.

screen shot 2018-06-29 at 1 04 58 pm


amir-qayyum-khan:
it works for user who are enroll in program but not in course?

Issue #3118, opened 2017-04-27
Fix Modal Titles on Profile


roberthouse54:
The modal titles for Education and Employment are quite clunky, the way they are currently implemented.

- All of the modals currently have two effective titles. (See below) For example, "Education" and then under that "Edit Education". There should only be a single title for each modal. 
![image](https://cloud.githubusercontent.com/assets/20047260/25495236/72b9ae76-2b4a-11e7-8fbb-920d4c6468ab.png)
- The Add/Edit Education modals also should describe which type of degree is being edited or added. For example, after you have clicked Add degree in the High School card, there is no indication in the modal that you are adding a High School degree. If the user forgets what link they clicked, they might enter the wrong information.

Solution:
- Each modal will have a single title.
- The title will include the type of degree being added/edited.

Ex.
Add a High School Degree
Edit bachelor's degree

On the user profile, if they click "Add degree" and no degree type is specified, the title will be "Add a degree".

For employment, the title will always be either "Add Employment" or "Edit Employment"

pdpinch:
@roberthouse54 does this still need to be fixed?

roberthouse54:
Yes it does.

PR #4059, opened 2018-07-09
Fixed education and employment dialog titles

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3118

#### What's this PR do?
This PR fixes makes:
- Each modal with a single title.
- The title will include the type of degree being added/edited. 

Go to profile page and add/edit degree and employment

@pdpinch 
- screen shot 2018-07-09 at 4 41 07 pm
- screen shot 2018-07-09 at 4 41 16 pm
- screen shot 2018-07-09 at 4 41 23 pm
- screen shot 2018-07-09 at 4 41 30 pm

### DIFF
--- a/static/js/components/EducationForm.js
+++ b/static/js/components/EducationForm.js
@@ -447,12 +447,8 @@ class EducationForm extends ProfileFormFields {
     const inFlight = profilePatchStatus === FETCH_PROCESSING
     const keySet = (key): any => ["education", educationDialogIndex, key]
     const id = _.get(profile, keySet("id"))
-    const degreeName = _.get(
-      R.find(R.propEq("value", _.get(profile, keySet("degree_name"))))(
-        EDUCATION_LEVELS
-      ),
-      "label"
-    )
+    const degreeName =
+      EDUCATION_LEVEL_LABELS[_.get(profile, keySet("degree_name"))]
     const title =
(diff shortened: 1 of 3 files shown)


static/js/components/EducationForm.js
```diff
@@ -440,11 +435,26 @@ class EducationForm extends ProfileFormFields {
...
 
     const inFlight = profilePatchStatus === FETCH_PROCESSING
+    const keySet = (key): any => ["education", educationDialogIndex, key]
+    const id = _.get(profile, keySet("id"))
+    const degreeName = _.get(
+      R.find(R.propEq("value", _.get(profile, keySet("degree_name"))))(
+        EDUCATION_LEVELS
+      ),
+      "label"
+    )
```
aliceriot: This is a bit of a convoluted way to get this out here - why don't we instead declare an `EDUCATION_LEVEL_LABELS` object in the constants file, so that here you can just do `EDUCATION_LEVEL_LABELS[_.get(profile, keySet("degree_name"))]`?

I think it could be something like

```js
const EDUCATION_LEVEL_LABELS = Object.assign(
  {},
  EDUCATION_LEVELS.map(({value, label}) => [value, label])
)
```

you could also use `R.fromPairs` instead of `Object.assign`

amir-qayyum-khan:
Done @aliceriot

PR #4062, opened 2018-07-16
Offer to pay again for exam when already passed

### LINKED ISSUE #4025: Exams: allow learner to pay for 3rd exam
As a learner, I'd like to take the proctored exam a third time, beyond the two attempts that are offered with one payment. 

Currently, we restrict learners to two attempts per payment. IIRC, this is implemented at the exam authorization stage, so we don't authorize learners who have had two grades. However, we don't seem to allow learners to pay again via the dashboard. 

See ZenDesk https://odl.zendesk.com/agent/tickets/17987

annagav:
Fix #4025

#### What's this PR do?
Show message about exams even when user has passed exam.

(Required)
Run `./scripts/test/run_snapshot_dashboard_states.sh  --match exams`

screen shot 2018-07-16 at 1 38 48 pm
screen shot 2018-07-16 at 1 52 15 pm
screen shot 2018-07-16 at 1 52 44 pm

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages.js
+++ b/static/js/components/dashboard/courses/StatusMessages.js
@@ -88,6 +88,27 @@ const messageForNotAttemptedExam = (course: Course) => {
   return message
 }
 
+const messageForAttemptedExams = (course: Course, passedExam: boolean) => {
+    const whenFailed = passedExam ? "" : "You did not pass the exam. "
+    const payAgain = course.has_to_pay ? "If you want to re-take the exam, you need to pay again." : ""
+    let whenToSchedule = ""
+    if (course.can_schedule_exam) {
+      // if can schedule exam now
+      if (!course.has_to_pay){
+        whenToSchedule = "Click above to reschedule an exam with Pearson."
+      }
+    } else if (R.isEmpty(course.exams_schedulable_in_future)) {
+      // no info about future exam runs
+       whenToSchedule = `There are currently no exams ` +
+       "available for scheduling. Please check back later."
+    } else {
+      // can not schedule now, but some time in the future
+      whenToSchedule = "You can sign up to re-take the exam starting " +
+        `on ${formatDate(course.exams_schedulable_in_future[0])}`
+    }
+    return { message : `${whenFailed}${payAgain}${whenToSchedule}`}
+  }
+
 const courseStartMessage = (run: CourseRun) => {
   const startDate = notNilorEmpty(run.course_start_date)
     ? formatDate(run.course_start_date)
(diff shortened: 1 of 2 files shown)


static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -88,6 +88,31 @@ const messageForNotAttemptedExam = (course: Course) => {
   return message
 }
 
+const messageForAttemptedExams = (course: Course, passedExam: boolean) => {
+  const whenFailed = passedExam ? "" : "You did not pass the exam. "
+  const payAgain = course.has_to_pay
+    ? "If you want to re-take the exam, you need to pay again. "
+    : ""
+  let whenToSchedule = ""
+  if (course.can_schedule_exam) {
```
amir-qayyum-khan: why not

```
if (course.can_schedule_exam && !course.has_to_pay) {
  whenToSchedule = "Click above to reschedule an exam with Pearson."
}
```

static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -88,6 +88,31 @@ const messageForNotAttemptedExam = (course: Course) => {
...
+    // if can schedule exam now
+    if (!course.has_to_pay) {
+      whenToSchedule = "Click above to reschedule an exam with Pearson."
+    }
+  } else if (R.isEmpty(course.exams_schedulable_in_future)) {
```
amir-qayyum-khan: use loadashhttps://lodash.com/docs/4.17.10#isEmpty because it tasked null and '' as empty.

static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -88,6 +88,31 @@ const messageForNotAttemptedExam = (course: Course) => {
   return message
 }
 
+const messageForAttemptedExams = (course: Course, passedExam: boolean) => {
+  const whenFailed = passedExam ? "" : "You did not pass the exam. "
```
amir-qayyum-khan: why not handle passed exam message here

static/js/components/dashboard/courses/StatusMessages.js
annagav: I think `exams_schedulable_in_future` should always be "" when no exams, but i'll change this just in case.

static/js/components/dashboard/courses/StatusMessages.js
annagav: what do you mean?

amir-qayyum-khan: just a question when user passed course we are showing "You passed this course.". 
Just saw in you PR description image. I was saying can we add/handle that message here?

static/js/components/dashboard/courses/StatusMessages.js
annagav: The rest of the `if else` statement should happen only based on `course.can_schedule_exam`

amir-qayyum-khan:
some tests are failing

static/js/components/dashboard/courses/StatusMessages.js
annagav: Sorry, I confused this with can_schedule_exam. This exams_schedulable_in_future is a list so what is wrong with using R.isEmpty?

amir-qayyum-khan: R.isEmpty is ok for list

PR #4076, opened 2018-07-26
Check if email is verified before creating account

### LINKED ISSUE #4072: Prevent edX users with is_active = False from creating accounts
At a minimum this is a call to `/api/users/v0/me` before `social_core.pipeline.user.create_user` is called to and checking `is_active` to abort the pipeline. We probably want to redirect them to a nice error page informing them what is required to proceed.

NOTE: We currently have an existing call [here](https://github.com/mitodl/micromasters/blob/328cce59ae5a9ff28850a9fff00f309bf187017d/backends/pipeline_api.py#L70-L75) to this API. We also have a call to the user_info API in the edx backend, when returns a subset of the users API. We should consolidate this to one upfront call.

  - [ ] Update MM to use new client library
  - [ ] Add `edx_is_active` field to `Profile` in MM to track if the email is verified on edX
  - [ ] Populate edx_is_active on login (checking with client library)

annagav:
Fix #4072

#### What's this PR do?
If user has not verified email on edX redirect to error page.

Try to register on MM without verifying your email on edX.

### DIFF
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -90,7 +90,7 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
     )
 
 
-def check_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
+def check_edx_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
     """Get account information to check if email was verified for account on edX"""
     if backend.name != EdxOrgOAuth2.name:
         return {}
@@ -99,7 +99,7 @@ def check_verified_email(backend, response, details, *args, **kwargs):  # pylint
     access_token = response.get('access_token')
     if not access_token:
         # this should never happen for the edx oauth provider, but just in case...
-        raise AuthException('Missing access token for the user {0}'.format(username))
+        raise AuthException('Missing access token for the edX user {0}'.format(username))
 
     user_profile_edx = backend.get_json(
         urljoin(backend.EDXORG_BASE_URL, '/api/user/v1/accounts/{0}'.format(username)),
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -93,7 +93,7 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
 def check_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
     """Get account information to check if email was verified for account on edX"""
     if backend.name != EdxOrgOAuth2.name:
-        return
+        return {}
 
     username = details.get('username')
     access_token = response.get('access_token')
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -6,10 +6,10 @@ from urllib.parse import urljoin
 
 from django.shortcuts import redirect
 from rolepermissions.checkers import has_role
+from social_core.exceptions import AuthException
 
 from backends.edxorg import EdxOrgOAuth2
 from micromasters.utils import now_in_utc
-from profiles.api import get_social_username
 from profiles.models import Profile
 from profiles.util import split_name
 from roles.models import (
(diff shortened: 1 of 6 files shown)


backends/pipeline_api.py
```diff
@@ -102,6 +103,26 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
     )
 
 
+def check_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
+    """Get account information to check if email was verified for account on edX"""
```
rhysyngsun: It's a PSA best practice to check `backend.name` if you're doing back-specific logic, so we should verify this is an edx backend.

backends/pipeline_api.py
```diff
@@ -102,6 +103,26 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
...
+    username = details.get('username')
+    access_token = response.get('access_token')
+    if not access_token:
+        # this should never happen for the edx oauth provider, but just in case...
+        log.error('Missing access token for the user %s', username)
+        return
```
rhysyngsun: I would have this raise an `AuthException`, since we would want to know this and have the flow abort too.

annagav:
@rhysyngsun, I addressed all your comments.

backends/pipeline_api.py
```diff
@@ -102,6 +90,30 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
...
+def check_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
+    """Get account information to check if email was verified for account on edX"""
+    if backend.name != EdxOrgOAuth2.name:
+        return
+
+    username = details.get('username')
+    access_token = response.get('access_token')
+    if not access_token:
+        # this should never happen for the edx oauth provider, but just in case...
+        raise AuthException('Missing access token for the user {0}'.format(username))
```
rhysyngsun: Suggestion: say "edx user" so we know it's the edx username

backends/pipeline_api.py
```diff
@@ -102,6 +90,30 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
     )
 
 
+def check_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
+    """Get account information to check if email was verified for account on edX"""
+    if backend.name != EdxOrgOAuth2.name:
+        return
```
rhysyngsun: Suggest you return `{}` here, I've found it's best to be consistent in always returning some kind of dict out of PSA actions in case you change it to a partial later.

backends/pipeline_api.py
```diff
@@ -102,6 +90,30 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
     )
 
 
+def check_verified_email(backend, response, details, *args, **kwargs):  # pylint: disable=unused-argument
```
rhysyngsun: Suggestion: rename to `check_edx_verified_email`

ui/templates/verify_email.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block title %}{% trans "MITx MicroMasters" %}{% endblock %}
+
+{% block error_header %}
+NEED TO VERIFY EMAIL
+{% endblock %}
+
+{% block error_content %}
+<div id="page-not-found">
+	<p>You need to verify your email before you can proceed.</p>
```
rhysyngsun: @pdpinch is this language ok?

backends/pipeline_api.py
rhysyngsun: Could you also add some tests around this new action?

backends/pipeline_api.py
rhysyngsun: Looks like `pylint` caught this anyway.

annagav:
screen shot 2018-07-31 at 11 48 59 am

annagav:
@rhysyngsun, comments addressed!

annagav:
@roberthouse54  suggested to change this to "You need to verify your email with edX before you can proceed."

Issue #3985, opened 2018-05-18
Missing exam authorizations


pdpinch:
A number of learners have written in to complain about missing exam authorizations when they have passed and paid for 14.73x. 

I've been able to deal with this on an ad hoc basis by updating their final grade record in order to force a signal to fire. 

I suspect the issue may be related to the fact that all the learners affected by this passed the first run of 14.73x, for which edX no longer returns enrollment information. I wonder if our batch process for creating authorizations has a dependency on enrollment status instead when a passing final grade should be sufficient.

rhysyngsun:
@pdpinch relevant code is here, it walks `CachedEnrollment`: https://github.com/mitodl/micromasters/blob/master/exams/api.py#L132-L150

pdpinch:
Do you think we could just change that to walk final grades instead?

pdpinch:
that would just be a matter of running before and after queries, right? 

the problem ticket is at https://odl.zendesk.com/agent/tickets/17069 which we can use to identify individual users who have a final grade but lack an enrollment.

pdpinch:
@amir-qayyum-khan we'll want to be careful with this. As mentioned above, I'll need to do some tests with production data before we merge it.

amir-qayyum-khan:
sure, can you suggest me how to test this with prod data?

PR #4083, opened 2018-07-31
Added exam authorizations on the base of final grade

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3985

#### What's this PR do?
- It picks latest passed final grade to authorize user on course for exam

- @pdpinch you can test it on prod data


exams/api.py
```diff
@@ -130,28 +129,30 @@ def authorize_for_exam_run(mmtrack, course_run, exam_run):
...
     """
-    This walks the CachedEnrollments backwards chronologically and authorizes the first eligible one.
+    This walks the FinalGrade backwards chronologically and authorizes the first eligible one.
 
     Args:
         mmtrack (dashboard.utils.MMTrack): An instance of all user information about a program
         exam_run (exams.models.ExamRun): the ExamRun to authorize the learner for
     """
-    enrollments_qset = CachedEnrollment.objects.filter(
+    final_grades = FinalGrade.objects.filter(
```
amir-qayyum-khan: @pdpinch @rhysyngsun do we need to check user has paid using `mmtrack` here?

amir-qayyum-khan:
blocked because @pdpinch need to test it first on prod.

exams/api.py
```diff
@@ -130,28 +129,30 @@ def authorize_for_exam_run(mmtrack, course_run, exam_run):
...
         user=mmtrack.user,
+        passed=True,
```
pdpinch: Why did you add `passed=True` to the filter?

amir-qayyum-khan: Dont we need only authorize for passed run? https://github.com/mitodl/micromasters/blob/master/exams/api.py#L104

I am filtering passed user above.

rhysyngsun:
@pdpinch can you run it against a copy of production data and check those results to be sure you're seeing the correct delta (i.e. gaining exam auths, but not losing any)?

pdpinch:
I rebased this with master to pick up the setting fix from Monday. Hopefully Heroku can build a review app now.

pdpinch:
Ok, I'm satisfied. Authorizations for 14.73x went up from 841 to 869 after this code change. Also, with the new code, the learner in [ZenDesk #19046](https://odl.zendesk.com/agent/tickets/19046) also got the expected authorization.

amir-qayyum-khan:
so @pdpinch is this ready for merge?

Issue #4056, opened 2018-07-06
Allow learner to unenroll from program


pdpinch:
As a a learner, I'd like to be able exit a program -- to stop getting email updates and to stop seeing it on my dashboard. 

Now that we're adding more programs, we're finding that some learners want to exit the program they are current enrolled in and register for another one. 

### questions
- is edX doing anything to track learner enrollment in programs now? 

### related issues
#2018 when manually deleting/retiring a user, we typically delete all their program enrollments

roberthouse54:
1. Add a link to the settings bage to "Leave a MicroMasters Program"

![1](https://user-images.githubusercontent.com/20047260/43271041-83bbe212-90c4-11e8-9944-7f1c954e12e7.png)

2. Open a modal, that allows the user to select which program they want to leave. This would show all programs the user is enrolled in. We can also have a "All Programs" option. 
![2](https://user-images.githubusercontent.com/20047260/43271123-b6b9d26e-90c4-11e8-8cae-b3d669a66c48.png)

3. On submission, show a "success" toast message. 
![3](https://user-images.githubusercontent.com/20047260/43271153-c941e3c2-90c4-11e8-8f5c-62d3ca07c526.png)

4. If the user has unenrolled from all MicroMasters programs, then their dashboard will show an empty state, with a link allowing them to enroll in a program.
![4](https://user-images.githubusercontent.com/20047260/43271191-e7516252-90c4-11e8-9051-52fd8a7488fe.png)

amir-qayyum-khan:
@pdpinch what i will Do

- remove user from model http://192.168.99.100:8079/admin/dashboard/programenrollment/
- Do i need to remove him from discussion as moderator if he is staff?
-  Show no program UI if he had single program. From there user can enroll.

pdpinch:
Could you be more specific about what you mean by "remove user from model?"

I don't think it's necessary to remove the user's staff role, unless it's going to cause exceptions. They should be removed from any discussions channels automatically, because they will no longer have the necessary program enrollment. 

How will you populate the drop-down list of programs the user can leave?

amir-qayyum-khan:
The model i was referring was program enrollment model.

i will load programs from dashboard api,
new api can also. be made.

pdpinch:
Is the dashboard API already being loaded on this account settings page?

The dashboard API is quite slow. It's fine if we're already loading it, but we should avoid it if we can. I think there's a programs API that might be do what we need?

[...]

PR #4084, opened 2018-08-01
Added unenroll program(s) feature

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/4056

#### What's this PR do?
It adds a wizard to unenroll from one or many programs. And update UI on unenroll.

- [x] Added api
- [x] Add unenroll card on settings
- [x] Add unenroll programs wizard
  - [x] display enrolled programs
  - [x] Add multi select programs widget
  - [x] api call
  - [x] show spinner on api call
- [x] dashboard card when no program selected
- [x] Unit tests for front end
- [x] Unit tests for backend
 
- Enroll programs one or many
- Go to settings and click unenroll link
- then select program(s)
- check dashboard
- when no program left i.e you unenroll all, check new card on dashboard with enroll link

@pdpinch 
-  screen shot 2018-08-03 at 5 25 22 pm
- screen shot 2018-08-02 at 3 07 14 pm
-  screen shot 2018-08-02 at 3 02 58 pm
-  screen shot 2018-08-02 at 3 08 47 pm
- screen shot 2018-08-03 at 4 30 50 pm

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -20,6 +20,8 @@ from rest_framework.generics import get_object_or_404
 from backends import utils
 from backends.edxorg import EdxOrgOAuth2
 from dashboard.permissions import CanReadIfStaffOrSelf
+from dashboard.serializers import UnEnrollProgramsSerializer
+from dashboard.models import ProgramEnrollment
 from dashboard.api import get_user_program_info
 from dashboard.api_edx_cache import CachedEdxDataApi
 from micromasters.exceptions import PossiblyImproperlyConfigured
@@ -149,4 +151,39 @@ class UserCourseEnrollment(APIView):
         CachedEdxDataApi.update_cached_enrollment(request.user, enrollment, enrollment.course_id, index_user=True)
         return Response(
             data=enrollment.json
+        )
+
+
+class UnEnrollPrograms(APIView):
+    """
+    api that unenroll user from one or more programs
+    """
+    authentication_classes = (
+        authentication.SessionAuthentication,
+        authentication.TokenAuthentication,
+    )
+    permission_classes = (permissions.IsAuthenticated, )
+
+    def post(self, request):
+        """
+        unenroll from MM program(s)
+        """
+        response = []
+        program_ids = UnEnrollProgramsSerializer(data=request.data).get_program_ids()
+
+        program_enrollments = ProgramEnrollment.objects.filter(
+            program_id__in=program_ids,
+            user=request.user
+        )
+
+        for program_enrollment in program_enrollments:
+            response.append({
+                'program_id': program_enrollment.program_id,
+                'title': program_enrollment.program.title
+            })
+            program_enrollment.delete()
+
+        return Response(
+            status=status.HTTP_200_OK,
+            data=response
         )
--- a/dashboard/views_test.py
+++ b/dashboard/views_test.py
@@ -16,7 +16,7 @@ from edx_api.enrollments.models import Enrollments, Enrollment
 
 from backends.utils import InvalidCredentialStored
 from courses.factories import ProgramFactory, CourseRunFactory
-from dashboard.factories import UserCacheRefreshTimeFactory
+from dashboard.factories import UserCacheRefreshTimeFactory, ProgramEnrollmentFactory
 from dashboard.models import ProgramEnrollment, CachedEnrollment
 from micromasters.exceptions import PossiblyImproperlyConfigured
 from micromasters.factories import UserFactory, SocialUserFactory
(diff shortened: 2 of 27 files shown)


pdpinch:
@roberthouse54 the pale blue background and shadow seems unusual. Is that what you had in mind?

amir-qayyum-khan:
Removed shadow. It was not part of mockups
screen shot 2018-08-07 at 3 37 14 pm

pdpinch:
Another question for @roberthouse54 -- Amir implemented an [error message when the user isn't logged into any programs](https://user-images.githubusercontent.com/10431250/43640777-a0d51964-973a-11e8-8ce9-c8fde7ded8f1.png). Does this work for you? Let us know if you'd like a review app to try it out.

pdpinch:
@amir-qayyum-khan I'm surprised to see how much has been added to the dashboard code. I was thinking of this as being part of the settings UI. Is that all bundled together?

amir-qayyum-khan:
> @amir-qayyum-khan I'm surprised to see how much has been added to the dashboard code. I was thinking of this as being part of the settings UI. Is that all bundled together?

@pdpinch yes this is all bundle.

static/js/components/LeaveProgramWizard.js
```diff
@@ -0,0 +1,171 @@
...
+import _ from "lodash"
+
+import ProfileFormFields from "../util/ProfileFormFields"
+import type { Profile } from "../flow/profileTypes"
+import type {
+  AvailableProgramsState,
+  AvailableProgram
+} from "../flow/enrollmentTypes"
+import type { UIState } from "../reducers/ui"
+import { showDialog, hideDialog, setProgramsToUnEnroll } from "../actions/ui"
```
mbertrand: Minor nitpick, can you move this import and the one below it above the type imports and below the `ProfileFormFields` import?
(https://github.com/mitodl/eslint-config#import-order)

amir-qayyum-khan:
Done @mbertrand

amir-qayyum-khan:
@mbertrand added more tests to fix coverage

static/js/components/LeaveProgramWizard.js
```diff
@@ -0,0 +1,171 @@
...
+  props: {
+    dispatch: Dispatch,
+    profile: Profile,
+    programs: AvailableProgramsState,
+    ui: UIState
+  }
+
+  setDialogVisibility = (visible: boolean) =>
+    visible
+      ? showDialog(UNENROLL_PROGRAM_DIALOG)
```
rhysyngsun: This is extra complexity that we don't need, you should just inline `showDialog` and `hideDialog`

static/js/components/LeaveProgramWizard.js
```diff
@@ -0,0 +1,171 @@
...
+        contentClassName="dialog unenroll-dialog"
+        className="unenroll-dialog-wrapper"
+        open={isVisible(dialogVisibility)}
+        onRequestClose={() => this.closeDialog()}
+        autoScrollBodyContent={true}
+        maxWidth="xs"
+        actions={dialogActions(
+          this.closeDialog,
+          this.unEnrollUserTask,
+          programsToUnEnrollInFlight,
```
rhysyngsun: Should probably also disable if `programsToUnEnroll.length === 0`

static/js/components/LeaveProgramWizard.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  setDialogVisibility = (visible: boolean) =>
+    visible
+      ? showDialog(UNENROLL_PROGRAM_DIALOG)
+      : hideDialog(UNENROLL_PROGRAM_DIALOG)
+
+  enrolledInPrograms = () => {
+    const { programs } = this.props
+    return R.filter((program: AvailableProgram) => program.enrolled, programs)
```
rhysyngsun: I think this would be better as a partial `const` outside the class and then call that in `renderUnenrollUI()`

static/js/actions/programs.js
```diff
@@ -100,5 +103,42 @@ export const addProgramEnrollment = (
...
 
+export const UNENROLL_PROGRAM_DIALOG = "unenrollProgramDialog"
 export const CLEAR_ENROLLMENTS = "CLEAR_ENROLLMENTS"
 export const clearEnrollments = createAction(CLEAR_ENROLLMENTS)
+
+export const unEnrollProgramEnrollments = (
+  programIds: []
+): Dispatcher<?AvailableProgram> => {
+  return (dispatch: Dispatch) => {
+    dispatch(setUnEnrollApiInFlightState(true))
```
rhysyngsun: this should probably `await` here to avoid any potential race conditions

dashboard/views.py
```diff
@@ -150,3 +152,37 @@ def post(self, request):
...
+    )
+    permission_classes = (permissions.IsAuthenticated, )
+
+    def post(self, request):
+        """
+        unenroll from MM program(s)
+        """
+        program_ids_serializer = UnEnrollProgramsSerializer(data=request.data)
+        program_ids_serializer.is_valid(raise_exception=True)
+        program_ids = program_ids_serializer.data['program_ids']
```
rhysyngsun: This code should all get moved into the serializer, otherwise the serializer is just being used to parse a list of ints.

dashboard/views.py
```diff
@@ -150,3 +152,37 @@ def post(self, request):
         return Response(
             data=enrollment.json
         )
+
+
+class UnEnrollPrograms(APIView):
+    """
+    api that unenroll user from one ore more programs
```
rhysyngsun: Typo here

dashboard/views_test.py
```diff
@@ -322,3 +322,106 @@ def test_enrollment(self, mock_refresh, mock_edx_enr, mock_index):  # pylint: di
...
+        cls.user = SocialUserFactory.create(social_auth__extra_data=social_extra_data)
+        cls.invalid_user = UserFactory.create()
+        cls.url = reverse('unenroll_programs')
+
+    def setUp(self):
+        super().setUp()
+        self.client.force_login(self.user)
+        self.program_enrollments = [ProgramEnrollmentFactory.create(user=self.user) for _ in range(2)]
+
+    def test_unenroll_all_programs(self):
```
rhysyngsun: Should add a test for a mix of valid and invalid program ids

dashboard/views.py
```diff
@@ -150,3 +152,37 @@ def post(self, request):
...
+            program_id__in=program_ids,
+            user=request.user
+        )
+        response = []
+        for program_enrollment in program_enrollments:
+            response.append({
+                'program_id': program_enrollment.program_id,
+                'title': program_enrollment.program.title
+            })
+            program_enrollment.delete()
```
rhysyngsun: Are we confident in the side effects of outright deleting the `ProgramEnrollment` here? We generally don't delete data so we should verify that no parts of our application (particularly any scheduled tasks). Easy ones to find will be any `ProgramEnrollment.objects.get()` queries that aren't error-handled, but I suspect there may be more difficult edge cases to tease out. We may want to consider just marking these and inactive and filtering them out instead, although I realize that might be just as laborious.

pdpinch: It's been my practice to delete ProgramEnrollment records when users wish to have their account deleted. I know this isn't exactly the same though, so I guess there are more cases to consider. 

If we take your suggestion of marking them inactive, will we need a migration, along with updating the code to filter out inactive ProgramEnrollments?

rhysyngsun: Yes, well need both. I guess my point was, the amount of effort to do that is probably equal to or possibly greater than the amount of effort needed to shore up existing code. that doesn't handle situations where the record doesn't exist and is expected to, because those are probably the minority of all `ProgramEnrollment` code.

TLDR: deleting is **probably** still the best option (dragons aside), but we definitely need to test and/or audit at least the obvious code points that use these records to head off any errors. We probably won't catch them all, but unhandled `get(user=?, program=?)` calls are obvious candidates for where you'd hit that kind of error.

amir-qayyum-khan: i tested all pages after removing all or single program, they are working fine.

amir-qayyum-khan:
@pdpinch @rhysyngsun should we remove staff role if user as well, to reindex work properly.

dashboard/views.py
amir-qayyum-khan: @rhysyngsun any thoughts

rhysyngsun: @amir-qayyum-khan you should test code that isn't on pages too, such as celery tasks, that query this.

pdpinch:
@amir-qayyum-khan what's the status of this?

amir-qayyum-khan:
celery tasks and signals were tested manually. I think i need to add new unit tests for them.

amir-qayyum-khan:
Done @rhysyngsun

amir-qayyum-khan:
Done @mbertrand,  I added tests related get(user=?, program=?) use in MM app and js test.

Issue #3917, opened 2018-03-27
link email to mail settings page without requiring login


pdpinch:
As a user, I'd like to be able to unsubscribe from bulk mailings without having to log in. 

This can be implemented similarly to Open Discussions in [PR 506](https://github.com/mitodl/open-discussions/pull/506/files).

amir-qayyum-khan:
@pdpinch , [here](https://github.com/mitodl/micromasters/blob/master/mail/views.py#L130) we are sending batch emails. I dont think so it is possible to create custom unsub link of every user in current state of code

pdpinch:
The `email_body` is being constructed from a template where we insert per-user variables, like name. Can we do that here?

@rhysyngsun how does this work in open-discussions?

amir-qayyum-khan:
- i dont think so we are using template. we are just forwarding text body. 
https://github.com/mitodl/micromasters/blob/master/mail/views.py#L130
- we have only footer as html.
https://github.com/mitodl/micromasters/blob/master/mail/utils.py#L110
- And email dont have name:


<img width="1010" alt="screen shot 2018-06-26 at 7 36 50 pm" src="https://user-images.githubusercontent.com/10431250/41919587-50a4747e-7978-11e8-8a96-4c93c55d0772.png">

rhysyngsun:
@pdpinch we only send one email at a time and don't use mailgun's batch API, so the email link issue will have to get solved differently here. The shortest path might be to implement mailgun's unsubscribe functionality and process those webhooks: https://help.mailgun.com/hc/en-us/articles/203306610-Do-I-control-the-unsubscribe-handling-or-does-Mailgun-

pdpinch:
We had to revert this. @gsidebo should be able to explain why.

PR #4094, opened 2018-08-03
Added mailgun unsub user support 

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3917

#### What's this PR do?
- If email-optin is updated to False then it adds user to mailgun unsub list.
  - User  will not receive any email from MM mailgun address.
  - **Question:** Is discussion uses MM mailgun address? Should it be block?
-  If email-optin is updated to True then it removes user from mailgun unsub list.

**To enable/disable unsubscribes:**
- Enable unsubscription feature for your domain.
- Remove text in the html and text footer templates so they won’t be appended automatically.
- Insert a variable in the html and text bodies of your email when you need unsubscribe links.
- This variable will be replaced by the corresponding unsubscribe link.

https://documentation.mailgun.com/en/latest/user_manual.html#tracking-unsubscribes

**Also set the webhook:**
- `api/v0/mail/unsub_webhook/`

@pdpinch 

#### Screenshoot:
screen shot 2018-07-04 at 7 41 54 pm

screen shot 2018-07-04 at 7 42 34 pm

### DIFF
--- a/mail/api.py
+++ b/mail/api.py
@@ -285,6 +285,76 @@ class MailgunClient:
         )
         return response
 
+    @staticmethod
+    def add_email_to_unsub_list(url, email):
+        """
+        It adds user email to mailgun unsub list.
+
+        Args:
+            url (str): mailgun api url:
+            email (str): user email
+        """
+        try:
+            response = requests.post(
+                url,
+                auth=('api', settings.MAILGUN_KEY),
+                data={
+                    'address': email,
+                    'tag': '*'
+                }
+            )
+        except requests.exceptions.RequestException:
+            log.exception(
+                "Unable to add email: %s to mailgun unsubscribes list.", email
+            )
+            return
+        if response.status_code == status.HTTP_200_OK:
+            log.debug(
+                "Added user's email: %s to mailgun unsubscribes list. Message received: %s",
+                email,
+                response.json()
+            )
+
+    @staticmethod
+    def remove_email_from_unsub_list(url, email):
+        """
+        It removes user email from mailgun unsub list.
+
+        Args:
+            url (str): mailgun api url:
+            email (str): user email
+        """
+        try:
+            response = requests.delete(url, auth=('api', settings.MAILGUN_KEY))
+        except requests.exceptions.RequestException:
+            log.exception(
+                "Unable to remove email: %s from mailgun unsubscribes list.", email
+            )
+            return
+        if response.status_code == status.HTTP_200_OK:
+            log.debug(
+                "Removed user's email: %s from mailgun unsubscribes list. Message received: %s",
+                email,
+                response.json()
+            )
+
+    @staticmethod
+    def toggle_mailing_list_subscription(email_optin, email):
+        """
+        it removes user from mailgun unsubscribes list.
+        https://documentation.mailgun.com/en/latest/api-suppressions.html#delete-a-single-unsubscribe
+
+        Args:
+            email_optin (bool): email optin flag
+            email (str): user email
+        """
+        url = "{base}/unsubscribes".format(base=settings.MAILGUN_URL)
+        if email_optin:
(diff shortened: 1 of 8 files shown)


amir-qayyum-khan:
@mitodl/devops there are some change on mailgun app require for this task. please see description.

profiles/views.py
```diff
@@ -37,6 +46,96 @@ class ProfileViewSet(RetrieveModelMixin, UpdateModelMixin, GenericViewSet):
...
+
+    def update(self, request, *args, **kwargs):
+        """
+        updates user profile.
+        """
+        if 'email_optin' in request.data and 'email' in request.data:
+            # perform mailgun action if email optin is set then remove user from mailgun unsubscription
+            # list otherwise add user to mailgun unsubscription.
+            ProfileViewSet.mailgun_action(request.data['email_optin'], request.data['email'])
+        return super().update(request, *args, **kwargs)
```
gsidebo: I think we can override DRF's `perform_update` method (documentation in [this section](http://www.django-rest-framework.org/api-guide/generic-views/#methods)) instead of overriding `update`. `perform_update` exists specifically to override and provide some custom saving functionality. Might look something like this:

```python
def perform_update(self, serializer):
  profile = serializer.save()
  if 'email_optin' in serializer.validated_data:
    MailgunClient.toggle_unsubscription(profile, serializer.validated_data['email_optin'])
```

amir-qayyum-khan:
Done @gsidebo

mail/api.py
```diff
@@ -285,6 +285,76 @@ def send_course_team_email(  # pylint: disable=too-many-arguments
...
+            return
+        if response.status_code == status.HTTP_200_OK:
+            log.debug(
+                "Removed user's email: %s from mailgun unsubscribes list. Message received: %s",
+                email,
+                response.json()
+            )
+
+    @staticmethod
+    def toggle_mailing_list_subscription(email_optin, email):
```
gsidebo: Sorry, I should have noticed this during the last review. I think this method should be removed, and `ProfileViewset.perform_update` can inspect the `email_optin` value and decide to call `add_email_to_unsub_list` or `remove_email_to_unsub_list`. The `add`/`remove` methods can declare the url on their own instead of passing it in, so both of them would take `email` as their only argument

mail/views.py
```diff
@@ -270,3 +270,30 @@ def post(self, request, *args, **kargs):  # pylint: disable=unused-argument
...
+
+        if recipient and event == "unsubscribed":
+            profiles_updated = Profile.objects.filter(user__email=recipient).update(email_optin=False)
+            if profiles_updated > 0:
+                log.debug(
+                    "Webhook event: '%s' received by Mailgun for recipient: %s:", event, recipient
+                )
+            else:
+                log.error(
+                    "Webhook event: Unable to update profile of user: "
```
gsidebo: Please change this message to: 

```
"Webhook event: '%s' failed. No profile exists for a user with email '%s'" % event, recipient
```

profiles/views.py
```diff
@@ -37,6 +43,17 @@ class ProfileViewSet(RetrieveModelMixin, UpdateModelMixin, GenericViewSet):
     serializer_class_filled_out = ProfileFilledOutSerializer
     serializer_class_limited = ProfileLimitedSerializer
 
+    def perform_update(self, serializer):
+        """Updates object"""
+        profile = serializer.save()
+        if 'email_optin' in serializer.validated_data:
+            # perform mailgun action if email optin is set then remove user from mailgun unsubscription
+            # list otherwise add user to mailgun unsubscription.
```
gsidebo: This comment can be removed when you make the changes that I recommended above. It will be obvious what the code is doing

amir-qayyum-khan:
Done @gsidebo

amir-qayyum-khan:
@mitodl/devops are you done with setting mention in description of PR

PR #4054, opened 2018-07-06
Removed  CYBERSOURCE_TRANSACTION_KEY key

amir-qayyum-khan:
#### What are the relevant tickets?
fixes https://github.com/mitodl/micromasters/issues/4039

#### What's this PR do?
it removes  CYBERSOURCE_TRANSACTION_KEY key from settings.py

#### How should this be manually tested?
- run mm and see everything works

@pdpinch

PR #4096, opened 2018-08-06
SESSION_ENGINE is not a required setting

pdpinch:

none

#### What's this PR do?

stops required SESSION_ENGINE for new apps

Does the review application for this PR build? Or, at least, does it fail somewhere else than on a missing value for SESSION_ENGINE?

### DIFF
--- a/app.json
+++ b/app.json
@@ -56,9 +56,6 @@
     "CYBERSOURCE_SECURITY_KEY": {
       "description": "CyberSource API key"
     },
-    "CYBERSOURCE_TRANSACTION_KEY": {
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -56,6 +56,9 @@
     "CYBERSOURCE_SECURITY_KEY": {
       "description": "CyberSource API key"
     },
+    "CYBERSOURCE_TRANSACTION_KEY": {
```
rhysyngsun: This is adding this key again, which was recently removed in https://github.com/mitodl/micromasters/pull/4054/files, is this intentional?

Issue #3839, opened 2018-02-14
Remove no-member from pylintrc disable list


noisecapella:
I'm not sure why we added it in the first place, but it's useful and we should investigate disabling warnings on a case by case basis instead, if possible

pdpinch:
Amir started a PR (#4101) for this, but it looks more complicated than we thought.

PR #4101, opened 2018-08-07
Remove no member

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/3839

#### What's this PR do?
it enables `no-member` on pylint 

travis-ci should work

@pdpinch

### DIFF
--- a/courses/factories.py
+++ b/courses/factories.py
@@ -51,12 +51,12 @@ class CourseFactory(DjangoModelFactory):
 class CourseRunFactory(DjangoModelFactory):
     """Factory for CourseRuns"""
     title = factory.LazyAttribute(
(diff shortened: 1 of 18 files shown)


pdpinch:
@noisecapella is this what you had in mind?

It's passing tests, so I think it's ok to merge. But the warning is disabled in a lot of places now.

noisecapella:
I'm not sure this is a good idea. I filed that issue in the hope that it was a false error and we didn't really need to disable it everywhere.

I did update pylint locally and dependent packages and the warning went away. However there is a significant change in behavior where the linting happens all at once in the beginning, which we might want to think about some more.

pdpinch:
I'm going to close this PR then.

PR #4102, opened 2018-08-07
Show semester year in GradeDetailPopup

### LINKED ISSUE #4078: Dashboard: add course dates to detail view
As a user, when I look at the detailed list of grades on my dashboard, I can't tell which course is which. 

screenshot 2018-07-27 08 28 17

It seems like the simplest thing to do would be to add the course start and end dates to the end of the course title.

annagav:
Fix #4078

#### What's this PR do?
Show semester year for a course run instead of course title.

screen shot 2018-08-07 at 12 04 07 pm

### DIFF
--- a/courses/utils.py
+++ b/courses/utils.py
@@ -91,6 +91,18 @@ def get_year_season_from_course_run(course_run):
         return season_year_tuple
 
 
+def format_season_year_for_course_run(course_run):
+    """
+    Args:
+        course_run (CourseRun): A CourseRun object
+
+    Returns:
+        string: containing season and year, eg: 'Spring 2018'
+    """
+    year_season_tuple = get_year_season_from_course_run(course_run)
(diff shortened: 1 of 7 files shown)


courses/utils.py
```diff
@@ -91,6 +91,18 @@ def get_year_season_from_course_run(course_run):
         return season_year_tuple
 
 
+def format_season_year_for_course_run(course_run):
+    """
+    Args:
+        course_run (CourseRun): A CourseRun object
+
+    Returns:
+        string: containing season and year, eg: 'Spring 2018'
```
rhysyngsun: Nit: change to `str` type

courses/utils.py
```diff
@@ -91,6 +91,18 @@ def get_year_season_from_course_run(course_run):
         return season_year_tuple
 
 
+def format_season_year_for_course_run(course_run):
+    """
```
rhysyngsun: Could you add a text description of the function?

rhysyngsun:
Functionality looks good

PR #4116, opened 2018-08-23
Audited passed, then audited failed course

### LINKED ISSUE #4117: Dashboard: wrong state when audit learner passed then failed course
See https://odl.zendesk.com/agent/tickets/20106

When an audit learner passes a course and then subsequently fails it, we should still tell them that they have passed.

annagav:
Fix https://odl.zendesk.com/agent/tickets/20106

fixes #4117

#### What's this PR do?
Two main fixes:

- In the situation where user has passed a course but did not pay, the course run status is `CAN_UPGADE` or `MISSED_DEADLINE`. The existing code would only check for PASSED status, which happens only when user has paid.
- If a user passed a course and then failed it, the most recent one will be returned first, and the status messages would be calculated incorrectly for the overall status of the course.

The changes are made on the assumption that the backend returns status PASSED only for paid courses, and for any course run that is/was audited the status would be CAN_UPGRADE, MISSED_DEADLINE, or NOT_PASSED.

Run `scripts/test/run_snapshot_dashboard_states.sh`

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages.js
+++ b/static/js/components/dashboard/courses/StatusMessages.js
@@ -36,7 +36,9 @@ import {
   userIsEnrolled,
   isOfferedInUncertainFuture,
   isPassedOrCurrentlyEnrolled,
-  notNilorEmpty, hasCanUpgradeCourseRun, hasMissedDeadlineCourseRun
+  notNilorEmpty,
+  hasCanUpgradeCourseRun,
+  hasMissedDeadlineCourseRun
 } from "./util"
 import {
   hasPassingExamGrade,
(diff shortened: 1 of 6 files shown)


static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -342,90 +345,81 @@ export const calculateMessages = (props: CalculateMessagesProps) => {
...
         )
       )
     }
+    return S.Just(messages)
+  } else if (hasCanUpgradeCourseRun(course)) {
+    //the course finished but can still pay
+    const dueDate = paymentDueDate.isValid()
+      ? ` (Payment due on ${paymentDueDate.format(DASHBOARD_FORMAT)})`
+      : ""
+    if (exams) {
```
mbertrand: Should this be `if (exams && !passedExam)`, for people who didn't pay and also failed the exam?  And then the message would be `The edX course is complete, but you need to pass the exam and pay to get credit.${dueDate}`?

PR #4123, opened 2018-08-29
Update progress message for staff

annagav:
Related to #4116 
#### What's this PR do?
Updates the progress message to give appropriate message to auditing learners. When learner audited the course and passed it, the status of course run will never be `STATUS_PASSED`, the possible values can only be `STATUS_MISSED_DEADLINE` or `STATUS_CAN_UPGRADE`.

Create a passed audited course for user, and a failed audited course that is more recent. Similar to the selenium test in #4116. Look at the ProgressMessage from staff view. It should say that the user passed but did not pay for the course.

### DIFF
--- a/static/js/components/dashboard/courses/ProgressMessage.js
+++ b/static/js/components/dashboard/courses/ProgressMessage.js
@@ -26,7 +26,9 @@ import {
   hasEnrolledInAnyRun,
   courseUpcomingOrCurrent,
   hasPaidForAnyCourseRun,
-  hasPassedCourseRun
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
it was stuck on `docker-compose run web ./manage.py migrate`, after 5 mins waiting i pressed `control z`. then it show me that all migrations are installed. I ran migrate command again and found all migrations were install. but it was stuck with no display.

screen shot 2018-09-04 at 6 54 09 pm

PR #4132, opened 2018-09-11
Oauth maintenance page on login

### LINKED ISSUE #4131: OAuth Login out of service maintenance message
edX is planning downtime for their OAuth API the morning of 9/18. During the planned one hour downtime the API will be unavailable or unresponsive (returning 403, 404 or 500 or just timing out). MicroMasters users (and, by extension, Discussions users) won't be able to log in during this time. 

It seems like the least impact would be to put the MicroMasters login into maintenance mode:

### acceptance criteria:
- [ ] add a setting for `OAUTH_MAINTENANCE_MODE` (don't forget docs & sane default)
- [ ] when `OAUTH_MAINTENANCE_MODE=True` the login endpoint should return a static message indicating that the site is down for maintenance: “MicroMasters login is unavailable now. Please try again later today.” 
- [ ] Review the site for any other endpoints that need a similar change
- [ ] Separately @mitodl/devops will shutdown the MIcroMasters workers during this time so that they don't attempt to download any data from edX.

annagav:
Fix #4131

#### What's this PR do?
If the flag `OAUTH_MAINTENANCE_MODE ` is on redirect to a page with the message.

Set env var `OAUTH_MAINTENANCE_MODE` and try to login.

screen shot 2018-09-11 at 2 38 41 pm

### DIFF
--- a/app.json
+++ b/app.json
@@ -187,7 +187,8 @@
     },
     "OAUTH_MAINTENANCE_MODE": {
       "description": "If true, on login redirects to a static page with message",
-      "value": "false"
+      "value": "False",
+      "required": false
     },
     "OPEN_DISCUSSIONS_API_USERNAME": {
       "description": "The username to use for authentication Open Discussions",
--- a/app.json
+++ b/app.json
@@ -185,6 +185,10 @@
       "description": "If false, disables the node_modules cache to fix yarn install",
       "value": "false"
     },
+    "OAUTH_MAINTENANCE_MODE": {
+      "description": "If true, on login redirects to a static page with message",
+      "value": "false"
+    },
     "OPEN_DISCUSSIONS_API_USERNAME": {
       "description": "The username to use for authentication Open Discussions",
       "required": false
--- a/backends/urls.py
+++ b/backends/urls.py
@@ -23,7 +23,7 @@ urlpatterns = [
     # authentication / association
 
     url(r'^login/(?P<backend>[^/]+){0}$'.format(extra),
-        RedirectView.as_view(url='/oauth_maintenance') if settings.OAUTH_MAINTENANCE_MODE else views.auth,
+        RedirectView.as_view(pattern_name='oauth_maintenance') if settings.OAUTH_MAINTENANCE_MODE else views.auth,
         name='begin'),
     # disconnection
     url(r'^disconnect/(?P<backend>[^/]+){0}$'.format(extra), views.disconnect,
--- a/backends/urls.py
+++ b/backends/urls.py
@@ -1,5 +1,6 @@
 """URLs module"""
 from django.conf.urls import url
+from django.conf import settings
 
 from social_django import views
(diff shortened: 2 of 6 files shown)


ui/views.py
```diff
@@ -188,6 +188,13 @@ def need_verified_email(request, *args, **kwargs):  # pylint: disable=unused-arg
     return standard_error_page(request, 401, "verify_email.html")
 
 
+def aouth_maintenance(request, *args, **kwargs):  # pylint: disable=unused-argument
```
rhysyngsun: Typo: 'oauth'

ui/views.py
```diff
@@ -188,6 +188,13 @@ def need_verified_email(request, *args, **kwargs):  # pylint: disable=unused-arg
...
+    """
+    Returns maintenance page during oauth downtime
+    """
+    return standard_error_page(request, 503, "oauth_maintenance.html")
```
rhysyngsun: I think this 503 status code might cause superfluous error reporting because it's a 5xx series. I agree 503 is probably the most accurate of the available statuses, I think this would be better as a 200, because it's an expected situation.

micromasters/settings.py
```diff
@@ -186,6 +186,8 @@
 LOGIN_URL = '/'
 LOGIN_ERROR_URL = '/'
 
+OAUTH_MAINTENANCE_MODE = get_bool('OAUTH_MAINTENANCE_MODE', False)
```
rhysyngsun: Could you add this to app.json too please?

ui/views.py
rhysyngsun: Actually, I think a slightly better approach would be to add the maintenance page at it's own url (e.g. /login-offline) and in urls.py use `RedirectView` conditionally: https://docs.djangoproject.com/en/dev/ref/class-based-views/base/#redirectview

app.json
```diff
@@ -185,6 +185,10 @@
       "description": "If false, disables the node_modules cache to fix yarn install",
       "value": "false"
     },
+    "OAUTH_MAINTENANCE_MODE": {
+      "description": "If true, on login redirects to a static page with message",
+      "value": "false"
```
rhysyngsun: I think you need to uppercase this to `"True"` and add `"required": false` as well.

backends/urls.py
```diff
@@ -19,7 +21,9 @@
 
     # These three go to views in social_core
     # authentication / association
-    url(r'^login/(?P<backend>[^/]+){0}$'.format(extra), views.auth,
+
+    url(r'^login/(?P<backend>[^/]+){0}$'.format(extra),
+        RedirectView.as_view(url='/oauth_maintenance') if settings.OAUTH_MAINTENANCE_MODE else views.auth,
```
rhysyngsun: I think this would be better as `pattern_name='oauth_maintenance'` instead of `url=?`

Issue #4139, opened 2018-09-26
management command for retiring users


pdpinch:
we have been following a simple manual process for retiring users who want to have their account removed. For consistency and to avoid errors, we should commit it to a management command. 

### acceptance criteria
- [x] replicate the current, manual process:
  - [x] set `is_active=False` on the user
  - [x] set `email_optin=False` on the user's profile
  - [x] delete the user's program enrollment object(s)
    - deleting the program enrollments should signal an update to the elastic search index
- [x] delete any social auths (notably edX oauth tokens)
   - this will prevent any asynchronous data updates

amir-qayyum-khan:
@pdpinch as you suggested i am putting this command in profile app

PR #4153, opened 2018-10-09
Added command to retire user

amir-qayyum-khan:
fixes https://github.com/mitodl/micromasters/issues/4139

#### What's this PR do?
Adds a command which retire user(s) from MM

- [ ] try to retire no user and test error message:
  - `docker-compose run web ./manage.py retire_users`
  - `docker-compose run web ./manage.py retire_users --user=`
- [ ] try to retire one user and check User, profile, ProgramEnrollment, User social auths pages in admin dashboard:
  - `docker-compose run web ./manage.py retire_users --user=fake.raul.ramirez`
- [ ] try to retire multiple users and check User, profile, ProgramEnrollment, User social auths pages in admin dashboard:
  - `docker-compose run web ./manage.py retire_users --user=fake.raul.ramirez --user=fake.hector.ramos`
- [ ] try to retire fake user and test error message:
  - `docker-compose run web ./manage.py retire_users --user=foo`

@pdpinch 

#### Screenshot
screen shot 2018-10-10 at 2 03 49 pm

screen shot 2018-10-10 at 2 05 46 pm

### DIFF
--- a/profiles/management/commands/retire_users.py
+++ b/profiles/management/commands/retire_users.py
@@ -115,4 +115,4 @@ For multiple users, add arg `--user` for each user i.e:\n
 
             # finish
             self.display_messages(f"User '{user_name}' is retired", log_messages)
-            log.info(", ".join(log_messages))
+            log.info("\n".join(log_messages[1:]))
--- a/profiles/management/commands/retire_users.py
+++ b/profiles/management/commands/retire_users.py
@@ -1,0 +1,90 @@
+"""
+Retire user from MM
+"""
+from argparse import RawTextHelpFormatter
+from django.contrib.auth.models import User
+from django.core.management import BaseCommand, CommandError
+from social_django.models import UserSocialAuth
+
+from dashboard.models import ProgramEnrollment
+
+
+class Command(BaseCommand):
+    """
+    Retire user from MicroMasters
+    """
+    help = """
+Retire one or multiple users. For single user use:\n
+`./manage.py retire_users --user=foo` or do \n
+`./manage.py retire_users -u foo` \n
+
+For multiple users, add arg `--user` for each user i.e:\n
+`./manage.py retire_users --user=foo --user=bar --user=baz` or do \n
+`./manage.py retire_users -u foo -u bar -u baz`
+"""
+
+    def create_parser(self, prog_name, subcommand):
+        """
+        create parser to add new line in help text.
+        """
+        parser = super(Command, self).create_parser(prog_name, subcommand)
+        parser.formatter_class = RawTextHelpFormatter
+        return parser
+
+    def add_arguments(self, parser):
+        """create args"""
+        # pylint: disable=expression-not-assigned
+        parser.add_argument(
+            '-u',
+            '--user',
+            action='append',
+            default=[],
+            dest='users',
+            help="Single or multiple user name"
+        ),
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        user_names = kwargs.get("users", [])
+
+        if len(user_names) <= 0:
+            # show error when no user selected.
+            raise CommandError("Please select user(s)")
+
+        for user_name in user_names:
+            # retire user
+            self.stdout.write(f"Retiring user {user_name}")
+
+            if not user_name:
+                # invalid user name, can be empty string
+                self.stdout.write(f"Invalid username: '{user_name}'", self.style.ERROR)
+                continue
+
+            try:
+                user = User.objects.get(username=user_name)
+            except User.DoesNotExist:
+                self.stdout.write(
+                    f"User '{user_name}' does not exist in MicroMasters",
+                    self.style.ERROR
+                )
+                continue
+
+            # mark user inactive
+            user.is_active = False
+            user.save()
+            self.stdout.write(f"User {user_name} is_active set to False")
+
+            # reset email_optin
+            user.profile.email_optin = False
+            user.profile.save()
+            self.stdout.write(f"User {user_name} email_optin set to False")
+
+            # reset program enrollments
+            enrollment_delete_count, _ = ProgramEnrollment.objects.filter(user=user).delete()
+            self.stdout.write(f"User {user_name}, {enrollment_delete_count} ProgramEnrollments rows deleted")
+
+            # reset user social
+            auth_delete_count, _ = UserSocialAuth.objects.filter(user=user).delete()
+            self.stdout.write(f"User {user_name}, {auth_delete_count} SocialAuth rows deleted")
+
+            # finish
+            self.stdout.write(f"User '{user_name}' is retired")
--- a/profiles/management/commands/retire_users.py
+++ b/profiles/management/commands/retire_users.py
@@ -1,12 +1,15 @@
 """
 Retire user from MM
 """
+import logging
 from argparse import RawTextHelpFormatter
 from django.contrib.auth.models import User
 from django.core.management import BaseCommand, CommandError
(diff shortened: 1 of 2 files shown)


profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,61 @@
...
+from social_django.models import UserSocialAuth
+
+from dashboard.models import ProgramEnrollment
+
+
+class Command(BaseCommand):
+    """
+    Validate prices and financial aid
+    """
+    help = "Retire one or multiple users"
```
pdpinch: Can we add a usage example to the help message?

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,61 @@
...
+from django.contrib.auth.models import User
+from django.core.management import BaseCommand, CommandError
+from social_django.models import UserSocialAuth
+
+from dashboard.models import ProgramEnrollment
+
+
+class Command(BaseCommand):
+    """
+    Validate prices and financial aid
```
pdpinch: please update this docstring

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,61 @@
...
+        user_names = kwargs.get("users", [])
+
+        if len(user_names) <= 0:
+            # show error when no user selected.
+            raise CommandError("Please select user(s)")
+
+        for user_name in user_names:
+            """retire user"""
+            if not user_name:
+                self.stdout.write("Invalid username.")
```
pdpinch: Please include the username in the error message

amir-qayyum-khan: this is for invalid user, i.e null or empty string. Can be produce with command
`docker-compose run web ./manage.py retire_users --user=`

profiles/management/commands/retire_users.py
amir-qayyum-khan: Done @pdpinch

amir-qayyum-khan:
Done @pdpinch

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,81 @@
...
+from django.contrib.auth.models import User
+from django.core.management import BaseCommand, CommandError
+from social_django.models import UserSocialAuth
+
+from dashboard.models import ProgramEnrollment
+
+
+class Command(BaseCommand):
+    """
+    Deactivate user from Micromasters
```
pdpinch: `Retire user from MicroMasters`

pdpinch:
Can you check to see what happens with the celery task that refreshes the edX cache? With the account deactivated and no social auth objects, I expect it can't update the cache. But does it throw an error (bad) or skip the user (good) or is the user not even included in the query (better)?

pdpinch:
Do we need some tests?

amir-qayyum-khan:
Added tests @pdpinch 

> Can you check to see what happens with the celery task that refreshes the edX cache? With the account deactivated and no social auth objects, I expect it can't update the cache. But does it throw an error (bad) or skip the user (good) or is the user not even included in the query (better)?

it logs exception and exit task

https://github.com/mitodl/micromasters/blob/master/dashboard/api.py#L618

amir-qayyum-khan:
@pdpinch are you ok with this PR?

profiles/management/commands/retire_users.py
pdpinch: But it's specifically for a username that's false, so the error could be clearer. How about "Username required"

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,98 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Retire user from MicroMasters
+    """
+    help = """
+Retire one or multiple users. For single user do like this:\n
```
pdpinch:
```suggestion
Retire one or multiple users. For single user use:\n
```

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,98 @@
...
+            raise CommandError("Please select user(s)")
+
+        for user_name in user_names:
+            # retire user
+            self.command_message(f"Retiring user {user_name}")
+
+            if not user_name:
+                # invalid user name, can be empty string
+                self.command_message(f"Invalid username: '{user_name}'", True)
+                continue
```
pdpinch: Actually, I think you could just omit this test, since if username is null, it's going to fail the next test anyway.

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,98 @@
...
+
+            try:
+                user = User.objects.get(username=user_name)
+            except User.DoesNotExist:
+                self.command_message(f"User '{user_name}' does not exist in MicroMasters", True)
+                continue
+
+            # mark user inactive
+            self.command_message(f"User {user_name} is_active set to False")
```
pdpinch: Should the message go after the action? For all 4 of these?

pdpinch:
I have a couple more changes, and I'd really like another opinion, like @annagav or @rhysyngsun

profiles/management/commands/retire_users.py
amir-qayyum-khan: i was thinking why hit database if username is null. I think this check will save operation cost.

amir-qayyum-khan:
refactored @pdpinch

amir-qayyum-khan:
Done @rhysyngsun. please review

profiles/management/commands/retire_users.py
```diff
@@ -0,0 +1,93 @@
...
+            default=[],
+            dest='users',
+            help="Single or multiple user name"
+        ),
+
+    def command_message(self, message, is_error=False):
+        """
+        display and log message
+        """
+        self.stdout.write(message, self.style.ERROR)
```
rhysyngsun: This is going to print everything as an error, which they're not all, you're probably better off just inlining this and doing the correct calls as-needed

amir-qayyum-khan:
Done @rhysyngsun

amir-qayyum-khan:
any update @rhysyngsun. By the way @pdpinch asked me to do logging as we as display information. He wants logging for future lookups.

pdpinch:
To be clear, what I would like is 1) immediate output that I can copy into a ticket as a record of me retiring the user and 2) similar output to the log. 

If stdout isn’t the write way to do it, I’m happy with anything that does these two things.

rhysyngsun:
Just want to flag that multiple log statements will possibly be interspersed with other log statements from workers and the web app (hence probably making it harder to search for all the log output for a particular run of the command). We may instead want to accumulate and output a single block of result report at the end (similar to a stacktrace) so that you can see all the output in one log statement.

pdpinch:
That's a good point. I think it makes sense to "accumulate and output a single block of result report at the end"

amir-qayyum-khan:
Done @rhysyngsun . Now logging like
```
 [2018-10-24 14:02:30] INFO 1 [profiles.management.commands.retire_users] retire_users.py:118 - [70fb69cc0352] - Retiring user fake.eugenio.fuentes, User fake.eugenio.fuentes is_active set to False, User fake.eugenio.fuentes email_optin set to False, For user fake.eugenio.fuentes: 0 ProgramEnrollments rows deleted, For user fake.eugenio.fuentes: 0 SocialAuth rows deleted, User 'fake.eugenio.fuentes' is retired
```

amir-qayyum-khan:
@rhysyngsun any update on this?

amir-qayyum-khan:
Done @rhysyngsun and @pdpinch

PR #4162, opened 2018-10-26
make the missed payment deadline message work for all learners

pdpinch:

none yet

#### What's this PR do?

- adds a word that was missing from the dashboard message
- makes the message work for all learners, whether or not they've passed the course

- waiting on a new dashboard test state from @annagav 

- Needs a screenshot


annagav:
screen shot 2018-11-02 at 11 14 57 am

screen shot 2018-11-02 at 11 14 22 am

pdpinch:
I'm still a little concerned about how we're choosing the dates, but I think that can be a separate PR (if it's needed at all).

PR #4189, opened 2018-11-29
Handle users who are inactive or have no profiles during populate_query_memberships

### LINKED ISSUE #4188: populate_query_memberships task processes inactive users and fails on missing profiles
The task should not process inactive users, and should not fail if a user has no profile (log an exception and continue instead).

mbertrand:
Fixes #4188 

#### What's this PR do?
Modifies `search.api.populate_query_memberships` to skip inactive users.
Modifies `search.api._search_percolate_queries` to gracefully handle users with missing profiles
Modifies `search.indexing_api.serialize_program_enrolled_user` to log an exception if a user has no profile.

For a given program, inactivate a user and add a user without a profile or delete the profile of an existing user in the program.
Create a new channel for the program.  The channel should be created successfully, with all contributors and moderators added (minus inactive user and user without profile).
You can also run the `populate_query_memberships` manually for the channel's `PercolateQuery` and make sure it finishes successfully.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -278,16 +278,17 @@ def _search_percolate_queries(program_enrollment):
     conn = get_conn()
     percolate_index = get_default_alias(PERCOLATE_INDEX_TYPE)
     doc = serialize_program_enrolled_user(program_enrollment)
-    if doc:
-        # We don't need this to search for percolator queries and
-        # it causes a dynamic mapping failure so we need to remove it
-        del doc['_id']
-        result = conn.percolate(percolate_index, GLOBAL_DOC_TYPE, body={"doc": doc})
-        failures = result.get('_shards', {}).get('failures', [])
-        if len(failures) > 0:
-            raise PercolateException("Failed to percolate: {}".format(failures))
-        return [int(row['_id']) for row in result['matches']]
-    return []
+    if not doc:
+        return []
+    # We don't need this to search for percolator queries and
+    # it causes a dynamic mapping failure so we need to remove it
+    del doc['_id']
+    result = conn.percolate(percolate_index, GLOBAL_DOC_TYPE, body={"doc": doc})
+    failures = result.get('_shards', {}).get('failures', [])
+    if len(failures) > 0:
+        raise PercolateException("Failed to percolate: {}".format(failures))
+    return [int(row['_id']) for row in result['matches']]
+
 
 
 def adjust_search_for_percolator(search):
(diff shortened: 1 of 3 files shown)


search/api_test.py
```diff
@@ -476,3 +477,27 @@ def test_populate_query_memberships(self, source_type, is_member, query_matches,
             membership = PercolateQueryMembership.objects.get(user=profile.user, query=query)
             assert membership.is_member is query_matches
             assert membership.needs_update is True
+
+    @ddt.data(*product(
+        [True, False],
+        [True, False]
+    ))
```
gsidebo: Instead of using `*product` here, can you just put in all the different combinations and add parameters for the expected `percolate` call count and the expected number of `PercolateQueryMembership` objects? If I'm not mistaken that would be just 4 sets of parameters and would be more readable

search/api.py
```diff
@@ -278,14 +278,16 @@ def _search_percolate_queries(program_enrollment):
...
+    if doc:
+        # We don't need this to search for percolator queries and
+        # it causes a dynamic mapping failure so we need to remove it
+        del doc['_id']
+        result = conn.percolate(percolate_index, GLOBAL_DOC_TYPE, body={"doc": doc})
+        failures = result.get('_shards', {}).get('failures', [])
+        if len(failures) > 0:
+            raise PercolateException("Failed to percolate: {}".format(failures))
+        return [int(row['_id']) for row in result['matches']]
+    return []
```
gsidebo: Nit: might be cleaner to return right away if `doc` is falsey.

```python
doc = serialize_program_enrolled_user(program_enrollment)
if not doc:
  return []

del doc['_id']
# ...
```

mbertrand:
@gsidebo to be able to create new channels from MM, I think you need to go to the admin page and create a new staff role for the user/program you want to create channels with/for.  Then there should be a 'New Channel' button on the `/learners` page for that channel.

pdpinch:
IIRC, your user needs to be a Django admin as well.

gsidebo:
@mbertrand My user is a superuser and staff for the given program. Problems begin when I fill out the new channel form. I get redirected to this URL, which results in a 404:
ss 2018-12-03 at 14 27 41

This is almost definitely an incorrectly configured settings value, but my .env matches the open discussions README. After that 404, I get a huge amount of repeated periodic errors from calls to `sync_channel_memberships`

PR #4202, opened 2018-12-07
fix: don't allow learners with a deleted exam run schedule an exam

### LINKED ISSUE #4201: Exams: don't allow learners with a deleted exam run schedule an exam
Related to #4196 

To delete an `ExamAuthorization` we need to set its operation to `OPERATION_DELETE` and then wait for the updated state to be acknowledged by Pearson. 

The learner should not be able to schedule an exam at any time while the operation is set to `OPERATION_DELETE`.

tasawernawaz:
fixes #4201 

#### What's this PR do?
The learner should not be able to schedule an exam at any time while the operation is set to `OPERATION_DELETE`.


pdpinch:
@annagav can you take a look at this?

PR #4204, opened 2018-12-14
Grade record view

### LINKED ISSUE #4206: Learner Record: basic learner view

annagav:
Fix #4206

#### What's this PR do?
Adds grade records view for a program.

Create a passed course with a grade by running `atler_data` command with `set_past_run_to_passed`.
```### Create exam run
from courses.models import *
from exams.factories import *
from exams.models import *

course = Course.objects.filter(title='Digital Learning 300')[0]
exam_run = ExamRunFactory.create(course=course, authorized=True)
au=ExamAuthorizationFactory.create(user=user, course=course, exam_run=exam_run, status= 'success', exam_taken=True)

exam_profile= ExamProfileFactory.create(status='success', profile=user.profile)
```
```
###### create a grade
from grades.factories import *

exam_grade = ProctoredExamGradeFactory.create(user=user, course=course, exam_run=exam_run, passed=True, percentage_grade=0.9)
exam_run.date_grades_available=now_in_utc()-timedelta(days=1)
exam_run.save()
```

To see the course as passed create a `CombinedFinalGrade` by running the task: ```create_combined_final_grades.delay()```


annagav:
@pdpinch this is what I've got so far: 
screen shot 2018-12-17 at 4 41 07 pm

annagav:
screen shot 2018-12-18 at 5 03 36 pm

annagav:
screen shot 2018-12-27 at 3 48 49 pm

certificates/views.py
```diff
@@ -100,3 +103,48 @@ def get_context_data(self, **kwargs):
...
+        }
+        context["js_settings_json"] = json.dumps(js_settings)
+
+        enrollment = ProgramEnrollment.objects.get(hash=kwargs.get('record_hash'))
+        user = enrollment.user
+        courses = enrollment.program.course_set.all()
+        mmtrack = get_mmtrack(user, enrollment.program)
+        context["program_title"] = enrollment.program.title
+        context["program_status"] = "completed" if MicromastersProgramCertificate.objects.filter(
+            user=user, program=enrollment.program) else "partially"
```
noisecapella: We should use `.exists()` here at the end of the queryset so that the query is specifically for whether the rows exist. https://docs.djangoproject.com/en/2.1/ref/models/querysets/#s-when-querysets-are-evaluated

certificates/views.py
```diff
@@ -100,3 +103,48 @@ def get_context_data(self, **kwargs):
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        js_settings = {
+            "gaTrackingID": settings.GA_TRACKING_ID,
+            "reactGaDebug": settings.REACT_GA_DEBUG,
+            "edx_base_url": settings.EDXORG_BASE_URL,
+        }
+        context["js_settings_json"] = json.dumps(js_settings)
+
+        enrollment = ProgramEnrollment.objects.get(hash=kwargs.get('record_hash'))
```
noisecapella: Can we use `get_object_or_404` here?

noisecapella:
Do you know if there's a mobile design for this, or are we just considering the desktop design here?

pdpinch:
In response to @noisecapella's comments:

- Can we just copy edX's CSS for this page?
- We do need 'date earned' or 'verified attempts'. i think 'date earned' should be the date the course certificate was created and 'verified attempts' should be a count of the proctored exam grades. (each grade corresponds to an attempt on the exam)
- I agree the 'passed/failed' labelling is confusing, but that's consistent with edX's implementation and they have consulted with the institutions that use this data.

annagav:
screen shot 2019-01-04 at 4 34 51 pm

dashboard/models.py
```diff
@@ -217,10 +219,19 @@ class ProgramEnrollment(Model):
     """
     user = ForeignKey(User, on_delete=CASCADE)
     program = ForeignKey(Program, on_delete=CASCADE)
+    hash = CharField(max_length=32, null=False, unique=True)
```
pdpinch: Can we talk about this? Is this the hash that you use for building the URL to the learner record? How will this work for non-FA programs?

dashboard/utils.py
```diff
@@ -329,23 +336,22 @@ def get_overall_final_grade_for_course(self, course):
...
            str: the overall final grade
         """
-        if settings.FEATURES.get('USE_COMBINED_FINAL_GRADE', False):
+        best_grade = self.get_best_final_grade_for_course(course)
+        if best_grade is None:
+            return ""
+        if not course.has_exam:
+            return str(round(best_grade.grade_percent))
 
+        if settings.FEATURES.get('USE_COMBINED_FINAL_GRADE', False):
```
pdpinch: This may be too late to be helpful, but I was just thinking that we could refactor out this `USE_COMBINED_FINAL_GRADE` feature flag. I don't think we still need to support have that set to False.

ui/urls.py
```diff
@@ -34,6 +34,8 @@
     url(r'^certificate/course/(?P<certificate_hash>[-\w.]+)?/?', CourseCertificateView.as_view(), name='certificate'),
     url(r'^certificate/program/(?P<certificate_hash>[-\w.]+)?/?', ProgramCertificateView.as_view(),
         name='program-certificate'),
+    url(r'^grade_records/(?P<record_hash>[-\w.]+)?/?', GradeRecordView.as_view(),
+        name='grade_records'),
```
pdpinch: Can we use a URL pattern that's more similar to edX's? I think they do something like /records/programs/shared/:hashid/ 

I don't know what the significant of the "shared" part of the path is. And I guess they must include the program id in the hash somehow -- which I guess we're doing as well since the hash is unique to a particular enrollment.

noisecapella:
@annagav or @pdpinch do you have a link to the design mockup? I can't find it

dashboard/utils.py
```diff
@@ -329,23 +336,22 @@ def get_overall_final_grade_for_course(self, course):
         Returns:
            str: the overall final grade
         """
-        if settings.FEATURES.get('USE_COMBINED_FINAL_GRADE', False):
+        best_grade = self.get_best_final_grade_for_course(course)
+        if best_grade is None:
+            return ""
+        if not course.has_exam:
+            return str(round(best_grade.grade_percent))
```
noisecapella: Can you add a test covering this case?

certificates/views.py
```diff
@@ -100,3 +104,52 @@ def get_context_data(self, **kwargs):
...
+
+
+class GradeRecordView(TemplateView):
+    """
+    View for grade records
+    """
+
+    template_name = 'grade_record.html'
+
+    def get_context_data(self, **kwargs):
```
noisecapella: Can we add a test for this to assert the context is set properly?

dashboard/utils.py
noisecapella: The code in the `else` is also untested, but it seems like we can just remove it according to this comment. We should also look for other uses of `USE_COMBINED_FINAL_GRADE` and deal with them appropriately

noisecapella:
UI looks great now

certificates/views.py
```diff
@@ -100,3 +103,48 @@ def get_context_data(self, **kwargs):
         context['certificate'] = certificate
 
         return context
+
+
+class GradeRecordView(TemplateView):
```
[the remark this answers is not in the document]
annagav: yes, I was going by certificate page as an example

annagav:
I added `FEATURE_PROGRAM_RECORD_LINK` flag.

dashboard/utils.py
noisecapella: @annagav Did you have a chance to look at this?

certificates/views_test.py
```diff
@@ -80,3 +81,22 @@ def test_valid_program_certificate_200(client):
         resp.context_data
     )
     assert reverse('program-certificate', args=[certificate.hash]) in resp.content.decode('utf-8')
+
+
+def test_program_record(client):
+    """Test"""
```
noisecapella: Can you add to this docstring?

static/js/components/ProgressWidget.js
```diff
@@ -61,6 +62,19 @@ export const circularProgressWidget = (
...
 
+export const gradeRecordsLink = (url: string): React$Element<*> => {
+  return (
+    <a
+      className="mm-minor-action"
+      href={url}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      View Grade Records
```
pdpinch: Let's change this to "View Program Record" to be consistent with edX. 

I'm still thinking about the paragraph of text that edX's program dashboard has. I don't really want to add it, but maybe we will have to.

annagav:
@noisecapella I addressed all the comments.

PR #4225, opened 2019-01-14
add support for tables in program tab page

### LINKED ISSUE #4194: Program Page: support for tables
As a member of a course team, I'd like to be able to add and edit tables on my program tab page (e.g. https://micromasters.mit.edu/dedp/pathways-masters/), and have it display as HTML. This will help accessibility, SEO and mobile rendering, and also make the content easier to maintain. 

The default table styling should be similar to [this graphic](https://do5zh7b0lqdye.cloudfront.net/images/schedule.width-800.png) embedded in the [POM program schedule page](https://micromasters.mit.edu/pom/program-schedule/) (_not_ the timeline/calendar).

Wagtail supports a [table block module](http://docs.wagtail.io/en/v2.3/reference/contrib/table_block.html). We may need to change the field type of the content field in the program tab page type to a streamfield to use this. 

At this time, we are expecting there to be text and links in the table.

tasawernawaz:
Fixes #4194 

#### What's this PR do?
Add support for tables on program tab pages.

Create program tab page, you should see adding table after content.

screen shot 2019-01-14 at 7 24 42 pm

#### Note:
@pdpinch This PR includes code for adding separate table field at the end of content (ready for review). However, if we want embedded table in content, we have to migrate content field from [richtext to stream field.](http://docs.wagtail.io/en/v2.0/topics/streamfield.html#streamfield-migrating-richtext)


pdpinch:
I spoke to the key user from the MIcroMasters team and, unfortunately, this won't work. The table needs to go in the middle of the existing content (after a page title and an image). 

I think we'll have to do the migration to a streamfield. However, let's limit it to just the "Program Tab" page type.

PR #4227, opened 2019-01-15
Tasawer/support for tables on programs page

### LINKED ISSUE #4194: Program Page: support for tables

tasawernawaz:
Fixes #4194 

#### What's this PR do?

- Add support for tables on program tab pages.
- Data migration added to convert current rich text data into streamfield.

**Critical to test (because this will effect existing tab pages in production):**
- While you are on master, create few published and drafts (if you don't have already) of tab page.
- Move to this branch.
- Run migrations.
- Test if already created tab page still rendering fine.
- Details about data migration can be found here https://docs.wagtail.io/en/v2.3/topics/streamfield.html#migrations

**Other things to test**
- Also test that while creating program tab page, you should see adding text/table under content. (While creating table select row headers checkbox)
- Test rendering of newly created tab page with table.

#### Screenshots (With table embedded between the text)

screen shot 2019-01-15 at 5 13 19 pm


pdpinch:
Naive question: will the migrations run automatically, like they normally do for django migrations? Or will we need manual intervention from @mitodl/devops ?

@tasawernawaz since Anna has approved, you can merge this PR into master. Once you do that, we should get another Travis check and then Heroku will do a build of the master branch on micromasters-ci.

We should probably test out the migration there before thinking of building a release candidate and testing on micromastesr-rc.

tasawernawaz:
@annagav thanks, 
@pdpinch I believe so, it should run automatically as other djnago migrations work. However, I don't know that how many records we have in production in this table. It could take time based on the number of records there. 
@mitodl/devops your input will be useful here.

PR #4238, opened 2019-01-21
update Readme to reference edX docker devstack

### LINKED ISSUE #4178: Update README for using docker instead of vagrant for edx-platform
Our current README file suggests that we should use vagrant instead of docker, which is outdated now. As we have [updated guide](https://github.com/mitodl/handbook/blob/master/edx-platform-development.md#odl-edx-devstack-development-guide) for setting up edx platform using docker.

pdpinch:

fixes #4178 

#### What's this PR do?

Updates the ReadMe to reference edX docker devstack instead of Vagrant devstack. 

I wasn't able to find a wholly satisfying solution to bridging the MicroMasters docker network to the edX devstack docker network. 

Also, I believe the instructions now will work for Linux and for OSX. They might be simplified even further if we can find a setting for `EDX_BASE_URL` that works for both. 

Follow the instructions. I tested this on MacOS. They should probably be tested by someone using MacOS and someone using Linux. If it's possible to standardize on one set of instructions that works for both, that would be desirable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,7 @@ need to use Django admin for the next step anyway.
 - **In Django admin**
 
     Run the server (discussed in step 2) and navigate to Django admin
-### change this URL
-    (eg: http://192.168.33.10:8000/admin). In the **Authentication and Authorization**
+    (eg: http://localhost:18000/admin/). In the **Authentication and Authorization**
     section, select a **User**, or add one then select it. In the **Permissions**
     section, check the **Superuser status** box and save.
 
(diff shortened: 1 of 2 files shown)


tasawernawaz:
@pdpinch After singing in local edx-platform, I am seeing following error.

screen shot 2019-01-30 at 2 31 52 pm

tasawernawaz:
@pdpinch Yes, I did. Previously I was logging in using `https://micromasters.d.mitx.mit.edu/`. Never tried with vagrant though.

pdpinch:
I think I might have skipped a step. Can you try to connect to the following, in a web browser while you have devstack and micromasters up and running?

1. http://edx.devstack.lms:18000
2. http://localhost:18000
3. http://127.0.0.1:18000

If one of these works, try changing `EDXORG_BASE_URL` in your MicroMasters .env file to it, and see if that works? 

For #1 to work above, you would need to change your /etc/hosts file on your Mac, but I'm not sure if that's necessary. I'd like to keep this to minimal set of instructions.

tasawernawaz:
> I think I might have skipped a step. Can you try to connect to the following, in a web browser while you have devstack and micromasters up and running?
> 
> 1. http://edx.devstack.lms:18000
> 2. http://localhost:18000
> 3. http://127.0.0.1:18000
> 
> If one of these works, try changing `EDXORG_BASE_URL` in your MicroMasters .env file to it, and see if that works?
> 
> For #1 to work above, you would need to change your /etc/hosts file on your Mac, but I'm not sure if that's necessary. I'd like to keep this to minimal set of instructions.

I have already updated my host file and I can use all three urls to login in edx-platform.

README.md
```diff
@@ -26,40 +26,28 @@ likely need to run it locally for other projects in the future.
...
- [several different components](http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/architecture.html#overview);
- MicroMaster's only depends on LMS.
+    # In your local edx_devstack/ directory, checkout the master branch of devstack
+    git checkout master
+    # assuming you want to run the master branch of edX, check it out
+    make dev.checkout
+    # update devstack, just in case:
+    make down
+    make pull
+    make dev.up
```
gsidebo: We shouldn't be repeating the edX instructions here. We should get rid of this section entirely and rename section 1 to `Install and run edX`

pdpinch:
@ahmed-arbisoft does your PR cover all of this? Should we close this PR?

pdpinch:
replaced by #4572

PR #4249, opened 2019-02-06
show signup dialog on start

### LINKED ISSUE #4248: When login is required, display the login dialog
As a user, when I access a page that requires login, I find it confusing to get directed to the MicroMasters home page. 

There are two possible UX solutions. The preferred solution is that the user is directed to a route that displays the home page with the login dialog display. The "next" parameter needs to be passed effectively through the edX login flow for this to work. 

A less desirable solution is to direct the user immediately to the edX log in page. Again, the "next" parameter needs to be passed, but it only needs to make it through one level of redirect. 

No matter what the solution is, it also has to work with users logging in from open-discussions. 

### related issue
- #4190 next parameter doesn't work if user is already logged in

tasawernawaz:
Fixes #4248 

#### What's this PR do?
Show login dialog at the start, if user is not logged in.

- Logout from micromaster
- Hit any of the following URLs, you will see login/signup popup.
  - http://0.0.0.0:8079/dashboard
  - http://0.0.0.0:8079/learners
- Signup will not be shown on home page 
  - https://0.0.0.0:8079

### DIFF
--- a/static/js/entry/public.js
+++ b/static/js/entry/public.js
@@ -83,6 +83,11 @@ const openDialog = () => store.dispatch(setDialogVisibility(true))
 
 const nodes = [...document.querySelectorAll(".open-signup-dialog")]
 
+const url = new URL(window.location.href)
+if (url.searchParams.get("next")) {
+  openDialog()
+}
+
 nodes.forEach(signUpButton => {
   signUpButton.onclick = openDialog
 })


annagav:
@tasawernawaz @pdpinch I don't know if I missunderstood the issue, but I don't think we want the sign up dialog to appear when hitting the homepage ` https://0.0.0.0:8079`.

pdpinch:
The dialog should only be appearing on routes that require login, i.e. /dashboard, /learners and maybe a couple of others. 

Definitely not the home page, and any routes served by wagtail.

tasawernawaz:
@annagav thanks for your review. @pdpinch I will update the PR.

tasawernawaz:
@annagav I am looking for the workaround here, I don't think its good idea to use `window.location.href` here, is there any other workaround given our codebase to differentiate that we are at home page or not ?

pdpinch:
Can I propose a simpler solution? The issue is that users who need to login are routed to /?next=<url> — why don’t we just display the pop up when that parameter is present?

tasawernawaz:
@pdpinch I have updated code according to your suggestion.

annagav:
@tasawernawaz some tests are failing

tasawernawaz:
> @tasawernawaz some tests are failing

Fix tests

PR #4253, opened 2019-02-12
fix `next` parameter issue for /discussions

### LINKED ISSUE #4190: next parameter doesn't work if user already logged in
### steps to reproduce
1. Log out of discussions, micromasters, and edx
2. navigate to `/login/edxorg/?next=/discussions/` in micromasters
3. entering credentials in edx
  you should be redirected to discussions (via micromasters)
4. Now, with active logged-in sessions, navigate _again_ to  `/login/edxorg/?next=/discussions/` in micromasters

### expected behavior
- redirected to discussions (same as the result of step 3 above)

### actual behavior
- redirected to your dashboard on micromasters (`/dashboard` for learners and `/learners` for staff)

It appears the `?next=` parameter is being ignored.

tasawernawaz:
fixes #4190 

#### What's this PR do?
User will redirect to correct `next` url even if they are already logged in to MM.

* Log out of discussions, micromasters, and edx
* navigate to `/login/edxorg/?next=/discussions/` in micromasters
* entering credentials in edx
  you should be redirected to discussions (via micromasters)
* Now, with active logged-in sessions, navigate _again_ to  `/login/edxorg/?next=/discussions/` in micromasters, again you should redirected to `/discussion`

### DIFF
--- a/backends/views.py
+++ b/backends/views.py
@@ -18,12 +18,14 @@ def complete(request, *args, **kwargs):
     # social_core can get confused on which User should get the SocialAuth object.
     if request.user.is_authenticated:
         key = "{}_state".format(request.backend.name)
+        redirect_url = request.session.get("next")
         backend_state = request.session.get(key)
         logout(request)
         # logout will clear the session, this preserves the backend session state. We need to do
         # this so that this workflow will validate correctly (EdxOrgOAuth2.validate_state).
         # key is the same one used in EdxOrgAuth2.get_session_state().
         request.session[key] = backend_state
+        request.session["next"] = redirect_url
 
     # Continue with social_core pipeline
     social_complete_rtn = social_complete(request, *args, **kwargs)
--- a/backends/pipeline_api.py
+++ b/backends/pipeline_api.py
@@ -45,9 +45,10 @@ def update_profile_from_edx(backend, user, response, is_new, *args, **kwargs):
     else:
         next_relative_url = "/dashboard"
 
-    next_url = backend.strategy.session.get('next')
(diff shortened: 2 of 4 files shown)


tasawernawaz:
@rhysyngsun Your guess was pretty close about change in pipeline, but along with that I have to update complete method as well. Please take a look at the solution. 

Thank you for your help.

backends/views.py
```diff
@@ -18,12 +18,14 @@ def complete(request, *args, **kwargs):
...
     if request.user.is_authenticated:
         key = "{}_state".format(request.backend.name)
+        redirect_url = request.session.get("next")
         backend_state = request.session.get(key)
         logout(request)
         # logout will clear the session, this preserves the backend session state. We need to do
         # this so that this workflow will validate correctly (EdxOrgOAuth2.validate_state).
         # key is the same one used in EdxOrgAuth2.get_session_state().
         request.session[key] = backend_state
+        request.session["next"] = redirect_url
```
rhysyngsun: Can you add a test to make sure that these two session keys are being copied over?

tasawernawaz: @rhysyngsun at this point, I am not sure how we can test this, however I have already added a [test](https://github.com/mitodl/micromasters/pull/4253/files#diff-58c973bba17544e492d75329ea514f14R300) for testing `next` parameter is preserved in `request.session` when user is already logged in.

rhysyngsun: I only see a test for the changes to `pipeline_api.py` and not one for these changes. There is an existing test for functionality in this method in `backends/views_test.py` so that should give you an idea of how to test this.

PR #4273, opened 2019-03-07
prioritize the syncing of channel memberships

### LINKED ISSUE #4197: prioritize the syncing of channel memberships
When we create a new channel, we create PercolateQueryMemberships for every user with a profile, currently on the order of 50k. It takes us a long time to run through all of these. 

We should prioritize the processing of these memberships to:
1. The ones that were more recently updated and
2. The ones that add a user to channel

Presumably this means ordering this query: 

https://github.com/mitodl/micromasters/blob/f675e338433e33d0865c3016e5dfea4bb88c01ec/discussions/api.py#L316-L320

by `updated_on` (most recent first) and by `is_member` (True before False)

tasawernawaz:
Fixes #4197 

#### What's this PR do?
Add `order_by` clause in query.

Test should pass.

### DIFF
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -313,11 +313,12 @@ def get_membership_ids_needing_sync():
     Returns:
         QuerySet: query for the list of database ids for memberships that need to be synced
     """
+    # Order by is_member (True before False) and updated_on (most recent first)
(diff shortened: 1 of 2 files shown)


tasawernawaz:
FYI @pdpinch

PR #4275, opened 2019-03-08
email mm admin if memberships is not been synced for more than 24 hours

### LINKED ISSUE #4166: alert when learners haven't been synced
We discovered an issue where learners' memberships in discussion channels weren't being synced to the discussions app. As an admin I'd like to be alerted when this membership sync has been delayed for too long, so we can take action. 

I think an alert on any memberships that haven't been synced after 24 hours would be sufficient notice. The alert could be an exception raised to Sentry, or an email (or something else). 

The task that synchronizes memberships is:
https://github.com/mitodl/micromasters/blob/master/discussions/tasks.py#L90-L102

Most of the work is done here:
https://github.com/mitodl/micromasters/blob/master/discussions/api.py#L323

If the PercolateQueryMembership is stale -- i.e. `needs_update` is True and it was created/modified over 24 hours ago -- I want to get an alert to investigate. In most cases, I'll convince myself that the user is an edge case, and I'll (manually) reset `needs_update` to False.

tasawernawaz:
Fixes #4166 

#### What's this PR do?
add code for alerting MM admin through email about stale memberships.

Test should pass.

### DIFF
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -316,11 +316,19 @@ def get_membership_ids_needing_sync():
         QuerySet: query for the list of database ids for memberships that need to be synced
     """
     # Order by is_member (True before False) and updated_on (most recent first)
-    return PercolateQueryMembership.objects.filter(
+    memberships_quest = PercolateQueryMembership.objects.filter(
         query__source_type=PercolateQuery.DISCUSSION_CHANNEL_TYPE,
         user__profile__isnull=False,
         needs_update=True
     ).order_by('-is_member', '-updated_on').values_list("id", flat=True)
+
+    # if membership is created/modified older then 24 hours, mark it as stale
+    stale_memberships = memberships_quest.filter(updated_on__lt=(datetime.now(tz=timezone.utc) - timedelta(hours=24)))
+
+    if stale_memberships:
+        email_mm_admin_about_stale_memberships(stale_memberships.values_list("id", flat=True))
+
+    return memberships_quest.values_list("id", flat=True)
 
 
 def sync_channel_memberships(membership_ids):
@@ -334,7 +342,7 @@ def sync_channel_memberships(membership_ids):
         channel_program.channel_id: channel_program.program_id
         for channel_program in ChannelProgram.objects.all()
     }
-    stale_memberships = []
+
     for membership_id in membership_ids:
         with transaction.atomic():
             membership = (
@@ -344,10 +352,6 @@ def sync_channel_memberships(membership_ids):
                 .select_for_update()
                 .first()
             )
-
-            # if membership is created/modified older then 24 hours, mark it as stale
-            if membership.updated_on < (datetime.now(tz=timezone.utc) - timedelta(hours=24)):
-                stale_memberships.append(membership_id)
 
             channel = membership.query.channels.first()
 
@@ -381,8 +385,6 @@ def sync_channel_memberships(membership_ids):
                 except DiscussionSyncException:
                     log.exception("Error updating channel membership")
 
-    if stale_memberships:
-        email_mm_admin_about_stale_memberships(stale_memberships)
 
 def add_channel(
         original_search, title, name, description, channel_type, program_id, creator_id,
--- a/discussions/api.py
+++ b/discussions/api.py
@@ -1,6 +1,5 @@
 """API for open discussions integration"""
 import logging
-from datetime import datetime, timedelta, timezone
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


discussions/api.py
```diff
@@ -314,12 +316,20 @@ def get_membership_ids_needing_sync():
         QuerySet: query for the list of database ids for memberships that need to be synced
     """
     # Order by is_member (True before False) and updated_on (most recent first)
-    return PercolateQueryMembership.objects.filter(
+    memberships_quest = PercolateQueryMembership.objects.filter(
```
pdpinch: `memberships_quest `? Did you mean `memberships_query` ?

discussions/api.py
```diff
@@ -314,12 +316,20 @@ def get_membership_ids_needing_sync():
...
         query__source_type=PercolateQuery.DISCUSSION_CHANNEL_TYPE,
         user__profile__isnull=False,
         needs_update=True
     ).order_by('-is_member', '-updated_on').values_list("id", flat=True)
```
pdpinch: If you want the query, I think you need to remove the `.values_list("id", flat=True)`

pdpinch:
@rhysyngsun is this kind of what you hand in mind, to notify on stale memberships? 

I'd be OK with raising it to sentry instead of an email, if that's simpler.

discussions/api.py
```diff
@@ -314,12 +316,20 @@ def get_membership_ids_needing_sync():
...
 
+    # if membership is created/modified older then 24 hours, mark it as stale
+    stale_memberships = memberships_quest.filter(updated_on__lt=(datetime.now(tz=timezone.utc) - timedelta(hours=24)))
+
+    if stale_memberships:
+        email_mm_admin_about_stale_memberships(stale_memberships.values_list("id", flat=True))
```
rhysyngsun: We shouldn't be doing this in this method because it's called by a task that's called as often as every minute, so this will quickly get annoying. Also, a failure to send the email would have the side affect of causing an error in any operation that calls this, so it's also not ideal in that regard.

This should at a minimum be a separate celery task that performs the check and sends the email to isolate failure cases of that from this code. It might also be feasible to accomplish this with a Redash alert depending on what @pdpinch needs.

discussions/utils.py
```diff
@@ -60,3 +63,29 @@ def get_token_for_request(request, force_create=False):
...
+        membership_ids(List): List of ids
+    """
+    if getattr(settings, 'ADMIN_EMAIL', None) is None:
+        raise ImproperlyConfigured('Setting ADMIN_EMAIL is not set')
+
+    subject = "List of stale memberships"
+    body = (
+        "Hi,\n"
+        "The following memberships were stale in the PercolateQueryMembership:\n\n"
+        "{membership_ids}"
```
rhysyngsun: This list could be large and I don't think the specific ids are useful. I think it's probably more useful to inform about the scope of the issue, which would probably be a per-channel breakdown of how many recipients haven't been synced.

discussions/utils.py
```diff
@@ -60,3 +63,29 @@ def get_token_for_request(request, force_create=False):
...
+    """
+    Email list of stale memberships to mm admin
+
+    Args:
+        membership_ids(List): List of ids
+    """
+    if getattr(settings, 'ADMIN_EMAIL', None) is None:
+        raise ImproperlyConfigured('Setting ADMIN_EMAIL is not set')
+
+    subject = "List of stale memberships"
```
rhysyngsun: Suggest changing this to "Alert: {num_not_synced} MicroMasters channel memberships not synced" or something like that so the recipient knows what the email is about at a glance

tasawernawaz:
Thank you @pdpinch  and @rhysyngsun for your suggestions, I will update the code and let you know.

discussions/utils.py
tasawernawaz: is there any field like channel `name/title` ?

discussions/api.py
tasawernawaz: @pdpinch Thoughts about alerting medium ?

@rhysyngsun Are we using Redash already in project ?

discussions/utils.py
rhysyngsun: `Channel` has an FK to `PercolateQuery`: https://github.com/mitodl/micromasters/blob/master/discussions/models.py#L29

pdpinch:
Thanks for the effort Tasawer, but I've decided this is better handled with a redash query and alert.

PR #4295, opened 2019-04-05
adding a management command to authorize users for expired exam runs

### LINKED ISSUE #3914: invitation-only exam runs
As a course team member, I would like to be able to set up an exam run for which learners must be authorized manually instead of automatically after having paid & passed a course. This would allow me to schedule "pop-up" proctored exams in regions where there is no Pearson exam center.

annagav:
Fix #3914

#### What's this PR do?
Adds a management command to authorize users for expired exam runs.

Create an exam run with the scheduling dates in the past, and current eligibility dates.
Run the command `./manage.py create_invitational_exam_runs <exam_run.id> <username>`


pdpinch:
I believe I mean eligibility, i.e. the dates the user can take the exam. 

Here’s a practical example:

Learners can sit for the 1T2019 exams between 5/1 and 5/15. We need to be manually authorize some leaners to sit for the exam 5/16-5/31. There’s more details on https://odl.zendesk.com/agent/tickets/26699

annagav:
@pdpinch Are the learners supposed to be able to see the exam run on the dashboard? In this comment it looks like we were specifically interested in exam runs that are expired: https://github.com/mitodl/micromasters/issues/3914#issuecomment-375663543

pdpinch:
No. We'd like this to be an invisible exam run, except for the blue "Schedule an exam" button for learners who are authorized. In all of these "ad-hoc" cases, the learner will be in contact with customer support, so they will know when they are authorized and that they should look for the button.

annagav:
The massages in the dashboard depend on the eligibility window of the exam run:
Screen Shot 2019-04-10 at 5 03 34 PM

annagav:
The `Final Proctored Exam` is not going to show schedulable exam runs unless the exam las legibility date is in the future. 
Screen Shot 2019-04-17 at 5 01 02 PM

pdpinch:
Thanks @annagav -- can you link me to the code that determines the display of the "schedule an exam" button? I thought it was based only on whether the user was authorized. I did get this to work once, using the code

pdpinch:
Maybe this is it: 

https://github.com/mitodl/micromasters/blob/1646d9fb0b9897be0b028aa7a2435a5eb4ba0021/dashboard/api.py#L496-L505

So I think I may have had the requirement for the exam run backwards. The schedulable dates will be in the past, but the eligible dates still need to be in the future. This prevents any users from self-scheduling in the exam run, but we can still manually schedule a date.

annagav:
This is used to compute the status for the exam card:
https://github.com/mitodl/micromasters/blob/1646d9fb0b9897be0b028aa7a2435a5eb4ba0021/dashboard/utils.py#L425

annagav:
@pdpinch When we create the authorization do we want to set the status to `success` or ` pending`?

pdpinch:
`pending` -- that way the authorization will get sent to Pearson. When they reply it will get updated to `success`

That might be why you're not seeing the button in your tests. @rhysyngsun may have some suggestions about how to short-circuit that for testing.

rhysyngsun:
@pdpinch @annagav https://github.com/mitodl/micromasters/blob/master/exams/management/commands/simulate_sftp_responses.py can be used to generate the responses, although I don't think it's been touched since I first wrote it almost 2 years ago so there may have been some drift between what that generates and what our code currently expects.

annagav:
@pdpinch I was able to see the button. So this is ready for review.

pdpinch:
Hm. I should have asked for a confirmation message.

PR #4335, opened 2019-06-06
Upgrading css-loader to get rid of js-yaml@3.7.0

annagav:
Fixing vulnerable dependency: `js-yaml`

#### What's this PR do?
Updates `css-loader` that was requiring js-yaml 3.7.0.

Everything should work!


aliceriot:
it looks like there's a flow error, I think you just need to add an appropriate ignore statement to `.flowconfig`

Issue #4251, opened 2019-02-07
Electives for FA courses


pdpinch:
In September 2019, DEDP would like to offer a new course 14.75x. The three courses 14.73x, 14.740x and 14.75x would form an elective group, with learners required to pass (at least) two of the three in order to complete the program. 

The MicroMasters team would like us to add the new 14.750x course to the dashboard by end of July 2019. MVP requires changes to program logic, but we probably won't have time to make significant changes to the UX of the learner dashboard. 

### MVP acceptance criteria
- [x] develop a data model for managing the configuration of electives (possibly some kind of set?) 
- [x] update the progress card on the learner dashboard to handle the elective properly
- [x] update the program credential creation logic to reflect the fact the the learner is required to pass only two of the three electives

### out of scope for MVP
- add a notice to the learner dashboard alerting learners to the fact that the two courses are elective (text and design TBD) 

### related issues (nice to have)
- only display courses in the learner record that the learner has attempted (diverging from edX current practice of showing all courses in the program) 

### questions
*Q:* How/does this affect program coupons?

pdpinch:
Updated the description with new dates and definition of MVP

pdpinch:
We can call this done.

PR #4347, opened 2019-07-11
Added RFC for electives in FA programs

annagav:
Related to #4251

#### What's this PR do?
Describes roughly the feature and some architectural changes.

Read the draft and propose changes.

Here it is [rendered](https://github.com/mitodl/micromasters/blob/523373c44ca6b5f699ba1c404a084e7a876f2e6c/docs/rfcs/0002-electives-for-fa-program.md)

### DIFF
--- a/docs/rfcs/0002-electives-for-fa-program.md
+++ b/docs/rfcs/0002-electives-for-fa-program.md
@@ -1,0 +1,40 @@
+## Electives for FA Programs
+
+#### Abstract
+This feature will allow the program team to offer learners to take a set of courses in the program as electives.
+A program will optionally have a set of electives. A set of electives will be defined by a number (a set) of courses,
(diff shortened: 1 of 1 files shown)


docs/rfcs/0002-electives-for-fa-program.md
```diff
@@ -0,0 +1,40 @@
...
+`ElectivesSet`
+- program
+- title
+- num_required_courses
+
+`ElectiveCourse`
+- course
+- electives_set
+
+Maybe these should be available to be set in Wagtail.
```
pdpinch: No, but we should be able to create and edit elective groups in the django admin.

pdpinch:
@gsidebo @noisecapella can you take a look at this?

Note that the data models have already merged in #4349

rhysyngsun:
@annagav @pdpinch I think we can just close this since the issue itself is closed?

PR #4349, opened 2019-07-16
Adding models for course electives

annagav:
Related to #4251

#### What's this PR do?
Adds two new models to support course electives in programs, modifies Program model, updates `generate_program_certificate`.

Run migrations. In the django admin create `ElectiveCourse`'s, `ElectivesSet` for a program you are testing with.
Create a course certificate that would satisfy the elective requirement. And for other courses that are not electives in the program. Run management command `generate_program_certificates`.


courses/models.py
```diff
@@ -312,3 +312,35 @@ def get_freezable(cls):
...
+    Links to a course to an ElectivesSet
+    """
+    course = models.OneToOneField(Course, on_delete=models.CASCADE)
+    electives_set = models.ForeignKey(ElectivesSet, on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('course', 'electives_set',)
+
+    def __str__(self):
+        return 'Course elective for course "{course_title}"'.format(
```
pdpinch: I think this is a little clearer: 

```suggestion
        return 'Elective course "{course_title}" in electives set "{electives_set}"'.format(
```

PR #4316, opened 2019-04-29
Remove authentication from courseruns endpoint, and update test

sesterki:
#### What are the relevant tickets?
#4311 

#### What's this PR do?
Remove the authentication requirements for the courseruns endpoint.

#### How should this be manually tested?
Manually navigate to the endpoint.

PR #4312, opened 2019-04-25
Adds viewset for courseruns API, required serializer, and related tests

sesterki:
#### What are the relevant tickets?
#4311 

#### What's this PR do?
This PR adds a course_run api endpoint, a new serializer, and associated tests to the courses app.

#### How should this be manually tested?
The new endpoint can be hit manually (`api/v0/courseruns/`).

PR #4380, opened 2019-08-07
Add catalog API for discussions

### LINKED ISSUE #4353: Create endpoint to query for program metadata
Similar to #4311 but for programs

rhysyngsun:
Closes #4353

#### What's this PR do?
Adds an endpoint to get catalog data about programs, courses, and course runs. Right now the data is fairly basic on courses and course runs because the details of those will all come from MITx on edX and this data will get merged with that.

Wrote this as a new API since none of the existing ones do quite what we need or were more user-oriented and I didn't want to fuss with making those safe and complete enough for public consumption.

Go to /api/v0/catalog/ and verify the data looks reasonable


pdpinch:
Will this replace the courseruns API?

rhysyngsun:
@pdpinch I'm not sure if it necessarily replaces anything, but none of our existing APIs fit squarely into what we needed on the discussions side, mostly because where we do generate such data it's returned statically on the dashboard view in the html. A lot of those APIs also vary the data they return dependent on the user, so I didn't want to meddle with those for this change and trigger a large set of regression testing.

I think going forward though, if the frontend is display program/course/course run data in a general way, this API would definitely be usable.

courses/views.py
```diff
@@ -124,3 +126,11 @@ class CourseRunViewSet(viewsets.ReadOnlyModelViewSet):
     """API for the CourseRun model"""
     serializer_class = CourseRunSerializer
     queryset = CourseRun.objects.all()
+
+
+class CatalogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """API for program/course catalog list"""
+    serializer_class = CatalogProgramSerializer
+    queryset = Program.objects.filter(live=True).prefetch_related(
+        "course_set__courserun_set", "programpage__thumbnail"
```
annagav: I think this is a wrong field name. This should be `programpage__thumbnail_image`.

pdpinch:
To follow up on my earlier question, in #4312 and #4316 we added a courseruns API, specifically for feeding courserun metadata to open-discussions. As far as I know, it was never used for anything else. 

If we are changing/removing the code in open-discussions that uses coursesruns, we should plan to remove it here as well.

rhysyngsun:
@pdpinch yeah, I'd seen those but figured they were used elsewhere since they aren't really effective for what we want to do, particularly since they require calling 3 different APIs then joining all the data in-memory via some kind of references to children objects. It would overly complicate things on the discussions end when a single API that returns the entire nested datastructure is far easier to walk. I'll open an issue to remove those other APIs if we don't need them since this PR has already been approved.

rhysyngsun:
Opened #4393 for the tech debt around the course runs API

PR #4382, opened 2019-08-08
Freeze grades scheduling update

annagav:
Need several runs of the task in order to mark the process as `complete`

#### What's this PR do?
Update the celery beat schedule to run freeze grades 4 times a night.

Make  sure that the task gets scheduled.


rhysyngsun:
Could you give some more detail on the problem the problem is the change is solving?

annagav:
@rhysyngsun, the freeze grade task doesn't usually complete in one because there are authentications failures, so with old schedule it was taking days to freeze the grades and mark the process as complete. In the new schedule I am trying to speed the process up by running it a few times.

PR #4387, opened 2019-08-09
Mark courses as electives in Program Records

### LINKED ISSUE #4386: Learner Record: Display elective/core status
Now that DEDP has core and elective courses, we need to update the Learner Record to indicate "core" and "elective" courses.

image

I think we can use the same labels that we use on the dashboard.

annagav:
Fix #4386

#### What's this PR do?
Add a tag to courses in the program records view.

Create electives for a program, and go to view program records.

Screen Shot 2019-08-09 at 1 37 32 PM


annagav:
@rhysyngsun I addressed you comments, thanks!

certificates/views.py
```diff
@@ -10,6 +10,7 @@
 from rest_framework.generics import Http404
 
 from cms.models import CourseCertificateSignatories, ProgramCertificateSignatories, ProgramLetterSignatory
+from courses.models import ElectiveCourse
```
rhysyngsun: This import needs to get removed for the lints to pass

PR #4396, opened 2019-08-19
Update redux and redux-asserts

annagav:
Fix critical security alert for `lodash-es`
#### What's this PR do?
Upgrades `redux` and `redux-asserts`
nothing should break


pdpinch:
Do we need to update this for any other repos? I see OVS has a similar alert about lodash-es

Issue #4318, opened 2019-05-07
Elasticsearch module raises illegal argument exception when sending enrollment email


markbreedlove:
Sentry has logged errors from Elasticsearch with HTTP 400 responses for an illegal argument in a `percolate` query. The search is triggered by https://github.com/mitodl/micromasters/blob/e890c10a1025fd5cdb450d48aa82a3d69314d318/search/api.py#L296

The error message is `'[percolate] query is missing required [document_type] parameter'`

I'm going to link this issue to the two Sentry issues that I see at the moment.

OpsGenie alerted me yesterday, 2019-05-06 at 4:52 PM EDT and it's occurred since then.

markbreedlove:
I'm befuddled by Sentry's U.I. and can't seem to link the issue quickly.  Here are links to the Sentry issues:
[MICROMASTERS-2PJ](https://sentry.io/organizations/mit-office-of-digital-learning/issues/1014377187/?referrer=github_plugin)
Sentry issue: [MICROMASTERS-2PK](https://sentry.io/organizations/mit-office-of-digital-learning/issues/1014377199/?referrer=github_plugin)

Issue #3748, opened 2018-01-08
Upgrade to elasticsearch 6


noisecapella:
We should do #2867 first, I'm not sure if skipping major versions is a good idea

blarghmatey:
The version that we are currently running was EOL last month so we should work on this upgrade as we are able to.

rhysyngsun:
The changes are getting reverted in #4319 due to the upgrade path being more complicated (see #4318).

tasawernawaz:
@rhysyngsun anything I can do on this ? or is this waiting on Devops ?

rhysyngsun:
@tasawernawaz someone needs to determine what the proper steps/order are for upgrading the cluster and the django dependencies in a way that won't break functionality, I think as we last left it, it should be one of us who looks into this instead of devops since it comes down to what the application needs in terms of compatibility at the different stages and how we would need to step through the migration as far as application code. That probably means this needs to become an epic with child issues for both the application and devops that need to be done in a precise order.

blarghmatey:
I think the easiest first step is to determine whether the code changes can be put behind a feature flag which we can flip once we get the infrastructure in place

blarghmatey:
@markbreedlove for this we need to do a few stages of rollout. First is to synchronize the data in the existing Elasticsearch indices for micromasters production to the cluster running in the production-apps environment, followed by validating that the app can connect to the other cluster, and then running a reindex after the cutover.

There are some other details to be considered, but that is the broad context.

PR #4405, opened 2019-09-13
Upgrade to Elasticsearch 6

markbreedlove:

https://github.com/mitodl/micromasters/issues/3748
https://github.com/mitodl/salt-ops/issues/867

#### What's this PR do?

It revives the changes that were attempted in these commits:
8886a8d (Rename wildcard)
c7d5797 (Bump Elasticsearch Version)
... both of which were introduced in https://github.com/mitodl/micromasters/pull/4303 and reverted in https://github.com/mitodl/micromasters/pull/4319

In addition, it changes the Elasticsearch Index Type, per "[Schedule for Removal of Mapping Types](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/removal-of-types.html)"

We need to make sure that the error reported in https://github.com/mitodl/micromasters/issues/4318 does not reoccur. The test suite did not catch this the last time around, and it is passing now for this branch.

I could use some guidance on adding a test to cover the query at https://github.com/mitodl/micromasters/blob/e890c10a1025fd5cdb450d48aa82a3d69314d318/search/api.py#L296 ... unless someone can update this PR documenting that this is unnecessary.

I'm reading `search.indexing_api.py`, specifically the [assignment of PERCOLATE_MAPPING](https://github.com/mitodl/micromasters/blob/c4de3969c5e66e9f884729896808114b3c9eb1d4/search/indexing_api.py#L222-L232) but it's not immediately apparent how this mapping is put into place. The indices that I'm seeing in my Elasticsearch Docker instance that were created during the test suite run do not have a `query` field.

#### Releasing

I'm flagging @noisecapella on this PR because he's the Release Manager next week. We'll need to coordinate before this is released and deployed to production, because I will first need to transfer the original Elasticsearch 5 indices to the Elasticsearch 6 cluster (to the `production-apps` cluster) and modify their mapping types in the process (from "doc" to "_doc"). Someone will then need to change the production app's configuration to point at the ES 6 cluster in sync with the deployment.

I believe that someone needs to run a Celery reindexing task after the deployment to pick up documents that may have been added to the indices since they was copied.


docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       - "6379"
 
   elastic:
-    build: elastic
+    image: elasticsearch:6.7.1
     command: elasticsearch -E network.host=0.0.0.0 -E http.cors.enabled=true -E http.cors.allow-origin=* -E rest.action.multi.allow_explicit_index=false
```
noisecapella: Can you add a line right after this one:

    user: elasticsearch

PR #4418, opened 2019-10-08
Add instructors to catalog API

noisecapella:
#### What are the relevant tickets?
Fixes #4411 

#### What's this PR do?
Reads the related `FacultyMember` instances and renders the `name` in the catalog API

#### How should this be manually tested?
Go to `/api/v0/catalog/` and check that instructors are listed like they are in the issue.

PR #4420, opened 2019-10-09
Add instructors, price, start_date, end_date, and enrollment_start to catalog API

### LINKED ISSUE #4406: Add program price and program dates to the catalog API
As a catalog API consumer, like the MIT Open site, I'd like to know the MicroMasters program price, enrollment start date, start date and end date of each program.

### Acceptance Criteria:

- [ ] Add a price field to the API response, based on the product of `program.price` and `program.num_required_courses`
- [ ] Add an start_date field to the API response, using the next start date of the first course in the program
- [ ] Add an end_date field  to the API response, using the next end date of the last course in the program. 
- [ ] Add an enrollment_start date field to the API response, using the next enrollment_start date of the first course in in the program.

### Related issues 

- Similar to https://github.com/mitodl/mitxpro/issues/1129
- API response will be consumed by (issue to be created in open-discussions)

noisecapella:
Fixes #4406 
Fixes #4418 

#### What's this PR do?
Calculates instructors, total price, some date fields and renders them in the catalog API

Go to `/api/v0/catalog/` and verify that everything looks right. The `total_price` field should be a string representing a cost value with the number of courses times the price per course. The dates should be ISO-8601 formatted strings, or None if the date is null or if no course runs are unexpired.

### DIFF
--- a/courses/catalog_serializers.py
+++ b/courses/catalog_serializers.py
@@ -3,6 +3,7 @@ Serializers for courses
 """
 from rest_framework import serializers
 
+from cms.models import ProgramPage
 from courses.models import Course, Program, CourseRun
 from courses.serializers import TopicSerializer
 
@@ -34,6 +35,7 @@ class CatalogProgramSerializer(serializers.ModelSerializer):
     """Serializer for Program objects"""
     programpage_url = serializers.SerializerMethodField()
     thumbnail_url = serializers.SerializerMethodField()
+    instructors = serializers.SerializerMethodField()
 
     courses = CatalogCourseSerializer(source="course_set", many=True)
     topics = TopicSerializer(many=True)
@@ -73,6 +75,15 @@ class CatalogProgramSerializer(serializers.ModelSerializer):
         except ProgramPage.DoesNotExist:
             return None
 
+    def get_instructors(self, program):
+        """Get the list of instructors from the program page"""
+        try:
+            page = program.programpage
+        except ProgramPage.DoesNotExist:
+            return []
+
+        return list(page.faculty_members.values("name"))
+
     class Meta:
         model = Program
         fields = (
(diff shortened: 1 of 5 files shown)


courses/catalog_serializers.py
```diff
@@ -71,12 +76,66 @@ def get_thumbnail_url(self, program):
...
+        try:
+            page = program.programpage
+        except ProgramPage.DoesNotExist:
+            return []
+
+        return list(page.faculty_members.values("name"))
+
+    def get_total_price(self, program):
+        """Get the combined price of all courses"""
+        return str(program.price * program.num_required_courses)
```
mbertrand: Can this be a numeric field instead, as in xpro?

courses/catalog_serializers.py
```diff
@@ -71,12 +76,66 @@ def get_thumbnail_url(self, program):
...
+        """Get the starting date of the first course in the program"""
+        course = program.course_set.order_by("position_in_program").first()
+        if not course:
+            return None
+        first_unexpired = first_matching_item(
+            course.courserun_set.all().order_by("start_date"), lambda run: run.is_unexpired
+        )
+        if not first_unexpired:
+            return None
+        return first_unexpired.start_date.isoformat() if first_unexpired.start_date else None
```
mbertrand: this can just be:
```python
return first_unexpired.start_date
```
which will return a date string in a format like `2019-09-12T19:23:03Z`, matching the date format returned by the xPro API, instead of `2019-09-12T19:23:03+00:00`

Ditto for the other date fields

courses/catalog_serializers.py
noisecapella: The result of `program.price * program.num_required_courses` is a `Decimal`, so if we render it as a string here and then parse as a `Decimal` in javascript, it removes the risk of floating point rounding errors

courses/catalog_serializers.py
noisecapella: Sure, will do

courses/catalog_serializers.py
mbertrand: 👍

noisecapella:
@mbertrand Done

PR #4506, opened 2020-01-29
course team update v2

tasawernawaz:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
  - [ ] Tag @ferdi or @pdpinch for review
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings
  - [ ] New settings are documented and present in `app.json`
  - [ ] New settings have reasonable development defaults, if applicable

https://trello.com/c/kdktSa7B/1-update-course-team-layout

#### What's this PR do?

Some more updates to course team page.

- The title should not be a required field. It should also be a rich text field, to allow for line breaks
- bio container is larger than image container, making it impossible to read longer bios (e.g. Regina Barzilay)
- page navigation shouldn't be hard-coded and it should match the section labels defined by the CMS user

Follow the same steps at #4502 

Relates to #4502 

(Optional)


cms/blocks.py
```diff
@@ -10,7 +10,7 @@ class CourseTeamBlock(blocks.StructBlock):
     """
 
     name = blocks.CharBlock(max_length=100, help_text="Name of the course team member.")
-    title = blocks.CharBlock(max_length=200, help_text="Title of the course team member.")
+    title = blocks.RichTextBlock(blank=True, null=True, help_text="Title of the course team member.")
```
ahmed-arbisoft: I believe you should disable the features not needed in this place, specifically link, document, image, lists etc. We can stay with the bold and italic if needed. Personally I think even h2, h3 etc need to be removed from this field because they have the potential to break the UI.

tasawernawaz: updated

PR #4601, opened 2020-04-14
Rename UWSGI_THREAD_COUNT and UWSGI_PROCESS_COUNT, and remove redundant if-not-env blocks

### LINKED ISSUE #4597: Fix environment variables so we can enable uWSGI strict mode
We have some environment variables starting with `UWSGI_` which are treated by uWSGI as settings. If there is no corresponding valid uWSGI setting, an error will occur on strict mode which causes uWSGI to exit.

I think we can solve this in a couple ways, for example with `UWSGI_THREAD_COUNT`:
 - Change `UWSGI_THREAD_COUNT` to `UWSGI_THREADS`, and only set the `if-not-env` case where the variable isn't already set.
 - Change `UWSGI_THREAD_COUNT` to `MICROMASTERS_THREAD_COUNT` or another non-conflicting environment variable, and set uWSGI similarly

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #4597 

#### What's this PR do?
- Renames `UWSGI_THREAD_COUNT` to `UWSGI_THREADS` and `UWSGI_PROCESS_COUNT` to `UWSGI_PROCESSES`
- Removes `if-env` blocks because uWSGI will populate those variables automatically

First, rename `UWSGI_THREAD_COUNT` and `UWSGI_PROCESS_COUNT` if you have them defined in your `.env` file. Then go to the homepage and verify that nothing has broken.

(Optional)

(Optional)

(Optional)

(Optional)


markbreedlove:
Please see mitodl/micromasters#4597 (comment)

> I recently made this change in cookiecutter-djangoapp: mitodl/cookiecutter-djangoapp#167

> My goal there was to preserve some kind of prefix, but not have uWSGI assign any configuration directives based on the variables.

markbreedlove:
should be 👍

PR #4621, opened 2020-07-17
fix phone number issues (came up w/ Vietnamese numbers)

### LINKED ISSUE #4620: issues with phone number validation
Some users have reported that their valid phone numbers are not being accepted.

aliceriot:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

closes #4620 

#### What's this PR do?

just bumps up the phone number validation lib we're using to the latest version. We were using a fairly old version, and we had some users having issues where valid phone numbers were being marked as invalid (particularly ones from Vietnam).

I also added a regression test that exposed the issue and is fixed by the dependency upgrade.

confirm that you can get past the first screen of the profile flow with a Vietnamese phone number like '+84-03-6123-4567'. Also confirm on master that you cannot do so.

### DIFF
--- a/static/js/lib/validation/profile_test.js
+++ b/static/js/lib/validation/profile_test.js
@@ -385,15 +385,32 @@ describe("Profile validation functions", () => {
       assert.deepEqual(personalValidation(profile), {})
     })
 
-    it("should complain if you enter an invalid phone number", () => {
-      const profile = {
-        ...USER_PROFILE_RESPONSE,
-        phone_number: "+1 222"
-      }
-      const errors = {
-        phone_number: "Please enter a valid phone number"
-      }
-      assert.deepEqual(personalValidation(profile), errors)
+    //
+    ;[
+      ["+84-03-6123-4567", true],
+      ["+1-213-223-3321", true],
+      ["+1-911-223-3321", false],
+      ["+1 222", false],
+      ["asdfasdf", false],
+      ["一二三四五六", false]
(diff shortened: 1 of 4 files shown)


static/js/lib/validation/profile_test.js
```diff
@@ -385,15 +385,29 @@ describe("Profile validation functions", () => {
...
-      }
-      const errors = {
-        phone_number: "Please enter a valid phone number"
-      }
-      assert.deepEqual(personalValidation(profile), errors)
+    //
+    ;[
+      ["+84-03-6123-4567", true],
+      ["+1 222", false],
+      ["asdfasdf", false]
```
ahmed-belal, line 392: It would be great if you could add a couple more test values, some US phone numbers as well as the Vietnamese one you've already included. Doesn't add much as far as functionality is concerned, just being verbose with the testing.

ahmed-belal:
FYI you can most probably fix the python build failures by pinning `isort==4.3.21` in `test_requirements.txt`, it's a known issue.

static/js/lib/validation/profile_test.js
aliceriot: sure, I'll add a few more - I didn't add too many other than a case that exposed the issue because the external library *should* be tested...

aliceriot:
@ahmed-belal I added a few more test cases and added the fix you suggested for the python build

ahmed-belal:
Yeah feel free to merge

PR #4671, opened 2020-10-27
Update populate_exam_coupons command to parse course number

### LINKED ISSUE #4669: Update the coupon population command to parse the course number
From Ned:

> I would provide a separate sheet for each course as the codes will only be valid for one course.
> 
> The sheets will be named for the course, but you can verify the course by comparing the course number in cell G2, "Catalog Query".
>

annagav:
Fix #4669 

#### What's this PR do?
Update populate_exam_coupons command to parse course number

Run the command with the new file that has the course number in the column "Catalog Query".

### DIFF
--- a/exams/management/commands/populate_edx_exam_coupons.py
+++ b/exams/management/commands/populate_edx_exam_coupons.py
@@ -34,7 +34,7 @@ class Command(BaseCommand):
         csvfile = kwargs.get('csvfile')
         reader = csv.DictReader(csvfile.read().splitlines())
         catalog_query = next(reader)['Catalog Query']
-        course_number = re.search(r"key:\(([A-Za-z0-9.]+)", catalog_query)
+        course_number = re.search(r"key:\(([A-Za-z0-9.]+)", catalog_query).group(1)
 
         validated_urls = validate_urls(reader)
         course = Course.objects.filter(course_number__startswith=course_number).first()
--- a/exams/management/commands/populate_edx_exam_coupons.py
+++ b/exams/management/commands/populate_edx_exam_coupons.py
@@ -3,6 +3,7 @@ Populates ExamAuthorizations with edx coupon urls for taking exam-course
 """
 import csv
 import argparse
+import re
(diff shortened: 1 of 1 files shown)


exams/management/commands/populate_edx_exam_coupons.py
```diff
@@ -32,16 +33,22 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument,too-many-l
 
         csvfile = kwargs.get('csvfile')
         reader = csv.DictReader(csvfile.read().splitlines())
-        edx_course_key = next(reader)['Course ID']
+        catalog_query = next(reader)['Catalog Query']
+        course_number = re.search(r"key:\(([A-Za-z0-9.]+)", catalog_query)
```
noisecapella, line 37: This returns a `Match` object which causes an error below in the `Course.objects.filter` line

noisecapella (changes requested):

annagav:
@noisecapella could you try again?

exams/management/commands/populate_edx_exam_coupons.py
```diff
@@ -32,16 +33,22 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument,too-many-l
...
         reader = csv.DictReader(csvfile.read().splitlines())
-        edx_course_key = next(reader)['Course ID']
+        catalog_query = next(reader)['Catalog Query']
+        course_number = re.search(r"key:\(([A-Za-z0-9.]+)", catalog_query).group(1)
 
         validated_urls = validate_urls(reader)
-        course = Course.objects.get(courserun__edx_course_key=edx_course_key)
+        course = Course.objects.filter(course_number__startswith=course_number).first()
+
+        if course is None:
```
noisecapella, line 42: Can we change this to use `Course.objects.get` instead? That way there is only one course which matches.

noisecapella (changes requested):

noisecapella (approved):

PR #4678, opened 2020-11-04
Updating the regex for parsing course codes from coupon report

annagav:
none

#### What's this PR do?
The coupon report file from edx have a new pattern for exam-course code in the 'Catalog Query` column. 
This PR updates the management command for parsing the coupon codes, with a new regex to parse the course code.

Run the management command with new coupon file.

### DIFF
--- a/exams/management/commands/populate_edx_exam_coupons.py
+++ b/exams/management/commands/populate_edx_exam_coupons.py
@@ -34,7 +34,7 @@ class Command(BaseCommand):
         csvfile = kwargs.get('csvfile')
         reader = csv.DictReader(csvfile.read().splitlines())
         catalog_query = next(reader)['Catalog Query']
-        course_number = re.search(r"key:\(([A-Za-z0-9.]+)", catalog_query).group(1)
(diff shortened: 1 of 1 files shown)


exams/management/commands/populate_edx_exam_coupons.py
```diff
@@ -34,7 +34,7 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument,too-many-l
         csvfile = kwargs.get('csvfile')
         reader = csv.DictReader(csvfile.read().splitlines())
         catalog_query = next(reader)['Catalog Query']
-        course_number = re.search(r"key:\(([A-Za-z0-9.]+)", catalog_query).group(1)
+        course_number = re.search(r"\+([A-Za-z0-9.]+)PEx", catalog_query).group(1)
```
noisecapella, line 37: Do we care about the old pattern anymore?

annagav: I don't think we need the old pattern. It looks like the exam code is going to be just like edx course key, but with PEx after the number, like this `MITx+13.456PEx+3T2020`

noisecapella (approved):

PR #4693, opened 2020-11-18
Add OWASP ZAP scanning with Github action

markbreedlove:
Add [OWASP ZAP](https://www.zaproxy.org/getting-started/) ["baseline"](https://github.com/marketplace/actions/owasp-zap-baseline-scan) and ["full"](https://github.com/marketplace/actions/owasp-zap-full-scan) scans on a schedule, with Github actions.

These are run on a schedule, not in response to code commits, in order to catch vulnerabilities that are announced in between code updates. A couple of tests are ignored which we have found generate false positives too often.


blarghmatey:
:+1:

PR #4696, opened 2020-11-19
Add git ref to Github action 'uses' specifier

markbreedlove:
In the "scanning.yml" workflow, add '@ref' notation to the 'uses' properties. This is required by Github and we are getting errors without them. I'm adding '@master' to the zaproxy actions in order to get the latest vulnerability definitions without having to update this file whenever they do a new release.

The Github documentation that I read yesterday said they "strongly recommend" the `@ref` syntax, not that it's required, but apparently it is.


blarghmatey:
:+1:

PR #4711, opened 2020-11-23
Populate exam coupons when authorizing for exam run

annagav:
N/A

#### What's this PR do?
This is a modification to the process of populating edx exam coupons. Here we are assuming that we are inevitably going to run out of coupons. If there will be a shortage the course team will have request more coupons.
The reason for this change is that an ExamRun gets authorized only when it reaches the schedulable window, but we need to populate the coupons ahead of time, before authorizations happen.

Run the command ./manage.py populate_edx_exam_coupons <file.csv>
a bunch of ExamRunCoupons should get created
Try running the task `authorize_exam_runs` or wait until it runs automatically.


pdpinch:
@annagav how does this approach compare to [Plan A in my DEDP Coupon Management document](https://docs.google.com/document/d/1imqiqK-l3VPUI04fmDyLqpDCaEcVsYhD9WIR3wVyd0Y/edit#heading=h.l6ek2ll9empq)?

pdpinch:
Should we close this PR?

PR #4712, opened 2020-11-24
Let the exam coupon command run many times

annagav:
none

#### What's this PR do?
This modifies the management command to run with any size coupon files.
Also it changes the the creation of the `ExamAuthorization`(s) to set the status for `SUCCESS` since we don't consult with Pearson anymore.

Run `./manage.py populate_edx_exam_coupons <coupons-report.csv>` 
Check the dashboard to see the link to edx to register for exam in that course.

### DIFF
--- a/exams/api.py
+++ b/exams/api.py
@@ -120,6 +120,7 @@ def authorize_for_exam_run(user, course_run, exam_run):
         user=mmtrack.user,
         course=course_run.course,
         exam_run=exam_run,
+        status=ExamAuthorization.STATUS_SUCCESS,
     )
     log.info(
         '[Exam authorization] user "%s" is authorized for the exam for course id "%s"',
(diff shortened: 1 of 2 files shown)


exams/api.py
```diff
@@ -120,6 +120,7 @@ def authorize_for_exam_run(user, course_run, exam_run):
         user=mmtrack.user,
         course=course_run.course,
         exam_run=exam_run,
+        status=ExamAuthorization.STATUS_SUCCESS,
```
noisecapella, line 123: Should this be in a defaults dict: `defaults={"status": ExamAuthorization.STATUS_SUCCESS}`? I think you would have multiple conflicting `ExamAuthorization` otherwise

noisecapella (changes requested):

noisecapella (approved):

PR #4740, opened 2021-01-06
Update command import_edx_exam_grades

annagav:
NA

#### What's this PR do?
Update the management command for importing exam grades.
Some of the updates include:
- different filel names
- the file contains grades for all courses that had exams
- adding the "no_show" boolean field

Run `./manage.py import_edx_exam_grades file_name`
with a provided file

### DIFF
--- a/exams/management/commands/import_edx_exam_grades.py
+++ b/exams/management/commands/import_edx_exam_grades.py
@@ -21,68 +21,61 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('csvfile', type=argparse.FileType('r'), help='')
 
-        parser.add_argument('course_code', help='Example: 14.100')
-
     def handle(self, *args, **kwargs):  # pylint: disable=unused-argument,too-many-locals
 
         csvfile = kwargs.get('csvfile')
-        course_number = kwargs.get('course_code')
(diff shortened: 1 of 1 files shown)


exams/management/commands/import_edx_exam_grades.py
```diff
@@ -21,68 +21,61 @@ class Command(BaseCommand):
...
+                exam_authorization.exam_no_show = True
                 exam_authorization.save()
             else:
-                existing_grades += 1
+                defaults = {
+                    'passing_score': exam_run.passing_score,
+                    'score': float(row['score']),
+                    'grade': row['grade'],
+                    'percentage_grade': float(row['score']) / 100.0 if row['score'] else 0,
+                    'passed': row['grade'].lower() == EXAM_GRADE_PASS,
```
noisecapella, line 63: Just double checking, is this correct? The previous code calculated this value by comparing the grade to `exam_run.passing_score`

annagav: Yes, before the it was using the passing score to decide whether the grade is passing. But Peter said that we could just get that value from the spreadsheet, I think Maya calculates this value when adjusting the grades.

noisecapella (approved):

PR #4747, opened 2021-01-12
Import edx grades: handle username does not exist

annagav:
none

#### What's this PR do?
 handle username does not exist when processing the task import_edx_exam_grades

nothing should break

### DIFF
--- a/exams/management/commands/import_edx_exam_grades.py
+++ b/exams/management/commands/import_edx_exam_grades.py
@@ -9,7 +9,7 @@ from django.core.management import BaseCommand, CommandError
 
 from courses.models import Course
 from exams.models import ExamRun, ExamAuthorization
-from exams.pearson.constants import EXAM_GRADE_PASS, EXAM_GRADE_FAIL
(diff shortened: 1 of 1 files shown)


exams/management/commands/import_edx_exam_grades.py
```diff
@@ -9,7 +9,7 @@
 
 from courses.models import Course
 from exams.models import ExamRun, ExamAuthorization
-from exams.pearson.constants import EXAM_GRADE_PASS, EXAM_GRADE_FAIL
+from exams.pearson.constants import EXAM_GRADE_PASS
```
pdpinch, line 12: Is this change necessary?

pdpinch (approved):
I left you a question, but if it doesn't affect anything, I'm a 👍

exams/management/commands/import_edx_exam_grades.py
annagav: no, it was unused import

PR #4751, opened 2021-01-13
upgrade redux-hammock

aliceriot:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

none

#### What's this PR do?

I recently upgraded some dependencies in redux-hammock (mitodl/redux-hammock#31). This just bumps us up to that version here.

We don't use redux-hammock for everything in MM, but we do use it for these reducers:

```
  automaticEmailsEndpoint,
  courseEnrollmentsEndpoint,
  coursePricesEndpoint,
  programLearnersEndpoint,
  discussionsFrontpageEndpoint,
  channelsEndpoint
```

so I think we should just check that the functionality for all of those things is working correctly.


annagav (approved):
👍

PR #4761, opened 2021-01-20
Exams: set number of attempts to 1

### LINKED ISSUE #4646: Exams: Purchase only buys one exam attempt
Currently, each order entitles the learner to two exam attempts. In order to ease the transition to edX exams, and to cut costs, we want to change this to one attempt. 

#### Designs and Mockups

We need to review the dashboard text to see where, if anywhere, the number of attempts is explicitly exposed. 

#### Acceptance Criteria:

- [ ] After date one (TBD), _new_ orders are good for only one purchase. The date should be configurable. 
- [ ] After date two (TBD but definitely in 2021), _all_ orders are good for only one purchase. I believe the number of attempts is already configurable.

annagav:
Fix #4646

#### What's this PR do?
Limit learners to one exam attempt for one payment

There are two dates to set up on the program itself `exam_attempts_first_date` and `exam_attempts_second_date`. First date is one after which all users who pay for the course after that date get 1 attempt.
one attempt.  The second date is for switching completely to 1 attempt for everybody independently of the payment date.

Set up a user with a past paid course and a proctored exam grade. Set up a new exam run that is current. The dashboard should ask the user to pay again.
If now is in between the two dates `exam_attempts_first_date` and `exam_attempts_second_date`, the learner gets 2 attempts for payment before the first date and 1 attempts after the first date. 
If now is after the `exam_attempts_second_date` then all users get 1 attempt for any payment.


pdpinch:
@annagav As far as I can tell, this would affect _all_ purchases, and not just new ones. I see how this would address the 2nd acceptance criteria on the issue, but not the first. 

Also, are there any user messages in the UI that mentions the number of attempts?

annagav:
@pdpinch I didn't find any messages about the number of attempts in the dashboard. 

The dates can be set on the program itself. Also, I made an assumption that we would give two attempts to a user if their first payment was before the first date, (even if they other payment after the first date).

dashboard/api.py
```diff
@@ -29,7 +29,8 @@
 from profiles.api import get_social_auth
 
 # maximum number of exam attempts per payment
-ATTEMPTS_PER_PAID_RUN = 2
+ATTEMPTS_PER_PAID_RUN_OLD = 2
```
pdpinch, line 32: We might want to leave a comment here to explain what this constant is for. Also, we'll be able to remove it at some point, right?

pdpinch:
Thanks for checking the text. 

> Also, I made an assumption that we would give two attempts to a user if their first payment was before the first date, (even if they other payment after the first date).

I don't think it makes a big difference, but it would be more accurate to give the user two attempts (`ATTEMPTS_PER_PAID_RUN_OLD`) per payment before the first date and one attempt payment (`ATTEMPTS_PER_PAID_RUN`) after the first date. 

Is that too complex?

dashboard/api.py
```diff
@@ -29,7 +29,8 @@
 from profiles.api import get_social_auth
 
 # maximum number of exam attempts per payment
-ATTEMPTS_PER_PAID_RUN = 2
+ATTEMPTS_PER_PAID_RUN_OLD = 2  # the number of attempts the user gets for payment before the first date
```
noisecapella, line 32: Is it possible to rename this? What makes it different from `ATTEMPTS_PER_PAID_RUN`?

dashboard/api.py
```diff
@@ -583,7 +584,20 @@ def has_to_pay_for_exam(mmtrack, course):
...
+    if not mmtrack.program.exam_attempts_first_date:
+        num_attempts = ATTEMPTS_PER_PAID_RUN_OLD
+    elif mmtrack.program.exam_attempts_first_date > now:
+        # still before the date
+        num_attempts = ATTEMPTS_PER_PAID_RUN_OLD
+    elif mmtrack.program.exam_attempts_second_date > now:
+        # in between the dates: check when the user paid for each course run
+        attempt_limit = mmtrack.get_custom_number_of_attempts_for_course(course)
+
+    if not attempt_limit:
```
noisecapella, line 599: What if `attempt_limit` is `0`?

dashboard/utils.py
```diff
@@ -239,6 +239,26 @@ def get_payments_count_for_course(self, course):
...
+        Returns:
+            int: a number of attempts
+        """
+        lines = Line.objects.filter(
+            order__status__in=Order.FULFILLED_STATUSES,
+            order__user=self.user,
+            course_key__in=course.courserun_set.values('edx_course_key'),
+        ).order_by('created_at')
+        first_date = course.program.exam_attempts_first_date
+        num_attempts = sum([2
```
noisecapella, line 257: Are the `1` and `2` here related to the `ATTEMPTS_PER_PAID_RUN` constant?

noisecapella (changes requested):

dashboard/api.py
annagav: I named like this because we are intending to remove it completely after `second date` passes. I thought it would be easy to distinguish.

dashboard/api.py
annagav: If it is 0 that means that the number of total exams attempts can be computed by multiplying the number of payments by the number of attempts per run. 
In the case where we are in between the dates, Peter asked to give 2 attempts to users who paid before the first date, and 1 attempt after the first date, so that is computed in `get_custom_number_of_attempts_for_course`.

dashboard/utils.py
annagav: yes, i'm going to fix that.

annagav:
@noisecapella I addressed all the comments. Thank you!

dashboard/api.py
noisecapella: Can we change `attempt_limit = None` and have this be `if attempt_limit is None`? In general we should use `None` rather than `0` as a sentinel value

noisecapella (changes requested):

annagav:
@noisecapella I addressed you last comment. Let me know if there is anything else. Thank you!

noisecapella:
@annagav Thanks, the code looks great. I just need to finish testing functionality before I approve

noisecapella (approved):

Issue #4753, opened 2021-01-14
Upgrade boto to boto3


rhysyngsun:
`boto` is unmaintained and hasn't been updated in over 2 years. It's also now caused the warning
```
/usr/local/lib/python3.7/site-packages/boto/plugin.py:40
  /usr/local/lib/python3.7/site-packages/boto/plugin.py:40: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp
```

The `imp` module has been deprecated since python 3.4 and is scheduled for remove in python 3.10 ([see PEP 594](https://www.python.org/dev/peps/pep-0594/#imp)).

There is some usages in pearson code which we _probably_ don't need anymore. Please verify this w/ Peter and if so we can probably also remove the pearson code as a pre-req.

pdpinch:
Which Pearson code in particular uses boto?

arslanashraf7:
@pdpinch I was able to take an initial look at this and it seems that we are using `boto` in our `ExamDataAuditor` class of Pearson. e.g.

- `exam/pearson/audit` https://github.com/mitodl/micromasters/blob/master/exams/pearson/audit.py

Other indirect usages are associated with either tests of audit class or the usages of `ExamDataAuditor`.

arslanashraf7:
Looks like the code containing boto usage might be removed in another legacy code removal PR https://github.com/mitodl/micromasters/pull/4765/files#diff-050dcae6dde23195d90b4f02491429cf2cd2a8d13c4406bddc867906b29660acL5. Putting this on hold till that PR is finalized after which we can decide if we will need to proceed with this or not.

annagav:
@arslanashraf7 The PR with the removal of the  legacy Pearson code was merged.

arslanashraf7:
Deployed on production now.

PR #4772, opened 2021-02-09
Upgrading boto to boto3

arslanashraf7:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

#4753 

#### What's this PR do?
Upgrades deprecating `boto` to `boto3`

To test it manually we might need to do A/B testing and see nothing broke after this change. More critical parts on the testing would be the ones actually using boto, In our case it could be wagtail.

(Copied from the main issue) 
>`boto` is unmaintained and hasn't been updated in over 2 years. It's also now caused the warning

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -450,7 +450,7 @@ if (
         'AWS_STORAGE_BUCKET_NAME'
     )
 if MICROMASTERS_USE_S3:
-    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
 else:
     # by default use django.core.files.storage.FileSystemStorage with
     # overwrite feature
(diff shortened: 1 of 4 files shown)


annagav (approved):
The CMS seems to work for me. I was able to login and create program pages and view them 👍

arslanashraf7:
@annagav, Sorry for asking it again since you've already reviewed the PR, I was testing it after rebasing on the master, and just before merging I saw some issues related to `Storages backends` & `AWS` keys in this PR.

I've updated the storage backends to use `boto3` as well since that step was somehow missed.

The issues I faced while re-testing were only regarding the uploading of images in wagtail/cms.

To reproduce the issues you might need to enable `MICROMASTERS_USE_S3=True` in env, I tested it with dummy AWS API keys though but you can test it if you have real ones from RC.

Let me know what you think of the new changes and if you could identify any other thing we might have missed for the boto3 upgrade.

annagav:
@arslanashraf7 I enabled `MICROMASTERS_USE_S3` and tested the cms again. I was able to successfully upload images and view my program pages. Code looks good 👍

arslanashraf7:
@annagav Thanks for taking the time to review it again, I'll go ahead and merge it now.

PR #4790, opened 2021-02-22
Exam Coupons: use edX coupons when user loads the dashboard

### LINKED ISSUE #4770: Exams: use edX coupons more efficiently
In our current implementation of edX proctored exams, we require one coupon per ExamAuthorization. Since there are many, many users with historical authorizations, and because users don't always take their proctored exam in the same term as the course, we end up requiring far more coupons than are actually used. This is problematic because edX requires several weeks notice to create coupons, and it is difficult to accurately predict how many we will need before the course concludes. 

A more efficient system would be to have a pool of coupons which we assign to ExamAuthorizations when they are needed. There are multiple ways to determine "when the coupon is needed" but the simplest seems to do it at the time a user loads their dashboard and gets the exam/coupon link. Typically, users who aren't taking the exam won't be loading their dashboard. 

#### Acceptance Criteria:

- [ ] create a model for edX coupons. It should carry enough metadata to manage the coupons, i.e.
       - course the exam is associated with (foreign key to course?)
       - expiration date
       - coupon code
       - URL for using coupon code
       - edX course key for the exam course 
       - type of coupon ("exam")
       - status? 
       - other foreign keys?
- [ ] management command for ingesting coupons from a csv
- [ ] update the dashboard API to assign a coupon to the ExamAuthorization if necessary (hopefully this won't make the dashboard API much slower than it already is) 
- [ ] confirm that the front-end doesn't require any additional changes

#### Related Issues

- Update dashboard message to include link to instructions on how user can get to exam course
- Improve error handling when the front-end logic expects an exam coupon link and there is none

annagav:
Fix #4770 

#### What's this PR do?

- Adds a model for edx exam coupons, 
- modifies the management command `populate_edx_exam_coupons` to create instances of `ExamRunCoupon`.
- modifies the dashboard api to assign a coupon to a user if the user is authorized for an exam run when the user loads the dashboard.

Ask me for an example coupon file.
Run the management command `populate_edx_exam_coupons`.
Make sure your user is eligible to take an exam for a course
Load the dashboard, and make sure you have the exam link available to you.


dashboard/api.py
```diff
@@ -532,10 +532,22 @@ def get_edx_exam_coupon_url(user, course):
...
+        expiration_date__gte=now.date(),
+        is_taken=False,
+        course=course
+    ).first()
+    if exam_coupon is None:
+        return ""
+
+    exam_auth.exam_coupon = exam_coupon
+    exam_auth.save()
+    return exam_coupon.use_coupon()
```
HamzaIbnFarooq, line 550: I don't know if it's something to be considered but seems like a bug to me so mentioning it here.

Let's say we have 10 ExamRunCoupon with is_taken=False, when a user loads its dashboard we assign the first ExamRunCoupon to ExamAuthorization and toggle is_taken to true.
If the user loads the dashboard again, this code will pick the second ExamRunCoupon with is_taken=False and toggles that, in this way, all the coupons will be marked as is_taken=True while only 1 of them will be getting used.

In the second scenario, if a user loads his dashboard and the system assigns user's ExamAuthorization object with a ExamRunCoupon. Now let's assume there are no  ExamRunCoupon with is_taken=False are left, now if the user loads his dashboard the edx exam_url will be empty for him although he was assigned a coupon earlier.

exams/management/commands/populate_edx_exam_coupons.py
```diff
@@ -48,30 +50,25 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument,too-many-l
             raise CommandError(
                 'There are multiple courses with given number "{}"'.format(course_number)
             )
-
-        exam_runs = ExamRun.get_currently_schedulable(course)
-
-        if not exam_runs.exists():
-            raise CommandError(
-                'There are no eligible exam runs for course "{}"'.format(course.title)
+        from datetime import datetime
```
HamzaIbnFarooq, line 53: Can we place this among other imports at the top?

dashboard/api.py
annagav: you are right, thanks for catching that.

HamzaIbnFarooq (approved):
LGTM 👍

PR #4800, opened 2021-02-24
Removing some pearson text

### LINKED ISSUE #4795: Dashboard: remove "Pearson" from Exam grades pop-up
The most recent round of exams weren't offered by Pearson. 

image

To keep things simple, I think we can just replace "Pearson test" with "Proctored exam," unless there's an easy way to differentiate from Pearson exams and edX exams.

annagav:
Fix #4795

#### What's this PR do?
Removes text about pearson test on the dashboard, and a test.

This should effect the user dashboard, where the user clicks on the test score to view their attempts.

Where to find the text: In the course row the user has a little box with the exam grade. If the user clicks on that box with the grade, then a little popup opens up with the list of all the exam attempts and grades for that course.

### DIFF
--- a/static/js/components/dashboard/courses/GradeDetailPopup.js
+++ b/static/js/components/dashboard/courses/GradeDetailPopup.js
@@ -91,7 +91,7 @@ const renderExamRow = (
 ) => (
   <div className="course-run-row" key={idx}>
     <div className="title">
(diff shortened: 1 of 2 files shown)


umarmughal824:
@annagav could you share the steps to manually test it. Or any UI screenshots could also be helpful.

umarmughal824 (approved):
LGTM 👍 

this is how it looks like

Screenshot 2021-02-26 at 12 42 24

Issue #4801, opened 2021-02-24
Home: CSS regression


pdpinch:
In PR #4781 we removed the background color on the Program FAQ page, a regression that had been introduced by the home page update PR #4749.

Unfortunately, when we removed the background color on the FAQ page, we _also_ removed the background color on the home page. I'm embarrassed that we overlooked this. 

### Acceptance criteria

- [ ] Grey background of #EBECED on the homage page, as displayed in [InVision](https://invis.io/YFZ4N7KMWT3#/435513872_Homepage_V2)
- [ ] No background on the program FAQ page (e.g. https://micromasters.mit.edu/scm/frequently-asked-questions/ )

umarmughal824:
@pdpinch home page itself has it own dedicated scss files it should not disturb any of the UI but we have changed styling for HubSpot form responsiveness in the faqs page which could be impacted other pages.

But in faq pages, earlier the material-UI grid system was not working properly that I fixed only that could have an impact on the other page but not the home page itself.

arslanashraf7:
@umarmughal824 I'm sure this discrepancy was introduced in https://github.com/mitodl/micromasters/pull/4781, Where we attempted to remove background from FAQ tab's content on the program, And due to how it was extended from the homepage we somehow missed the regression of FAQ background on the homepage on part of reusability.
I'm going to put up a PR to fix this.

umarmughal824:
> @umarmughal824 I'm sure this discrepancy was introduced in #4781, Where we attempted to remove background from FAQ tab's content on the program, And due to how it was extended from the homepage we somehow missed the regression of FAQ background on the homepage on part of reusability.
> I'm going to put up a PR to fix this.

okay, understood. thanks

PR #4781, opened 2021-02-11
Fixed FAQ background & link underline decoration

arslanashraf7:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
#4779 

#### What's this PR do?
- Removes the background from the FAQ content panel
- Removes the tab underlines on Program page

#### How should this be manually tested?
- Add and FAQ page in a Program through CMS
- Open program page in the web app
- Check the background color on FAQs it should be white
- Check the Tabs, their text shouldn't be underlined

#### Any background context you want to provide?
We recently did major design changes in the HomePage for which you can have a context thrugh this PR https://github.com/mitodl/micromasters/pull/4749

#### Screenshots (if appropriate)
Screenshot 2021-02-11 at 4 00 08 PM

PR #4749, opened 2021-01-13
home page redesign

umarmughal824:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
  - [x] Tag @ferdi or @pdpinch for review
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
fixes #4736
fixes #4758

#### What's this PR do?
Home Page re-design

#### How should this be manually tested?
Just visit the home page

PR #4803, opened 2021-02-25
Fixed the FAQ background regression

arslanashraf7:
#### Pre-Flight checklist

  - [x] Desktop screenshots
  - [x] Mobile width screenshots

- [x] Testing
  - [x] Code is tested

#4801 

[`Invision link`](https://projects.invisionapp.com/share/YFZ4N7KMWT3#/screens/435513872_Homepage_V2) for HomePage

#### What's this PR do?
Fixes the FAQ background color discrepancy in the program and homepage.

- Check that the FAQ section on the HomePage has a background (`#EBECED`)
- Check the FAQs tab on the program page doesn't have a grey background.

In PR #4781 we removed the background color on the Program FAQ page, a regression that had been introduced by the home page update PR #4749.
While fixing that we somehow missed the color on the homepage since the SCSS for both had the same class names for FAQs and its background removal got overridden on the homepage too.

**Home Page(Desktop)**
Screenshot 2021-02-25 at 1 31 20 PM

**Home Page(Mobile)**
Screenshot 2021-02-25 at 1 31 35 PM

**Program Page(Desktop)**
Screenshot 2021-02-25 at 1 51 00 PM

**Program Page(Mobile)**
Screenshot 2021-02-25 at 1 51 15 PM

I wish I could find a very relevant one for this 😥

### DIFF
--- a/static/scss/program-page.scss
+++ b/static/scss/program-page.scss
@@ -372,6 +372,7 @@
 
   .faqs {
     margin-top: 20px;
+    background-color: transparent;
 
     @include breakpoint(phone) {
       margin: 18px auto;
--- a/static/scss/homepage.scss
+++ b/static/scss/homepage.scss
@@ -684,6 +684,7 @@ a {
 
 .faqs {
   padding: 106px 0;
+  background-color: #ebeced;
 
   .faq-holder {
     max-width: 600px;


static/scss/program-page.scss
```diff
@@ -372,6 +372,7 @@
 
   .faqs {
     margin-top: 20px;
+    background-color: transparent;
```
umarmughal824, line 375: @arslanashraf7 is that intentional? It doesn't seem to show any changes, can you confirm that after removing it?

arslanashraf7: Yes, It to keep the background transparent on FAQ tab of the program page.

umarmughal824 (approved):
👍

arslanashraf7:
@pdpinch can you please have a look at this as well, I'm pretty sure that we shouldn't be seeing a ripple on this now.

pdpinch (approved):
Thanks. The screenshots look good to me. I think adding the `transparent` style is a good idea because it makes the intention clear. 

> Are there any other potential pages we should test for FAQs except HomePage and ProgramPage?

It looks like there are two templates that use `class="faqs"`. I don't remember exactly how the [categeorized_faqs_page](https://github.com/mitodl/micromasters/blob/0f4caa22263468959415a9ec3e161f7bbb710730/cms/templates/cms/categorized_faqs_page.html) is used, but I think it's part of a page like this one: https://micromasters.mit.edu/scm/frequently-asked-questions/

In any case, the styling should be the same as [faqs_page](https://github.com/mitodl/micromasters/blob/0f4caa22263468959415a9ec3e161f7bbb710730/cms/templates/cms/faqs_page.html) with a transparent background.

arslanashraf7:
@pdpinch, yes, the usage you mentioned above is part of the FAQ tab's content on the program page where we list all the FAQs. Going ahead with merging the PR.

Issue #4805, opened 2021-02-25
Update Certificate sharing like edX


umarmughal824:
Make Certificates social sharing identical to edX

#### Designs and Mockups

<img width="687" alt="facebook" src="https://user-images.githubusercontent.com/4043989/109150664-fab09200-778a-11eb-92df-e6be7d25bd41.png">
<img width="1523" alt="linkedin" src="https://user-images.githubusercontent.com/4043989/109150667-fbe1bf00-778a-11eb-969b-3b1cc2659f10.png">
<img width="752" alt="twitter" src="https://user-images.githubusercontent.com/4043989/109150668-fd12ec00-778a-11eb-92ac-2cc9dcf2521c.png">

#### Acceptance Criteria:

- [ ] Sharing should fill-up the form as shown above

umarmughal824:
@HamzaIbnFarooq as I am going to review @annagav second [PR](https://github.com/mitodl/micromasters/pull/4790) for today, so I don't have enough capacity to work on it. So it would be better for you to pick that.

HamzaIbnFarooq:
@pdpinch
Can you please share the url made by edx.org for linkedin certificate sharing and also let me know what is the URN for the organization (micromasters)?

pdpinch:
Does this URL help? https://www.linkedin.com/in/pdpinch/edit/certification/new/?certId=a4f596e7e0444e959a4ccb9401fb5721&certUrl=https%3A%2F%2Fcourses.edx.org%2Fcertificates%2Fa4f596e7e0444e959a4ccb9401fb5721&isFromA2p=true&issueMonth=4&issueYear=2016&name=edX%20Verified%20Certificate%20for%20Information%20and%20Communication%20Technology%20(ICT)%20Accessibility%20&organizationId=2746406

If that works, I think we should keep the same organizationId.

PR #4812, opened 2021-03-03
add certificate sharing context

HamzaIbnFarooq:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/micromasters/issues/4805

#### What's this PR do?
Adds Certificate sharing context

Create a certificate and share it on facebook, twitter and linkedin.

**LinkedIn:**
course:
image

program:
image

**Twitter:**
image

**Facebook:**


certificates/views.py
```diff
@@ -74,6 +75,15 @@ def get_context_data(self, **kwargs):
...
         context['certificate'] = certificate
+        context[
+            'linkedin_share_context'
+        ] = 'certId={certId}&issueMonth={issueMonth}&issueYear={issueYear}&name={certName}&organizationId={organizationId}&isFromA2p=true'.format(
+            certId=certificate.hash,
+            issueMonth=certificate.created_on.month,
+            issueYear=certificate.created_on.year,
+            certName=course.title,
+            organizationId="2746406",
+        )
```
arslanashraf7, line 86: I think we can move this chunk to a common method since all the values and string format is the same for both `Courses` and `Programs` except `certName` which we can pass to that LinkedIn context generator method.

Up to you if you decide to create the method in the parent class https://github.com/mitodl/micromasters/blob/hamza/4805-add_certificate_sharing_context/certificates/views.py#L27 to dynamically use it or as an independent function.

certificates/views.py
```diff
@@ -108,6 +118,15 @@ def get_context_data(self, **kwargs):
...
         context['signatories'] = list(signatories)
         context['certificate'] = certificate
+        context[
+            'linkedin_share_context'
+        ] = 'certId={certId}&issueMonth={issueMonth}&issueYear={issueYear}&name={certName}&organizationId={organizationId}&isFromA2p=true'.format(
+            certId=certificate.hash,
+            issueMonth=certificate.created_on.month,
+            issueYear=certificate.created_on.year,
+            certName=program.title,
+            organizationId="2746406",
```
arslanashraf7, line 128: I think it would be a good idea that we use this value through some settings/env variable, Obviously that solely depends upon the probability of this value's change. 
It will give us the freedom of changing this value in the future without any code change & deployment.

arslanashraf7 (changes requested):

certificates/views.py
pdpinch: Good suggestion, but I doubt that we will ever be changing the organizationID so it's not worth making it configurable.

pdpinch:
Based on the screenshot, it looks like we're missing some of the Facebook metadata, in particular page title and page description. 

Looking at the source for https://courses.edx.org/certificates/a4f596e7e0444e959a4ccb9401fb5721 I see `<title>` tags, but I don't see a description, so I'm not sure how edX is populating that. The content appears to be the same as the text content of the paragraph with`class="accomplishment-statement-lead"`

HamzaIbnFarooq:
> Looking at the source for https://courses.edx.org/certificates/a4f596e7e0444e959a4ccb9401fb5721 I see `<title>` tags, but I don't see a description, so I'm not sure how edX is populating that. The content appears to be the same as the text content of the paragraph with`class="accomplishment-statement-lead"`

I am not sure about this but is it possible that when facebook render's the URL it populates all that stuff from page, in my screenshot's case the URL was referring to my localhost that's why it didn't render the text?

arslanashraf7 (approved):
Looks good 👍

I was also able to test the Facebook content render for a certificate by replacing the localhost URL with the [URL](https://courses.edx.org/certificates/a4f596e7e0444e959a4ccb9401fb5721) that Peter gave earlier.

HamzaIbnFarooq:
@pdpinch
I deployed my branch and after deployment when I tried sharing the certificate on facebook it looks like this:
image

Is this satisfactory?

pdpinch:
I think the MITx logo is appropriate, but can we do something about the cropping? What's Facebook's preferred aspect ratio, so the image doesn't get cropped on the left and the right?

HamzaIbnFarooq:
> What's Facebook's preferred aspect ratio, so the image doesn't get cropped on the left and the right?

According to https://developers.facebook.com/docs/sharing/webmasters/images/:
`Try to keep your images as close to 1.91:1 aspect ratio as possible to display the full image in News Feed without any cropping.
 `

HamzaIbnFarooq:
@pdpinch 
I have updated the PR and now the Facebook sharing content is rendered like the image below, does this looks appropriate?

pdpinch:
Looks good, thanks!

pdpinch (approved):

Issue #4788, opened 2021-02-22
Certificates: update styling to match current edX certificates


pdpinch:
As a user, I'd like certificates from my FA courses and programs to look the same as edX certificates. 

#### Designs and Mockups

Current [course certificates look like this](https://micromasters-rc.odl.mit.edu/certificate/course/74fa750f2d30c4cfa8bfc21052d88056) and [program certificates look like this](https://micromasters-rc.odl.mit.edu/certificate/program/0d54d09954752e507775550841644584).

They both need to be updated to match the [current edx.org certificates](https://courses.edx.org/certificates/dd6838d94f4b41038c03be77aed91172)

#### Acceptance Criteria:

- [ ] Remove the background logo
- [ ] Replace the MicroMasters logo with "Verified Certificate"
- [ ] Make sure that when the page is printed it look similar to the on-screen certificate

#### Related Issue

- #4784 certificates share and print bar

HamzaIbnFarooq:
@pdpinch, when we try to print the edX certificate the background image doesn't get included. Are we going to replicate that as well?
![image](https://user-images.githubusercontent.com/42243411/108716934-82ee2780-753e-11eb-916b-a95bf74928dc.png)

HamzaIbnFarooq:
Oh, sorry, that seems to be a browser-specific thing, background images need to be allowed by browsers, doesn't have anything to do with development.

pdpinch:
The goal is to make this certificate behave the same as the edX one. 

Maybe take a look at the edX help page for printing certificates and see what they recommend as far as printing backgrounds. 

https://support.edx.org/hc/en-us/articles/207206647-How-are-edX-certificates-delivered-

https://support.edx.org/hc/en-us/articles/206504068-Can-I-get-a-paper-copy-of-my-certificate-

HamzaIbnFarooq:
@pdpinch 
Sorry I didn't consider this scenario earlier, so mentioning it now, when there are more than 3 signatories the printer can't print that onto a single page (e.g: https://micromasters-rc.odl.mit.edu/certificate/program/0d54d09954752e507775550841644584) should we do something about it before the design release (https://github.com/mitodl/micromasters/pull/4813)?

pdpinch:
In production, the max number of signatures we have is 4. 

What can we do?

- Peter

HamzaIbnFarooq:
@ Pdpinch, the latest certificate design has been deployed to prod, kindly have a look into it so we can close this issue.

[...]

PR #4818, opened 2021-03-08
add support for multiple signatories in certificate design

HamzaIbnFarooq:
#### Pre-Flight checklist

  - [ ] Desktop screenshots

https://github.com/mitodl/micromasters/issues/4788

#### What's this PR do?
add support for 4 signatories in certificate design, previously the design started to break at 4.

Create a certificate with 4 signatories and see if the print works well.

#### Screenshots (Printed Version)
image

### DIFF
--- a/static/scss/certificate.scss
+++ b/static/scss/certificate.scss
@@ -164,7 +164,9 @@
     font-style: italic;
   }
 
-  .signatory,
+  .signatory {
+    margin-top: 0rem !important;
+  }
   .wrapper-accomplishment-orgs {
     margin-top: 1.5rem !important;
   }
@@ -337,6 +339,11 @@
     max-height: 150px;
   }
 
+  .signatory .signatory-name {
+    margin-top: 0px;
+    margin-bottom: 2px;
+  }
+
   .signatory .signatory-signature {
     max-width: 60%;
     max-height: 60px;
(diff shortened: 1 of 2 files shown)


arslanashraf7:
@HamzaIbnFarooq, Although the code changes are fine I believe there is a noticeable difference in the look & feel of the spacing and overall border of certificate & signatories and that best can be seen when we have less than 4 signatories e.g 3.

I think it would be a good idea to add more screenshots(**Old vs New**) for both normal view and print view in the description of the PR and I think we'll have to wait for @pdpinch to review them.

Just to add a thought, A solution could be to support the old design for 3 signatories, and when there are 4 signatories we add these styling classes on the runtime. 

I'm adding some screenshots that I took just to give an idea.
**(Old Print with 3 signatories)**
Screenshot 2021-03-09 at 7 37 21 PM
**(New Print with 3 signatories)**
Screenshot 2021-03-09 at 7 38 37 PM

HamzaIbnFarooq:
Thanks @arslanashraf7 for pointing this out. I think rather than looking at a comparison with previous versions, we should have a comparison with edX:
**3 Signatories in edX:**
image

**3 Signatories in MM:**
image


**4 signatories in edX:**
image

**4 Signatories in MM:**
image

pdpinch:
Thank you both for your care and diligence on this. 

The goal is to have our certificates look the same as edX’s. Looking at the screenshots from Hamza, I think this is OK as is, with one stylesheet and the signatures crowded at the top when there are only 2.

arslanashraf7 (approved):
Looks good 👍

PR #4829, opened 2021-03-16
Update existing FAs with current tier programs

### LINKED ISSUE #4823: User can't checkout with old financial aid
### Steps to Reproduce

1. Create a user
2. View the dashboard for a program with finacial aid
3. Click "skip and pay for price" which will create a full price financial aid object for the user
4. The "pay now" button should be active for your user, click it.
5. You should navigate to the checkout page, click the checkout button
6. There should be no API error and you should advance to the Cybersource site. Don't pay for the course. 
7. Now, in the django admin for financial aid program tiers, uncheck the "current" box for the corresponding tier. This prevents new users from accessing this tier 
5. As the user, return to the dashboard 
2. The "pay now" button should still be active for your user, click it. 
3. You should navigate to the checkout page, click the checkout button

### Expected Behavior

- Similar to the first time through, there should be no API error and you should advance to the Cybersource site

### Actual Behavior

- the checkout API returns a 400 error

### Stacktrace

See https://sentry.io/organizations/mit-office-of-digital-learning/issues/2270428820/?environment=production&project=104497&query=is%3Aunresolved

appears to be coming from this line of code: https://github.com/mitodl/micromasters/blob/0f4caa22263468959415a9ec3e161f7bbb710730/ecommerce/api.py#L84

annagav:
Fix #4823

#### What's this PR do?
Adds a management command for finding all financial aid objects that have outdated program tiers and updating them with new ones.

Try to reproduce the bug as described in the issue. 
Then create new tier_program with the same income_threshold as the one you were testing with. Then run the management command `./manage.py  migrate_finaid_program_tiers`

### DIFF
--- a/financialaid/management/commands/migrate_finaid_program_tiers.py
+++ b/financialaid/management/commands/migrate_finaid_program_tiers.py
@@ -28,11 +28,14 @@ class Command(BaseCommand):
 
             except TierProgram.DoesNotExist:
                 raise CommandError(
-                    'Could not find a current tier program with threshold "{}"'.format(threshold)
+                    'Could not find a current tier program with threshold "{}" for financial aid {}'.format(
+                        threshold,
+                        financial_aid.id
+                    )
                 )
             except TierProgram.MultipleObjectsReturned:
                 raise CommandError(
-                    'There are multiple courses with given number "{}"'.format(threshold)
+                    'There are multiple tier programs with threshold "{}"'.format(threshold)
                 )
 
             financial_aid.tier_program = tier_program
--- a/financialaid/management/commands/migrate_finaid_program_tiers.py
+++ b/financialaid/management/commands/migrate_finaid_program_tiers.py
@@ -1,0 +1,42 @@
+"""
+Update FinancialAid objects with current tier program
+"""
+from django.core.management import BaseCommand, CommandError
+
+from financialaid.models import FinancialAid, TierProgram
+
+
+class Command(BaseCommand):
+    """
+    Updates the existing financial aid objects to current tier programs
+    """
+    help = "Updates the existing financial aid objects to current tier programs"
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+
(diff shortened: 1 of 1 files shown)


financialaid/management/commands/migrate_finaid_program_tiers.py
```diff
@@ -0,0 +1,42 @@
...
+            try:
+                threshold = financial_aid.tier_program.income_threshold
+                tier_program = TierProgram.objects.get(
+                    income_threshold=threshold,
+                    current=True,
+                )
+
+            except TierProgram.DoesNotExist:
+                raise CommandError(
+                    'Could not find a current tier program with threshold "{}"'.format(threshold)
```
pdpinch, line 31: Can we include some information about the user and/or financialaid object that failed here?

financialaid/management/commands/migrate_finaid_program_tiers.py
```diff
@@ -0,0 +1,42 @@
...
+                )
+            except TierProgram.MultipleObjectsReturned:
+                raise CommandError(
+                    'There are multiple courses with given number "{}"'.format(threshold)
```
pdpinch, line 35: I think this is the wrong error message?
```suggestion
                    'There are multiple TierProgram with given threshold "{}"'.format(threshold)
```

financialaid/management/commands/migrate_finaid_program_tiers.py
```diff
@@ -0,0 +1,42 @@
...
+                )
+
+            financial_aid.tier_program = tier_program
+            financial_aid.save_and_log(None)
```
pdpinch, line 39: Does this work with 'None'? Is there some way to pass a user into the management command? Not super critical.

pdpinch (changes requested):
I can think of other things that might be nice to have, but I don't think they're useful enough to justify the complexity -- like a dryrun option or the ability to run it for just on user. 

I think you should merge after you consider my other comments.

PR #4830, opened 2021-03-17
Fix sentry Errors -- typeError on Checkout & IE 11.0 forEach error

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

- https://github.com/mitodl/micromasters/issues/4834
- https://github.com/mitodl/micromasters/issues/4835
- https://github.com/mitodl/micromasters/issues/4614

#### What's this PR do?
 - Fixes an undefined payload error in case of checkout API failure response
 - Adds a polyfill & Fixes IE 11.0 error on property method forEach on the program page

- To reproduce and test checkout error you can follow the steps defined in https://github.com/mitodl/micromasters/issues/4823, On the order summary page, when you click `Continue` you'll see the error logged in the console in case of 400 API response.

- To reproduce the forEach property error, you'll need to open any program page which has rich text in it. Open the page in IE 11.0 and you'll see the error logged in the console.

**Note:**
I've run the seed commands and added other pages for testing in the [CI deployment](https://micromasters-ci-pr-4830.herokuapp.com/) for this PR, let me know if you'll need anything on that.

### DIFF
--- a/static/js/containers/OrderSummaryPage.js
+++ b/static/js/containers/OrderSummaryPage.js
@@ -23,6 +23,9 @@ import type { RestState } from "../flow/restTypes"
 import type { CheckoutResponse } from "../flow/checkoutTypes"
 
 export const processCheckout = (result: CheckoutResponse) => {
+  if (!result) {
+    return
+  }
   const { payload, url, method } = result
   if (method === "POST") {
     const form = createForm(url, payload)
--- a/static/js/entry/public.js
+++ b/static/js/entry/public.js
@@ -23,6 +23,11 @@ import {
 import SignupDialog from "../containers/SignupDialog"
 import CopyLinkDialog from "../containers/CopyLinkDialog"
 
+// Adding forEach polyfill for IE
+if (window.NodeList && !NodeList.prototype.forEach) {
+  NodeList.prototype.forEach = Array.prototype.forEach
+}
+
 // Program Page course list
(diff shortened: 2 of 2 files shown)


static/js/containers/OrderSummaryPage.js
```diff
@@ -23,6 +23,9 @@ import type { RestState } from "../flow/restTypes"
 import type { CheckoutResponse } from "../flow/checkoutTypes"
 
 export const processCheckout = (result: CheckoutResponse) => {
+  if (!result) {
```
annagav, line 26: I think this issue  #4823 had already beed fixed, or is this fixing something differnt?

annagav (changes requested):
Just one comment about the order summary page

static/js/containers/OrderSummaryPage.js
arslanashraf7: @annagav, You're right, These issues are connected (https://github.com/mitodl/micromasters/issues/4834 & https://github.com/mitodl/micromasters/issues/4823), The only difference is that the one you mentioned is related to the checkout API(Backend) and the one fixed in this PR is related to manipulation of the API Frontend response in the frontend.

We can keep this fix if we think the checkout API might give an error in the future, otherwise, we can remove it. I'm okay with both but I'll wait on you, either to remove it or keep it.

annagav (approved):
👍

PR #4841, opened 2021-03-22
Missed deadline message for exams

annagav:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
  - [x] Tag @ferdi or @pdpinch for review
- [ ] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Fix #4824

#### What's this PR do?
Updates the some logic regarding the exam message when the user missed deadline to pay for the course. 

#### How should this be manually tested?
Set up a course
`./manage.py alter_data set_to_needs_upgrade --username <username> --course-title 'Digital Learning 100' --missed-deadline
`
Then set up an ExamRun that happens the same term as the course run.
You should get a message like this:
Screen Shot 2021-03-21 at 10 00 43 PM
And if you create another ExamRun in the future, then:
Screen Shot 2021-03-21 at 10 07 41 PM

PR #4844, opened 2021-03-23
Restrict exam authorization if user missed payment deadline

annagav:
#### Pre-Flight checklist

- [ ] Testing
  - [x] Code is tested

Depends on #4841 

#### What's this PR do?
Restrict exam authorization if user missed payment deadline

Set a course to enrolled but needs to pay using the alter_date management command:
`./manage.py alter_data set_to_needs_upgrade --username <username> --course-title 'Digital Learning 100' --missed-deadline
`
Then create exam runs, but not exam authorization since user gets authorized only once they paid for the course.
```
exam_run = ExamRunFactory.create(course=course, authorized=True, scheduling_past=False, scheduling_future=False)
```
This will create an exam that that is schedulable now. Which means that once there is a payment the user will get authorized for the exam run.
This restriction applies only user pays for current course run:
find the course run for which your user is auditing(it was printed out when you ran the alter_data command), and set the `end_date` for that course run to 2 weeks ago:
```
from courses.models import *
edx_course_key=<>
course_run=CourseRun.objects.get(edx_course_key=edx_course_key)
course_run.end_date = now_in_utc - datetime.timedelta(weeks=2)
```
Go ahead and pay for the course:
```
from seed_data.lib import *
set_course_run_to_paid(user, course_run)
```

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -289,6 +289,21 @@ class MMTrack:
         """
         return self.has_final_grade(edx_course_key) and self.has_paid(edx_course_key)
 
+    def paid_but_missed_deadline(self, course_run):
+        """
+        Checks if user paid for this run only after the deadline
+        """
+        if self.has_paid(course_run.edx_course_key):
+            orders = Order.objects.filter(
+                status__in=Order.FULFILLED_STATUSES,
+                user=self.user,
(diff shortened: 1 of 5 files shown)


umarmughal824 (approved):
LGTM 👍

PR #4868, opened 2021-04-12
Standardize Log In / Sign Up buttons

asadiqbal08:
#### What are the relevant tickets?
fixes: #4860 

#### What's this PR do?
Standardize Log In / Sign Up buttons

screenshot-localhost_8079-2021 04 13-15_03_33

---------------------- 

screenshot-localhost_8079-2021 04 13-13_39_46

PR #4870, opened 2021-04-13
Program Page: Learn more button

### LINKED ISSUE #4853: Direct users to sign up for marketing mail
As a marketer, I'd like MM users to sign up for more information as quickly as possible so I can see if I can drive more enrollments. 

#### Design 

n/a

#### Acceptance Criteria:

- [ ] If "Program subscribe link" is set in the CMS, display a "Learn More" button that links to that URL instead of the current "Sign Up Now" button in the hero section, which initiates an edX registration flow. 
- [ ] If there is no "Program subscribe link" set in the CMS, maintain the current behavior. 

#### out of scope

No change to the sign up / login buttons, but we will standardize this in another issue

annagav:
Fix #4853

#### What's this PR do?
Updates the "I'm insterested" button into "Learn more" button. Now show the button whenever the subscribe link is filled out for the program.

Checkout the program page, make sure to view it when logged in and logged out.
Has the subscribe link:
Screen Shot 2021-04-13 at 11 12 51 AM
Does not have the link and user logged in:
Screen Shot 2021-04-13 at 11 14 03 AM
Does not have the link and user not authenticated:
Screen Shot 2021-04-13 at 11 14 26 AM


cms/templates/cms/program_page.html
```diff
@@ -58,7 +58,7 @@ <h1 class="title">
         <div class="description">
           {{ page.title_over_image|richtext }}
         </div>
-        {% if courses %}
+        {% if not page.program_subscribe_link %}
           {% if not authenticated %}
             <button class="mdl-button hero-button open-signup-dialog mdl-cell--hide-phone">
               Sign Up Now
```
pdpinch, line 64: I think we're going to get a merge conflict with #4868. Would you mind merging that first and then rebasing this?

cms/templates/cms/program_page.html
```diff
@@ -58,7 +58,7 @@ <h1 class="title">
         <div class="description">
           {{ page.title_over_image|richtext }}
         </div>
-        {% if courses %}
+        {% if not page.program_subscribe_link %}
```
pdpinch, line 61: This logic is pretty confusing. I know you're trying to make the minimal change, but would you mind refactoring it to put `if page.program_subscribe_link` first?

pdpinch (approved):
Wow, that was pretty easy. I made a couple of comments, but you can decide how you want to handle them.

cms/templates/cms/program_page.html
annagav: It is actually first here. I am removing the courses check since it is redundant now.

annagav: I see what you mean :)

cms/migrations/0050_helper_text_change.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('cms', '0049_benefitspage_content_table_block'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='programpage',
+            name='program_subscribe_link',
+            field=models.URLField(blank=True, help_text='A url for Learn More button, if there is no course(s) available.', null=True),
```
pdpinch, line 16: Why do we need a migration here?

If we must do a migration, we might as well change the help text, I think?

cms/models.py
```diff
@@ -356,7 +356,7 @@ class ProgramPage(Page):
     program_subscribe_link = models.URLField(
         blank=True,
         null=True,
-        help_text="A url for I'm interested button, if there is no course(s) available."
+        help_text="A url for Learn More button, if there is no course(s) available."
```
pdpinch, line 359:
```suggestion
        help_text="A url for Learn More button. If none, then the button won't display."
```

is that clear?

PR #4924, opened 2021-06-11
raising 404 for records

HamzaIbnFarooq:
https://github.com/mitodl/internal/issues/31

#### What's this PR do?
raise 404 for anonymous users on records

Just anonymously access some records, it should be giving 404.

### DIFF
--- a/certificates/views.py
+++ b/certificates/views.py
@@ -205,6 +205,10 @@ class GradeRecordView(TemplateView):
 
         enrollment = get_object_or_404(ProgramEnrollment, hash=kwargs.get('record_hash'))
         user = enrollment.user
+
+        if user.is_anonymous or self.request.user != user:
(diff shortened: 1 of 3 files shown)


arslanashraf7 (approved):
👍 Works well.

arslanashraf7:
@HamzaIbnFarooq, I've reviewed and tested your latest commit and things work as intended. With the latest change the `Send Learner Record` and `Share` and functionality would probably become invalid as well, Should we hide these buttons?

cc: @pdpinch

pdpinch:
> With the latest change the Send Learner Record and Share functionalities would probably become invalid as well, Should we hide these buttons?

This is true, but we can address that in a follow-up.

arslanashraf7 (approved):
👍

PR #4930, opened 2021-06-11
verify micromasters.mit.edu on Yandex

pdpinch:

None 

#### What's this PR do?

Verifies micromasters.mit.edu for a couple of search engines.

It has to be released, and then I can check to see if we have ownership of the site in each search engine's webmaster tools.


annagav (approved):
👍

PR #4937, opened 2021-06-15
Create new hash for shared record access

HamzaIbnFarooq:
#### Pre-Flight checklist
  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing

https://github.com/mitodl/internal/issues/32

#### What's this PR do?
* Adds allow-sharing and revoke-sharing records feature.
* Creates two record views one for the owner and one for shared access
* Uses UUID for the hash generation to make hashes unpredictable

Create a ProgramEnrollment and try accessing it through various ways and test the whole functionality

Owner's Record View:
image

Owner's Shared Record View:
image

Owner's Shared Record View (On Revoking Button Click):
image

Owner's Shared Record View (After Revoking):
image

Anonymous Shared Record View:
image


static/js/entry/public_jquery.js
```diff
@@ -0,0 +1,6 @@
+__webpack_public_path__ = `${SETTINGS.public_path}` // eslint-disable-line no-undef, camelcase
```
HamzaIbnFarooq, line 1: `public.js` requires these lines to be executed and previously they were executed through `style_public.js`. `style_public` and link sharing `rrssb` buttons depend upon `is_public` context passed from view to template but in our scenario as the record page `/records/programs/enrollment_id/` is not going to be shared with any one so we can't pass `is_public` which in return causes a crash of `public.js`. So this file was included just to fulfill that requirement.

rhysyngsun:
The backend code looks good.

annagav (approved):
👍 looks good. I think this is good to go. The only thing that might be a concern is if a learner sent the record to an institution and they revoked for some reason, I hope it would make sense to the user that they have to re-share the record again.

HamzaIbnFarooq:
> The only thing that might be a concern is if a learner sent the record to an institution and they revoked for some reason, I hope it would make sense to the user that they have to re-share the record again.

Yes, good point, the current implementation will revoke everything. So let me add this re-sharing part explicitly to the warning text.

Furthermore, once this feature is release, all pre-shared hashes will be revoked and we might need to inform all the learners about that. @pdpinch do you have any thoughts on it?

HamzaIbnFarooq:
**Update**: I have squashed commits and updated the warning text.

pdpinch:
Screeshots look good to me. Can you give examples of the URLs?

1. Owner's Record View
2. Shared Record View

HamzaIbnFarooq:
@pdpinch  they will be looking like this:

> 1. Owner's Record View

https://micromasters.com/records/programs/1

> 2. Shared Record View

https://micromasters.com/records/programs/1/shared/UUID

PR #4948, opened 2021-06-17
fix: change sharable record url

HamzaIbnFarooq:
#### Description
Previously we had only one URL for program records (`/records/programs/HASH`) so that same URL was used by the learner to look at his records and the same was shared across schools.

But now, we have two different record URLs and I saw that the old one was being used for sharing records through emails although it's not public anymore.

#### What's this PR do?
This PR will replace the new sharable link with the old link.

Try sending a record sharing email through `/records/programs` page


annagav (approved):
👍

Issue #4952, opened 2021-06-18
Update trim-newlines


HamzaIbnFarooq:
off-topic question: why do we need an npm package to trim newlines? Is that really not a standard JS function?

See https://github.com/mitodl/micromasters/security/dependabot/yarn.lock/trim-newlines/open

HamzaIbnFarooq:
> off-topic question: why do we need an npm package to trim newlines? Is that really not a standard JS function?

Apparently, JS doesn't have such a feature and even if there is one, micromasters is not directly using it rather it's a child dependency of `node-sass`.

arslanashraf7:
@HamzaIbnFarooq, Just FYI, There is are similar alerts in [Bootcamp](https://github.com/mitodl/bootcamp-ecommerce/security/dependabot/yarn.lock/trim-newlines/open) & [xPRO](https://github.com/mitodl/mitxpro/security/dependabot/yarn.lock/trim-newlines/open) as well.

It might be a good idea to create similar PRs for there too.

HamzaIbnFarooq:
closed by https://github.com/mitodl/micromasters/pull/4953

PR #4953, opened 2021-06-18
Upgrade trim-newlines to v3.0.1

HamzaIbnFarooq:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/micromasters/issues/4952

#### What's this PR do?
Upgrade trim-newlines to v3.0.1

Just check everything is working, specifically webpack build compilations

`node-sass` requires `meow` and `meow` requires `trim-newlines`. To upgrade `trim-newlines` the ideal way is to upgrade node-sass from 5.0.0 to 6.0.0 but that requires the node version from 10 to be upgraded to >12. Even if we upgrade the node version we have to upgrade a lot of other dependencies as well.

That's why we have to use resolution for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,5 +152,8 @@
   "devDependencies": {
     "enzyme": "^3.10.0",
     "enzyme-adapter-react-16": "^1.15.2"
+  },
+  "resolutions": {
+    "node-sass/**/trim-newlines": "^3.0.1"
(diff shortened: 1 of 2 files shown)


arslanashraf7 (approved):
👍

PR #4954, opened 2021-06-18
searchkit version bump to fix axios sercurity alert

### LINKED ISSUE #4383: Can't upgrade `axios` as a searchkit dependency
Searchkit needs to update this [dependency](https://github.com/searchkit/searchkit/blob/staging/yarn.lock#L1084).  To get rid of the security alert need to `Upgrade axios to version 0.18.1 or later`.

umarmughal824:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

fix #4383 

#### What's this PR do?
searchkit version bum to fix axios sercurity alert

Just check everything is up

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "sass": "^1.35.1",
     "sass-lint": "^1.10.2",
     "sass-loader": "10",
-    "searchkit": "2.3.0",
+    "searchkit": "2.4.4",
     "serialize-javascript": "^3.1.0",
     "sinon": "^2.3.1",
     "slick-carousel": "^1.6.0",
--- a/static/js/components/ProgramFilter.js
+++ b/static/js/components/ProgramFilter.js
@@ -61,6 +61,7 @@ export default class ProgramFilter extends SearchkitComponent {
   }
 
   componentDidMount() {
+    this._initAccessor()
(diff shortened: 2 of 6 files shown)


umarmughal824:
@noisecapella I tried to run it with an accessor but it was not getting into that accessor code at all so I did it in some other way as per my knowledge and understanding so let me know your suggestion on that. As I don't have much details on the searchkit working so it might need some tweaks. The weird thing is its functionality is working now but a couple of tests are not showing the expected output so your input would be very handy on that. @annagav that would be nice, if you could answer that too.

noisecapella:
I experimented with this and made a commit in a separate branch to upgrade searchkit: https://github.com/mitodl/micromasters/commit/2c1a14212fb06a19d955267d119a05e7f2b4cf1c

In general:
 - We should upgrade searchkit to 2.4.4 since it looks like that's the latest stable version
 - Some functions are overriding functions in searchkit components. There were places where they made checks on internal state, and if internal state wasn't ready the function would exit early. I copied these checks into our own functions where we needed to
 - CustomNoHits stubs out `componentDidMount` since `componentWillMount` wasn't present anymore

Can you update this PR with those changes?

noisecapella (approved):

Issue #4959, opened 2021-06-21
Search: move reindex into a celery task


pdpinch:
We tried to rebuild the learner index, but the Heroku dyno timed-out before it could complete. Reindex should be set up to run as a celery task.

arslanashraf7:
@pdpinch, While working on this I saw a few things and had some assumptions that I'm posting here for more clarity.

**_Assumption:_** I'm assuming that we usually run the reindexing through https://github.com/mitodl/micromasters/blob/master/search/management/commands/recreate_index.py command which then uses the https://github.com/mitodl/micromasters/blob/master/search/indexing_api.py#L575 for further operations. You can let me know if I'm missing something in this assumption.

**Questions1:** If we are trying to run this `recreate_index` asynchronously, we might need to keep a `synchronous` version of this because in some usages, e.g. https://github.com/mitodl/micromasters/blob/master/search/base.py#L21 we might still need to run reindexing synchronously. Should we keep this concern in mind?

**Questions2:** This is partially related to question#1, Instead of making the whole `create_reindex` running asynchronously through celery, shouldn't we go for creating subtasks for https://github.com/mitodl/micromasters/blob/master/search/indexing_api.py#L604 and https://github.com/mitodl/micromasters/blob/master/search/indexing_api.py#L612 and then group and run these subtasks parallelly inside re-index?

Although, I've created a WIP [PR](https://github.com/mitodl/micromasters/pull/4960/files) that just runs the `recreate_index` command through a celery task as a whole.

pdpinch:
@rhysyngsun or @noisecapella can please advise?

rhysyngsun:
@arslanashraf7 the `recreate_index` command in Open is probably your best reference for how this should be implemented. The use case in that project is much more complicated than we need here. The general strategy is what you suggested in Question 2:

https://github.com/mitodl/open-discussions/blob/master/search/tasks.py#L699

I suspect you'll see some differences between some of the lower-level elasticsearch APIs, so this might be something we need to split across a couple of PRs to bring MicroMasters up to speed with the umproved implementation in Open.

arslanashraf7:
@rhysyngsun Thanks for providing the details on this one. I'll take a look at the referenced Open-discussion link.

>so this might be something we need to split across a couple of PRs

[...]

PR #4960, opened 2021-06-22
Run recreate index through a celery task

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#4959 

#### What's this PR do?
- It moves the execution of `recreate_index` out of the main thread and puts it to a separate celery task

- Running the command e.g.  `./manage.py recreate_index` should now run separately in a celery task.

(Optional)

Previously we were facing timeouts when running this command with high amount of records.

### DIFF
--- a/search/management/commands/recreate_index.py
+++ b/search/management/commands/recreate_index.py
@@ -16,4 +16,5 @@ class Command(BaseCommand):
         """
         Recreates the index
         """
-        recreate_index_async.delay()
+        recreate_index_task = recreate_index_async.delay()
+        self.stdout.write('Running reindexing task with Id: {}'.format(recreate_index_task.id))
--- a/search/management/commands/recreate_index.py
+++ b/search/management/commands/recreate_index.py
@@ -1,13 +1,8 @@
 """
 Management command to recreate the Elasticsearch index
 """
-import logging
 
 from django.core.management.base import BaseCommand
-
-from search.indexing_api import (
-    __name__ as indexing_api_name,
-)
 from search.tasks import recreate_index_async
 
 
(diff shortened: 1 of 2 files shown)


pdpinch:
Arslan, I know there are still some open questions about the issue this addresses, but you should still put in some minimal details in the PR description.

arslanashraf7:
> Arslan, I know there are still some open questions about the issue this addresses, but you should still put in some minimal details in the PR description.

Thanks for pointing this out, I've added the initial details.

search/tasks.py
```diff
@@ -136,3 +137,11 @@ def populate_query_memberships(percolate_query_id):
         percolate_query_id (int): Database id for the PercolateQuery to populate
     """
     api.populate_query_memberships(percolate_query_id)
+
+
+@app.task
```
rhysyngsun, line 142: So we can ensure this task doesn't get dropped.

```suggestion
@app.task(acks_late=True)
```

search/management/commands/recreate_index.py
```diff
@@ -41,10 +41,10 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
             import uuid
             profile = cProfile.Profile()
             profile.enable()
-            recreate_index()
+            recreate_index_async.delay()
```
rhysyngsun, line 44: A couple things:

- remove the profiling code, it's not doing anything anymore
- print out the task id so if it doesn't finish we can investigate

search/management/commands/recreate_index.py
```diff
@@ -1,50 +1,19 @@
...
-            profile = cProfile.Profile()
-            profile.enable()
-            recreate_index()
-            profile.disable()
-            filename = 'recreate_index_{}.profile'.format(uuid.uuid4())
-            profile.dump_stats(filename)
-            self.stdout.write('Output profiling data to: {}'.format(filename))
-        else:
-            recreate_index()
+        recreate_index_async.delay()
```
noisecapella, line 19: Can we write out the task id to the console here?

noisecapella (changes requested):

search/management/commands/recreate_index.py
arslanashraf7: @noisecapella, Thanks for pointing this out I was writing the task id within the task but the command will be a better place for it. I've now updated the PR as per your suggestion and this is ready for review.

Also, Just FYI, There is another [PR](https://github.com/mitodl/micromasters/pull/4968) that is actually rebased on this one. That PR is the successor part of what we wanted to achieve in the ticket. The associated ticket has more details regarding that. You can take a look at that as well and we can decide either to merge this PR and check out the feasibility of the other one or close this PR in favor of the other [PR](https://github.com/mitodl/micromasters/pull/4968) which is rebased on this.

noisecapella (approved):

search/management/commands/recreate_index.py
noisecapella: I think it makes sense to merge this and then continue the work in #4968

PR #4963, opened 2021-06-23
use uuid based hashes for certificates

HamzaIbnFarooq:
#### Pre-Flight checklist
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/micromasters/issues/4964

#### What's this PR do?
It uses UUID based hashes for certificates to make the URLs unpredictable

Create program and course certificates the hash should be working fine, secondly, the old certificates and their hashes should not be disturbed.

### DIFF
--- a/grades/factories.py
+++ b/grades/factories.py
@@ -80,6 +80,7 @@ class MicromastersProgramCertificateFactory(DjangoModelFactory):
 
     user = SubFactory(UserFactory)
     program = SubFactory(ProgramFactory)
+    hash = uuid.uuid4().hex
 
     class Meta:  # pylint: disable=missing-docstring,no-init,too-few-public-methods
         model = MicromastersProgramCertificate
--- a/grades/factories.py
+++ b/grades/factories.py
@@ -1,5 +1,6 @@
 """Factories for the grades app"""
 import datetime
+import uuid
 
 from factory import (
     SubFactory,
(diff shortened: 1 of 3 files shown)


annagav (changes requested):
We also need to update hash generation in factories.py the creation of `MicromastersCourseCertificateFactory` and `MicromastersProgramCertificateFactory`

annagav:
Also please remove `generate_md5` function fro `micromasters/utils.py`. I don't think it is needed any longer.

HamzaIbnFarooq:
Thanks, Anna for the review.

> We also need to update hash generation in factories.py the creation of `MicromastersCourseCertificateFactory` and `MicromastersProgramCertificateFactory`

I have already changed the hash mechanism for `MicromastersCourseCertificateFactory` in this PR at https://github.com/mitodl/micromasters/pull/4963/files#diff-75b41e76db697a24ee58812d32897e2257d2c215d173d2cafcde3e0054bbe42bR72. Although I will add the hash for `MicromastersProgramCertificateFactory` now.

> Also please remove generate_md5 function fro micromasters/utils.py. I don't think it is needed any longer.

generate_md5 is being used in two migrations 1) https://github.com/mitodl/micromasters/blob/master/dashboard/migrations/0008_programenrollment_hash.py and 2) https://github.com/mitodl/micromasters/blob/master/grades/migrations/0014_remove_micromasterscoursecertificate_final_grade.py so I don't think it's possible to get rid of that without applying some hack here, do you have any thoughts on this?

annagav (approved):
Looks good 👍

Issue #4956, opened 2021-06-21
Upgrade glob-parent to version 5.1.2 or later


arslanashraf7:
Alert Link: https://github.com/mitodl/micromasters/security/dependabot/yarn.lock/glob-parent/open

arslanashraf7:
Update:

The package `glob-parent` was a 3rd & 4th level dependency of `babel-plugin-istanbul` and `webpack ` accordingly. 

However, for `babel-plugin-istanbul` this upgrade was directly available and was fixed with simple upgrade to `6.0.0`.

But for `webpack` this upgrade was only available in `webpack5` requiring major `webpack` migration which seems to have many breaking changes. So, In order to fix it for webpack without webpack 5 migration, I've added a resolution for it in the PR https://github.com/mitodl/micromasters/pull/4957/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R158. This resolution will only affect the 4th level `glob-parent` version and nowhere else.

pdpinch:
Sorry, I don't know enough to comment. @annagav or @aliceriot can you comment?

Separately, when should we think about investing the time in a webpack upgrade?

aliceriot:
as long as the app is building and everything is working I think that's a reasonable approach to take (the resolution I mean).

we're using webpack 4 in most other places, mainly because of those pretty large differences in how it's configured and whatnot. They have a fairly extensive guide for the upgrade here: https://webpack.js.org/migrate/5/

I think we probably should eventually, but if we can get the upgrade we want without doing so then it's no problem to wait on the webpack upgrade for now.

arslanashraf7:
@aliceriot Thanks for commenting and strategizing the way ahead, I had something similar in mind while creating the PR and adding the resolution.

@annagav, On the above note, I think we can go ahead with reviewing the [associated PR](https://github.com/mitodl/micromasters/pull/4957). Could you take a look when possible?

arslanashraf7:
Closing, Associated PR has been merged & deployed.

PR #4965, opened 2021-06-23
remove node-sass in favor of sass, remove unused package

aliceriot:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings
  - [ ] New settings are documented and present in `app.json`
  - [ ] New settings have reasonable development defaults, if applicable

none, just noticed after looking at #4956

#### What's this PR do?

`node-sass` is now deprecated, and the pure-JS `sass` package is preferred. Getting rid of the former in favor of the latter removes a native code dependency (hooray!) which has often been a major headache for us in the past.

I also realized that I don't think we are using `babel-plugin-istanbul` for anything so it can be removed. 

build should pass, there shouldn't be any CSS issues.


arslanashraf7 (approved):
👍

PR #4970, opened 2021-06-29
Make client_authorization_id field optional in ProctoredExamGrade

### LINKED ISSUE #4936: Exams: client_authorization_id shouldn't be a required field
As an admin, I sometimes need to update user's proctored exam grades. When I tried to do this in the django admin, I was (temporarily) blocked because client_authorization_id, a field that is only relevant to Pearson grades, is a required field. 

We shouldn't remove the field entirely, since it's still useful for the legacy Pearson grades. But it shouldn't be a required field. We could also add a hint in the admin that it's a legacy field.

annagav:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
  
Fix #4936

#### What's this PR do?
Make client_authorization_id field optional in ProctoredExamGrade

Run migrations. In Django admin create a ProctoredExamGrade for your learner. `client_authorization_id` should not be required.


HamzaIbnFarooq (approved):
LG

PR #4978, opened 2021-07-08
Revert "Bump wagtail from 2.12.4 to 2.12.5"

arslanashraf7:
Reverts mitodl/micromasters#4951


pdpinch (approved):
I don't know why there are so many changes in requirements.txt, but it seems like it's just reordering things. 

The only real change is wagtail pin

Issue #4989, opened 2021-07-23
mitX Online Integration: Add social auth backend


rhysyngsun:
Add a social auth backend for mitX Online, this should largely be the same as the edx.org one, with some differences:

### Design
- If the learner has an mitX Online social auth, always use that as the source of truth (i.e. the logic to update the email address from edx.org should only be run if that’s the only social auth for that learner).
- The mitX Online social auth should pull user profile data from the mitX Online user api not the OpenedX user api. If there’s gaps here, let’s discuss, but I believe MicroMasters should be ok as long as it’s getting at least basic email/name data.
- For now, the edx.org social auth should continue to be the default one.

### Acceptance Criteria
- [x] The learner should be able to create a new account with edx.org or mitX Online
- [x] The learner should be able to login with existing accounts for all combinations of having edx.org and/or mitX Online social auths.
- [x] The learner should be able to register with both edx.org and mitX Online using an existing MicroMasters account. Social auth should provide this behavior out-of-box, but we should verify this is configured and functions without issue.

PR #4997, opened 2021-07-27
Remove usages of get_social_username

rhysyngsun:
#### Pre-Flight checklist

- [ ] Testing
  - [x] Code is tested

Part of #4989 

#### What's this PR do?
This PR removes usages of `get_social_username` in favor of simply using `User.username` so that we can support users with social auth other than edx. I also renamed the existing functions in `profiles/api.py` so it's obvious those are edx-specific.

Generally speaking, nothing should break.

You should still be able to:
- View CMS pages
- Create an account
- Login with an existing account
- View your dashboard
- View your profile
- Edit your profile
- Make a payment

### DIFF
--- a/backends/pipeline_api_test.py
+++ b/backends/pipeline_api_test.py
@@ -7,7 +7,7 @@ import ddt
 
 from backends import pipeline_api, edxorg
 from courses.factories import ProgramFactory
-from profiles.api import get_social_username
+from profiles.api import get_edxorg_social_username
 from profiles.models import Profile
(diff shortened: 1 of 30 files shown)


annagav:
@rhysyngsun is this ready for review? it seems to be marked as a draft.

rhysyngsun:
@annagav I think generally this is looking good, I still need to figure out how to test out the payments but i think we can probably test that on RC if need be

annagav (approved):
👍

PR #5006, opened 2021-08-05
Update Cache for enrollments and grades for mitxonline

### LINKED ISSUE #4992: MITx Online Integration: Enrollments and grades should support alternative backends
For enrollments and grades:
- When querying edx/openedx for enrollment and grade data, make the endpoint/hostname for `edx-api-client` conditionally configured based on the value of `CourseRun.courseware_backend`.
- Add settings in `settings.py` as required for the above.

### Acceptance Criteria

- [ ] Learner with only edxorg course runs can update their dashboard
- [ ] Learner with only mitxonline course runs can update their dashboard
- [ ] Learner with course runs from both backends can update their dashboard

annagav:
Fix #4992 

#### What's this PR do?
Add support of cached enrollements and grades with another courseware backend mitxonline.

Register with edx org and enroll in courses there.
Register and enroll in courses with mitxonline as the courseware backend.
Test that:
Learner with only edxorg course runs can update their dashboard
Learner with only mitxonline course runs can update their dashboard

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -8,12 +8,13 @@ import pytz
 
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.core.exceptions import ImproperlyConfigured
+from django.core.exceptions import ImproperlyConfigured, ObjectDoesNotExist
 from django.db import transaction
 from django.urls import reverse
 from django_redis import get_redis_connection
 from edx_api.client import EdxApi
 
+from backends.constants import COURSEWARE_BACKEND_URL, BACKEND_EDX_ORG, BACKEND_MITX_ONLINE
 from backends.exceptions import InvalidCredentialStored
 from backends import utils
 from courses.models import Program, ElectiveCourse, CourseRun
@@ -26,9 +27,10 @@ from grades.models import FinalGrade
 from grades.serializers import ProctoredExamGradeSerializer
 from exams.models import ExamAuthorization, ExamRun, ExamRunCoupon
 from micromasters.utils import now_in_utc
-from profiles.api import get_edxorg_social_auth
 
 # key that stores user_key and number of failures in a hash
+from profiles.api import get_social_auth
+
 CACHE_KEY_FAILURE_NUMS_BY_USER = "update_cache_401_failure_numbers"
 # key that stores user ids to exclude from cache update
 CACHE_KEY_FAILED_USERS_NOT_TO_UPDATE = "failed_cache_update_users_not_to_update"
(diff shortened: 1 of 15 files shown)


dashboard/models.py
```diff
@@ -41,11 +41,12 @@ class Meta:
         abstract = True
 
     @classmethod
-    def user_qset(cls, user, program=None):
+    def user_qset(cls, user, program=None, provider='edxorg'):
```
rhysyngsun, line 44: I think maybe we should split this into 2 methods so our code is more explicit about when edx-only is being called:

```python
    @classmethod
    def user_qset(cls, user): ...

    @classmethod
    def program_qset(cls, program, provider): ...
```

So existing calls of `user_qset()` would remain as such and other calls could turn into `user_qset().program_qset(program, provider)`.

annagav: Thanks! I updated this, not the way you suggested though, but I think it makes sense for the usages of these method. Let me know what you think.

dashboard/api_edx_cache.py
```diff
@@ -341,9 +348,9 @@ def update_all_cached_grade_data(cls, user):
...
         # get the credentials for the current user for edX
-        user_social = get_edxorg_social_auth(user)
+        user_social = get_social_auth(user, 'edxorg')
         utils.refresh_user_token(user_social)
         # create an instance of the client to query edX
         edx_client = EdxApi(user_social.extra_data, settings.EDXORG_BASE_URL)
-        cls.update_cached_certificates(user, edx_client)
-        cls.update_cached_current_grades(user, edx_client)
+        cls.update_cached_certificates(user, edx_client, 'edxorg')
+        cls.update_cached_current_grades(user, edx_client, 'edxorg')
```
rhysyngsun, line 356: Should this be doing this for all social auths the learner has?

dashboard/models.py
```diff
@@ -54,23 +55,15 @@ def user_qset(cls, user, program=None):
...
         if program is not None:
-            query_params.update(dict(course_run__course__program=program))
+            query_params.update(dict(
+                course_run__course__program=program,
+            ))
+        if provider is not None:
+            query_params.update(dict(
+                course_run__courseware_backend=provider
+            ))
         return cls.objects.filter(**query_params)
```
rhysyngsun, line 65: Take it or leave it: this would probably be a bit cleaner with multiple query = query.filter(...) lines instead of building up a dict for a single call.

dashboard/api_edx_cache.py
```diff
@@ -251,29 +256,30 @@ def update_cached_certificates(cls, user, edx_client):
...
             None
         """
 
-        course_ids = models.CachedEnrollment.active_course_ids(user)
+        course_ids = models.CachedEnrollment.active_course_ids(user, provider=provider)
 
         # Current Grades are out of date, so fetch new data from edX.
         current_grades = edx_client.current_grades.get_student_current_grades(
-            get_edxorg_social_username(user), course_ids)
+            get_social_username(user, provider), course_ids)
```
rhysyngsun, line 282: `get_social_username` could potentially raise an exception here because it may not exist. I'm thinking we should return from this function if that's the case.

It might also be worth changing `get_social_username` to do a `filter(provider=provider).first()` so it returns `None`, rather than handling it via exception catching, but that's obviously more work so I'll leave that to your judgement if it makes things easier here.

dashboard/api.py
```diff
@@ -131,29 +133,18 @@ def __repr__(self):
...
-            for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
-                CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type)
-        except InvalidCredentialStored:
-            # this needs to raise in order to force the user re-login
-            raise
-        except:  # pylint: disable=bare-except
-            log.exception('Impossible to refresh edX cache')
-
+    update_cache_for_backend(user, 'edxorg')
+    update_cache_for_backend(user, 'mitxonline')
```
rhysyngsun, line 147: Could you use the constants for these instead?

rhysyngsun (changes requested):

dashboard/api_edx_cache.py
annagav: it looks like the exception is handled already and it returns None if object not found.

rhysyngsun (approved):
LGTM

PR #5012, opened 2021-08-13
Edx coupons: update regex

annagav:
N/A

#### What's this PR do?
Edx coupons csv file comes with a variety of formats for the course key field under "catalog query". This PR modifies the regex to match `key:(edx_course_key)` and `key:"edx_course_key"`

Ask me for an sample file
run the management command `./manage.py populate_edx_exam_coupons file.csv`


pdpinch:
Thanks Anna, but it's unlikely we'll ever use this, as the next exam offering should be on mitxonline.

PR #5017, opened 2021-08-19
reshesh cache update error

### LINKED ISSUE #5013: MITx Online Integration: dashboard-up-to-date error message
When a learner's enrollments and grades are out of date, the following error message is displayed:

> Sorry, edx.org is not responding as expected, so some of the information on this page may not be up to date.
>
> Please refresh the browser, or check back later.

Now that we're adding MITx Online as an alternative backend, we need to update this:

- [ ] Update the frontend to display a more generic error message:
  > Sorry, we're unable to update your dashboard at this time, so some of the information on this page may not be up to date.
  >
  > Please refresh the browser, or check back later.
- [ ] Update the `are_all_caches_fresh` to return `False` only if the caches for backends the learner has social auth for are out of date. If the learner doesn't have social auth for edx or mitx online, that should be treated as `True`

annagav:
Fix #5013

#### What's this PR do?
- Update the frontend to display a more generic error message
- Update the are_all_caches_fresh to return False only if the caches for backends the learner has social auth for are out of date. If the learner doesn't have social auth for edx or mitx online, that should be treated as True

Create a user that has course enrollments in both courseware backends, and make sure the dashboard loads with no error.
Then make the one of the backends not reachable, and refresh the dashboard, you should see a message at the top saying that the data might not be fresh.


backends/utils.py
```diff
@@ -56,3 +61,20 @@ def update_email(user_profile_edx, user):
     """
     user.email = user_profile_edx.get('email')
     user.save()
+
+
+def has_social_auth(user, provider):
```
rhysyngsun, line 66: Could you add a test for this?

dashboard/api_edx_cache.py
```diff
@@ -133,7 +137,7 @@ def is_cache_fresh(cls, user, cache_type):
         Returns:
             bool
         """
-        if cache_type not in cls.SUPPORTED_CACHES:
+        if cache_type not in cls.SUPPORTED_CACHES + cls.MITXONLINE_SUPPORTED_CACHES:
```
rhysyngsun, line 140: Could you update the tests for the modifications in this module?

rhysyngsun (changes requested):

annagav:
@rhysyngsun the tests should be there now. Thank you

rhysyngsun (approved):
LGTM

PR #5033, opened 2021-09-01
Update all cache types in update_cache_for_backend

annagav:
Related to the recent changes in adding cache for mitxonline backend

#### What's this PR do?
This fixes a bug in the release #5030

Create an account for edx or use an existing account.
When you signup for mitxonline your cache should be fresh and you should not see an error message that data is not fresh on the dashboard.


dashboard/api_edx_cache.py
```diff
@@ -74,6 +74,10 @@ class CachedEdxDataApi:
     # the sorting of the supported caches matters for refresh
     SUPPORTED_CACHES = (ENROLLMENT, CERTIFICATE, CURRENT_GRADE,)
     MITXONLINE_SUPPORTED_CACHES = (ENROLLMENT_MITXONLINE, CURRENT_GRADE_MITXONLINE,)
```
rhysyngsun, line 76: I think this needs to be this:
```suggestion
    MITXONLINE_SUPPORTED_CACHES = (ENROLLMENT, CURRENT_GRADE,)
```

because otherwise I'm getting this error:

```
web_1      | [2021-09-02 15:02:47] ERROR 14 [dashboard.api] api.py:784 - [b9898e1cf8e7] - Impossible to refresh edX cache
web_1      | Traceback (most recent call last):
web_1      |   File "./dashboard/api.py", line 779, in update_cache_for_backend
web_1      |     CachedEdxDataApi.update_cache_if_expired(user, edx_client, cache_type, provider)
web_1      |   File "./dashboard/api_edx_cache.py", line 344, in update_cache_if_expired
web_1      |     raise ValueError("{} is an unsupported cache type".format(cache_type))
web_1      | ValueError: enrollment_mitxonline is an unsupported cache type
```

rhysyngsun (changes requested):

annagav:
@rhysyngsun I think I addressed the issue. Thank you

dashboard/api.py
```diff
@@ -729,7 +729,7 @@ def refresh_user_data(user_id, provider=BACKEND_EDX_ORG):
         log.exception("Unable to create an edX client object for student %s", user.username)
         return
 
-    for cache_type in CachedEdxDataApi.SUPPORTED_CACHES:
+    for cache_type in CachedEdxDataApi.EDX_SUPPORTED_CACHES:
```
rhysyngsun, line 732: Should this be this instead?
```suggestion
    for cache_type in CachedEdxDataApi.CACHE_TYPES_BACKEND[provider]:
```

annagav: this function is only used by the batch_update_user_data task, so I thought to leave it as is. But I think I can update it.

rhysyngsun (approved):

PR #5035, opened 2021-09-02
Update the circular progress to show passed courses

### LINKED ISSUE #5022: Donut charts don't always accurately show how many courses remain
### Steps to Reproduce

- In SDS, complete and pass the three required courses and the two electives. Do not complete the capstone. 
- Go to SDS in your dashboard (or the `/learner` page for this user)

### Expected Behavior

- In this particular instance, since the learner still has a core course to complete, the donut chart should say 4/5
- A first pass at the logic for calculating the donut chart is:
   - minimum of:
      - [number of required courses] - [number of not-completed non-elective courses]
      - [number of courses completed]
   - In the example above: `min([5-1], 5)` => `min(4,5)` => `4`

### Actual Behavior

- The donut chart shows 5/5  courses complete
- Note that the donut chat is on the learner's dashboard as well as in the `/learner` page

### Related Issues

https://github.com/mitodl/micromasters/issues/5021

### Screenshot or Screencast

Screen Shot 2021-08-24 at 11 45 22 PM

annagav:
Fix #5022

#### What's this PR do?
Update the circular progress component to show a number of courses that the user passed towards the completion of the program.

Create an ElectivesSet for your program that your user is enrolled in.
Have your user pass all the courses in the elective set, or at least more than what is required.
Check to see the dashboard that all the required number of courses is reflected on the dashboard.

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -182,7 +182,8 @@ def get_info_for_program(mmtrack):
             mmtrack.program.num_required_courses
             if mmtrack.program.electives_set.exists()
             else mmtrack.program.course_set.count()
-        )
+        ),
+        "number_courses_passed": mmtrack.get_number_of_passed_courses_for_completion()
     }
     if mmtrack.financial_aid_available:
         data["financial_aid_user_info"] = FinancialAidDashboardSerializer.serialize(mmtrack.user, mmtrack.program)
--- a/dashboard/api_test.py
+++ b/dashboard/api_test.py
@@ -1824,6 +1824,7 @@ class InfoProgramTest(MockedESTestCase):
             'calculate_final_grade_average.return_value': 91,
             'get_program_certificate_url.return_value': "",
             'get_program_letter_url.return_value': "",
+            'get_number_of_passed_courses_for_completion.return_value': 0,
         })
         mock_info_course.return_value = {'position_in_program': 1}
         res = api.get_info_for_program(self.mmtrack)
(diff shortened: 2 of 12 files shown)


arslanashraf7:
The code changes look good but while testing it a ran into a scenario and I think it's worth mentioning it so that you can comment. I saw that the total courses required that show is always equal to the number defined in the program. E.g. in scenarios considering a program "Test Program":
- The program has 3 Core courses
- The program has 3 Elective courses
- The program has 1 elective set
- The required number of courses set in the program is 3 
- The required number of courses set in the elective set is 2
- Passed cores courses is 3
- Passed elective courses is 2

I believe the total courses should have be `5` in this case e.g. `required_in_program + required_in_elective [3] + [2] `, but you can comment better.

And the results are as follows:
Screenshot 2021-09-08 at 6 17 20 PM

annagav:
@arslanashraf7 yes, we can compute the number of required courses as well, but I think the people who create the program know pretty well how many courses need to be passed. If we want to change this I would say open another issue, where we can think of removing this field from the program model all together and just compute this number everytime a dashboard is accessed.

arslanashraf7:
> @arslanashraf7 yes, we can compute the number of required courses as well, but I think the people who create the program know pretty well how many courses need to be passed. If we want to change this I would say open another issue, where we can think of removing this field from the program model all together and just compute this number everytime a dashboard is accessed.

@annagav That makes sense. Thanks for clarifying that.

arslanashraf7 (approved):
👍

PR #5057, opened 2021-09-22
URL escape usernames for profile urls (#5056)

### LINKED ISSUE #5056: APIs with learner username are returning a 404 for MITx Online users
Some usernames that are possible on MITx Online contain characters that aren't currently being matched by the url regex, resulting in a 404 and a visible error message to the learner.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #5056 

#### What's this PR do?

Initially I thought this was because we weren't escaping the username, but those should be fine. Instead it was a url matching issue, so this updates the regex we use in the API urls so that it matches on newer username patterns coming from MITx Online. See the [corresponding regex](https://github.com/mitodl/mitxonline/blob/main/users/serializers.py#L30) in MITx Online.

- On `master`, verify you can reproduce the bug by changing your username to contain a `+` character. You should see 404 responses from the APIs
- Checkout this branch and verify the APIs now return 200 and you can load the dashboard and profile pages.

### DIFF
--- a/dashboard/urls.py
+++ b/dashboard/urls.py
@@ -7,9 +7,10 @@ from dashboard.views import (
     UnEnrollPrograms,
     ToggelProgramEnrollmentShareHash,
 )
+from profiles.constants import USERNAME_RE_PARTIAL
 
 urlpatterns = [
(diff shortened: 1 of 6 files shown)


annagav (approved):
LGTM

PR #5062, opened 2021-09-27
Don't refresh learner dashboards for staff (#5061)

### LINKED ISSUE #5061: Only update the dashboard if the current user is the same as the dashboard being loaded
Staff can view learner profiles, which causes the learner's dashboard to be loaded (`/api/v0/dashboard/USERNAME/`). This works fine most of the time, but if the learner's edX auth credentials are invalid it can cause the staff user to get [logged out](https://github.com/mitodl/micromasters/blob/master/static/js/lib/api.js#L61).

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #5061 

#### What's this PR do?
This fixes the following things:
- Updates the appropriate view test to not make API calls to edX
- Updates the view.py and api.py to conditionally update the dashboard only the dashboard user matches the requesting user

For both of the following verify the UI makes a request for the user's dashboard when viewing the profile page and that it does not error. You can shut down openedx or disable your networking if you're using edx staging to ensure that a request isn't being made.
- As any user, view your dashboard, it should load
- As a staff user, view the profile of any learner

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -133,18 +133,20 @@ class CourseRunUserStatus:
         )
 
 
-def get_user_program_info(user):
+def get_user_program_info(user, *, update_cache=True):
     """
     Provides a detailed serialization all of a User's enrolled Programs with enrollment/grade info
 
     Args:
         user (User): A User
+        update_cache (bool): if True, update the cache for the backends
 
     Returns:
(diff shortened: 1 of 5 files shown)


annagav (approved):
👍

PR #5065, opened 2021-09-29
Add support for discontinued runs (#5064)

### LINKED ISSUE #5064: Add support for discontinued course runs
We need to be able to discontinue a course run during a run. This would involve setting `enrollment_end_date` and `end_date` to the past, but a lot of our logic is based around course runs always receiving frozen grades. Because of this a past run without frozen grades is considered a current run. We need an additional state between that is for a discontinued run so that learners can enroll in subsequent runs without us having to generate fake grades.

This should be a boolean field, defaulted to `False`, and `dashboard.api.get_status_for_courserun` updated to handle this before the check for `has_frozen_grades`. Theremay be other place where we need to handle this so some additional due diligence is probably warranted.

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [ ] Testing
  - [ ] Code is tested

Fixes #5064 

#### What's this PR do?
This PR reconfigures the data models and queries to ignore discontinued course runs so the application behaves as if they don't exist, except for django admin that can still see them

- Mark a current run as discontinued in django admin
- The run should no longer show up in the dashboard
- Create a new run that is future, the dashboard should reflect the future run
- Change that run to be current, the dashboard should let the learner enroll now


annagav (approved):
👍

PR #5116, opened 2021-11-11
fix Cybersource cancellation or success redirect failures

### LINKED ISSUE #5113: Cybersource cancellation or success redirect fails
### Steps to Reproduce

We're observing that the POST request from cybersource to `/dashboard/` is currently prompting the user to login with edX. It's very likely that before this wouldn't have been observed because the user would've been redirected directly to edx to login and then redirected back to the MicroMasters dashboard, so it would appear to "work". Since we've now added an intermediate `/signin/` this redirect is now visible.

The redirect itself is happening because the POST request is triggered by javascript which means the browser treats this similarly to an XHR cross origin request, which causes it to not send most of the user's cookies, particularly the session cookie. This means the django server sees the user as logged out and trying to reach `/dashboard/` so it redirects to signin.

### Expected Behavior

We don't want to redirect to `/signin/`, but we can't change anything about the request either. So the proposed solution is to add another url (e.g. `/payment-callback`, the path doesn't mater much since it'll immediately redirect) that we send to cybersource as the redirect url for receipts and cancellation. This view will then redirect to `/dashboard/` along with a copy of the querystring from cybersource. This means we ignore the POST payload, but that was already being ignored anyway.

- [ ] Add a new view at `/payment-callback` that redirects to `/dashboard/` with the same querystring
   - This view should **not** require the user to be logged in
- [ ] Update the dashboard url passed in [here](https://github.com/mitodl/micromasters/blob/master/ecommerce/api.py#L231-L232) to point to the new view above.

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes #5113 

#### What's this PR do?
fix Cybersource cancellation or success redirect failures

### DIFF
--- a/ecommerce/views.py
+++ b/ecommerce/views.py
@@ -281,8 +281,7 @@ class UserCouponsView(APIView):
 
 class PaymentCallBackView(RedirectView):
     """
-    payment callback view that will redirect to dashboard
+    payment callback view that will redirect to dashboard url
     """
     url = DASHBOARD_URL
     query_string = True
-    permanent = True
--- a/ecommerce/views.py
+++ b/ecommerce/views.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.db import transaction
 from django.http.response import Http404
 from django.shortcuts import get_object_or_404
+from django.views.generic.base import RedirectView
 from ipware import get_client_ip
 from rest_framework.authentication import SessionAuthentication, TokenAuthentication
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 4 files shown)


ecommerce/views.py
```diff
@@ -275,3 +277,12 @@ def post(self, request, code, *args, **kwargs):  # pylint: disable=unused-argume
...
             )
+
+
+class PaymentCallBackView(RedirectView):
+    """
+    payment callback view that will redirect to dashboard
+    """
+    url = DASHBOARD_URL
+    query_string = True
+    permanent = True
```
rhysyngsun, line 288: This shouldn't be permanent because the browser could cache it which would probably work without issue now but could cause issues later if that changes.

rhysyngsun (changes requested):

rhysyngsun (approved):
LGTM, we'll test in RC since local testing is difficult.

umarmughal824:
> LGTM, we'll test in RC since local testing is difficult.

I had tested it locally, seems to be working fine

PR #5118, opened 2021-11-15
added sentry webpack plugin

### LINKED ISSUE #5117: add sentry-webpack-plugin
As an engineer, it would be easier to debug issues reported in Sentry if we had sourcemaps available. 

If we add `sentry-webpack-plugin` it should share sourcemaps with Sentry. 

#### related issues 

- #5106

Wassaf-Shahzad:
Closes #5117 

#### What's this PR do?
This PR adds a sentry Webpack plugin do add source map to sentry.

Source map should be uploaded in sentry.

#### Introduces a new settings in CI
This PR adds the following environment variables
    - ORG_NAME ( organisation name as defined in sentry)
    - PROJECT_NAME (name of the project as defined in sentry)
    - SENTRY_RELEASE (same as the one passed to sentry init call in sentry client)
    - SENTRY_AUTH_TOKEN (sentry auth token with project:write scope)


Wassaf-Shahzad:
i have updated the  webpack config and added some environment variables as well( I didn't chose a dot file as we could add the token in GitHub secrets). 
I do need one config value from sentry (SENTRY_AUTH token). I can create that on a developer level but still thought I should wait for you go ahead.

Wassaf-Shahzad:
Made a small change in webpack config for prod. I made the sentry plugin conditional so as not to upload releases to sentry during CI builds.

webpack.config.prod.js
```diff
@@ -52,7 +54,15 @@ module.exports = Object.assign(prodConfig, {
     new webpack.optimize.AggressiveMergingPlugin(),
     new MiniCssExtractPlugin({
       filename: "styles-[name]-[contenthash].css"
-    })
+    }),
+    ...process.env.IS_NOT_CI ? [new SentryWebpackPlugin({
+      authToken: process.env.SENTRY_AUTH_TOKEN,
+      org : process.env.ORG_NAME,
+      project : process.env.PROJECT_NAME,
+      release : process.env.SENTRY_RELEASE,
```
rhysyngsun, line 62: A few things here:
- You should open a ticket with devops to set these values
- Could you namespace them all under `SENTRY_*`?
- Currently the release version comes from `settings.py`, there's no environment variable for that and we don't currently have any automation around setting this var either. So I think some work needs to go into pulling that  out into a version file or otherwise getting it extracted somehow during the heroku build. The release bot expects it to live in `settings.py` though so the latter might be best for the time being.

rhysyngsun (changes requested):

webpack.config.prod.js
Wassaf-Shahzad: I have addressed the first two issues However I am not sure I understand you last point. Is there any related document or commit I could look in to

rhysyngsun: I've put up a fix to the release script here: https://github.com/mitodl/release-script/pull/358, that will allow you to move the version into a `VERSION` file and then replace the current definition of `VERSION` in `settings.py` with a few lines of code that read that file's contents. Then do the same in `webpack.config.prod.js` and use that for the value of `release`.

rhysyngsun: Ok, that PR is approved, but I'm going to hold off on merging it until after this merged, let me know when i can review this again.

Wassaf-Shahzad: Yeah, i had addressed the feedback a while back but forgot tor re-request the review. Sorry for the delay

rhysyngsun: @Wassaf-Shahzad I'm not seeing those changes here, specifically the usage of a `VERSION` file.

rhysyngsun (approved):
LGTM

PR #5119, opened 2021-11-16
Enroll user in an exam_run on edx and redirect to edX

### LINKED ISSUE #5111: Exams: support MITx Online exams
We need to be able to link learners to the exam course in MITx Online. Since we don't have an ecommerce flow configured on openedx, we can't use the `ExamRunCoupon` strategy already implemented. Instead, we'll need to have MicroMasters enroll the learner directly into the exam course run on openedx

- [ ] Add a new column to `ExamRun` to store the exam course run edx key, this should be configurable via django-admin
- [ ] Add a new configuration value for the service worker access token to use for enrolling users in the exam course
- [ ] Add a new view that accepts an exam run id (e.g. `/enroll-in-exam/<exam_run_id>/`, feel free to choose something better) and performs the following:
  - Verifies the learner is authorized for that exam
  - Enrolls the learner in the exam course run on openedx using the service worker access token
    - If the learner is already enrolled, skip this step 
  - Redirect the learner to the exam course run on openedx
- [ ] Update `dashboard.api.get_edx_exam_coupon_url` to redirect to the new view defined above if the exam openedx run id is set

annagav:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [ ] Testing
  - [ ] Code is tested

Fix #5111

#### What's this PR do?
When user is authorized for an ExamRum enroll the user and redirect them to the platform.

Enroll a user in a current course, create an ExamRun for that course and authorize the user to take the exam.

Screen Shot 2021-11-16 at 5 44 25 PM

Screen Shot 2021-11-16 at 5 44 34 PM


dashboard/views.py
```diff
@@ -143,6 +145,86 @@ def post(self, request):
...
+        """
+        enrolls the user in a exam run in edx
+        """
+        edx_exam_course_id = request.data.get('exam_course_id')
+
+        if edx_exam_course_id is None:
+            raise ValidationError('exam course id missing in the request')
+
+        exam_run = ExamRun.objects.get(edx_exam_course_key=edx_exam_course_id)
+        if not ExamAuthorization.objects.filter(user=request.user, status=ExamAuthorization.STATUS_SUCCESS).exists():
```
rhysyngsun, line 168: I think this needs to check for an exam authorization for the exam run?

dashboard/views.py
```diff
@@ -143,6 +145,86 @@ def post(self, request):
...
+    def post(self, request):
+        """
+        enrolls the user in a exam run in edx
+        """
+        edx_exam_course_id = request.data.get('exam_course_id')
+
+        if edx_exam_course_id is None:
+            raise ValidationError('exam course id missing in the request')
+
+        exam_run = ExamRun.objects.get(edx_exam_course_key=edx_exam_course_id)
```
rhysyngsun, line 167: This would be a bit cleaner if the `exam_course_id` was part of the url, and I would just do a simple `get_object_or_404` on it.

dashboard/views.py
```diff
@@ -143,6 +145,86 @@ def post(self, request):
...
+                data={'error': str(exc)}
+            )
+
+        # create an instance of the client to query edX
+        edx_client = EdxApi(user_social.extra_data, COURSEWARE_BACKEND_URL[provider])
+        enrollments = edx_client.enrollments.get_student_enrollments()
+        all_enrolled_course_ids = enrollments.get_enrolled_course_ids()
+        # if user already enrolled in this exam course
+        if ExamRun.objects.filter(edx_exam_course_key__in=all_enrolled_course_ids).exists():
+            return Response({'url': url})
```
rhysyngsun, line 192: I think this might cause problems if the user was enrolled in a previous exam run but is trying to take it a second time and isn't enrolled yet. We'd end up not enrolling them in the newer run but linking them to it.

dashboard/views.py
```diff
@@ -143,6 +145,86 @@ def post(self, request):
...
+    """
+    Create an exam enrollment for the user in a given exam run identified by course_id.
+    """
+    authentication_classes = (
+        authentication.SessionAuthentication,
+        authentication.TokenAuthentication,
+    )
+    permission_classes = (permissions.IsAuthenticated, )
+
+    def post(self, request):
```
rhysyngsun, line 158: This is a bit of a lengthy implementation for a view, if there's a way to push some of this into an api function and then do the error response handling in the view still that'd be a little cleaner.

rhysyngsun (changes requested):
Some initial feedback.

dashboard/views.py
annagav: I tried taking some things out, but the majority of the view is just error messages. Let me know if you think, we can cut something out.

dashboard/views.py
annagav: 👍

annagav:
@rhysyngsun I addressed your comments, thank you

rhysyngsun (changes requested):

rhysyngsun:
Also what we discussed yesterday - making the enrollment call via a staff user's api key (or whatever auth functionality accomplishes this securely).

rhysyngsun (approved):
LGTM

rhysyngsun:
@annagav looks like there's a test failure, but otherwise this looks good to me functionally.

PR #5122, opened 2021-11-24
switch to depending on our forked iso-3166-2 package

alicewriteswrongs:

none

#### What's this PR do?

I set up our forked repo (https://github.com/mitodl/iso-3166-2.js) to be published as an npm package (https://github.com/mitodl/iso-3166-2.js/pull/4) so we don't have to include the package from github directly. Doing this will let us upgrade to the newest version of yarn, and do some of the build / CI setup simplifications we've done on other projects.

The features that use this project should work as normal, and tests should pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "dependencies": {
     "@fancyapps/fancybox": "^3.5.7",
     "@material-ui/core": "^4.2.1",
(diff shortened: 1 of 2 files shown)


umarmughal824 (approved):
looks good 👍

Issue #5129, opened 2021-12-14
update frontend tooling


alicewriteswrongs:
A lot of the frontend stuff in Micromasters is pretty behind what we're using elsewhere. We're on a pretty old version of node, using flow, outdated versions of babel, etc. Some stuff we should take care of:

- [ ] update nodejs version
- [ ] update eslint config
- [ ] upgrade yarn to v3
- [ ] convert Flow code to Typescript

PR #5130, opened 2021-12-14
upgrade node to v14

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

part of #5129 

#### What's this PR do?

This upgrades our Node version to 14.18.2. This isn't fully updated (current version is 17.x.x) but this will get us to the point where we can upgrade yarn to v3 and run some other upgrades that will then, I think, let us upgrade to a more recent nodejs version and so on. We're climbing the ladder here :)

everything should work!


umarmughal824 (approved):
👍 looks good

PR #5146, opened 2022-01-07
update index users tasks to acks late so that last unfinished didn't get ignored

### LINKED ISSUE #5114: user has an account and profile, but does not appear in learner search
We have had multiple reports from customer support of users that have accounts, but cannot be found in the learner search. 

When they have alerted us to the issue, we have been able to locate the accounts via the django admin. Running the following code in the django shell will then add the user to the learner search without issue:

```
from search.tasks import index_users 
index_users([user.id],check_if_changed=True)
```

As far as I know, these users should be added to the index as soon as they complete their profile. Why is this failing, especially when running it in the shell corrects the issue immediately? I can't find any relevant errors in sentry.

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes #5114 

#### What's this PR do?
update index users tasks to acks late so that last unfinished didn't get ignored

just test everything is working and the build is passing

### DIFF
--- a/search/tasks.py
+++ b/search/tasks.py
@@ -87,7 +87,7 @@ def index_program_enrolled_users(program_enrollment_ids):
     post_indexing_handler(program_enrollments)
 
 
-@app.task
+@app.task(acks_late=True)
 def index_users(user_ids, check_if_changed=False):
     """
     Index users' ProgramEnrollment documents


arslanashraf7 (approved):
👍

PR #5147, opened 2022-01-10
Readme: how to update final grade for user

annagav:
none

#### What's this PR do?
Adds instructions on how to update final grade for user

View it [here](https://github.com/mitodl/micromasters/blob/1b7d5a0759549b9baab383d87e14c0c9decaae4f/grades/README.md)

### DIFF
--- a/grades/README.md
+++ b/grades/README.md
@@ -188,10 +188,10 @@ Course certificates are generated every hour (on the hour). If this is too long 
 
 
 ##Sync edx current cached grade and FinalGrade
-It happens that after the grades were frozen, account relinking can occur 
-which results in grades being out of sync, since `FinalGrade` doesn't get updated
-after grades had been frozen.
-
+Occasionally, a learner's grade may need to be updated manually. One way this
+could happen is if a learner's MITx Online account was linked to their Micromasters
+account after some grades were frozen. This would result in grades being out of sync,
+since `FinalGrade` doesn't get updated after grades had been frozen.
 ```
 from django.contrib.auth.models import User
 user = User.objects.get(username=username)
--- a/grades/README.md
+++ b/grades/README.md
@@ -185,3 +185,38 @@ the proctored exam being released. So you will need to run the management comman
 When a learner's fifth and final course certificate is created, it triggers the creation of a program certificate. 
 Course certificates are generated every hour (on the hour). If this is too long to wait, use the 
 `generate_program_certificates` management command. Each learner can only earn one Program certificate. 
+
+
+##Sync edx current cached grade and FinalGrade
+It happens that after the grades were frozen, account relinking can occur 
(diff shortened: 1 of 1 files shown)


grades/README.md
```diff
@@ -185,3 +185,39 @@ the proctored exam being released. So you will need to run the management comman
...
+##Sync edx current cached grade and FinalGrade
+It happens that after the grades were frozen, account relinking can occur 
+which results in grades being out of sync, since `FinalGrade` doesn't get updated
+after grades had been frozen.
+
+```
+from django.contrib.auth.models import User
+user = User.objects.get(username=username)
+
+# If you don't know which course_run you need to sinc you can find it here
```
pdpinch, line 199:
```suggestion
# If you don't know which course_run you need to sync you can find it here
```

grades/README.md
```diff
@@ -185,3 +185,39 @@ the proctored exam being released. So you will need to run the management comman
...
+
+from grades.constants import *
+final_grade, _ = FinalGrade.objects.update_or_create(user=user,course_run=course_run,
+    defaults={
+        'grade': edx_grade.grade,
+        'passed': edx_grade.passed,
+        'status': FinalGradeStatus.COMPLETE
+})
+final_grade.save()
+```
```
pdpinch, line 223: How/is this different from running `freeze_user_final_grade(user, course_run)` ?

(maybe it avoids the issue where an incorrect final grade already exists?)

pdpinch (approved):
a copy edit and a question. 

@briangrossman you should review this too

grades/README.md
annagav: the freeze_user_final_grade function would not update the existing grade, which of course can be modified. But this code was more intended for trouble shooting the problem manually, checking on the learners grades for the course and seeing if the grade just needs to be updated.

grades/README.md
```diff
@@ -185,3 +185,39 @@ the proctored exam being released. So you will need to run the management comman
 When a learner's fifth and final course certificate is created, it triggers the creation of a program certificate. 
 Course certificates are generated every hour (on the hour). If this is too long to wait, use the 
 `generate_program_certificates` management command. Each learner can only earn one Program certificate. 
+
+
+##Sync edx current cached grade and FinalGrade
+It happens that after the grades were frozen, account relinking can occur 
+which results in grades being out of sync, since `FinalGrade` doesn't get updated
+after grades had been frozen.
```
briangrossman, line 193: Slight wording suggestion

```suggestion
Occasionally, a learner's grade may need to be updated manually. One way this 
could happen is if a learner's MITx Online account was linked to their Micromasters 
account after some grades were frozen. This would result in grades being out of sync, 
since `FinalGrade` doesn't get updated after grades had been frozen.
```

annagav:
@briangrossman I addressed your comment. Thank you!

PR #5165, opened 2022-02-04
Stop automated updates of edX data for users in certain programs

### LINKED ISSUE #5124: Stop automated updates of edX data for users in certain programs
For users who are only enrolled in the DEDP program, there is no point in automatically updating their edX enrollments, grades and certificates any more. Any new data for these users will only be coming from mitxonline. 

However, if a user is enrolled in multiple programs, we should continue to update their edx.org data. 

If there is a reasonable way to filter out the DEDP-only users, this should be a configurable option, if/when other programs end on edX.

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings
  - [x] New settings are documented and present in `app.json`
  - [x] New settings have reasonable development defaults, if applicable

fixes #5124 

#### What's this PR do?
Stop automated updates of edX data for users in certain programs

Just need to confirm that only non DEDP( with no other enrollment but only in DEDP program) users are being passed to update edX data

You can change the flag to see if it's being sent to update edX for DEDP users if  `UPDATE_EDX_DATA_FOR_DEDP_PROGRAM_USERS=True`


pdpinch:
I'm not sure @umarmughal824 but I think this data is updated in two places:

1. In a periodic celery task that is run multiple times per day
2. User-initiated, when the user loads their dashboard 

The goal of the issue is to stop the automated update because it's very resource intensive. I think there are 60,000-80,000 users whose data we attempt to update multiple times per day. 

If it's possible to keep the dashboard update, that would be best, because we'd like to handle new users who have old edX data.

dashboard/api.py
```diff
@@ -42,6 +44,9 @@
 # pylint: disable=too-many-branches
 
 
+DEDP = "Data, Economics, and Development Policy"
```
annagav, line 47: Could you rename this to DEDP_PROGRAM_TITLE?

dashboard/api.py
```diff
@@ -684,10 +689,20 @@ def calculate_users_to_refresh_in_bulk():
         usercacherefreshtime__current_grade__gte=refresh_time_limit
     )
 
+    only_dedp_users = []
+    if not settings.UPDATE_EDX_DATA_FOR_DEDP_PROGRAM_USERS:
+        dedp_users = ProgramEnrollment.objects.filter(
+            program__title=DEDP
+        ).values_list('user__id', flat=True)
+        non_dedp_users = ProgramEnrollment.objects.filter(
+            Q(user__id__in=dedp_users), ~Q(program__title=DEDP)
```
annagav, line 698: could you add a test for this?

annagav: you could also do .exclude(program__title=DEDP).filter(user_id__in=dedp_users) which I think is more readable

annagav (approved):
👍

PR #5199, opened 2022-05-05
Rewrite the algorithm for exam attempts

### LINKED ISSUE #5198: Exams: Logic to determine if learner needs to pay for an exam doesn't work in some cases
### Steps to Reproduce

In certain scenarios, the logic to determine if a learner needs to pay for an exam authorization is incorrect, asking a learner to pay when they don't need to. This is due to the code that was introduced to go from two authorizations per order to one (https://github.com/mitodl/micromasters/issues/4646). 

Steps:
- Learner has an order for a DEDP course from prior to the date when purchases switched to a single authorization 
- Learner uses both authorization for the above order
- Learner orders the same DEDP course

### Expected Behavior

- Learner should have an authorization
   - The two used authorizations were both covered by the initial order
   - The new order should award them an authorization

### Actual Behavior

- The learner is not authorized

### Stacktrace

Here's some hints

Here's the code that determines if a learner needs to pay: https://github.com/mitodl/micromasters/blob/9f52ab55cf2381dcb6d886ddb78dca0ed9466b27/dashboard/api.py#L615-L639

Note the left side of the conditional is simply checking the number of authorizations used:
`return ExamAuthorization.objects.filter(user=mmtrack.user, course=course, exam_taken=True).count()` >= `attempt_limit`

Something like this might work.
`Round( 0.5 * number of authorizations used prior to exam_attempts_first_date) + number of authorizations used after exam_attempts_first_date` >= `attempt_limit`

Also, the code that issues authorizations should be checked as well.

### Related Issues

- https://github.com/mitodl/micromasters/issues/4646

annagav:
#### Pre-Flight checklist
- [ ] Testing
  - [x] Code is tested

Fix #5198

#### What's this PR do?
Update the algorithm for determining the number of attempts still available to the user and whether they have to pay for the course again.

Take a look at the tests, make sure all cases are covered. Tests should pass.
The new algorithm for determining the number of attempts left:
```
Find all payments before the "first date"
Find the number of used attempts before the "second date"
if the number of used attempts is less than the amount of attempts given:
    Compute the carry over amount of attempts, converting them from double attempts to single attempts
if the number of unused attempts (before second date) is 0:
    Then we only compute the number of attempts in the new way, i.e. payment = attempt
else
    # the user has more attempts than payments
    # this means the user made at least one payment and attempt both after the first date
    find the number of attempts without the double attempts and subtract that from new payments

### DIFF
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -244,9 +244,11 @@ class MMTrack:
 
     def get_number_of_attempts_left(self, course):
         """
-        Checks for each payment from before first date, when payment used to give double attempts
-        and find if any unused attempts are carried over to the single attempt era. Calculte the number of aytempts
-        left based on the carry over number of attempts.
+        Checks for each payment from before first date (when a payment provided two attempts)
+        and find if any unused attempts are carried over to the single attempt era (when a payment
+        provides one attempt).
+        Calculate the number of attempts remaining based on the calculated carry over and the other
+        payments and attempts.
 
         Args:
             course (courses.models.Course): a course
@@ -278,19 +280,25 @@ class MMTrack:
             exam_run__date_first_eligible__gte=self.program.exam_attempts_second_date
         ).count()
 
+        # calculate any unused attempts from for the period when one payment provided two attempts
         unused_double_attempts = old_payments * ATTEMPTS_PER_PAID_RUN_OLD - old_attempts
         if unused_double_attempts > 0:
             # the user has unused attempts from before the first date
-            # total payments after second date - total attempts after second date
+            # divide unused attempts from before the first date by two since payments are
+            # only valid for one attempt after the second date. This is the carryover.
             attempts_carryover = floor(unused_double_attempts / 2)
+            # Calculate the number of remaining attempts
+            #   the number of payments after the first date
+            #   minus the number of attempts after the second date
+            #   plus the carryover attempts
             return new_lines - new_attempts + attempts_carryover
         elif unused_double_attempts == 0:
             # there is no carry over
             return new_lines - new_attempts
 
         else:
-            # the user has more attempts than payments
-            # this means the user made at least one payment and attempt both after the first date
+            # the user has more used attempts (before the second date) than payments (before the first date)
+            # this can happen if the user made at least one payment and attempt both after the first date
             # find the number of attempts without the double attempts and subtract that from new payments
             single_attempts = used_attempts_qset.count() - old_payments * ATTEMPTS_PER_PAID_RUN_OLD
             return new_lines - single_attempts
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -260,7 +260,6 @@ class MMTrack:
             course_key__in=course.courserun_set.values('edx_course_key')
         ).order_by('created_at')
         used_attempts_qset = ExamAuthorization.objects.filter(user=self.user, course=course, exam_taken=True)
-
         # if for some reason the first date is not set, return the difference between payments and used attempts
         if first_date is None:
             return payments_qset.count() - used_attempts_qset.count()
--- a/dashboard/utils.py
+++ b/dashboard/utils.py
@@ -3,6 +3,7 @@ Utility functions and classes for the dashboard
 """
 import logging
 from decimal import Decimal
+from math import floor
 
 from django.db import transaction
 from django.db.models import Q, Count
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dashboard/utils.py [hunk 1]
```diff
@@ -241,25 +242,59 @@ def get_payments_count_for_course(self, course):
...
         ).values('order_id').distinct().count()
 
-    def get_custom_number_of_attempts_for_course(self, course):
+    def get_number_of_attempts_left(self, course):
         """
-        Get a custom number of exam attempts for given course. If payment was made before
-        the first date, learner get 2 attempts. Otherwise, 1.
+        Checks for each payment from before first date, when payment used to give double attempts
+        and find if any unused attempts are carried over to the single attempt era. Calculte the number of aytempts
+        left based on the carry over number of attempts.
```
briangrossman, line 249:
```suggestion
        Checks for each payment from before first date (when a payment provided two attempts)
        and find if any unused attempts are carried over to the single attempt era (when a payment
        provides one attempt). 
        
        Calculate the number of attempts remaining based on the calculated carry over and the other 
        payments and attempts.
```

dashboard/utils.py
```diff
@@ -241,25 +242,59 @@ def get_payments_count_for_course(self, course):
...
+        # number of used attempts before the second date
+        old_attempts = used_attempts_qset.filter(
+            exam_run__date_first_eligible__lt=self.program.exam_attempts_second_date
+        ).count()
+        # number of used attempts after the second date
+        new_attempts = used_attempts_qset.filter(
+            exam_run__date_first_eligible__gte=self.program.exam_attempts_second_date
+        ).count()
+
+        unused_double_attempts = old_payments * ATTEMPTS_PER_PAID_RUN_OLD - old_attempts
```
briangrossman, line 281:
```suggestion
        # calculate any unused attempts from for the period when one payment provided two attempts
        unused_double_attempts = old_payments * ATTEMPTS_PER_PAID_RUN_OLD - old_attempts
```

dashboard/utils.py
```diff
@@ -241,25 +242,59 @@ def get_payments_count_for_course(self, course):
...
+        if unused_double_attempts > 0:
+            # the user has unused attempts from before the first date
+            # total payments after second date - total attempts after second date
+            attempts_carryover = floor(unused_double_attempts / 2)
+            return new_lines - new_attempts + attempts_carryover
```
briangrossman, line 286:
```suggestion
            # the user has unused attempts from before the first date
            
            # divide unused attempts from before the first date by two since payments are 
            # only valid for one attempt after the second date. This is the carryover.
            attempts_carryover = floor(unused_double_attempts / 2)
            
            # Calculate the number of remaining attempts
            #   the number of payments after the first date
            #   minus the number of attempts after the second date
            #   plus the carryover attempts
            return new_lines - new_attempts + attempts_carryover
```

dashboard/utils.py
```diff
@@ -241,25 +242,59 @@ def get_payments_count_for_course(self, course):
...
+            # there is no carry over
+            return new_lines - new_attempts
 
-        return num_attempts
+        else:
+            # the user has more attempts than payments
+            # this means the user made at least one payment and attempt both after the first date
+            # find the number of attempts without the double attempts and subtract that from new payments
+            single_attempts = used_attempts_qset.count() - old_payments * ATTEMPTS_PER_PAID_RUN_OLD
+            return new_lines - single_attempts
```
briangrossman, line 296:
```suggestion
        else:
            # the user has more used attempts (before the second date) than payments (before the first date)
            # this can happen if the user made at least one payment and attempt both after the first date
            
            # find the number of attempts without the double attempts and subtract that from new payments
            single_attempts = used_attempts_qset.count() - old_payments * ATTEMPTS_PER_PAID_RUN_OLD
            return new_lines - single_attempts
```

dashboard/utils.py
```diff
@@ -241,25 +242,66 @@ def get_payments_count_for_course(self, course):
...
         ).order_by('created_at')
-        first_date = course.program.exam_attempts_first_date
-        num_attempts = sum([ATTEMPTS_PER_PAID_RUN_OLD
-                            if line.modified_at < first_date else ATTEMPTS_PER_PAID_RUN for line in lines])
+        used_attempts_qset = ExamAuthorization.objects.filter(user=self.user, course=course, exam_taken=True)
+        # if for some reason the first date is not set, return the difference between payments and used attempts
+        if first_date is None:
+            return payments_qset.count() - used_attempts_qset.count()
+
+        # number of payments before the fist date
```
arslanashraf7, line 269:
```suggestion
        # number of payments before the first date
```

dashboard/utils.py
```diff
@@ -241,25 +242,66 @@ def get_payments_count_for_course(self, course):
...
+        old_payments = payments_qset.filter(modified_at__lt=first_date).count()
+        # number of payments after the first date
+        new_lines = payments_qset.filter(modified_at__gte=first_date).count()
```
arslanashraf7, line 272: We can use `new_payments` here instead since we are using `old_payments ` on the previous line.

dashboard/utils.py
```diff
@@ -241,25 +242,66 @@ def get_payments_count_for_course(self, course):
...
-        first_date = course.program.exam_attempts_first_date
-        num_attempts = sum([ATTEMPTS_PER_PAID_RUN_OLD
-                            if line.modified_at < first_date else ATTEMPTS_PER_PAID_RUN for line in lines])
+        used_attempts_qset = ExamAuthorization.objects.filter(user=self.user, course=course, exam_taken=True)
+        # if for some reason the first date is not set, return the difference between payments and used attempts
+        if first_date is None:
+            return payments_qset.count() - used_attempts_qset.count()
+
+        # number of payments before the fist date
+        old_payments = payments_qset.filter(modified_at__lt=first_date).count()
```
arslanashraf7, line 270: What happens if the first_date(exam_attempts_first_date) is not `set/None`. I think the query might throw errors.

dashboard/utils.py
```diff
@@ -241,25 +242,66 @@ def get_payments_count_for_course(self, course):
             course_key__in=course.courserun_set.values('edx_course_key'),
         ).values('order_id').distinct().count()
 
-    def get_custom_number_of_attempts_for_course(self, course):
+    def get_number_of_attempts_left(self, course):
```
arslanashraf7, line 245: I'm sorry if I'm missing something basic but I have a confusion here for a case with the new Logic:

A user placing an Order before the first date and then using both his attempts before the first date.

Now the user places another order after the second date as a scenario mentioned in [this doc](https://docs.google.com/document/d/1A_fZemxKx6haTJj0euWgaZXjFvcsWX7RgP6cIJR8XbY). With the new order shouldn't it return at least 1 attempt?
This is related to how/when we create the `ExamAuthorization` for an Order, I believe that part is [here](https://github.com/mitodl/micromasters/blob/master/exams/api.py#L84-L89). Looking at this will it create the new ExamAuthorization for a New Order with our updated logic?

arslanashraf7 (changes requested):
I requested some minor changes and a question based on the code review.

dashboard/utils.py
annagav: there is a check for `if first_date is None:` right above the line

dashboard/utils.py
```diff
@@ -241,25 +242,66 @@ def get_payments_count_for_course(self, course):
...
+            return payments_qset.count() - used_attempts_qset.count()
+
+        # number of payments before the fist date
+        old_payments = payments_qset.filter(modified_at__lt=first_date).count()
+        # number of payments after the first date
+        new_lines = payments_qset.filter(modified_at__gte=first_date).count()
+
+        # number of used attempts before the second date
+        old_attempts = used_attempts_qset.filter(
+            exam_run__date_first_eligible__lt=self.program.exam_attempts_second_date
```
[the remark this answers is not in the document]
annagav: there is a check for first_date above, and there is an assumption that both of them have to be set, because that is how the transition was made to single attempt per payment.

dashboard/utils.py
annagav: Yes, if a learner had a payment and two attempts before the first date, if they make another purchase they get another exam attempt and has_to_pay_for_exam would return False.

dashboard/utils.py
arslanashraf7: Thanks for pointing to that, I definitely missed that.

dashboard/utils.py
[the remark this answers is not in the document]
arslanashraf7: I see, If this was assumed in the previous work it's fine. Although we should have not allowed blank/null for this date making sure that Django admin doesn't allow empty end dates. I'm fine either way. But I'm fine either way. Thanks for clarifying that.

dashboard/utils.py
arslanashraf7: Thanks for adding the details. Just let me know when the PR is ready for review again

dashboard/utils.py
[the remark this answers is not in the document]
annagav: it would be null for programs that don't have any exams

annagav:
@arslanashraf7 this is ready for another review. Thanks

arslanashraf7 (approved):
Looks good 👍

Issue #5213, opened 2022-08-09
Store order reference numbers in the database


rhysyngsun:
Related to https://github.com/mitodl/mitxonline/issues/799, we need to store the reference numbers for import. See that issue for more details.

To address this, we should do the following:

- [ ] Add a new `generated_reference_number` column to `Order`.
- [ ] Update `Order` creation code to write to this column when an order is made.
- [ ] Backpopulate this column with values based on how `Order.generated_reference_number` would be computed.

PR #5214, opened 2022-08-09
Added order reference numbers

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Part of #5213 

#### What's this PR do?

This adds the `reference_number` column and the code to populate that value on create going forward.

- Go to your dashboard and perform a checkout
- You only need to get as far as being redirected over to CyberSource (you don't need a functional Cybersource integration), once that happens you can go to django-admin and confirm your order has a reference number populated


ecommerce/models.py
```diff
@@ -68,6 +68,8 @@ class Order(AuditableModel):
     modified_at = DateTimeField(auto_now=True)
     total_price_paid = DecimalField(decimal_places=2, max_digits=20)
 
+    reference_number = CharField(null=True, unique=True, max_length=255)
```
arslanashraf7, line 71: Is this something we might want to add to Django Admin's [display list of Order model](https://github.com/mitodl/micromasters/blob/master/ecommerce/admin.py#L41).

arslanashraf7 (approved):
Looks good 👍 , I was also able to test. I just added a small question/suggestion.

ecommerce/models.py
rhysyngsun: I think for now we'll leave it to keep the changeset down and we haven't needed it yet.

Issue #5212, opened 2022-08-08
Temporarily set dashboard updates to run weekly instead of 4x per day 


pdpinch:
We update the cached edx (and mitxonline) data for all users every six hours. These tasks can interfere with other scheduled tasks, like finalizing grades and created exam authorizations.

We should change the frequency to once a week, on Fridays, before the next DEDP exam registration 8/16 (grades frozen) to 8/17 (exam authorizations created).  

We should put the dashboard update tasks on hold before the week of 815 to focus our resources on final grade updates and exam authorizations.

### acceptance criteria

- [x] one PR to change the task to once a week. 
- [ ] a 2nd PR to revert change. The 2nd PR should include notes about the timing and criteria for merging it.

pdpinch:
The PR to revert this change hasn't been written. I don't know how serious an issue it is though.

PR #5217, opened 2022-08-11
only update user dashboards on Friday

pdpinch:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

part of #5212 

#### What's this PR do?

Stops updating user dashboard data every day. Only updates user dashboard data on Fridays. 

I don't think it can be manually tested

We need to authorize exams on Tuesday 8/16. We don't want the exam authorization task to be slowed down waiting for these dashboard updates.

### DIFF
--- a/micromasters/settings.py
+++ b/micromasters/settings.py
@@ -480,7 +480,7 @@ CELERY_TASK_ALWAYS_EAGER = get_bool("CELERY_TASK_ALWAYS_EAGER", False) or get_bo
 CELERY_TASK_EAGER_PROPAGATES = (get_bool("CELERY_TASK_EAGER_PROPAGATES", True) or
                                 get_bool("CELERY_EAGER_PROPAGATES_EXCEPTIONS", True))
 CELERY_BEAT_SCHEDULE = {
-    'batch-update-user-data-every-6-hrs': {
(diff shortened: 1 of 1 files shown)


micromasters/settings.py
```diff
@@ -482,7 +482,7 @@
 CELERY_BEAT_SCHEDULE = {
     'batch-update-user-data-every-6-hrs': {
```
arslanashraf7, line 483: Think we should update the task name to match the updated schedule.

```suggestion
    'batch-update-user-data-every-friday-every-6-hrs': {
```

arslanashraf7 (approved):
👍 Changes look reasonable, Just a comment, I can add that change in a new commit to this PR if you'd like. Let me know.

pdpinch:
I accepted your change. If we change the task name, does it need to be updated anywhere else? I searched but didn't find anything.

arslanashraf7:
> I accepted your change. If we change the task name, does it need to be updated anywhere else? I searched but didn't find anything.

I don't think it would need to be reflected anywhere else in code though. I do think that it will be loaded when the celery restarts.

arslanashraf7 (approved):
👍

Issue #5226, opened 2022-09-29
Change the text references to "edx.org" to be "MITx Online"


pdpinch:
this should be based on the courses' and program's courseware backend

<img width="775" alt="image" src="https://user-images.githubusercontent.com/430126/193129005-086da0d4-85d5-486f-85ac-b177ad0ba045.png">

PR #5231, opened 2022-10-04
Remove mention of edx.org

annagav:
Fix https://github.com/mitodl/micromasters/issues/5226
#### What's this PR do?
Remove mention of edx.org

Checkout the dashboard of your learner.

Before:
Screen Shot 2022-10-04 at 11 07 34 AM

After:
Screen Shot 2022-10-04 at 11 07 49 AM

### DIFF
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -16,7 +16,10 @@ import {
   STATUS_PASSED,
   EDUCATION_LEVELS,
   PROFILE_STEP_LABELS,
-  PROFILE_STEP_ORDER
+  PROFILE_STEP_ORDER,
+  COURSEWARE_BACKEND_MITXONLINE,
+  COURSEWARE_BACKEND_EDXORG,
+  COURSEWARE_BACKEND_NAMES
 } from "../constants"
 import type {
   Profile,
@@ -575,6 +578,18 @@ export function sortedCourseRuns(program: Program): Array<CourseRun> {
   )
 }
 
+export function programBackendName(program: Program): string {
+  const courses = program.courses
+  if (
+    R.any(
+      R.equals(COURSEWARE_BACKEND_MITXONLINE),
+      R.pluck("courseware_backend", R.unnest(R.pluck("runs", courses)))
+    )
+  ) {
+    return COURSEWARE_BACKEND_NAMES[COURSEWARE_BACKEND_MITXONLINE]
+  }
+  return COURSEWARE_BACKEND_NAMES[COURSEWARE_BACKEND_EDXORG]
+}
 // lets you edit both keys and values
 // just pass a function that expects and returns [key, value]
 export const mapObj = R.curry((fn, obj) =>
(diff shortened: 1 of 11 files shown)


arslanashraf7 (commented):
These changes look fine as per the scenario that we'll be retiring the MicroMasters soon and we have hardcoded it. 

But the ticket still mentions that we should decide on this text based on the courseware backend so I'm confused if we should be based on the backend. Let's get confirmation on this from @pdpinch .

annagav:
@arslanashraf7 @pdpinch since the courseware backend is individual per course run, in this PR I am determining the backend for this message based on hopefully valid assumption: If at least one course run in the program is on MITx Online, then the message is going to refer to MITx Online. Otherwise, it's edX.

static/js/util/util.js
```diff
@@ -575,6 +578,18 @@ export function sortedCourseRuns(program: Program): Array<CourseRun> {
...
+  const courses = program.courses
+  if (
+    R.any(
+      R.equals(COURSEWARE_BACKEND_MITXONLINE),
+      R.pluck("courseware_backend", R.unnest(R.pluck("runs", courses)))
+    )
+  ) {
+    return COURSEWARE_BACKEND_NAMES[COURSEWARE_BACKEND_MITXONLINE]
+  }
+  return COURSEWARE_BACKEND_NAMES[COURSEWARE_BACKEND_EDXORG]
```
arslanashraf7, line 591: I was looking around and i found a place where we already have an added property as for checking if a program is MITxOnline based where we already have a logic to calculate, I think we can reuse that to keep it in sync everywhere. So what it will be like is:

1. After [this line](https://github.com/mitodl/micromasters/blob/master/dashboard/api.py#L196) we add a property like: 
```
"has_mitxonline_courses": mmtrack.program.has_mitxonline_courses
```

2. After [this line](https://github.com/mitodl/micromasters/blob/master/static/js/flow/programTypes.js#L43) we add :
```
has_mitxonline_courses:     boolean
```

3. Our fuction `programBackendName` will change something like:
```
export function programBackendName(program: Program): string {
  if (program.has_mitxonline_courses) {
    return COURSEWARE_BACKEND_NAMES[COURSEWARE_BACKEND_MITXONLINE]
  }
  else {
    return COURSEWARE_BACKEND_NAMES[COURSEWARE_BACKEND_EDXORG]
  }
}
```

arslanashraf7 (commented):
I think we can standardize the logic for checking if the program is MITxOnline backend based. I added a comment. @annagav What do you think of that?

arslanashraf7:
@annagav To pace up and get it merged on Tuesday, For my above suggestion, I've created a branch and a commit in it. (This is the commit https://github.com/mitodl/micromasters/commit/fd43d6a4247c3c774c4cf107dc485e4641d57d2d), If this seems feasible you could just cherry-pick this commit into this branch, I've fixed the tests and stuff accordingly. Let me know what you think.

arslanashraf7 (approved):
Looks great 👍  thanks for making all the changes.

Issue #5227, opened 2022-09-29
restore the link to the program record for DEDP


pdpinch:
### acceptance criteria

- [ ] only for DEDP
- [ ]  The program record link can be conditional on the existence of combined final grades.

pdpinch:
@annagav I'm seeing this for some users, but not all. What is the condition(s) for when the link gets displayed?

annagav:
Could you provide an example user foe whom this doesn't work?

pdpinch:
See https://odl.zendesk.com/agent/tickets/135560

pdpinch:
Oh wait. That's different. Sorry.

pdpinch:
There's no link for my user.

pdpinch:
And there's no link for the user in https://odl.zendesk.com/agent/tickets/135615

PR #5232, opened 2022-10-05
Updating usages of fin_aid_availability

annagav:
Fix https://github.com/mitodl/micromasters/issues/5227

#### What's this PR do?
Updated all usages of the program field `financial_aid_availability`, that are not related to financial aid so much as much as to the existence of exams, combined final grades, and payments on our MM portal for courses.
In the past all of these were attributes of only one program: DEDP. But now that we are switching the payment system off, and financial aid is no longer available.

Run the app, nothing should break. Go the the dashboard make sure everything works. 
You need to create some exam runs for any courses in your program. 
The "Digital Learning" program by default has 'financial_aid_availability' turned on.

```
from courses.models import *
from exams.factories import *
course = Course.objects.filter(title='Digital Learning 100').first()
exam_run = ExamRunFactory.create(course=course, authorized=True, scheduling_past=False, scheduling_future=False)
```

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -184,6 +184,7 @@ def get_info_for_program(mmtrack):
         "description": mmtrack.program.description,
         "title": mmtrack.program.title,
         "financial_aid_availability": mmtrack.financial_aid_available,
+        "has_exams": mmtrack.has_exams,
         "courses": [],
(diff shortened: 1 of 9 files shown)


arslanashraf7 (approved):
Looks good 👍 , I've added some very minor comments.

Issue #5235, opened 2022-10-07
List of users who have paid, but have not yet passed a course


pdpinch:
As a user who has paid for DEDP course, I need a discount/enrollment code for mitxonline, so I can enroll as verified and take a proctored exam. 

This is similar to https://github.com/mitodl/mitxonline/issues/961 but on that issue we limited the list to users who were authorized. This omitted anyone who had not yet passed the course. 

#### Acceptance Criteria:

- [ ] csv of username, email, course key

#### related issues

- use this list as an input for creating discount codes in mitxonline

arslanashraf7:
I was looking around and it looks like @annagav might already have created a command that pretty much does the same and doesn't take exam authorizations into account for identifying upgradable users. That command is [list_currently_enrolled_paid](https://github.com/mitodl/micromasters/blob/master/exams/management/commands/list_currently_enrolled_paid.py). Is this exactly what we need? @annagav what's your opinion on this?

annagav:
@pdpinch as I am understanding, the previous command where we were picking authorized users, is a subset of all users that paid and haven't used their attempt. So it will be one list of all users paid, and need a coupon.

annagav:
This is not completed yet.

PR #5239, opened 2022-10-12
List learners paid, have exam, not enrolled

annagav:
Fix https://github.com/mitodl/micromasters/issues/5235

#### What's this PR do?
A management command to receive a file with a List of learners paid, have exam, not enrolled.

Run` ./manage.py list_paid_learners.py`


exams/management/commands/list_paid_learners.py
```diff
@@ -0,0 +1,70 @@
...
+            enrolled_user_ids = set(CachedEnrollment.get_cached_users(next_course_run))
+            paid_not_enrolled_users_ids = list(Line.objects.filter(
+                order__status__in=Order.FULFILLED_STATUSES,
+                course_key__in=course.courserun_set.values('edx_course_key'),
+            ).exclude(order__user__in=enrolled_user_ids).values_list('order__user', flat=True))
+            paid_not_enrolled_users = User.objects.filter(id__in=paid_not_enrolled_users_ids)
+            for user in paid_not_enrolled_users:
+                mmtrack = get_mmtrack(user, course.program)
+                if not has_to_pay_for_exam(mmtrack, course):
+                    username = get_social_username(user, BACKEND_MITX_ONLINE)
```
arslanashraf7, line 53: If `username` is None, it still adds the users to the list with None username which I think it shouldn't or might be overthinking in this.

arslanashraf7 (approved):
Looks good 👍  I added a small comment.

arslanashraf7:
I just noticed, The original ticket has been closed https://github.com/mitodl/micromasters/issues/5235

exams/management/commands/list_paid_learners.py
annagav: we do want to keep the record even if it is none, I think it is helpful for looking people up, even if the don't have mitxonline social auth.

annagav:
@pdpinch In this PR we are combining all learners that need a coupon into one list. Instead of the "authorized" list, that included only people who have pass the course before, this will be a list of people who paid, did not use their attempt, and are not currently enrolled.

Issue #5257, opened 2022-12-01
Enhance Program letter generation w.r.t their active status


arslanashraf7:
This won't be the case for our regular flow since the letters will only be generated when a program is completed, but due to a bug in the past, a lot of program letters were created which shouldn't have been. We invalidated those in https://github.com/mitodl/micromasters/pull/5253. We also introduced the is_active field there and made amendments in the code to serve the letters based on their active status only.

The existing implementation of the [letter generation function](https://github.com/mitodl/micromasters/blob/master/grades/api.py#L282) checks if a letter already exists before creating a new one. This could be a problem for the letters that we invalidated as mentioned above it's possible that those users can complete these program(s) at some point in the future. In that case, we should mark their existing letter as active instead of creating a new one. So [this logic](https://github.com/mitodl/micromasters/blob/master/grades/api.py#L282) needs to be changed a bit to handle that case.

### Steps to Reproduce
You need manual intervention to generate this case.

1. Create a Program letter for a user manually and mark it as inactive, also make sure that the program is not complete for that user
2. Now complete the program, add a final grade entry and you should notice that the Program letter is still inactive.

### Expected Behavior
For existing letters that are inactive, when the user completes the program they should be marked active

### Actual Behavior
The existing letters won't be marked active in case of completion

PR #5253, opened 2022-11-23
fix: Validate program letters and serve active letters only

arslanashraf7:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [ ] Mobile width screenshots
- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested
- [ ] Settings
  - [ ] New settings are documented and present in `app.json`
  - [ ] New settings have reasonable development defaults, if applicable

#### What are the relevant tickets?
#5250

#### What's this PR do?
- Adds a new field `is_active` in `MicromastersProgramCommendation`.
- Validates the existing Program letters based on program completion and marks them as `active or inactive`.
- Updates the logic to serve the program letters on basis of `is_active` through the URL.
- Updates the logic on the dashboard to show `View Program Letter` link only if Program letter is active.

#### How should this be manually tested?
- Before moving to this PR make sure that you have some program letters in your local setup, Create a couple for which the user has completed the program, and also create some letters for which the program isn't complete
- Now, Move to this PR, And see that the migration is successful and the Program letters that don't have completed the program. are marked as `is_active=False` (Verify this behavior)
- Now open the dashboard for this Program and confirm that you only see `View Program Letter` link for active letters(`MicromastersProgramCommendation`)
- Now open a program letter for an active letter it should be opened in a link something like `http://mm.odl.local:8079/letter/program/0b91f501-0583-4171-ac8e-7b82a4cf3c59`
- Now mark the above letter as inactive, Make sure that there is no link(`View Program Letter`) on the dashboard, and also opening the letter directly from URL(e.g. `http://mm.odl.local:8079/letter/program/0b91f501-0583-4171-ac8e-7b82a4cf3c59`) throws 404.

#### Where should the reviewer start?
(Optional)

#### Any background context you want to provide?
(Optional)

#### Screenshots (if appropriate)
image

When a program letter is active(Ignore the completion of courses, i created this manually):
image

When a program letter is Inactive:
image

PR #5262, opened 2022-12-02
fix: enhance the program letter generation for active status

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#5257 

#### What's this PR do?
- Enhances the logic to generate letters and also activate existing ones

There are a couple of test cases for this:

1. Create an inactive program letter for a **fa** program, Now let's complete the program by passing all the grades and make sure a certificate object is created, also make sure that this existing letter has been activated.
2. Create an inactive Program letter for a **non fa** program, Now let's add final grades for the user and complete the program for the user. The program letter should be activated.
3. Now perform the above two steps without an existing letter, in this case instead of activating, a new letter should be created
4. This is the sum of all above, the logged information should be matched. e.g. the log message should be clear about when a letter is created and when it's activated.
5. Optionally, make sure that the active letter can be seen and in the active letter shows 404 to the user. (There is no specific change in its logic but just a smoke test on this would be good)

Create program letter

This is kind of a second part of https://github.com/mitodl/micromasters/pull/5253

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -283,19 +283,20 @@ def generate_program_letter(user, program):
         log.info('User [%s] already has a letter for program [%s]', user, program)
         return
 
-    if (program.financial_aid_availability and
-            MicromastersProgramCertificate.objects.filter(user=user, program=program).exists()):
-        MicromastersProgramCommendation.objects.update_or_create(user=user, program=program,
-                                                                 defaults={"is_active": True})
-        return
-    if completed_program(user, program):
-        MicromastersProgramCommendation.objects.update_or_create(user=user, program=program,
-                                                                 defaults={"is_active": True})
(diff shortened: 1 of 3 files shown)


annagav (approved):
LGTM 👍 

Could you please update the description for `generate_program_letter`.

arslanashraf7:
>Could you please update the description for generate_program_letter.

I updated the description in https://github.com/mitodl/micromasters/pull/5262/commits/b40e57ea4671ac0bd1331ccafc7e5e1e8cf19dde. @annagav Does this look good to you?

arslanashraf7:
I merged the PR but if you'd have any suggestions regarding the message we can do in a separate PR.

Issue #5271, opened 2022-12-12
Update dashboard api course status 


annagav:
The user dashboard calculates  status for each course in the program.

In the past payments for courses with exams used to be paid on micromasters. Therefore calculation of course status relies on payment information.

Since the payments have moved to MITxOnline, we have to switch to rely on edx enrollment data for enrollment status of the course.

In particular in this function `is_enrolled_mmtrack`

PR #5274, opened 2022-12-19
Update has_paid to use edx data

annagav:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
Fix https://github.com/mitodl/micromasters/issues/5271

#### What's this PR do?
Since the payments have moved to MITxOnline, we have to switch to rely on edx enrollment data for enrollment status of the course.
`has_paid` is now checking for verified status of the enrollment to mean the user paid.

(Required)

make sure the course you are testing with has exams
`exam_run = ExamRunFactory.create(course=course, authorized=True, scheduling_past=False, scheduling_future=False)`
enroll your user in a course in 'audit' mode:
`./manage.py alter_data set_to_enrolled --username staff --course-title 'Analog Learning 100' --audit`

Then manually upgrade their cached enrollment to verified.

### DIFF
--- a/dashboard/fixtures/user_enrollments.json
+++ b/dashboard/fixtures/user_enrollments.json
@@ -62,9 +62,9 @@
                     "description": null,
                     "expiration_datetime": null,
                     "min_price": 0,
-                    "name": "Audit",
+                    "name": "Verified",
                     "sku": null,
-                    "slug": "audit",
+                    "slug": "Verified",
                     "suggested_prices": ""
                 }
             ],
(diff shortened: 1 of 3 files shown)


arslanashraf7 (approved):
Looks good 👍 . I have some thoughts though.

>Then check their dashboard api, the course status should be "status": "currently-enrolled", instead of "status": "can-upgrade",

Following the steps mentioned in the PR description, I was able to test the above, So this works.

Some other thoughts, I still see 
image

I was assuming that I would not see the `Pay Now` button since my enrollment is now paid. So, I did some digging around and noticed that the response from dashboard API has two fields that defined the payment 1) for a course 2) for a course run 
dashboard_api
 And the course payment indicator is based on if I have any exam attempts or not which I don't. Just wanted to make sure to pass this information through you and see if this behavior is expacted.

PR #5286, opened 2023-01-13
Remove the display of coupon messages

### LINKED ISSUE #5281: Messages about coupons
There seems to be some messages related to discounts on courses. Given that MM does not get payments for courses anylonger, are they still relevant?

Screen Shot 2023-01-09 at 2 24 28 PM

annagav:
Fix #5281

#### What's this PR do?
Removes just the rendering of the coupon messages

Nothing should break.

### DIFF
--- a/static/js/components/dashboard/courses/StatusMessages.js
+++ b/static/js/components/dashboard/courses/StatusMessages.js
@@ -102,7 +102,6 @@ const enrollmentDateMessage = (run: CourseRun) => {
 // sometimes there really isn't a better way :/
 export const calculateMessages = (props: CalculateMessagesProps) => {
   const {
-    coupon,
     courseAction,
     financialAid,
     hasFinancialAid,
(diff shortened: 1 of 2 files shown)


static/js/components/dashboard/courses/StatusMessages.js
```diff
@@ -10,9 +10,7 @@ import type {
   CourseRun,
   FinancialAidUserInfo
 } from "../../../flow/programTypes"
-import { makeCouponMessage } from "../../../lib/coupon"
```
arslanashraf7, line 13: I think we should remove this function from `lib/coupon.js` as well.

arslanashraf7 (approved):
👍

static/js/components/dashboard/courses/StatusMessages.js
annagav: Yes, we can remove it. But I think It would be better to include that in removing the coupons completely. There is a lot refactoring that is involved so I am trying to avoid it for now.

Issue #5301, opened 2023-02-06
Update freezing of final grades for MITxonline courses


annagav:
When freezing courses for MITxOnline backend we want to update the code not to check for certificate.

PR #5303, opened 2023-02-08
Update computation of final grades for courses with exams

annagav:
Fix https://github.com/mitodl/micromasters/issues/5301

#### What's this PR do?
Update the computation of the final grade for courses that have exams during the freeze process of final grades.

Take a look at the code and the tests, that they make sense.

The main change is that we don't want to rely on certificate data when we are freezing grades for MITxonline courses.
Few assumptions made:

- [ ] If the enrollment status is ['verified', 'honor'], then paid_on_edx is true 
- [ ] If the current grade passed=True then we the learner has a passing final grade, and also have passed the course.

nothing should brake.

### DIFF
--- a/grades/api.py
+++ b/grades/api.py
@@ -39,14 +39,8 @@ def _compute_grade_for_fa(user_edx_run_data):
         UserFinalGrade: a namedtuple of (float, bool,) representing the final grade
             of the user in the course run and whether she passed it
     """
-    if user_edx_run_data.certificate is not None:
-        run_passed = user_edx_run_data.certificate.status == 'downloadable'
-        # the following line should be updated when
-        # we add support for honor enrollments and certificates in the edx-api-client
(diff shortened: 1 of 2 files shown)


pdpinch:
> If the enrollment status is ['verified', 'honor'], then paid_on_edx is true

This is correct. "honor" certificates were a legacy certificate mode before edX started taking payment circa 2015. 

> If the current grade passed=True then we the learner has a passing final grade, and also have passed the course.

This is _almost always_ correct. The one exception will be the DEDP 3T2022 courses, when the course team didn't plan ahead for changes to the grading policy. We can handle this manually by updating the passing state of the users who "passed" the course but did not pass the final exam. We should probably write up an issue to address this, and implement it as soon as possible after this PR merges.

arslanashraf7 (approved):
Looks good 👍

PR #5331, opened 2023-04-14
Prompt learner to reauthenticate witht the correct social auth if invalid

### LINKED ISSUE #5327: Dashboard redirects to original edxorg login when mitxonline auth is needed.
### Steps to Reproduce

- Ensure your refresh token is expired in openedx
- Try to load the dashboard

### Expected Behavior

The user is prompted to login with the appropriate backend (edxorg or mitxonline) depending on which requires fresh credentials.

### Actual Behavior

The user is always redirected to the default (edxorg) login.

### Proposed Fix

I think we should update the dashboard API to soft-fail when we hit an error from edx/openedx. The dashboard should load, although the data will be stale, but that's ok I think since right now we're just failing _hard_. We should no longer logout the learner and redirect them to the login page.

If it's a 4xx error the UI should display a dialog similar to `SocialAuthDialog` that informs them they need to reauthenticate. This probably means the dashboard API needs to return some additional information in terms of which API auths are failing. This dialog should be dismissable  so the learner can proceed without fixing it should they wish.

rhysyngsun:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Fixes #5327 

#### What's this PR do?
This updates the dashboard logic to not cause a user to get in a circular login/logout loop if their MITx online authentication is no longer valid. It instead prompts them to reauthenticate with the correct provider.

For both MITx Online and EdX
- Authenticate with the social auth for that provider (`/login/mitxonline` and `/login/edxorg`)
- Go to the dashboard and verify you don't see any messages or alerts.

---

- Update the social auth record for that provider, changing the `extra_data` to have incorrect values for the refresh token and access token, this will simulate an expired or otherwise invalid token.
- Go to the dashboard and verify you get the popup dialog prompting you to login.
- You should be able to opt out of this and dismiss the dialog without authenticating and continue to use the dashboard.
- You should also be able to reload the page, which will prompt you again, and then successfully reauthenticate.

---

- In `dashboard/api.py`, put a `raise Exception("error")` on the first line of `update_cache_for_backend`. 
- Verify when you load the dashboard you still get an error message indicating the data is out of date.

(Optional)

Screenshot 2023-04-20 at 16-38-23 Learner Dashboard MITx MicroMasters
Screenshot 2023-04-20 at 16-37-17 Learner Dashboard MITx MicroMasters

If edx or mitxonline are unreachable regardless of social auth validity, you should see this:

Screenshot 2023-04-20 at 09-51-56 Learner Dashboard MITx MicroMasters

### DIFF
--- a/dashboard/api.py
+++ b/dashboard/api.py
@@ -15,7 +15,7 @@ from django.urls import reverse
 from django_redis import get_redis_connection
 from edx_api.client import EdxApi
 
-from backends.constants import COURSEWARE_BACKEND_URL, BACKEND_EDX_ORG, BACKEND_MITX_ONLINE
+from backends.constants import COURSEWARE_BACKEND_URL, COURSEWARE_BACKENDS
 from backends.exceptions import InvalidCredentialStored
 from backends import utils
 from courses.models import Program, ElectiveCourse, CourseRun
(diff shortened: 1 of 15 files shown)


annagav (approved):
Works for me 👍

PR #5332, opened 2023-04-18
Fix deployments due to conflict with a few library versions.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
This fixes an issue trying to build the app for heroku deployment. See the trace in the build log:

https://github.com/mitodl/micromasters/actions/runs/4699543907/jobs/8400688419#step:3:29935

The issue here is that heroku appears to have an pre-installed version of `setuptools`. There must've been a recent update that caused this to be upgraded to a version incompatible with how `l18n` invokes it. So I've pinned both `setuptools` (to the version current for Heroku) and `l18n`.

(Required)

- On `master` branch, add `setuptools==67.6.1` to `requirements.in` and run `pip-compile`. This should error with the same error on the failed build linked above.
- Checkout this branch, rebuild your containers, and smoke test that the app still runs.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -24,6 +24,7 @@ opensearch-dsl==1.0.0
 factory_boy
 faker
 html5lib==0.999999999
+l18n==2021.3
 ipython
 jsonfield==2.0.2
 jsonpatch==1.16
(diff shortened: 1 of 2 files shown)


annagav (approved):
👍

PR #3711, opened 2017-11-28
Release 0.82.0

annagav:
## George Schneeloch
  - [x] Propagate 409 response from open-discussions when creating a channel (#3708) ([8fc8c524](../commit/8fc8c524a80231e9137801502a32e6d1005805af))

PR #5370, opened 2024-03-05
api endpoint for program page config

shanbady:

Closes [#3711](https://github.com/mitodl/hq/issues/3711)

This PR has changes that integrate wagtail built-in api endpoints so we can retrieve content used in program letters from mit-open.

### How can this be tested?
1. Checkout this branch.
2. If you dont already have some mock data, create a ProgramPage via /cms - make sure it is not sitting in the 'root' (should be a child of the "Micromasters Home" page. Fill out the required fields in addition to the fields starting with "program_letter_*"
3. Save and publish the page
4. navigate to the wagtail endpoint and specify the pagetype and wildcard for the fields http://localhost:8079/api/v0/wagtail/pages/?type=cms.ProgramPage&fields=*
5. See that all the program letter fields are shown (inckuding the list of signatories) in addition to links for the signature images

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -1,3 +1,6 @@
+"""
+Integrate Wagtail's builtin api endpointss
+"""
 from wagtail.api.v2.views import PagesAPIViewSet
 from wagtail.api.v2.router import WagtailAPIRouter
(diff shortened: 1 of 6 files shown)


annagav (approved):
LGTM 👍

PR #5382, opened 2024-04-11
Test should not try to refresh token

annagav:
Related to #3878

Updates a test not call edx api and patch the call instead.
The test was not working properly because we didn't patch ExamAuthorization to return no object.
This test is checking to see that a user is authorized before we try enroll them for an exam run.

### How can this be tested?
Tests should pass.


shanbady:
there is a js failure. dont think its related to this change though

PR #5389, opened 2024-05-01
Removing sign in buttons from DEDP page

annagav:
Fix https://github.com/mitodl/hq/issues/3936

Removes the sign in buttons from the DEDP program page. 
I used a hardcoded id for the program since it is never going to change. I didn't see much value in introducing a constant. We are moving away from the micromsters site.
After:

Screenshot 2024-05-01 at 5 15 05 PM

Before:

Screenshot 2024-05-01 at 5 16 57 PM

### How can this be tested?
Make sure other pages are unaffected.

### DIFF
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -64,7 +64,7 @@
                   target="blank">Learn More</a>
         {% else %}
             {% if page.program.id == 2 %}
-                <a href="https://mitxonline.mit.edu/signin/" class="mdl-button hero-button mdl-cell--hide-phone">
+                <a href="https://mitxonline.mit.edu/create-account/" class="mdl-button hero-button mdl-cell--hide-phone">
                   Enroll on MITx Online
                 </a>
             {% else %}
--- a/cms/templates/cms/program_page.html
+++ b/cms/templates/cms/program_page.html
@@ -64,7 +64,7 @@
                   target="blank">Learn More</a>
         {% else %}
             {% if page.program.id == 2 %}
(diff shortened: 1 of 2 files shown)


cms/templates/cms/program_page.html
```diff
@@ -63,17 +63,23 @@ <h1 class="title">
...
-            </a>
-            <div class="log-in mdl-cell--hide-phone">
-              Already a Member?
-              <a href="{% url 'signin' %}?program={{page.program.id}}">
-                Sign In
-              </a>
-            </div>
-          {% endif %}
+            {% if page.program.id == 2 %}
+                <a href="https://mitxonline.mit.edu/" class="mdl-button hero-button mdl-cell--hide-phone">
```
pdpinch, line 67: What do you think is the best URL to use for this? Could we go directly to the create-account page?

cms/templates/cms/program_page.html
```diff
@@ -63,17 +63,23 @@ <h1 class="title">
...
-            </a>
-            <div class="log-in mdl-cell--hide-phone">
-              Already a Member?
-              <a href="{% url 'signin' %}?program={{page.program.id}}">
-                Sign In
-              </a>
-            </div>
-          {% endif %}
+            {% if page.program.id == 2 %}
+                <a href="https://mitxonline.mit.edu/create-account/" class="mdl-button hero-button mdl-cell--hide-phone">
```
shanbady, line 67: not sure how this works locally but could is it possible to use a reverse here?

shanbady (changes requested):

shanbady (approved):
looks good. left a minor optional comment

PR #5391, opened 2024-05-16
make "MITx Online" mixed case

pdpinch:

see https://github.com/mitodl/hq/issues/3936#issuecomment-2103361040

"ENROLL ON MITX ONLINE" => "ENROLL ON MITx ONLINE"


pdpinch:
@annagav let me know if this is too hacky. We don't necessarily have to do it.

PR #5399, opened 2024-09-24
Add warning message to DEDP dashboard

annagav:
Fix https://github.com/mitodl/hq/issues/5519

Adds a warning message.

The program records link was shown only to programs that had exams (aka DEDP) and can be turned off through a flag `FEATURE_PROGRAM_RECORD_LINK`.

I decided not to disable the course rows, because there is no downside to having it enabled. And I didn't want to have to dismantle all JS tests that we have.
I also removed the FA calculation card that appeared before the course rows.

Screenshot 2024-09-25 at 4 11 55 PM


rachellougee (approved):
I didn't setup MM to test locally but the code looks good to me.

pdpinch:
Since we're working on this, let's apply this warning message to _all_ programs. We just need to change the link for edx.org programs to go to edx.org instead of MITx Online.

rachellougee (commented):
I was able to test this locally. The message looks good on DEDP program id=2. You may want to update the message for all program per Peter's comment.

 One issue I noticed that there is a trailing space in the modified file `FinancialAidCard_test.js ` https://github.com/mitodl/micromasters/pull/5399/files#diff-ad8eb2397d0a3bb35daab569f07d3d3dca98ea6b2ed090ad5eb79eb9446f2a88. Not sure why the test didn't warn that.

pdpinch:
@pdpinch what should we link users to in mitxonline and in edx.org? 

for edx.org, we can link them to the program dashboard, https://courses.edx.org/dashboard/programs/

for mitxonline, we can link them to the dashboard, https://mitxonline.mit.edu/dashboard 

for mitxonline users without a socialauth, we should tell them to create an account with the same email address. We can display the email address we have on record for them.

annagav:
Screenshot 2024-10-02 at 10 41 57 AM

annagav:
I think we should include their email address for everyone, in case the forgot which email they registered with.

rachellougee (approved):
The message looks good to me based on the user's social auth.

static/js/components/dashboard/CourseListCard.js
```diff
@@ -39,22 +42,30 @@ export default class CourseListCard extends React.Component {
...
-        To get credit for the courses in this program, you must pay for a
-        verified certificate from {programBackendName(program)}. If you want to
-        audit courses for FREE and upgrade later, click Enroll then choose the
-        audit option.
-      </p>
+      <div className="callout callout-warning">
+        <img src="/static/images/c-warning-1.svg" alt="Warning" />
+        <div>
+          The following course, enrollment, and grade information is outdated.
+          <br />
```
pdpinch, line 55: Is this line break necessary here? I think the message would flow better in one single paragraph.

pdpinch:
I thought there would be an overlay over the dashboard itself, to make it look "greyed out" and to prevent clicking on it.

static/js/components/dashboard/CourseListCard.js
annagav: The break was in the design, but I can remove it.

PR #5430, opened 2025-08-21
update docker python & nginx images

cp-at-mit:
NA

Previously, `docker compose build` would fail since the python docker image being used in this project is no longer supported.  This PR updates the python docker image, as well as the nginx docker image which was throwing an error for being out of date.

### How can this be tested?
`docker compose build` should run without any errors or failure.


annagav (approved):
Works 👍

PR #5443, opened 2026-01-05
chore: upgrade django to v52 and wagtail to v7

Anas12091101:
https://github.com/mitodl/hq/issues/5344

This PR upgrades django to v52 and wagtail to v7.2.1

- [ ] Desktop screenshots

### DIFF
--- a/micromasters/tests/test_settings.py
+++ b/micromasters/tests/test_settings.py
@@ -45,8 +45,9 @@ class TestSettings(MockedESTestCase):
             'MICROMASTERS_USE_S3': 'False'
         }, clear=True):
             settings_vars = self.reload_settings()
+            storages = settings_vars.get('STORAGES', {})
             self.assertNotEqual(
-                settings_vars.get('DEFAULT_FILE_STORAGE'),
+                storages.get('default', {}).get('BACKEND'),
                 'storages.backends.s3boto3.S3Boto3Storage'
             )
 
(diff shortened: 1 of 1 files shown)


micromasters/urls.py
```diff
@@ -1,52 +1,50 @@
...
-    url('', include('courses.urls')),
-    url("", include("cms.urls")),
-    url('', include('dashboard.urls')),
-    url('', include('ecommerce.urls')),
-    url('', include('financialaid.urls')),
-    url('', include('search.urls')),
-    url('', include('mail.urls')),
-    url('', include('profiles.urls')),
-    url('', include('discussions.urls')),
-    url(r'^status/', include('server_status.urls')),
```
Anas12091101: Since https://github.com/mitodl/django-server-status has been archived and we are no longer maintaining it, I have removed it here in this PR, as it was causing some dependency resolution issues

annagav:
I got this:
``` 
2026-01-14 15:17:48.015 UTC [1] FATAL:  database files are incompatible with server

2026-01-14 15:17:48.015 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 12, which is not compatible with this version 16.11 (Debian 16.11-1.pgdg13+1).
```

Anas12091101:
This is due to the Postgres version being upgraded to v16 in this PR. To resolve the issue, you will need to back up your local data, drop the database service, bring it up again, and then restore the data.

annagav (approved):
👍

PR #5501, opened 2026-04-08
Add scheduled task to sync course runs

annagav:

N/A 

Add scheduled task to sync course runs daily with learn

### How can this be tested?
Make sure the task is syncing course runs for existing courses correctly.

### DIFF
--- a/courses/management/commands/import_course_runs_for_course.py
+++ b/courses/management/commands/import_course_runs_for_course.py
@@ -45,6 +45,7 @@ class Command(BaseCommand):
             self.stdout.write(
                 self.style.ERROR(f"Course {course_id} not found in MIT Learn API.")
             )
+            return None
 
         try:
             course = Course.objects.get(edx_key=course_id)
--- a/courses/management/commands/import_course_runs_for_course.py
+++ b/courses/management/commands/import_course_runs_for_course.py
@@ -59,9 +59,16 @@ class Command(BaseCommand):
                 f"Updating course runs for {course_id}: {course.title}"
             )
         )
-        enrollment_url = raw_course.get("enrollment_url", "")
+        if not raw_course:
+            self.stdout.write(
+                self.style.ERROR(f"Course {course_id} not found.")
+            )
+            return None
 
-        course_runs_created = sync_mit_learn_courseruns_for_course(course, enrollment_url, raw_courseruns=raw_course["runs"])
-        self.stdout.write(self.style.SUCCESS(f"{course_runs_created} course runs created"))
+        course_runs_created = sync_mit_learn_courseruns_for_course(course, raw_course)
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"{course_runs_created} course runs created")
+        )
 
         return None
--- a/courses/management/commands/import_course_runs_for_course.py
+++ b/courses/management/commands/import_course_runs_for_course.py
@@ -36,6 +36,7 @@ class Command(BaseCommand):
             self.stdout.write(
                 self.style.ERROR("You must provide a course_id to import all the associated course runs.")
             )
+            return None
         self.stdout.write(
                 self.style.SUCCESS(f"Fetching course run data from MIT Learn API for course {course_id}")
             )
--- a/courses/management/commands/import_course_runs_for_course.py
+++ b/courses/management/commands/import_course_runs_for_course.py
@@ -41,7 +41,7 @@ class Command(BaseCommand):
             )
         raw_course = fetch_course_from_mit_learn(course_id)
 
-        if raw_course is None:
+        if not raw_course:
             self.stdout.write(
                 self.style.ERROR(f"Course {course_id} not found in MIT Learn API.")
             )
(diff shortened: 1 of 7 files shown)


grades/tasks.py
```diff
@@ -139,3 +140,15 @@ def freeze_users_final_grade_async(user_ids, course_run_id):
...
+
+def sync_course_run_info_from_learn():
+    """
+    Sync course run info from the learn API.
+    Exclude Finance courses
+    """
+    for course in Course.objects.exclude(program__title="Finance"):
+        course_id = course.edx_key
+        raw_course = fetch_course_from_mit_learn(course_id)
+        sync_mit_learn_courseruns_for_course(course, raw_course)
```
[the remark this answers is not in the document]
annagav: there are only 37 courses

courses/mit_learn_api.py
```diff
@@ -49,40 +52,60 @@ def fetch_course_from_mit_learn(course_id) -> dict[str, Any]:
...
+                "enrollment_end") else None,
             "start_date": parse_datetime(raw_courserun.get("start_date")) if raw_courserun.get("start_date") else None,
-            "end_date": parse_datetime(raw_courserun.get("end_date"))if raw_courserun.get("end_date") else None,
-            "upgrade_deadline": parse_datetime(raw_courserun.get("upgrade_deadline")) if raw_courserun.get("upgrade_deadline") else None,
-            "courseware_backend": raw_courserun.get("courseware_backend", ""),
-            "enrollment_url": enrollment_url,
+            "end_date": parse_datetime(raw_courserun.get("end_date")) if raw_courserun.get("end_date") else None,
+            "upgrade_deadline": parse_datetime(raw_courserun.get("upgrade_deadline")) if raw_courserun.get(
+                "upgrade_deadline") else None,
+            "courseware_backend": "edxorg" if platform_code == "edx" else "mitxonline",
```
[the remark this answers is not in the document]
annagav: not a big deal, but will add validation

courses/management/commands/import_course_runs_for_course.py
```diff
@@ -59,9 +59,16 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
                 f"Updating course runs for {course_id}: {course.title}"
             )
         )
-        enrollment_url = raw_course.get("enrollment_url", "")
+        if not raw_course:
+            self.stdout.write(
+                self.style.ERROR(f"Course {course_id} not found.")
+            )
+            return None
```
Anas12091101: We’re already performing this check on L44. Also, the condition there should be `if not raw_course:` rather than `if raw_course is None`, since `fetch_course_from_mit_learn` returns an empty dict when the course isn’t found.

grades/tasks.py
```diff
@@ -139,3 +142,20 @@ def freeze_users_final_grade_async(user_ids, course_run_id):
...
+        if not course.edx_key:
+            continue
+        course_id = course.edx_key
+        try:
+            raw_course = fetch_course_from_mit_learn(course_id)
+            if not raw_course:
+                continue
+            sync_mit_learn_courseruns_for_course(course, raw_course)
+        except MITLearnAPIError:
+            log.exception('Error syncing MIT Learn course data for course "%s"', course_id)
```
Anas12091101: It doesn’t make sense for this task to live in the grades module. Instead, we should create a `tasks.py` file within the courses module and place this task there.

grades/tasks.py
```diff
@@ -139,3 +142,20 @@ def freeze_users_final_grade_async(user_ids, course_run_id):
                 'Impossible to freeze final grade for user "%s" in course %s',
                 user.username, course_run.edx_course_key
             )
+
+@app.task
+def sync_course_run_info_from_learn():
```
Anas12091101: We should add tests for this task.

Anas12091101 (changes requested):
Some changes

courses/mit_learn_api.py
```diff
@@ -1,20 +1,56 @@
...
     """Custom exception for MIT Learn API errors."""
 
+
 LEARN_API_COURSES_LIST_URL = "https://api.learn.mit.edu/api/v1/courses/"
 
+EDX_PLATFORM_CODES = {"edx", "edxorg"}
+MITXONLINE_PLATFORM_CODES = {"mitxonline"}
+
+
+def get_courseware_backend(platform_code, course_key):
```
Anas12091101, line 27: Can we add tests for this function as well?

courses/mit_learn_api.py
```diff
@@ -49,40 +85,63 @@ def fetch_course_from_mit_learn(course_id) -> dict[str, Any]:
 
 
 
-
-
-
-
-def sync_mit_learn_courseruns_for_course(course, enrollment_url, raw_courseruns: list[dict[str, Any]]) -> list[dict[str, Any]]:
+def sync_mit_learn_courseruns_for_course(course, raw_course) -> int:
```
Anas12091101, line 88: I just noticed that this function is also missing tests. Could we add them, please?

Anas12091101 (changes requested):
Thanks for making the changes. Just a couple of requests:

Anas12091101 (approved):
Thanks for making all the changes.. LGTM 👍

PR #5503, opened 2026-04-13
Update PUBLIC POLICY ELECTIVES requirement

annagav:
Fix https://github.com/mitodl/hq/issues/10403

add clarifying language to the header of the Public Policy Elective section after (Complete 2 of 4).
### How can this be tested?

Have a program setup with an elective set or multiple.
Name the elective set `title` "PUBLIC POLICY ELECTIVES"
View the product page.
Screenshot 2026-04-13 at 6 43 15 PM

### DIFF
--- a/static/js/components/CourseListWithPopover.js
+++ b/static/js/components/CourseListWithPopover.js
@@ -6,6 +6,7 @@ import type {
   ProgramPageCourse,
   ProgramPageElectiveSet
 } from "../flow/programTypes"
+import { EXTRA_ELECTIVE_REQUIREMENT_TITLES } from "../constants"
 
 const listItems = courses => {
   return courses.map((course, index) => (
@@ -22,7 +23,9 @@ function populateElectivesList(electiveSet) {
         <h2 className="required-completion">
           (Complete {electiveSet.required_number} of{" "}
           {electiveSet.courses.length}
-          {electiveSet.title === "PUBLIC POLICY ELECTIVES"
+          {EXTRA_ELECTIVE_REQUIREMENT_TITLES.includes(
+            electiveSet.title.toLowerCase()
+          )
             ? ", one elective must be advanced"
             : ""}
           )
--- a/static/js/components/CourseListWithPopover.js
+++ b/static/js/components/CourseListWithPopover.js
@@ -21,7 +21,8 @@ function populateElectivesList(electiveSet) {
         <h3 className="elective">{electiveSet.title}</h3>
         <h2 className="required-completion">
           (Complete {electiveSet.required_number} of{" "}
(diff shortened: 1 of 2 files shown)


static/js/components/CourseListWithPopover.js
```diff
@@ -21,7 +21,11 @@ function populateElectivesList(electiveSet) {
         <h3 className="elective">{electiveSet.title}</h3>
         <h2 className="required-completion">
           (Complete {electiveSet.required_number} of{" "}
-          {electiveSet.courses.length})
+          {electiveSet.courses.length}
+          {electiveSet.title === "PUBLIC POLICY ELECTIVES"
+            ? ", one elective must be advanced"
+            : ""}
+          )
```
Anas12091101, line 31: A cleaner approach would be to define a setting (e.g., `ADVANCED_ELECTIVE_REQUIRED_TITLES`) that contains a list like `["public policy electives"]`, and then check against it here:

```suggestion
          {electiveSet.courses.length}
          {electiveSet.title.lower() in ADVANCED_ELECTIVE_REQUIRED_TITLES
            ? ", one elective must be advanced"
            : ""}
          )
```

This keeps the logic configurable and avoids hardcoding specific titles in the code.

Anas12091101 (changes requested):

static/js/components/CourseListWithPopover.js
```diff
@@ -21,7 +22,13 @@ function populateElectivesList(electiveSet) {
         <h3 className="elective">{electiveSet.title}</h3>
         <h2 className="required-completion">
           (Complete {electiveSet.required_number} of{" "}
-          {electiveSet.courses.length})
+          {electiveSet.courses.length}
+          {EXTRA_ELECTIVE_REQUIREMENT_TITLES.includes(
+            electiveSet.title.toLowerCase()
+          )
```
Anas12091101, line 28: Just a small safety check to ensure we don’t hit errors if `electiveSet.title` is null.

```suggestion
          {EXTRA_ELECTIVE_REQUIREMENT_TITLES.includes(
            (electiveSet.title || "").toLowerCase()
          )
```

Anas12091101 (changes requested):

static/js/components/CourseListWithPopover.js
annagav: The title can't be None as defined by 
```
export type ProgramPageElectiveSet = {
  required_number:              number,
  title:                        string,
  courses:                      Array<ProgramPageCourse>,
}
```
Title is a required field on the model as well

Anas12091101 (approved):
Thanks for making all the changes, LGTM 👍

PR #5518, opened 2026-06-03
Update DEDP and SCM to Login to MIT Learn

annagav:
Fix https://github.com/mitodl/hq/issues/11606

Remove login buttons.
Link to MIT Learn for DEDP and SCM

This is hardcoded to the program ids since those will never change.

Screenshot 2026-06-03 at 1 48 40 PM
Screenshot 2026-06-03 at 1 52 34 PM


rhysyngsun (approved):
LGTM

